PR #2386, opened 2021-07-16
Added reporting of why Plutus scripts fail.

TimSheard:
Exended  UtxosPredicateFailure, by adding a Text field to its constructor  ValidationTagMismatch.
Altered the functions runPLCScript  and  evalScripts return a ScriptResult rather than a Bool.

data ScriptResult = Passes | Fails ![String]

So when evalScripts fails in the Utxos rule, we add the reason why to ValidationTagMismatch.
Right now the text string is a rather crude concatenation of a bunch of show results. In the next step we will construct the Text object by using the PrettyA class, so the result is more informative.

### DIFF
--- a/alonzo/impl/src/Cardano/Ledger/Alonzo/Rules/Utxos.hs
+++ b/alonzo/impl/src/Cardano/Ledger/Alonzo/Rules/Utxos.hs
@@ -34,6 +34,7 @@ import Cardano.Ledger.Alonzo.Tx
     txouts,
   )
 import qualified Cardano.Ledger.Alonzo.TxBody as Alonzo
+import Cardano.Ledger.Alonzo.TxInfo (ScriptResult (..))
 import qualified Cardano.Ledger.Alonzo.TxWitness as Alonzo
 import Cardano.Ledger.BaseTypes
   ( Globals,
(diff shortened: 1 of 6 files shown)


JaredCorduan (approved):
Looks great to me!

JaredCorduan:
@TimSheard - this is not a WIP anymore, is it?

alonzo/impl/src/Cardano/Ledger/Alonzo/Rules/Utxos.hs
```diff
@@ -232,8 +236,8 @@ scriptsNotValidateTransition = do
 data UtxosPredicateFailure era
   = -- | The 'isValidating' tag on the transaction is incorrect. The tag given
     --   here is that provided on the transaction (whereas evaluation of the
-    --   scripts gives the opposite.)
-    ValidationTagMismatch IsValidating
+    --   scripts gives the opposite.). The Text tries to explain why it failed.
+    ValidationTagMismatch IsValidating Text
```
nc6: Rather than packing in using `unlines`, why not have an array here? Allows more choice for pretty-printers down the line

alonzo/impl/src/Cardano/Ledger/Alonzo/TxInfo.hs
```diff
@@ -345,21 +345,30 @@ valContext ::
   Data era
 valContext txinfo sp = Data (P.toData (P.ScriptContext txinfo (transScriptPurpose sp)))
 
+data ScriptResult = Passes | Fails ![String]
+
+andResult :: ScriptResult -> ScriptResult -> ScriptResult
```
nc6, line 365: Could just add a `Semigroup` instance instead. Then we get other combining functions for free

nc6 (approved):
Made a couple of suggestions, but in general looks good. I'm assuming no longer WIP

alonzo/impl/src/Cardano/Ledger/Alonzo/TxInfo.hs
TimSheard: good idea, done.

PR #2390, opened 2021-07-20
unit test for the size of an ada-only utxo

JaredCorduan:
This number is helpful for comparing the protocol parameter `utxoCostPerWord` with the old parameter `minUTxOValue`.


polinavino (approved):
LGTM!

PR #2391, opened 2021-07-21
Update cardano-base and derive Measure for ExUnits

nfrisby:
`cardano-base` now defines the `Measure` class and the Consensus Layer needs `ExUnits` to instantiate it. This PR does so.

### DIFF
--- a/alonzo/impl/src/Cardano/Ledger/Alonzo/Scripts.hs
+++ b/alonzo/impl/src/Cardano/Ledger/Alonzo/Scripts.hs
@@ -69,7 +69,9 @@ import Cardano.Ledger.ShelleyMA.Timelocks
 import Control.DeepSeq (NFData (..))
 import Data.ByteString.Short (ShortByteString, fromShort)
 import Data.Coders
+import Data.DerivingVia (InstantiatedAt (..))
 import Data.Map (Map)
+import Data.Measure (BoundedMeasure, Measure)
 import Data.Text (Text)
 import Data.Typeable
 import Data.Word (Word64, Word8)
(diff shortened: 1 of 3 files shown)


alonzo/impl/src/Cardano/Ledger/Alonzo/Scripts.hs
```diff
@@ -134,17 +136,17 @@ data ExUnits = ExUnits
     exUnitsSteps :: !Word64
   }
   deriving (Eq, Generic, Show) -- It is deliberate that there is NO ORD instance.
+  deriving
+    (BoundedMeasure, Measure)
+    via (InstantiatedAt Generic ExUnits)
+  deriving
+    (Monoid, Semigroup)
+    via (InstantiatedAt Measure ExUnits)
```
nfrisby, line 144: These links are for context (ie I'm pretending to be the `haddock` exe here).

https://github.com/input-output-hk/cardano-base/blob/8c732560b201b5da8e3bdf175c6eda73a32d64bc/base-deriving-via/src/Data/DerivingVia.hs#L27-L47

https://github.com/input-output-hk/cardano-base/blob/8c732560b201b5da8e3bdf175c6eda73a32d64bc/measures/src/Data/Measure/Class.hs#L80

https://github.com/input-output-hk/cardano-base/blob/8c732560b201b5da8e3bdf175c6eda73a32d64bc/measures/src/Data/Measure/Class.hs#L155-L160

https://github.com/input-output-hk/cardano-base/blob/8c732560b201b5da8e3bdf175c6eda73a32d64bc/measures/src/Data/Measure/Class.hs#L187-L193

nc6 (approved):
LGTM

PR #2393, opened 2021-07-23
redeemers check mistake

polinavino:
as @WhatisRT points out, the current check for extra redeemers allowed redeemers for non-native scripts : (
this fixes that

### DIFF
--- a/alonzo/impl/src/Cardano/Ledger/Alonzo/Rules/Utxow.hs
+++ b/alonzo/impl/src/Cardano/Ledger/Alonzo/Rules/Utxow.hs
@@ -18,11 +18,7 @@ import Cardano.Ledger.Address (Addr (..), bootstrapKeyHash, getRwdCred)
 import Cardano.Ledger.Alonzo.Data (DataHash)
 import Cardano.Ledger.Alonzo.Language (Language (..))
 import Cardano.Ledger.Alonzo.PParams (PParams)
-import Cardano.Ledger.Alonzo.PlutusScriptApi
-  ( checkScriptData,
-    language,
-    scriptsNeeded,
-  )
+import Cardano.Ledger.Alonzo.PlutusScriptApi (language, scriptsNeeded)
 import Cardano.Ledger.Alonzo.Rules.Utxo (AlonzoUTXO)
 import qualified Cardano.Ledger.Alonzo.Rules.Utxo as Alonzo (UtxoPredicateFailure)
 import Cardano.Ledger.Alonzo.Scripts (Script (..))
@@ -50,7 +46,6 @@ import Cardano.Ledger.Credential (Credential (KeyHashObj))
 import Cardano.Ledger.Era (Crypto, Era, ValidateScript (..))
 import Cardano.Ledger.Keys (GenDelegs, KeyHash, KeyRole (..), asWitness)
 import Cardano.Ledger.Rules.ValidationMode ((?!#))
-import Control.DeepSeq (NFData (..))
 import Control.Iterate.SetAlgebra (domain, eval, (⊆), (◁), (➖))
 import Control.State.Transition.Extended
 import Data.Coders
@@ -100,7 +95,7 @@ import Shelley.Spec.Ledger.UTxO (UTxO (..), txinLookup)
 --   failure type of the Shelley Era, as they share some failure modes.
 data AlonzoPredFail era
   = WrappedShelleyEraFailure !(UtxowPredicateFailure era)
-  | UnRedeemableScripts ![(ScriptPurpose (Crypto era), ScriptHash (Crypto era))]
+  | MissingRedeemers ![(ScriptPurpose (Crypto era), ScriptHash (Crypto era))]
   | MissingRequiredDatums
       !(Set (DataHash (Crypto era))) -- Set of missing data hashes
       !(Set (DataHash (Crypto era))) -- Set of received data hashes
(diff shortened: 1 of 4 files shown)


JaredCorduan (dismissed):
good catch!

JaredCorduan:
The unit test for "missing 2-phase script witness" is now also returning an addition predicate failure, namely:
```
ExtraRedeemers [RdmrPtr Mint 1,RdmrPtr Cert 1,RdmrPtr Rewrd 0]
```
Something seems wrong...

polinavino:
those were redeemers for phase-1 scripts, there is only 1 phase-2 script (used for 1 purpose, so needs 1 redeemer, and the transaction has a total of 4)

polinavino:
The example failure is fixed.. but now I do not understand why this is happening.. is it something to do with recent changes?:

@nc6 it looks like the same error as in the latest merged PR build?

`[6 of 9] Compiling Test.Cardano.Ledger.Alonzo.Tools ( test/Test/Cardano/Ledger/Alonzo/Tools.hs, dist/build/cardano-ledger-alonzo-test/cardano-ledger-alonzo-test-tmp/Test/Cardano/Ledger/Alonzo/Tools.o, dist/build/cardano-ledger-alonzo-test/cardano-ledger-alonzo-test-tmp/Test/Cardano/Ledger/Alonzo/Tools.dyn_o )

test/Test/Cardano/Ledger/Alonzo/Tools.hs:22:1: error:
    Could not find module ‘Cardano.Ledger.Tx’
    Perhaps you meant
      Cardano.Ledger.Era (from cardano-ledger-core-0.1.0.0)
      Cardano.Ledger.Val (from cardano-ledger-core-0.1.0.0)
      Cardano.Ledger.Coin (from cardano-ledger-core-0.1.0.0)
    Use -v (or `:set -v` in ghci) to see a list of the files searched for.
   |
22 | import qualified Cardano.Ledger.Tx as Core`

nc6:
Ah, maybe suggests an out-of-order merge issue. I'll fix separately.

WhatisRT:
I think we can cut down on the complexity now, if we really don't want the size check. `checkScriptData` can simply check if all the required scripts are present (and no redeemer checks, so the name should probably be changed) and we can simply do all the checks on the redeemers in this set comparison. I'll push a change to the spec that does this to this PR

WhatisRT:
I made the changes to the formal spec we just discussed

WhatisRT:
These changes also made it obvious that we've been checking that all the scripts are present twice, so I've removed that duplicate as well. UTXOW looks a lot simpler now :)

polinavino:
Thank you!!

JaredCorduan (dismissed):
LGTM!

cardano-ledger-test/src/Test/Cardano/Ledger/Examples/TwoPhaseValidation.hs
```diff
@@ -1583,6 +1583,12 @@ alonzoUTXOWexamples =
                       ],
                     WrappedShelleyEraFailure . MissingScriptWitnessesUTXOW . Set.singleton $
                       (alwaysSucceedsHash 2 pf),
+                    -- these redeemers are associated with phase-1 scripts
+                    ExtraRedeemers
+                      [ RdmrPtr Tag.Mint 1,
+                        RdmrPtr Tag.Cert 1,
+                        RdmrPtr Tag.Rewrd 0
```
JaredCorduan, line 1588: So this predicate failure is happening because the lack of script witness means that we do not know that the script hash is a plutus script, and hence these look superfluous?

alonzo/impl/src/Cardano/Ledger/Alonzo/Rules/Utxow.hs
```diff
@@ -310,8 +310,17 @@ alonzoStyleWitness = do
          in seq (rnf ans) ans
   null unredeemed ?! UnRedeemableScripts unredeemed
 
-  {-  || { (sp, h) ∈ \fun{scriptsNeeded} utxo tx | h\mapsto s ∈ txscripts txw, s ∈ ScriptPhTwo } || = || fun{txrdmrs} tx ||  -}
-  let rdptrs = Set.fromList [el | (sp, _) <- sphs, SJust el <- [rdptr @era txbody sp]]
+  {-  ∀ (ptr \mapsto _ ) ∈ txrdmrs tx, ptr ∈ { rdptr txb sp | (sp, h) ∈ scriptsNeeded utxo tx,
+    h \mapsto s ∈ txscripts txw, s ∈ Scriptph2}  -}
```
WhatisRT: This comment is outdated, we want those sets to be equal (cf. the changes to the spec in this PR)

alonzo/impl/src/Cardano/Ledger/Alonzo/Rules/Utxow.hs
```diff
@@ -310,8 +310,17 @@ alonzoStyleWitness = do
...
+  let rdptrs =
+        Set.fromList
+          [ el
+            | (sp, hs) <- sphs,
+              SJust el <- [rdptr @era txbody sp],
+              (hs', script) <- Map.toList (getField @"scriptWits" tx),
```
WhatisRT: Seems like a lookup of `hs` in the map would be nicer here?

alonzo/impl/src/Cardano/Ledger/Alonzo/Rules/Utxow.hs
```diff
@@ -310,8 +310,17 @@ alonzoStyleWitness = do
...
+              hs == hs',
+              (not . isNativeScript @era) script
+          ]
       extraRdmrs :: [RdmrPtr]
       extraRdmrs = Map.keys $ Map.withoutKeys (unRedeemers $ txrdmrs $ wits tx) rdptrs
```
WhatisRT: This is only one of the two inclusions required. It seems like the other one is handled by the previous block using `checkScriptData`, but we removed that function from the spec entirely. So instead, there should be a `missingRdmrs` that's defined the same way as `extraRdmrs`, except that the two sets here are flipped.

WhatisRT (commented):
I think what's in here is technically equivalent to what's in the spec but it's not obvious. If someone else agrees, we could merge it and fix it later (probably as part of our current ongoing audit)

cardano-ledger-test/src/Test/Cardano/Ledger/Examples/TwoPhaseValidation.hs
polinavino: yes, there is no plutus/phase-2 script validating those three things

JaredCorduan (approved):
LGTM

WhatisRT:
I think we can merge this now, I'd probably like Polina to have another look at this but we probably don't want to wait till she's back

PR #2396, opened 2021-07-26
rename IsValidating to IsValid

JaredCorduan:


alonzo/impl/src/Cardano/Ledger/Alonzo/Tx.hs
```diff
@@ -161,16 +165,23 @@ import qualified Shelley.Spec.Ledger.UTxO as Shelley
 
 -- ===================================================
 
+{-# DEPRECATED IsValidating "Use IsValid instead" #-}
+type IsValidating = IsValid
+
+{-# DEPRECATED isValidating "Use isValid instead" #-}
+isValidating :: ValidatedTx era -> IsValid
+isValidating = isValid
```
nc6, line 173: For completeness we should also have a deprecated `HasField` instance.

nc6 (approved):
One minor comment

nc6 (approved):
One minor comment

alonzo/impl/src/Cardano/Ledger/Alonzo/Tx.hs
JaredCorduan: great catch, thank you!

JaredCorduan: @nc6 is it possible to deprecate a class instance? It doesn't seem like it to me. I can add the instance, and leave a TODO to remove it when we remove the other deprecated functions.

PR #2403, opened 2021-07-30
move defaultCostModel to tests

JaredCorduan:
The Plutus default cost model should only be used in testing, since the value set in the protocol parameters is the only source of truth. For this reason, I am moving it to the test package (and deprecating the existing one).


nc6 (approved):
Not sure if we need the deprecation here, but doesn't hurt!

PR #2405, opened 2021-08-02
Expose event logging downstream

nc6:
This PR exposes the event logging mechanism for use downstream, as well as making some slight changes to its interface.

No changes should be visible to clients using the existing interfaces.

### DIFF
--- a/semantics/executable-spec/src/Control/State/Transition/Extended.hs
+++ b/semantics/executable-spec/src/Control/State/Transition/Extended.hs
@@ -50,6 +50,7 @@ module Control.State.Transition.Extended
     liftSTS,
     tellEvent,
     tellEvents,
+    mapEventReturn,
 
     -- * Apply STS
     AssertionPolicy (..),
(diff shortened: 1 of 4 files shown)


goolord (approved):
lgtm

semantics/executable-spec/src/Control/State/Transition/Extended.hs
```diff
@@ -253,6 +256,15 @@ type family EventReturnType ep sts a :: Type where
   EventReturnType 'EventPolicyReturn sts a = (a, [Event sts])
   EventReturnType _ _ a = a
 
+class EventReturnTypeRep ert where
```
goolord, line 259: maybe `EventReturnTypeRep` -> `ReifyEventPolicy` or something

semantics/executable-spec/src/Control/State/Transition/Extended.hs
```diff
@@ -431,14 +462,12 @@ applySTS = applySTSOptsEither defaultOpts
     defaultOpts =
       ApplySTSOpts
         { asoAssertions = globalAssertionPolicy,
-          asoValidation = ValidateAll
+          asoValidation = ValidateAll,
+          asoEvents = EPDiscard
         }
 
-#ifdef STS_ASSERT
```
lehins, line 437: If `STS_ASSERT` flag is no longer used we might want to also remove it from the cabal file

semantics/executable-spec/src/Control/State/Transition/Extended.hs
nc6: I went for consistency with `RuleTypeRep`, in the same file.

semantics/executable-spec/src/Control/State/Transition/Extended.hs
nc6: Oop, good catch, not sure how this got deleted. Totally unintentional!

JaredCorduan (approved):
Thank you @nc6 , this looks great!

If I understand correctly, @DavidEichmann can now use `applyBlockOpts` with the right `EventReturnTypeRep` to make a version of `foldBlocks` that can be used to make a client that receives all the blocks, txs, together with the events. yea?

nc6:
Yep, that's the idea!

PR #2409, opened 2021-08-03
better collateral predicate failure

### LINKED ISSUE #2408: [BUG] - InsufficientCollateral error doesn't consider collateral percentage
**Internal/External**
*Internal*   if an IOHK staff member.

**Area**
*Plutus*    Related to Plutus Scripts (Alonzo).

**Summary**
Collateral is influenced by the `collateralPercentage` parameter. But if insufficient collateral is provided the error does not take this into account, it always displays using fee value.

**Steps to reproduce**
1. On a network configured with `collateralPercentage` to something other than '100'. e.g. alonzo-white is currently set to '150'
2. Submit transaction to execute non-native (plutus) script using collateral utxo equal to fee amount (not 150% of fee)
3. Observe error
```
Command failed: transaction submit  Error: Error while submitting tx: ShelleyTxValidationError ShelleyBasedEraAlonzo (ApplyTxError [UtxowFailure (WrappedShelleyEraFailure (UtxoFailure (InsufficientCollateral (Coin 1400000000) (Coin 1400000000))))])
```

**Expected behavior**
The InsufficientCollateral error should consider  `collateralPercentage` so it is clear to the user exactly how much collateral is expected.

**System info (please complete the following information):**
- OS Name: Mint
- OS Version: 20.2 (Uma)
- Node version: cardano-node 1.27.0 - linux-x86_64 - ghc-8.10 git rev acefba14ef54235ab43105b421754065152da165
- CLI version: cardano-cli 1.27.0 - linux-x86_64 - ghc-8.10 git rev acefba14ef54235ab43105b421754065152da165

**Screenshots and attachments**
None

**Additional context**
None

JaredCorduan:
The previous `InsufficientCollateral` predicate failure was reporting the given fee, not the required collateral, which is confusing. We now report the required collateral.

The golden/example test changes are a bit noisy since I made it now possible to use different protocol parameters, and I also change the amount in the collateral output.

Additionally, in a separate commit, I added an unrelated test. While debugging a separate issue, I made the test for the API function `evaluateTransactionFee` better and I also now test it in alonzo.

closes #2408

### DIFF
--- a/cardano-ledger-test/src/Test/Cardano/Ledger/Examples/TwoPhaseValidation.hs
+++ b/cardano-ledger-test/src/Test/Cardano/Ledger/Examples/TwoPhaseValidation.hs
@@ -45,6 +45,7 @@ import Cardano.Ledger.Alonzo.Tx
     ScriptPurpose (..),
     ValidatedTx (..),
     hashWitnessPPData,
+    minfee,
   )
 import Cardano.Ledger.Alonzo.TxInfo (txInfo, valContext)
 import Cardano.Ledger.Alonzo.TxWitness (RdmrPtr (..), Redeemers (..), TxDats (..), unRedeemers)
@@ -100,6 +101,7 @@ import Shelley.Spec.Ledger.API
   ( BHBody (..),
     BHeader (..),
     Block (..),
+    CLI (..),
     DPState (..),
     DState (..),
     KESPeriod (..),
@@ -1927,3 +1929,29 @@ alonzoBBODYexamples =
     ]
   where
     pf = Alonzo Mock
+
+testEvaluateTransactionFee :: Assertion
+testEvaluateTransactionFee =
+  evaluateTransactionFee @A
+    pparams
+    validatingTxNoWits
+    1
+    @?= minfee pparams (validatingTx pf)
+  where
+    pf = Alonzo Mock
+    pparams = newPParams pf $ defaultPPs ++ [MinfeeA 1]
+    validatingTxNoWits =
+      newTx
+        Override
+        pf
+        [ Body (validatingBody pf),
+          Witnesses'
+            [ ScriptWits [always 3 pf],
+              DataWits [datumExample1],
+              RdmrWits validatingRedeemersEx1
+            ]
+        ]
+
+alonzoAPITests :: TestTree
+alonzoAPITests =
+  testGroup "Alonzo API" [testCase "evaluateTransactionFee" testEvaluateTransactionFee]
--- a/cardano-ledger-test/src/Test/Cardano/Ledger/Examples/TwoPhaseValidation.hs
+++ b/cardano-ledger-test/src/Test/Cardano/Ledger/Examples/TwoPhaseValidation.hs
@@ -164,22 +164,24 @@ testSystemStart = SystemStart $ posixSecondsToUTCTime 0
 freeCostModel :: CostModel
 freeCostModel = CostModel $ 0 <$ fromJust defaultCostModelParams
 
-pp :: Proof era -> Core.PParams era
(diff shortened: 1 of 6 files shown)


shelley/chain-and-ledger/shelley-spec-ledger-test/test/Test/Shelley/Spec/Ledger/Fees.hs
```diff
@@ -508,12 +509,19 @@ testEvaluateTransactionFee :: Assertion
 testEvaluateTransactionFee =
   API.evaluateTransactionFee @(ShelleyEra C_Crypto)
     pp
-    (txSimpleUTxO @C_Crypto)
+    txSimpleUTxONoWit
     1
-    @?= Coin 103
+    @?= minfee pp (txSimpleUTxO @C_Crypto)
```
lehins, line 514: I don't think it is a good idea to switch to `@?= minfee ...` because then this unit test only checks in the end that `evaluateMinFee` indeed calls `minfee`. Hard coding the expected value is better for a unit test like that. 

If it was a property test it would make more sense, but not for a unit test.

~Also,  I might be misunderstanding something, but `evaluateTransactionFee` adds a dummy witnesses, but `minfee`  is getting a different witness, namely `makeWitnessesVKey (hashAnnotated $ txbSimpleUTxO @c) [alicePay]`, so it seem a bit hacky to me.~ Nevermind, it seem like the whole purpose of `evaluateTransactionFee` is to estimate the fee without knowing witness' keys ahead of time.

cardano-ledger-test/src/Test/Cardano/Ledger/Examples/TwoPhaseValidation.hs
```diff
@@ -1889,3 +1929,29 @@ alonzoBBODYexamples =
...
   where
     pf = Alonzo Mock
+
+testEvaluateTransactionFee :: Assertion
+testEvaluateTransactionFee =
+  evaluateTransactionFee @A
+    pparams
+    validatingTxNoWits
+    1
+    @?= minfee pparams (validatingTx pf)
```
lehins, line 1939: Same here, without providing expected value we aren't testing the implementation of `minfee`, because `evaluateTransactionFee` itself calls `minfee`. In other words if `minfee` has a bug, we won't catch it. Hard coding expected values in tests is better.

lehins (approved):
Minor comments on the test cases, but the rest looks good.

shelley/chain-and-ledger/shelley-spec-ledger-test/test/Test/Shelley/Spec/Ledger/Fees.hs
JaredCorduan: exactly, `evaluateTransactionFee` is really just hoping to get `minfee` correct without having all the key witnesses yet.

cardano-ledger-test/src/Test/Cardano/Ledger/Examples/TwoPhaseValidation.hs
JaredCorduan: maybe I should just have both tests?

lehins: I can add a property test just like that in the work that I am doing. So I can change this test to a hard coded value in a later PR, if that works for you.

JaredCorduan: that would be great, thank you!

nc6 (approved):

PR #2410, opened 2021-08-03
WIP second iteration pool rankings

JaredCorduan:
This is a draft PR that @HeinrichApfelmus can use to start thinking about what would be the best interface for the wallet for the second iteration of the stake pool rankings.

The functions `getMyopicROS` and `saturationROS` provide the two new values that we want to display. Looking at the parameters they need, we can decide how best to make something usable for the wallet.


JaredCorduan:
I'm going to close this, and re-open once I hear back from the researchers.

HeinrichApfelmus:
> I'm going to close this, and re-open once I hear back from the researchers.

Thanks, sounds good to me! I have reimplemented the (old) ranking scheme in the wallet using the new `GetRewardInfoPools` query, so we can quickly pivot to new ranking indicators.

PR #2413, opened 2021-08-04
rename wpphash

JaredCorduan:
Rename `wppHash` and friends to ` scriptIntegrityHash`.

### DIFF
--- a/cardano-ledger-core/src/Cardano/Ledger/Hashes.hs
+++ b/cardano-ledger-core/src/Cardano/Ledger/Hashes.hs
@@ -19,9 +19,12 @@ module Cardano.Ledger.Hashes
     EraIndependentScriptData,
     EraIndependentAuxiliaryData,
     EraIndependentPParamView,
-    EraIndependentWitnessPPData,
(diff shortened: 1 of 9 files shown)


cardano-ledger-core/src/Cardano/Ledger/Hashes.hs
```diff
@@ -59,6 +62,10 @@ data EraIndependentScriptData
 
 data EraIndependentPParamView
 
+data EraIndependentScriptIntegrity
+
+{-# DEPRECATED EraIndependentWitnessPPData "Use EraIndependentScriptIntegrity instead" #-}
```
nc6, line 67: We should make this a type alias, surely? We can certainly alias empty data types.

nc6 (changes requested):

cardano-ledger-core/src/Cardano/Ledger/Hashes.hs
JaredCorduan: yes indeed, thanks!

nc6 (approved):

PR #2417, opened 2021-08-09
Add more event types

goolord:
- [ ] NewEra event
- [x] NewEpoch event
- [x] Rewards event
- [ ] StakeDist event


JaredCorduan:
This is looking good, thanks @goolord !

I was talking with @erikd , and the most important event for him right now is something that provides the sum of all rewards given on the epoch boundary. This will probably have to be two events, since this needs to include the rewards from the MIR certs.

First, we should provide the the sum of the rewards calculated inside `applyRUpd`, namely the sum of `regRU`:

https://github.com/input-output-hk/cardano-ledger-specs/blob/e622ba7d26ab16c6186255cfe62178a87538f6b8/shelley/chain-and-ledger/executable-spec/src/Shelley/Spec/Ledger/LedgerState.hs#L1058

I think `applyRU` will need to return this value, so that the `NEWEPOCH` rule can provide it as an event.

Second, we should provide the the sum of both the reserves MIR and treasury MIR, `totR + totT`:

https://github.com/input-output-hk/cardano-ledger-specs/blob/e622ba7d26ab16c6186255cfe62178a87538f6b8/shelley/chain-and-ledger/executable-spec/src/Shelley/Spec/Ledger/STS/Mir.hs#L104-L105

JaredCorduan (approved):
Looks great, thank you!

JaredCorduan:
@goolord - when this goes green, are you okay merging this now and getting to the other two bullet points in a follow up PR? I think it would be great to get this in now with what you have.

goolord:
yes, absolutely, that was my plan.

PR #2418, opened 2021-08-09
Small formal spec things

polinavino:
- rename `wppHash` to `scriptIntegrityHash` and change its calculation to match the code
- add corollary saying that if the collateral percent parameter is greater than 100, the transaction always pays at least the `minfee`
- change `Prices` from `Integer` to `Rational`

### DIFF
--- a/alonzo/formal-spec/protocol-parameters.tex
+++ b/alonzo/formal-spec/protocol-parameters.tex
@@ -6,28 +6,38 @@ in addition to those that are already defined in the Shelley specification~\cite
 
 \vspace{12pt}
 \begin{tabular}{lp{5in}}
+  $\CostMod$ &
+  $c\in\CostMod$ is the vector of coefficients that are used to compute
+  a value $exu\in\ExUnits$ given a vector of some resource primitives. The mapping is defined
+  concretely by the specific version of the Plutus interpreter that is associated with $\Language$.
+  We keep this type as abstract in the specification, see \cite{plutuscore} and \cite{plutustech}
+  for details.
+  \\
   $\Language$ &
-  This represents the language name/tag, including the
-  version number.
+  This represents phase-2 language types. Currently there is only $\PlutusVI$.
+  \\
+  $\Prices$ &
+  $\var{(pr_{mem}, pr_{steps})}\in \Prices$ consists of two rational numbers
+  that correspond to the components of $\ExUnits$:
+  $pr_{mem}$ is the price per unit of memory, and $pr_{steps}$ is the price per
+  reduction step. This is used to calculate the cost for a specific script execution.
   \\
   $\ExUnits$ &
-  A term of this type contains two integer values,
-  $(mem, steps)$.
+  $(mem, steps)\in \ExUnits$ is made up of two integer values.
   These represent abstract notions of the relative memory usage and script execution steps,
   respectively.
   \\
-  $\CostMod$ &
-  A term of this type represents the vector of coefficients that are used to generate
-  a term of type $\ExUnits$ given a vector of some resource primitives. The mapping is defined
-  concretely by the specific version of the Plutus interpreter that is associated with $\Language$.
-  We keep this type as abstract in the specification.
-  \\
-  $\Prices$ &
-  A term of this type comprises two integer values that correspond to the components of $\ExUnits$,
-  $\var{(pr_{mem}, pr_{steps})}$:
-  $pr_{mem}$ is the price (in Ada) per unit of memory, and $pr_{steps}$ is the price (in Ada) per
-  reduction step. This is used to calculate the Ada cost for a specific script execution.
+  $\LangDepView$ &
+  A pair of two byte strings, where the first represents the serialized language tag (eg. the tag for $\PlutusVI$),
+  and the second, the protocol parameters that must be fixed (by the transaction) when carrying a phase-2 script
+  in that language.
 \end{tabular}
+
+The function $\fun{serialize}$ is left abstract, as
+these is an implementation-dependent serialization function. It must
+be implemented for all serializable types, including all
+collections of protocol parameters needed to construct a $\LangDepView$ for
+each existing language.
 
 \subsection{Language Versions and Backwards Compatibility Requirements}
 \label{sec:versions}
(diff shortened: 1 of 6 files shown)


alonzo/formal-spec/transactions.tex
```diff
@@ -162,8 +162,7 @@ \section{Transactions}
        & \times~ \Wdrl  & \fun{txwdrls} &\text{Reward withdrawals}\\
        & \times ~\Update^?  & \fun{txUpdates} & \text{Update proposals}\\
        & \times ~\hldiff{\powerset{\KeyHash}} & \hldiff{\fun{reqSignerHashes}} & \text{Hashes of VKeys passed to scripts}\\
-       & \times ~\hldiff{\WitnessPPDataHash^?} & \hldiff{\fun{wppHash}} & \text{Hash of script-related data}\\
-       & \times ~\AuxiliaryDataHash^? & \fun{txADhash} & \text{Auxiliary data hash}\\
```
WhatisRT, line 166: This line seems to have been removed by accident?

alonzo/formal-spec/protocol-parameters.tex
```diff
@@ -182,9 +205,8 @@ \subsection{Script Evaluation Cost Model and Prices}
   %
   \begin{align*}
     & \fun{getLanguageView} \in \PParams \to \Language \to \LangDepView \\
-    & \fun{getLanguageView}~\var{pp}~\PlutusVI = \var{costmdls}~\mapsto~ (\fun{costmdls}~{pp}~\{\PlutusVI\})
+    & \fun{getLanguageView}~\var{pp}~\PlutusVI = (\fun{serialize}~\PlutusVI, ~ (\fun{serializeEncoding}~(\fun{costmdls}~{pp}~\{\PlutusVI\})))
```
WhatisRT: I really don't think this should have been changed. The only reason why it isn't implemented like this is because Haskell can't properly express it. This adds a bunch of complexity and extra primitive functions that don't add any extra clarity.

alonzo/formal-spec/protocol-parameters.tex
```diff
@@ -108,6 +122,15 @@ \subsection{Script Evaluation Cost Model and Prices}
 signatures that must be checked during validation.
 
 \begin{figure*}[htb]
+  \emph{Abstract functions}
+  %
+  \begin{equation*}
+    \begin{array}{r@{~\in~}l@{\qquad\qquad\qquad}r}
+      \fun{serialize},~\fun{serializeEncoding} & \mathsf{Type} & \ByteString \\
```
WhatisRT, line 129: ByteString isn't defined, so this doesn't compile

alonzo/formal-spec/protocol-parameters.tex
```diff
@@ -6,29 +6,42 @@ \section{Language Versions and Cost Models}
...
-  These represent abstract notions of the relative memory usage and script execution steps,
-  respectively.
-  \\
   $\CostMod$ &
   A term of this type represents the vector of coefficients that are used to generate
   a term of type $\ExUnits$ given a vector of some resource primitives. The mapping is defined
   concretely by the specific version of the Plutus interpreter that is associated with $\Language$.
   We keep this type as abstract in the specification.
+  \begin{note}
+    Add a link to $\CostMod$ documentation
```
WhatisRT: I think we shouldn't add TODO's in the repo at this point, we probably want this document to be presentable to the public as much as possible

WhatisRT (changes requested):
The type `WitnessPPDataHash` and the function `hashWitnessPPData` also need to be renamed

alonzo/formal-spec/protocol-parameters.tex
polinavino: How do you propose we make the types match up? `LangDepView` should now be a type common to all languages, ie just a bytestring of all the relevant info

WhatisRT: I think regarding the types it was just fine before. In the Shelley spec, there are also `PParams` and `PParamsUpdate` types that don't work like that in Haskell. There is no reason to limit ourselves to what's possible to express in Haskell.

Adding the `PlutusVI` tag is probably necessary, but again, no byte strings necessary.

WhatisRT: As a compromise, I propose we make the type `LangDepView` abstract, and say that we require it to have a `hash` function and that it should be able to hold the data we put into it in `getLanguageView`. Then, if necessary, we can have an appendix that describes the details.

polinavino: i forget, is this still your opinion or are you ok with using `serialize` (which I already used in defining `serSize` in ShelleyMA)

alonzo/formal-spec/protocol-parameters.tex
```diff
@@ -6,29 +6,42 @@ \section{Language Versions and Cost Models}
...
-  These represent abstract notions of the relative memory usage and script execution steps,
-  respectively.
-  \\
   $\CostMod$ &
   A term of this type represents the vector of coefficients that are used to generate
   a term of type $\ExUnits$ given a vector of some resource primitives. The mapping is defined
   concretely by the specific version of the Plutus interpreter that is associated with $\Language$.
   We keep this type as abstract in the specification.
+  % \begin{note}
+  %   Add a link to $\CostMod$ documentation
```
WhatisRT: I could find two documents related to cost models, one in the plutus repo and the other one that was recently posted in the plutus slack channel. I think we can probably link to them already, maybe with a remark that it's unpublished

alonzo/formal-spec/protocol-parameters.tex
WhatisRT: I think `serialize` is fine. I'd prefer it if it was just a single `ByteString` though that contains everything, and no `serializeEncoding` function.

alonzo/formal-spec/protocol-parameters.tex
```diff
@@ -6,29 +6,40 @@ \section{Language Versions and Cost Models}
 
 \vspace{12pt}
 \begin{tabular}{lp{5in}}
+  $\CostMod$ &
+  A term of this type represents the vector of coefficients that are used to generate
```
WhatisRT: I've already mentioned on the other PR that 'term of this type' isn't great, because it actually means something different. I think 'element' or 'member' is just fine here and at the end of the sentence.

alonzo/formal-spec/protocol-parameters.tex
```diff
@@ -6,29 +6,40 @@ \section{Language Versions and Cost Models}
...
+  We keep this type as abstract in the specification, see \cite{plutuscore} and \cite{plutustech}
+  for details.
+  \\
   $\Language$ &
-  This represents the language name/tag, including the
-  version number.
+  This represents phase-2 language types. Currently there is only $\PlutusVI$.
+  \\
+  $\Prices$ &
+  A term of this type comprises two rational number values that correspond to the components of $\ExUnits$,
```
WhatisRT: Same here

alonzo/formal-spec/protocol-parameters.tex
```diff
@@ -58,13 +69,14 @@ \subsection{Determinism of Script Evaluation}
...
-protocol parameter $\var{ppm}~\in~\Ppm$, and that $\PParams~=~\Ppm~\mapsto~\Tppm$.
-The type $\LangDepView$ is a subrecord of $\PParams$.
-%
-At the time of writing, the only parameter that needs to be passed to the $\PlutusVI$
-interpreter is the cost model, specifically only the cost model for the $\PlutusVI$ language.
+The $\fun{getLanguageView}$ function is used in the computation of this integrity hash.
+Shown in (Figure~\ref{fig:defs:protocol-parameters}), it filters the data relvant
+to a given language from the protocol parameters, and returns a pair of byte strings :
+the serialized representation of
+the language tag, and the serialized representation of the relevant PPs.
```
WhatisRT: I'd spell out protocol parameters

WhatisRT (commented):
The only remaining thing is spelling out PP now, everything else looks good!

alonzo/formal-spec/protocol-parameters.tex
WhatisRT: I.e. here it should say 'of the relevant protocol parameters'

WhatisRT (approved):
LGTM

PR #2419, opened 2021-08-09
Fix `WitVKey` instance for `Ord`

lehins:
Current Ord instance allows breakage of referential transparency  because this law doesn't hold: `compare x y == EQ ==> x == y` 
This PR fixes this problem and adds couple property tests that ensure the fix.

There are also a couple of other minor improvements included in the PR.

### DIFF
--- a/alonzo/impl/src/Cardano/Ledger/Alonzo/TxSeq.hs
+++ b/alonzo/impl/src/Cardano/Ledger/Alonzo/TxSeq.hs
@@ -54,6 +54,7 @@ import Cardano.Ledger.Serialization
 import Control.Monad (unless)
 import Data.ByteString (ByteString)
 import qualified Data.ByteString.Lazy as BSL
+import Data.ByteString.Short (fromShort)
 import Data.Coders
   ( decodeList,
     decodeMap,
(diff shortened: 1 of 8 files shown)


alonzo/impl/src/Cardano/Ledger/Alonzo/TxSeq.hs
```diff
@@ -180,16 +181,18 @@ hashTxSeq ::
...
 hashTxSeq (TxSeq' _ bodies ws md vs) =
   coerce $
-    hashStrict
-      ( hashPart bodies
-          <> hashPart ws
-          <> hashPart md
-          <> hashPart vs
-      )
+    hashStrict $
+      fromShort $
```
JaredCorduan, line 185: why go by why of a short bytestring here?

JaredCorduan (approved):
LGTM, thanks!

alonzo/impl/src/Cardano/Ledger/Alonzo/TxSeq.hs
lehins: I was hoping this question would come up :)

That is because `hashToBytes` copies the underlying `ShortByteString` into a `ByteString`, while `hashToBytesShort` is a simple type cast.

In other words we save 4 memory copies when switching to `ShortByteString` and 2 more when switching to `mconcat`.

JaredCorduan: oh great, thank you for explaining!

PR #2422, opened 2021-08-13
Remove usage of partial `decodeUtf8`

lehins:
It is widely advised against usage of `decodeUtf8`. In place were it is used we are decoding base 16 encoded text anyways, so we should be using faster and total `decodeLatin1`.

* [x] There are also two more uses in Byron code, these could also be swapped for `decodeLatin1`, but I am not sure how amendable is the old code. Done.

https://github.com/input-output-hk/cardano-ledger-specs/blob/6f070a120c8e5a35e773a95ce076e10a0727b419/byron/ledger/impl/src/Cardano/Chain/Common/Address.hs#L208

https://github.com/input-output-hk/cardano-ledger-specs/blob/6f070a120c8e5a35e773a95ce076e10a0727b419/byron/ledger/impl/src/Cardano/Chain/Common/Address.hs#L235

### DIFF
--- a/byron/ledger/impl/src/Cardano/Chain/Common/Address.hs
+++ b/byron/ledger/impl/src/Cardano/Chain/Common/Address.hs
@@ -86,6 +86,7 @@ import Data.ByteString.Base58
     decodeBase58,
     encodeBase58,
   )
+import Data.Text.Encoding (decodeLatin1)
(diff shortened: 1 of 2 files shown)


JaredCorduan (approved):
Thanks @lehins ! and thanks for explain this to me!

PR #2425, opened 2021-08-17
update plutus to alonzo/rc-2

JaredCorduan:
This PR ended up being less trivial than I expected, so I will leave some notes here for posterity.

* Plutus no longer restricts the bytestring inside `Data` to 64 bytes, so our tests and CDDL have been adapted accordingly.
* The Plutus Data Constructor `102` is now serialized as `#6.102([uint, [* a]])` instead of `#6.102([uint, * a])`.
* The serialized size of the Plutus cost model was problematically large. We no longer transmit the keys, only the values (and assume they are sent in ascending order by key name).
* I have moved the Alonzo trace classifier test to the nightly tests, there is a discard problem that we cannot locate and there were far too many failed CI tests to be worth it (for now).


alonzo/test/cddl-files/alonzo.cddl
```diff
@@ -264,7 +264,7 @@ plutus_data = ; New
   / { * plutus_data => plutus_data }
   / [ * plutus_data ]
   / big_int
```
michaelpj, line 266: `big_int` can be any big int now for the same reason.

redxaxder (approved):

alonzo/impl/src/Cardano/Ledger/Alonzo/Scripts.hs
```diff
@@ -187,8 +194,31 @@ checkCostModel cm =
...
+decodeCostModel PlutusV1 =
+  case defaultCostModelParams of
+    Nothing -> fail "Default Plutus Cost Model is corrupt."
+    Just dcm -> do
+      checked <- checkCostModel <$> decodeArrayAsMap (Map.keysSet dcm) fromCBOR
+      case checked of
+        Left e -> fail e
+        Right cm -> pure cm
+
+decodeArrayAsMap :: Ord a => Set a -> Decoder s b -> Decoder s (Map a b)
```
michaelpj, line 210: Okay, so we gave you the cost model as `Map Text Integer` to make it very homogenous, and so you could work with it somewhat independently of the API we give you. But if we're using `defaultCostModelParams` during deserialization, then that's kind of gone out the window. 

In which case: it seems like this would be easier if it was `Map ModelKey Integer` where `ModelKey` has a `Enum`/`Bounded` instances? It also would make for a nicer, more generic function:
```
decodeArrayAsMap :: (Enum k, Bounded k, Ord k) => Decoder s v -> Decoder s (Map k v)
```

JaredCorduan: Yep, I agree, that would probably be nicer to work with and maintain.

michaelpj: https://jira.iohk.io/browse/SCP-2696

JaredCorduan: Thanks @michaelpj ! The enumeration will have to match what we are currently doing (ascending order by key) if we want to make this change sometime before the next hardfork after alonzo (since the cost model is a part of the script integrity hash).

michaelpj: Good point!

PR #2427, opened 2021-08-17
Reorganize property tests to reside in Cardano.Ledger.Test

TimSheard:
Reorganized some things to make mutual dependencies less likely when things are moved in the next step


JaredCorduan:
Sorry this never got merged. :( at this point, I think it would be easier to redo it. But thank you for trying to clean things up!

PR #2428, opened 2021-08-18
Minor improvements

lehins:
Nothing terribly exciting, just a few minor things I noticed while reading code.

Change to `feesOK` just makes it look a little closer like the spec, so no change in behavior.


nc6 (approved):
Looks great, thanks!

PR #2430, opened 2021-08-20
Debug help for failed Plutus scripts

JaredCorduan:
This adds all the necessary information for re-runing a failed plutus script inside the `ValidationTagMismatch` predicate failure.

Additionally, I've replaced the `[Text]` field in `ValidationTagMismatch` with descriptive types for all the different scenarios.  Previously, the `Eq` instance for `ValidationTagMismatch` was ignoring this field, I suspect to make testing easier. I've made the `Eq` instance care about all the fields, and gave the tests a way to ignore the inner details of the plutus error message.

### DIFF
--- a/alonzo/impl/src/Cardano/Ledger/Alonzo/Rules/Utxos.hs
+++ b/alonzo/impl/src/Cardano/Ledger/Alonzo/Rules/Utxos.hs
@@ -1,4 +1,5 @@
 {-# LANGUAGE DataKinds #-}
+{-# LANGUAGE DeriveAnyClass #-}
 {-# LANGUAGE DeriveGeneric #-}
 {-# LANGUAGE DerivingStrategies #-}
 {-# LANGUAGE FlexibleContexts #-}
@@ -15,6 +16,7 @@ module Cardano.Ledger.Alonzo.Rules.Utxos
     UtxosPredicateFailure (..),
     constructValidated,
     lbl2Phase,
+    TagMismatchDescription (..),
   )
 where
 
@@ -34,7 +36,7 @@ import Cardano.Ledger.Alonzo.Tx
     txouts,
   )
 import qualified Cardano.Ledger.Alonzo.TxBody as Alonzo
-import Cardano.Ledger.Alonzo.TxInfo (ScriptResult (..))
+import Cardano.Ledger.Alonzo.TxInfo (FailureDescription (..), ScriptResult (..))
 import qualified Cardano.Ledger.Alonzo.TxWitness as Alonzo
 import Cardano.Ledger.BaseTypes
   ( Globals,
@@ -59,7 +61,6 @@ import Data.Foldable (toList)
 import qualified Data.Map.Strict as Map
 import Data.Sequence.Strict (StrictSeq)
 import Data.Set (Set)
-import Data.Text
 import GHC.Generics (Generic)
 import GHC.Records (HasField (..))
 import NoThunks.Class (NoThunks)
@@ -183,7 +184,11 @@ scriptsValidateTransition = do
   case collectTwoPhaseScriptInputs ei sysSt pp tx utxo of
     Right sLst ->
       case evalScripts @era tx sLst of
-        Fails sss -> False ?!## ValidationTagMismatch (getField @"isValidating" tx) (pack (Prelude.unlines sss))
+        Fails sss ->
+          False
+            ?!## ValidationTagMismatch
+              (getField @"isValidating" tx)
+              (FailedUnexpectedly sss)
         Passes -> pure ()
     Left info -> failBecause (CollectErrors info)
   pup' <-
@@ -228,7 +233,7 @@ scriptsNotValidateTransition = do
   case collectTwoPhaseScriptInputs ei sysSt pp tx utxo of
     Right sLst ->
       case (evalScripts @era tx sLst) of
-        Passes -> False ?!## ValidationTagMismatch (getField @"isValidating" tx) (pack ("Script expected to fail, passes."))
+        Passes -> False ?!## ValidationTagMismatch (getField @"isValidating" tx) PassedUnexpectedly
         Fails _sss -> pure ()
     Left info -> failBecause (CollectErrors info)
   pure $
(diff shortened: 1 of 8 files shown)


michaelpj (approved):
Seems fine. We could move the executable into `plutus` if we wanted: we have an executable that does this sort of thing, it's only not set up to take its arguments as a single serialized blob. But here is fine too.

JaredCorduan:
@michaelpj I've removed your review since I've taken it out of draft and changed it a lot. Don't feel obligated to review it again (unless you want to), I can get a review from the usual suspects.

alonzo/impl/src/Cardano/Ledger/Alonzo/Rules/Utxos.hs
```diff
@@ -237,11 +242,27 @@ scriptsNotValidateTransition = do
         _fees = fees <> Val.coin (balance @era (eval (getField @"collateral" txb ◁ utxo)))
       }
 
+data TagMismatchDescription
+  = PassedUnexpectedly
+  | FailedUnexpectedly [FailureDescription]
+  deriving (Show, Eq, Generic, NoThunks)
```
nc6, line 248: Deriving `NoThunks` here is a bit weird, since it's not obvious that we _will_ have no thunks. Are we forcing the descriptions during construction?

alonzo/impl/src/Cardano/Ledger/Alonzo/TxInfo.hs
```diff
@@ -362,7 +381,35 @@ valContext ::
   Data era
 valContext txinfo sp = Data (P.toData (P.ScriptContext txinfo (transScriptPurpose sp)))
 
-data ScriptResult = Passes | Fails ![String]
+data FailureDescription
+  = OnePhaseFailure Text
+  | PlutusFailure Text ByteString
```
nc6, line 386: Label fields here?

alonzo/impl/src/Cardano/Ledger/Alonzo/TxInfo.hs
```diff
@@ -362,7 +381,35 @@ valContext ::
...
+  toCBOR (PlutusFailure s b) = encode $ Sum PlutusFailure 1 !> To s !> To b
+
+instance FromCBOR FailureDescription where
+  fromCBOR = decode (Summands "FailureDescription" dec)
+    where
+      dec 0 = SumD OnePhaseFailure <! From
+      dec 1 = SumD PlutusFailure <! From <! From
+      dec n = Invalid n
+
+data ScriptResult = Passes | Fails ![FailureDescription]
```
nc6, line 400: This bang will only force the head of the list...

alonzo/impl/src/Cardano/Ledger/Alonzo/TxInfo.hs
```diff
@@ -395,81 +488,102 @@ runPLCScript proxy (CostModel cost) scriptbytestring units ds =
...
+              pack $
+                unlines
+                  [ "\nThe 3 arg plutus script (" ++ name ++ ") fails.",
+                    show e,
+                    "The data is: " ++ show dat,
+                    "The redeemer is: " ++ show redeemer,
+                    "The third data argument, does not decode to a context\n" ++ show info
+                  ]
+            db = B64.encode . serialize' $ PlutusDebug cm eu scriptbytestring ds PlutusV1
+        Just info2 -> Fails [PlutusFailure line db]
```
nc6, line 526: `line` and `db` will be thunks here...

nc6 (approved):
So there's some weird stuff potentially with evaluation in this PR, and `NoThunks` checks on types that will be decidedly thunky. But none of that should change semantics, and in the name of getting a RC today, I'll merge and we can update this later.

alonzo/impl/src/Cardano/Ledger/Alonzo/TxInfo.hs
```diff
@@ -395,81 +488,102 @@ runPLCScript proxy (CostModel cost) scriptbytestring units ds =
...
-                  "The redeemer is: " ++ show redeemer,
-                  "The second data argument, does not decode to a context\n" ++ show info
-                ]
-        Just info2 -> Fails [line]
+              pack $
+                unlines
+                  [ "\nThe 2 arg plutus script (" ++ name ++ ") fails.",
+                    show e,
+                    "The redeemer is: " ++ show redeemer,
+                    "The second data argument, does not decode to a context\n" ++ show info
```
michaelpj, line 554: Is the textual information here now redundant with the debug information?

JaredCorduan: yep, it is. maybe the human readable part should be whittled down, good point, thanks.

alonzo/impl/src/Cardano/Ledger/Alonzo/TxInfo.hs
JaredCorduan: do you mean add some comments? that seems a good idea to me, thanks

alonzo/impl/src/Cardano/Ledger/Alonzo/TxInfo.hs
```diff
@@ -395,81 +488,102 @@ runPLCScript proxy (CostModel cost) scriptbytestring units ds =
...
+  ScriptResult
+explain_plutus_failure _proxy scriptbytestring e ds@[dat, redeemer, info] cm eu =
   -- A three data argument script.
   let ss :: Script era
       ss = PlutusScript scriptbytestring
       name :: String
       name = show ss
    in case P.fromData info of
-        Nothing -> Fails [line]
+        Nothing -> Fails [PlutusFailure line db]
```
michaelpj, line 514: also I forget, but could we put the structured type in here and just print it specially later?

JaredCorduan: I think the train has left the station, sorry. But as for you question, I don't know, we would have to write our own show for it to produce the base64 in the logs messages, which is maybe a weird/not-always-what-you-want show instance?

michaelpj: You're using `Show` for log messages? I'm not totally sure that's a good idea, but if that's the situation then okay.

JaredCorduan: I'm not actually sure, to be honest.

PR #2431, opened 2021-08-23
Expand events around MIR transfers.

nc6:
Add two new events that provide full detail about MIR transfers being
made. These subsume the functionality of the existing event.


JaredCorduan (approved):
LGTM, thanks!

goolord (approved):

PR #2433, opened 2021-08-24
startStep now computes an initial RewardProvenance.

TimSheard:
Several of the fields were not being populated in the reward provenance. In particular, the `startStep` function was computing several of them, but was not in the provenance monad.


JaredCorduan (approved):

nc6:
What's the motivation behind this change? Can you add it to the commit/PR messages?

JaredCorduan:
@nc6 I added a description. I'll be better remembering to leave good notes in PR when doing code reviews, thanks!

PR #2434, opened 2021-08-24
Alonzo utxow Tx generator and properties

lehins:
This is a self contained PR that implements QuickCheck generators for `ValidatedTx`,  `DPState` and other types necessary for running `"LEDGER"` sts rules, which among others also include `"UTXOW"`.

There aren't many properties, which can be fixed with a follow up PR, but generators are fairly powerful. The unique approach in this PR is that all the data inside and outside LEDGER state needed for a valid  transaction is built up on the go as needed, instead of a more common approach of generating a huge world state and picking out random things from it.

Further potential improvements:

* Parameterize for different eras
* Expand generation (not all fields of ValidatedTx are explored, only ones that are related in some way to UTXOW)
* Some form of custom shrinking

### DIFF
--- a/cardano-ledger-test/src/Test/Cardano/Ledger/Properties.hs
+++ b/cardano-ledger-test/src/Test/Cardano/Ledger/Properties.hs
@@ -1,0 +1,243 @@
+{-# LANGUAGE DataKinds #-}
+{-# LANGUAGE TupleSections #-}
+{-# LANGUAGE TypeApplications #-}
+
+module Test.Cardano.Ledger.Properties where
+
+--import Debug.Trace (trace)
+import qualified Data.Sequence.Strict as Seq
+import Cardano.Ledger.Hashes (ScriptHash (..))
+import Cardano.Ledger.SafeHash (hashAnnotated)
+--import Cardano.Ledger.Mary.Value (Value (..))
+import Data.ByteString.Short (ShortByteString)
+import Cardano.Ledger.Alonzo.Language (Language (..))
+import Cardano.Slotting.Slot (SlotNo (..))
+import Data.Maybe (fromJust)
+import Plutus.V1.Ledger.Api (defaultCekCostModelParams)
+import Control.State.Transition.Extended hiding (Assertion)
+import Cardano.Ledger.Alonzo.Scripts (CostModel (..))
+import Shelley.Spec.Ledger.STS.Utxo (UtxoEnv (..))
+import Cardano.Ledger.Alonzo.PParams (PParams, PParams' (..))
+import Cardano.Ledger.Alonzo.TxWitness (TxWitness (..), Redeemers (..))
+import Test.Shelley.Spec.Ledger.Utils
+  ( applySTSTest,
+    mkKeyPair',
+    runShelleyBase,
+  )
+import Cardano.Ledger.Alonzo.Tx
+  ( IsValidating (..),
+    --ScriptPurpose (..),
+    ValidatedTx (..),
+    --hashWitnessPPData,
+  )
+import Cardano.Ledger.ShelleyMA.Timelocks (Timelock (..), ValidityInterval (..))
+import Cardano.Ledger.Alonzo.Rules.Utxow (AlonzoUTXOW)
+import Data.Default.Class (Default (def))
+import Shelley.Spec.Ledger.LedgerState (UTxOState (..))
+import Test.QuickCheck (
+  Property, (===), elements, oneof,
+  --Positive (..),
+  chooseInt, vectorOf,
+  counterexample, property, choose)
+import Test.Tasty (TestTree, testGroup)
+import Test.Tasty.QuickCheck (Gen, testProperty, forAll, Arbitrary (..))
+import Shelley.Spec.Ledger.API
+  ( UTxO (..), TxIn, Addr (..), Credential (..), StakeReference (..), Wdrl (..))
+import Shelley.Spec.Ledger.UTxO (makeWitnessVKey, balance)
+import Cardano.Ledger.Alonzo (AlonzoEra)
+import Cardano.Ledger.Keys (
+  coerceKeyRole, hashKey, KeyHash (..), KeyPair (..), KeyRole(..), GenDelegs (..))
+import Test.Shelley.Spec.Ledger.ConcreteCryptoTypes (C_Crypto)
+import Test.Shelley.Spec.Ledger.Serialisation.EraIndepGenerators ()
+import Test.Cardano.Ledger.Alonzo.Serialisation.Generators ()
+import qualified Data.Set as Set
+import Data.Set (Set)
+import qualified Data.Map as Map
+import Data.Map (Map)
+import Cardano.Ledger.BaseTypes (Network (Testnet), StrictMaybe (..))
+import Cardano.Ledger.Alonzo.TxBody (TxOut (..), TxBody (..))
+import Cardano.Ledger.Val (coin, inject, (<->))
+import Cardano.Ledger.Coin (Coin (..))
+import Cardano.Ledger.Alonzo.Data (DataHash, Data)
+
+type A = AlonzoEra C_Crypto
+
+newtype KeySpace = KeySpace [(KeyHash 'Witness C_Crypto, KeyPair 'Witness C_Crypto)]
+  deriving (Show)
+
+genListBetween :: Int -> Int -> Gen a -> Gen [a]
+genListBetween a b g = chooseInt (a, b) >>= (flip vectorOf g)
+
+data Vault a b =
+  Vault { sid :: [a]
+        , sec :: Map a b
+        }
+
+data BagOSecrets =
+  BagOSecrets
+    { keys :: Vault (KeyHash 'Witness C_Crypto) (KeyPair 'Witness C_Crypto)
+    , timelocks :: Vault (ScriptHash C_Crypto) (Timelock A, [KeyPair 'Witness C_Crypto])
+    , plutons :: Vault (ScriptHash C_Crypto) ShortByteString
+    , datums :: Vault (DataHash C_Crypto) (Data A)
+    }
+
+findKey :: KeySpace -> KeyHash 'Witness C_Crypto -> KeyPair 'Witness C_Crypto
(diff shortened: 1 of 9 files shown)


shelley/chain-and-ledger/shelley-spec-ledger-test/src/Test/Shelley/Spec/Ledger/Utils.hs
```diff
@@ -182,7 +183,8 @@ data RawSeed = RawSeed !Word64 !Word64 !Word64 !Word64 !Word64
 
 instance Arbitrary RawSeed where
   arbitrary =
-    RawSeed <$> arbitrary <*> arbitrary <*> arbitrary <*> arbitrary <*> arbitrary
+    RawSeed
+      <$> chooseAny <*> chooseAny <*> chooseAny <*> chooseAny <*> chooseAny
```
JaredCorduan, line 189: I'd love to learn from the `Random` maintainer about when to use `arbitrary` vs `chooseAny`. :smile: (if it's long, you can inform me later, or give me a reading assignment)

lehins: It is not long at all, but it is an extremely important differece! If you use `Random`/`Uniform` class then you'll get uniform distribution, but when you use `Arbitrary` instance then you get some sort of randomness that depends on `size` For example `chooseAny` for `Int` will pick a truly pseudo random number uniformly on the whole range `minBound .. maxBound`:
```haskell
λ> sample (chooseAny :: Gen Int)
8210185546773420831
518266691867196564
-6264935124352629180
1513173028515106965
6464989126909467446
-3226533972854067600
-2010886522517333871
-7398707487487345588
5672178984769180970
-5781351238750652131
244116361716631734
```
while `arbitrary` will give you small values that can be scaled with size:
```haskell
λ> sample (((,) <$> arbitrary <*> resize 300 arbitrary) :: Gen (Int, Int))
(0,-296)
(-2,156)
(0,-141)
(5,-153)
(-3,-147)
(-10,-242)
(7,-77)
(4,288)
(-12,167)
(-18,151)
(6,-80)
```

Latter is useful for testing, but sometimes it gives you a false sense of security because the range of values that is tested is not what most people expect, eg:

```haskell
λ> quickCheck $ forAll (arbitrary :: Gen Int) (< 100)
+++ OK, passed 100 tests.
```

cardano-ledger-test/src/Test/Cardano/Ledger/Properties.hs
```diff
@@ -0,0 +1,519 @@
...
+  bogusCollateralTxId <- lift arbitrary
+  let bogusCollateralTxIns =
+        Set.fromList
+          [ TxIn bogusCollateralTxId (fromIntegral i)
+            | i <- [maxBound, maxBound - 1 .. maxBound - maxCollateralCount - 1]
+          ]
+  collateralAddresses <-
+    replicateM maxCollateralCount $
+      genRecipientWithPayment . coerceKeyRole . KeyHashObj =<< genCredential
+  bogusCollateralKeyWitsMakers <-
```
JaredCorduan, line 677: you are just filling out the field so that the cost in bytes is over-estimated, right?

shelley/chain-and-ledger/shelley-spec-ledger-test/src/Test/Shelley/Spec/Ledger/Utils.hs
JaredCorduan: ah, makes sense, thanks!

cardano-ledger-test/src/Test/Cardano/Ledger/Properties.hs
lehins: yep

JaredCorduan:
This looks really great! Using Reader, State and Gen together seems to have worked out very well. I like generating things as needed, but re-using previously generated things at some point.

It's interesting to see how you build up the transactions, it's more "wallet-like" than our previous generators. This has me wondering if the CLI helper class would be of any help here. https://github.com/input-output-hk/cardano-ledger-specs/blob/8efcfc755faae4db3a64ad45343235fce3ed5c47/shelley/chain-and-ledger/executable-spec/src/Shelley/Spec/Ledger/API/Wallet.hs#L359

I'll be digesting this some more, but it's definitely the test I was hoping for, thanks!

JaredCorduan (approved):
This look great! I appreciate all the comments, they made following along much easier.

I'll make a note here about future improvements while I am thinking about it. It would be great to add this property test here

https://github.com/input-output-hk/cardano-ledger-specs/blob/e59e6ae405e9054966180a26d759b078f8f87329/alonzo/test/test/Test/Cardano/Ledger/Alonzo/PropertyTests.hs#L122-L141

would be a great addition to these new tests.

cardano-ledger-test/src/Test/Cardano/Ledger/Properties.hs
```diff
@@ -0,0 +1,794 @@
...
+    DState (..),
+    PState (..),
+    RewardAccounts,
+    UTxOState (..),
+  )
+import Shelley.Spec.Ledger.STS.Utxo (UtxoEnv (..))
+import Shelley.Spec.Ledger.Tx (hashScript)
+import Shelley.Spec.Ledger.TxBody (DCert (..), DelegCert (..), Delegation (..), PoolParams (..))
+import Shelley.Spec.Ledger.UTxO (balance, makeWitnessVKey)
+import Test.Cardano.Ledger.Alonzo.Serialisation.Generators ()
```
nc6, line 95: I guess I'm a bit worried when I see the serialisation generators imported in the real generators, since they're intended to be completely "dumb", whereas I can believe that in this module, we actually rely on certain properties of them.

cardano-ledger-test/src/Test/Cardano/Ledger/Properties.hs
```diff
@@ -0,0 +1,794 @@
...
+{-# LANGUAGE DataKinds #-}
+{-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE LambdaCase #-}
+{-# LANGUAGE NamedFieldPuns #-}
+{-# LANGUAGE RecordWildCards #-}
+{-# LANGUAGE TupleSections #-}
+{-# LANGUAGE TypeApplications #-}
+{-# LANGUAGE TypeFamilies #-}
+
+module Test.Cardano.Ledger.Properties where
```
nc6, line 11: Are the generators in this module intended to be used elsewhere? If so, I'd prefer we split apart the generators (test infrastructure) from the properties themselves (which should not be included in the library)

nc6 (approved):
I think in general this looks great. I am a little worried however about the proliferation of generators and tests. This is now adding another new method atop the existing one and the one we've asked the Obsidian folks to build. At the moment, it doesn't seem to be testing anything the existing tests don't. So what's the plan here? If I were to add a new test, I would now struggle to identify where or how to add one, and this makes it much harder to survey exactly what we're testing.

cardano-ledger-test/src/Test/Cardano/Ledger/Properties.hs
lehins: You bring up a great point. We must not have multiple instances of the same type for the same class in the same codebase, this is a bug. So, if generators in serialization are dumb, they need to be fixed to be smarter. In fact it would be wise to aggregate `Arbitrary` instances in one place per each era. 

I can certainly work on that in a separate PR.

lehins:
@nc6 This particular task was for creating generators and properties for UTXOW for Alonzo. For context, this is the commit I started off of: 441b1a84a6a1eb87c0bec9c7d1e5d15df8cbc3c5, which I then expanded from. I took care of generators that create a valid transaction and state needed for LEDGER rules transition. So, now I can write some property tests for UTXOW and add the properties that Jared implemented in another PR. This is described as a TODO in the description for this PR along a few others.

> At the moment, it doesn't seem to be testing anything the existing tests don't.

If that was the case then logic in this PR wouldn't help us find the problem with redeemers for duplicate scripts certificates, wouldn't it? Probably because current tests use the same logic they should be testing, in particular `scriptsNeededFromBody` ;) Sarcasm aside, if you think that this task was not needed and our test framework already tests UTXOW then I am not sure why this task was even assigned to me or anyone else for that matter. If there is something missing that you think should be here, then please, I am all ears. Moreover, now reading in more depth the trace testing system, I can't find where it generates plutus scripts for rewards and certificates, which is present in this PR. If am mistaken please point me to the location in the codebase.

> this makes it much harder to survey exactly what we're testing.

Do you mean to say "it makes harder to see what have been tested and what hasn't"? I think setting up code coverage would help us with this. If you really mean exactly what you said, I'd argue that it is already quite hard to see what is being tested, because generators  are spread throughout eras/packages and arguing about data being generated is far from trivial, although limited time of my exposure to the codebase might have something to do with this :)

cardano-ledger-test/src/Test/Cardano/Ledger/Properties.hs
lehins: Generators in this module only work in `RWST GenEnv () GenState Gen` monad, so as it stands right now they can't be used anywhere else.

> This is now adding another new method atop the existing one and the one we've asked the Obsidian folks to build.

I believe this is orthogonal to what Obsidian folks are doing. This is intended for testing a particular component, rather than creating a framework that will test all transition rules from epoch to epoch. What do you mean by this: "atop the existing one"? I don't really see what is the bottom method it is building atop off?

> I'd prefer we split apart the generators (test infrastructure) from the properties themselves (which should not be included in the library)

I certainly agree. This module was originally named `Properties`, but I think it should be renamed to something like `Generators.Ledger` or something along those lines. I guess the problem I see is there no concrete structure or documented plan to what `cardano-ledger-test` should be. Has it been outlined anywhere?

JaredCorduan:
@nc6 any problems / issues with the existence of these tests is my fault, not @lehins, I gave him this task . The reason that I thought that these tests would be helpful is because when you are testing a single transaction at the `UTXOW` / `LEDGER` level, as opposed to a trace of blocks at the `CHAIN` level, the generation can be quite random. The trace generators are of course random, but they are much more massaged than these. And indeed these tests did catch something that the trace generation tests have yet to catch.

PR #2437, opened 2021-08-26
more events, AnnotatedBlock

goolord:


nc6:
What's the logic behind `AnnotatedBlock`? Having another top-level function to call is annoying, because it requires plumbing all the way through the system.

goolord:
theres discussion here https://github.com/input-output-hk/cardano-db-sync/pull/668/ and AnnotatedBlock seemed relatively uncontroversial. I think it could probably be an event or something so we don't need to do all of this plumbing again

nc6:
Yep, having this data as an event makes lots of sense. I'd rather avoid the current implementation.

nc6:
This somehow has a very large number of unrelated changes integrated into it... please can you rebase just the relevant parts onto master?

goolord:
oh yeah, i don't know what's up with that. i'll fix that

eras/shelley/impl/src/Cardano/Ledger/Shelley/API/Validation.hs
```diff
@@ -64,7 +84,8 @@ class
     Environment (Core.EraRule "BBODY" era) ~ STS.BbodyEnv era,
     State (Core.EraRule "BBODY" era) ~ STS.BbodyState era,
     Signal (Core.EraRule "BBODY" era) ~ (Block BHeaderView era),
-    ToCBORGroup (TxSeq era)
+    ToCBORGroup (TxSeq era),
+    Coercible (Event (Core.EraRule "BBODY" era)) (STS.BbodyEvent era)
```
nc6: I appreciate that this is potentially more general, but everywhere else we've used equality constraints, so I'd prefer to stick with that here unless there really _is_ an isomorphic type.

eras/shelley/impl/src/Cardano/Ledger/Shelley/API/Validation.hs
```diff
@@ -104,20 +125,36 @@ class
     m (EventReturnType ep (Core.EraRule "BBODY" era) (NewEpochState era))
   default applyBlockOpts ::
     forall ep m.
-    (EventReturnTypeRep ep, MonadError (BlockTransitionError era) m) =>
+    ( EventReturnTypeRep ep,
+      MonadError (BlockTransitionError era) m,
+      Era era,
+      HasField "inputs" (Core.TxBody era) (Set (TxIn (Crypto era))),
```
nc6: The point of these API methods is that they wrap up all of the era constraints such that they don't propagate to downstream users. So we cannot have additional constraints here.

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Bbody.hs
```diff
@@ -85,8 +89,25 @@ data BbodyPredicateFailure era
...
+    abEpochSlot :: !SlotNo, -- The slot within the epoch (starts at 0 for first slot of each epoch
+    abTimeStamp :: !UTCTime, -- The slot number converted to UTCTime
+    abEpochSize :: !EpochSize, -- Number of slots in current epoch
+    abTxs :: [AnnotatedTx] -- All fields in the superset of all block types
+  }
+
+data AnnotatedTx = AnnotatedTx
+  { atInputSum :: !Coin, -- Sum of the tx inputs
+    atOutSum :: !Coin, -- Sum of the tx outputs
+    txFees :: !Coin -- All fields in the superset of all tx types
```
nc6: Speaking with Erik, he would also like to have the actual resolved inputs here.

eras/shelley/impl/src/Cardano/Ledger/Shelley/API/Validation.hs
```diff
@@ -104,20 +125,36 @@ class
...
+      HasField "certs" (Core.TxBody era) (StrictSeq (DCert (Crypto era))),
+      HasField "wdrls" (Core.TxBody era) (Wdrl (Crypto era)),
+      HasField "_keyDeposit" (Core.PParams era) Coin,
+      HasField "_poolDeposit" (Core.PParams era) Coin
```
nc6: Unfortunately we cannot do this - the intention of this API is that it wraps up all of the constraints needed so they're not exposed to downstream clients. So there should be no new constraints here.

eras/shelley/impl/src/Cardano/Ledger/Shelley/API/Validation.hs
nc6: I appreciate this is more general, but for consistency with everything else I'd rather stick with the equality constraint.

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Bbody.hs
```diff
@@ -85,8 +89,25 @@ data BbodyPredicateFailure era
   | LedgersFailure (PredicateFailure (Core.EraRule "LEDGERS" era)) -- Subtransition Failures
   deriving (Generic)
 
-newtype BbodyEvent era
+data AnnotatedBlock = AnnotatedBlock
+  { -- abEra :: !Era
+    abEpochNo :: !EpochNo,
```
nc6:
```suggestion
  { abEpochNo :: !EpochNo,
```

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Bbody.hs
```diff
@@ -85,8 +89,25 @@ data BbodyPredicateFailure era
...
+    abSlotNo :: !SlotNo,
+    abEpochSlot :: !SlotNo, -- The slot within the epoch (starts at 0 for first slot of each epoch
+    abTimeStamp :: !UTCTime, -- The slot number converted to UTCTime
+    abEpochSize :: !EpochSize, -- Number of slots in current epoch
+    abTxs :: [AnnotatedTx] -- All fields in the superset of all block types
```
nc6, line 98: Don't understand the comment here

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Snap.hs
```diff
@@ -60,7 +68,11 @@ snapTransition = do
   TRC (lstate, s, _) <- judgmentContext
 
   let LedgerState (UTxOState utxo _ fees _) (DPState dstate pstate) = lstate
+      stake :: SnapShot (Crypto era)
       stake = stakeDistr utxo dstate pstate
+
+  tellEvent $ StakeDistEvent stake
```
nc6: This is rather a giant event, since it contains a full stake snapshot:

```
data SnapShot crypto = SnapShot
  { _stake :: !(Stake crypto),
    _delegations :: !(VMap VB VB (Credential 'Staking crypto) (KeyHash 'StakePool crypto)),
    _poolParams :: !(VMap VB VB (KeyHash 'StakePool crypto) (PoolParams crypto))
  }
  ```

This is a really large chunk of internal stuff, not really something that we want to make part of a public API. What's the use case for this full structure?

File: eras/shelley/impl/src/Cardano/Ledger/Shelley/API/Validation.hs [hunk 8]
```diff
@@ -254,3 +291,51 @@ deriving stock instance
...
+    annotateTx :: Core.Tx era -> STS.AnnotatedTx
+    annotateTx tx =
+      let txBody = getField @"body" tx
+          ledgerState = esLState $ nesEs state
+          pp_ = esPp $ nesEs state
+          u = _utxoState ledgerState
+          pools = _pParams $ _pstate $ _delegationState $ ledgerState
+          certs = Foldable.toList (getField @"certs" txBody)
+          ins =
+            Val.coin (balance @era (eval ((getField @"inputs" txBody) <| _utxo u)))
```
nc6, line 314: Doing this within the API, rather than during the appropriate ledger rule, means we end up resolving the inputs twice. The correct place for this would surely be in the `Utxo` rule?

nc6 (changes requested):
I guess I can see why you've taken this approach, but I think it's going to be better to have these as conventional events, each generated at the correct place. So we would have a bunch of `AnnotatedTx` events, plus a block-level one.

eras/shelley/impl/src/Cardano/Ledger/Shelley/API/Validation.hs
goolord: some instances don't compile when this is an equality constraint, it's necessary

goolord: Some instances won't compile if this is in equality constraint. This constraint will get floated outside of typeclass. Maybe it could be an `Embed` constraint instead, but the `Coercible` constraint is not superficially more general

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Bbody.hs
goolord: that's from https://github.com/input-output-hk/cardano-db-sync/blob/54852e7981da92c3ba089d3863977eeffdb862e6/doc/ledger-consensus-api.md

the `AnnotatedTx` is meant to contain fields from the `TxIn/TxOut/TxBody` of each i believe

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Snap.hs
goolord: ah yeah, this is not the appropriate subset we want here. this should be something more like https://github.com/input-output-hk/cardano-db-sync/blob/689253c7736dfdf2d1f63e05b692313817b9b1bd/cardano-sync/src/Cardano/Sync/Era/Shelley/Generic/StakeDist.hs#L41

eras/shelley/impl/src/Cardano/Ledger/Shelley/API/Validation.hs
goolord: Is this expensive? I can't imagine this `AnnotatedBlock` would be very useful if it was split up into several events. I don't see where in `Cardano.Ledger.Alonzo.Rules.Utxo.hs` would be more appropriate, I feel like I'm missing something

goolord:
closed in favor of https://github.com/input-output-hk/cardano-ledger/pull/2652

PR #2441, opened 2021-09-01
MIR event should filter to extant reward accounts.

nc6:
Previously we just returned the `InstantaneuousRewards`, but this
doesn't account for deregistered credentials.


JaredCorduan (approved):
LGTM!

PR #2442, opened 2021-09-01
Make Alonzo TxOut a sum type to save memory

JaredCorduan:
While investigating our options to save memory for the hashes, @redxaxder noticed that we can save memory in Alonzo by making the TxOut a ~GADT~ sum type. (I have regrets about this branch name now :) ).


lehins (approved):
Looks great!

PR #2443, opened 2021-09-02
Skeleton of passing redeemers in TxInfo

michaelpj:
@JaredCorduan this is as far as I've got, I'll see if I can do the conversion functionality after lunch, otherwise it's in your court!

The plutus commit is based on our `release/alonzo` branch, so should just be one commit on top of what you had previously, and all that does is add the extra field to `TxInfo`.

As we discussed, Jared, the type of `txInfoRedeemers` is `[(ScriptPurpose, Redeemer)]`.

(I'm targeting `plutus-staging` so we can review this PR independently without merging it into master, then we can merge `plutus-staging` into master if we decide to ship the change.)

### DIFF
--- a/alonzo/impl/src/Cardano/Ledger/Alonzo/TxInfo.hs
+++ b/alonzo/impl/src/Cardano/Ledger/Alonzo/TxInfo.hs
@@ -29,7 +29,7 @@ import Cardano.Ledger.Alonzo.TxBody
     wdrls',
   )
 import qualified Cardano.Ledger.Alonzo.TxBody as Alonzo (TxBody (..), TxOut (..))
-import Cardano.Ledger.Alonzo.TxWitness (TxWitness, unTxDats)
+import Cardano.Ledger.Alonzo.TxWitness (RdmrPtr, TxWitness (..), unRedeemers, unTxDats)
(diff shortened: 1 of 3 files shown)


alonzo/impl/src/Cardano/Ledger/Alonzo/TxInfo.hs
```diff
@@ -315,6 +272,20 @@ getWitVKeyHash =
...
+    HasField "wdrls" (Core.TxBody era) (Wdrl (Crypto era)),
+    HasField "certs" (Core.TxBody era) (StrictSeq (DCert (Crypto era)))
+  ) =>
+  (Core.TxBody era) ->
+  (RdmrPtr, (Data era, ExUnits)) ->
+  Maybe (P.ScriptPurpose, P.Redeemer)
+transRedeemer txb (ptr, (d, _)) =
+  case rdptrInv txb ptr of
+    SNothing -> Nothing
+    SJust sp -> Just (transScriptPurpose sp, undefined d)
```
JaredCorduan: oops, let me remove that

JaredCorduan:
The changes in this PR will appear in https://github.com/input-output-hk/cardano-ledger-specs/pull/2485.

PR #2445, opened 2021-09-03
Separating the protocol from the ledger

JaredCorduan:
This noisy :drum: PR creates an new package `cardano-protocol-tpraos` and moves the following into it:

* Block Header and associated types and helper functions
* Operational Certificate types and helper functions
* Overlay Schedule support
* The following rules: `PRTCL`, `OCERT`, `OVERLAY`, `Updn`


alonzo/impl/cardano-ledger-alonzo.cabal
```diff
@@ -61,6 +61,7 @@ library
     cardano-ledger-core,
     cardano-ledger-shelley-ma,
     cardano-prelude,
+    cardano-protocol-tpraos,
```
nc6, line 64: Ultimately I guess we'd like to drop this dependency.

cardano-protocol-tpraos/src/Cardano/Protocol/TPraos/BHeader.hs
```diff
@@ -0,0 +1,490 @@
...
+{-# LANGUAGE NamedFieldPuns #-}
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE PatternSynonyms #-}
+{-# LANGUAGE RankNTypes #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE StandaloneDeriving #-}
+{-# LANGUAGE TypeApplications #-}
+{-# OPTIONS_GHC -Wno-orphans #-}
+
+-- BlockNo is an orphan until the instance is added to cardano-base
```
nc6: Is this comment in the right place?

shelley/chain-and-ledger/executable-spec/src/Shelley/Spec/Ledger/STS/Chain.hs
```diff
@@ -53,6 +53,15 @@ import Cardano.Ledger.Serialization (ToCBORGroup)
 import Cardano.Ledger.Shelley.Constraints (UsesValue)
 import Cardano.Ledger.Slot (EpochNo)
 import qualified Cardano.Ledger.Val as Val
+import Cardano.Protocol.TPraos (PoolDistr (..))
+import Cardano.Protocol.TPraos.BHeader (BHeader, LastAppliedBlock (..), bHeaderSize, bhHash, bhbody, bheaderBlockNo, bheaderSlotNo, hBbsize, lastAppliedHash, prevHashToNonce)
```
nc6: Break over multiple lines?

nc6 (approved):
So I think we'll need to pull things apart further, to remove the dependency at all. But this certainly looks like a solid first step!

cardano-protocol-tpraos/src/Cardano/Protocol/TPraos/BHeader.hs
JaredCorduan: let me see if I can find a home for it where ormolu won't move it...

shelley/chain-and-ledger/executable-spec/src/Shelley/Spec/Ledger/STS/Chain.hs
JaredCorduan: oh wow, I didn't realize how long that one got :laughing:

JaredCorduan:
@nc6 do you think now is an okay time to merge this possibly disruptive PR?

nc6:
Yep, seems reasonable to me!

PR #2447, opened 2021-09-04
Test multiple, identicle Plutus-locked certs

JaredCorduan:
Document, by way of unit tests, the current behavior of using multiple, identical Plutus-locked stake credential certs in the same transaction. These are pretty odd scenarios that no one would probably ever intend to enact, namely 1) delegating from the same Plutus-locked stake address to the same pool more than once in a single transaction, and 2) de-registering a Plutus locked stake credential, then re-registering it, and de-registering it again, all in the same transaction.

In these cases, only one redeemer in the witnesses can be used for all occurrences of the identical certificate in the transaction, and it must point to the first occurrence of the certificate.

In the future, we may wish to just forbid such scenarios (ie not allow multiple, identical certs in the same transaction).

### DIFF
--- a/cardano-ledger-test/src/Test/Cardano/Ledger/Examples/TwoPhaseValidation.hs
+++ b/cardano-ledger-test/src/Test/Cardano/Ledger/Examples/TwoPhaseValidation.hs
@@ -1018,6 +1018,61 @@ utxoStEx10 ::
   UTxOState era
 utxoStEx10 pf = UTxOState (utxoEx10 pf) (Coin 0) (Coin 5) def
 
+-- ====================================================================================
+--  Example 11: A transaction with multiple identical certificates
+-- ====================================================================================
+
+multipleEqualCertsRedeemers :: Era era => Redeemers era
+multipleEqualCertsRedeemers =
+  Redeemers $
+    Map.fromList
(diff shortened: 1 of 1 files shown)


cardano-ledger-test/src/Test/Cardano/Ledger/Examples/TwoPhaseValidation.hs
```diff
@@ -1018,6 +1018,61 @@ utxoStEx10 ::
   UTxOState era
 utxoStEx10 pf = UTxOState (utxoEx10 pf) (Coin 0) (Coin 5) def
 
+-- ====================================================================================
+--  Example 11: A transaction with multiple identicle certificates
```
lehins:
```suggestion
--  Example 11: A transaction with multiple identical certificates
```

lehins (approved):

cardano-ledger-test/src/Test/Cardano/Ledger/Examples/TwoPhaseValidation.hs
JaredCorduan: there were a few of these :grimacing: , thanks!

PR #2449, opened 2021-09-08
Add neededTxInsForBlock function for UTxO on-disk

JaredCorduan:
The validity of any individual block depends only on a subset of the UTxO stored in the ledger state. We introduce a new function `neededTxInsForBlock` that take a block and returns the transaction inputs corresponding to the required UTxO for the given Block.

This function will be used by the consensus layer to enable storing the UTxO on disk. In particular, given a block, the consensus layer will use 'neededTxInsForBlock' to retrived the needed UTxO from disk and present only those to the ledger.

A property test is included, which tests that: if you partition the UTxO into those that are needed and those that are irrelevant, then running the `LEDGER` rule on the the full UTxO results in the same mapping as running the `LEDGER` rule on the the restricted UTxO and then combining the results with the irrelevant UTxO (we also check that the irrelevant UTxO are disjoint from the updated restricted UTxO).

### DIFF
--- a/shelley/chain-and-ledger/executable-spec/src/Shelley/Spec/Ledger/BlockChain.hs
+++ b/shelley/chain-and-ledger/executable-spec/src/Shelley/Spec/Ledger/BlockChain.hs
@@ -27,6 +27,7 @@ module Shelley.Spec.Ledger.BlockChain
     bbHash,
     bBodySize,
     slotToNonce,
+    neededTxInsForBlock,
     -- accessor functions
     bheader,
     bbody,
@@ -100,19 +101,23 @@ import Data.ByteString (ByteString)
 import qualified Data.ByteString.Char8 as BS
 import qualified Data.ByteString.Lazy as BSL
 import Data.Coerce (coerce)
+import Data.Foldable (toList)
 import Data.Map.Strict (Map)
 import qualified Data.Map.Strict as Map
 import Data.Sequence (Seq)
 import qualified Data.Sequence as Seq
 import Data.Sequence.Strict (StrictSeq)
 import qualified Data.Sequence.Strict as StrictSeq
+import Data.Set (Set)
+import qualified Data.Set as Set
 import Data.Typeable
 import GHC.Generics (Generic)
 import GHC.Records (HasField (..))
 import NoThunks.Class (AllowThunksIn (..), NoThunks (..))
 import Numeric.Natural (Natural)
 import Shelley.Spec.Ledger.EpochBoundary (BlocksMade (..))
-import Shelley.Spec.Ledger.Tx (Tx, segwitTx)
+import Shelley.Spec.Ledger.Tx (Tx, TxIn (..), segwitTx)
+import Shelley.Spec.Ledger.UTxO (txid)
 
 data TxSeq era = TxSeq'
   { txSeqTxns' :: !(StrictSeq (Tx era)),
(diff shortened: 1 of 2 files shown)


shelley/chain-and-ledger/executable-spec/src/Shelley/Spec/Ledger/BlockChain.hs
```diff
@@ -850,3 +855,26 @@ incrBlocks isOverlay hk b'@(BlocksMade b)
...
+  ) =>
+  Block era ->
+  Set (TxIn (Crypto era))
+neededTxInsForBlock (Block' _ txsSeq _) = Set.filter isNotNewInput allTxIns
+  where
+    getTxIn = getField @"inputs" . getField @"body"
+    txs = Era.fromTxSeq txsSeq
+    allTxIns = foldl' (\acc tx -> acc `Set.union` (getTxIn tx)) mempty txs
+    newTxIds = map (txid . getField @"body") (toList txs)
+    isNotNewInput (TxIn txID _) = txID `notElem` newTxIds
```
lehins: One thing that is bad here is that `notElem` O(n), instead of `Set.notMember` which O(log n), otherwise seems good. At the same time I think rewriting it this way is a bit clearer:

```suggestion
    txBodies = map (getField @"body") $ toList $ Era.fromTxSeq txsSeq
    allTxIns = Set.unions $ map txins txBodies
    newTxIds = Set.fromList $ map txid txBodies
    isNotNewInput (TxIn txID _) = txID `Set.notMember` newTxIds
```

shelley/chain-and-ledger/shelley-spec-ledger-test/test/Test/Shelley/Spec/Ledger/Rules/TestChain.hs
```diff
@@ -600,6 +602,28 @@ preserveOutputsTx SourceSignalTarget {source = chainSt, signal = block} =
...
+  conjoin $
+    map outputPreserved $
+      zip (sourceSignalTargets ledgerTrFull) (sourceSignalTargets ledgerTrRestr)
+  where
+    (_, ledgerTrFull) = ledgerTraceFromBlock @era @ledger chainSt block
+    (UTxO irrelevantUTxO, ledgerTrRestr) =
+      ledgerTraceFromBlockWithRestrictedUTxO @era @ledger chainSt block
+    outputPreserved
+      ( SourceSignalTarget {target = (UTxOState {_utxo = (UTxO uFull)}, _), signal = _},
+        SourceSignalTarget {target = (UTxOState {_utxo = (UTxO uRestr)}, _), signal = _}
```
lehins: I might be wrong, but I don't think you need to bind signal field you don't need it, also redundant parenthesis :
```suggestion
      ( SourceSignalTarget {target = (UTxOState {_utxo = UTxO uFull}, _)},
        SourceSignalTarget {target = (UTxOState {_utxo = UTxO uRestr}, _)}
```

shelley/chain-and-ledger/shelley-spec-ledger-test/test/Test/Shelley/Spec/Ledger/Rules/TestChain.hs
```diff
@@ -600,6 +602,28 @@ preserveOutputsTx SourceSignalTarget {source = chainSt, signal = block} =
...
+        ) =
+        uRestr `Map.disjoint` irrelevantUTxO && uFull == uRestr `Map.union` irrelevantUTxO
```
lehins: I always try to use QuickChecks functions, since then we'll get the output printed if it fails:

```suggestion
        (uRestr `Map.disjoint` irrelevantUTxO) .&&. uFull === (uRestr `Map.union` irrelevantUTxO)
```

shelley/chain-and-ledger/shelley-spec-ledger-test/test/Test/Shelley/Spec/Ledger/Rules/TestChain.hs
```diff
@@ -907,6 +931,31 @@ ledgerTraceFromBlock chainSt block =
...
+  ( UTxO irrelevantUTxO,
+    runShelleyBase $
+      Trace.closure @ledger ledgerEnv ledgerSt0' txs
+  )
+  where
+    (_tickedChainSt, ledgerEnv, ledgerSt0, txs) = ledgerTraceBase chainSt block
+    tx_ins = neededTxInsForBlock block
+    (utxoSt, delegationSt) = ledgerSt0
+    utxo = unUTxO . _utxo $ utxoSt
+    (relevantUTxO, irrelevantUTxO) = Map.partitionWithKey (const . (`Set.member` tx_ins)) utxo
```
lehins: Haskell Camels :wink:  
```suggestion
    txIns = neededTxInsForBlock block
    (utxoSt, delegationSt) = ledgerSt0
    utxo = unUTxO . _utxo $ utxoSt
    (relevantUTxO, irrelevantUTxO) = Map.partitionWithKey (const . (`Set.member` txIns)) utxo
```

lehins (changes requested):
Few suggestions otherwise LGTM

JaredCorduan:
@lehins thanks for the review! I made all of the changes you suggested.

lehins (approved):

PR #2451, opened 2021-09-10
At PV6, translate time for Plutus correctly.

nc6:
The relevant part of this change is in TxInfo.hs (and the enabling switch in
HardForks.hs). The rest is required to plumb through the protocol parameters to
allow this switch.


JaredCorduan (approved):
looks great!

investorguddi:
Cool

PR #2453, opened 2021-09-11
upgrading cardano base

JaredCorduan:
This brings in the new `PackedBytes` implementation of the `Hash` type for Blake2b hashes.

### DIFF
--- a/cardano-protocol-tpraos/src/Cardano/Protocol/TPraos/BHeader.hs
+++ b/cardano-protocol-tpraos/src/Cardano/Protocol/TPraos/BHeader.hs
@@ -97,11 +97,10 @@ import Cardano.Protocol.TPraos (ProtVer (..))
 import Cardano.Protocol.TPraos.OCert (OCert (..))
 import Cardano.Slotting.Slot (WithOrigin (..))
 import Control.DeepSeq (NFData)
+import qualified Data.ByteString as BS
 import qualified Data.ByteString.Builder as BS
 import qualified Data.ByteString.Builder.Extra as BS
-import qualified Data.ByteString.Char8 as BS
 import qualified Data.ByteString.Lazy as BSL
-import Data.Coerce (coerce)
 import Data.Ratio ((%))
 import Data.Typeable
 import Data.Word (Word64)
@@ -157,8 +156,6 @@ instance
         decodeNull
         pure GenesisHash
       _ -> BlockHash <$> fromCBOR
-
-instance NFData BlockNo -- TODO remove once instance exists in cardano-base
 
 deriving newtype instance CC.Crypto crypto => FromCBOR (HashHeader crypto)
 
@@ -337,8 +334,17 @@ bhHash ::
 bhHash = HashHeader . Hash.hashWithSerialiser toCBOR
 
 -- | HashHeader to Nonce
+-- What is going on here?
+-- This is here because the surrounding code is parametrized in the hash algorithm used,
+-- but the nonce is hard-coded to Blake2b_256.
+-- We require the nonce to have the right length (the size of a Blake2b_256 hash), so
+-- if the hash size differs, we pad or remove bytes accordingly.
 hashHeaderToNonce :: HashHeader crypto -> Nonce
-hashHeaderToNonce = Nonce . coerce
+hashHeaderToNonce (HashHeader h) = case Hash.hashFromBytes bytes of
+  Nothing -> Nonce (Hash.castHash (Hash.hashWith id bytes))
+  Just hash -> Nonce hash
+  where
+    bytes = Hash.hashToBytes h
 
 prevHashToNonce ::
   PrevHash crypto ->
--- a/alonzo/impl/src/Cardano/Ledger/Alonzo/TxInfo.hs
+++ b/alonzo/impl/src/Cardano/Ledger/Alonzo/TxInfo.hs
@@ -12,7 +12,7 @@ module Cardano.Ledger.Alonzo.TxInfo where
 -- =============================================
 
 import Cardano.Binary (FromCBOR (..), ToCBOR (..), decodeFull', serialize')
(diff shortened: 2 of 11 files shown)


cardano-protocol-tpraos/src/Cardano/Protocol/TPraos/BHeader.hs
```diff
@@ -337,8 +335,18 @@ bhHash ::
...
+-- We require the nonce to have the right length (the size of a Blake2b_256 hash), so
+-- if the hash size differs, we pad or remove bytes accordingly.
 hashHeaderToNonce :: HashHeader crypto -> Nonce
-hashHeaderToNonce = Nonce . coerce
+hashHeaderToNonce (HashHeader h) = case Hash.hashFromBytes bytes of
+  Nothing -> Nonce (scale bytes)
+  Just hash -> Nonce hash
+  where
+    bytes = Hash.hashToBytes h
+    scale b = fromMaybe (error "impossible") $ Hash.hashFromBytes (BS.take 32 (b <> BS.replicate 32 0))
```
lehins: I still don't think this is a safe approach, because it looks to me like a vulnerability waiting to happen. It would be much safer to ensure the length by hashing the mismatched length hash instead of padding/cropping it:
```suggestion
hashHeaderToNonce (HashHeader h) = case Hash.hashFromBytes bytes of
  Nothing -> Nonce (Hash.castHash (Hash.hashWith id bytes))
  Just hash -> Nonce hash
  where
    bytes = Hash.hashToBytes h
```

lehins (changes requested):
As it stands this PR is correct, but I would highly recommend avoiding cropping/padding hashes, especially since there is a safer solution.

cardano-protocol-tpraos/src/Cardano/Protocol/TPraos/BHeader.hs
JaredCorduan: We could do that instead. It's a bit weird to force the tests to hash things twice, but I suppose it is not really a performance concern for the tests since blake2b is fast and this is only for the block headers.

lehins: Yes, I am not really worried about tests, performance implications for tests will be minuscule. 

I am more worried about the future of this code. Any change to `HashHeader` algorithm could have serious implications. For instance changing the algorithm for the header in the future to a different one with smaller digest size would reduce security of  the nonce. I know what people usually say: this is never going to happen, until it does.

Padding and cropping of anything in crypto should be thought through very carefully. However hashing twice will rarely result in a less secure hash value. 

As an extra side note, with this implementation: `fromMaybe (error "impossible") $ Hash.hashFromBytes (BS.take 32 (b <> BS.replicate 32 0))` all it takes is a change to hashing algorithm used in the Nonce that has different digest size and  the impossible becomes possible. Using `hashSize` would have alleviated it.

JaredCorduan: Is this definitely safer? It would hide a lack of entropy, whereas just padding is more obvious and detectable.

JaredCorduan: I'll apply your suggestion, and make a ticket to consider making `Nonce` parametric in the underlying crypto. Hard coding Blake2b is awkward, and I suspect there is no good reason for it.

lehins: > Is this definitely safer?

Of course it is! What do you think more secure this:
```
7d17d7742385e89a8003ad91bce6f43f00000000000000000000000000000000
```
or this:
```
91cb0802981d2aaa52e26537cc1111a59b1ef521addeb52f8bd57134e503ceb6
```

Further more nonces are xored, right? When they are zeros reveal the previous value of the hash of a nonce, which could potentially lead to trouble too.

> It would hide a lack of entropy, whereas just padding is more obvious and detectable.

It is not about entropy, but about ability to reproduce a collision, i.e. the same hash value from some input. It is almost  as hard to go from the latter to former as if there were no zeros at the end, precisely because secure hashes are non-invertible. Even if an attacker is aware that there are zeros at the end of the input. If there are only a few first bytes are non-zero, then of course you could just brute force and precompute all hashes, but even if 128 bits are non-zero then precomputing becomes practically infeasible.

JaredCorduan: > If there are only a few first bytes are non-zero, then of course you could just brute force and precompute all hashes, but even if 128 bits are non-zero then precomputing becomes practically infeasible.

Yep, this is the kind of scenario that I was worried about. The testing the header hash is 80 bits, if that was to ever somehow end up outside of testing, it might look random after being black2b hashed.

lehins: Even 80bits is a lot for brute force. Also think about this case once again, but without the extra hashing. If it did leak somehow, would you rather have only 80bits and the rest zeros opening the system to an obvious attack or the hash of that value, which would still be very hard to crack?

> It might looked random after being black2b hashed.

A 256bit hash is more random than 80bits padded with zeros to 256bits, wouldn't you agree?

JaredCorduan: That's fair. I've already applied your suggestion.

I look at a lot of hashes _before_ things go to mainnet. So does QA. Consistent zeros at the end of a hash is an huge and obvious sign that something is wrong, and I found that thought comforting. Of course I realize that we cannot depend on someone noticing such a thing.

The real plan, though, is to make the nonce parametric in the hash https://github.com/input-output-hk/cardano-ledger-specs/issues/2454 to avoid all of this (provided we can't find any reason for it having been hard coded in the first place).

lehins (approved):

PR #2435, opened 2021-08-25
Apply hlint sugestions to alonzo and shelley

lehins:
This PR applies most suggestions from hlint, which mostly just:

* redundant parenthesis
* unused extensions
* eta reduction
* single argument data types changed to newtypes

Feel free to just mention on any line if the change needs to be reversed, I am not going to push back.

There are a few minor, but notable improvements from me:

* `explain_plutus_failure` -> `explainPlutusFailure` rename with deprecation (although quick search didn't show any usage from outside)
* pretty printing module got most fixups, in particular instances got formatted to be more consistent
* some long lines that I found which didn't fit on a screen without wrapping around got reformatted.
* Removed usage of deprecated `"isValidating"` in a couple of places
* haddock improvements

**Important** - none of the changes in this PR affects the semantics of any functions (there might be a bit more strictness due to  newtypes, but that should be a desired improvement IMHO)

I'll be happy to apply this sort of cleanup to the rest of the repository once this PR is merged.

PR #2455, opened 2021-09-14
Apply hlint suggestions and other cleanup to shelley-ma

lehins:
HLint round 2 follow up to #2435. This time cleanup was done on shelley-ma codebase including tests


JaredCorduan (approved):
thanks again! looks like ormolu still needs appeasing.

lehins:
> looks like ormolu still needs appeasing.

Actually looks like a nix build error? :open_mouth:

JaredCorduan:
> > looks like ormolu still needs appeasing.
> 
> Actually looks like a nix build error?

indeed. I re-ran it.

nc6 (approved):

PR #2456, opened 2021-09-14
Fix transaction cddl to include IsValid

JaredCorduan:
When we changed the mempool serialization of the transaction, we did not update the CDDL. cc @SebastienGllmt

### DIFF
--- a/alonzo/test/cddl-files/alonzo.cddl
+++ b/alonzo/test/cddl-files/alonzo.cddl
@@ -13,6 +13,7 @@ block =
 transaction =
   [ transaction_body
   , transaction_witness_set
+  , bool
   , auxiliary_data / null
   ]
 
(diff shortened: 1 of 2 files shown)


alonzo/test/cddl-files/alonzo.cddl
```diff
@@ -13,6 +13,7 @@ block =
 transaction =
   [ transaction_body
   , transaction_witness_set
+  , bool
```
SebastienGllmt, line 16: I don't think this is actually correct because Alonzo is backwards compatible which means that wallets that didn't upgrade anything for Alonzo are still working. Transactions created by these wallets aren't adding this field and so I assume it means that the node actually accepts transactions that don't specific `isValid` in their transaction. Am I missing something maybe?

JaredCorduan: I would need to check with the consensus folks to be sure, but I believe Mary transactions get translated to Alonzo by:

https://github.com/input-output-hk/cardano-ledger-specs/blob/0eb9b38bb125ec8eedf0f7aabf9e09f9b7bf79f7/alonzo/impl/src/Cardano/Ledger/Alonzo/Translation.hs#L106-L124

SebastienGllmt (approved):

nc6 (approved):
LGTM

PR #2486, opened 2021-09-16
add alonzo min-value calc

polinavino:
Example calculations of Alonzo min-values for documentation

Also incorporates the PR https://github.com/input-output-hk/cardano-ledger-specs/pull/2492


doc/explanations/min-utxo-alonzo.rst
```diff
@@ -0,0 +1,99 @@
...
+| ``coinsPerUTxOWord`` (in lovelace)       |34,482               |
++------------------------------------------+---------------------+
+| ``dataHashSize (dh)``, ``dh = Nothing``  |0                    |
++------------------------------------------+---------------------+
+| ``dataHashSize (dh)``, ``dh <> Nothing`` |10                   |
++------------------------------------------+---------------------+
+
+** NO datum hash: **
+
++---------------------+-----------------+-----------------+-------------------+------------------+------------------+---------------------------------+
```
WhatisRT: I think this table also needs an entry for the entire size of the output. At the top, this document talks about how the size of the full output is computed, but down here it's all about the size of Value, which I found confusing.

doc/explanations/min-utxo-alonzo.rst
```diff
@@ -0,0 +1,99 @@
...
+|                     | asset name (i)  | asset name (ii) | asset names (iii) | name (iv)        | name for each (v)| names between them (total) (vi) |
++---------------------+-----------------+-----------------+-------------------+------------------+------------------+---------------------------------+
+| size of value       | 11              | 12              | 15                | 16               | 17               | 173                             |
++---------------------+-----------------+-----------------+-------------------+------------------+------------------+---------------------------------+
+| ``minUTxO``         | 1,310,316       | 1,344,798       | 1,448,244         | 1,482,726        | 1,517,208        | 6,896,400                       |
++---------------------+-----------------+-----------------+-------------------+------------------+------------------+---------------------------------+
+| ``minUTxO`` (in ada)| 1.310316        | 1.344798        | 1.448244          | 1.482726         | 1.517208         | 6.896400                        |
```
WhatisRT: I'm not sure we need the same number twice

doc/explanations/min-utxo-alonzo.rst
```diff
@@ -0,0 +1,99 @@
...
++---------------------+-----------------+--------------------+------------------+
+| size of value       | 11              |  26                | 16               |
++---------------------+-----------------+--------------------+------------------+
+| ``minUTxO``         | 1,655,136       | 2,172,366          | 1,827,546        |
++---------------------+-----------------+--------------------+------------------+
+| ``minUTxO`` (in ada)| 1.655136        |  2.172366          | 1.827546         |
++---------------------+-----------------+--------------------+------------------+
+
+
+* (i) : ``6 + FLOOR(((1 * 12) + 0 + (1 * 28) + 7)/8, 1) = 11``
```
WhatisRT, line 91: These calculations are not explained at all by the above document. In fact, since there is a formula above, the reader could think that these are examples of the formulas above (which was my first thought!)

doc/explanations/min-utxo-alonzo.rst
```diff
@@ -0,0 +1,99 @@
+Min-Ada-Value Calculation in Alonzo
+===================================
+
+The calculation
+#################
+
+There is no longer a ``minUTxOValue`` protocol parameter in Alonzo, it was
+deprecated at the end of the Mary era. However, the requirement that each
```
WhatisRT, line 8: The sentence starting with 'However' is weird. Something needs to be fixed here

doc/explanations/min-utxo-alonzo.rst
```diff
@@ -0,0 +1,99 @@
...
+
+``utxoEntrySize (txout) * coinsPerUTxOWord``
+
+where
+
+``utxoEntrySize (txout) = utxoEntrySizeWithoutVal + size (v) + dataHashSize (dh)``
+
+The ``v`` variable is the token bundle inside the ``txout`` output, and
+``dh`` is the hash of the datum that is contained in the output. If there is
+no datum, the hash is represented by ``Nothing``, and takes up ``0`` space.
```
WhatisRT: This explains one branch of `dataHashSize` but not the other (which is given by the table below). Either both should be explained here, or none (and referring to the table below)

WhatisRT (changes requested):

doc/explanations/min-utxo-alonzo.rst
polinavino: the size calculation is explained in the Mary document (and should stay there for non-duplication)

WhatisRT: I'm not talking about `size`, but about `dataHashSize`. This paragraph gives half of its definition, and it should either be all of it, or none of it.

doc/explanations/min-utxo-alonzo.rst
```diff
@@ -92,8 +97,24 @@ The following table gives the values of the constants used in the calculation ab
 
 * (vi) : ``6 + FLOOR(((96 * 12) + 96 + (3 * 28) + 7)/8, 1) = 173``
 
-* (vii) : ``6 + FLOOR (((1 * 12) + 32 + (1 * 28) + 7) / 8) = 15``
```
WhatisRT: Was this removed by accident?

polinavino: it's both wrong and a repeat of (i)

WhatisRT (approved):
LGTM

PR #2487, opened 2021-09-16
RetiredPools event

goolord:
emits RetiredPools which contains 2 maps from the retired pools to the amount refunded and the amount unclaimed

### DIFF
--- a/shelley/chain-and-ledger/executable-spec/src/Shelley/Spec/Ledger/STS/PoolReap.hs
+++ b/shelley/chain-and-ledger/executable-spec/src/Shelley/Spec/Ledger/STS/PoolReap.hs
@@ -144,7 +144,7 @@ poolReapTransition = do
     let rewardAcntsWithPool =
           Map.foldlWithKey'
             ( \acc sp (ra, coin) ->
-                Map.insertWith (<>) (getRwdCred ra) (Map.singleton sp coin) acc
+                Map.insertWith (Map.unionWith (<>)) (getRwdCred ra) (Map.singleton sp coin) acc
             )
             Map.empty
             rewardAcnts_
--- a/shelley/chain-and-ledger/executable-spec/src/Shelley/Spec/Ledger/STS/PoolReap.hs
+++ b/shelley/chain-and-ledger/executable-spec/src/Shelley/Spec/Ledger/STS/PoolReap.hs
@@ -71,8 +71,8 @@ data PoolreapPredicateFailure era -- No predicate failures
   deriving (Show, Eq, Generic)
 
 data PoolreapEvent era = RetiredPools
-  { refundPools :: Map.Map (KeyHash 'StakePool (Crypto era)) Coin,
-    unclaimedPools :: Map.Map (KeyHash 'StakePool (Crypto era)) Coin
+  { refundPools :: Map.Map (Credential 'Staking (Crypto era)) (Map.Map (KeyHash 'StakePool (Crypto era)) Coin),
+    unclaimedPools :: Map.Map (Credential 'Staking (Crypto era)) (Map.Map (KeyHash 'StakePool (Crypto era)) Coin)
   }
 
 instance NoThunks (PoolreapPredicateFailure era)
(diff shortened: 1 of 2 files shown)


JaredCorduan (approved):
Looks good to me! Out of curiosity, the type annotations were not necessary, right? just for clarity?

goolord:
yes, it was just to help me grok this

shelley/chain-and-ledger/executable-spec/src/Shelley/Spec/Ledger/STS/PoolReap.hs
```diff
@@ -65,6 +70,11 @@ deriving stock instance
 data PoolreapPredicateFailure era -- No predicate failures
   deriving (Show, Eq, Generic)
 
+data PoolreapEvent era = RetiredPools
+  { refundPools :: Map.Map (KeyHash 'StakePool (Crypto era)) Coin,
```
nc6: I don't believe this has the information we need, since this doesn't identify _who_ will be credited with the rewards

nc6 (dismissed):
I don't think this has the needed information. Having `refunds` and `mRefunds` would I think be more useful, since these associate with the accounts they are paid to. Even better would be something like
```
refunds :: Map (Credential 'Staking crypto) (Map (KeyHash 'StakePool crypto) Coin)
```

This would allow DB-sync to recreate all the needed information

shelley/chain-and-ledger/executable-spec/src/Shelley/Spec/Ledger/STS/PoolReap.hs
JaredCorduan: does @erikd want the reward account, and not the stake pool id?

shelley/chain-and-ledger/executable-spec/src/Shelley/Spec/Ledger/STS/PoolReap.hs
```diff
@@ -101,25 +112,40 @@ instance
...
       refunded = fold $ Map.elems refunds
       unclaimed = fold $ Map.elems mRefunds
 
+  tellEvent $
+    let refundPools' = Map.mapMaybe (\ra -> Map.lookup (getRwdCred ra) refunds) rewardAcnts
+        unclaimedPools' = Map.mapMaybe (\ra -> Map.lookup (getRwdCred ra) mRefunds) rewardAcnts
+     in RetiredPools
+          { refundPools = refundPools',
+            unclaimedPools = unclaimedPools'
+          }
```
lehins: Just a minor suggestion, let bindings seem superfluous:
```suggestion
     RetiredPools
        { refundPools = Map.mapMaybe (\ra -> Map.lookup (getRwdCred ra) refunds) rewardAcnts
          unclaimedPools = Map.mapMaybe (\ra -> Map.lookup (getRwdCred ra) mRefunds) rewardAcnts
        }
```

goolord:
ok, refunds and unclaimed are supplied as `Map.Map (Credential 'Staking (Crypto era)) (Map.Map (KeyHash 'StakePool (Crypto era)) Coin)` now

shelley/chain-and-ledger/executable-spec/src/Shelley/Spec/Ledger/STS/PoolReap.hs
```diff
@@ -101,25 +112,51 @@ instance
...
+          Map.foldlWithKey'
+            ( \acc sp (ra, coin) ->
+                Map.insertWith (Map.unionWith (<>)) (getRwdCred ra) (Map.singleton sp coin) acc
+            )
+            Map.empty
+            rewardAcnts_
+        (refundPools', unclaimedPools') =
+          Map.partitionWithKey
+            (\k _ -> eval (k ∈ dom (_rewards ds)))
+            rewardAcntsWithPool
```
goolord, line 154: pretty sure this is at least an order of magnitude slower than it needs to be, but it was the easiest way for my fried brain to reason about constructing this at the time. might be worth revisiting later : )

PR #2491, opened 2021-09-17
move TICKN rule to tpraos and CHAIN to tests

### LINKED ISSUE #2458: Move TICKN
Move the `TICKN` rule into the `cardano-protocol-tpraos` package.

### LINKED ISSUE #2461: Move CHAIN
Move the `CHAIN` rule into a test package.

JaredCorduan:
This is more effort towards separating out the protocol concerns from the ledger code.

Moving the CHAIN rule was a bit involved, due to some parts of the module needing to stay in the main package:

  * The "Ada Pots" code, which provides some accounting support.
  * The "chainChecks" code, which provides some checks that are use in the Shelley.Spec.Ledger.API.Protocol module. In particular, `chainChecks` was using the `CHAIN` predicate failure type, so I had make a new type for `chainChecks` which includes only the failures it uses, and re-used this type for the `CHAIN` predicate failure.

I removed some of the `CHAIN` types from the API, namely `CHAIN` and `ChainState`. I also removed `initialShelleyState`.

closes #2458
closes #2461


shelley/chain-and-ledger/executable-spec/src/Cardano/Ledger/Chain.hs
```diff
@@ -0,0 +1,159 @@
...
+    DPState (..),
+    DState (..),
+    EpochState (..),
+    LedgerState (..),
+    UTxOState (..),
+  )
+import Shelley.Spec.Ledger.PParams (ProtVer (..))
+import Shelley.Spec.Ledger.UTxO (balance)
+
+data ChainChecksData = ChainChecksData
```
nc6: Maybe call this `ChainChecksPParms` to make it clear that this is a subset of the protocol parameters?

shelley/chain-and-ledger/executable-spec/src/Cardano/Ledger/Chain.hs
```diff
@@ -0,0 +1,159 @@
...
+  unless (fromIntegral (bHeaderSize bh) <= ccMaxBHSize ccd) $
+    throwError $
+      HeaderSizeTooLargeCHAIN (fromIntegral $ bHeaderSize bh) (ccMaxBHSize ccd)
+  unless (hBbsize (bhbody bh) <= ccMaxBBSize ccd) $
+    throwError $
+      BlockSizeTooLargeCHAIN (hBbsize (bhbody bh)) (ccMaxBBSize ccd)
+  where
+    (ProtVer m _) = ccProtocolVersion ccd
+
+data AdaPots = AdaPots
```
nc6: Do we actually use this anywhere? If so, I'd suggest that putting it into the wallet API module or similar might be a better place. At the moment I can only find its usage in tests.

nc6 (approved):

shelley/chain-and-ledger/executable-spec/src/Cardano/Ledger/Chain.hs
JaredCorduan: It's probably only in tests. I'll move it to the tests if so, and otherwise to same API module, thanks!

JaredCorduan: db-sync uses it, I'll move it to the wallet API module

PR #2494, opened 2021-09-21
TxInfo chapter

polinavino:


alonzo/formal-spec/txinfo.tex
```diff
@@ -3,5 +3,232 @@ \section{TxInfo Construction}
 
 This section specifies exactly what parts of the transaction and ledger
 state are used by the $\fun{txInfo}$ function to construct the
-$\TxInfo$ term that gets passed as an
-argument to the Plutus interpreter.
+$\TxInfo$ term that, together with the script purpose for which the script
```
WhatisRT: element, not term

alonzo/formal-spec/txinfo.tex
```diff
@@ -3,5 +3,232 @@ \section{TxInfo Construction}
...
+is being run, gets passed as a $\Data$ argument to the Plutus interpreter.
+
+The functions in Figure \ref{fig:txinfo-funcs} are needed to build a $\Data$ summary of a transaction.
+In this Section we use the notation $\type{P.PlutusType}$ for the Plutus type
+$\type{PlutusType}$ (specified in the Plutus library) to distinguish it from ledger types.
+The type $\type{P.ScriptContext}$ represents
+the script context. The function $\type{P.toData}$ converts a Plutus term
```
WhatisRT: same as above

alonzo/formal-spec/txinfo.tex
```diff
@@ -3,5 +3,232 @@ \section{TxInfo Construction}
...
+into into a $\Data$ term.
+
+$\fun{txInfo}$ summarizes all the necessary transaction and chain state info
```
WhatisRT: info -> information

alonzo/formal-spec/txinfo.tex
```diff
@@ -3,5 +3,232 @@ \section{TxInfo Construction}
...
+    The $\Language$ argument
+    is required because different languages have different expectations of the
+    format and contents of the $\TxInfo$ summary.
+
+    Note that $\fun{txInfo}$ has a $\UTxO$ argument. Even though the full ledger UTxO
+    is passed to it, we define this function in such a way that the only
+    entries in the ledger UTxO map that a Plutus script
+    actually sees via the argument $\fun{txInfo}$ builds are the ones corresponding to the transaction
+    inputs (ie. its realized inputs). This is done in order to maintain
+    deterministic evaluation. For details, see~\ref{sec:txinfo}.
```
WhatisRT: This link points to the current section :) Also, this is more about locality than about determinism.

alonzo/formal-spec/txinfo.tex
```diff
@@ -3,5 +3,232 @@ \section{TxInfo Construction}
...
+
+    $\fun{valContext}$ constructs the \emph{validation context}, also referred to as the
+    script context. A validation context is
+    a $\Data$ term which encodes both the summary of the transaction and ledger information
```
WhatisRT: Again, not term

alonzo/formal-spec/txinfo.tex
```diff
@@ -3,5 +3,232 @@ \section{TxInfo Construction}
...
+  \caption{Transaction Summarization Functions}
+  \label{fig:txinfo-funcs}
+\end{figure}
+
+We give the Plutus types corresponding to the ledger counterparts in Figures \ref{fig:txinfo-types}
+and \ref{fig:txinfo-types-two}. Details for types that have non-identity translation
+functions between ledger and Plutus types are in Figure \ref{fig:txinfo-translations}.
+
+Certain types, such as the bootstrap address type, cannot be passed to scripts, and
+are therefore translated as $\Nothing$. All terms made up of untranslatable
```
WhatisRT: Change to something like: Anything that cannot be fully translated is considered untranslatable.

alonzo/formal-spec/txinfo.tex
```diff
@@ -3,5 +3,232 @@ \section{TxInfo Construction}
...
+terms also translate to $\Nothing$. For example, since a bootstrap address $a \in \AddrBS$
+is not translatable, neither is $a\in\Addr$, and $(a,\wcard,\wcard)~\in~\TxOut$
```
WhatisRT: Remove `~` around `\in` for better spacing

alonzo/formal-spec/txinfo.tex
```diff
@@ -3,5 +3,232 @@ \section{TxInfo Construction}
...
+also translates to $\Nothing$.
+
+We denote a term $t$ of a ledger type $\type{LT}$ that has been translated to the
```
WhatisRT: I think this is phrased a bit awkwardly, but term is used correctly here. Maybe: If $t$ is a term of type $LT$, then we denote the translated term of type $P.LT$ by $t_P$.

alonzo/formal-spec/txinfo.tex
```diff
@@ -3,5 +3,232 @@ \section{TxInfo Construction}
...
+Is the maximum possible value of a POSIX time interval. This will be adjusted in
+future eras to reflect the time being translated.
+
+Note that the $\Ptr$ addresses translated and passed to Plutus scripts are
+not resolvable to their corresponding key or script staking credencials. This
+is because doing a lookup in the resolution map, which part of the ledger state,
+could break determinism.
+
+During the transaction encoding and decoding process, a transaction is discarded if it is not
+encoded correctly. This includes, in particular, a check that scripts and $\Data$
```
WhatisRT: Either $\Script$ and $\Data$ elements, or scripts and datums, but don't mix them

alonzo/formal-spec/txinfo.tex
```diff
@@ -3,5 +3,232 @@ \section{TxInfo Construction}
...
+terms therein are constructed with constructors of terms of these types, themselves
```
WhatisRT: This is confusing. I'd just say 'are well-formed' or something like this.

alonzo/formal-spec/txinfo.tex
```diff
@@ -3,5 +3,232 @@ \section{TxInfo Construction}
...
+applied to correctly typed terms. Functions that perfom these checks are
+in Figure \ref{fig:data-script-check}. The $\fun{P.validateScript}$ function
+is a Plutus-library function which checks whether a bytestring represents a
+Plutus script.
+
+\begin{figure*}[htb]
```
WhatisRT: The alignment is this figure is weird. Both columns should be aligned to the left, for better readability. There's also too much white space between them.

alonzo/formal-spec/txinfo.tex
```diff
@@ -3,5 +3,232 @@ \section{TxInfo Construction}
...
+    ~\\
+    & \emph{Data-dropping Certificate Types} \\
+    & \DCertGen & \type{P.DCertGenesis} \\
+    & \DCertMir & \type{P.DCertMir} \\
+  \end{align*}
+  \caption{TxInfo and Constituent Types}
+  \label{fig:txinfo-types}
+\end{figure*}
+
+\begin{figure*}[htb]
```
WhatisRT: Same as the previous figure.

alonzo/formal-spec/txinfo.tex
```diff
@@ -3,5 +3,232 @@ \section{TxInfo Construction}
...
+    & \mathsf{Mint} & \type{P.Minting} \\
+    & \mathsf{Spend} & \type{P.Spending} \\
+    ~\\
+    & \emph{Execution Budget Types} \\
+    & \Integer & \type{P.ExCPU} \\
+    & \Integer  & \type{P.ExMemory}\\
+    & \ExUnits & \type{P.ExBudget}  \\
+    ~\\
+    & \emph{TxInfo} \\
+    & \TxInfo = \\
```
WhatisRT: This definition is very difficult to read. If the comments don't fit in the same line, they should be shortened until they do, with extra information given in the prose. For example, the comment for the mint line should just be `mint` or `mint field`.

alonzo/formal-spec/txinfo.tex
```diff
@@ -3,5 +3,232 @@ \section{TxInfo Construction}
...
+  \end{align*}
+  \caption{TxInfo and Constituent Types}
+  \label{fig:txinfo-types-two}
+\end{figure*}
+
+\begin{figure*}[htb]
+  \begin{align*}
+    & \fun{transAddr} \in \Addr  \to  \type{P.Address} \\
+    & \fun{transAddr} ~a = \begin{cases}
+      \Nothing & \text{if}~a~\in~\AddrBS \\
```
WhatisRT: Remove `~` around `\in` and `=`, there's too much white space

alonzo/formal-spec/txinfo.tex
```diff
@@ -3,5 +3,232 @@ \section{TxInfo Construction}
...
+      (ob,~st) & \text{if}~a~=~(\wcard, ob_P, st_P)
+    \end{cases} \\
+    & \text{Address translation}
+    \nextdef
+    & \fun{transValue} \in \Value \to \type{P.Value} \\
+    & \fun{transValue}~ (c,~ mp) = \{~ pid_P\mapsto (aid_P \mapsto q_P) ~\mid~ pid \mapsto (aid \mapsto q) ~\in ~mp ~\} \\
```
WhatisRT: Too much white space around `\in` here too

alonzo/formal-spec/txinfo.tex
```diff
@@ -3,5 +3,232 @@ \section{TxInfo Construction}
...
+    & ~~~~\cup \{~\type{P.adaSymbol}\mapsto (\type{P.adaToken} \mapsto c) ~\} \\
```
WhatisRT: So Plutus has a pid and asset name for Ada? We should mention that in the MA spec!

alonzo/formal-spec/txinfo.tex
```diff
@@ -3,5 +3,232 @@ \section{TxInfo Construction}
...
+    & \text{Value translation}
+    \nextdef
+    & \fun{transDCert} \in \DCertRegPool \to \type{P.DCertPoolRegister} \\
+    & \fun{transDCert}~\var{pool} = (\fun{poolId}~\var{pool},~\fun{poolVrf}~\var{pool}) \\
+    & \text{Pool registration certificate translation}
+    \nextdef
+    & \fun{txInfoIn} \in \UTxO \to \TxIn \to \type{P.TxInInfo} \\
+    & \fun{txInfoIn}~ utxo~ txin = \begin{cases}
+        (\var{txin}_P, \var{txout}_P) & \text{when $(\var{txin}\mapsto\var{txout})\in\var{utxo}$} \\
```
WhatisRT: We use `if` instead of `when` everywhere

alonzo/formal-spec/txinfo.tex
```diff
@@ -3,5 +3,232 @@ \section{TxInfo Construction}
...
+      \end{cases} \\
+      & ~~~~\where \\
+      & ~~~~~~~~\fun{slotToPOSIXTime}~pp~ei~sysS~vs~=~\type{ResolutionConstant} \\
+    & \text{Translate validity interval to a POSIX time range}
+  \end{align*}
+  \caption{TxInfo Constituent Type Translation Functions}
+  \label{fig:txinfo-translations}
+\end{figure*}
+
+\begin{figure*}[htb]
```
WhatisRT: There's too much white space around all the `=` signs in this figure (remove `~`)

WhatisRT (changes requested):
This does not actually define the `txInfo` function though. Also, since `txInfo` needs to invoke `transVITime`, it needs to have `EpochInfo` and `SystemStart` as parameters

alonzo/formal-spec/txinfo.tex
polinavino: maybe we should do that in the other (Value) PR?

alonzo/formal-spec/txinfo.tex
polinavino: the description of the translation is not trivial - it needs to be mentioned in the comment (i dont have a better idea of where to put it)

eras/alonzo/formal-spec/txinfo.tex
```diff
@@ -3,5 +3,284 @@ \section{TxInfo Construction}
...
+    $\fun{valContext}$ constructs the \emph{validation context}, also referred to as the
+    script context. A validation context is
+    a $\Data$ element which encodes both the summary of the transaction and ledger information
+    (this is supplied by the $\fun{txInfo}$ summarization function), and the script purpose.
+
+\begin{figure}
+  \emph{Plutus Library Types and Functions}
+  %
+  \begin{align*}
+      &\type{P.ScriptContext} ~=~ \type{P.TxInfo} \times \type{P.ScriptPurpose} \\
```
WhatisRT: This type is unused. If `valContext` was uncurried then it could use that type, but I don't think it should be. So I think this should just be removed.

eras/alonzo/formal-spec/txinfo.tex
```diff
@@ -3,5 +3,284 @@ \section{TxInfo Construction}
...
+      \nextdef
+      &\fun{P.toData} \in \type{P.T} \to \Data \\
+      &\text{Constructs a Data element from a Plutus-library-type element of type P.T}
+  \end{align*}
+    \emph{Ledger Functions}
+    %
+    \begin{align*}
+        &\fun{txInfo} \in \Language \to \PParams \to \EpochInfo \to \SystemStart \to \UTxO \to \Tx \to \TxInfo \\
+        &\fun{txInfo}~ pp~ ei~ sysS~ utxo~tx = \\
+        & ~~~~ (\{~\fun{txInfoIn}~\var{txin}~\mid~\var{txin}\in(\fun{txinputs}~tx)~\}, \\
```
WhatisRT: I think this needs an `_P`?

WhatisRT: Also, it's a bit awkward since the only reason this won't have `Nothing` in it is because of the ledger rules. I think it might be best to inline the definition of `txInfoIn` here. It's the only place where it's used anyway

eras/alonzo/formal-spec/txinfo.tex
```diff
@@ -3,5 +3,284 @@ \section{TxInfo Construction}
...
+
+We give the Plutus types corresponding to the ledger counterparts in Figures \ref{fig:txinfo-types}
+and \ref{fig:txinfo-types-two}. Details for types that have non-identity translation
+functions between ledger and Plutus types are in Figure \ref{fig:txinfo-translations}.
+
+\textbf{Notation. }
+In this Section we use the notation $\type{P.PlutusType}$ for the Plutus type
+$\type{PlutusType}$ (specified in the Plutus library) to distinguish it from ledger types.
+
+If $t$ is a term of type $LT$, then we denote the translated term of type $P.LT$ by $t_P$.
```
WhatisRT: I realized that `_P` isn't a function on terms as I originally thought. It's just a regular function. So it should probably be treated as such, and be formally stated in one of the figures. We could for example call it `toPlutusType_LT : LT -> P.LT` which we require to be defined for each ledger type `LT`, and then just use the `_P` notation that hides the `LT` index.

WhatisRT (changes requested):
I think this section should go into an appendix. It's not really required to understand the ledger and its properties. In fact, none of the properties we've proven depends on how `txinfo` is defined.

I've also made some small formatting changes & fixes, I hope you like them!

WhatisRT (approved):
I've done a commit on top of yours that streamlines everything a bit and restores the `_P` notation (I liked it, it just needed to be formally justified). If you like it the way it is now, we can merge it!

WhatisRT (changes requested):
Actually, I realized that the threading through of the `EpochInfo` and `SystemStart` isn't finished. `collectTwoPhaseScriptInputs` needs to pass these to `txinfo`, which it needs to get from `UTXOS`, which means that we need to adjust the context of the entire family of `UTXO*` rules...

I don't even know where these parameters come from in the first place, I just hope they don't have to be threaded through all the way to `CHAIN`...

eras/alonzo/formal-spec/utxo.tex
```diff
@@ -480,6 +480,9 @@ \subsection{The UTXOS transition system}
 
 There are two rules, corresponding to the two possible state changes of the
 UTxO state in the UTXOS transition system (Figure~\ref{fig:rules:utxo-state-upd}).
+Note here that the constants $\mathsf{ei}$ and $\mathsf{sysSt}$ are available
+inside the definition of all ledger rules. They are supplied by the consensus
```
WhatisRT: No, this is not true. The ledger rules are simply relations, so to make this true we'd have to break our entire theory. Those need to be provided by the contexts, even if the code does it differently.

WhatisRT (changes requested):

eras/alonzo/formal-spec/utxo.tex
polinavino: we act as if it's hard-coded, like other constants. But this is obscuring where it comes from. It is not part of the context

polinavino: i can re-word this and point out it's an implementation detail (ie it's actually hard-coded)

WhatisRT: Hmm, if they're constants then maybe we can just stick them into the global constants?

WhatisRT: Also, I'm confused. Why is `slotToPOSIXTime` a constant?

polinavino: The use of EpochInfo is similar to the use of NetworkID

WhatisRT: And the NetworkID is a global constant, see Fig 8 of the Shelley spec

WhatisRT (approved):
LGTM!

JaredCorduan:
The commits need to be signed before we can merge. Perhaps we can also clean up the commit history? Thanks!

PR #2496, opened 2021-09-22
fix typo in faq.rst

vpatov:


vpatov:
This typo is being fixed in an existing PR.

PR #2497, opened 2021-09-22
Few minor improvements and a fix for 'Reward Provenance' property test:

lehins:
This PR fixes occasional test failure that is caused by the `genBlocksMade` producing an empty map sometimes, which in turn causes no reward provenance produced either. Also:

* Turns printer into generator
* Get rid of non-reproducable `generate` action
* Consolidate `reward` and `rewardPulser` functions into one reducing clutter.


JaredCorduan (approved):
looks great, thank you for cleaning this up!

PR #2499, opened 2021-09-24
use github action to run ormolu

JaredCorduan:
And also remove the no longer used `check-cabal-project` pipeline from buildkite. This will let us get rid of buildkite altogether.


.buildkite/pipeline.yml
```diff
@@ -1,13 +1,2 @@
 # The Buildkite pipeline definition.
 # See https://buildkite.com/docs/pipelines for documentation.
```
JaredCorduan: I do not know what else I need to do to disable Buildkite (and if it is okay to delete this yml file yet).

lehins (approved):
:+1:

lehins:
@JaredCorduan I was thinking about this and maybe we should define it as a separate `job` in `haskell.yaml` instead of a whole separate workflow. This has a benefit of the same github settings as our normal CI. So placing it in `haskell.yaml` like so:

```yaml
jobs:
  build:
  .... all the stuff from Haskell CI
  ormolu:
    runs-on: ubuntu-latest

    defaults:
      run:
        shell: bash

    strategy:
      fail-fast: false
    steps:
    - uses: actions/checkout@v1

    - name: Install ormolu
      run: |
        mkdir -p "$HOME/.local/bin"
        curl -sL https://github.com/tweag/ormolu/releases/download/0.1.4.1/ormolu-Linux -o "$HOME/.local/bin/ormolu"
        chmod a+x "$HOME/.local/bin/ormolu"
        echo "$HOME/.local/bin" >> $GITHUB_PATH
    - name: Run ormolu
      run: ./scripts/ormolise.sh
```

JaredCorduan:
That seems pretty reasonable @lehins . I guess the question is, do we want to make that haskell CI action required? I think we do want the ormolu check required (though I'm having trouble figuring out how atm).

lehins:
> Not sure I understand. Isn't it already required?

if you look at the checks below, you'll see that the haskell action doesn't have `required` to the left of details. and indeed you can merge the PR if this check fails.

JaredCorduan:
oops, I did not mean to edit your comment @lehins , I didn't even know that was possible :scream:

lehins:
> I did not mean to edit your comment 

lol. It's one way to communicate :) I see what you mean now. Looks like it should be possible to protect it in settings, see [SO answer](https://stackoverflow.com/questions/60776412/github-actions-is-there-a-way-to-make-it-mandatory-for-pull-request-to-merge) And yeah, we want to protect on all of the CI checks

JaredCorduan:
> lol. It's one way to communicate :) I see what you mean now. Looks like it should be possible to protect it in settings, see [SO answer](https://stackoverflow.com/questions/60776412/github-actions-is-there-a-way-to-make-it-mandatory-for-pull-request-to-merge) And yeah, we want to protect on all of the CI checks

yea, I saw that, but the ormolu check isn't showing up there, only hydra.

PR #2503, opened 2021-09-27
Use GADT extension in TxBody.

nc6:
For some reason this works in some places, but others complain about the
missing extension.


nc6:
I'm a very confused person, sorry

Issue #2462, opened 2021-09-15
Remove dependency on cardando-protocol-tpraos


JaredCorduan:
We can remove the dependency on `cardano-protocol-tpraos` once #2459 #2460 #2461 are complete and when:

- [x] the ledger has its own types for `PoolDistr`, ` IndividualPoolStake` and `ProtVer`
- [x] the `Pretty` instances for protocol types are moved
- [x] we have deleted the deprecated modules which re-export the protocol modules
- [x] move the `Cardano.Ledger.Shelley.API.Protocol` elsewhere
- [x] move the `getLeaderSchedule` function from `Cardano.Ledger.Shelley.API.Wallet` elsewhere
- [x] move the protocol types out of `Cardano.Ledger.Shelley.API.Types`
- [x] figure out somewhere else to put the Shelley and Alonzo `EraMapping` modules

PR #2505, opened 2021-09-30
replace block header with block header view where appropriate

### LINKED ISSUE #2460: BBODY to use BHView
Create a type class, a block header view `BHView`, which contains all the information that the `BBODY` rule needs from the block header. Change the signal of the `BBODY` rule from a block to a pair consisting of the block body and the block header view.

Finally, create an instance of the `BHView` class for the `TPraos` protocol block header.

JaredCorduan:
A new type `BHeaderView` has been added to `cardano-ledger-core`, it contains the information from the block header that the `BBODY` rule needs. The function `isOverlaySlot` has also been moved next to the `BHeaderView`  definition (it would be nice to move it to `cardano-protocol-tpraos`, but it is still needed in the `BBODY` rule).

The BBODY rule no longer takes a `Block` as a signal, but a pair of a `BHeaderView` and a `TxSeq`. Additionally, the `applyBlock` API function now uses the `BHeaderView` as well. The `cardano-protocol-tpraos` package provides a translation from transitional-praos block headers to the view.

closes #2460

### DIFF
--- a/libs/cardano-ledger-core/src/Cardano/Ledger/BHeaderView.hs
+++ b/libs/cardano-ledger-core/src/Cardano/Ledger/BHeaderView.hs
@@ -1,0 +1,49 @@
+{-# LANGUAGE DataKinds #-}
+{-# LANGUAGE MultiParamTypeClasses #-}
+
+module Cardano.Ledger.BHeaderView where
+
+import Cardano.Ledger.BaseTypes (BoundedRational (..), UnitInterval)
+import Cardano.Ledger.Hashes (EraIndependentBlockBody)
+import Cardano.Ledger.Keys (Hash, KeyHash, KeyRole (..))
+import Cardano.Ledger.Slot (SlotNo (..), (-*))
+import Numeric.Natural (Natural)
+
+-- | 'BHeaderView' provides an interface between block headers
+-- from different Cardano protocols and packages that should be
+-- agnostic of Cardano protocol specific details,
+-- such as those in TPraos, Praos, Genesis, etc.
+--
+-- In particular, the 'BBODY' rule comprises most of the ledger logic
+-- and should work independently of the protocol. The values in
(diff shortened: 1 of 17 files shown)


libs/cardano-ledger-core/src/Cardano/Ledger/BHeaderView.hs
```diff
@@ -0,0 +1,30 @@
...
+
+module Cardano.Ledger.BHeaderView where
+
+import Cardano.Ledger.BaseTypes (BoundedRational (..), UnitInterval)
+import Cardano.Ledger.Hashes (EraIndependentBlockBody)
+import Cardano.Ledger.Keys (Hash, KeyHash, KeyRole (..))
+import Cardano.Ledger.Slot (SlotNo (..), (-*))
+import Numeric.Natural (Natural)
+
+data BHeaderView crypto = BHeaderView
```
nc6, line 21: I think it would be helpful to add some explanatory documentation here!

nc6 (approved):
Looks good! Can we now drop the dependency on `cardano-protocol-tpraos` in `cardano-ledger-shelley` and beyond?

nc6:
Seems like the answer to that question is "not yet".

JaredCorduan:
> Seems like the answer to that question is "not yet".

yea, but we are getting closer! I'll have a follow up PR that will parameterize the `Block` type by the header type. After that we will be very close, maybe just `PoolDistr` and a couple other stragglers.

libs/cardano-ledger-core/src/Cardano/Ledger/BHeaderView.hs
JaredCorduan: yes indeed, thanks!

JaredCorduan:
@nc6 I updated #2462 with all the work left to remove `cardano-protocol-tpraos` from all the era `impl` packages.

PR #2509, opened 2021-10-04
update cardano-base

JaredCorduan:
This bring in:
 * https://github.com/input-output-hk/cardano-base/pull/240

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -40,8 +40,8 @@ test-show-details: streaming
 source-repository-package
   type: git
   location: https://github.com/input-output-hk/cardano-base
-  tag: 42525d64502f41f4753d9efc98b9a936ac5ba2ac
-  --sha256: 1mn7ivszn2652vqm8hbkw9fwawf70qh2mzn0lbqig6basc5xdr75
+  tag: 654f5b7c76f7cc57900b4ddc664a82fc3b925fb0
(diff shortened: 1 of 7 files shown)


lehins (approved):

nc6:
The golden files are changed here. Do we expect this?

JaredCorduan:
> The golden files are changed here. Do we expect this?

definitely not, they should not have changed. I do not understand why removing those rewrite rules would change the representation...

lehins:
I know exactly why these test fail!

How many bytes is this as hex: `"01020304"` ... 4 bytes, but our hashes are expected to be 8 bytes. So tests where always buggy

lehins:
Instead of `UnsafeHash` we should use `hashFromBytes`, this way at least we'll get a proper failure when hashing algorithm changes in tests. In other words `UnsafeHash` should never be used and IMHO it shouldn't even be exported.

JaredCorduan:
> Instead of `UnsafeHash` we should use `hashFromBytes`, this way at least we'll get a proper failure when hashing algorithm changes in tests. In other words `UnsafeHash` should never be used and IMHO it shouldn't even be exported.

So a previous PR changed the hashing algorithm, and the use of `UnsafeHash` prevented the golden tests from noticing this?

JaredCorduan:
thanks for cleaning this up @lehins , looks good to me!

PR #2510, opened 2021-10-05
parameterize Block by the header type

### LINKED ISSUE #2459: Parameterize Block type by the header type
Move the `Block` type to the `cardano-protocol-tpraos` package.

JaredCorduan:
This changes the `Block` type to be parameterized by the block header type. I did not attempt to make the tests more generic, but just used the TPraos `BHeader` directly.

closes #2459 

EDIT -  I've also now:

* Moved `TxIn` and `TxId` to cardano-ledger-core
* Moved `BlocksMade` to cardano-ledger-core (I placed it in `BaseTypes`)
* Moved `Block` to cardano-ledger-core
* Moved `LaxBlock` to the shelley tests
* The signal for `BBODY` is once again a `Block`, but now we use the `BHeaderView` for the block header in the `BBODY` rule. There is one awkward snag with this (but I think it is worth it), the `Block` pattern requires a `ToCBOR` instance on the block header. I didn't want to add a `ToCBOR` for the BHeaderView, since we don't actually ever want it to be serialized, so when giving `BBODY` a `Block`, we need to use the constructor `Block'` with something bogus in the bytes field.


eras/shelley/impl/src/Cardano/Ledger/Shelley/BlockChain.hs
```diff
@@ -250,27 +250,33 @@ bbHash (TxSeq' _ bodies wits md) =
     hashStrict = Hash.hashWith id
     hashPart = Hash.hashToBytes . hashStrict . BSL.toStrict
 
-data Block era
-  = Block' !(TP.BHeader (Crypto era)) !(Era.TxSeq era) BSL.ByteString
+data Block h era
```
nc6: Now that `Block` is parametric over the header type, I wonder whether we should move it to `cardano-ledger-core`

nc6 (changes requested):
This looks great, but now that `Block` is parametrised by the header, can we revert to using `Block` as the signal for `BBODY` rather than the tuple?

eras/shelley/impl/src/Cardano/Ledger/Shelley/BlockChain.hs
JaredCorduan: These are some great ideas, thanks @nc6 ! I'll do some experimenting. :scientist:

JaredCorduan: @nc6 this turned out mostly great, see my notes in the PR notes at the top.

libs/cardano-ledger-core/src/Cardano/Ledger/Block.hs
```diff
@@ -0,0 +1,154 @@
...
+import Data.Foldable (toList)
+import Data.Set (Set)
+import qualified Data.Set as Set
+import Data.Typeable
+import GHC.Generics (Generic)
+import GHC.Records (HasField (..))
+import NoThunks.Class (NoThunks (..))
+
+data Block h era
+  = Block' !(h (Crypto era)) !(Era.TxSeq era) BSL.ByteString
```
nc6, line 53: Rather than exposing the constructor directly, how about having a pair of patterns which we use in cases where we don't care about serialisation: a unidirectional pattern `UnserialisedBlock` that allows us to pull out the header and the TxSeq, and a bidirectional `UnsafeUnserialisedBlock` which constructs the block with an error in the bytestring? This way it's a bit clearer what's going on, and we don't have to expose the potentially unsafe constructor?

nc6 (changes requested):
Looks great, I think the changes make this much nicer! I have left one final suggestion. Also, is this file intentionally checked in?
- eras/shelley/impl/dist/cabal-config-flags

libs/cardano-ledger-core/src/Cardano/Ledger/Block.hs
JaredCorduan: Absolutely, that's much better!

JaredCorduan:
> Looks great, I think the changes make this much nicer! I have left one final suggestion. Also, is this file intentionally checked in?
> 
> * eras/shelley/impl/dist/cabal-config-flags

oops! I'll remove it, thanks! Edit: done.

nc6 (approved):

PR #2511, opened 2021-10-06
Ledger API function for stake distribution and pool parameters query

HeinrichApfelmus:
### Issue Number

[CAD-3480](https://input-output.atlassian.net/browse/CAD-3480)

### Summary

module `Cardano.Ledger.Shelley.API.Wallet`:
* Introduce and export types `RewardParams` and `RewardInfoPool`
* The `getRewardInfoPools` function is essentially a refactoring of the `getNonMyopicMemberRewards` function.
* Also reorder the functions in the file and export list more thematically.

### Comments

Exporting `getRewardInfoPools` helps the wallet to calculate reward predictions (e.g. stake pool desirability, non-myopic rewards, but also future reward measures) for stake pools. It's intended to become a local state query in the node.

I have made a best effort to conform with the coding conventions in the ledger, but I'm probably off. I did not know where to put the types. Also, I did not know whether or how to write unit tests (though it seems that the old `GetRewardProvenance` does not have many tests either.)

### DIFF
--- a/eras/shelley/impl/src/Cardano/Ledger/Shelley/API/Wallet.hs
+++ b/eras/shelley/impl/src/Cardano/Ledger/Shelley/API/Wallet.hs
@@ -11,18 +11,24 @@
 
 module Cardano.Ledger.Shelley.API.Wallet
   ( getNonMyopicMemberRewards,
+    -- * UTxOs
     getUTxO,
     getUTxOSubset,
     getFilteredUTxO,
+    -- * Stake Pools
     getLeaderSchedule,
     getPools,
     getPoolParameters,
     getTotalStake,
     poolsByTotalStakeFraction,
+    RewardInfoPool (..),
+    RewardParams (..),
+    getRewardInfoPools,
     getRewardInfo,
+    -- * Transaction helpers
     CLI (..),
     addShelleyKeyWitnesses,
-    -- | Ada Pots
+    -- * Ada Pots
     AdaPots (..),
     totalAdaES,
     totalAdaPotsES,
@@ -68,6 +74,7 @@ import Cardano.Ledger.Shelley.PParams (PParams, PParams' (..), ProtVer)
 import Cardano.Ledger.Shelley.RewardProvenance (RewardProvenance)
 import Cardano.Ledger.Shelley.Rewards
   ( NonMyopic (..),
+    PerformanceEstimate (..),
     StakeShare (..),
     getTopRankedPools,
     nonMyopicMemberRew,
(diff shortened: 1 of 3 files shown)


JaredCorduan (approved):
Looks good to me! Thank you for leaving this module in better shape than you found it, much appreciated!

eras/shelley/impl/src/Cardano/Ledger/Shelley/API/Wallet.hs
```diff
@@ -103,9 +135,73 @@ import Data.Ratio ((%))
...
+  UTxO $
+    fullUTxO `Map.restrictKeys` txins
+  where
+    UTxO fullUTxO = getUTxO ss
+
+--------------------------------------------------------------------------------
+-- Stake pools and pool rewards
+--------------------------------------------------------------------------------
+
+-- | Get the /current/ registered stake pool parameters for a given set of
```
nc6: This haddock is incorrect!

eras/shelley/impl/src/Cardano/Ledger/Shelley/API/Wallet.hs
```diff
@@ -103,9 +135,73 @@ import Data.Ratio ((%))
...
+-- stake pools. The result map will contain entries for all the given stake
+-- pools that are currently registered.
```
nc6:
```suggestion
-- | Get the /current/ registered stake pools.
```

eras/shelley/impl/src/Cardano/Ledger/Shelley/API/Wallet.hs
```diff
@@ -238,92 +334,121 @@ currentSnapshot ss =
...
+            )
+            mempty
+            (_poolOwners poolp)
 
+-- | Calculate stake pool rewards from the snapshot labeled `go`.
+-- Also includes information on how the rewards were calculated
+-- ('RewardProvenance').
+--
+-- For a calculation of rewards based on the current stake distribution,
+-- see 'getRewardInfo'.
```
nc6: This function _is_ `getRewardInfo`...

nc6: Do you mean `getRewardInfoPools`? If so, I would say that the names are rather confusing! I'm guessing the intent is to deprecate this function in favour of the new one? If so, maybe we should name it accordingly.

nc6 (changes requested):
Code looks nice, but a few places where the comments need to be updated/clarified!

eras/shelley/impl/src/Cardano/Ledger/Shelley/API/Wallet.hs
HeinrichApfelmus: Oops, sorry.

eras/shelley/impl/src/Cardano/Ledger/Shelley/API/Wallet.hs
HeinrichApfelmus: > Do you mean getRewardInfoPools? 

Yes, sorry.

```suggestion
-- see 'getRewardInfoPools'.
```

> If so, I would say that the names are rather confusing! I'm guessing the intent is to deprecate this function in favour of the new one? If so, maybe we should name it accordingly.

Well, the confusion that `getRewardInfo` returns a `RewardProvenance` is something that I inherited from the previous code 😅. But we can rename `getRewardInfo` to `getRewardProvenance` to disambiguate better. How would I (we) go about deprecation?

nc6 (approved):

jackrousian:
[CAD-3480]

PR #2514, opened 2021-10-08
babbage phase 1-2 rearrangement

polinavino:
Making a draft PR for the babbage changes i linked in slack some time ago so that they dont get lost.


polinavino:
Not doing either change

JaredCorduan:
thanks for spec-ing this out @polinavino , if/when we return to this we'll have a place to start! :)

PR #2515, opened 2021-10-08
make ExUnit type parameteric, but hide details

JaredCorduan:
This should make it easier for consensus to add a `BoundedMeasure` instance to `ExUnits`.

### DIFF
--- a/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Scripts.hs
+++ b/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Scripts.hs
@@ -1,12 +1,15 @@
 {-# LANGUAGE DataKinds #-}
+{-# LANGUAGE DeriveFunctor #-}
 {-# LANGUAGE DeriveGeneric #-}
 {-# LANGUAGE DerivingVia #-}
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE FlexibleInstances #-}
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
 {-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE MultiParamTypeClasses #-}
 {-# LANGUAGE NamedFieldPuns #-}
 {-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE PatternSynonyms #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE StandaloneDeriving #-}
 {-# LANGUAGE TypeApplications #-}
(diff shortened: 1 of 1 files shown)


eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Scripts.hs
```diff
@@ -148,23 +151,39 @@ isPlutusScript (TimelockScript _) = False
...
   deriving
     (Monoid, Semigroup)
-    via (InstantiatedAt Measure ExUnits)
+    via (InstantiatedAt Measure (ExUnits' a))
+
+instance NoThunks a => NoThunks (ExUnits' a)
+
+instance NFData a => NFData (ExUnits' a)
+
+newtype ExUnits = WrappExUnits {unWrap :: ExUnits' Natural}
```
nc6:
```suggestion
newtype ExUnits = WrapExUnits {unWrap :: ExUnits' Natural}
```

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Scripts.hs
```diff
@@ -148,23 +151,39 @@ isPlutusScript (TimelockScript _) = False
...
+  deriving (Eq, Generic, Show)
+  deriving newtype (Monoid, Semigroup)
 
 instance NoThunks ExUnits
 
 instance NFData ExUnits
 
+pattern ExUnits :: Natural -> Natural -> ExUnits
```
nc6, line 196: I think we might want some comments here explaining why this is needed, otherwise it's a bit weird!

nc6 (commented):
How does this help consensus? Presumably they still have no `BoundedMeasure` for `ExUnits`. Do they convert to another format which has a bound?

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Scripts.hs
JaredCorduan: great point!

JaredCorduan:
> How does this help consensus? Presumably they still have no `BoundedMeasure` for `ExUnits`. Do they convert to another format which has a bound?

That right, they will use `ExUnits (Infinite Natural)`, which just slaps `ω` on top of `Natural`.

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Scripts.hs
JaredCorduan: I've added some prose. Let me know if you think it makes sense.

nc6 (approved):

nfrisby:
@JaredCorduan @nc6 Apologies for the delayed response!

I think Consensus will also need `unWrap` and `ExUnits'` to be exported (I think that's the minimal set we'd need). Relatedly, I would suggest `unWrapExUnits` instead of just `unWrap`. Thanks!

PR #2518, opened 2021-10-13
More TPraos separation

JaredCorduan:
This PR completes a few of the tasks remaining for separating TPraos from the ledger code (see #2462)

* `PoolDistr`, `IndividualPoolStake` and `ProtVer` have been moved to the `cardano-ledger-core` package.
* The Wallet API function `getLeaderSchedule` was moved to the Protocol API module, so that it does not need any TPraos imports. I'm not sure what to do with the Protocol API module, it may need to move to consensus.
* All the pretty printers have been moved to a new package `cardano-ledger-pretty'. There are a few things worth mentioning:
  - I wanted to keep the `Cardano.Ledger.Pretty` namespace, so I was not able to deprecate the old module. I think this is okay, I don't think any downstream libraries are using the pretty printers.
  - I added some unidirectional patterns since the instances are no longer couple to the hidden module definitions. I also had to add some `COMPLETE` pragma to some existing patterns.
* ~I removed the TPraos rules (`OCERT`, `OVERLAY` and `TICKN`) from the era mappings.~


eras/shelley/impl/src/Cardano/Ledger/Shelley/RewardUpdate.hs
```diff
@@ -21,14 +21,7 @@ import Cardano.Binary
...
   )
-import Cardano.Ledger.BaseTypes
-  ( ActiveSlotCoeff,
-    NonNegativeInterval,
-    ShelleyBase,
-    UnitInterval,
-    boundedRationalFromCBOR,
-    boundedRationalToCBOR,
-  )
+import Cardano.Ledger.BaseTypes (ActiveSlotCoeff, NonNegativeInterval, ProtVer (..), ShelleyBase, UnitInterval, boundedRationalFromCBOR, boundedRationalToCBOR)
```
nc6: [minor] long line

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rewards.hs
```diff
@@ -48,14 +48,7 @@ import Cardano.Binary
...
   )
-import Cardano.Ledger.BaseTypes
-  ( ActiveSlotCoeff,
-    BoundedRational (..),
-    NonNegativeInterval,
-    UnitInterval,
-    activeSlotVal,
-    invalidKey,
-  )
+import Cardano.Ledger.BaseTypes (ActiveSlotCoeff, BoundedRational (..), NonNegativeInterval, ProtVer, UnitInterval, activeSlotVal, invalidKey)
```
nc6: [minor] long line

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Rupd.hs
```diff
@@ -22,18 +22,7 @@ module Cardano.Ledger.Shelley.Rules.Rupd
...
-    ShelleyBase,
-    StrictMaybe (..),
-    UnitInterval,
-    activeSlotCoeff,
-    epochInfo,
-    maxLovelaceSupply,
-    randomnessStabilisationWindow,
-    securityParameter,
-  )
+import Cardano.Ledger.BaseTypes (BlocksMade, NonNegativeInterval, ProtVer, ShelleyBase, StrictMaybe (..), UnitInterval, activeSlotCoeff, epochInfo, maxLovelaceSupply, randomnessStabilisationWindow, securityParameter)
```
nc6: [minor] long line

libs/cardano-ledger-core/src/Cardano/Ledger/PoolDistr.hs
```diff
@@ -0,0 +1,83 @@
...
+{-# LANGUAGE DerivingVia #-}
+{-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE FlexibleInstances #-}
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE MultiParamTypeClasses #-}
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE UndecidableInstances #-}
+{-# OPTIONS_GHC -Wno-redundant-constraints #-}
+
+module Cardano.Ledger.PoolDistr where
```
nc6, line 20: Can we document this module a bit? Just to provide a general idea of what a `PoolDistr` is and why it's a fundamental concept.

nc6 (approved):
Minor comments, otherwise looks good!

eras/shelley/impl/src/Cardano/Ledger/Shelley/RewardUpdate.hs
JaredCorduan: good catch! I would not do that on purpose, I must have done something to trigger omolu.

libs/cardano-ledger-core/src/Cardano/Ledger/PoolDistr.hs
JaredCorduan: absolutely! Thank you for reminding me recently about things that could be documented better, I need to do this more by reflex!

PR #2519, opened 2021-10-13
expose more of the ExUnits internals

JaredCorduan:


eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Scripts.hs
```diff
@@ -30,6 +30,8 @@ module Cardano.Ledger.Alonzo.Scripts
     -- * Cost Model
     CostModel (..),
     ExUnits (ExUnits, exUnitsMem, exUnitsSteps, ..),
+    ExUnits',
+    WrapExUnits (unWrapExUnits),
```
nfrisby: This line should be `ExUnits (unWrapExUnits),`

JaredCorduan: ha ha, `unWrap` was actually already being exported by `ExUnits (ExUnits, exUnitsMem, exUnitsSteps, ..)`. :)

nfrisby (approved):

PR #2520, opened 2021-10-13
Added the Compact data modules

TimSheard:
Some new data structures to be used instead of Data.Map
Take advantage of some properties
  1) Keys are Cryptographic Hashes, so they are evenly distributed, and  never have collisions
  2) Aggressively using {-# UNPACK #-}   to reduce space
   3) Use Trie based structures based on SmallArray


lehins (approved):
This is still work in progress, but is already functional and shows really good results with respect to reducing memory overhead when comparing to `Data.Map.Strict` from `containers`. We still need to add tests and reduce clutter by removing unused code. We will continue improving this `compact-map` package. However, considering that it is not being used by any other packages it is already safe to merge it in, despite its WIP state.

PR #2525, opened 2021-10-19
Fix the alonzo tx witness serialization

JaredCorduan:
Missing witness group should be entirely absent on the wire, not encoded as an empty list. We were encoding an empty list in the scenario that one type of Plutus script was present, but another Plutus type was absent, since the `Omit` check was checking to see if there were no scripts at all.

I am hoping this unblocks https://github.com/input-output-hk/ouroboros-network/pull/3438


redxaxder (approved):

nc6 (approved):

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxWitness.hs
```diff
@@ -362,7 +374,12 @@ encodeWitnessRaw vkeys boots scripts dats rdmrs =
     unwrapPS1 _ = Nothing
     unwrapPS2 (PlutusScript PlutusV2 x) = Just x
     unwrapPS2 _ = Nothing
-    (plutusScripts, timelocks) = Map.partition isPlutusScript scripts
+
+    isTimelock (TimelockScript _) = True
+    isTimelock (PlutusScript _ _) = False
+
+    isPlutus _ (TimelockScript _) = True
```
JaredCorduan, line 381: oh goodness :facepalm: I will need to do a follow-up PR. cc @nc6

erikd: I can wait.

PR #2526, opened 2021-10-20
correct a mistake in 9ab85598f0cf1c780665b8b3bc52b

JaredCorduan:
Timelock scripts were not filtered out of the Plutus witnesses on the wire. I will add more golden tests (round trip tests would not catch this mistake) soon.


erikd (approved):

erikd:
Obviously wrong before and obviously correct now, so just merging it.

PR #2527, opened 2021-10-21
at PV 7 make MissingScriptWitnessesUTXOW clearer

JaredCorduan:
The `MissingScriptWitnessesUTXOW` failure is confusing, it reports an empty set in the case of extraneous scripts. Starting at major protocol version 7, we use a new failure, `ExtraneousScriptWitnessesUTXOW` for the other side of the symmetric difference.

### DIFF
--- a/eras/shelley/impl/src/Cardano/Ledger/Shelley/HardForks.hs
+++ b/eras/shelley/impl/src/Cardano/Ledger/Shelley/HardForks.hs
@@ -59,7 +59,7 @@ translateTimeForPlutusScripts ::
 translateTimeForPlutusScripts pp = pvMajor (getField @"_protocolVersion" pp) > 5
 
 -- | Starting with protocol version 7, the UTXO rule predicate failure
--- MissingScriptWitnessesUTXOW will not be used for unnecessary scripts
+-- MissingScriptWitnessesUTXOW will not be used for extraneous scripts
 missingScriptsSymmetricDifference ::
   (HasField "_protocolVersion" pp ProtVer) =>
   pp ->
--- a/eras/shelley/impl/src/Cardano/Ledger/Shelley/HardForks.hs
+++ b/eras/shelley/impl/src/Cardano/Ledger/Shelley/HardForks.hs
@@ -8,6 +8,7 @@ module Cardano.Ledger.Shelley.HardForks
     validatePoolRewardAccountNetID,
     allowScriptStakeCredsToEarnRewards,
     translateTimeForPlutusScripts,
+    missingScriptsSymmetricDifference,
   )
 where
 
@@ -56,3 +57,11 @@ translateTimeForPlutusScripts ::
   pp ->
   Bool
 translateTimeForPlutusScripts pp = pvMajor (getField @"_protocolVersion" pp) > 5
+
+-- | Starting with protocol version 7, the UTXO rule predicate failure
+-- MissingScriptWitnessesUTXOW will not be used for unnecessary scripts
+missingScriptsSymmetricDifference ::
+  (HasField "_protocolVersion" pp ProtVer) =>
+  pp ->
+  Bool
+missingScriptsSymmetricDifference pp = pvMajor (getField @"_protocolVersion" pp) > 6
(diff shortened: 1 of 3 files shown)


eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Utxow.hs
```diff
@@ -137,6 +138,8 @@ data UtxowPredicateFailure era
       !(AuxiliaryDataHash (Crypto era)) -- hash of the full metadata
       -- Contains out of range values (strings too long)
   | InvalidMetadata
+  | UnnecessaryScriptWitnessesUTXOW
```
lehins: In order to relay the intention better I'd name it `ExtraneousScriptWitnessesUTXOW`. This is a suggestion, feel free to resolve for any reason.

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Utxow.hs
```diff
@@ -320,8 +329,12 @@ shelleyStyleWitness collectVKeyWitnesses embed = do
   {-  { s | (_,s) ∈ scriptsNeeded utxo tx} = dom(txscripts txw)          -}
   let sNeeded = scriptsNeeded utxo tx
       sReceived = Map.keysSet (getField @"scriptWits" tx)
-  sNeeded == sReceived
+  sNeeded `Set.isSubsetOf` sReceived
     ?! embed (MissingScriptWitnessesUTXOW (sNeeded `Set.difference` sReceived))
+  sReceived `Set.isSubsetOf` sNeeded
+    ?! if HardForks.missingScriptsSymmetricDifference pp
+      then embed (UnnecessaryScriptWitnessesUTXOW (sReceived `Set.difference` sNeeded))
+      else embed (MissingScriptWitnessesUTXOW mempty)
```
lehins: This is how I'd write it:
```suggestion
  when (HardForks.missingScriptsSymmetricDifference pp) $ do
    let extraneous = sReceived `Set.difference` sNeeded
    Set.null extraneous ?! embed (UnnecessaryScriptWitnessesUTXOW extraneous)
  sNeeded == sReceived
    ?! embed (MissingScriptWitnessesUTXOW (sNeeded `Set.difference` sReceived))
```

First of all for protocol that is before the hardfork we'll do exactly the same thing and we avoid the redundant subset checking. While for the protocol after the hardfork we'll shortcircuit on the extraneous script

lehins (approved):
Minor suggestions, but as far as correctness is concerned it is good to go as is.

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Utxow.hs
JaredCorduan: yea, extraneous is better, thanks!

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Utxow.hs
JaredCorduan: great!

JaredCorduan: I think this even fixes the unit test that just broke :)

JaredCorduan: @lehins I forgot that the failures accumulate, so the `when` has the effect that an extraneous script would hit two failures, `ExtraneousScriptWitnessesUTXOW` and `MissingScriptWitnessesUTXOW`, even when no needed script was missing. I've pushed a change, let know what you think

lehins: Yeah, makes sense. Looks good.

eras/shelley/impl/src/Cardano/Ledger/Shelley/HardForks.hs
```diff
@@ -56,3 +57,11 @@ translateTimeForPlutusScripts ::
   pp ->
   Bool
 translateTimeForPlutusScripts pp = pvMajor (getField @"_protocolVersion" pp) > 5
+
+-- | Starting with protocol version 7, the UTXO rule predicate failure
+-- MissingScriptWitnessesUTXOW will not be used for unnecessary scripts
```
lehins:
```suggestion
-- MissingScriptWitnessesUTXOW will not be used for extraneous scripts
```

PR #2530, opened 2021-10-25
Remove compacting of TxIn. Add Keyed instances for hashes

lehins:
This PR requires: https://github.com/input-output-hk/cardano-base/pull/243

### DIFF
--- a/libs/cardano-ledger-core/src/Cardano/Ledger/Hashes.hs
+++ b/libs/cardano-ledger-core/src/Cardano/Ledger/Hashes.hs
@@ -34,6 +34,7 @@ import Cardano.Ledger.Crypto (ADDRHASH)
 import qualified Cardano.Ledger.Crypto as CC (Crypto)
 import Control.DeepSeq (NFData)
 import Data.Aeson
+import Data.Compact.HashMap (Keyed)
 import GHC.Generics (Generic)
 import NoThunks.Class (NoThunks (..))
 
@@ -75,13 +76,11 @@ newtype ScriptHash crypto
   deriving (Show, Eq, Ord, Generic)
   deriving newtype (NFData, NoThunks)
 
-deriving newtype instance
-  CC.Crypto crypto =>
-  ToCBOR (ScriptHash crypto)
+deriving newtype instance CC.Crypto crypto => Keyed (ScriptHash crypto)
 
-deriving newtype instance
-  CC.Crypto crypto =>
-  FromCBOR (ScriptHash crypto)
+deriving newtype instance CC.Crypto crypto => ToCBOR (ScriptHash crypto)
+
+deriving newtype instance CC.Crypto crypto => FromCBOR (ScriptHash crypto)
 
 deriving newtype instance CC.Crypto crypto => ToJSON (ScriptHash crypto)
 
--- a/libs/cardano-ledger-core/src/Cardano/Ledger/TxIn.hs
+++ b/libs/cardano-ledger-core/src/Cardano/Ledger/TxIn.hs
@@ -6,7 +6,6 @@
 {-# LANGUAGE ExistentialQuantification #-}
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE FlexibleInstances #-}
-{-# LANGUAGE GADTs #-}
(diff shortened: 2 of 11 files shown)


lehins:
@redxaxder Here is the reason why we need to remove compacted `TxIn`, we need an instance of `Keyed` for `TxId`

@TimSheard if you are ok with this PR feel free to "Squash and Merge| it, it is opened into your `ts-compact` branch.

lehins:
After some experiments we found out that using HashMap that was implemented in #2520 is qyuite a bit more memory efficient then other approaches that involved manual unpacking of `TxIn`. Here is a comparison:

```
  Case                                  Max        MaxOS         Live       Allocated     GCs
  IntMap (KeyMap TxId ())       232,127,304  503,316,480  232,127,304  42,208,026,560  40,753
  KeyMap TxId (IntMap TxId ())  305,991,112  627,048,448  305,991,112  46,528,677,952  44,916
  IntMap (Map TxId ())          324,725,624  701,497,344  324,725,624  34,313,290,760  33,026
  Map TxIn ()                   383,746,832  896,532,480  383,746,832  33,154,708,560  31,937
```

In order for us to be able and use the most efficient from the top of the list, this PR is needed.

libs/cardano-ledger-core/src/Cardano/Ledger/Hashes.hs
```diff
@@ -75,13 +76,11 @@ newtype ScriptHash crypto
...
-  CC.Crypto crypto =>
-  ToCBOR (ScriptHash crypto)
+deriving newtype instance CC.Crypto crypto => Keyed (ScriptHash crypto)
 
-deriving newtype instance
-  CC.Crypto crypto =>
-  FromCBOR (ScriptHash crypto)
+deriving newtype instance CC.Crypto crypto => ToCBOR (ScriptHash crypto)
+
+deriving newtype instance CC.Crypto crypto => FromCBOR (ScriptHash crypto)
```
nc6, line 83: Where's the underlying `Keyed` instance coming from?

nc6: Ah, added in this PR :+1:

libs/cardano-ledger-core/src/Cardano/Ledger/TxIn.hs
```diff
@@ -72,101 +71,59 @@ newtype TxId crypto = TxId {_unTxId :: SafeHash crypto EraIndependentTxBody}
...
-  TxInCompact32 ::
-    HS.SizeHash (CC.HASH crypto) ~ 32 =>
-    {-# UNPACK #-} !Word64 -> -- Hash part 1/4
-    {-# UNPACK #-} !Word64 -> -- Hash part 2/4
-    {-# UNPACK #-} !Word64 -> -- Hash part 3/4
-    {-# UNPACK #-} !Word64 -> -- Hash part 4/4
-    {-# UNPACK #-} !Word64 -> -- Index
-    TxIn crypto
-  TxInCompactOther :: !(TxId crypto) -> {-# UNPACK #-} !Word64 -> TxIn crypto
+data TxIn crypto = TxInCompact !(TxId crypto) {-# UNPACK #-} !Int
```
nc6, line 87: Why the change to `Int` here?

nc6 (approved):
Everything looks reasonable. I have two questions, but they're not really review comments:

- The comment about needing a `Keyed` instance doesn't really motivate the non-unpacking.
- Why the change from `Word64` to `Int` for the index? Is this motivated by wanting to use `IntMap`?

libs/cardano-ledger-core/src/Cardano/Ledger/TxIn.hs
lehins: Because we will use it with `IntMap`, which needs an `Int`, rather than a `Word64`. I don't think it would a problem, right? Considering that max value on mainnet is 249 and I don't see tx index going larger than `maxBound :: Int` any time in the foreseeable future :smile:

libs/cardano-ledger-core/src/Cardano/Ledger/Hashes.hs
lehins: Yep :+1:

PR #2531, opened 2021-10-26
example of non-script output with datum hash

JaredCorduan:
This PR adds documentation (by way of a unit test / example) for the creation of non-script outputs containing a datum hash.


polinavino (approved):
LGTM! great way to address the double satisfaction problem haha

PR #2532, opened 2021-10-27
Add initial working implementation of `ledger-state` benchmarking tool

lehins:


JaredCorduan (approved):
This is great, thank you @lehins ! We can polish this library up later, I'm happy to merge this is now.

PR #2533, opened 2021-10-28
only store the go snapshot in the reward pulser

JaredCorduan:
The reward pulser is currently storing all three snapshots. It only needs to store the `go` stapshot and the fees, so this PR makes the adjustment.

Note that this is a change to the ledger state serialization (causing nodes to have to replay from genesis after upgrading). I think it is worth it, though. Though sharing might usually mean no extra memory is used, if a node is shut down and brought back up while the pulser is pulsing, the sharing is lost. In practice this means that if Deadalus is started up on day three of an epoch, it is holding in memory two stake distributions and two delegation mappings that it doesn't need to.


lehins (approved):
LGTM

PR #2534, opened 2021-10-28
Create a more compact TxOut using unpacked Word64s

DavidEichmann:
I've special cased when the address hash size is 28 bytes and the data
hash size is 32 bytes and the value is Ada only.

### DIFF
--- a/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxBody.hs
+++ b/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxBody.hs
@@ -80,7 +80,7 @@ import Cardano.Ledger.Coin (Coin (..))
 import Cardano.Ledger.Compactible
 import Cardano.Ledger.Core (PParamsDelta)
 import qualified Cardano.Ledger.Core as Core
-import Cardano.Ledger.Credential (Credential (..), PaymentCredential, StakeReference)
+import Cardano.Ledger.Credential (Credential (..), PaymentCredential, StakeReference (..))
 import qualified Cardano.Ledger.Crypto as CC
 import Cardano.Ledger.Era (Crypto, Era)
 import Cardano.Ledger.Hashes
@@ -141,7 +141,7 @@ data TxOut era
       !(DataHash (Crypto era))
   | SizeHash (CC.ADDRHASH (Crypto era)) ~ 28 =>
     TxOut_AddrHash28_AdaOnly
-      !(StakeReference (Crypto era))
+      !(Credential 'Staking (Crypto era))
       {-# UNPACK #-} !Word64 -- Payment Addr
       {-# UNPACK #-} !Word64 -- Payment Addr
       {-# UNPACK #-} !Word64 -- Payment Addr
@@ -151,7 +151,7 @@ data TxOut era
       SizeHash (CC.HASH (Crypto era)) ~ 32
     ) =>
     TxOut_AddrHash28_AdaOnly_DataHash32
-      !(StakeReference (Crypto era))
+      !(Credential 'Staking (Crypto era))
       {-# UNPACK #-} !Word64 -- Payment Addr
       {-# UNPACK #-} !Word64 -- Payment Addr
       {-# UNPACK #-} !Word64 -- Payment Addr
@@ -183,14 +183,14 @@ decodeAddress28 ::
   ( SizeHash (CC.ADDRHASH crypto) ~ 28,
     HashAlgorithm (CC.ADDRHASH crypto)
   ) =>
-  StakeReference crypto ->
+  Credential 'Staking crypto ->
   Word64 ->
   Word64 ->
   Word64 ->
   Word64 ->
   Addr crypto
 decodeAddress28 stakeRef a b c d =
-  Addr network paymentCred stakeRef
+  Addr network paymentCred (StakeRefBase stakeRef)
   where
     network = if d `testBit` 1 then Mainnet else Testnet
     paymentCred =
@@ -330,14 +330,16 @@ pattern TxOut addr vl dh <-
   (viewTxOut -> (addr, vl, dh))
   where
     TxOut (Addr network paymentCred stakeRef) vl SNothing
-      | Just adaCompact <- getAdaOnly (Proxy @era) vl,
+      | StakeRefBase stakeCred <- stakeRef,
+        Just adaCompact <- getAdaOnly (Proxy @era) vl,
         Just (Refl, a, b, c, d) <- encodeAddress28 network paymentCred =
-        TxOut_AddrHash28_AdaOnly stakeRef a b c d adaCompact
+        TxOut_AddrHash28_AdaOnly stakeCred a b c d adaCompact
     TxOut (Addr network paymentCred stakeRef) vl (SJust dh)
-      | Just adaCompact <- getAdaOnly (Proxy @era) vl,
+      | StakeRefBase stakeCred <- stakeRef,
+        Just adaCompact <- getAdaOnly (Proxy @era) vl,
         Just (Refl, a, b, c, d) <- encodeAddress28 network paymentCred,
         Just (Refl, e, f, g, h) <- encodeDataHash32 dh =
-        TxOut_AddrHash28_AdaOnly_DataHash32 stakeRef a b c d adaCompact e f g h
+        TxOut_AddrHash28_AdaOnly_DataHash32 stakeCred a b c d adaCompact e f g h
     TxOut addr vl mdh =
       let v = fromMaybe (error "Illegal value in txout") $ toCompact vl
           a = compactAddr addr
--- a/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxBody.hs
+++ b/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxBody.hs
@@ -4,6 +4,7 @@
 {-# LANGUAGE DerivingVia #-}
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE FlexibleInstances #-}
+{-# LANGUAGE GADTs #-}
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
 {-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE MultiParamTypeClasses #-}
@@ -12,9 +13,11 @@
 {-# LANGUAGE PatternSynonyms #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE StandaloneDeriving #-}
+{-# LANGUAGE TypeApplications #-}
 {-# LANGUAGE TypeFamilies #-}
 {-# LANGUAGE UndecidableInstances #-}
 {-# LANGUAGE ViewPatterns #-}
+{-# OPTIONS_GHC -Wno-redundant-constraints #-}
 
 module Cardano.Ledger.Alonzo.TxBody
   ( TxOut (TxOut, TxOutCompact, TxOutCompactDH),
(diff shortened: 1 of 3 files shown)


lehins:
Whats all the bit twiddling is about? Is there any reason you can't use `PackedBytes` exposed in https://github.com/input-output-hk/cardano-base/pull/243

dcoutts (commented):
Nice. I realise we cannot use `PackedBytes` (because that itself cannot be unpacked due to it being a multi-constructor type), but I think the code would be a little cleaner if we factored out a 4-word 28byte hash type (which would be single-constructor and thus unpackable). It'd give us the same representation, but would factor the code more concisely.

DavidEichmann:
Humm I have another branch that tries to extract things into something like `PackedBytes` but with a single constructor so it's unpackable. I'll update this PR to use that. I'm busy at the moment, so expect an update on Tuesday evening.

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxBody.hs
```diff
@@ -64,87 +67,341 @@ import Cardano.Binary
...
+        b26,
+        b27,
+        b28
+        ] ->
+          ( toWord64 b1 b2 b3 b4 b5 b6 b7 b8,
+            toWord64 b9 b10 b11 b12 b13 b14 b15 b16,
+            toWord64 b17 b18 b19 b20 b21 b22 b23 b24,
+            toWord64 b25 b26 b27 b28 0 0 0 (networkBit .&. payCredTypeBit)
+          )
+      _ -> error "Impossible! Wrong number of bytes"
```
lehins: By using `PackedBytes` we can avoid creating an intermediate `ByteString` and remove the need to convert hashes through lists and doing it byte-by-byte, all of which is much slower than it has to be.
```suggestion
    go :: SizeHash (CC.ADDRHASH crypto) ~ 28 => Hash (CC.ADDRHASH crypto) a -> (Word64, Word64, Word64, Word64)
    go h =
      case hashToPackedBytes h of
        PackedBytes28 a64 b64 c64 d32 -> (a64, b64, c64, (fromIntegral d32 `shiftL` 32) .|. fromIntegral (networkBit .&. payCredTypeBit))
        _ -> error "Impossible! Wrong number of bytes"
```

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxBody.hs
```diff
@@ -64,87 +67,341 @@ import Cardano.Binary
...
+  DataHash crypto
+decodeDataHash32 a b c d =
+  unsafeMakeSafeHash $
+    fromJust $
+      hashFromBytes $
+        BS.pack $
+          [ fromIntegral (w64 `shiftR` offset)
+            | w64 <- [a, b, c, d],
+              offset <- [56, 48 .. 0]
+          ]
```
lehins: Same here. Using lists in such a critical part of code for conversions of hash representation could have serious impact on performance. Also I recommend everyone to ban `fromJust` from their vocabulary. Not only it is a partial function, but it also does not provide any information on the location of the failure, when it does happen (no matter how impossible it is)

```suggestion
  hashFromPackedBytes (PackedBytes32 a b c d)
```

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxBody.hs
```diff
@@ -64,87 +67,341 @@ import Cardano.Binary
...
+    b30,
+    b31,
+    b32
+    ] ->
+      ( toWord64 b1 b2 b3 b4 b5 b6 b7 b8,
+        toWord64 b9 b10 b11 b12 b13 b14 b15 b16,
+        toWord64 b17 b18 b19 b20 b21 b22 b23 b24,
+        toWord64 b25 b26 b27 b28 b29 b30 b31 b32
+      )
+  _ -> error "Impossible! Wrong number of bytes"
```
lehins:
```suggestion
encodeDataHash32 dataHash = 
  case hashToPackedBytes (extractHash dataHash) of
    PackedBytes32 a b c d -> (a, b, c, d)
```

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxBody.hs
```diff
@@ -64,87 +67,341 @@ import Cardano.Binary
...
+      fromJust $
+        hashFromBytes $
+          BS.pack $
+            [ fromIntegral (w64 `shiftR` offset)
+              | w64 <- [a, b, c],
+                offset <- [56, 48 .. 0]
+            ]
+              ++ [ fromIntegral (d `shiftR` offset)
+                   | offset <- [24, 16 .. 0]
+                 ]
```
lehins:
```suggestion
    addrHash =
        hashFromPackedBytes $ PackedBytes28 a b c (fromIntegral (d `shiftR` 32))
```

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxBody.hs
```diff
@@ -64,87 +67,341 @@ import Cardano.Binary
...
     decodeMint,
     decodeNonNegative,
     encodeMint,
     isZero,
   )
+import Data.Bits
+import qualified Data.ByteString as BS
 import Data.Coders
-import Data.Maybe (fromMaybe)
+import Data.Maybe (fromJust, fromMaybe)
```
lehins: This function along with other like `head`, `tail`, etc. should be banned in production code `base`. It is better to use an explicit `error`, because not only it is more descriptive, but it also provides a stack trace.

```suggestion
import Data.Maybe (fromMaybe)
```

Here is a good example that I saw recently that depicts the problem very well: https://github.com/haskell/haskell-language-server/issues/1618

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxBody.hs
```diff
@@ -64,87 +67,341 @@ import Cardano.Binary
...
+    ] ->
+      ( toWord64 b1 b2 b3 b4 b5 b6 b7 b8,
+        toWord64 b9 b10 b11 b12 b13 b14 b15 b16,
+        toWord64 b17 b18 b19 b20 b21 b22 b23 b24,
+        toWord64 b25 b26 b27 b28 b29 b30 b31 b32
+      )
+  _ -> error "Impossible! Wrong number of bytes"
+
+toWord64 :: Word8 -> Word8 -> Word8 -> Word8 -> Word8 -> Word8 -> Word8 -> Word8 -> Word64
+toWord64 b1 b2 b3 b4 b5 b6 b7 b8 =
```
lehins: This function can be removed with suggested changes in this review

lehins (changes requested):
Here are more detailed suggestions that I meant by my https://github.com/input-output-hk/cardano-ledger-specs/pull/2534#issuecomment-954286006

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxBody.hs
DavidEichmann: Yes! great suggestion. I'll do that. I think that `.&.` is meant to be a `.|.` I'll also elaborate on the error too.

lehins: Oh yeah, you are right `.&.` is meant to be `.|.`. I was modifying your code when making the suggestion and didn't give that part much thought.

DavidEichmann:
@lehins after much fiddling with getting my gpg keys working, I've updated this PR with your suggestions.

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxBody.hs
```diff
@@ -64,87 +67,249 @@ import Cardano.Binary
...
+  where
+    network = if d `testBit` 1 then Mainnet else Testnet
+    paymentCred =
+      if d `testBit` 0
+        then KeyHashObj (KeyHash addrHash)
+        else ScriptHashObj (ScriptHash addrHash)
+    addrHash :: Hash (CC.ADDRHASH crypto) a
+    addrHash =
+      hashFromPackedBytes $
+        PackedBytes32 a b c (fromIntegral (d `shiftR` offset))
```
lehins: I'll be really surprised if this compiles. `PackedBytes32` is a 32 byte hash constructor, but we are decoding an address with 28 bytes. Also `offset` doesn't seem to be defined

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxBody.hs
```diff
@@ -168,17 +335,58 @@ pattern TxOut ::
...
+  ) =>
+  CompactAddr (Crypto era) ->
+  CompactForm (Core.Value era) ->
+  TxOut era
+pattern TxOutCompact addr vl <-
+  (viewCompactTxOut -> (addr, vl, SNothing))
+  where
+    TxOutCompact = TxOutCompact'
+
+-- TODO deprecate
```
lehins, line 365: I think it makes sense to add `DEPRECATED` pragmas right away.


```suggestion
{-# DEPRECATED TxOutCompactDH "In favor of `TxOut`. Do not use compacted versions directly." #-}
```

lehins:
Looks much better! :+1: However, there are still too many partial functions and some compilation errors that should be easy to fix.

So, here is a challenge, it is possible to get rid of all partial functions from this PR, if you won't be able to get to it tomorrow I'll make a commit with the fix, I just don't have any more time to get to it today.

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxBody.hs
DavidEichmann: So these patterns are still used in this repo which causes warnings. Do you know if we allow warning in the codebase (That doesn't seem like a good idea)? I was aiming to simply do a follow up PR removing this. Would that be sufficient?

lehins: Yeah, you are right. Disregard my comment, we'll deal with this at some later point.

DavidEichmann:
In terms of partial functions there is `valueToCompactErr` and in one other place I assume that compacting a Value will always succeed. I think this is an assumption we already make in master: https://github.com/input-output-hk/cardano-ledger-specs/blob/master/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxBody.hs#L174. Do you have some other solution in mind? I guess we could have a second non-partial version of the `Compactible` class with `toCompact :: a -> CompactForm a`, but that seems out of scope of this PR.

I also have `error` cases when matching on `PackedBytes` even though the number of bytes is known statically in the type. This is because the `PackedBytes#` constructor is defined for **all** sizes so we must have a catch all case and rely on the invariant that we don't use `PackedBytes#` when we can used the unpacked constructors. One solution would be to reimplement the constructors of PackedBytes into pattern synonyms that discharge the invariant as follows. Thoughts? Or is there somet else I'm missing?

```
{-# COMPLETE  PackedBytes32 #-}
pattern PackedBytes32 :: Word64 -> Word64 -> Word64 -> Word64 -> PackedBytes 32
pattern PackedBytes32 a b c d = view32 -> (a, b, c, d)

view32 :: PackedBytes 32 -> (Word64, Word64, Word64, Word64)
view32 (PackedBytes32 a b c d) = (a,b,c,d)
view32 (PackedBytes# _) = error "Impossible!"
```

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxBody.hs
lehins: These patterns are actually completely wrong, because when using them the new compact form cannot be constructed. In other words in all places where `TxOutCompact` and `TxOutCompactDH` are used for constructing `TxOut` will always construct the less optimal `TxOutCompact'` and `TxOutCompactDH'`

lehins (approved):
:+1:

PR #2536, opened 2021-11-03
Reward event to distinguish members and leaders

JaredCorduan:
DB-sync needs the reward event to distinguish between the member and leader rewards.


goolord (approved):
lgtm

PR #2537, opened 2021-11-03
Remove partial functions

lehins:
Also fixes `TxOutCompact` and  `TxOutCompactDH`  pattern synonyms

Removes `Show` constraint for `TxOut` pattern, since there is no reason to carry extra constraint for reporting errors on impossible cases.

Also does no uncompacting on Null and Ptr staking credentials.

Overall this gives us no memory overhead when compared with previous uncompacting, but allows us to share `Credential 'Staking`:

```
  Case                                   Max          MaxOS           Live        Allocated      GCs
  Original TxOut               1,091,506,912  2,864,709,632  1,091,506,912  196,483,571,888  173,309
  Adjusted TxOut no sharing    1,091,506,728  2,831,155,200  1,091,506,728  175,260,154,776  164,113
  Adjusted TxOut with sharing    969,811,552  2,312,110,080    969,811,552  179,412,416,608  168,094
```


JaredCorduan (approved):
Good catch, thanks @lehins !

DavidEichmann:
Ah, great! Thanks @lehins !

PR #2538, opened 2021-11-03
Introduced IncrementalStake, for all Eras 

TimSheard:
incrementally maintains the raw stake distribution as well as the UTxO. 
Adds InrementalStake field to UTxOState
Utxo rule (Utxos rule in Alonzo) and the Snap rule are modified to use this new state
Added a property test that computes both ways at every Epoch Boundary.

### DIFF
--- a/eras/shelley-ma/impl/src/Cardano/Ledger/ShelleyMA/Rules/Utxo.hs
+++ b/eras/shelley-ma/impl/src/Cardano/Ledger/ShelleyMA/Rules/Utxo.hs
@@ -39,7 +39,7 @@ import Cardano.Ledger.Shelley.Constraints
     UsesTxOut,
     UsesValue,
   )
-import Cardano.Ledger.Shelley.LedgerState (PPUPState)
+import Cardano.Ledger.Shelley.LedgerState (PPUPState, updateStakeDistribution)
 import qualified Cardano.Ledger.Shelley.LedgerState as Shelley
 import Cardano.Ledger.Shelley.PParams (PParams, PParams' (..), Update)
 import Cardano.Ledger.Shelley.Rules.Ppup (PPUP, PPUPEnv (..), PpupPredicateFailure)
@@ -248,7 +248,7 @@ utxoTransition ::
   TransitionRule (UTXO era)
 utxoTransition = do
   TRC (Shelley.UtxoEnv slot pp stakepools genDelegs, u, tx) <- judgmentContext
-  let Shelley.UTxOState utxo deposits' fees ppup = u
+  let Shelley.UTxOState utxo deposits' fees ppup incStake = u
   let txb = getField @"body" tx
 
   inInterval slot (getField @"vldt" txb)
@@ -336,13 +336,17 @@ utxoTransition = do
   let refunded = Shelley.keyRefunds pp txb
   let txCerts = toList $ getField @"certs" txb
   let depositChange = totalDeposits pp (`Map.notMember` stakepools) txCerts Val.<-> refunded
+  let utxoAdd = txouts txb -- These will be inserted into the UTxO
+  let utxoDel = eval (txins @era txb ◁ utxo) -- These will be deleted from the UTxO
+  let newIncStakeDistro = updateStakeDistribution @era incStake utxoDel utxoAdd
 
   pure
     Shelley.UTxOState
-      { Shelley._utxo = eval ((txins @era txb ⋪ utxo) ∪ txouts txb),
+      { Shelley._utxo = eval ((txins @era txb ⋪ utxo) ∪ utxoAdd),
         Shelley._deposited = deposits' <> depositChange,
         Shelley._fees = fees <> getField @"txfee" txb,
-        Shelley._ppups = ppup'
+        Shelley._ppups = ppup',
+        Shelley._stakeDistro = newIncStakeDistro
       }
 
 --------------------------------------------------------------------------------
--- a/eras/shelley/impl/src/Cardano/Ledger/Shelley/API/ByronTranslation.hs
+++ b/eras/shelley/impl/src/Cardano/Ledger/Shelley/API/ByronTranslation.hs
@@ -148,7 +148,8 @@ translateToShelleyLedgerState genesisShelley epochNo cvs =
               { _utxo = utxoShelley,
                 _deposited = Coin 0,
                 _fees = Coin 0,
-                _ppups = def
+                _ppups = def,
+                _stakeDistro = IStake mempty Map.empty
               },
           _delegationState =
             DPState
--- a/eras/shelley/impl/src/Cardano/Ledger/Shelley/API/Genesis.hs
+++ b/eras/shelley/impl/src/Cardano/Ledger/Shelley/API/Genesis.hs
@@ -27,6 +27,8 @@ import Cardano.Ledger.Shelley.API.Types
     word64ToCoin,
   )
 import Cardano.Ledger.Shelley.EpochBoundary (emptySnapShots)
+import Cardano.Ledger.Shelley.LedgerState (updateStakeDistribution)
+import Cardano.Ledger.Shelley.UTxO (UTxO (..))
 import Cardano.Ledger.Val (Val ((<->)))
 import Control.State.Transition (STS (State))
 import Data.Default.Class (Default, def)
@@ -67,6 +69,7 @@ instance
                   (Coin 0)
                   (Coin 0)
                   def
+                  (updateStakeDistribution mempty (UTxO mempty) initialUtxo)
               )
               (DPState (def {_genDelegs = GenDelegs genDelegs}) def)
           )
--- a/eras/shelley/impl/src/Cardano/Ledger/Shelley/EpochBoundary.hs
+++ b/eras/shelley/impl/src/Cardano/Ledger/Shelley/EpochBoundary.hs
@@ -26,7 +26,6 @@ module Cardano.Ledger.Shelley.EpochBoundary
     SnapShots (..),
     emptySnapShot,
     emptySnapShots,
-    aggregateUtxoCoinByCredential,
     poolStake,
     obligation,
     maxPool,
@@ -35,7 +34,6 @@ module Cardano.Ledger.Shelley.EpochBoundary
 where
 
 import Cardano.Binary (FromCBOR (..), ToCBOR (..), encodeListLen)
-import Cardano.Ledger.Address (Addr (..))
 import Cardano.Ledger.BaseTypes (BoundedRational (..), NonNegativeInterval)
 import Cardano.Ledger.Coin
   ( Coin (..),
@@ -43,23 +41,18 @@ import Cardano.Ledger.Coin
     rationalToCoinViaFloor,
   )
 import Cardano.Ledger.Compactible
-import qualified Cardano.Ledger.Core as Core
-import Cardano.Ledger.Credential (Credential, Ptr, StakeReference (..))
+import Cardano.Ledger.Credential (Credential)
 import qualified Cardano.Ledger.Crypto as CC (Crypto)
-import Cardano.Ledger.Era
 import Cardano.Ledger.Keys (KeyHash, KeyRole (..))
 import Cardano.Ledger.Serialization (decodeRecordNamedT)
 import Cardano.Ledger.Shelley.TxBody (PoolParams)
-import Cardano.Ledger.Shelley.UTxO (UTxO (..))
 import Cardano.Ledger.Val ((<+>), (<×>))
-import qualified Cardano.Ledger.Val as Val
 import Control.DeepSeq (NFData)
 import Control.Monad.Trans (lift)
 import Control.SetAlgebra (dom, eval, setSingleton, (▷), (◁))
 import Data.Compact.VMap as VMap
 import Data.Default.Class (Default, def)
 import Data.Map.Strict (Map)
-import qualified Data.Map.Strict as Map
 import Data.Ratio ((%))
 import Data.Sharing
 import Data.Typeable
@@ -87,40 +80,6 @@ instance CC.Crypto crypto => FromSharedCBOR (Stake crypto) where
 
 sumAllStake :: Stake crypto -> Coin
 sumAllStake = VMap.foldMap fromCompact . unStake
-
--- A TxOut has 4 different shapes, depending on the shape its embedded of Addr.
--- Credentials are stored in only 2 of the 4 cases.
--- 1) TxOut (Addr _ _ (StakeRefBase cred)) coin   -> HERE
--- 2) TxOut (Addr _ _ (StakeRefPtr ptr)) coin     -> HERE
--- 3) TxOut (Addr _ _ StakeRefNull) coin          -> NOT HERE
--- 4) TxOut (AddrBootstrap _) coin                -> NOT HERE
--- Unfortunately TxOut is a pattern, that deserializes the address. This can be expensive, so if
--- we only deserialize the parts that we need, for the 2 cases that count, we can speed
--- things up considerably. That is the role of deserialiseAddrStakeRef. It returns (Just stake)
--- for the two cases that matter, and Nothing for the other two cases.
-
--- | Sum up all the Coin for each staking Credential
-aggregateUtxoCoinByCredential ::
-  forall era.
-  ( Era era,
-    HasField "address" (Core.TxOut era) (Addr (Crypto era))
-  ) =>
-  Map Ptr (Credential 'Staking (Crypto era)) ->
-  UTxO era ->
-  Map (Credential 'Staking (Crypto era)) Coin ->
-  Map (Credential 'Staking (Crypto era)) Coin
-aggregateUtxoCoinByCredential ptrs (UTxO u) initial =
-  Map.foldl' accum initial u
-  where
-    accum !ans out =
-      case (getField @"address" out, getField @"value" out) of
-        (Addr _ _ (StakeRefPtr p), c) ->
-          case Map.lookup p ptrs of
-            Just cred -> Map.insertWith (<>) cred (Val.coin c) ans
-            Nothing -> ans
-        (Addr _ _ (StakeRefBase hk), c) ->
-          Map.insertWith (<>) hk (Val.coin c) ans
-        _other -> ans
 
 -- | Get stake of one pool
 poolStake ::
--- a/eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState.hs
+++ b/eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState.hs
@@ -1,4 +1,5 @@
 {-# LANGUAGE AllowAmbiguousTypes #-}
+{-# LANGUAGE BangPatterns #-}
 {-# LANGUAGE ConstraintKinds #-}
 {-# LANGUAGE DataKinds #-}
 {-# LANGUAGE DeriveAnyClass #-}
@@ -29,6 +30,7 @@ module Cardano.Ledger.Shelley.LedgerState
   ( AccountState (..),
     DPState (..),
     DState (..),
+    emptyDState,
     EpochState (..),
     UpecState (..),
     PulsingRewUpdate (..),
@@ -43,6 +45,8 @@ module Cardano.Ledger.Shelley.LedgerState
     RewardUpdate (..),
     RewardSnapShot (..),
     UTxOState (..),
+    smartUTxOState,
+    IncrementalStake (..),
     depositPoolChange,
     emptyRewardUpdate,
     pvCanFollow,
@@ -70,6 +74,9 @@ module Cardano.Ledger.Shelley.LedgerState
 
     -- * Epoch boundary
     stakeDistr,
+    incrementalStakeDistr,
+    updateStakeDistribution,
+    aggregateUtxoCoinByCredential,
     applyRUpd,
     applyRUpd',
     createRUpd,
@@ -123,7 +130,7 @@ import Cardano.Ledger.Coin
 import Cardano.Ledger.Compactible
 import Cardano.Ledger.Core (PParamsDelta)
 import qualified Cardano.Ledger.Core as Core
-import Cardano.Ledger.Credential (Credential (..))
+import Cardano.Ledger.Credential (Credential (..), StakeReference (StakeRefBase, StakeRefPtr))
 import qualified Cardano.Ledger.Crypto as CC (Crypto)
 import Cardano.Ledger.Era (Crypto, Era)
 import Cardano.Ledger.Keys
@@ -139,7 +146,7 @@ import Cardano.Ledger.Keys
   )
 import Cardano.Ledger.PoolDistr (PoolDistr (..))
 import Cardano.Ledger.SafeHash (HashAnnotated, extractHash, hashAnnotated)
-import Cardano.Ledger.Serialization (decodeRecordNamedT, mapToCBOR)
+import Cardano.Ledger.Serialization (decodeRecordNamedT, mapFromCBOR, mapToCBOR)
 import Cardano.Ledger.Shelley.Address.Bootstrap
   ( BootstrapWitness (..),
     bootstrapWitKeyHash,
@@ -158,7 +165,6 @@ import Cardano.Ledger.Shelley.EpochBoundary
   ( SnapShot (..),
     SnapShots (..),
     Stake (..),
-    aggregateUtxoCoinByCredential,
     sumAllStake,
   )
 import qualified Cardano.Ledger.Shelley.HardForks as HardForks
@@ -246,7 +252,7 @@ import qualified Data.Compact.VMap as VMap
 import Data.Constraint (Constraint)
 import Data.Default.Class (Default, def)
 import Data.Foldable (fold, toList)
-import Data.Group (invert)
+import Data.Group (Group, invert)
 import Data.Kind (Type)
 import Data.Map.Strict (Map)
 import qualified Data.Map.Strict as Map
@@ -567,11 +573,62 @@ pvCanFollow _ SNothing = True
 pvCanFollow (ProtVer m n) (SJust (ProtVer m' n')) =
   (m + 1, 0) == (m', n') || (m, n + 1) == (m', n')
 
+-- =============================
+
+-- | Incremental Stake, Stake along with possible missed coins from danging Ptrs.
+--   Transactions can use Ptrs to refer to a stake credential in a TxOut. The Ptr
+--   does not have to point to anything until the epoch boundary, when we compute
+--   rewards and aggregate staking information for ranking. This is unusual but legal.
+--   In a non incremental system, we use what ever 'legal' Ptrs exist at the epoch
+--   boundary. Here we are computing things incrementally, so we need to remember Ptrs
+--   that might point to something by the time the epoch boundary is reached. When
+--   the epoch boundary is reached we 'resolve' these pointers, to see if any have
+--   become non-dangling since the time they came into the incremental computation.
+data IncrementalStake crypto = IStake
+  { credMap :: !(Map (Credential 'Staking crypto) Coin),
+    ptrMap :: !(Map Ptr Coin)
+  }
+  deriving (Generic, Show, Eq, Ord, NoThunks, NFData)
+
+instance CC.Crypto crypto => ToCBOR (IncrementalStake crypto) where
+  toCBOR (IStake st dangle) =
+    encodeListLen 2 <> mapToCBOR st <> mapToCBOR dangle
+
+instance CC.Crypto crypto => FromSharedCBOR (IncrementalStake crypto) where
+  type Share (IncrementalStake crypto) = Interns (Credential 'Staking crypto)
+  fromSharedCBOR credInterns = do
+    decodeRecordNamed "Stake" (const 2) $ do
+      stake <- fromSharedCBOR (credInterns, mempty)
+      dangle <- mapFromCBOR
+      pure $ IStake stake dangle
+
+instance Semigroup (IncrementalStake c) where
+  (IStake a b) <> (IStake c d) = IStake (Map.unionWith (<>) a c) (Map.unionWith (<>) b d)
+
+instance Monoid (IncrementalStake c) where
+  mempty = (IStake Map.empty Map.empty)
+
+instance Data.Group.Group (IncrementalStake c) where
+  invert (IStake m1 m2) = IStake (Map.map invert m1) (Map.map invert m2)
+
+instance Default (IncrementalStake c) where
+  def = IStake Map.empty Map.empty
+
+-- =============================
+
+-- | There is a serious invariant that we must maintain in the UTxOState.
+--   Given (UTxOState utxo _ _ _ istake) it must be the case that
+--   istake == (updateStakeDistribution (IStake Map.empty Map.empty) (UTxO Map.empty) utxo)
+--   Of course computing the RHS of the above equality can be very expensive, so we only
+--   use this route in the testing function smartUTxO. But we are very carefull, wherever
+--   we update the UTxO, we carefully make INCREMENTAL changes to istake to maintain
+--   this invariant. This happens in the UTxO rule.
 data UTxOState era = UTxOState
   { _utxo :: !(UTxO era),
     _deposited :: !Coin,
     _fees :: !Coin,
-    _ppups :: !(State (Core.EraRule "PPUP" era))
+    _ppups :: !(State (Core.EraRule "PPUP" era)),
+    _stakeDistro :: !(IncrementalStake (Crypto era))
   }
   deriving (Generic)
 
@@ -601,8 +658,8 @@ deriving stock instance
 instance TransUTxOState NoThunks era => NoThunks (UTxOState era)
 
 instance TransUTxOState ToCBOR era => ToCBOR (UTxOState era) where
-  toCBOR (UTxOState ut dp fs us) =
-    encodeListLen 4 <> toCBOR ut <> toCBOR dp <> toCBOR fs <> toCBOR us
+  toCBOR (UTxOState ut dp fs us sd) =
+    encodeListLen 5 <> toCBOR ut <> toCBOR dp <> toCBOR fs <> toCBOR us <> toCBOR sd
 
 instance
   ( TransValue FromCBOR era,
@@ -617,12 +674,13 @@ instance
     Share (UTxOState era) =
       Interns (Credential 'Staking (Crypto era))
   fromSharedCBOR credInterns =
-    decodeRecordNamed "UTxOState" (const 4) $ do
+    decodeRecordNamed "UTxOState" (const 5) $ do
       _utxo <- fromSharedCBOR credInterns
       _deposited <- fromCBOR
       _fees <- fromCBOR
       _ppups <- fromCBOR
-      pure UTxOState {_utxo, _deposited, _fees, _ppups}
+      _stakeDistro <- fromSharedCBOR credInterns
+      pure UTxOState {_utxo, _deposited, _fees, _ppups, _stakeDistro}
 
 -- | New Epoch state and environment
 data NewEpochState era = NewEpochState
@@ -758,6 +816,7 @@ genesisState genDelegs0 utxo0 =
         (Coin 0)
         (Coin 0)
         def
+        (IStake mempty Map.empty)
     )
     (DPState dState def)
   where
@@ -855,6 +914,8 @@ consumed pp u tx =
     refunds = keyRefunds pp tx
     withdrawals = fold . unWdrl $ getField @"wdrls" tx
 
+-- ====================================================
+
 newtype WitHashes crypto = WitHashes
   {unWitHashes :: Set (KeyHash 'Witness crypto)}
   deriving (Eq, Generic)
@@ -897,8 +958,7 @@ witsVKeyNeeded ::
     HasField "wdrls" (Core.TxBody era) (Wdrl (Crypto era)),
     HasField "certs" (Core.TxBody era) (StrictSeq (DCert (Crypto era))),
     HasField "inputs" (Core.TxBody era) (Set (TxIn (Crypto era))),
-    HasField "update" (Core.TxBody era) (StrictMaybe (Update era)),
-    HasField "address" (Core.TxOut era) (Addr (Crypto era))
+    HasField "update" (Core.TxBody era) (StrictMaybe (Update era))
   ) =>
   UTxO era ->
   tx ->
@@ -1037,9 +1097,11 @@ reapRewards dStateRewards withdrawals =
 -- epoch boundary calculations --
 ---------------------------------
 
+-- | Compute the current Stake Distribution. This was called at the Epoch boundary in the Snap Rule.
+--   Now its is called in the tests to see that its incremental analog 'incrementaStakeDistr' agrees.
 stakeDistr ::
   forall era.
-  (Era era, HasField "address" (Core.TxOut era) (Addr (Crypto era))) =>
+  Era era =>
   UTxO era ->
   DState (Crypto era) ->
   PState (Crypto era) ->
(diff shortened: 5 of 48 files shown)


eras/shelley/impl/src/Cardano/Ledger/Shelley/EpochBoundary.hs
```diff
@@ -66,49 +60,12 @@ import Numeric.Natural (Natural)
...
-  forall era.
-  ( Era era,
-    HasField "address" (Core.TxOut era) (Addr (Crypto era))
-  ) =>
-  Map Ptr (Credential 'Staking (Crypto era)) ->
-  UTxO era ->
-  Map (Credential 'Staking (Crypto era)) Coin ->
-  Map (Credential 'Staking (Crypto era)) Coin
-aggregateUtxoCoinByCredential ptrs (UTxO u) initial =
-  Map.foldl' accum initial u
```
TimSheard, line 113: This function has been moved to LedgerState.hs where it has an Incremental analog

libs/small-steps-test/src/Control/State/Transition/Trace.hs
```diff
@@ -495,7 +495,7 @@ applySTSTest = applySTSOptsEither defaultOpts
   where
     defaultOpts =
       ApplySTSOpts
-        { asoAssertions = AssertionsAll,
+        { asoAssertions = AssertionsOff,
           asoValidation = ValidateAll,
           asoEvents = EPDiscard
```
TimSheard: Not sure why this was changed. Not sure I did this.

JaredCorduan: I think you started with my original branch, where I was doing that for debugging purposes. We shouldn't check this in.

eras/alonzo/test-suite/test/Tests.hs
```diff
@@ -56,10 +62,18 @@ nightlyTests :: TestTree
 nightlyTests =
   testGroup
     "Alonzo tests"
-    [ alonzoPropertyTests, -- These are the full property tests
+    [ incrementalTests,
```
JaredCorduan: can we put `incrementalTests` inside `alonzoPropertyTests`?

eras/shelley-ma/impl/src/Cardano/Ledger/ShelleyMA/Rules/Utxo.hs
```diff
@@ -337,13 +337,17 @@ utxoTransition = do
...
   let txCerts = toList $ getField @"certs" txb
   let depositChange = totalDeposits pp (`Map.notMember` stakepools) txCerts Val.<-> refunded
+  let utxoAdd = txouts txb -- These will be inserted into the UTxO
+  let utxoDel = eval (txins @era txb ◁ utxo) -- These will be deleted fromthe UTxO
+  let newIncStakeDistro = updateStakeDistribution @era incStake utxoDel utxoAdd
 
   pure
     Shelley.UTxOState
-      { Shelley._utxo = eval ((txins @era txb ⋪ utxo) ∪ txouts txb),
+      { Shelley._utxo = eval ((txins @era txb ⋪ utxo) ∪ utxoAdd), -- Domain exclusion (a ⋪ b) deletes 'a' from 'b'
```
JaredCorduan:
```suggestion
      { Shelley._utxo = eval ((txins @era txb ⋪ utxo) ∪ utxoAdd), -- Domain exclusion (a ⋪ b) deletes 'a' from the domain of 'b'
```

eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState.hs
```diff
@@ -530,11 +536,46 @@ pvCanFollow _ SNothing = True
 pvCanFollow (ProtVer m n) (SJust (ProtVer m' n')) =
   (m + 1, 0) == (m', n') || (m, n + 1) == (m', n')
 
+-- =============================
+
+-- | Incremental Stake, Stake along with possible missed coins from danging Ptrs
+data IncrementalStake crypto = IStake
+  { getStake :: !(Map (Credential 'Staking crypto) Coin),
+    dangling :: !(Map Ptr Coin)
```
JaredCorduan: These are not really dangling anymore. Originally, we were only including the dangling pointers in this structure, but now it is _all_ the pointers. I think we need a new name. :)

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Snap.hs
```diff
@@ -52,18 +54,19 @@ instance (UsesTxOut era, UsesValue era) => STS (SNAP era) where
...
-  ) =>
+  forall era.
   TransitionRule (SNAP era)
 snapTransition = do
   TRC (lstate, s, _) <- judgmentContext
 
-  let LedgerState (UTxOState utxo _ fees _) (DPState dstate pstate) = lstate
-      stake = stakeDistr utxo dstate pstate
+  let LedgerState (UTxOState _utxo _ fees _ incStake) (DPState dstate pstate) = lstate
+      -- stakeSnap = stakeDistr @era utxo dstate pstate
```
JaredCorduan: we can delete this comment. bye bye :wave: `stakeSnap` :tada:

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Utxo.hs
```diff
@@ -41,6 +41,7 @@ import Cardano.Ledger.BaseTypes
     networkId,
   )
 import Cardano.Ledger.Coin (Coin (..))
+-- import Cardano.Ledger.Mary (Value(..))
```
JaredCorduan: let's delete this

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Utxo.hs
```diff
@@ -59,20 +60,24 @@ import Cardano.Ledger.Shelley.Constraints
...
+    updateStakeDistribution,
   )
 import Cardano.Ledger.Shelley.PParams (PParams, PParams' (..), Update)
 import Cardano.Ledger.Shelley.Rules.Ppup (PPUP, PPUPEnv (..), PpupEvent, PpupPredicateFailure)
-import Cardano.Ledger.Shelley.Tx (Tx (..), TxIn)
+import Cardano.Ledger.Shelley.Tx (Tx (..), TxIn, TxOut (..))
 import Cardano.Ledger.Shelley.TxBody
   ( DCert,
     PoolParams,
+    -- Ptr,
```
JaredCorduan: delete

libs/cardano-ledger-core/src/Cardano/Ledger/TerseTools.hs
```diff
@@ -0,0 +1,107 @@
+{-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE UndecidableInstances #-}
+
+module Cardano.Ledger.TerseTools where
```
JaredCorduan: maybe this could live in `libs/cardano-ledger-pretty` instead?

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Utxo.hs
```diff
@@ -124,8 +129,9 @@ data UtxoEnv era
   = UtxoEnv
       SlotNo
       (Core.PParams era)
-      (Map (KeyHash 'StakePool (Crypto era)) (PoolParams (Crypto era)))
+      (Map (KeyHash 'StakePool (Crypto era)) (PoolParams (Crypto era))) -- DO WE NEED THIS, IN DPSTATE
```
JaredCorduan: We do need the pool parameters in the UTXO environment, so that we can recognize which stake pool registration certificates are new and which are re-registrations (only the former pay a deposit).

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Utxo.hs
```diff
@@ -124,8 +129,9 @@ data UtxoEnv era
...
       (GenDelegs (Crypto era))
+      (DPState (Crypto era)) -- (Map Ptr (Credential 'Staking (Crypto era)))
```
JaredCorduan: I don't think we need this in the UTXO environment anymore.

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Utxow.hs
```diff
@@ -269,8 +269,8 @@ initialLedgerStateUTXOW ::
   ) =>
   InitialRule (UTXOW era)
 initialLedgerStateUTXOW = do
-  IRC (UtxoEnv slots pp stakepools genDelegs) <- judgmentContext
-  trans @(Core.EraRule "UTXO" era) $ IRC (UtxoEnv slots pp stakepools genDelegs)
+  IRC (UtxoEnv slots pp stakepools genDelegs stakeDistr) <- judgmentContext
```
JaredCorduan: isn't the last value in `UTxOEnv` a `DPState`? either way, I don't think we need it.

JaredCorduan (commented):
@TimSheard overall this is really awesome!! I made some minor comments.

libs/small-steps-test/src/Control/State/Transition/Trace.hs
TimSheard: I set it back to AssertionsAll

eras/alonzo/test-suite/test/Tests.hs
TimSheard: Done

eras/shelley-ma/impl/src/Cardano/Ledger/ShelleyMA/Rules/Utxo.hs
TimSheard: done

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Snap.hs
TimSheard: I was thinking we might leave it as a bit of history, if we ever wanted to back this out, this tells us where we would have to make the one big important change. Think about it.

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Utxo.hs
TimSheard: I thought I had already did that. Well it is done now.

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Utxo.hs
TimSheard: done

eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState.hs
TimSheard: We now use credMap and ptrMap
`data IncrementalStake crypto = IStake
  { credMap :: !(Map (Credential 'Staking crypto) Coin),
    ptrMap :: !(Map Ptr Coin)
  }`

libs/cardano-ledger-core/src/Cardano/Ledger/TerseTools.hs
TimSheard: The problem with that is, that it can't be used when debugging Shelley code, because of mutual recursion problems.  The point of TerseTools is to make "terse" representations of the differences between two objects/ Not to pretty print them.

JaredCorduan: ah. how about `eras/shelley/impl/src/Cardano/Ledger/Terse.hs`?

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Snap.hs
JaredCorduan: How about it we just pretty it up a bit:

```
-- | The stake distribution was previously computed as in the spec:
--
-- @
--  stakeDistr @era utxo dstate pstate
-- @
--
-- but is now computed incrementally.
```
```

File: eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState.hs [hunk 13]
```diff
@@ -564,8 +606,8 @@ deriving stock instance
 instance TransUTxOState NoThunks era => NoThunks (UTxOState era)
 
 instance TransUTxOState ToCBOR era => ToCBOR (UTxOState era) where
-  toCBOR (UTxOState ut dp fs us) =
-    encodeListLen 4 <> toCBOR ut <> toCBOR dp <> toCBOR fs <> toCBOR us
+  toCBOR (UTxOState ut dp fs us sd) =
+    encodeListLen 5 <> toCBOR ut <> toCBOR dp <> toCBOR fs <> toCBOR us <> toCBOR sd
```
lehins, line 662: If I understood @dcoutts correctly, in order to be able to implement sharing upon deserialization this incremental stake map needs to come before the utxo. Not sure what are the implications of this would be. Duncan, could you please comment on this.

eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState.hs
```diff
@@ -1009,6 +1056,139 @@ stakeDistr u ds ps =
...
+  IncrementalStake (Crypto era)
+incrementalAggregateUtxoCoinByCredential mode (UTxO u) initial =
+  Map.foldl' accum initial u
+  where
+    keepOrDelete new Nothing = case (mode new) of Coin 0 -> Nothing; final -> Just final
+    keepOrDelete new (Just old) =
+      case (mode new) <> old of
+        Coin 0 -> Nothing
+        final -> Just final
+    accum ans@(!(IStake stake ptrs)) out =
```
lehins: Pattern matching already makes it strict, right?

```suggestion
    accum ans@(IStake stake ptrs) out =
```

eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState.hs
```diff
@@ -1009,6 +1056,139 @@ stakeDistr u ds ps =
...
+      case (getField @"address" out, getField @"value" out) of
+        (Addr _ _ (StakeRefPtr p), c) -> IStake stake (Map.alter (keepOrDelete (Val.coin c)) p ptrs)
+        (Addr _ _ (StakeRefBase hk), c) -> IStake (Map.alter (keepOrDelete (Val.coin c)) hk stake) ptrs
+        _other -> ans
```
lehins: Reducing duplication and avoiding unnecessary tuple pattern matching:
```suggestion
      let c = Val.coin (getField @"value" out)
      in case getField @"address" out of
           Addr _ _ (StakeRefPtr p) -> IStake stake (Map.alter (keepOrDelete c) p ptrs)
           Addr _ _ (StakeRefBase hk) -> IStake (Map.alter (keepOrDelete c) hk stake) ptrs
           _other -> ans
```

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Snap.hs
```diff
@@ -3,6 +3,9 @@
 {-# LANGUAGE EmptyDataDeriving #-}
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE FlexibleInstances #-}
+{-# LANGUAGE RankNTypes #-}
```
lehins: Was this extension added intentionally? It doesn't look like it is being used:
```suggestion
```

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Snap.hs
lehins: I have no idea if this line would be useful in the future or not, but leaving a piece of code commented out without an explanation what it is for is never a good idea. It look like dead code. Having a comment about why such thing was left over is almost like a requirement in my books.

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Utxo.hs
```diff
@@ -431,13 +438,18 @@ utxoInductive = do
   let totalDeposits' = totalDeposits pp (`Map.notMember` stakepools) txCerts
   tellEvent $ TotalDeposits totalDeposits'
   let depositChange = totalDeposits' <-> refunded
+  let utxoAdd = txouts txb -- These will be inserted into the UTxO
+  let utxoDel = eval (txins @era txb ◁ utxo) -- These will be deleted fromthe UTxO
```
lehins:
```suggestion
  let utxoDel = eval (txins @era txb ◁ utxo) -- These will be deleted from the UTxO
```

eras/shelley/test-suite/src/Test/Cardano/Ledger/Shelley/Rules/TestChain.hs
```diff
@@ -182,6 +186,76 @@ collisionFreeComplete =
...
+                  "\nfinal sd\n",
+                  show sd',
+                  "\noriginal ptrs\n",
+                  show ptrs,
+                  "\nfinal ptrs\n",
+                  show ptrs'
+                ]
+              )
+          )
+          $ counterExampleTooBig u' .||. utxoBal === incrStakeBal
```
lehins: Why are we passing a test on large input? If this is for performance reasons, then can we resize a generator to produce smaller utxo and discard tests with sizes that are too large? If not where did value `50` came from?
```suggestion
          $ counterExampleTooBig u' ==> utxoBal === incrStakeBal
```

libs/cardano-ledger-core/src/Cardano/Ledger/TerseTools.hs
lehins: From what I can see this TerseTools module is used only in testing. This means it has no business being neither in `cardano-ledger-core` nor in `cardano-ledger-shelley`. It's ought to be added to `cardano-ledger-shelley-test`

libs/cardano-ledger-core/src/Cardano/Ledger/TerseTools.hs
```diff
@@ -0,0 +1,107 @@
...
+{-
+tersediffincremental :: String -> IncrementalStake crypto -> IncrementalStake crypto -> String
+tersediffincremental message (IStake a b) (IStake c d) =
+   tersemapdiffs (message++" "++"hashes") a c ++
+   tersemapdiffs (message++" "++"ptrs") b d
+
+terseutxo :: (Era era,Terse (Core.TxOut era)) => String -> UTxO era -> String
+terseutxo message (UTxO mp) = terselist message (Map.toList mp)
+
+-}
```
lehins: Dead code? Delete?

eras/shelley/test-suite/src/Test/Cardano/Ledger/Shelley/Rules/TestChain.hs
```diff
@@ -1141,3 +1215,58 @@ sameEpoch SourceSignalTarget {source, target} =
...
+  LedgerState era ->
+  LedgerState era ->
+  Property
+testIncrementalStake _ (LedgerState (UTxOState utxo _ _ _ incStake) (DPState dstate pstate)) =
+  let stake = stakeDistr @era utxo dstate pstate
+
+      istake = incrementalStakeDistr @era incStake dstate pstate
+   in counterexample
+        ( "\nIncremental stake distribution does not match old style stake distribution"
+            ++ tersediffincremental "differences: Old vs Incremental" (unSnapshot stake) (unSnapshot istake)
```
lehins: Any reason why we can't use the existing accessor functions?
```suggestion
            ++ tersediffincremental "differences: Old vs Incremental" (_stake stake) (_stake istake)
```

eras/shelley-ma/impl/src/Cardano/Ledger/ShelleyMA/Rules/Utxo.hs
```diff
@@ -337,13 +337,18 @@ utxoTransition = do
   let refunded = Shelley.keyRefunds pp txb
   let txCerts = toList $ getField @"certs" txb
   let depositChange = totalDeposits pp (`Map.notMember` stakepools) txCerts Val.<-> refunded
+  let utxoAdd = txouts txb -- These will be inserted into the UTxO
+  let utxoDel = eval (txins @era txb ◁ utxo) -- These will be deleted fromthe UTxO
```
lehins:
```suggestion
  let utxoDel = eval (txins @era txb ◁ utxo) -- These will be deleted from the UTxO
```

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Utxo.hs
TimSheard: I see, we will leave it.

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Utxo.hs
TimSheard: Good observation. It is now gone!

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Utxow.hs
TimSheard: Its gone so the name doesn't matter any more. I think it harks back to the original branch, which we changed about 10 times, but since the UtxoEnv had the same number of fields, and the only use we made of that variable, was to pass it to another call to the constructor UtxoEnv, the name didn't matter.

libs/cardano-ledger-core/src/Cardano/Ledger/TerseTools.hs
TimSheard: So moved.

eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState.hs
```diff
@@ -1009,6 +1056,139 @@ stakeDistr u ds ps =
...
+  IncrementalStake (Crypto era) ->
+  IncrementalStake (Crypto era)
+incrementalAggregateUtxoCoinByCredential mode (UTxO u) initial =
+  Map.foldl' accum initial u
+  where
+    keepOrDelete new Nothing = case (mode new) of Coin 0 -> Nothing; final -> Just final
+    keepOrDelete new (Just old) =
+      case (mode new) <> old of
+        Coin 0 -> Nothing
+        final -> Just final
```
lehins: I can't imagine that we allow negative Coins. But if it ever happened this sort of arithmetic always make me worry that someone can exploit to make the sum of coins to be zero, while coins themselves are equal in magnitude but not in the sign. I am not quite sure, maybe this is even desired 

In any case this is just a suggestion:

```suggestion
    keepOrDelete new mOld = do
      let c = mode new
      guard (c /= Coin 0)
      guard (mOld /= Just (Coin 0))
      Just c <> mOld
```

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Snap.hs
TimSheard: Done

lehins (approved):
I don't know enough about the staking logic to comment on the implementation, but here are my comments with respect to Haskell.

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Snap.hs
TimSheard: Probably was added one some change was made, which was then later retracted. I have removed it.

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Utxo.hs
TimSheard: fixed it.

libs/cardano-ledger-core/src/Cardano/Ledger/TerseTools.hs
TimSheard: These are very usefulldebugging functions, which can't be defined here since UTxO and IncrementalStake, cannot be imported into this function. Someone will need them again, so I leave then here. so the source is not lost.

eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState.hs
TimSheard: Good eyes. It is now fixed.

eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState.hs
TimSheard: At some point I was pattern matching that the coin was 0, and that got reverted when we added the function keepOrDelete.  So I agree this is a better version.

eras/shelley/test-suite/src/Test/Cardano/Ledger/Shelley/Rules/TestChain.hs
TimSheard: I guess, at the time, it was easier to define it, than lookup what it was called. Now it is done.

eras/shelley-ma/impl/src/Cardano/Ledger/ShelleyMA/Rules/Utxo.hs
TimSheard: done

eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState.hs
TimSheard: What is going on here, is we are deleting any entries where the coins drop to 0. This will naturally happen in the course of things. The Nothing means delete this entry. So I think this is better code. In fact the IStake keeps track of both positive and negative Coins, think of this as the change in the UtxO. When we apply the change, negative coins are subtracted. There is an invariant, that is being maintained here, which we work hard to maintain, and that ensures, that after applying the change all values are strictly greater than 0.

eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState.hs
TimSheard: Right now there are CDDL invariants that are being maintained. If we change those, we can change the order in which things are serialized. I can see why that would be a good idea. But we have to wait for the change in the CDDL to do that.

eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState.hs
lehins: > In fact the IStake keeps track of both positive and negative Coins

This sort of arithmetic on money always makes me cringe. But that is a totally different discussion, so I am not gonna say anything more about that.


With regards to this function, consider this change for brevity that preserves the semantics:

```haskell
    keepOrDelete new mOld = do
      final <- Just (mode new) <> mOld
      guard (final /= Coin 0)
      Just final
```

JaredCorduan (approved):

eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState.hs
lehins: If you'd rather keep the version you have then we should at least use consistent formatting and remove redundant parenthesis:
```haskell
    keepOrDelete new Nothing = 
      case mode new of
        Coin 0 -> Nothing
        final -> Just final
    keepOrDelete new (Just old) =
      case mode new <> old of
        Coin 0 -> Nothing
        final -> Just final
```

TimSheard: OK lets use this one.

libs/cardano-ledger-core/src/Cardano/Ledger/TerseTools.hs
TimSheard: In the end, I moved the whole file to the test-suite, so these can actually be defined.

eras/shelley/test-suite/src/Test/Cardano/Ledger/Shelley/Rules/TestChain.hs
TimSheard: I inherited this from @JaredCorduan , perhaps he can better explain this.

JaredCorduan: good catch @lehins ! sorry I only just now noticed this. This was never intended to be merged! I put it there for my own sanity while debugging, but now that tests are passing we need to remove it.

JaredCorduan:
```suggestion
          $ utxoBal === incrStakeBal
```

nc6 (changes requested):
Just marking this as request changes so we don't accidentally merge - this branch is known to cause problems during benchmark run.

eras/shelley/impl/src/Cardano/Ledger/Shelley/API/ByronTranslation.hs
```diff
@@ -148,7 +148,8 @@ translateToShelleyLedgerState genesisShelley epochNo cvs =
               { _utxo = utxoShelley,
                 _deposited = Coin 0,
                 _fees = Coin 0,
-                _ppups = def
+                _ppups = def,
+                _stakeDistro = IStake mempty Map.empty
```
JaredCorduan, line 152: I think this is wrong, when we translate from Byron, the initial incremental stake needs to be
```
aggregateUtxoCoinByCredential Map.empty utxoShelley
```

We can use `Map.empty` for the pointer map since there will be no registered stake credentials at the start of Shelley.

eras/shelley/impl/src/Cardano/Ledger/Shelley/API/ByronTranslation.hs
```diff
@@ -148,7 +149,8 @@ translateToShelleyLedgerState genesisShelley epochNo cvs =
               { _utxo = utxoShelley,
                 _deposited = Coin 0,
                 _fees = Coin 0,
-                _ppups = def
+                _ppups = def,
+                _stakeDistro = updateStakeDistribution mempty (UTxO mempty) utxoShelley
```
JaredCorduan: I am now thinking that this is not needed, Byron lovelace contains no stake!

eras/shelley/impl/src/Cardano/Ledger/Shelley/API/Genesis.hs
```diff
@@ -67,6 +68,7 @@ instance
                   (Coin 0)
                   (Coin 0)
                   def
+                  (IStake mempty Map.empty)
```
JaredCorduan: I bet this is what is causing trouble with the benchmarking. the stake in `initialUtxo` needs to be accounted for in the incremental stake.

deepfire:
@nc6, this is green from benchmarking standpoint.

eras/shelley/impl/src/Cardano/Ledger/Shelley/API/Genesis.hs
deepfire: Turned out that was spot-on, thanks a lot @JaredCorduan !

JaredCorduan: :raised_hands:

eras/shelley-ma/impl/src/Cardano/Ledger/ShelleyMA/Rules/Utxo.hs
```diff
@@ -337,13 +337,18 @@ utxoTransition = do
...
   let depositChange = totalDeposits pp (`Map.notMember` stakepools) txCerts Val.<-> refunded
+  let utxoAdd = txouts txb -- These will be inserted into the UTxO
+  let utxoDel = eval (txins @era txb ◁ utxo) -- These will be deleted from the UTxO
+  let newIncStakeDistro = updateStakeDistribution @era incStake utxoDel utxoAdd
 
   pure
     Shelley.UTxOState
-      { Shelley._utxo = eval ((txins @era txb ⋪ utxo) ∪ txouts txb),
+      { Shelley._utxo = eval ((txins @era txb ⋪ utxo) ∪ utxoAdd),
+        -- Domain exclusion (a ⋪ b) deletes 'a' from the domain of 'b'
```
nc6:
```suggestion
```

This is a standard function used in various places, and the documentation can be seen through haddock etc

eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState.hs
```diff
@@ -536,11 +543,53 @@ pvCanFollow _ SNothing = True
 pvCanFollow (ProtVer m n) (SJust (ProtVer m' n')) =
   (m + 1, 0) == (m', n') || (m, n + 1) == (m', n')
 
+-- =============================
+
+-- | Incremental Stake, Stake along with possible missed coins from danging Ptrs
```
nc6: Can you elaborate the scenario described in this comment?

eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState.hs
```diff
@@ -995,9 +1047,11 @@ reapRewards dStateRewards withdrawals =
 -- epoch boundary calculations --
 ---------------------------------
 
+-- | Compute the current Stake Distribution. This was called at the Epoch boundary in the Snap Rule.
```
nc6, line 1100: If it's only used in tests, can we move it to a testing package?

eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState.hs
```diff
@@ -1014,10 +1068,149 @@ stakeDistr u ds ps =
     stakeRelation = aggregateUtxoCoinByCredential (forwards ptrs') u rewards'
     activeDelegs :: Map (Credential 'Staking (Crypto era)) (KeyHash 'StakePool (Crypto era))
     activeDelegs = eval ((dom rewards' ◁ delegs) ▷ dom poolParams)
-    compactCoinOrError c =
-      case toCompact c of
-        Nothing -> error $ "Invalid ADA value in staking: " <> show c
-        Just compactCoin -> compactCoin
+
+-- A TxOut has 4 different shapes, depending on the shape of it's embedded Addr.
```
nc6:
```suggestion
-- A TxOut has 4 different shapes, depending on the shape of its embedded Addr.
```

eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState.hs
```diff
@@ -1014,10 +1068,149 @@ stakeDistr u ds ps =
...
+       in case getField @"address" out of
+            Addr _ _ (StakeRefPtr p) -> IStake stake (Map.alter (keepOrDelete c) p ptrs)
+            Addr _ _ (StakeRefBase hk) -> IStake (Map.alter (keepOrDelete c) hk stake) ptrs
+            _other -> ans
+
+-- | Resolve inserts and deletes which were indexed by Ptrs, by looking them up in 'ptrs'
+--   and combining them with the ordinary stake. Zero out the Ptr indexed stake in the result.
+--   This function is meant to only be called at the Epoch boundary in the Snap Rule. It is not meant
+--   to be used to transform the IncrementalStake.
+resolveIncrementalPtrs :: Map Ptr (Credential 'Staking (Crypto era)) -> IncrementalStake (Crypto era) -> IncrementalStake (Crypto era)
```
nc6: Is there a reason in this case for returning `IncrementalStake` rather than simply `Map Credential Coin`? I assume we zero out the incremental stuff at the epoch boundary anyway?

eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState.hs
```diff
@@ -1014,10 +1068,149 @@ stakeDistr u ds ps =
...
+resolveIncrementalPtrs ptrs (IStake stake byPtr) = IStake (Map.foldlWithKey' accum stake byPtr) Map.empty
+  where
+    accum ans ptr coin =
+      case Map.lookup ptr ptrs of
+        Nothing -> ans
+        Just hash -> Map.insertWith (<>) hash coin ans
+
+compactCoinOrError :: Coin -> CompactForm Coin
```
nc6: Can we move this inside `incrementalStakeDistr`, since it's a bit of a weird function and in any case only used there?

eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState.hs
```diff
@@ -1014,10 +1068,149 @@ stakeDistr u ds ps =
...
+compactCoinOrError c =
+  case toCompact c of
+    Nothing -> error $ "Invalid ADA value in staking: " <> show c
+    Just compactCoin -> compactCoin
+
+-- | Compute the current state distribution by using the IncrementalStake, which is an aggregate
+--   of the current UTxO. This function has a non-incremental analog 'takeDistr'
```
nc6:
```suggestion
--   of the current UTxO. This function has a non-incremental analog 'stakeDistr'
```

eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState.hs
```diff
@@ -1014,10 +1068,149 @@ stakeDistr u ds ps =
...
+--   analog 'aggregateUtxoCoinByCredential'
```
nc6: > his function has a non-incremental analog 'takeDistr' analog 'aggregateUtxoCoinByCredential'

This comment seems weird and wrong?

eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState.hs
```diff
@@ -69,6 +73,10 @@ module Cardano.Ledger.Shelley.LedgerState
 
     -- * Epoch boundary
     stakeDistr,
+    incrementalStakeDistr,
+    updateStakeDistribution,
+    resolveIncrementalPtrs,
```
nc6: Is there a reason to export this? Seems like it's only used in `incrementalStakeDistr`, and since it has a fairly heavy use restriction on it, would be better to keep it internal.
```suggestion
```

eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState.hs
nc6: Alternatively, just move this function inside of `incrementalStakeDistr`, which is the only thing to use it anyway.

eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState.hs
```diff
@@ -1014,10 +1068,149 @@ stakeDistr u ds ps =
...
+
+-- | A valid (or self-consistent) UTxOState{_utxo, _deposited, _fees, _ppups, _stakeDistro}
+--   maintains an invariant between the _utxo and _stakeDistro fields. the _stakeDistro field is
+--   the aggregation of Coin over the StakeReferences in the UTxO. It can be computed by a pure
+--   function from the _utxo field. In some situations, mostly unit or example tests, or when
+--   initializing a small UTxO, we want to create a UTxOState that computes the _stakeDistro from
+--   the _utxo. This is aways safe to do, but if the _utxo field is big, this can be very expensive,
+--   which defeats the purpose of memoizing the _stakeDistro field. So use of this function should be
+--   restricted to tests and initializations, where the invariant should be maintained.
+smartUTxOState ::
```
nc6, line 1248: Better name? Not that I have any good ideas...

eras/shelley/test-suite/src/Test/Cardano/Ledger/Shelley/Address/Bootstrap.hs
```diff
@@ -156,7 +157,8 @@ utxoState0 =
     { _utxo = utxo0,
       _deposited = Coin 0,
       _fees = Coin 0,
-      _ppups = PPUPState (ProposedPPUpdates mempty) (ProposedPPUpdates mempty)
+      _ppups = PPUPState (ProposedPPUpdates mempty) (ProposedPPUpdates mempty),
+      _stakeDistro = IStake mempty mempty
```
nc6:
```suggestion
      _stakeDistro = mempty
```

Should use the monoid instance

eras/shelley/test-suite/src/Test/Cardano/Ledger/Shelley/Address/Bootstrap.hs
```diff
@@ -171,7 +173,8 @@ utxoState1 =
     { _utxo = UTxO $ Map.fromList [bobResult, aliceResult],
       _deposited = Coin 0,
       _fees = Coin 10,
-      _ppups = PPUPState (ProposedPPUpdates mempty) (ProposedPPUpdates mempty)
+      _ppups = PPUPState (ProposedPPUpdates mempty) (ProposedPPUpdates mempty),
+      _stakeDistro = IStake mempty mempty
```
nc6, line 177:
```suggestion
      _stakeDistro = mempty
```

eras/shelley/test-suite/src/Test/Cardano/Ledger/Shelley/BenchmarkFunctions.hs
```diff
@@ -154,6 +155,7 @@ initUTxO n =
     (Coin 0)
     (Coin 0)
     def
+    (IStake mempty mempty)
```
nc6:
```suggestion
    mempty
```

eras/shelley/test-suite/src/Test/Cardano/Ledger/Shelley/Examples/Consensus.hs
```diff
@@ -324,7 +324,8 @@ exampleNewEpochState value ppp pp =
                             ],
                       _deposited = Coin 1000,
                       _fees = Coin 1,
-                      _ppups = def
+                      _ppups = def,
+                      _stakeDistro = IStake mempty mempty -- Maybe this needs to agree with _utxo?
```
nc6: Use `smartUtxoState` here?

nc6 (approved):
Left a bunch of comments, which I think it would be good to address. However, since this PR has been open for so long, happy for them to be addressed in a separate PR if that's easier.

eras/shelley-ma/impl/src/Cardano/Ledger/ShelleyMA/Rules/Utxo.hs
TimSheard: done

eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState.hs
TimSheard: It now reads like this.

#-- | Incremental Stake, Stake along with possible missed coins from danging Ptrs. Transactions can use Ptrs to refer to a stake credential in a TxOut. The Ptr does not have to point to anything until the epoch boundary, when we compute rewards and aggregate staking information for ranking. This is unusual but legal. In a non incremental system, we use what ever 'legal' Ptrs exist at the epoch boundary. Here we are computing things incrementally, so we need to remember Ptrs that might point to something by the time the epoch boundary is reached. When the epoch boundary is reached we 'resolve' these pointers, to see if any have become non-dangling since the time they came into the incremental computation.

eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState.hs
TimSheard: It originally appeared only in tests. But is now used in two places were things need to be initialized correctly. So it is usefull here both because it is used, and it provides documentation.

eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState.hs
TimSheard: done

eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState.hs
TimSheard: Actually it is used twice, both in stakeDistr and in incrementalstakeDistr

eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState.hs
TimSheard: done

eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState.hs
TimSheard: it has been fixed. It was caused by a copy-paste problem

eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState.hs
TimSheard: done, I also realized it does not have to return an IncrementalStake, it needs only return the resolved map. So that removes its 'heavy' restriction, and since it is not used elsewhere I removed it from the exports.

eras/shelley/test-suite/src/Test/Cardano/Ledger/Shelley/Address/Bootstrap.hs
TimSheard: good idea.

eras/shelley/test-suite/src/Test/Cardano/Ledger/Shelley/BenchmarkFunctions.hs
TimSheard: done

eras/shelley/test-suite/src/Test/Cardano/Ledger/Shelley/Examples/Consensus.hs
TimSheard: yes!

eras/shelley/test-suite/src/Test/Cardano/Ledger/Shelley/Address/Bootstrap.hs
TimSheard: done

PR #2539, opened 2021-11-05
Miscellaneous fixes

nc6:
Each commit can be reviewed independently.

### DIFF
--- a/libs/cardano-ledger-core/src/Cardano/Ledger/Hashes.hs
+++ b/libs/cardano-ledger-core/src/Cardano/Ledger/Hashes.hs
@@ -20,7 +20,8 @@ module Cardano.Ledger.Hashes
     EraIndependentAuxiliaryData,
     EraIndependentPParamView,
     EraIndependentScriptIntegrity,
-    -- $scriptHash
+
+    -- * Script hashes
     ScriptHash (..),
 
     -- * deprecated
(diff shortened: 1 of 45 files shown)


libs/cardano-ledger-core/src/Cardano/Ledger/Hashes.hs
```diff
@@ -68,7 +68,7 @@ data EraIndependentScriptIntegrity
 
 type EraIndependentWitnessPPData = EraIndependentScriptIntegrity
 
--- $scriptHash
+-- scriptHash
```
lehins: This was meant to be documentation for the header file. See `$scriptHash` anchor in the exports section. I think it would make sense to delete it from there too. A better alternative would be to revert this change and some haddock here.

lehins (approved):
Just one comment, otherwise looks great!

lehins:
I think you also need to update ormolu in github actions https://github.com/input-output-hk/cardano-ledger/blob/4347e54b2827090e898702ccfb93cc731b4a1909/.github/workflows/haskell.yml#L145

libs/cardano-ledger-core/src/Cardano/Ledger/Hashes.hs
nc6: I went to add haddock, but there's really nothing to say, so I've just removed it.

JaredCorduan (approved):
thanks @nc6 !

nc6:
Turns out we don't use the nix specified version in CI, so I've downbumped it to the 3.1.0 release

PR #2540, opened 2021-11-05
Snapshots with array base map

lehins:
Implement Vector based map for snapshots to save memory

### DIFF
--- a/eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/NewEpoch.hs
+++ b/eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/NewEpoch.hs
@@ -22,6 +22,7 @@ where
 
 import Cardano.Ledger.BaseTypes
 import Cardano.Ledger.Coin
+import Cardano.Ledger.Compactible (Compactible (fromCompact))
 import qualified Cardano.Ledger.Core as Core
 import Cardano.Ledger.Credential (Credential)
 import Cardano.Ledger.Era (Crypto, Era)
@@ -38,9 +39,10 @@ import Cardano.Ledger.Slot
 import qualified Cardano.Ledger.Val as Val
 import Control.Provenance (runProvM)
 import Control.State.Transition
+import Data.Compact.VMap as VMap
 import Data.Default.Class (Default, def)
 import qualified Data.Map.Strict as Map
-import Data.Maybe (catMaybes)
+import Data.Ratio
 import Data.Set (Set)
 import GHC.Generics (Generic)
 import GHC.Records
@@ -171,16 +173,21 @@ newEpochTransition = do
 
 calculatePoolDistr :: SnapShot crypto -> PoolDistr crypto
 calculatePoolDistr (SnapShot (Stake stake) delegs poolParams) =
-  let Coin total = Map.foldl' (<>) mempty stake
+  let Coin total = VMap.foldMap fromCompact stake
+      -- total could be zero (in particular when shrinking)
+      nonZeroTotal = if total == 0 then 1 else total
       sd =
         Map.fromListWith (+) $
-          catMaybes
-            [ (,fromIntegral c / fromIntegral (if total == 0 then 1 else total))
-                <$> Map.lookup hk delegs -- TODO mgudemann total could be zero (in
-                -- particular when shrinking)
-              | (hk, Coin c) <- Map.toList stake
-            ]
-   in PoolDistr $ Map.intersectionWith IndividualPoolStake sd (Map.map _poolVrf poolParams)
+          [ (d, c % nonZeroTotal)
+            | (hk, compactCoin) <- VMap.toAscList stake,
+              let Coin c = fromCompact compactCoin,
+              Just d <- [VMap.lookup hk delegs]
+          ]
+   in PoolDistr $
+        Map.intersectionWith
+          IndividualPoolStake
+          sd
+          (toMap (VMap.map _poolVrf poolParams))
 
 instance
   ( UsesTxOut era,
--- a/eras/shelley/test-suite/test/Test/Cardano/Ledger/Shelley/Examples/TwoPools.hs
+++ b/eras/shelley/test-suite/test/Test/Cardano/Ledger/Shelley/Examples/TwoPools.hs
@@ -3,6 +3,7 @@
 {-# LANGUAGE DataKinds #-}
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE GADTs #-}
+{-# LANGUAGE OverloadedLists #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE TypeApplications #-}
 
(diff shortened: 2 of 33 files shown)


eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/NewEpoch.hs
```diff
@@ -171,16 +172,20 @@ newEpochTransition = do
 
 calculatePoolDistr :: SnapShot crypto -> PoolDistr crypto
 calculatePoolDistr (SnapShot (Stake stake) delegs poolParams) =
-  let Coin total = Map.foldl' (<>) mempty stake
+  let Coin total = VMap.fold stake
+      -- TODO mgudemann total could be zero (in particular when shrinking)
```
JaredCorduan:
```suggestion
      -- total could be zero (in particular when shrinking)
```

eras/shelley/test-suite/test/Test/Cardano/Ledger/Shelley/Examples/TwoPools.hs
```diff
@@ -482,13 +480,13 @@ blockEx5 =
     (mkOCert Cast.alicePoolKeys 0 (KESPeriod 10))
 
 activeStakeEx5 :: Integer
-activeStakeEx5 = sum $ unCoin <$> [aliceCoinEx1, bobInitCoin, carlInitCoin]
+activeStakeEx5 = sum $ map unCoin [aliceCoinEx1, bobInitCoin, carlInitCoin]
```
JaredCorduan, line 483: just curious, is this just stylistic, making it read easier?

eras/shelley/test-suite/test/Test/Cardano/Ledger/Shelley/Serialisation/Golden/Encoding.hs
```diff
@@ -426,7 +427,7 @@ tests =
     "CBOR Serialization Tests (Encoding)"
     [ checkEncodingCBOR
         "list"
-        [1 :: Integer]
+        ([1] :: [Integer])
```
JaredCorduan, line 430: just style?

libs/cardano-ledger-core/src/Cardano/Ledger/Coin.hs
```diff
@@ -86,9 +88,23 @@ integerToWord64 c
...
   fromCBOR = CompactCoin <$> fromCBOR
+
+instance Storable Coin where
+  sizeOf _ = sizeOf (undefined :: Word64)
+  {-# INLINE sizeOf #-}
+  alignment _ = alignment (undefined :: Word64)
+  {-# INLINE alignment #-}
+  poke ptr (Coin c) =
+    case integerToWord64 c of
+      Nothing -> error $ "Impossible Coin overflow: " ++ show c
```
JaredCorduan: ah, this is the unfortunate part you mentioned to me earlier. I don't see a way around it, unfortunately.

libs/ledger-state/bench/Memory.hs
```diff
@@ -48,33 +48,36 @@ main = do
...
+--   io "IntMap (KeyMap TxId ())" (loadDbUTxO txIxSharingKeyMap_) dbFp
+--   io "KeyMap TxId (IntMap TxId ())" (loadDbUTxO txIdSharingKeyMap_) dbFp
+--   io "IntMap (Map TxId ())" (loadDbUTxO txIxSharing_) dbFp
+--   io "Map TxIn ()" (loadDbUTxO noSharing_) dbFp
+-- wgroup "LedgerState" $ do
+--   wgroup "UTxO (Share DState)" $ do
+--     io "IntMap (KeyMap TxId TxOut)" getLedgerStateDStateTxIxSharingKeyMap dbFp
+--     io "KeyMap TxId (IntMap TxOut)" getLedgerStateDStateTxIdSharingKeyMap dbFp
+--     io "IntMap (Map TxId TxOut)" getLedgerStateDStateTxIxSharing dbFp
+--     io "Map TxIn TxOut" getLedgerStateDStateSharing dbFp
```
JaredCorduan, line 80: are these comments helpful to have around? normally I like to not keep such things.

libs/small-steps/src/Data/Coders.hs
```diff
@@ -273,6 +277,18 @@ wrapCBORArray len contents =
...
 -- cannonical improvements easy to use.
 
+encodeVMap ::
+  (VMap.Vector vk k, VMap.Vector vv v) =>
+  (k -> Encoding) ->
+  (v -> Encoding) ->
+  VMap.VMap vk vv k v ->
+  Encoding
+encodeVMap encodeKey encodeValue vmap =
+  --encodeMap encodeKey encodeValue (toMap vec)
```
JaredCorduan: remove?

JaredCorduan (approved):
Looks good to me! I only had a few trivial suggestions and questions.

eras/shelley/test-suite/test/Test/Cardano/Ledger/Shelley/Examples/TwoPools.hs
lehins: This is a very good question. It is actually not due to a stylistic choice, but rather because I turned on `OverloadedLists` extension and using `map` here instead of `fmap` will restrict its type to `[a]`, rather than `IsList (t a) => t a`. 

So instead of manually restricting the type:

```haskell
activeStakeEx5 = sum $ unCoin <$> ([aliceCoinEx1, bobInitCoin, carlInitCoin] :: [Coin])
```

I chose to use `map` that will do that for me :)

eras/shelley/test-suite/test/Test/Cardano/Ledger/Shelley/Serialisation/Golden/Encoding.hs
lehins: Again, `OverloadedLists`

libs/ledger-state/bench/Memory.hs
lehins: These are still useful because these aren't really comments. `weigh` package unfortunately does not have a command line interface that allows selection of which benchmarks to run . Which means the only way right now to select benchmarks is to comment the ones that you don't want to run. I did a simple cli interface that takes some paths, which can be expanded to perform benchmark selection, but the set of benchmarks hasn't really solidified for me to do that. Let's keep that as comment for now.

libs/small-steps/src/Data/Coders.hs
lehins: Removed

eras/shelley/test-suite/test/Test/Cardano/Ledger/Shelley/Examples/TwoPools.hs
JaredCorduan: ah, thanks for the explanation!

libs/ledger-state/bench/Memory.hs
JaredCorduan: works for me!

PR #2541, opened 2021-11-08
fixed typo in variable length encoding description

JaredCorduan:


eras/shelley/formal-spec/address-binary.tex
```diff
@@ -105,4 +105,4 @@ \subsection{Payloads}
 
 The variable-length encoding used in pointers addresses is the base-128 representation
 of the number, with the the most significant bit of each byte indicating continuation.
```
Craphtex:
```suggestion
of the number, with the most significant bit of each byte indicating continuation.
```
This is probably the best opportunity to fix this minor grammar error too.

As this is formally called [VLQ](https://en.wikipedia.org/wiki/Variable-length_quantity), it might be worth adding a note about that too just to be extra clear?

JaredCorduan: thanks @craphtex! I think making mention of the name VLQ sounds helpful.

lehins (approved):

PR #2543, opened 2021-11-10
Alonzo spec fixes

WhatisRT:


JaredCorduan (approved):
looks fantastic, thanks @WhatisRT !

JaredCorduan:
I am going to re-run CI, we hit :slot_machine: a failure in the "We are elected as leader proportional to our stake" test, which does still sometimes fail and is only a worry if it starts failing often. (and is obviously totally unrelated to this latex PR).

PR #2548, opened 2021-11-12
Ts intern stake credential

TimSheard:
Serialization now interns (Credential 'Staking crypto) objects

### DIFF
--- a/libs/small-steps/src/Data/Sharing.hs
+++ b/libs/small-steps/src/Data/Sharing.hs
@@ -18,6 +18,7 @@ module Data.Sharing
     Interned (..),
     Internable (..),
     fromCBOR',
+    decodeAndIntern,
     -- Extra for experts
     Arity (..),
     Nat (..),
@@ -189,3 +190,9 @@ decodeMapAsBimap maps = decodeMapSkel (biMapFromAscDistinctList . List.map (\(k,
 instance (Ord k, FromCBOR k, FromCBOR t, VMap.Vector v2 t) => FromCBORShare (VMap VB v2 k t) k 'N1 where
   fromShare A1 maps = decodeVMap (decodeAndIntern maps) fromCBOR
   fromSharePlus A1 maps = do mp <- fromShare A1 maps; pure (mp, include mp maps)
+
+instance (Ord k, Ord t, FromCBOR k, FromCBOR t, VMap.Vector v2 t) => FromCBORShare (VMap VB v2 k t) (k, t) 'N2 where
+  fromShare A2 (km, tm) = decodeVMap (decodeAndIntern km) (decodeAndIntern tm)
+  fromSharePlus A2 (kms, tms) = do
+    mp <- decodeVMap (decodeAndIntern kms) (decodeAndIntern tms)
+    pure (mp, (include mp kms, tms))
--- a/libs/small-steps/src/Data/Sharing.hs
+++ b/libs/small-steps/src/Data/Sharing.hs
@@ -1,0 +1,191 @@
+{-# LANGUAGE BangPatterns #-}
+{-# LANGUAGE DataKinds #-}
+{-# LANGUAGE FlexibleInstances #-}
+{-# LANGUAGE GADTs #-}
+{-# LANGUAGE KindSignatures #-}
+{-# LANGUAGE LambdaCase #-}
+{-# LANGUAGE MultiParamTypeClasses #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TupleSections #-}
+{-# LANGUAGE TypeApplications #-}
+
+module Data.Sharing
+  ( emptyA,
+    include,
+    internI,
+    internUsingMap,
+    FromCBORShare (..),
+    Interned (..),
+    Internable (..),
+    fromCBOR',
+    -- Extra for experts
+    Arity (..),
+    Nat (..),
+    IList,
+  )
+where
+
+import Cardano.Binary (Decoder, FromCBOR (..), decodeListLen, decodeMapSkel, dropMap)
+import Control.Iterate.SetAlgebra (BiMap, biMapFromAscDistinctList)
+import Control.Monad (void)
+import Control.SetAlgebra (backwards)
+import Data.Coders (decodeMap, decodeVMap, invalidKey)
+import Data.Compact.VMap (VB, VMap)
+import qualified Data.Compact.VMap as VMap
+import qualified Data.List as List
+import qualified Data.Map.Strict as Map
+import Data.Map.Strict.Internal
+
+-- =======================================
+
+class Internable t k where
+  makeIntern :: t -> Interned k
+
+data Interned k = Interned
+  { -- | Partially applied intern function
+    internFunc :: k -> Maybe k,
+    -- | Corresponds to the size of a map
+    internWeight :: Int
+  }
+
+instance Ord k => Internable (Map k v) k where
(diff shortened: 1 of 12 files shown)


libs/small-steps/src/Data/Sharing.hs
```diff
@@ -0,0 +1,184 @@
...
+import qualified Data.Compact.VMap as VMap
+import qualified Data.List as List
+import qualified Data.Map.Strict as Map
+import Data.Map.Strict.Internal
+
+-- =======================================
+
+class Intern map where
+  findKey :: Ord k => k -> map k v -> Maybe k
+  internSize :: map k v -> Int
```
lehins:
```suggestion
class Internable t k | t -> k where
  mkIntern :: t ->  Inter k
```

libs/small-steps/src/Data/Sharing.hs
```diff
@@ -0,0 +1,184 @@
...
+-- | include a new map in the IList, the idea is to keep the IList sorted by map size, so
+--   the largest maps come first. This should decrease the number of maps searched when
+--   looking to intern a 'k' object.
+include :: Intern map => map k a -> IList k -> IList k
+include m INil = ICons m INil
+include m (x@(ICons n ns)) =
+  case compare (internSize m) (internSize n) of
+    EQ -> ICons n (include m ns) -- When the size is the same, keep the older map at the front
+    GT -> ICons m x
+    LT -> ICons n (include m ns)
```
lehins: A bit of simplification
```suggestion
include m x@(ICons n ns)
  | internSize m > internSize n = ICons m x
  | otherwise = ICons n (include m ns)
```

libs/small-steps/src/Data/Sharing.hs
```diff
@@ -0,0 +1,184 @@
...
+
+-- | Like (FromCBOR t), but able to perform sharing on objects of type k, because of the Arity parameter
+--   k can be x or (x,y) or (x,y,z). That means we can carry around up to 3 tables to use to intern 3 differenet types.
+--   An Instance can define either method 'fromShare' or 'fromSharePlus'. The latter accumulates new (Ilist tables).
+class FromCBORShare t k n where
+  fromShare :: Arity n k table -> table -> Decoder s t
+  fromShare arity maps = fmap fst (fromSharePlus arity maps)
+
+  fromSharePlus :: Arity n k table -> table -> Decoder s (t, table)
+  fromSharePlus arity maps = do x <- fromShare arity maps; pure (x, maps)
```
lehins: I really don't like the `;` It is a remnant from imperative language that makes the code less readable. One  way or another there is no need for `do` here.
```suggestion
  fromSharePlus arity maps = (,) maps <$> fromShare arity maps
```

libs/small-steps/src/Data/Sharing.hs
```diff
@@ -0,0 +1,184 @@
...
+
+-- | Decode 't' then look it up in the intern table (IList) 'ts'
+decodeAndIntern :: (Ord t, FromCBOR t) => (IList t) -> Decoder s t
+decodeAndIntern ts = do t <- fromCBOR; pure (internI ts t)
```
lehins:
```suggestion
decodeAndIntern ts = internI ts <$> fromCBOR
```

libs/small-steps/src/Data/Sharing.hs
```diff
@@ -0,0 +1,184 @@
...
+-- ==================================
+
+internUsingMap :: Ord k => k -> Map k a -> k
+internUsingMap = go
+  where
+    go !k Tip = k
+    go !k (Bin _ kx _ l r) = case compare k kx of
+      LT -> go k l
+      GT -> go k r
+      EQ -> kx
```
lehins, line 91: This function doesn't seem to be use, so we can remove it. If we do need it we can create a helper `fromMaybe k . findKey k` instead

libs/small-steps/src/Data/Sharing.hs
```diff
@@ -0,0 +1,184 @@
...
+    pure (mp, include mp ks)
+
+-- | Decode a Map where there is an intern tables for just the range 'v'
+instance (Ord k, Ord v, FromCBOR k, FromCBOR v) => FromCBORShare (Map k v) v 'N1 where
+  fromShare A1 vs = decodeMap fromCBOR (decodeAndIntern vs)
+
+-- Can be used inside FromCBOR instances, when you only have a (FromShare k t) instance
+-- use:  fromCBOR' A1   or fromCBOR' A2      or    fromCBOR' A3
+fromCBOR' :: forall tuple t n table s. FromCBORShare t tuple n => Arity n tuple table -> Decoder s t
+fromCBOR' arity = fromShare arity (emptyA arity)
```
lehins, line 150: Maybe a more explicit name `fromCBORNoShare` or something that carries a hint that no sharing will happen

libs/small-steps/src/Data/Sharing.hs
```diff
@@ -0,0 +1,184 @@
...
+      -- Previous encoding of 'BiMap' encoded both the forward and reverse
+      -- directions. In this case we skip the reverse encoding. Note that,
+      -- further, the reverse encoding was from 'b' to 'a', not the current 'b'
+      -- to 'Set a', and hence the dropper reflects that.
+      2 -> do
+        !x <- decodeMapAsBimap maps
+        dropMap (void $ fromCBOR @b) (void $ fromCBOR @a)
+        return x
+      k -> invalidKey (fromIntegral k)
+  fromSharePlus A1 maps = do bimap <- fromShare A1 maps; pure (bimap, include (backwards bimap) maps)
```
lehins: The line is already pretty long
```suggestion
  fromSharePlus A1 maps = do
    bimap <- fromShare A1 maps
    pure (bimap, include (backwards bimap) maps)
```

libs/small-steps/src/Data/Sharing.hs
```diff
@@ -0,0 +1,184 @@
...
+  A1 :: Arity 'N1 x (IList x)
+  A2 :: Arity 'N2 (x, y) (IList x, IList y)
+  A3 :: Arity 'N3 (x, y, z) (IList x, IList y, IList z)
+
+data Nat = N1 | N2 | N3
+
+-- | Like (FromCBOR t), but able to perform sharing on objects of type k, because of the Arity parameter
+--   k can be x or (x,y) or (x,y,z). That means we can carry around up to 3 tables to use to intern 3 differenet types.
+--   An Instance can define either method 'fromShare' or 'fromSharePlus'. The latter accumulates new (Ilist tables).
+class FromCBORShare t k n where
```
lehins, line 115: So the compiler can protect us from divergent instances
```suggestion
class FromCBORShare t k n where
  {-# MINIMAL (fromShare|fromSharePlus) #-}
```

lehins (changes requested):
The whole arity approach seems a little bit complex and not scalable, I have some ideas for a an alternative interface. Instead of trying to explain I it created a PR: https://github.com/input-output-hk/cardano-ledger/pull/2557 Let me know what you think.

libs/small-steps/src/Data/Sharing.hs
TimSheard: This is the function one will use, when you want to share when you are not deserializing

libs/small-steps/src/Data/Sharing.hs
TimSheard: done

JaredCorduan:
closed in favor of #2557

PR #2549, opened 2021-11-12
Filter before returning rewards in RewardEvent.

nc6:
This does currently call `filterEvent` twice, but this should have no
effect, and saves introducing another function.


JaredCorduan (approved):
LGTM!

PR #2552, opened 2021-11-15
Added property tests for the KeyMap data structure.

TimSheard:
Adds property tests for KeyMap. Also adds some tools for computing statistics about a KeyMap.


libs/compact-map/test/Test/Compact/KeyMap.hs
```diff
@@ -25,9 +64,409 @@ instance Arbitrary a => Arbitrary (KeyMap a) where
...
+    statSum :: n,
+    statMax :: (Maybe n),
+    statMin :: (Maybe n)
+  }
+
+liftM :: (t -> t -> t) -> Maybe t -> Maybe t -> Maybe t
+liftM f (Just x) (Just y) = Just (f x y)
+liftM _ Nothing (Just y) = Just y
+liftM _ (Just x) Nothing = Just x
+liftM _ Nothing Nothing = Nothing
```
lehins: Let's rename it to something else maybe `liftMaybes` or something. `liftM` is a historic `fmap` and taking over known names is not a good idea for the future readers of the code. Also compacting it a bit maybe?

```suggestion
liftMaybes :: (t -> t -> t) -> Maybe t -> Maybe t -> Maybe t
liftMaybes f mx my = (f <$> mx <*> my) <|> my <|> mx
```

libs/compact-map/test/Test/Compact/KeyMap.hs
```diff
@@ -25,9 +64,409 @@ instance Arbitrary a => Arbitrary (KeyMap a) where
...
+makeKeys :: Int -> Int -> [Key]
+makeKeys seed cnt = loop (mkStdGen seed) cnt []
+  where
+    loop g i !ans =
+      if i <= 0
+        then ans
+        else case genKey g of (!key, !g2) -> loop g2 (i - 1) (key : ans)
+
+randKM :: Int -> Int -> KeyMap Int
+randKM seed n = fromList (zip (makeKeys seed n) [0 :: Int ..])
```
lehins: Dead code
```suggestion
```

libs/compact-map/test/Test/Compact/KeyMap.hs
```diff
@@ -1,19 +1,58 @@
...
 
+-- ==============================================================
+
+genKey :: RandomGen b => b -> (Key, b)
+genKey g = (Key w0 w1 w2 w3, g4)
+  where
+    (w0, g1) = genWord64 g
+    (w1, g2) = genWord64 g1
+    (w2, g3) = genWord64 g2
+    (w3, g4) = genWord64 g3
```
lehins: Lets use the new interface from `System.Random.Stateful`
```suggestion
instance Uniform Key where
  uniformM g = Key <$> uniformM g <*> uniformM g <*> uniformM g <*> uniformM g
```

libs/compact-map/test/Test/Compact/KeyMap.hs
```diff
@@ -1,19 +1,58 @@
...
+
 instance Arbitrary Key where
   arbitrary =
     oneof
       [ Key <$> arbitrary <*> arbitrary <*> arbitrary <*> arbitrary,
         Key <$> chooseAny <*> chooseAny <*> chooseAny <*> chooseAny
       ]
 
+instance HM.Keyed Key where
```
lehins, line 48: I think it will be good to move this instance right next to `Keyed` class definition, no point in defining an orphan here.

libs/compact-map/test/Test/Compact/KeyMap.hs
```diff
@@ -25,9 +64,409 @@ instance Arbitrary a => Arbitrary (KeyMap a) where
...
+    pathArr = fromlist pathsbulk
+
+-- ==================================================================
+-- Data structure for storing statictic like values about KeyMaps
+
+data Stat n = Stat
+  { statCount :: n,
+    statSum :: n,
+    statMax :: (Maybe n),
+    statMin :: (Maybe n)
```
lehins: Redundant parenthesis"
```suggestion
    statMax :: Maybe n,
    statMin :: Maybe n
```

libs/compact-map/test/Test/Compact/KeyMap.hs
```diff
@@ -25,9 +64,409 @@ instance Arbitrary a => Arbitrary (KeyMap a) where
...
+
+instance (Integral n, Show n) => Show (Stat n) where
+  show (Stat c s mx mn) =
+    "{count= " ++ show c ++ ", sum=" ++ show s ++ ", max=" ++ show mx
+      ++ ", min="
+      ++ show mn
+      ++ (if c == 0 then "}" else ", avg=" ++ show (div s c) ++ "}")
+
+add :: (Num n, Ord n) => n -> Stat n -> Stat n
+add n stat = (Stat 1 n (Just n) (Just n)) <> stat
```
lehins: redundant parenthesis:
```suggestion
add n stat = Stat 1 n (Just n) (Just n) <> stat
```

libs/compact-map/test/Test/Compact/KeyMap.hs
```diff
@@ -25,9 +64,409 @@ instance Arbitrary a => Arbitrary (KeyMap a) where
...
+-- ============================================================
+
+alltests :: TestTree
+alltests =
+  testGroup
+    "KeyMap tests"
+    [keyMapTests, keyMapEquivDataMap]
+
+test :: IO ()
+test = defaultMain alltests
```
lehins: Dead code
```suggestion
```

libs/compact-map/test/Test/Compact/KeyMap.hs
```diff
@@ -25,9 +64,409 @@ instance Arbitrary a => Arbitrary (KeyMap a) where
...
+      testPropertyN 1000 "intersectWith is commutative" $ commutes @Key @Int (HM.intersectionWith (+)), -- (unionwith f) is commutative if 'f' is commutative
+      testProperty "ascending fold == descending fold with commutative operator" ascFoldDescFold,
+      testProperty "lookupMin finds the smallest key" (minKey @Int),
+      testProperty "lookupMax finds the largest key" (maxKey @Int),
+      testProperty "(mapWithKey f) applies 'f' to every value" mapWorks,
+      testProperty "foldOverIntersection folds over the intersection" foldintersect,
+      testProperty "restrictKeys and withoutKeys partition a KeyMap" withoutRestrict,
+      testPropertyN 500 "splitLookup pieces add to the whole" splitwhole,
+      testProperty "bulk insert == incremental insert" (\n -> testbulk n 1000),
+      testCaseInfo "statistics gathering" (testStatistics 1000000)
```
lehins: This is not really a test. It is a function that can be used for debugging, but because it doesn't test anything there is no point in running every time we run the test suite and on every CI build.
```suggestion
```

libs/compact-map/test/Test/Compact/KeyMap.hs
```diff
@@ -25,9 +64,409 @@ instance Arbitrary a => Arbitrary (KeyMap a) where
...
+  k ->
+  t k a ->
+  Property
+lookupdelete look del k km = look k (del k km) === Nothing
+
+assoc :: forall k v t. (Eq (t k v), Show (t k v)) => (t k v -> t k v -> t k v) -> t k v -> t k v -> t k v -> Property
+assoc oper x y z = oper (oper x y) z === oper x (oper y z)
+
+commutes :: forall k v t. (Eq (t k v), Show (t k v)) => (t k v -> t k v -> t k v) -> t k v -> t k v -> Property
+commutes oper x y = (oper x y) === (oper y x)
```
lehins: redundant parenthesis
```suggestion
commutes oper x y = oper x y === oper y x
```

libs/compact-map/src/Data/Compact/SmallArray.hs
```diff
@@ -0,0 +1,65 @@
...
+mindex = mboundsCheck Small.readSmallArray
+
+msize = Small.sizeofSmallMutableArray
+
+mnew size = Small.newSmallArray size (error "uninitialized index, allocated by 'mnew', is referenced")
+
+mfreeze = Small.unsafeFreezeSmallArray
+
+mwrite arr i a =
+  if i >= 0 && i < (msize arr)
```
lehins: Redundant parenthesis
```suggestion
  if i >= 0 && i < msize arr
```

libs/compact-map/src/Data/Compact/SmallArray.hs
```diff
@@ -0,0 +1,65 @@
...
+    then Small.writeSmallArray arr i a
+    else error ("mwrite error, " ++ show i ++ ", not in bounds (0.." ++ show (msize arr -1) ++ ").")
+
+mcopy = Small.copySmallArray
+
+mboundsCheck :: (MArray s a -> Int -> p) -> MArray s a -> Int -> p
+mboundsCheck indexf arr i | i >= 0 && i < msize arr = indexf arr i
+mboundsCheck _ arr i = error ("mboundscheck error, " ++ show i ++ ", not in bounds (0.." ++ show (msize arr -1) ++ ").")
```
lehins:
```suggestion
mboundsCheck _ arr i = error ("mboundscheck error, " ++ show i ++ ", not in bounds (0.." ++ show (msize arr - 1) ++ ").")
```

libs/compact-map/src/Data/Compact/SmallArray.hs
```diff
@@ -0,0 +1,65 @@
...
+
+boundsCheck :: (PArray a -> Int -> p) -> PArray a -> Int -> p
+boundsCheck indexf arr i | i >= 0 && i < isize arr = indexf arr i
+boundsCheck _ arr i = error ("boundscheck error, " ++ show i ++ ", not in bounds (0.." ++ show (isize arr -1) ++ ").")
```
lehins:
```suggestion
boundsCheck _ arr i = error ("boundscheck error, " ++ show i ++ ", not in bounds (0.." ++ show (isize arr - 1) ++ ").")
```

libs/compact-map/src/Data/Compact/SmallArray.hs
```diff
@@ -0,0 +1,65 @@
...
+isize :: PArray a -> Int
+fromlist :: [a] -> PArray a
+tolist :: PArray a -> [a]
+index = boundsCheck Small.indexSmallArray
+
+isize = Small.sizeofSmallArray
+
+fromlist = Small.smallArrayFromList
+
+tolist arr = foldr (:) [] arr
```
lehins: `SmallArray` has `Foldable` instance (this way it will work well with list fusion):
```suggestion
tolist = Data.Foldable.toList
```

libs/compact-map/src/Data/Compact/KeyMap.hs
```diff
@@ -93,6 +106,11 @@ data Key
       {-# UNPACK #-} !Word64
   deriving (Eq, Ord, Show, NFData, Generic)
 
+{-
+instance Show Key where
+  show (Key x0 _ _ _) = show x0
+-}
+
```
lehins: Dead code
```suggestion
```

libs/compact-map/src/Data/Compact/KeyMap.hs
```diff
@@ -278,9 +293,31 @@ toList km = foldWithDescKey accum [] km
 -- =================================================================
 -- Deletion
 
+delete2 :: Path -> Key -> KeyMap v -> (KeyMap v -> KeyMap v) -> KeyMap v
+delete2 path key km continue = case3 (continue Empty) leafF arrayF km
+  where
+    leafF k2 _ = if key == k2 then (continue Empty) else (continue km)
```
lehins: Redundant parenthesis
```suggestion
    leafF k2 _ = if key == k2 then continue Empty else continue km
```

libs/compact-map/src/Data/Compact/KeyMap.hs
```diff
@@ -278,9 +293,31 @@ toList km = foldWithDescKey accum [] km
...
+                else continue km
+
+delete :: Key -> KeyMap v -> KeyMap v
+delete key km = delete2 (keyPath key) key km id
+
+{-
+
 -- | Delete the Key encoded in the Path from the KeyMap
 delete' :: Path -> Key -> KeyMap v -> KeyMap v
-delete' [] _key hm = hm -- Removing a bogus key, leaves 'hm' unchanged
```
lehins, line 283: Looks like dead code. Do we have a reason to keep it around?

libs/compact-map/src/Data/Compact/KeyMap.hs
```diff
@@ -38,11 +46,16 @@ import Data.Word (Word64)
...
-import System.Random (RandomGen, genWord64, mkStdGen)
 
-type PArray = Small.SmallArray
+-- =============================
+
+-- | binary encoding of 'n', LEAST significant bit on the front of the list.
+binary :: Integral n => n -> [n]
+binary 0 = []
+binary 1 = [(1)]
+binary n = (mod n 2) : binary (div n 2)
```
lehins: Redundant parenthesis
```suggestion
binary 1 = [1]
binary n = mod n 2 : binary (div n 2)
```

libs/compact-map/src/Data/Compact/KeyMap.hs
```diff
@@ -502,218 +527,307 @@ array1 x = fst (withMutArray 1 (\marr -> mwrite marr 0 x))
...
+([14,13,49, 4, 7,37, 4,23, 5,28], (L 16388898632001935134 6))
+([14,43,28,10,55,12,51,63,56,24], (L 16923449273545098794 7))
+
+For each column we break the code into groups where the Segment matches
+on that column. Above we have grouped the 6's, 9's and 14's together by column 1
+-}
+
+-- | Make a (KeyMap v) out of the input. Works by  focusing on a particular range of rows ('lo' .. 'hi')
+--   It calls it self recursively, by chooing a smaller range, and increasing the column number 'n' by 1.
+bulkInsert :: Int -> PArray (Path, KeyMap v) -> Int -> Int -> KeyMap v
```
lehins: From what I understand bulk insertion is not yet complete, right? I think it would be really good to extract anything related to bulk insertion into a separate branch and exclude it from this PR. I suggest this for a few important reasons:

1. Someone might think it is working correctly and use it, which could cause problems
2. Whenever it is complete we can actually review it in isolation.
3. This PR is already big enough
4. If it never gets finished, we will not end up dragging around code that no one can use.

lehins (changes requested):
I've looked through a portion of this PR, made few minor suggestion, but haven't finished it yet. Over all it looks great. I am requesting some changes before I review it again just because this PR includes much code that is not ready for master yet, namely bulk insertion. I left a comment about it in the review.

Problem with ormolu is that we now use newer version on CI, make sure you have `ormolu-0.3.1.0` installed. I think doing `cabal install ormolu` should do the trick

libs/compact-map/test/Test/Compact/KeyMap.hs
```diff
@@ -25,9 +64,409 @@ instance Arbitrary a => Arbitrary (KeyMap a) where
...
+-- ====================================================
+-- Generating random KeyMaps
+
+makeKeys :: Int -> Int -> [Key]
+makeKeys seed cnt = loop (mkStdGen seed) cnt []
+  where
+    loop g i !ans =
+      if i <= 0
+        then ans
+        else case genKey g of (!key, !g2) -> loop g2 (i - 1) (key : ans)
```
lehins: With `Uniform` instance this becomes as easy as:
```suggestion
makeKeys :: Int -> Int -> [Key]
makeKeys seed cnt = runStateGen_ (mkStdGen seed) (uniformListM cnt)
```

libs/compact-map/test/Test/Compact/KeyMap.hs
TimSheard: This would have to go in the file KeyMap.hs where the type Key is defined. Best to keep the Arbitrary instances in a test file. Arbitrary instances are almost always orphans.

lehins: Yep, that's exactly what I meant.

libs/compact-map/test/Test/Compact/KeyMap.hs
TimSheard: I made it into a function that tests if keys are uniformly distributed. It take about 0.2 seconds to run, and tests something important.

libs/compact-map/test/Test/Compact/KeyMap.hs
TimSheard: done

libs/compact-map/src/Data/Compact/KeyMap.hs
TimSheard: Bulk insertion is complete. There are tests that show that it gives the same result as sequential insertion, and that it uses less time and space.  It just could use even less time and space by using a in place sort.

TimSheard: Nonetheless, I have removed bulk insert.

libs/compact-map/src/Data/Compact/KeyMap.hs
```diff
@@ -24,11 +24,19 @@ import Data.Bits
     (.&.),
     (.|.),
   )
-import Data.Compact.Class
```
lehins, line 27: We can remove the `Data/Compact/Class.hs` module  from the repo now, right?

TimSheard: yes

TimSheard:
Rebased on master, and all conversations resolved

libs/compact-map/test/Test/Compact/KeyMap.hs
```diff
@@ -1,19 +1,62 @@
...
 
+-- ==============================================================
+
+genKey :: RandomGen b => b -> (Key, b)
+genKey g = (Key w0 w1 w2 w3, g4)
+  where
+    (w0, g1) = genWord64 g
+    (w1, g2) = genWord64 g1
+    (w2, g3) = genWord64 g2
+    (w3, g4) = genWord64 g3
```
lehins: No longer used.
```suggestion
```

libs/compact-map/src/Data/Compact/KeyMap.hs
```diff
@@ -321,19 +356,54 @@ delete k hm = delete' (keyPath k) k hm
...
--- Float Empty's up over BitmpIndexed, Note that if the size of the arr
--- becomes 2, then rebuild with Two rather than BitmapIndexed
-bitmapE :: Bitmap -> PArray (KeyMap v) -> KeyMap v
-bitmapE bmap arr | isize arr == 2 = twoE bmap (index arr 0) (index arr 1)
-bitmapE bmap arr = bitmapIndexedOrFull bmap arr
+-- | Create a 'BitmapIndexed' or 'Full' or 'One' or 'Two' node depending on the size of 'arr'
+--   and dropping all Empty nodes.  Use this only where things can become empty (delete, intersect, etc)
+dropEmpty :: Bitmap -> PArray (KeyMap v) -> KeyMap v
+dropEmpty _ arr | isize arr == 0 = Empty
+dropEmpty b arr | isize arr == 1 = oneE (head (bitmapToList b)) (index arr 0)
```
lehins: Partial functions like `head` should never be used, even if they look impossible. It is not obvious that `b /= 0`.

```haskell
> head $ bitmapToList (0 :: Int)
*** Exception: Prelude.head: empty list
```

 It is very hard to find the source of these sort of errors, because all you get is this when it happens: `*** Exception: Prelude.head: empty list`, without any information where in the codebase it originated from.

libs/compact-map/src/Data/Compact/KeyMap.hs
```diff
@@ -502,218 +522,229 @@ array1 x = fst (withMutArray 1 (\marr -> mwrite marr 0 x))
...
+-- | The view of the KeyMap of the smallestKey and its value, and the map that results from removing that Leaf.
+minViewWithKeyHelp :: KeyMap a -> (KeyMap a -> KeyMap a) -> Maybe ((Key, a), KeyMap a)
+minViewWithKeyHelp x continue = case3 Nothing leafF arrayF x
+  where
+    leafF k v = Just ((k, v), continue Empty)
+    arrayF bm arr = minViewWithKeyHelp (index arr slicepoint) (continue . largeSide i bmMinusi slicepoint arr)
+      where
+        slicepoint = 0
+        seglist = bitmapToList bm
+        i = head seglist
```
lehins: Same thing here.

IMHO `head` should be removed from `base`, so no one could use it.

libs/compact-map/src/Data/Compact/KeyMap.hs
```diff
@@ -737,14 +768,32 @@ sparseIndex b m = popCount (b .&. (m - 1))
...
-bitmapIndexedOrFull :: Bitmap -> PArray (KeyMap v) -> KeyMap v
-bitmapIndexedOrFull _ arr | isize arr == 0 = Empty
-bitmapIndexedOrFull b arr | isize arr == 1 = One (head (bitmapToList b)) (index arr 0)
-bitmapIndexedOrFull b arr | isize arr == 2 = Two b (index arr 0) (index arr 1)
-bitmapIndexedOrFull b arr
+buildKeyMap :: Bitmap -> PArray (KeyMap v) -> KeyMap v
+buildKeyMap _ arr | isize arr == 0 = Empty
+buildKeyMap b arr | isize arr == 1 = case (index arr 0) of
+  x@(Leaf _ _) -> x
+  x -> One (head (bitmapToList b)) x
```
lehins: One more occurrence of `head`.

lehins (changes requested):
It should be possible to rewrite all those cases that use `head` with something like [`listToMaybe`](https://downloads.haskell.org/~ghc/9.0.1/docs/html/libraries/base-4.15.0.0/Data-Maybe.html#v:listToMaybe) or direct pattern matching. Worst case scenario we can use `error`, as it provides location of the error and information about why it should have been impossible.

libs/compact-map/src/Data/Compact/KeyMap.hs
TimSheard: The KeyMap data structure maintains an invariant that the number of bits set in a BitMap is the same as the size of the Array. I removed the calls to 'head' and replaced them with a case analysis that calls error with a message that locates the call if this invariant is violated. I also make the invariant explicit in a few cases where it is maintained.

libs/compact-map/src/Data/Compact/KeyMap.hs
TimSheard: handled as described above, there was also a 3rd case which was handed the same way.

lehins (approved):
Beautiful!

libs/compact-map/src/Data/Compact/KeyMap.hs
TimSheard: done

libs/compact-map/src/Data/Compact/KeyMap.hs
TimSheard: It has been removed.

libs/compact-map/test/Test/Compact/KeyMap.hs
TimSheard: done

libs/compact-map/src/Data/Compact/KeyMap.hs
TimSheard: done

PR #2553, opened 2021-11-16
Improve tx out compacting

lehins:
After some investigation I found out why we didn't get as much as expected on the compacting `TxOut`. It turns out constraints on data types do not come for free. Dropping those constraints in favor of couple impossible error calls we regain the expected savings on the TxOut:
```
  Case                                             Max          MaxOS           Live        Allocated      GCs  Wall Time
  IntMap (KeyMap TxId TxOut) (original)  1,185,790,008  3,191,865,344  1,194,345,304  199,779,514,568  176,472   398.977s
  IntMap (KeyMap TxId TxOut)             1,185,789,440  3,132,096,512  1,194,344,248  174,999,006,728  163,813   366.926s
  IntMap (KeyMap TxId TxOut) (sharing)   1,063,937,144  2,557,476,864  1,072,490,464  180,082,957,552  168,734   356.143s
  IntMap (KeyMap TxId TxOut')            1,163,115,976  3,128,950,784  1,171,668,776  182,641,648,320  171,162   337.480s
  IntMap (KeyMap TxId TxOut') (sharing)  1,041,263,648  2,487,222,272  1,049,814,984  189,083,461,088  177,386   334.723s
```
Legend:
* `TxOut'` is what implemented in this PR
* `TxOut` is what we have on master
* and the `original` what was there before #2534

Only this commit is important in this PR: 3da6065da1999f22de96040a7e7f73798726d42c the rest are changes to the benchmarking tool that were needed for getting this problem figured out.


JaredCorduan (approved):
LGTM, thanks @lehins !

libs/ledger-state/src/Cardano/Ledger/State/Transform.hs
```diff
@@ -1,54 +1,87 @@
...
+--     Alonzo.TxOutCompactDH cAddr cVal dh
+--       | Just (cAddr', sr) <- restructureAddr cAddr ->
+--         TxOut' (Alonzo.TxOutCompactDH cAddr' cVal dh) sr
+--     _ -> TxOutNoStake' txOut
+--   where
+--     restructureAddr cAddr =
+--       case decompactAddr cAddr of
+--         Addr ni pc (StakeRefBase sr) ->
+--           Just (compactAddr (Addr ni pc StakeRefNull), intern sr m)
+--         _ -> Nothing
```
nc6: Dead code?

lehins: You are right, it is no longer needed. I dumped most of the content in that module, since it was just experimentation with a better compaction of TxOut

PR #2554, opened 2021-11-16
add traces to time plutus script evalutaiton

goolord:
this should make it easier to view how long it takes plutus scripts to evaluate/validate

### DIFF
--- a/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/PlutusScriptApi.hs
+++ b/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/PlutusScriptApi.hs
@@ -68,6 +68,7 @@ import Cardano.Slotting.Time (SystemStart)
 import Data.Coders
 import Data.Foldable (foldl')
 import Data.Functor.Identity (Identity, runIdentity)
+import Data.List (intercalate)
 import qualified Data.Map as Map
 import Data.Maybe (mapMaybe)
 import Data.Proxy (Proxy (..))
@@ -75,6 +76,7 @@ import Data.Sequence.Strict (StrictSeq)
 import Data.Set (Set)
 import qualified Data.Set as Set
 import Data.Text (pack)
+import Debug.Trace (traceEvent)
 import GHC.Generics
 import GHC.Records (HasField (..))
 import NoThunks.Class (NoThunks)
@@ -242,8 +244,21 @@ evalScripts tx ((AlonzoScript.TimelockScript timelock, _, _, _) : rest) =
     lift True = Passes
     lift False = Fails [OnePhaseFailure . pack . show $ timelock]
 evalScripts tx ((AlonzoScript.PlutusScript lang pscript, ds, units, cost) : rest) =
-  runPLCScript (Proxy @era) lang cost pscript units (map getPlutusData ds)
-    `andResult` evalScripts tx rest
+  let beginMsg =
+        intercalate
+          ","
+          [ "[LEDGER][PLUTUS_SCRIPT]",
+            "BEGIN"
+          ]
+      !res = traceEvent beginMsg $ runPLCScript (Proxy @era) lang cost pscript units (map getPlutusData ds)
+      endMsg =
+        intercalate
+          ","
+          [ "[LEDGER][PLUTUS_SCRIPT]",
+            "END",
+            "res = " <> show res
+          ]
+   in (traceEvent endMsg res) `andResult` evalScripts tx rest
 
 -- Collect information (purpose and hash) about all the scripts in a Tx.
 -- THE SPEC CALLS FOR A SET, BUT THAT NEEDS A BUNCH OF ORD INSTANCES (DCert)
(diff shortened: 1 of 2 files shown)


eras/alonzo/impl/src/Cardano/Ledger/Alonzo/PlutusScriptApi.hs
```diff
@@ -242,8 +246,24 @@ evalScripts tx ((AlonzoScript.TimelockScript timelock, _, _, _) : rest) =
     lift True = Passes
     lift False = Fails [OnePhaseFailure . pack . show $ timelock]
 evalScripts tx ((AlonzoScript.PlutusScript lang pscript, ds, units, cost) : rest) =
-  runPLCScript (Proxy @era) lang cost pscript units (map getPlutusData ds)
-    `andResult` evalScripts tx rest
+  let ~traceIdentifier = show $ hashWithSerialiser @ShortHash toCBOR pscript
```
nc6: We're not using `-XStrict`, so surely the laziness markers are redundant here?

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Rules/Utxos.hs
```diff
@@ -182,6 +188,23 @@ scriptsValidateTransition = do
         totalDeposits pp (`Map.notMember` poolParams) txcerts <-> refunded
   sysSt <- liftSTS $ asks systemStart
   ei <- liftSTS $ asks epochInfo
+  let ~traceIdentifier =
+        (castHash $ hashWithSerialiser @ShortHash toCBOR slot)
+          `xor` (castHash $ hashWithSerialiser @ShortHash toCBOR txb)
+          `xor` (castHash $ hashWithSerialiser @ShortHash toCBOR poolParams)
+          `xor` (castHash $ hashWithSerialiser @ShortHash toCBOR genDelegs)
+          `xor` (castHash $ hashWithSerialiser @ShortHash toCBOR deposited)
+          `xor` (castHash $ hashWithSerialiser @ShortHash toCBOR utxo)
```
nc6: This seems like an awful lot of hashing - for what reason? Is this identifier going to be something consumers of this event will actually use?

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Rules/Utxos.hs
```diff
@@ -232,12 +265,35 @@ scriptsNotValidateTransition = do
   let txb = body tx
   sysSt <- liftSTS $ asks systemStart
   ei <- liftSTS $ asks epochInfo
+  let ~traceIdentifier =
+        (castHash $ hashWithSerialiser @ShortHash toCBOR txb)
+          `xor` (castHash $ hashWithSerialiser @ShortHash toCBOR utxo)
```
nc6: Erm.... this is hashing the _entire_ UTxO set. Please do not do this!

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Rules/Utxos.hs
nc6: And on second look, this hashes the UTxO!

nc6 (changes requested):

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/PlutusScriptApi.hs
goolord: yes, i just put them there to make it obvious on first read that the intention is that these don't get evaluated if the -eventlog flag is off. if it's having the opposite of that effect I'll just remove it

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Rules/Utxos.hs
goolord: right, i guess the txbody should be good enough to uniquely identify each call? i wasn't really concerned with performance since this should only get evaluated with `-eventlog` is passed and i was not aware of how insane this was lol

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Rules/Utxos.hs
goolord: yes, this is just for the GHC event log. it's important to be able to correlate the "BEGIN"/"END" events so that we have useful information about how long script validation/execution takes. this shouldn't be evaluated without the appropriate rts flag on. I just removed hashing the utxo since maybe the txbody is enough

nc6: From the `Debug.Trace` module:
> Note that the string message is always evaluated, whether or not profiling is available or enabled.

So this will be evaluated for every transaction. The ledger is single-threaded, so correlation shouldn't be an issue.

I would suggest, if anything, just using the txbody hash. This is at least already serialised. However, I'd question whether we really need this, since events should not be interleaved.

nc6 (approved):

nc6:
bors merge

nc6:
bors retry

PR #2556, opened 2021-11-17
make deserialiseAddrFull function

JaredCorduan:
This PR adds a function `deserialiseAddrFull`, which is nearly identical to `deserialiseAddr`, but does not allow for extra bytes.

### DIFF
--- a/libs/cardano-ledger-core/src/Cardano/Ledger/Address.hs
+++ b/libs/cardano-ledger-core/src/Cardano/Ledger/Address.hs
@@ -141,14 +141,19 @@ serialiseAddr = BSL.toStrict . B.runPut . putAddr
 deserialiseAddr :: CC.Crypto crypto => ByteString -> Maybe (Addr crypto)
 deserialiseAddr bs = case B.runGetOrFail getAddr (BSL.fromStrict bs) of
   Left (_remaining, _offset, _message) -> Nothing
-  Right (_remaining, _offset, result) -> Just result
+  Right (remaining, _offset, result) ->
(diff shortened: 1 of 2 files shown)


lehins (approved):
:+1:

libs/cardano-ledger-core/src/Cardano/Ledger/Address.hs
```diff
@@ -141,14 +141,20 @@ serialiseAddr = BSL.toStrict . B.runPut . putAddr
...
+  Right (remaining, _offset, result) ->
+    if BSL.null remaining
+      then Just result
+      else Nothing
 
 -- | Deserialise a stake refence from a address. This will fail if this
 -- is a Bootstrap address (or malformed).
 deserialiseAddrStakeRef :: CC.Crypto crypto => ByteString -> Maybe (StakeReference crypto)
 deserialiseAddrStakeRef bs = case B.runGetOrFail getAddrStakeReference (BSL.fromStrict bs) of
   Left (_remaining, _offset, _message) -> Nothing
```
lehins: Just a more compact suggestion, feel free to ignore it if you don't like it:

```haskell
deserialiseAddrStakeRef bs =
  case B.runGetOrFail getAddrStakeReference (BSL.fromStrict bs) of
    Right (remaining, _offset, result)
      | not (BSL.null remaining)  -> result
    _ -> Nothing
```

JaredCorduan: I like it!

PR #2557, opened 2021-11-19
 Sharing of values during deserialization

lehins:
This is an alternative approach to the one implemented in #2548 

Applied sharing to all places possible in EpochState, including TxOuts in the UTxO. These are the results we get:
```
Case                                 Max          MaxOS           Live        Allocated      GCs
EpochState (FromCBOR)      1,199,503,544  3,557,818,368  1,199,503,544   70,260,877,216   63,501
EpochState (no-sharing)    1,970,142,176  4,795,138,048  1,910,834,120  408,093,302,920  388,393
```


TimSheard:
A bit of an overview might help. As I see it, we introduce a new class, 
  FromSharedCBOR t where ...
with an associated type family:  Share t = ...
that specifies what is being shared when t is being de-serialized.
We need this flexibility, because different 't' may be sharing several different things. Usually (Share t) is a single type or a pair of types.
We also introduce some new monadic mechanisms, that hide  (Share t) inside a state transformer. This introduces a lot of 'lifting" in the instances to move in and out of the monad.

The old PR https://github.com/input-output-hk/cardano-ledger/pull/2548 dose similar things, but does them explicitly and manually.  we trade lifting and monads, for pattern matching over tuples, and for explicit state transformers: fromCBORShare :: state -> (Decoder, state) in the few places it is needed.
My opinion is that the old PR is easier to understand and use.

libs/small-steps/src/Data/Sharing.hs
```diff
@@ -0,0 +1,194 @@
...
+    x <- fromSharedPlusCBOR
+    x <$ put s
+
+  -- | Deserialize with sharing and add the state that used for sharing. Default
+  -- implementation will add value returned by `getShare` for adding to the
+  -- state.
+  fromSharedPlusCBOR :: StateT (Share a) (Decoder s) a
+  fromSharedPlusCBOR = do
+    x <- fromSharedCBOR
+    x <$ mapM_ (\s -> modify' (s <>)) (getShare x)
```
lehins: @TimSheard I took your suggestion and brought in the `fromSharedPlusCBOR` function into the  class. This allows us to implement decoding with sharing in two ways:

1. implementing `fromSharedCBOR` and `getShare` (example instances are in this module)
2. implementing only `fromSharedPlusCBOR`, in which case the state used for sharing needs to be modified manually and `getShare` is completely ignored.

`MINIMAL` pragma will warn the programmer if it wasn't done right

eras/shelley/impl/src/Cardano/Ledger/Shelley/EpochBoundary.hs
```diff
@@ -79,8 +84,10 @@ deriving newtype instance Typeable crypto => NoThunks (Stake crypto)
 deriving newtype instance
   CC.Crypto crypto => ToCBOR (Stake crypto)
 
-deriving newtype instance
-  CC.Crypto crypto => FromCBOR (Stake crypto)
+instance CC.Crypto crypto => FromSharedCBOR (Stake crypto) where
+  type Share (Stake crypto) = Share (VMap VB VP (Credential 'Staking crypto) (CompactForm Coin))
```
nc6, line 84: Can we not use newtype deriving for this?

eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState.hs
```diff
@@ -407,15 +422,22 @@ instance
...
-
-instance
-  CC.Crypto crypto =>
-  FromCBOR (DPState crypto)
-  where
-  fromCBOR =
-    decodeRecordNamed "DPState" (const 2) $ DPState <$> fromCBOR <*> fromCBOR
+  toCBOR DPState {_pstate, _dstate} =
+    encodeListLen 2
+      <> toCBOR _pstate -- We get better sharing when encoding pstate before dstate
```
nc6, line 427: OK, so we need to rebuild the ledger state. We should definitely coordinate the changes with the other things that update the ledger state serialisation.

eras/shelley/impl/src/Cardano/Ledger/Shelley/RewardUpdate.hs
```diff
@@ -10,7 +10,9 @@
 {-# LANGUAGE MultiParamTypeClasses #-}
 {-# LANGUAGE NamedFieldPuns #-}
 {-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE ScopedTypeVariables #-}
```
nc6:
```suggestion
```

eras/shelley/impl/src/Cardano/Ledger/Shelley/RewardUpdate.hs
```diff
@@ -10,7 +10,9 @@
...
 {-# LANGUAGE StandaloneDeriving #-}
+{-# LANGUAGE TypeApplications #-}
```
nc6:
```suggestion
```

eras/shelley/impl/src/Cardano/Ledger/Shelley/TxBody.hs
```diff
@@ -1014,15 +1015,25 @@ instance-- use the weakest constraint necessary
       <> toCBOR addr
       <> toCBOR coin
 
-instance-- use the weakest constraint necessary
-
+-- use the weakest constraint necessary
```
nc6:
```suggestion
```

eras/shelley/impl/src/Cardano/Ledger/Shelley/TxBody.hs
```diff
@@ -1014,15 +1015,25 @@ instance-- use the weakest constraint necessary
...
   (Era era, TransTxOut DecodeNonNegative era, Show (Core.Value era)) =>
   FromCBOR (TxOut era)
   where
-  fromCBOR = decodeRecordNamed "TxOut" (const 2) $ do
-    cAddr <- fromCBOR
-    coin <- decodeNonNegative
-    pure $ TxOutCompact cAddr coin
+  fromCBOR = fromNotSharedCBOR
+
+-- This instance does not do any sharing and is isomorphic to FromCBOR
```
nc6, line 1020: If we're not sharing here, why bother with this instance?

libs/ledger-state/app/Main.hs
```diff
@@ -14,7 +14,10 @@ import System.IO
 data Opts = Opts
   { -- | Path to the CBOR encoded NewEpochState data type, which will be used to
     -- load into sqlite database
-    optsLedgerStateBinaryFile :: Maybe FilePath,
+    optsNewEpochStateBinaryFile :: Maybe FilePath,
+    -- | Path to the CBOR encoded NewEpochState data type, which will be used to
+    -- load into sqlite database
```
nc6: Comment is wrong here? Sounds like this is for the `EpochState`? Why would one use this separately?

libs/ledger-state/app/Main.hs
```diff
@@ -14,7 +14,10 @@ import System.IO
 data Opts = Opts
   { -- | Path to the CBOR encoded NewEpochState data type, which will be used to
     -- load into sqlite database
-    optsLedgerStateBinaryFile :: Maybe FilePath,
+    optsNewEpochStateBinaryFile :: Maybe FilePath,
```
nc6, line 26: Duplicate options?

libs/ledger-state/app/Main.hs
```diff
@@ -55,15 +68,22 @@ main = do
...
+      let dbFp = T.pack dbFpStr
+      epochState <- loadEpochState dbFp
+      putStrLn "Written EpochState into the database"
+      writeEpochState binFp epochState
+
+-- forM_ (optsSqliteDbFile opts) $ \dbFpStr -> do
+--   let dbFp = T.pack dbFpStr
+--   km <- loadDbUTxO txIdSharingKeyMap dbFp
+--   m <- loadDbUTxO noSharing dbFp
+--   testKeyMap km m
```
nc6, line 98: Dead code?

libs/ledger-state/bench/Memory.hs
```diff
@@ -12,7 +13,10 @@ import Weigh
 data Opts = Opts
   { -- | Path to the CBOR encoded NewEpochState data type, which will be used to
     -- load into sqlite database
-    optsLedgerStateBinaryFile :: Maybe FilePath,
+    optsNewEpochStateBinaryFile :: Maybe FilePath,
+    -- | Path to the CBOR encoded EpochState data type, which will be used to
+    -- load into sqlite database
+    optsEpochStateBinaryFile :: Maybe FilePath,
```
nc6, line 19: Ditto comment?

libs/ledger-state/bench/Memory.hs
```diff
@@ -48,22 +59,26 @@ main = do
...
+
+-- forM_ mEpochStateEntity $ \_ese ->
+--   -- wgroup "EpochState" $ do
+--   --   io "SnapShots - no sharing" (loadSnapShotsNoSharingM dbFp) _ese
+--   --   io "SnapShots - with sharing" (loadSnapShotsWithSharingM dbFp) _ese
+--   --   io "SnapShots (Vector) - no sharing" (loadSnapShotsNoSharing dbFp) _ese
+--   --   io "SnapShots (Vector) - with sharing" (loadSnapShotsWithSharing dbFp) _ese
+--   wgroup "DState+UTxO" $ do
+--     io "IntMap (KeyMap TxId TxOut)" getLedgerStateNoSharingKeyMap dbFp
+--     io "IntMap (KeyMap TxId TxOut) (sharing)" getLedgerStateWithSharingKeyMap dbFp
```
nc6, line 81: More dead code? Or is this maintained for a reason?

libs/small-steps/src/Data/Sharing.hs
```diff
@@ -0,0 +1,208 @@
...
+import Data.Kind
+import qualified Data.Map.Strict as Map
+import Data.Map.Strict.Internal
+import Data.Primitive.Types (Prim)
+import Lens.Micro
+
+-- =======================================
+
+data Intern a = Intern
+  { internMaybe :: a -> Maybe a,
```
nc6: Can we add a comment on this function? I guess it's the critical function here

nc6 (changes requested):
Made a bunch of comments, but otherwise I think this looks good!

libs/ledger-state/bench/Memory.hs
lehins: Running all benchmarks takes ages. With `weigh` it is unfortunately not possible to select through cli which benchmarks you want to run, so commenting out code for now is the poor man's selection. Once benchmarks stabilize I can add manual selection options, but for now let's not worry about it. This `ledger-state` package is only used as a tool, no one depends or will depend on it.

libs/ledger-state/app/Main.hs
lehins: It can still be useful. I will do a final cleanup of ledger-state package once we are done with memory optimizations

libs/small-steps/src/Data/Sharing.hs
lehins: Indeed it could use a comment. I'll add one

libs/ledger-state/app/Main.hs
lehins: Do you mean options in `bench` and `app` are the same? If that's the question than they look the same but they mean different things (eg. one is for reading, while another one is for writing). Moreover I suspect they will diverge in the future.

libs/ledger-state/app/Main.hs
lehins: Yep, copy and paste error. Fixed and added a comment about how these options are used:

A bit more insight into options, at least into how they are currently used in this PR. This will  change after new serialization in this PR is realeased:

* `optsNewEpochStateBinaryFile` is for reading a previously serialized `NewEpochState` produced by cadano-cli` and is used to populate sqlite database
* `optsEpochStateBinaryFile` is used for grabbing data from sqlite, constructing `EpochState` (in a new format) and writing it into the cbor serialized file

eras/shelley/impl/src/Cardano/Ledger/Shelley/EpochBoundary.hs
lehins: Unfortunately we cannot use GND, which is caused by associated type family and it's usage in `StateT`. Beats me why ghc can't figure it out, but this is the error we get if we try to derive it:
```haskell
src/Cardano/Ledger/Shelley/EpochBoundary.hs:(87,1)-(88,51): error: …
    • Couldn't match type ‘VMap
                             VB VP (Credential 'Staking crypto) (CompactForm Coin)’
                     with ‘Stake crypto’
        arising from a use of ‘ghc-prim-0.6.1:GHC.Prim.coerce’
    • In the expression:
          ghc-prim-0.6.1:GHC.Prim.coerce
            @(Control.Monad.Trans.State.Strict.StateT (Share (VMap VB VP (Credential Staking crypto) (CompactForm Coin))) (Codec.CBOR.Decoding.Decoder s) (VMap VB VP (Credential Staking crypto) (CompactForm Coin)))
            @(Control.Monad.Trans.State.Strict.StateT (Share (Stake crypto)) (Codec.CBOR.Decoding.Decoder s) (Stake crypto))
            (fromSharedPlusCBOR
               @(VMap VB VP (Credential Staking crypto) (CompactForm Coin))) ::
            forall (s :: TYPE ghc-prim-0.6.1:GHC.Types.LiftedRep).
            Control.Monad.Trans.State.Strict.StateT (Share (Stake crypto)) (Codec.CBOR.Decoding.Decoder s) (Stake crypto)
```

eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState.hs
lehins: Most definitely! We should talk about this, cause I am not sure what is the process when CBOR is changed in a backwards incompatible manner

eras/shelley/impl/src/Cardano/Ledger/Shelley/TxBody.hs
lehins: Because of the `FromSharedCBOR (Core.TxOut era)` constraint on this instance:

```
instance
  ( CC.Crypto (Crypto era),
    FromSharedCBOR (Core.TxOut era),
    Share (Core.TxOut era) ~ Interns (Credential 'Staking (Crypto era))
  ) =>
  FromSharedCBOR (UTxO era)
  where
 ....
````

In other words Alonzo `TxOut` (and so will Babbage) has sharing capabilities, while Shelley TxOut is unshareable

File: libs/ledger-state/app/Main.hs (hunk 9 above)
Comment (nc6) [reply libs/ledger-state/app/Main.hs:26]: I meant as with the other comment, sounds like it's addressed

File: eras/shelley/impl/src/Cardano/Ledger/Shelley/EpochBoundary.hs (hunk 2 above)
Comment (lehins) [reply eras/shelley/impl/src/Cardano/Ledger/Shelley/EpochBoundary.hs:84]: Yeah, if it wasn't for this problem this whole sharing approach could be pushed further up the stream into `FromCBOR` class and it would have been backwards compatible. But inability to derive for `newtypes` is a deal breaker on doing that, althoughusing an extra `FromSharedCBOR` is that terrible IMHO

File: libs/small-steps/src/Data/Sharing.hs (hunk 13 above)
Comment (lehins) [reply libs/small-steps/src/Data/Sharing.hs]: I added some haddock, let me know if it makes sense and if it is enough of an explanation

File: libs/ledger-state/bench/Memory.hs (hunk 11 above)
Comment (lehins) [reply libs/ledger-state/bench/Memory.hs:19]: Fixed comment

File: libs/ledger-state/app/Main.hs [hunk 14]
```diff
@@ -11,10 +11,22 @@ import Data.Text as T (pack)
 import Options.Applicative
 import System.IO
 
+-- | Insigt into options:
```
nc6:
```suggestion
-- | Insight into options:
```

libs/small-steps/src/Data/Sharing.hs
```diff
@@ -0,0 +1,220 @@
...
+import qualified Data.Map.Strict as Map
+import Data.Map.Strict.Internal
+import Data.Primitive.Types (Prim)
+import Lens.Micro
+
+-- =======================================
+
+-- | This is an abstract interface that does the interning. In other words it
+-- does the actual sharing by looking up the supplied value in some existing
+-- data structure and uses that value instead. Relying this interface gives us
```
nc6:
```suggestion
-- data structure and uses that value instead. Relying on this interface gives us
```

libs/small-steps/src/Data/Sharing.hs
```diff
@@ -0,0 +1,220 @@
...
+-- the benefit of ignoring the type of underlying data structure and allows us
+-- to compose many `Intern`s with the monoidal interface provided by `Interns`
+-- wrapper. In order to create an `Intern` see the `internsFromMap` or
+-- `internsFromVMap` functions.
+data Intern a = Intern
+  { -- | Function that will do the interning. Is value is not available then
```
nc6:
```suggestion
  { -- | Function that will do the interning. If value is not available then
```

libs/small-steps/src/Data/Sharing.hs
```diff
@@ -0,0 +1,220 @@
...
+    -- `Nothing` is returned.
+    internMaybe :: a -> Maybe a,
+    -- | Used for sorting. Normally set to the size of the underlying data
+    -- structure. Keeping interns sorted with respect to how many elements
+    -- is in the underlying data structure in theory gives a better chance of
+    -- successfull intern hit sooner rather than later.
```
nc6:
```suggestion
    -- successful intern hit sooner rather than later.
```

nc6 (approved):
Misc typos, otherwise looks good!

PR #2560, opened 2021-11-23
Babbage: TxBody + TxOut

### LINKED ISSUE #2463: Babbage TxBody
Implement the new Babbage `TxBody` type and all necessary instances.

Babbage adds three new fields:
- [ ] [reference inputs](https://github.com/cardano-foundation/CIPs/pull/159)
- [ ] collateral return output
- [ ] total collateral

### LINKED ISSUE #2465: Babbage TxOut
Implement the new Babbage `TxOut` type and all necessary instances.

Babbage adds two things to the Alonzo `TxOut' type.

- [x] [Inline datums](https://github.com/cardano-foundation/CIPs/pull/160)
- [x] [reference scripts](https://github.com/cardano-foundation/CIPs/pull/161)

goolord:
this is the beginning of the babbage era work so this pr includes all of the necessary changes to build the new era + the TxBody module.

resolves https://github.com/input-output-hk/cardano-ledger/issues/2465
resolves https://github.com/input-output-hk/cardano-ledger/issues/2463


eras/babbage/impl/src/Cardano/Ledger/Babbage/TxBody.hs
```diff
@@ -0,0 +1,886 @@
...
+{-# COMPLETE TxOut #-}
+
+-- ======================================
+
+type ScriptIntegrityHash crypto = SafeHash crypto EraIndependentScriptIntegrity
+
+data TxBodyRaw era = TxBodyRaw
+  { _inputs :: !(Set (TxIn (Crypto era))),
+    _collateral :: !(Set (TxIn (Crypto era))),
+    _collateralAddresses :: !(StrictMaybe (Addr (Crypto era))),
```
nc6: This should be singular (there's only one) and should have a more helpful name - maybe `collateralReturn`, to indicate its function

eras/babbage/impl/src/Cardano/Ledger/Babbage/TxBody.hs
```diff
@@ -0,0 +1,886 @@
...
+import Data.Typeable (Proxy (..), Typeable, (:~:) (Refl))
+import Data.Word
+import GHC.Generics (Generic)
+import GHC.Records (HasField (..))
+import GHC.Stack (HasCallStack)
+import GHC.TypeLits
+import NoThunks.Class (InspectHeapNamed (..), NoThunks)
+import Prelude hiding (lookup)
+
+data TxOut era
```
nc6, line 138: Where is the burn address here?

nc6 (dismissed):
Looks reasonable, but I don't see where the burn address has been defined.

eras/babbage/impl/src/Cardano/Ledger/Babbage/TxBody.hs
goolord: oh, i spoke with jared and he said that we aren't adding the burn address anymore

eras/babbage/impl/src/Cardano/Ledger/Babbage/TxBody.hs
```diff
@@ -0,0 +1,886 @@
...
+  | TxOutCompactDH
+      {-# UNPACK #-} !(CompactAddr (Crypto era))
+      !(CompactForm (Core.Value era))
+      !(DataHash (Crypto era))
+  | TxOutCompactDatum
+      {-# UNPACK #-} !(CompactAddr (Crypto era))
+      !(CompactForm (Core.Value era))
+      !(Data era)
+  | SizeHash (CC.ADDRHASH (Crypto era)) ~ 28 =>
+    TxOut_AddrHash28_AdaOnly
```
lehins: Note that TxOut definition has changed since you copied it over from Alonzo.

I suggest recopying the whole thing once again and adding `TxOutCompactDatum` from scratch. This is the price we usually have to pay for copy pasting code :(

eras/babbage/impl/src/Cardano/Ledger/Babbage/TxBody.hs
```diff
@@ -0,0 +1,886 @@
...
+encodeDataHash32 ::
+  forall crypto.
+  (HashAlgorithm (CC.HASH crypto)) =>
+  DataHash crypto ->
+  Maybe (SizeHash (CC.HASH crypto) :~: 32, Word64, Word64, Word64, Word64)
+encodeDataHash32 dataHash = do
+  refl@Refl <- sameNat (Proxy @(SizeHash (CC.HASH crypto))) (Proxy @32)
+  case hashToPackedBytes (extractHash dataHash) of
+    PackedBytes32 a b c d -> Just (refl, a, b, c, d)
+    _ -> Nothing
```
lehins: There is no reason to copy all of those functions, it is better to export them from `Alonzo.TxBody` and reuse them here.

lehins (changes requested):
@nc6 I have to disagree with you, it doesn't look reasonable, there is way too much duplicated functionality in this PR.

We need to try and figure out how we can reduce duplication as much as possible. Perfect example why this is important is the fact that definition of `TxOut` has recently changed as well as many of the helper functions that deal with it. Soon CBOR deserialization will change as well (due to sharing)

Maybe we can approach this problem from a different angle and instead of creating a new `TxOut` type for babbage we can create a new type family for datums, eg:

```haskell
type family Datum era :: Type
...
  | TxOutCompactDH
      {-# UNPACK #-} !(CompactAddr (Crypto era))
      !(CompactForm (Core.Value era))
      !(Datum era)

... 
type instance Datum Alonzo = DataHash (Crypto Alonzo)
type instance Datum Babbage = Either (DataHash (Crypto Aonzo)) (Data Babbage)
-- ^ Note that `Either` is just an example, a custom type would be better
```

or something along those lines. This way we could avoid copying this whole monstrosity. 
@goolord , @nc6 , @JaredCorduan Any thoughts on this?

goolord:
more code reuse would be nice. a type family is not like an open union though, don't just end up with the same problem? 

we are already reasonably polymorphic over `TxOut` for each era when destructing using `HasField`, or in functions like `decodeDataHash32` which just accept the parameters of the constructor as arguments. there are other techniques which don't change the semantics of `TxOut`

goolord:
there is still a maintainership burden but it seems small, I don't really have any strong opinions either way

eras/babbage/impl/src/Cardano/Ledger/Babbage/TxBody.hs
```diff
@@ -0,0 +1,805 @@
...
+import Prelude hiding (lookup)
+
+data TxOut era
+  = TxOutCompact
+      {-# UNPACK #-} !(CompactAddr (Crypto era))
+      !(CompactForm (Core.Value era))
+  | TxOutCompactDH
+      {-# UNPACK #-} !(CompactAddr (Crypto era))
+      !(CompactForm (Core.Value era))
+      !(DataHash (Crypto era))
```
lehins: Note that Alonzo's TxOut has separate pattern synonym defined for `TxOutCompact` and `TxOutCompactDH`. It is crucial not export these type contractors because they allow to bypass the optimization of unpacking ada only cases

eras/babbage/impl/src/Cardano/Ledger/Babbage/TxBody.hs
```diff
@@ -0,0 +1,805 @@
...
+      Just 2 ->
+        TxOutCompact
+          <$> fromCBOR
+          <*> decodeNonNegative
+      Just 3 ->
+        TxOutCompactDH
+          <$> fromCBOR
+          <*> decodeNonNegative
+          <*> fromCBOR
+      Just _ -> cborError $ DecoderErrorCustom "txout" "wrong number of terms in txout"
```
lehins: Here is a good example of the problem from my other comment. Using TxOutCompact and TxOutCompactDH completely bypasses the ada only optimizations.

Also this deserialization does not account for decoding TxOut with datums

eras/babbage/impl/src/Cardano/Ledger/Babbage/TxBody.hs
```diff
@@ -0,0 +1,805 @@
...
+  ( Era era,
+    Compactible (Core.Value era),
+    Val (Core.Value era),
+    HasCallStack
+  ) =>
+  Addr (Crypto era) ->
+  Core.Value era ->
+  StrictMaybe (DataHash (Crypto era)) ->
+  TxOut era
+pattern TxOut addr vl dh <-
```
lehins: This doesn't make sense. How anyone would ever construct a TxOut with `Data`? If I had to guess this should be something like this instead:
```suggestion
  Addr (Crypto era) ->
  Core.Value era ->
  Either (DataHash (Crypto era)) (Data era) ->
  TxOut era
pattern TxOut addr vl ed <-
```

eras/babbage/impl/src/Cardano/Ledger/Babbage/TxBody.hs
```diff
@@ -0,0 +1,805 @@
...
+    encodeListLen 3
+      <> toCBOR addr
+      <> toCBOR cv
+      <> toCBOR dh
+  toCBOR x =
+    let (addr, cv, dh) = viewCompactTxOut x
+     in encodeListLen 3
+          <> toCBOR addr
+          <> toCBOR cv
+          <> toCBOR dh
```
lehins: Serialization of `TxOutCompactDatum` is missing

eras/babbage/impl/src/Cardano/Ledger/Babbage/TxBody.hs
```diff
@@ -0,0 +1,805 @@
...
+    _reqSignerHashes :: Set (KeyHash 'Witness (Crypto era)),
+    _mint :: !(Value (Crypto era)),
+    -- The spec makes it clear that the mint field is a
+    -- Cardano.Ledger.Mary.Value.Value, not a Core.Value.
+    -- Operations on the TxBody in the AlonzoEra depend upon this.
+    _scriptIntegrityHash :: !(StrictMaybe (ScriptIntegrityHash (Crypto era))),
+    _adHash :: !(StrictMaybe (AuxiliaryDataHash (Crypto era))),
+    _txnetworkid :: !(StrictMaybe Network)
+  }
+  deriving (Generic, Typeable)
```
lehins: @JaredCorduan or @nc6 can you comment on this? I think we could do something like this instead?

```suggestion
data TxBodyRaw era = TxBodyRaw
  { txBodyAlonzo :: Alonzo.TxBodyRaw era
  , txBodyCollateralReturn :: !(StrictMaybe (TxOut era))
  }
```

This way we could reuse all the serialization/deserialization logic from alonzo. All this copy pasting is hurting my eyes. I will of course not stand in a way if that is how `era` parameterization was designed to be used.

eras/babbage/impl/src/Cardano/Ledger/Babbage/TxBody.hs
```diff
@@ -0,0 +1,805 @@
...
+      _adHash,
+      _txnetworkid
+    } =
+    Keyed
+      ( \i ifee ca o f t c w u b rsh mi sh ah ni ->
+          TxBodyRaw i ifee ca o c w f (ValidityInterval b t) u rsh mi sh ah ni
+      )
+      !> Key 0 (E encodeFoldable _inputs)
+      !> Key 13 (E encodeFoldable _collateral)
+      !> Key 13 (To _collateralReturn)
```
lehins: Correct me if I am wrong, but I assumed that key should be unique and reflect the position of the field during serlization/deserialization.

eras/babbage/impl/src/Cardano/Ledger/Babbage/TxBody.hs
```diff
@@ -0,0 +1,805 @@
...
+    Typeable (Core.Script era),
+    Typeable (Core.AuxiliaryData era),
+    Compactible (Core.Value era),
+    Show (Core.Value era),
+    DecodeNonNegative (Core.Value era),
+    FromCBOR (Annotator (Core.Script era)),
+    FromCBOR (PParamsDelta era),
+    ToCBOR (PParamsDelta era)
+  ) =>
+  FromCBOR (TxBodyRaw era)
```
lehins: Looks like deserialziation of `collateralReturn` is missing

lehins (changes requested):
I think we need a test suite added in this PR that at least checks serialization rountripping. It would have caught most of my comments.

eras/babbage/impl/src/Cardano/Ledger/Babbage/TxBody.hs
goolord: this is the current pattern synonym for `TxOutCompact`
```hs
-- TODO deprecate
pattern TxOutCompact ::
  ( Era era,
    HasCallStack
  ) =>
  CompactAddr (Crypto era) ->
  CompactForm (Core.Value era) ->
  TxOut era
pattern TxOutCompact addr vl <-
  (viewCompactTxOut -> (addr, vl, SNothing))
  where
    TxOutCompact cAddr cVal = TxOut (decompactAddr cAddr) (fromCompact cVal) SNothing
```
the `TODO deprecate` comment and viewpattern lead me to believe 1. the pattern synonym is mostly for convenience 2. unrolling this pattern synonym will behave identically 3. the intention is that these pattern synonyms will eventually be removed

goolord: i'll just revert it to be more like it is in alonzo for now, but i'm still curious what the motivation for keeping these is

goolord:
yeah i neglected to introduce the test suite yet since it would be a lot of additional work to do so until the rest of the babbage era stuff is done, but i can add tests gradually

goolord:
ok, adding tests at this point would actually be *way* more work than i thought. if it's alright i will fix what i can now and revisit the tests / correctness of this module later just to avoid adding way too much to this pr

JaredCorduan:
> ok, adding tests at this point would actually be _way_ more work than i thought. if it's alright i will fix what i can now and revisit the tests / correctness of this module later just to avoid adding way too much to this pr

work for me

eras/babbage/impl/src/Cardano/Ledger/Babbage/TxBody.hs
JaredCorduan: that correct, we are not implementing burn addresses in babbage

eras/babbage/impl/src/Cardano/Ledger/Babbage/TxBody.hs
JaredCorduan: I don't think either works either, since there are three cases: no datum, datum hash, or inline datum. Maybe we use `Maybe Either`?

eras/babbage/impl/src/Cardano/Ledger/Babbage/TxBody.hs
JaredCorduan: hmm. I like the idea of removing code copy when we can, but it's also weird to switch to this nested format now.

JaredCorduan:
> I think we need a test suite added in this PR that at least checks serialization rountripping. It would have caught most of my comments.

That's a good idea. @goolord I think it would be pretty easy to add a few round trip tests, you can reuse the existing tools, like here in Alonzo: https://github.com/input-output-hk/cardano-ledger/blob/e7b7773655f6f1acc600abd63a3de1c6599c6d98/eras/alonzo/test-suite/test/Test/Cardano/Ledger/Alonzo/Serialisation/Tripping.hs#L70-L104

I'm also fine with adding tests in a follow up PR.

eras/babbage/impl/src/Cardano/Ledger/Babbage/TxBody.hs
nc6: We could have the datatype defined like this, and unpack in the patterns.

eras/babbage/impl/src/Cardano/Ledger/Babbage/TxBody.hs
lehins: Makes sense. I think a custom data type would be more appropriate:

```haskell
data Datum era
  = NoDatum 
  | DatumHash !(DataHash (Crypto era))
  | Datum !(Data era)
```

goolord:
ok, i got the whole era working on a branch and the tripping tests passing, i'll push the correct version of this module after I pr some of the related changes i needed to make

goolord:
https://github.com/input-output-hk/cardano-ledger/pull/2593

goolord:
ok, this is up to date with the latest version of the spec, and has changes from this branch https://github.com/input-output-hk/cardano-ledger/tree/zachc/babbage-big which has the test suite running (+ most of the other babbage changes which i'm goin to port over)

this pr https://github.com/input-output-hk/cardano-ledger/pull/2593 is necessary for the module to actually be useful

eras/babbage/impl/src/Cardano/Ledger/Babbage/TxBody.hs
```diff
@@ -0,0 +1,897 @@
...
+  toCBOR (TxOutCompact addr cv) =
+    encodeListLen 2
+      <> toCBOR addr
+      <> toCBOR cv
+  toCBOR (TxOutCompactDatum addr cv d) =
+    encodeListLen 4
+      <> toCBOR True
+      <> toCBOR addr
+      <> toCBOR cv
+      <> toCBOR d
```
goolord, line 606: this feels kinda hacky, but putting a break before the `encodeListLen` makes tests fail

JaredCorduan (approved):
Looks good to me! This has all the new features I was expecting.

eras/babbage/impl/src/Cardano/Ledger/Babbage/TxBody.hs
```diff
@@ -0,0 +1,907 @@
...
+            <$> fromCBOR
+            <*> decodeNonNegative
+      Just 3 ->
+        fmap pure $
+          TxOutCompactDH
+            <$> fromCBOR
+            <*> decodeNonNegative
+            <*> fromCBOR
+      Just 4 -> do
+        _ <- fromCBOR @Bool
```
lehins: I understand that it is not required, but just to be safe let's utilize MonadFail and pattern match to ensure we do get the expected value.

```suggestion
        True <- fromCBOR @Bool
```

lehins (approved):
LGTM. One minor suggestion.

PR #2561, opened 2021-11-24
Benchmarking

nc6:
- Update the Tx benchmarks (formats had changed and they were no longer running)
- Update the description of how to generate the files
- Add a benchmarking workflow


lehins (approved):
That is pretty cool. TIL about benchmarks with github actions.

FYI. See #2563 for the fix for CI failure. It is benign so feel free to rerun the CI make it pass

PR #2563, opened 2021-11-24
Minor bugfixes

lehins:
This PR fixes two bugs:

1. utxow property tests failed recently on CI: https://github.com/input-output-hk/cardano-ledger/runs/4311566920?check_suite_focus=true
  Reason is `collateralPercentage` was genrated to be 0, which caused the test to not generate any collateral, despite that it was required.
* Deserialization of BiMap expects keys to be distinct and ordered. Ensure that this is indeed the case and produce deserialization error when it is violated.


nc6 (approved):
Thanks!

PR #2564, opened 2021-11-25
Reorganized SetAlgebra by breaking it into 4 modules. 

TimSheard:
As we introduce new datatypes that are more compact, we will need new SetAlgebra Instances.  This ha become problematic since SetAgebra is a huge file, and the pieces of code pertinent to one type are spread all over.

So we reorganize by breaking it into four modules. The new ones are Control.Iterate.BaseTypes, Control.Iterate.Exp, and Control.Iterate.BiMap

To add a new type XXX, add Control.Iterate.XXX where the new type is defined.
Then add instances in Control.Iterate.BaseTypes. 
The other modules Control.Iterate.SetAlgebra and Control.Iterate.Exp are independent from what the Base types are.


nc6:
Can you talk a little bit about the general structure? E.g. what's the difference between `Control.Iterate.SetAlgebra` and `Control.SetAlgebra`? At the moment I still don't have a general conception of how things break down here.

TimSheard:
Over view of system structure.

Normal users, who just want to use Set algebra expressions should use:
import Control.SetAlgebra(eval, setSingleton, (∩), (⊆), (▷) )  

The modules
import Control.Iterate.SetAlgebra
import Control.Iterate.BaseTypes
import Control.Iterate.Exp
import Control.Iterate.BiMap

divide the system into components. imports of these modules should be limited to the SetAlgebra systems, and tests and benchmarks of the system that need lower level operations.  The organization of these modules is this.

Control.Iterate.BaseTypes  defines the classes that types that use the SetAlgebra system must defined instances. The file also contains instances for things like lists, sets, Data.Map etc.

Control.Iterate.BiMap defines a new type that we use in the system. Instances for this type are included in Control.Iterate.BaseTypes.

import Control.Iterate.Exp defines the internal intermediate data structures that the system uses to perform runtime pattern matching. Users do not need to understand these types.

Control.Iterate.SetAlgebra contains the evaluation strategies, eval, compute, compile etc.

JaredCorduan (approved):
Thank you for splitting this up @TimSheard !

libs/small-steps/src/Control/Iterate/BiMap.hs
```diff
@@ -0,0 +1,144 @@
...
+        !x <- decodeMapAsBimap
+        dropMap (void $ fromCBOR @b) (void $ fromCBOR @a)
+        return x
+      k -> invalidKey (fromIntegral k)
+
+-- | Decode a serialised CBOR Map as a Bimap
+decodeMapAsBimap ::
+  (FromCBOR a, FromCBOR b, Ord a, Ord b) =>
+  Decoder s (BiMap b a b)
+decodeMapAsBimap = decodeMapSkel biMapFromAscDistinctList
```
lehins: This function is broken again, see the fix in 7ef96a683bc0ce6ab1be25ac0ced5c212dea90a7

lehins (changes requested):
It looks great overall, however it seems that you reverted one of the resent PRs: #2563

libs/small-steps/src/Control/Iterate/BiMap.hs
TimSheard: Good catch. I didn't realise this. I have fixed it back.

lehins (approved):

PR #2565, opened 2021-11-25
Eq for MemoBytes should compare the bytes.

nc6:
Two instances of a memoised type are equal iff they have the same bytes.
Justification:
- We require that deserialisation is left inverse to serialisation. So
  if we have
  `serialise x == serialise y`
  `deserialise (serialise x) == deserialise (serialise y)`
  `x == y`
- `Eq` should be compatible with hashing (e.g. if `a == b`, then
  `hash a == hash b`). Currently, this is broken. If I serialise a
  `TxBodyRaw` in two different valid ways (as is perfectly allowed),
  then the resulting `TxBody` will currently compare as equal but hash
  to a different value.
- In choosing not to require a canonical serialisation, the serialised
  bytes are an integral part of the type, and must be considered in
  comparing for equality.

### DIFF
--- a/libs/small-steps/src/Data/MemoBytes.hs
+++ b/libs/small-steps/src/Data/MemoBytes.hs
@@ -69,7 +69,7 @@ instance (Typeable t, FromCBOR (Annotator t)) => FromCBOR (Annotator (MemoBytes 
     (Annotator getT, Annotator getBytes) <- withSlice fromCBOR
     pure (Annotator (\fullbytes -> Memo (getT fullbytes) (toShort (toStrict (getBytes fullbytes)))))
 
-instance Eq t => Eq (MemoBytes t) where (Memo x _) == (Memo y _) = x == y
+instance Eq (MemoBytes t) where (Memo _ x) == (Memo _ y) = x == y
 
 instance Show t => Show (MemoBytes t) where show (Memo y _) = show y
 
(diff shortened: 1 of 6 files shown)


libs/small-steps/src/Data/MemoBytes.hs
```diff
@@ -69,7 +69,7 @@ instance (Typeable t, FromCBOR (Annotator t)) => FromCBOR (Annotator (MemoBytes
     (Annotator getT, Annotator getBytes) <- withSlice fromCBOR
     pure (Annotator (\fullbytes -> Memo (getT fullbytes) (toShort (toStrict (getBytes fullbytes)))))
 
-instance Eq t => Eq (MemoBytes t) where (Memo x _) == (Memo y _) = x == y
+instance Eq (MemoBytes t) where (Memo _ x) == (Memo _ y) = x == y
```
KtorZ, line 72: Not comparing both constituents of `Memo` implicitly supposes that if two binary representations are equal, then their Haskell representation must be equal a well; is there anything truly enforcing that? While it sounds plausible in my head, is it really the case :thinking: ?

nc6: This was what I was justifying above:
```
serialise x == serialise y
deserialise (serialise x) == deserialise (serialise y)
x == y
```

The property that `deserialise` is left-inverse to `serialise` _is_ a required property for any serialisable type (as opposed to the other way around!). And we have tests verifying this property for serialised types.

redxaxder (approved):

PR #2566, opened 2021-11-29
Export an unsafe constructor for Validated

### LINKED ISSUE #2558: Export an unsafe constructor for Validated tx
This will allow tests with Txs that are not valid, ie they miss some witnesses, but are good enough to be included in blocks that are reapplied.

kderme:
Fixes https://github.com/input-output-hk/cardano-ledger/issues/2558


nc6 (approved):

nc6:
Thanks for sorting this!

PR #2568, opened 2021-12-02
Move ApplyTx benchmarks to using a fixed seed.

nc6:
This should prevent problems where serialisation changes result in
broken benchmarks. Additionally, we now test Allegra and Mary era
transactions.

Unfortunately this means a discontinuity with existing benchmark runs,
but should prevent future discontinuities.


lehins (approved):
I have one comment about benchmarks, which hasn't been touched by this PR, but it is important: 

`applyTxsTransition` returns a tuple, which means it is not enough to evaluate it to `whnf`, it has to be `nf` in this benchmark:

https://github.com/input-output-hk/cardano-ledger/blob/4e2d082d3a01d8a172d0659b9c6d7f56689570ca/libs/cardano-ledger-test/bench/Bench/Cardano/Ledger/ApplyTx.hs#L120-L127

If `NFData` instance is a problem at least replacing `id` with 
```haskell
\(x, y) -> x `seq` y `seq` (x, y)
```
 would be a good idea.

PR #2569, opened 2021-12-02
fix errata 17.2 in protocol version 7

### LINKED ISSUE #2477: Correct Shelly ledger 17.2
Fix the errata issue 17.2 of the Shelley spec using the `HardFork` protocol version mechanism.

JaredCorduan:
namely, no longer prefilter the rewards by those who are currently not registered at the start of the reward calculation. cc @ashisherc

closes #2477

### DIFF
--- a/eras/shelley/impl/src/Cardano/Ledger/Shelley/HardForks.hs
+++ b/eras/shelley/impl/src/Cardano/Ledger/Shelley/HardForks.hs
@@ -9,6 +9,7 @@ module Cardano.Ledger.Shelley.HardForks
     allowScriptStakeCredsToEarnRewards,
     translateTimeForPlutusScripts,
     missingScriptsSymmetricDifference,
+    forgoRewardPrefilter,
   )
 where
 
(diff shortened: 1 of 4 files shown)


eras/shelley/test-suite/test/Test/Cardano/Ledger/Shelley/Rewards.hs
```diff
@@ -511,7 +511,11 @@ rewardOnePool
           else Map.insert
       potentialRewards =
         f (getRwdCred $ _poolRAcnt pool) lReward mRewards
-      rewards' = Map.filter (/= Coin 0) $ eval (addrsRew ◁ potentialRewards)
+      potentialRewards' =
+        if HardForks.forgoRewardPrefilter pp
+          then potentialRewards
+          else eval (addrsRew ◁ potentialRewards)
```
JaredCorduan, line 517: reassuringly, the tests did not pass without this modification. :)

eras/shelley/impl/src/Cardano/Ledger/Shelley/HardForks.hs
```diff
@@ -65,3 +66,12 @@ missingScriptsSymmetricDifference ::
   pp ->
   Bool
 missingScriptsSymmetricDifference pp = pvMajor (getField @"_protocolVersion" pp) > 6
+
+-- | Starting with protocol version 7, the reward calculation no longer
+-- filters out unregistered stake addresses during the calculation.
```
nc6: During the calculation, or at the start of the calculation? I don't think we've ever had anything filter _during_, since that would have a different effect in the pulsed and un-pulsed version.

nc6 (changes requested):
Code looks fine, but one comment seems dodgy!

eras/shelley/impl/src/Cardano/Ledger/Shelley/HardForks.hs
JaredCorduan: good catch! We are filtering out addresses during the computation depending on whether they were registered at the start of the calculation. I'll update the comment, thank you!

nc6 (approved):

PR #2570, opened 2021-12-03
Update index state

newhoggy:


eras/byron/ledger/executable-spec/src/Byron/Spec/Ledger/Update.hs
```diff
@@ -232,9 +233,13 @@ data UProp = UProp
   }
   deriving (Eq, Generic, Show, Hashable, Data, Typeable, NoThunks)
 
+#if MIN_VERSION_hashable(1,3,4)
+-- Instance Hashable (Set a) is provided by the hashable package
+#else
 -- We need the Hashable instance before making lenses.
 instance Hashable a => Hashable (Set a) where
   hashWithSalt = H.hashUsing Set.toList
+#endif
```
newhoggy, line 242: Please pay special attention to the change in hash function to ensure this is safe to do.

lehins:
Thank you for doing this. This part can be removed in this PR as well:
https://github.com/input-output-hk/cardano-ledger/blob/4e2d082d3a01d8a172d0659b9c6d7f56689570ca/cabal.project#L106-L108

eras/byron/ledger/executable-spec/src/Byron/Spec/Ledger/Update.hs
lehins: I doubt that this change will cause problems, but I think we need to investigate it in a bit more detail. In particular someone who is familiar with how Hashable instances are being used in byron. This is important because hash value produced by the new hashable version is different. For example if it is used with `unordered-containers` then the order in which values are stored in the HashSet/HasMap will be different and it will affect things like folding and conversion to list:

* `hashable-1.3.4.1`
```haskell
Prelude> import Data.HashSet as HashSet
Prelude HashSet> HashSet.fromList ["foo", "bar", "baz"]
fromList ["bar","foo","baz"]
````
* `hashable-1.3.0.0`
```haskell
Prelude> import Data.HashSet as HashSet
Prelude HashSet> HashSet.fromList ["foo", "bar", "baz"]
fromList ["baz","foo","bar"]
```

I used `unordered-containers` just as an example, I don't think it is used directly in byron. Updating to the new `hashable` version does already change the outcome of `hashWithSalt` function for many types, so that in itself is worth investigating. The only thing about the particular change in this PR that I can find interesting is that hash of a set will not match the one of a matching list (i.e. `xs == Set.toList (Set.fromList xs)) ==> hash (Set.fromList xs) /= hash xs`), which is probably not a big of a deal at all considering that hash values for other types changed as well.

Here is the new version:

* `hashable-1.3.4.1`
```haskell
Prelude> import Data.Set as Set
Prelude Set> import Data.Hashable
Prelude Set Data.Hashable> let s = Set.fromList [1,2,3,4] :: Set Int
Prelude Set Data.Hashable> hashWithSalt (-3750763034362895579) s
6642510073022299387
Prelude Set Data.Hashable> let xs = [1,2,3,4] :: [Int]
Prelude Set Data.Hashable> hashWithSalt (-3750763034362895579) xs
3708964778940489643
```
Note that hash values for list and set differ. 

While this is the old behavior:
* `hashable-1.3.0.0`
```haskell
Prelude> import Data.Set as Set
Prelude Set> import Data.Hashable
Prelude Set Data.Hashable> let s = Set.fromList [1,2,3,4] :: Set Int
Prelude Set Data.Hashable> hashWithSalt (-3750763034362895579) s

<interactive>:4:1: error:
    • No instance for (Hashable (Set Int))
        arising from a use of ‘hashWithSalt’
    • In the expression: hashWithSalt (- 3750763034362895579) s
      In an equation for ‘it’:
          it = hashWithSalt (- 3750763034362895579) s
Prelude Set Data.Hashable> hashUsing Set.toList (-3750763034362895579) s
-6108972173346416629
Prelude Set Data.Hashable> let xs = [1,2,3,4] :: [Int]
Prelude Set Data.Hashable> hashWithSalt (-3750763034362895579) xs
-6108972173346416629
````
Naturally hash values for list and set are the same.

File: eras/byron/ledger/executable-spec/src/Byron/Spec/Ledger/Update.hs [hunk 2]
```diff
@@ -232,9 +233,13 @@ data UProp = UProp
   }
   deriving (Eq, Generic, Show, Hashable, Data, Typeable, NoThunks)
 
+#if MIN_VERSION_hashable(1,3,4)
+-- Instance Hashable (Set a) is provided by the hashable package
+#else
```
lehins, line 238:
```suggestion
-- Instance Hashable (Set a) is provided by the hashable package
#if !MIN_VERSION_hashable(1,3,4)
```

lehins (commented):
This is great. However, it would be nice to either:

* get someone who is familiar with byron codebase to invesigate the usage of Hashable to ensure this upgrade does not affect us. 
* or sync mainnet all the way from the beginning with this change.

eras/byron/ledger/executable-spec/src/Byron/Spec/Ledger/Update.hs
newhoggy: I put the comment inside the `if` because it has an implied:

"Instance Hashable (Set a) is provided by the hashable package from 1.3.4.0 onwards so we don't need to define the instance in this case"

Perhaps I should keep it in the `if` but make it more explicit?

lehins: Don't worry about it. It is not terribly important. If you really prefer the `if` comment `else` then let's keep it.

eras/byron/ledger/executable-spec/src/Byron/Spec/Ledger/Update.hs
nc6: I had the same comment. Since this is for Byron, I guess we can get away with just testing it out. If it syncs mainnet, that's all we really care about.

I'm pretty certain the hashes aren't used in anywhere important.

newhoggy:
Verified that `mainnet` syncs with `cardano-ledger` update here: https://github.com/input-output-hk/cardano-node/pull/3418

```
$ uname -a
Darwin johns-mbp.lan 21.1.0 Darwin Kernel Version 21.1.0: Wed Oct 13 17:33:01 PDT 2021; root:xnu-8019.41.5~1/RELEASE_ARM64_T6000 arm64
```

```
$ CARDANO_NODE_SOCKET_PATH=local-node/socket cardano-cli query tip --mainnet
{
    "epoch": 306,
    "era": "Alonzo",
    "slot": 47142155,
    "syncProgress": "100.00",
    "hash": "d66d4a38f833ff6234615e89f5151fe0b7826424865f12f8e92d47fc403dac39",
    "block": 6588621
}
```

lehins (approved):

nc6 (approved):
LGTM. Ledger requires signed commits - if you're not set up for this, we can adopt the patch and resign it.

newhoggy:
Pushed a signed commit.

PR #2573, opened 2021-12-03
Benchmark custom maps

lehins:
This PR adds benchmarks, applies hlint and some renameing for consistency.

Most importantly it also improves speed of insertion into the KeyMap and fixes a bug in `filterArrayWithBitmap` that resulted in `intersection` throwing an exception on some input.

### DIFF
--- a/libs/compact-map/compact-map.cabal
+++ b/libs/compact-map/compact-map.cabal
@@ -43,6 +43,7 @@ library
                      , deepseq
                      , prettyprinter
                      , primitive
+                     , random
                      , text
                      , nothunks
                      , vector
@@ -71,3 +72,16 @@ test-suite tests
                      , quickcheck-classes-base
                      , random
   ghc-options:        -threaded -O
+
+benchmark bench
+  type:                exitcode-stdio-1.0
+  hs-source-dirs:      bench
+  main-is:             Bench.hs
+  ghc-options:         -Wall -threaded -O2 -rtsopts
+  build-depends:       base
+                     , criterion
+                     , compact-map
+                     , containers
+                     , random
+  default-language:    Haskell2010
+
--- a/libs/compact-map/compact-map.cabal
+++ b/libs/compact-map/compact-map.cabal
@@ -71,7 +71,7 @@ test-suite tests
                      , QuickCheck
                      , quickcheck-classes-base
                      , random
-  ghc-options:        -threaded -O
+  ghc-options:        -threaded
 
 benchmark bench
   type:                exitcode-stdio-1.0
--- a/libs/compact-map/src/Data/Compact/KeyMap.hs
+++ b/libs/compact-map/src/Data/Compact/KeyMap.hs
@@ -46,6 +46,7 @@ import Data.Word (Word64)
 import GHC.Exts (isTrue#, reallyUnsafePtrEquality#, (==#))
 import Prettyprinter
 import qualified Prettyprinter.Internal as Pretty
+import System.Random.Stateful (Uniform (..))
 
 -- =============================
 
@@ -105,6 +106,15 @@ data Key
       {-# UNPACK #-} !Word64
       {-# UNPACK #-} !Word64
   deriving (Eq, Ord, Show, NFData, Generic)
+
+
+instance Uniform Key where
+  uniformM g = do
+    w0 <- uniformM g
+    w1 <- uniformM g
+    w2 <- uniformM g
+    w3 <- uniformM g
+    pure (Key w0 w1 w2 w3)
 
 -- | The number of Word64 per key
 wordsPerKey :: Int
--- a/libs/compact-map/src/Data/Compact/KeyMap.hs
+++ b/libs/compact-map/src/Data/Compact/KeyMap.hs
@@ -418,10 +418,10 @@ filterArrayWithBitmap p bm0 arr =
     -- i ranges over all possible elements of a Bitmap [0..63], only some are found in 'bm'
     -- j ranges over the slots in the new array [0..n-1]
     loop i j bm marr
-      | i < 63 && not (testBit bm0 i) =
+      | i <= 63 && not (testBit bm0 i) =
         loop (i + 1) j bm marr -- Skip over those not in 'bm'
     loop i j bm marr
-      | i < 63 =
+      | i <= 63 =
         let slot = indexFromSegment bm0 i -- what is the index in 'arr' for this Bitmap element?
             item = index arr slot -- Get the array item
          in if not (p item) -- if it does not meet the 'p' then move it to the answer.
@@ -429,7 +429,7 @@ filterArrayWithBitmap p bm0 arr =
               else -- if it meets 'p' then don't copy, and clear it from 'bm'
                 loop (i + 1) j (clearBit bm i) marr
     loop _i j _bm _marr
-      | j /= n = error $ "Left over plank space at. j= " ++ show j ++ ", n= " ++ show n
+      | j /= n = error $ "Left over blank space at. j= " ++ show j ++ ", n= " ++ show n
     loop _i _j bm _marr = pure bm
 
 -- ================================================================
--- a/libs/compact-map/src/Data/Compact/KeyMap.hs
+++ b/libs/compact-map/src/Data/Compact/KeyMap.hs
@@ -2,12 +2,53 @@
 {-# LANGUAGE DeriveAnyClass #-}
 {-# LANGUAGE DeriveGeneric #-}
 {-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE MagicHash #-}
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE UnboxedTuples #-}
 {-# OPTIONS_GHC -Wno-orphans #-}
 
-module Data.Compact.KeyMap where
+module Data.Compact.KeyMap
+  ( KeyMap(..),
+    Key (..),
+    empty,
+    size,
+    lookup,
+    lookupMax,
+    lookupMin,
+    splitLookup,
+    insert,
+    insertWith,
+    insertWithKey,
+    delete,
+    mapWithKey,
+    restrictKeys,
+    withoutKeys,
+    intersection,
+    intersectionWith,
+    intersectionWithKey,
+    union,
+    unionWith,
+    unionWithKey,
+    foldWithAscKey,
+    foldWithDescKey,
+    fromList,
+    toList,
+    lub,
+    maxViewWithKey,
+    minViewWithKey,
+    foldOverIntersection,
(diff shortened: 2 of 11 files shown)


libs/compact-map/src/Data/Compact/KeyMap.hs
```diff
@@ -71,28 +106,32 @@ bitsPerSegment :: Int
...
-type Path = [Segment]
-
--- | The maximum value of a Segment, as an Int
-intSize :: Int
-intSize = 2 ^ bitsPerSegment
-{-# INLINE intSize #-}
+type Path = VP.Vector Segment
 
 -- | The maximum value of a segment, as a Word64
 wordSize :: Word64
```
nc6: A little confusingly named! But I guess it's not exported

nc6 (approved):

libs/compact-map/src/Data/Compact/KeyMap.hs
lehins: Very good point. I've renamed it to `segmentMaxValue`

libs/compact-map/compact-map.cabal
```diff
@@ -68,4 +69,17 @@ test-suite tests
                      , QuickCheck
                      , quickcheck-classes-base
                      , random
-  ghc-options:        -threaded -O
+  ghc-options:        -threaded
+
```
TimSheard, line 75: I would have thought that one would want to have used compiler flag -O on a benchmark?

libs/compact-map/src/Data/Compact/KeyMap.hs
```diff
@@ -35,29 +79,21 @@ import Data.Compact.SmallArray
...
-
--- | binary encoding of 'n', LEAST significant bit on the front of the list.
-binary :: Integral n => n -> [n]
-binary 0 = []
-binary 1 = [1]
-binary n = mod n 2 : binary (div n 2)
-
--- | Show 'n' as a binary number with MOST significant bits on the front of the list.
-bin :: Integral n => n -> [n]
-bin x = reverse (binary x)
```
TimSheard, line 60: Pretty sure there were some error messages that used bin and binary to show BitMaps, showing them in Binary is much more illustrative than just printing a BitMap as an Int.  I think we should restore that functionality.

libs/compact-map/src/Data/Compact/KeyMap.hs
```diff
@@ -248,67 +297,86 @@ insertWithKey' n0 combine path k x kmap = go n0 kmap
...
-    if i < j
-      then Two (setBits [i, j]) leaf1 (Leaf k2 v2)
-      else Two (setBits [i, j]) (Leaf k2 v2) leaf1
+twoLeaf path1 leaf1 path2 k2 v2 = go path1 path2
+  where
+    leaf2 = Leaf k2 v2
+    go p1 p2
+      | Just (i, is) <- VP.uncons p1,
+        Just (j, js) <- VP.uncons p2 =
+        if i == j
```
TimSheard, line 327: There seems to be some tension between thinking of Path as a list or as an array. Sometimes I think a list is better (like here, where we are pattern matching) and other times an array (when we need something like (path !! n))   The 'n' is usually needed to handle the (Leaf __ ) case because unlike all the other constructors  Leaf's do not track which segement they are under. So we do some adustment to the Path, to recover this lost information.

libs/compact-map/src/Data/Compact/KeyMap.hs
```diff
@@ -248,67 +297,86 @@ insertWithKey' n0 combine path k x kmap = go n0 kmap
...
+insert !k = insertWithKey' 0 (\_key new _old -> new) (keyPath k) k
+{-# INLINE insert #-}
 
 fromList :: [(Key, v)] -> KeyMap v
-fromList ps = foldl' accum Empty ps
+fromList = foldl' accum Empty
   where
-    accum ans (k, v) = insert k v ans
+    accum !ans (k, v) = insert k v ans
+{-# INLINE fromList #-}
```
TimSheard, line 361: Isn't fold'  strict in the accumulator argument (that is what the ' is used for), so is the ! on ans needed?

libs/compact-map/src/Data/Compact/KeyMap.hs
```diff
@@ -248,67 +297,86 @@ insertWithKey' n0 combine path k x kmap = go n0 kmap
...
     accum k v ans = (k, v) : ans
 
 -- =================================================================
 -- Deletion
 
-delete2 :: Path -> Key -> KeyMap v -> (KeyMap v -> KeyMap v) -> KeyMap v
-delete2 path key km continue = case3 (continue Empty) leafF arrayF km
+deleteInternal :: Path -> Key -> KeyMap v -> (KeyMap v -> KeyMap v) -> KeyMap v
+deleteInternal path key km continue = case3 (continue Empty) leafF arrayF km
   where
```
TimSheard, line 372: to be consistent with insert', shouldn't we call this delete'

libs/compact-map/src/Data/Compact/KeyMap.hs
```diff
@@ -417,60 +496,49 @@ foldWithDescKey accum !ans0 (Full arr) = loop ans0 (n - 1)
...
-        else Nothing
-      where
-        i = indexFromSegment bm j
-    go (j : js) (Full arr) =
-      -- Every possible bit is set, so no testBit call necessary
-      go js (index arr i)
-      where
-        i = indexFromSegment fullNodeMask j
+lookup :: Key -> KeyMap v -> Maybe v
+lookup key = searchPath key (keyPath key)
```
TimSheard, line 509: Good to catch that searchPath and lookup were very similar

libs/compact-map/src/Data/Compact/KeyMap.hs
```diff
@@ -484,22 +552,25 @@ instance Functor KeyMap where
...
-array2 x y = fst (withMutArray 2 (\marr -> mwrite marr 0 x >> mwrite marr 1 y))
+array2 x y = withMutArray_ 2 (\marr -> mwrite marr 0 x >> mwrite marr 1 y)
+{-# INLINE array2 #-}
 
-union4 :: Int -> (Key -> v -> v -> v) -> KeyMap v -> KeyMap v -> KeyMap v
-union4 _n _combine Empty Empty = Empty
-union4 n combine x y = case3 emptyC1 leafF1 arrayF1 x
+unionInternal :: Int -> (Key -> v -> v -> v) -> KeyMap v -> KeyMap v -> KeyMap v
+unionInternal _n _combine Empty Empty = Empty
+unionInternal n combine x y = case3 emptyC1 leafF1 arrayF1 x
```
TimSheard, line 573: union' ?

libs/compact-map/src/Data/Compact/KeyMap.hs
```diff
@@ -508,101 +579,111 @@ union4 n combine x y = case3 emptyC1 leafF1 arrayF1 x
...
 
 -- ===========================================
 -- intersection operators
 
-intersect3 :: Int -> (Key -> u -> v -> w) -> KeyMap u -> KeyMap v -> KeyMap w
-intersect3 _ _ Empty Empty = Empty
-intersect3 n combine x y = case3 Empty leafF1 arrayF1 x
+intersectInternal :: Int -> (Key -> u -> v -> w) -> KeyMap u -> KeyMap v -> KeyMap w
+intersectInternal _ _ Empty Empty = Empty
+intersectInternal n combine x y = case3 Empty leafF1 arrayF1 x
```
TimSheard: intersect', or maybe use insertInternal instead of insert', so all use the same naming convention.

TimSheard (commented):
Cleans up quite a bit of loose ends, so a good job. I have some concerns about using arrays for Path in every function.  keyPath computes the list lazily in in most functions, if we use a list, on most calls we will look at only the first 6-8 items in the path, while the full path may contain 32+ items. So laziness is our friend.  There is 1 exception to the rule in insertWithKey!  where we need to access particular locations in the path with VP.!.  We might have two kinds of paths, one an array for use in insertWithKey' and one a lazy list for every where else a path is used. 

We ought to talk about this.

libs/compact-map/compact-map.cabal
lehins: `-O` is always supplied by default in cabal projects. It is only needed when it is desired to disable optimizations with `-O0` or increase the level i.e. `-O2`

libs/compact-map/src/Data/Compact/KeyMap.hs
lehins: I would not have been able to remove these functions if they were used somewhere. It was really dead code.

libs/compact-map/src/Data/Compact/KeyMap.hs
lehins: I understand your reasoning, but I was using benchmarks as a guide, rather than logical reasoning. In the end what we think the performance should be is not necessarily the performance that we'll get. Using lists for Path makes the worst case 5 times slower than when using primitive vector.

We can certainly discuss it in more detail. If you can come up with the worst case example where vector would perform worse that would be tremendously useful, because then we could benchmark it and reason about the performance hit if any. One way or another `insert` is one of the most important operations in this map, and that is the one that I've optimized.

libs/compact-map/src/Data/Compact/KeyMap.hs
lehins: It isn't technically necessary to fix memory leaks, but there is a subtle difference between just `foldl'` and `foldl'` with bang on accumulator: https://www.well-typed.com/blog/2014/04/fixing-foldl/#a-postscript-which-foldl

With all honesty I added it because I was benchmarking the function and I was tapping into all possible optimizations I could find. So this change is not strictly necessary (pun intended), but it won't hurt either.

libs/compact-map/src/Data/Compact/KeyMap.hs
lehins: Good point, consistency is our friend. I'll rename `insertWithKey'` -> `insertWithKeyInternal`

I don't like ticks in top level functions since they don't carry information, except in strict folds, which was a poor decision on itself IMHO, but that ship has sailed for Haskell.

libs/compact-map/src/Data/Compact/KeyMap.hs
lehins: Yep, that was my thinking too :) Done

libs/compact-map/src/Data/Compact/KeyMap.hs
lehins: switched insert to insertInternal

libs/compact-map/src/Data/Compact/KeyMap.hs
lehins: Switched to builtin function for showing numbers as binary and added it to error message

PR #2575, opened 2021-12-04
remove final dependencies on tpraos

### LINKED ISSUE #2462: Remove dependency on cardando-protocol-tpraos

JaredCorduan:
This PR finishes the task of removing the dependency of `cardano-protocol-tpraos` in `cardano-ledger-shelley`. In fact, `cardano-protocol-tpraos` now depends on `cardano-ledger-shelley`.

- `Cardano/Ledger/Shelley/API/Protocol.hs` was moved to `Cardano/Protocol/TPraos/API.hs`
- the constraint `PraosCrypto` was removed from `ShelleyBasedEra`.
- `mkInitialShelleyLedgerView`, `GetLedgerView`, and `PraosCrypto` were moved to `Cardano.Protocol.TPraos.API`
- all the TPraos types in `Cardano/Ledger/Shelley/API/Types.hs` are removed

closes #2462


eras/shelley/impl/src/Cardano/Ledger/Shelley/API/Mempool.hs
```diff
@@ -183,7 +183,25 @@ class
...
+      (Core.EraRule "LEDGER" (ShelleyEra c))
+      ~ LedgerPredicateFailure (ShelleyEra c),
+    STS (Core.EraRule "LEDGER" (ShelleyEra c)),
+    FromCBOR (PredicateFailure (Core.EraRule "DELEGS" (ShelleyEra c))),
+    Show (PredicateFailure (Core.EraRule "DELEGS" (ShelleyEra c))),
+    FromCBOR (PredicateFailure (Core.EraRule "UTXOW" (ShelleyEra c))),
+    ToCBOR (PredicateFailure (Core.EraRule "DELEGS" (ShelleyEra c))),
+    ToCBOR (PredicateFailure (Core.EraRule "UTXOW" (ShelleyEra c)))
+  ) =>
+  ApplyTx (ShelleyEra c)
```
JaredCorduan: Do we like the `ApplyTx` remaining here, but changing the `PraosCrypto` constraint? I would love to understand why removing `PraosCrypto` forced all the constraints above, they do not seem to be covered by `PraosCrypto`.

libs/cardano-protocol-tpraos/src/Cardano/Protocol/TPraos/API.hs
```diff
@@ -516,3 +529,67 @@ getLeaderSchedule globals ss cds poolHash key pp = Set.filter isLeader epochSlot
...
+mkInitialShelleyLedgerView genesisShelley =
+  LedgerView
+    { lvD = Shelley._d . sgProtocolParams $ genesisShelley,
+      lvExtraEntropy = Shelley._extraEntropy . sgProtocolParams $ genesisShelley,
+      lvPoolDistr = PoolDistr Map.empty,
+      lvGenDelegs = GenDelegs $ sgGenDelegs genesisShelley,
+      lvChainChecks = pparamsToChainChecksPParams . sgProtocolParams $ genesisShelley
+    }
+
+instance PraosCrypto c => ApplyTx (AllegraEra c)
```
JaredCorduan: Unlike the shelley instance, I moved the other era instances here, and kept the `PraosCrypto` constraint. Do we like this better?

nc6: No, this definitely seems wrong. `ApplyTx` has nothing to do with the protocol.

eras/shelley-ma/impl/src/Cardano/Ledger/Allegra.hs
```diff
@@ -39,12 +39,6 @@ import qualified Data.Map.Strict as Map
 
 type AllegraEra = ShelleyMAEra 'Allegra
 
-instance PraosCrypto c => ApplyTx (AllegraEra c)
-
-instance PraosCrypto c => ApplyBlock (AllegraEra c)
-
-instance PraosCrypto c => GetLedgerView (AllegraEra c)
```
nc6, line 46: This one's a bit annoying. I am _tempted_ to say the correct thing to do with this is drop it entirely, and move it into consensus (in the way that I am doing with Praos). That would more accurately reflect its nature as an "integration" piece, rather than part of either ledger or protocol.

eras/shelley-ma/impl/src/Cardano/Ledger/Allegra.hs
```diff
@@ -39,12 +39,6 @@ import qualified Data.Map.Strict as Map
 
 type AllegraEra = ShelleyMAEra 'Allegra
 
-instance PraosCrypto c => ApplyTx (AllegraEra c)
-
-instance PraosCrypto c => ApplyBlock (AllegraEra c)
```
nc6, line 44: Both `ApplyTx` and `ApplyBlock` are pure ledger concerns, right? `Block` actually calls `BBODY`. So these should stay ledger related.

eras/shelley/impl/src/Cardano/Ledger/Shelley/API.hs
```diff
@@ -6,49 +6,12 @@
...
-    UsesAuxiliary era,
-    UsesTxBody era,
-    UsesTxOut era,
-    UsesPParams era,
-    ChainData (State (Core.EraRule "PPUP" era)),
-    SerialisableData (State (Core.EraRule "PPUP" era)),
-    Share (Core.TxOut era) ~ Interns (Credential 'Staking (Crypto era)),
-    FromSharedCBOR (Core.TxOut era)
-  ) =>
-  ShelleyBasedEra era
```
nc6, line 52: I think we should split this constraint block in two, rather than moving it. For one thing, as it is right now, it will not be satisfied once we move to Babbage anyway.

eras/shelley/impl/src/Cardano/Ledger/Shelley/API/Mempool.hs
nc6: Odd

nc6 (changes requested):
Maybe let's discuss this in a call!

JaredCorduan:
@nc6 The plan worked out well!

nc6 (approved):
LGTM!

PR #2578, opened 2021-12-07
Fix nix-shell and nix-build

nc6:
- Alters the nix configuration so as to actually make use of the
  separate hackage.nix pin.
- Update the hackage.nix and haskell.nix versions sufficiently for the
  new index-state. We bump haskell.nix only to the same version as
  cardano-node, to take advantage of the cache.

### DIFF
--- a/eras/shelley/test-suite/test/Golden/ShelleyGenesis
+++ b/eras/shelley/test-suite/test/Golden/ShelleyGenesis
@@ -1,70 +1,61 @@
 {
-    "activeSlotsCoeff": 0.259,
+    "maxKESEvolutions": 28899,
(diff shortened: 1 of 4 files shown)


lehins:
@nc6 Since you updating nix config can you also update GHC to 8.10.7 in this PR?

https://github.com/input-output-hk/cardano-ledger/blob/7155301bcddc9cc2efe9cd98ad9c2521e9a0d085/nix/haskell.nix#L11

nc6:
I can, though this is ultimately something we should coordinate across the various components.

JaredCorduan (approved):
thanks @nc6 !

PR #2579, opened 2021-12-07
add MIR changes to Alonzo spec

JaredCorduan:
Additionally, added a new check to the Alonzo MIR rule that ensures that the Coin valued used in transfers (reserves <-> treasury) is always non-negative. This does mean that the predicate serialization has changed, but as this is a governance feature we should be fine (cc @nc6).

### DIFF
--- a/eras/alonzo/formal-spec/mir.tex
+++ b/eras/alonzo/formal-spec/mir.tex
@@ -1,0 +1,288 @@
+\section{MIR Certificates}
+\label{sec:mir}
+
+\newcommand{\InstantaneousRewards}{\type{InstantaneousRewards}}
+\newcommand{\MIRPot}{\type{MIRPot}}
+\newcommand{\ReservesMIR}{\type{ReservesMIR}}
+\newcommand{\TreasuryMIR}{\type{TreasuryMIR}}
+
+In the Alonzo era, MIR certs have:
+
+\begin{itemize}
+  \item Slightly different behavior -
+  prior to the Alonzo era, MIR certificates within the same epoch override
+  values for repeated stake addresses.
+  In the Alonzo era, repetitions are handled by adding the values.
+  Moreover, negative values are allowed, so long as the change never results in
+  an overall negative sum.
+  See Figure \ref{fig:dcert-mir}.
+  \item Extended functionality -
+  MIR certificates can now be used to transfer funds between the reserves and the treasury.
+  See Figure \ref{fig:dcert-mir-trans}.
+\end{itemize}
+
+Figure \ref{fig:defs:alonzo-mir} describes the necessary changes to the types.
+Note that $\fun{mirTarget}$ was named $\fun{credCoinMap}$ in the Shelley specification.
+Figure \ref{fig:dcert-mir} is a modification to the DELEG rule, and
+Figure \ref{fig:dcert-mir-trans} is an addition to the DELEG rule.
+Lastly, Figure \ref{fig:rules:mir} is the updated MIR rule.
+
+
+\begin{figure*}[htb]
+  \emph{Derived types}
+  %
+  \begin{equation*}
+    \begin{array}{r@{~~}l@{~~}l@{\qquad}l}
+      \InstantaneousRewards =
+      & (\StakeCredential \mapsto \Coin)
+      & \text{rewards from reserves}
+      \\
+      & \times (\StakeCredential \mapsto \Coin)
+      & \text{rewards from treasury}
+      \\
+      & \times \hldiff{\Coin}
+      & \text{pending modification to reserves}
+      \\
+      & \times \hldiff{\Coin}
+      & \text{pending modification to treasury}
+    \end{array}
+  \end{equation*}
+  %
+  \emph{New MIR Cert accessor}
+  %
+  \begin{equation*}
+    \begin{array}{r@{~:~}lr}
+      \fun{mirTarget}
+      & \DCertMir \to (\StakeCredential \mapsto \Coin) \uniondistinct \hldiff{\MIRPot}
+    \end{array}
+  \end{equation*}
+  \caption{Alonzo Types for MIR updates}
+  \label{fig:defs:alonzo-mir}
+\end{figure*}
+
+\begin{figure}[htp]
+  \centering
+  \begin{equation}\label{eq:deleg-mir}
+    \inference[Deleg-Mir]
+    {
+      \var{c}\in \DCertMir
+      &
+      \hldiff{\fun{mirTarget}~\var{c}\in \StakeCredential \mapsto \Coin}
+      \\
+      slot < \fun{firstSlot}~((\epoch{slot}) + 1) - \fun{StabilityWindow}\\
+      (\var{irR},~\var{irT}\hldiff{,~\var{dR},~\var{dT}})\leteq\var{i_{rwd}}
+      &
+      (\var{treasury},~\var{reserves})\leteq\var{acnt}
+      \\
+      (\var{pot}\hldiff{,~\var{deltaPot}},~\var{irPot})\leteq
+      {\begin{cases}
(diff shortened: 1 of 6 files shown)


eras/alonzo/formal-spec/mir.tex
```diff
@@ -0,0 +1,415 @@
...
+Figure \ref{fig:dcert-mir-trans} is an addition to the DELEG rule.
+Lastly, Figure \ref{fig:rules:mir} is the updated MIR rule.
+
+
+\begin{figure*}[htb]
+  \emph{Derived types}
+  %
+  \begin{equation*}
+    \begin{array}{llr}
+      \InstantaneousRewards =
```
WhatisRT, line 36: Spacing for the `=` sign is wrong, and I think for `\times` as well

eras/alonzo/formal-spec/mir.tex
```diff
@@ -0,0 +1,415 @@
...
+        \varUpdate{\var{i_{rwd}'}} \\
+      \end{array}
+      \right)
+    }
+  \end{equation}
+  \caption{MIR transfer Inference Rule}
+  \label{fig:dcert-mir-trans}
+\end{figure}
+
+\begin{figure}[ht]
```
WhatisRT, line 205: Not sure if it's worth it, but we could get rid of the code duplication in MIR by having a single rule for it and defining an `acnt'` to be either of the two options depending on the condition.

eras/alonzo/formal-spec/mir.tex
```diff
@@ -0,0 +1,415 @@
...
+    \begin{array}{r@{~:~}lr}
+      \fun{mirTarget}
+      & \DCertMir \to (\StakeCredential \mapsto \Coin) \uniondistinct \hldiff{\MIRPot}
+    \end{array}
+  \end{equation*}
+  \caption{Alonzo Types for MIR updates}
+  \label{fig:defs:alonzo-mir}
+\end{figure*}
+
+\begin{figure}[htp]
```
WhatisRT, line 63: There's a lot of duplication in this and the next figure. I see two options of simplifying them:
1. Define `i'_rwd` depending on the two cases for `mirPot c` and drop the condition. This would give us two rules with still some duplication, but a lot less than now.
2. Make a new STS that just transitions the state of `i_rwd`. Then there would only be a single rule for `DELEG` that handles all the common stuff and the new STS handles all the special cases. This would be more complicated but not have any duplication. Probably not worth it, except if we expect more changes in the future.

JaredCorduan: yea, #1 as actually how the code works, but I wasn't sure how to make it readable in the spec.

https://github.com/input-output-hk/cardano-ledger/blob/fdcb83f35f58d0320f8a0bcf6c1ae54395bb984c/eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Deleg.hs#L328-L335

eras/alonzo/formal-spec/mir.tex
JaredCorduan: that's true, as I mention in the comment above, that is actually how the code works.

eras/alonzo/formal-spec/mir.tex
JaredCorduan: option 2 would be too different from the implementation for my comfort...

Let me think about how to do 1 nicely in latex w/o programming jargon :thinking:

WhatisRT: Let me recommend the `cases` environment :)

eras/alonzo/formal-spec/mir.tex
JaredCorduan: yea, it's pretty ugly. I just tried using the array settings used in `TxWitness`, which does look better, but still a bit off.

eras/alonzo/formal-spec/mir.tex
JaredCorduan: in a let binding? or use helper functions?

JaredCorduan: ok @WhatisRT , I've tried out using cases, let me know what you think. it's nice that it actually matches the code a lot better now.

WhatisRT: Yeah, this is pretty much what I had in mind

eras/alonzo/formal-spec/mir.tex
```diff
@@ -0,0 +1,331 @@
...
+      \\
+      \var{combined}\leteq(\fun{mirTarget}~\var{c})\hldiff{\unionoverridePlus}\var{irPot}
+      \\
+      \sum\limits_{\wcard\mapsto\var{val}\in\var{combined}} val \leq\var{pot} \hldiff{+ \var{deltaPot}}
+      \\
+      \hldiff{\forall \var{r}\in\range{\var{combinedR}},~r\geq 0}
+      \\
+      \var{i_{rwd}'}\leteq
+      {\begin{cases}
+      (\var{combined},~\var{irT},~\var{dR},~\var{dT}) \\
```
WhatisRT: The condition for the two cases is missing here

eras/alonzo/formal-spec/mir.tex
```diff
@@ -0,0 +1,331 @@
...
+          \left(\sum\limits_{\wcard\mapsto\var{val}\in\var{irT}} val\right)
+          & \fun{mirPot}~\var{c}=\TreasuryMIR
+       \end{cases}
+      }
+      \\
+      \var{coin} \leq \var{available}
+      \\
+      \var{i_{rwd}'}
+      {\begin{cases}
+         \leteq(\var{irR},~\var{irT},~\var{dR}-\var{coin},~\var{dT}+\var{coin})
```
WhatisRT: `\leteq` should be outside of the `cases`

eras/alonzo/formal-spec/mir.tex
```diff
@@ -0,0 +1,331 @@
...
+\begin{figure}[htp]
+  \centering
+  \begin{equation}\label{eq:deleg-mir-r-to-t}
+    \inference[Deleg-Mir-Trans]
+    {
+      \var{c}\in \DCertMir
+      &
+      \fun{mirTarget}~\var{c}\in \Coin
+      &
+      \fun{mirPot}~\var{c}=\ReservesMIR
```
WhatisRT: This condition needs to be dropped

eras/alonzo/formal-spec/mir.tex
WhatisRT: I think it gets the seal of 'good enough' :)

JaredCorduan: :sweat_smile:

eras/alonzo/formal-spec/mir.tex
JaredCorduan: yes, thank you!

eras/alonzo/formal-spec/mir.tex
JaredCorduan: yes! thank you.

eras/alonzo/formal-spec/mir.tex
JaredCorduan: most definitely, thank you

eras/alonzo/formal-spec/mir.tex
JaredCorduan: I forgot to attempt this after the first review, I've done it now, see what you think. It was a bit hard to keep the horizontal space within the table, but I think it looks okay-ish. :)

JaredCorduan: I should probably make these "reduce copy" changes to the shelley spec as well now...

JaredCorduan: ok, I've changed the Shelley MIR transitions as well now. The compare nicely now.

JaredCorduan:
I somehow broke the latex build, I will fix that tomorrow

WhatisRT (approved):
LGTM!

PR #2583, opened 2021-12-09
Lehins/attempt to fix regression

lehins:


JaredCorduan (approved):
That did the trick! :raised_hands: 

(I benchmarked with https://github.com/input-output-hk/cardano-ledger/pull/2583/commits/dc60dbc78d9332a70855e5614460a7e6131b41c0 by itself)

JaredCorduan (changes requested):
There's a problem with the optimization in `sumStakePerPool`. ie commit https://github.com/input-output-hk/cardano-ledger/pull/2583/commits/acff545b1dd8af11c7ae1c2b2b810e5db58b6f67

It is computing a different result, and prevent the node from syncing.

lehins:
> There's a problem with the optimization in sumStakePool. ie commit acff545

Dropped the commit, I'll investigate it in a separate PR. For now this PR fixes the regression and that was the priority.

JaredCorduan (approved):

nc6 (approved):
Nice catch!

PR #2584, opened 2021-12-10
Ts mono unifiedmaps

TimSheard:
Introduces UnifiedMap, ViewMap, Triple
Adds SetAlgebra Instances
Adds FromSharedCBOR instances
Fixes files so the new types are actually used

### DIFF
--- a/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Rules/Ledger.hs
+++ b/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Rules/Ledger.hs
@@ -30,6 +30,7 @@ import Cardano.Ledger.Shelley.LedgerState
     DState (..),
     PState (..),
     UTxOState (..),
+    rewards,
   )
 import Cardano.Ledger.Shelley.Rules.Delegs (DELEGS, DelegsEnv (..), DelegsEvent, DelegsPredicateFailure)
 import Cardano.Ledger.Shelley.Rules.Ledger (LedgerEnv (..), LedgerEvent (..), LedgerPredicateFailure (..))
@@ -156,7 +157,7 @@ instance
         "Deposit pot must equal obligation"
         ( \(TRC (LedgerEnv {ledgerPp}, _, _))
            (utxoSt, DPState {_dstate, _pstate}) ->
-              obligation ledgerPp (_rewards _dstate) (_pParams _pstate)
+              obligation ledgerPp (rewards _dstate) (_pParams _pstate)
                 == _deposited utxoSt
         )
     ]
--- a/eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState.hs
+++ b/eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState.hs
@@ -31,6 +31,9 @@ module Cardano.Ledger.Shelley.LedgerState
     DPState (..),
     DState (..),
     emptyDState,
+    rewards,
+    delegations,
+    ptrsMap,
     EpochState (..),
     UpecState (..),
     PulsingRewUpdate (..),
@@ -235,15 +238,15 @@ import Cardano.Ledger.Slot
     SlotNo (..),
   )
 import Cardano.Ledger.TxIn (TxIn (..))
+import Cardano.Ledger.UnifiedMap (Trip (..), Triple, UMap (..), UnifiedMap, View (..), ViewMap)
 import Cardano.Ledger.Val ((<+>), (<->), (<×>))
 import qualified Cardano.Ledger.Val as Val
 import Cardano.Prelude (rightToMaybe)
 import Control.DeepSeq (NFData)
-import Control.Monad.State.Strict (evalStateT)
+import Control.Monad.State.Strict (StateT (..), evalStateT)
 import Control.Monad.Trans
-import Control.Monad.Trans.State.Strict (StateT)
 import Control.Provenance (ProvM, liftProv, modifyM)
-import Control.SetAlgebra (Bimap, biMapEmpty, dom, eval, forwards, (∈), (∪+), (▷), (◁))
+import Control.SetAlgebra (dom, eval, (∈), (▷), (◁))
 import Control.State.Transition (STS (State))
 import Data.Coders
   ( Annotator (..),
@@ -255,7 +258,7 @@ import Data.Coders
     (<!),
     (<*!),
   )
-import qualified Data.Compact.VMap as VMap
+import qualified Data.Compact.ViewMap as VMap
 import Data.Constraint (Constraint)
 import Data.Default.Class (Default, def)
 import Data.Foldable (fold, toList)
@@ -265,6 +268,7 @@ import Data.Group (Group, invert)
 import Data.Kind (Type)
 import Data.Map.Strict (Map)
 import qualified Data.Map.Strict as Map
+import Data.MapExtras (filterMaybe)
 import Data.Pulse (Pulsable (..), completeM)
 import Data.Ratio ((%))
 import Data.Sequence.Strict (StrictSeq)
@@ -272,6 +276,7 @@ import Data.Set (Set)
 import qualified Data.Set as Set
 import Data.Sharing
 import Data.Typeable
+import qualified Data.UMap as UM
 import Data.Word (Word64)
 import GHC.Generics (Generic)
 import GHC.Records (HasField (..))
@@ -310,7 +315,7 @@ instance CC.Crypto crypto => FromCBOR (FutureGenDeleg crypto) where
 -- the rewards which will be paid out from the reserves and the rewards
 -- which will be paid out from the treasury. It also consists of
 -- two coin values which represent the transfer of coins from
--- one pot the the other pot.
+-- one pot to the other pot.
 -- NOTE that the following property should always hold:
 --   deltaReserves + deltaTreasury = 0
 data InstantaneousRewards crypto = InstantaneousRewards
@@ -350,12 +355,8 @@ instance CC.Crypto crypto => FromSharedCBOR (InstantaneousRewards crypto) where
 
 -- | State of staking pool delegations and rewards
 data DState crypto = DState
-  { -- | The active reward accounts.
-    _rewards :: !(RewardAccounts crypto),
-    -- | The current delegations.
-    _delegations :: !(Map (Credential 'Staking crypto) (KeyHash 'StakePool crypto)),
-    -- | The pointed to hash keys.
-    _ptrs :: !(Bimap Ptr (Credential 'Staking crypto)),
+  { -- | Unified Reward Maps
+    _unified :: UnifiedMap crypto,
     -- | Future genesis key delegations
     _fGenDelegs :: !(Map (FutureGenDeleg crypto) (GenDelegPair crypto)),
     -- | Genesis key delegations
@@ -365,16 +366,30 @@ data DState crypto = DState
   }
   deriving (Show, Eq, Generic)
 
+-- ========================
+-- Virtual selectors, which get the appropriate view from a DState from the embedded UnifiedMap
+
+rewards :: DState crypto -> ViewMap crypto (Credential 'Staking crypto) Coin
+rewards (DState unified _ _ _) = Rewards unified
+
+delegations :: DState crypto -> ViewMap crypto (Credential 'Staking crypto) (KeyHash 'StakePool crypto)
+delegations (DState unified _ _ _) = Delegations unified
+
+-- | get the actual ptrs map, we don't need a view
+ptrsMap :: DState crypto -> Map Ptr (Credential 'Staking crypto)
+ptrsMap (DState (UnifiedMap _ ptrmap) _ _ _) = ptrmap
+
+-- =======================
+-- CBOR instances
+
 instance NoThunks (DState crypto)
 
 instance NFData (DState crypto)
 
 instance CC.Crypto crypto => ToCBOR (DState crypto) where
-  toCBOR (DState rw dlg p fgs gs ir) =
+  toCBOR (DState unified fgs gs ir) =
     encodeListLen 6
-      <> toCBOR rw
-      <> toCBOR dlg
-      <> toCBOR p
+      <> toCBOR unified
       <> toCBOR fgs
       <> toCBOR gs
       <> toCBOR ir
@@ -385,13 +400,11 @@ instance CC.Crypto crypto => FromSharedCBOR (DState crypto) where
       (Interns (Credential 'Staking crypto), Interns (KeyHash 'StakePool crypto))
   fromSharedPlusCBOR = do
     decodeRecordNamedT "DState" (const 6) $ do
-      rw <- fromSharedPlusLensCBOR (toMemptyLens _1 _1)
-      dlg <- fromSharedPlusCBOR
-      p <- fromSharedPlusLensCBOR (toMemptyLens _2 _1)
+      unified <- fromSharedPlusCBOR
       fgs <- lift fromCBOR
       gs <- lift fromCBOR
       ir <- fromSharedPlusLensCBOR _1
-      pure $ DState rw dlg p fgs gs ir
+      pure $ DState unified fgs gs ir
 
 -- | Current state of staking pools and their certificate counters.
 data PState crypto = PState
@@ -416,12 +429,11 @@ instance CC.Crypto crypto => FromSharedCBOR (PState crypto) where
   type
     Share (PState crypto) =
       Interns (KeyHash 'StakePool crypto)
-  fromSharedPlusCBOR =
-    decodeRecordNamedT "PState" (const 3) $ do
-      _pParams <- fromSharedPlusLensCBOR (toMemptyLens _1 id)
-      _fPParams <- fromSharedPlusLensCBOR (toMemptyLens _1 id)
-      _retiring <- fromSharedPlusLensCBOR (toMemptyLens _1 id)
-      pure PState {_pParams, _fPParams, _retiring}
+  fromSharedPlusCBOR = decodeRecordNamedT "PState" (const 3) $ do
+    _pParams <- fromSharedPlusLensCBOR (toMemptyLens _1 id)
+    _fPParams <- fromSharedPlusLensCBOR (toMemptyLens _1 id)
+    _retiring <- fromSharedPlusLensCBOR (toMemptyLens _1 id)
+    pure PState {_pParams, _fPParams, _retiring}
 
 -- | The state associated with the current stake delegation.
 data DPState crypto = DPState
@@ -449,11 +461,10 @@ instance CC.Crypto crypto => FromSharedCBOR (DPState crypto) where
       ( Interns (Credential 'Staking crypto),
         Interns (KeyHash 'StakePool crypto)
       )
-  fromSharedPlusCBOR =
-    decodeRecordNamedT "DPState" (const 2) $ do
-      _pstate <- fromSharedPlusLensCBOR _2
-      _dstate <- fromSharedPlusCBOR
-      pure DPState {_pstate, _dstate}
+  fromSharedPlusCBOR = decodeRecordNamedT "DPState" (const 2) $ do
+    _pstate <- fromSharedPlusLensCBOR _2
+    _dstate <- fromSharedPlusCBOR
+    pure DPState {_pstate, _dstate}
 
 data AccountState = AccountState
   { _treasury :: !Coin,
@@ -612,11 +623,11 @@ pvCanFollow (ProtVer m n) (SJust (ProtVer m' n')) =
 --   Transactions can use Ptrs to refer to a stake credential in a TxOut. The Ptr
 --   does not have to point to anything until the epoch boundary, when we compute
 --   rewards and aggregate staking information for ranking. This is unusual but legal.
---   In a non incremental system, we use what ever 'legal' Ptrs exist at the epoch
+--   In a non incremental system, we use whatever 'legal' Ptrs exist at the epoch
 --   boundary. Here we are computing things incrementally, so we need to remember Ptrs
 --   that might point to something by the time the epoch boundary is reached. When
 --   the epoch boundary is reached we 'resolve' these pointers, to see if any have
---   become non-dangling since the time they came into the incremental computation.
+--   become non-dangling since the time they were first used in the incremental computation.
 data IncrementalStake crypto = IStake
   { credMap :: !(Map (Credential 'Staking crypto) Coin),
     ptrMap :: !(Map Ptr Coin)
@@ -827,7 +838,7 @@ getGKeys nes = Map.keysSet genDelegs
   where
     NewEpochState _ _ _ es _ _ = nes
     EpochState _ _ ls _ _ _ = es
-    LedgerState _ (DPState (DState _ _ _ _ (GenDelegs genDelegs) _) _) = ls
+    LedgerState _ (DPState (DState _ _ (GenDelegs genDelegs) _) _) = ls
 
 -- | The state associated with a 'Ledger'.
 data LedgerState era = LedgerState
@@ -1188,12 +1199,12 @@ depositPoolChange ls pp tx = (currentPool <+> txDeposits) <-> txRefunds
     txRefunds = keyRefunds pp tx
 
 reapRewards ::
+  UnifiedMap crypto ->
   RewardAccounts crypto ->
-  RewardAccounts crypto ->
-  RewardAccounts crypto
-reapRewards dStateRewards withdrawals =
-  Map.mapWithKey removeRewards dStateRewards
+  UnifiedMap crypto
+reapRewards (UnifiedMap tmap ptrmap) withdrawals = (UnifiedMap (Map.mapWithKey g tmap) ptrmap)
   where
+    g k (Triple x y z) = Triple (fmap (removeRewards k) x) y z
     removeRewards k v = if k `Map.member` withdrawals then Coin 0 else v
 
 ---------------------------------
@@ -1212,14 +1223,18 @@ stakeDistr ::
 stakeDistr u ds ps =
   SnapShot
     (Stake $ VMap.fromMap (compactCoinOrError <$> eval (dom activeDelegs ◁ stakeRelation)))
-    (VMap.fromMap delegs)
+    (VMap.fromMap (UM.unUnify delegs))
     (VMap.fromMap poolParams)
   where
-    DState rewards' delegs ptrs' _ _ _ = ds
+    rewards' = rewards ds
+    delegs = delegations ds
+    ptrs' = ptrsMap ds
     PState poolParams _ _ = ps
     stakeRelation :: Map (Credential 'Staking (Crypto era)) Coin
-    stakeRelation = aggregateUtxoCoinByCredential (forwards ptrs') u rewards'
-    activeDelegs :: Map (Credential 'Staking (Crypto era)) (KeyHash 'StakePool (Crypto era))
+    stakeRelation = aggregateUtxoCoinByCredential ptrs' u (UM.unUnify rewards')
+    -- The use of  (UM.unUnify rewards') looks exspensive, but since we now incrementally
+    -- compute stake distribution, this function is ONLY used in tests
+    activeDelegs :: ViewMap (Crypto era) (Credential 'Staking (Crypto era)) (KeyHash 'StakePool (Crypto era))
     activeDelegs = eval ((dom rewards' ◁ delegs) ▷ dom poolParams)
 
 compactCoinOrError :: Coin -> CompactForm Coin
@@ -1275,9 +1290,11 @@ updateStakeDistribution incStake0 utxoDel utxoAdd = incStake2
     incStake2 = incrementalAggregateUtxoCoinByCredential invert utxoDel incStake1
 
 -- | Incrementally sum up all the Coin for each staking Credential, use different 'mode' operations
---   for inserts (id) and deletes (invert). Never store a (Coin 0) balance, since these do not occur
---   in the non-incremental stye that works directly from the UTxO. This function has a non-incremental
---   analog 'aggregateUtxoCoinByCredential'
+--   for UTxO that are inserts (id) and UTxO that are deletes (invert). Never store a (Coin 0) balance,
+--   since these do not occur in the non-incremental stye that works directly from the whole UTxO.
+--   This function has a non-incremental analog 'aggregateUtxoCoinByCredential' . In this incremental
+--   version we expect the size of the UTxO to be fairly small. I.e the number of inputs and outputs
+--   in a transaction, which is aways < 4096, not millions, and very often < 10).
 incrementalAggregateUtxoCoinByCredential ::
   forall era.
   ( Era era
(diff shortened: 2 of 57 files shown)


eras/shelley/test-suite/src/Test/Cardano/Ledger/Shelley/Generator/Trace/Chain.hs
```diff
@@ -241,11 +242,11 @@ registerGenesisStaking
...
-          { _rewards =
-              Map.map (const $ Coin 0)
-                . Map.mapKeys KeyHashObj
-                $ sgsStake,
-            _delegations = Map.mapKeys KeyHashObj sgsStake
+          { _unified =
+              UM.unify
+                (Map.map (const $ Coin 0) . Map.mapKeys KeyHashObj $ sgsStake)
+                (Map.mapKeys KeyHashObj sgsStake)
+                (Map.empty)
```
TimSheard: I wonder if the (Map.empty) should be the Ptrs equivalent of the oldDPState, we should probably fix this.

libs/cardano-data/src/Data/MapExtras.hs
```diff
@@ -0,0 +1,146 @@
...
+-- | - fold over the intersection of a Map and a Set
+intersectMapSetFold :: Ord k => (k -> v -> ans -> ans) -> Map k v -> Set k -> ans -> ans
+intersectMapSetFold _accum Tip _ !ans = ans
+intersectMapSetFold _accum _ set !ans | Set.null set = ans
+intersectMapSetFold accum (Bin _ k v l1 l2) set !ans =
+  intersectMapSetFold accum l1 s1 (addKV k v (intersectMapSetFold accum l2 s2 ans))
+  where
+    (s1, found, s2) = Set.splitMember k set
+    addKV k1 v1 !ans1 = if found then accum k1 v1 ans1 else ans1
+{-# INLINEABLE intersectMapSetFold #-}
```
TimSheard, line 118: I think Alexey had some thoughts about this function that used to be in SetAlgebra.hs, It's been deleted there and added here.   Maybe it needs to be changed.

libs/set-algebra/src/Control/Iterate/BaseTypes.hs
```diff
@@ -74,8 +72,9 @@ class Iter f => Basic f where
   removekey :: (Ord k) => k -> f k v -> f k v
   domain :: Ord k => f k v -> Set k
   range :: Ord v => f k v -> Set v
-  emptyc :: Ord k => f k v
-  emptyc = error ("emptyc only works on some types.")
+
+-- emptyc :: Ord k => f k v
+-- emptyc = error ("emptyc only works on some types.")
```
TimSheard: delete this old code.

libs/set-algebra/src/Control/Iterate/SetAlgebra.hs
```diff
@@ -439,129 +461,6 @@ rewrite (RRestrict (Base SplitR xs) (Base SetR (Sett s))) = Just $ filterWithKey
...
--- ===========================================================
--- Some times we need to write our own version of functions
--- over  Map.Map that do not appear in the library
--- For example
--- 1) version of Map.withoutKeys where both parts are Map.Map
--- 2) Comparing that two maps have exactly the same set of keys
--- 3) The intersection of two maps guarded by a predicate.
---    ((dom stkcred) ◁ deleg) ▷ (dom stpool))   ==>
---    intersectDomP (\ k v -> Map.member v stpool) stkcred deleg
--- ============================================================
```
TimSheard, line 451: Moved to Data.MapExtras

libs/set-algebra/src/Control/Iterate/BaseTypes.hs
TimSheard: done

eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState.hs
```diff
@@ -1226,17 +1259,73 @@ incrementalStakeDistr ::
...
-    stake0, stake1 :: Map (Credential 'Staking (Crypto era)) Coin
-    stake0 = resolveIncrementalPtrs @era (forwards bimap) incstake
-    stake1 = Map.unionWith (<>) stake0 rewards'
-    activeDelegs :: Map (Credential 'Staking (Crypto era)) (KeyHash 'StakePool (Crypto era))
-    activeDelegs = eval ((dom rewards' ◁ delegs) ▷ dom poolParams)
+    delegs = delegations ds
+    step1 = resolveActiveIncrementalPtrs (activeP tripmap) ptrmap incstake
+    step2 = aggregateActiveStake tripmap step1
+
+-- | A credential is active, only if the the third part of the triple is (SJust _)
```
JaredCorduan:
```suggestion
-- | A credential is active, only if the third part of the triple is (SJust _)
```

eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState.hs
```diff
@@ -1226,17 +1259,73 @@ incrementalStakeDistr ::
...
+activeP :: Map (Credential 'Staking crypt0) (Triple crypto) -> Credential 'Staking crypt0 -> Bool
```
JaredCorduan, line 1372: what does the `P` stand for?

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Deleg.hs
```diff
@@ -255,34 +256,30 @@ delegationTransition = do
...
-        ds
-          { _rewards = eval (setSingleton hk ⋪ _rewards ds),
-            _delegations = eval (setSingleton hk ⋪ _delegations ds),
-            _ptrs = eval (_ptrs ds ⋫ setSingleton hk)
-          }
+        let u0 = _unified ds
+            u1 = (Set.singleton hk UM.⋪ Rewards u0)
+            u2 = (Set.singleton hk UM.⋪ Delegations u1)
+            u3 = (Ptrs u2 UM.⋫ Set.singleton hk)
+         in ds {_unified = u3}
```
JaredCorduan, line 295: is there not a way to just remove `hk` from the unified map?

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Ledger.hs
```diff
@@ -189,7 +190,7 @@ instance
         "Deposit pot must equal obligation"
         ( \(TRC (LedgerEnv {ledgerPp}, _, _))
            (utxoSt, DPState {_dstate, _pstate}) ->
-              obligation ledgerPp (_rewards _dstate) (_pParams _pstate)
+              obligation ledgerPp (rewards _dstate) (_pParams _pstate) -- FIX ME
```
JaredCorduan, line 213: what needs fixing?

libs/cardano-data/src/Data/UMap.hs
```diff
@@ -0,0 +1,691 @@
...
+data Trip coin pool ptr = Triple
+  { coinT :: !(StrictMaybe coin),
+    ptrT :: !(Set ptr),
+    poolidT :: !(StrictMaybe pool)
+  }
+  deriving (Show, Eq, Generic, NoThunks, NFData)
+-}
+
+data Trip coin pool ptr
+  = TEEE
```
JaredCorduan, line 94: `E` for empty, `F` for full?

JaredCorduan (commented):
@TimSheard this really is a great abstraction! It's fantastic how the new code reads so similar to the old! I only had a few trivial comments. I'll wait to approve until my eyes have a change to go over this. Thanks for all this hard work!

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Ledger.hs
TimSheard: _rewards  to rewards, notice no  underscore

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Rules/Ledger.hs
```diff
@@ -52,6 +53,7 @@ import Data.Kind (Type)
 import Data.Sequence (Seq)
 import Data.Sequence.Strict (StrictSeq)
 import qualified Data.Sequence.Strict as StrictSeq
+import Data.UMap ()
```
lehins: Redundant import
```suggestion
```

libs/cardano-data/src/Data/Sharing.hs
```diff
@@ -136,6 +137,9 @@ class Monoid (Share a) => FromSharedCBOR a where
     x <- lift $ fromSharedCBOR s
     x <$ put (getShare x <> s)
 
+  fromSharedPlusCBOR' :: Share a -> Decoder s (a, Share a)
+  fromSharedPlusCBOR' = runStateT fromSharedPlusCBOR
+
```
lehins: It is wrong to add another equivalent function to a class. It is always Possible to use this manual state passing style at the callsite if it is really necessary:

```haskell
  fromSharedPlusCBOR = StateT $ \(share :: Share a) -> undefined :: Decoder s (a, Share a)
```

```suggestion
```

libs/cardano-data/src/Data/Sharing.hs
```diff
@@ -125,7 +126,7 @@ class Monoid (Share a) => FromSharedCBOR a where
   -- | Utilize sharing when decoding, but do not add anything to the state for
   -- future sharing.
   fromSharedCBOR :: Share a -> Decoder s a
-  fromSharedCBOR = evalStateT fromSharedPlusCBOR
+  fromSharedCBOR s = fst <$> fromSharedPlusCBOR' s
```
lehins: Reverting. See my comment on `fromSharedPlusCBOR'`

```suggestion
  fromSharedCBOR = evalStateT fromSharedPlusCBOR
```

libs/cardano-data/src/Data/Sharing.hs
```diff
@@ -112,7 +113,7 @@ instance Semigroup (Interns a) where
         | otherwise = i : a : as
 
 class Monoid (Share a) => FromSharedCBOR a where
-  {-# MINIMAL (fromSharedCBOR | fromSharedPlusCBOR) #-}
+  {-# MINIMAL (fromSharedCBOR | fromSharedPlusCBOR | fromSharedPlusCBOR') #-}
```
lehins: Reverting. See my comment on `fromSharedPlusCBOR'`
```suggestion
  {-# MINIMAL (fromSharedCBOR | fromSharedPlusCBOR) #-}
```

libs/cardano-data/src/Data/UMap.hs
```diff
@@ -0,0 +1,691 @@
...
+instance
+  (Tbor coin, Tbor ptr, Tbor cred, ToCBOR pool, Ord ptr) =>
+  ToCBOR (UMap coin cred pool ptr)
+  where
+  toCBOR (UnifiedMap tripmap ptrmap) =
+    encodeListLen 2 <> encodeMap toCBOR toCBOR tripmap <> encodeMap toCBOR toCBOR ptrmap
+
+-- One can do this using fromSharedPlusCBOR and Lenses and lift, but I think this is much clearer
+-- using fromSharedPlusCBOR'. By giving a name to the shared state, i.e. (a,b) we can use
+-- individual pieces of it where we need, without involving Lense combinators or lift. The only
```
lehins:
```suggestion
-- using manual state passing. By giving a name to the shared state, i.e. (a,b) we can use
-- individual pieces of it where we need, without involving Lens combinators or lift. The only
```

libs/cardano-data/src/Data/UMap.hs
```diff
@@ -0,0 +1,691 @@
...
+
+data Trip coin pool ptr = Triple
+  { coinT :: !(StrictMaybe coin),
+    ptrT :: !(Set ptr),
+    poolidT :: !(StrictMaybe pool)
+  }
+  deriving (Show, Eq, Generic, NoThunks, NFData)
+-}
+
+data Trip coin pool ptr
```
lehins: Would it make sense for consistency with the `pattern` change the order of parameters? Also the order of `E` and `F` in constructors otherwise are a bit confusing (eg. `TFEF` means that `ptr` are empty, not `pool`)
```suggestion
data Trip coin ptr pool
```

libs/cardano-data/src/Data/UMap.hs
```diff
@@ -0,0 +1,691 @@
...
+      case Map.lookup ptr ptrmap of
+        Nothing -> all (\(Triple _ set _) -> not (Set.member ptr set)) tripmap
+        Just cred ->
+          case Map.lookup cred tripmap of
+            Nothing -> False
+            Just (Triple _ set _) -> Set.member ptr set
+
+-- =====================================================
+
+data VMap coin cr pl ptr k v where
```
lehins: We already have a `VMap`. it is a very bad idea to have a name clash like that in the same codebase.

Either one of them need to be renamed. I suggest either renaming this one to `ViewMap` (or anything else that doesn't clash) or the one from `Data.Compact.VMap` to `VecMap`?

libs/cardano-data/src/Data/UMap.hs
```diff
@@ -0,0 +1,691 @@
...
+    accum2 ans _ = ans
+foldl' accum ans0 (Delegations (UnifiedMap tmap _)) = Map.foldl' accum2 ans0 tmap
+  where
+    -- accum2 ans (Triple _ _ (SJust p)) = accum ans p -- Don't use pattern Triple in tight loop.
+    accum2 ans (TFFF _ _ p) = accum ans p
+    accum2 ans (TFEF _ p) = accum ans p
+    accum2 ans (TEFF _ p) = accum ans p
+    accum2 ans (TEEF p) = accum ans p
+    accum2 ans _ = ans
+foldl' accum ans (Ptrs (UnifiedMap _ ptrmap)) = Map.foldl' accum ans ptrmap
```
lehins: Is this implementation more efficient than the one provided by `Foldable`? If it is not then is the reason to specify it? But if it is (which I suspect it might be since it is using `Map.foldl'`) then we should override default `foldl'` implementation if `Foldable`

libs/cardano-data/src/Data/UMap.hs
```diff
@@ -0,0 +1,691 @@
...
+instance Foldable (VMap coin cred pool ptr k) where
+  foldMap f (Rewards (UnifiedMap tmap _)) = Map.foldlWithKey' accum mempty tmap
+    where
+      accum ans _ (Triple (SJust c) _ _) = ans <> (f c)
+      accum ans _ _ = ans
+  foldMap f (Delegations (UnifiedMap tmap _)) = Map.foldlWithKey' accum mempty tmap
+    where
+      accum ans _ (Triple _ _ (SJust c)) = ans <> (f c)
+      accum ans _ (Triple _ _ SNothing) = ans
+  foldMap f (Ptrs (UnifiedMap _ ptrmap)) = foldMap f ptrmap
```
lehins: Using `foldlWithKey'` isn't strictly correct (pun intended), since `foldMap` should be lazy and there is a strict version `foldMap'`. But I don't think we care too much about laziness in this codebase.

libs/cardano-data/src/Data/UMap.hs
```diff
@@ -0,0 +1,691 @@
...
+    Nothing -> Nothing
+    Just (k, Triple _ _ (SJust poolid), tripmap2) -> Just (k, poolid, delegations tripmap2 Map.empty)
+    Just (_, Triple _ _ SNothing, tripmap2) -> next (delegations tripmap2 Map.empty)
+next (Ptrs (UnifiedMap tripmap ptrmap)) =
+  case mapNext ptrmap of
+    Nothing -> Nothing
+    Just (k, stakeid, m2) -> Just (k, stakeid, ptrs (pick tripmap Map.empty) m2)
+
+pick :: x -> x -> x
+pick _old new = new
```
lehins: There is a special function for this in Prelude: `asTypeOf`, except its arguments are flipped.

```suggestion
```

libs/cardano-data/src/Data/UMap.hs
```diff
@@ -0,0 +1,691 @@
...
+    Just (_, Triple SNothing _ _, tripmap2) -> next (rewards tripmap2 Map.empty)
+next (Delegations (UnifiedMap tripmap _)) =
+  case mapNext tripmap of
+    Nothing -> Nothing
+    Just (k, Triple _ _ (SJust poolid), tripmap2) -> Just (k, poolid, delegations tripmap2 Map.empty)
+    Just (_, Triple _ _ SNothing, tripmap2) -> next (delegations tripmap2 Map.empty)
+next (Ptrs (UnifiedMap tripmap ptrmap)) =
+  case mapNext ptrmap of
+    Nothing -> Nothing
+    Just (k, stakeid, m2) -> Just (k, stakeid, ptrs (pick tripmap Map.empty) m2)
```
lehins:
```suggestion
    Just (k, stakeid, m2) -> Just (k, stakeid, ptrs (Map.empty `asTypeOf` trimap) m2)
```

libs/cardano-data/src/Data/UMap.hs
```diff
@@ -0,0 +1,691 @@
...
+    Just (_, Triple SNothing _ _, tripmap2) -> next (rewards tripmap2 Map.empty)
+leastUpperBound stakeid (Delegations (UnifiedMap tripmap _)) =
+  case mapLub stakeid tripmap of
+    Nothing -> Nothing
+    Just (k, Triple _ _ (SJust poolid), tripmap2) -> Just (k, poolid, delegations tripmap2 Map.empty)
+    Just (_, Triple _ _ SNothing, tripmap2) -> next (Delegations (UnifiedMap tripmap2 Map.empty))
+leastUpperBound ptr (Ptrs (UnifiedMap tripmap ptrmap)) =
+  case mapLub ptr ptrmap of
+    Nothing -> Nothing
+    Just (k, stakeid, m2) -> Just (k, stakeid, ptrs (pick tripmap Map.empty) m2)
```
lehins:
```suggestion
    Just (k, stakeid, m2) -> Just (k, stakeid, ptrs (Map.empty `asTypeOf` trimap) m2)
```

libs/cardano-data/src/Data/UMap.hs
```diff
@@ -0,0 +1,691 @@
...
+
+instance (Ord pool, Monoid coin) => Semigroup (Trip coin ptr pool) where
+  (<>) (Triple c1 ptrs1 x) (Triple c2 ptrs2 y) = Triple (addC c1 c2) (Set.union ptrs1 ptrs2) (add x y)
+    where
+      add SNothing SNothing = SNothing
+      add (SJust w) SNothing = SJust w
+      add SNothing (SJust z) = SJust z
+      add (SJust w) (SJust _) = SJust w
+
+addC :: Monoid x => StrictMaybe x -> StrictMaybe x -> StrictMaybe x
```
lehins: We should add `Monoid` and `Semigroup` instances fro `StrictMaybe`. For now renaming this to reflect would be good IMHO: 

```suggestion
appendStrictMaybe :: Semigroup x => StrictMaybe x -> StrictMaybe x -> StrictMaybe x
```

libs/cardano-data/src/Data/UMap.hs
```diff
@@ -0,0 +1,691 @@
...
+  FromSharedCBOR (UMap coin cred pool ptr)
+  where
+  type
+    Share (UMap coin cred pool ptr) =
+      (Interns cred, Interns pool)
+  fromSharedPlusCBOR' (a, b) =
+    decodeRecordNamed "UnifiedMap" (const 2) $ do
+      tripmap <- decodeMap (interns a <$> fromCBOR) (fromSharedCBOR b)
+      ptrmap <- decodeMap fromCBOR (interns a <$> fromCBOR)
+      pure (UnifiedMap tripmap ptrmap, (internsFromMap tripmap <> a, b))
```
lehins: No need to add special function to the class. `StateT` is just a newtype around this whole manual state passing style. There is nothing wrong with using this style if you think it is cleaner or simply find it easier to use: 
```haskell
  fromSharedPlusCBOR = StateT $ \ (a, b) ->
```

However, I think you missed out on some sharing opportunity:

```suggestion
  fromSharedPlusCBOR = StateT $ \ (a, b) ->
    decodeRecordNamed "UnifiedMap" (const 2) $ do
      tripmap <- decodeMap (interns a <$> fromCBOR) (fromSharedCBOR b)
      let a' = internsFromMap tripmap <> a
      ptrmap <- decodeMap fromCBOR (interns a' <$> fromCBOR)
      pure (UnifiedMap tripmap ptrmap, (a', b))
```

libs/cardano-data/src/Data/UMap.hs
```diff
@@ -0,0 +1,691 @@
...
+-- derived operations
+
+findWithDefault :: (Ord cred, Ord ptr) => a -> k -> VMap coin cred pool ptr k a -> a
+findWithDefault a1 k vm =
+  case lookup k vm of
+    Just a2 -> a2
+    Nothing -> a1
+
+size :: VMap coin cred pool ptr k a -> Int
+size x = foldl' (\count _v -> count + 1) 0 x
```
lehins: `Map.size` is O(1), but this implementation is O(n). I think manual pattern match on underlying maps and using `Map.size`  would be a good optimization

lehins (changes requested):
This is a pretty big PR. 

Here is what I have so far. I'll do a second pass when I am back from vacation.

eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState.hs
TimSheard: done

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Deleg.hs
TimSheard: We could combine these into one operation, but then we don't have the nice continuation between the old code, the spec,  and the new code.

libs/cardano-data/src/Data/UMap.hs
TimSheard: Yes Exactly, I added some comments to explain the convention.

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Rules/Ledger.hs
TimSheard: Done.

eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState.hs
TimSheard: Predicate?  Its a Lisp thing. predicates (functions to True and False) often end in P.

libs/cardano-data/src/Data/Sharing.hs
TimSheard: As per our discussion, We now export only one Plus method. I fixed things so this compiles, but I think it only emphasizes my point. There are two different views, and both are usefull.

libs/cardano-data/src/Data/Sharing.hs
TimSheard: no longer applicable

libs/cardano-data/src/Data/UMap.hs
TimSheard: no longer applicable

libs/cardano-data/src/Data/UMap.hs
TimSheard: done. I didn't know about this. It comes in handy.

libs/cardano-data/src/Data/UMap.hs
TimSheard: I changed the name. The problem with the Semigroup instance is there are several possible instances (left biased, right biased, and as illustrated here the instance where the type variable is a monoid)  It is not my type, so choosing one of these arbitrarily seems wrong.

libs/cardano-data/src/Data/UMap.hs
TimSheard: THe sharing needs some thought, as this type has a bunch of invariants, so testing for extra sharing might just be a waste of time.

libs/cardano-data/src/Data/UMap.hs
TimSheard: The problem is that there is no constant time size for KeyMap, so you will aways resort to expensive algorithms.  This approach avoids tricky interations at the boundary of IntMap and KeyMap

lehins: I think you might have mixed up two PRs, in particular mine with utxo+SplitMap and this one. `UMap` doesn't use `KeyMap`, instead it relies on `Map` which hash O(1) size op

libs/cardano-data/src/Data/MapExtras.hs
lehins: Sorry, I don't remember having any thoughts about this function. :man_shrugging:

libs/cardano-data/test/Test/Data/UMap.hs
```diff
@@ -0,0 +1,474 @@
...
+-- local types for Testing
+
+data Coin = Coin Integer
+  deriving (Eq, Ord)
+
+instance Show Coin where show (Coin n) = "(Coin " ++ show n ++ ")"
+
+instance Semigroup Coin where (Coin x) <> (Coin y) = Coin (x + y)
+
+instance Monoid Coin where mempty = Coin 0
```
lehins: We can avoid defining all those instances and reuse `Data.Semigroup.Sum`  instead
```suggestion
type Coin = Sum Integer
```

libs/cardano-data/test/Test/Data/UMap.hs
```diff
@@ -0,0 +1,474 @@
...
+
+type View = VMap Coin Cred Pool Ptr
+
+type UnifiedMap = UMap Coin Cred Pool Ptr
+
+-- ======================================================
+-- Generators to make Arbitrary instances for Testing
+
+genPool :: Gen Pool
+genPool = oneof (map pure "abcdefghijklmnopqrstuvwxyz")
```
lehins:
```suggestion
genPool = elements ['a'..'z']
```

libs/cardano-data/test/Test/Data/UMap.hs
```diff
@@ -0,0 +1,474 @@
...
+      testPropertyN 1000 "domain restrict Ptrs" domRestrictPtr,
+      testPropertyN 1000 "unify unUnify roundtrip" unifyRoundTrip1,
+      testPropertyN 1000 "unUnify unify roundtrip" unifyRoundTrip2,
+      testPropertyN 1000 "size is accurate" sizeUMap
+    ]
+
+-- =======================================================
+
+test :: IO ()
+test = defaultMain alltests
```
lehins: dead code
```suggestion
-- =======================================================

test :: IO ()
test = defaultMain alltests
```

libs/compact-map/src/Data/Compact/VMap.hs
```diff
@@ -117,6 +118,10 @@ fromMap :: (VG.Vector kv k, VG.Vector vv v) => Map.Map k v -> VMap kv vv k v
 fromMap = VMap . KV.fromMap
 {-# INLINE fromMap #-}
 
+fromMapWith :: (VG.Vector kv k, VG.Vector vv u) => (v -> u) -> Map.Map k v -> VMap kv vv k u
+fromMapWith f = VMap . KV.fromMapWith f
+{-# INLINE fromMapWith #-}
+
```
lehins: It doesn't look like it is being used anywhere in this PR. Any reason why we need this function? This is equivalent to `fromMap . Map.map f`
```suggestion
```

libs/compact-map/src/Data/Compact/KVVector.hs
```diff
@@ -70,6 +71,11 @@ fromMap ::
 fromMap m = fromDistinctAscListN (Map.size m) $ Map.toAscList m
 {-# INLINE fromMap #-}
 
+-- | Convert a `Map.Map` into a sorted key/value vector where we change the type of the range
+fromMapWith ::
+  (VG.Vector kv k, VG.Vector vv u) => (v -> u) -> Map.Map k v -> KVVector kv vv (k, u)
+fromMapWith f m = fromDistinctAscListN (Map.size m) $ map (\(k, v) -> (k, f v)) $ Map.toAscList m
+
```
lehins: See my other comment. This function is not being used
```suggestion
```

lehins (changes requested):
Few more comments

libs/cardano-data/test/Test/Data/UMap.hs
TimSheard: The problem with that is, one cannot control the Show and Arbitrary instances if Coin is a type synonym. How about this, which gets the best of both worlds
`newtype Coin = Coin Integer
  deriving (Ord,Eq)
  deriving (Monoid,Semigroup) via (Sum Integer)`

libs/cardano-data/test/Test/Data/UMap.hs
TimSheard: Its is only dead code, if your not trying to run the tests from inside the repl

libs/compact-map/src/Data/Compact/VMap.hs
TimSheard: Originally this was going to be used at the epoch boundary when we made the SnapShots. But it never got used. So I guess we can delete it.

libs/compact-map/src/Data/Compact/KVVector.hs
TimSheard: Same reason as the other fromMapWith. it is now deleted.

libs/cardano-data/src/Data/UMap.hs
TimSheard: You are right, I got these mixed up. But, there is another subtle thing we need to consider. A VMap is a view of a UnifiedMap, Which is a pair of maps, So we would need something like
`size (Delegations (Unified tripmap ptrmap)) = ...
 size (Rewards (Unified tripmap ptrmap)) = ...
 size (Ptrs (Unified tripmap ptrmap)) = ... `
The size of the view for Delegations and Rewards depends upon the shape of the Triple in the tripmap, NOT the size of tripmap. So the best we can do is inspect each of the triples for their shape, and count the ones with the right shape. That inspecting the Triples is exactly what the foldl' does.  So we can do a bit better, because the Ptr view size can be computed directly from the 'ptrmap'.  So here is a better, but not optimal solution.
`size :: VMap coin cred pool ptr k a -> Int
size (Ptrs (UnifiedMap _ ptrmap)) = Map.size ptrmap
size x = foldl' (\count _v -> count + 1) 0 x
`

libs/cardano-data/test/Test/Data/UMap.hs
TimSheard: clever. done

libs/cardano-data/test/Test/Data/UMap.hs
lehins: I realize that, but if you are the only one running it from repl it's not that hard for you to run `> defaultMain alltests` instead of `> test`, right?

libs/cardano-data/test/Test/Data/UMap.hs
lehins: :100:

libs/cardano-data/src/Data/MapExtras.hs
TimSheard: OK, we will just leave it alone.

libs/cardano-data/src/Data/Sharing.hs
TimSheard: This has been reverted to the way it was.

libs/cardano-data/src/Data/UMap.hs
TimSheard: After further review, the sharing looks ike a good idea. It is now done

libs/cardano-data/test/Test/Data/UMap.hs
TimSheard: I do not agree with this, but I have removed it in order to get things moving. I think that every test file should have 
'test = defaultMain alltests'
This is extremely useful in a debugging workflow, and I don't think it is dead code at all, It is all a matter of perspective.

libs/cardano-data/src/Data/UMap.hs
TimSheard: now use asTypeOf

libs/cardano-data/src/Data/UMap.hs
TimSheard: So it now overrides the default foldl' in the Foldable instance

libs/cardano-data/src/Data/UMap.hs
TimSheard: So now VMap is View, which I think is more descriptive

libs/cardano-data/src/Data/UMap.hs
TimSheard: Good idea, it makes things clearer.

libs/cardano-data/src/Data/UMap.hs
TimSheard: I changed it to  Map.foldlWithKey , so it is now lazy. I don't think it is ever used, except it is required for the Foldable instance. What really matters is for foldl' to be strict, and now it overrides the default implementation, so everything should be good.

lehins (approved):
Sweet!

PR #2585, opened 2021-12-10
Attempt to improve the stake pool calculation:

lehins:
Performance should improve due to reduction of number of iterations over
the total stake pool.


JaredCorduan (approved):
Wow, this seemed to make a pretty dramatic improvement for me! I was able to sync from the end of byron to the beginning of Mary in about two thirds the time! :raised_hands:

weebl2000:
Yeah initial ledger creation is lightning quick compared to previous versions.

PR #2586, opened 2021-12-14
Fix the `Ord` instance for `TxIn`

lehins:


nc6:
I'll test this on sync

nc6 (approved):
Looks great, chain extends past the broken point :-)

PR #2591, opened 2021-12-15
fix bug in taylorExpCmp

JaredCorduan:
When checking a negative value with `taylorExpCmp`, the error term `errorTerm` needs to be an absolute value.

The ledger code [never calls](https://github.com/input-output-hk/cardano-ledger/blob/1a9ec4ae9e0b09d54e49b2a40c4ead37edadcce5/libs/cardano-protocol-tpraos/src/Cardano/Protocol/TPraos/BHeader.hs#L423) `taylorExpCmp` with a negative value (note that c is negative, it is the [natural log](https://github.com/input-output-hk/cardano-ledger/blob/1a9ec4ae9e0b09d54e49b2a40c4ead37edadcce5/libs/cardano-ledger-core/src/Cardano/Ledger/BaseTypes.hs#L561) of a number less than one), but the hazard needs to be removed for the saftey of the future.

It's nice to know that the property test that I wrote for this change fails with the absolute value is not taken. :)


iquerejeta (approved):
Looks good to me 👍

PR #2592, opened 2021-12-17
Use splitmap for utxo

lehins:
This PR switches UTxO to `SplitMap`, which is in itself is based on a nested data structure `IntMap` and custom `KeyMap`  that is indexed by 32byte keys.


eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Rules/Utxo.hs
```diff
@@ -257,11 +260,12 @@ feesOK ::
   Core.Tx era ->
   UTxO era ->
   Rule (AlonzoUTXO era) 'Transition ()
-feesOK pp tx (UTxO m) = do
+feesOK pp tx (UTxO utxo) = do
   let txb = getField @"body" tx
       theFee = getField @"txfee" txb -- Coin supplied to pay fees
       collateral = getField @"collateral" txb -- Inputs allocated to pay theFee
-      utxoCollateral = eval (collateral ◁ m) -- restrict Utxo to those inputs we use to pay fees.
+      -- restrict Utxo to those inputs we use to pay fees.
```
JaredCorduan, line 269:
```suggestion
      -- restrict Utxo to those inputs we use to pay fees.
      -- (collateral ◁ utxo)
```

eras/shelley-ma/impl/src/Cardano/Ledger/ShelleyMA/Rules/Utxo.hs
```diff
@@ -212,20 +211,12 @@ consumed ::
...
   Core.Value era
-consumed pp u tx =
-  balance @era (eval (txins @era tx ◁ u))
-    <> getField @"mint" tx
-    <> Val.inject (refunds <> withdrawals)
-  where
-    -- balance (UTxO (Map.restrictKeys v (txins tx))) + refunds + withdrawals
-    refunds = Shelley.keyRefunds pp tx
-    withdrawals = fold . unWdrl $ getField @"wdrls" tx
+consumed pp u tx = Shelley.consumed pp u tx <> getField @"mint" tx
```
JaredCorduan, line 215: :clap:

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Rules/Utxo.hs
```diff
@@ -257,11 +260,13 @@ feesOK ::
...
+      -- (collateral ◁ utxo)
+      utxoCollateral = utxo `SplitMap.restrictKeysSet` collateral
```
TimSheard: I'd suggest we do something like this. In the SplitMap.hs we define
(◁ ) dom map = map `SplitMap.restrictKeysSet` dom 

And in the file Utxo.hs we say
import qualified Cardano.Data.SplitMap as Split
      
utxoCollateral =  (collateral Split.◁ utxo)

This maintains the correspondence between the spec, the old code and the new code.

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Rules/Utxo.hs
```diff
@@ -347,26 +352,27 @@ utxoTransition = do
   {-   txins txb ≠ ∅   -}
   not (Set.null (getField @"inputs" txb)) ?!# InputSetEmptyUTxO
 
-  {-   feesOKp p tx utxo   -}
+  {-   feesOK pp tx utxo   -}
   feesOK pp tx utxo -- Generalizes the fee to small from earlier Era's
 
-  {-   (txins txb) ∪ (collateral txb)  ⊆ dom utxo   -}
-  eval (inputsAndCollateral ⊆ dom utxo)
-    ?! BadInputsUTxO (eval (inputsAndCollateral ➖ dom utxo))
```
TimSheard, line 357: we could do something similar to before like this
(inputsAndCollateral Split.⊆ dom utxo)
    ?! BadInputsUTxO ((inputsAndCollateral Split.➖ dom utxo))

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Rules/Utxow.hs
```diff
@@ -264,10 +265,11 @@ alonzoStyleWitness = do
 
   {-  { h | (_ → (a,_,h)) ∈ txins tx ◁ utxo, isNonNativeScriptAddress tx a} = dom(txdats txw)   -}
   let inputs = getField @"inputs" txbody :: (Set (TxIn (Crypto era)))
-      smallUtxo = eval (inputs ◁ utxo) :: Map.Map (TxIn (Crypto era)) (Core.TxOut era)
+      smallUtxo :: [(TxIn (Crypto era), Core.TxOut era)]
+      smallUtxo = SplitMap.toList (unUTxO utxo `SplitMap.restrictKeysSet` inputs)
       twoPhaseOuts =
```
TimSheard: Perhaps like this?

smallUtxo = SplitMap.toList (inputs Split.◁ utxo)

eras/alonzo/test-suite/src/Test/Cardano/Ledger/Alonzo/AlonzoEraGen.hs
```diff
@@ -415,11 +415,12 @@ instance Mock c => EraGen (AlonzoEra c) where
...
           mapScriptWit
           -- (dataMapFromTxOut (Prelude.foldr (:) [] (outputs' txbody)) (TxDats (getDataMap scriptinfo mapScriptWit)))
-          (dataMapFromTxOut (Prelude.foldr (:) [] (Map.elems smallUtxo)) (TxDats (getDataMap scriptinfo mapScriptWit)))
+          (dataMapFromTxOut smallUtxo (TxDats (getDataMap scriptinfo mapScriptWit)))
           -- The data hashes come from two places
           (Redeemers rdmrMap)
       txinputs = inputs' txbody
-      smallUtxo = eval (txinputs ◁ utxo) :: Map.Map (TxIn c) (Core.TxOut (AlonzoEra c))
+      smallUtxo :: [Core.TxOut (AlonzoEra c)]
+      smallUtxo = SplitMap.elems (unUTxO utxo `SplitMap.restrictKeysSet` txinputs)
```
TimSheard, line 423: Perhaps like this

smallUtxo = (txinputs Split.◁ utxo)

eras/alonzo/test-suite/src/Test/Cardano/Ledger/Alonzo/AlonzoEraGen.hs
```diff
@@ -486,7 +487,7 @@ sumCollateral ::
   UTxO era ->
   Coin
 sumCollateral tx (UTxO utxo) =
-  coin . balance @era . UTxO . eval $ collateral_ ◁ utxo
+  coin . balance @era . UTxO $ SplitMap.restrictKeysSet utxo collateral_
```
TimSheard, line 490: Perhaps like this?

coin . balance @era . UTxO  $ (collateral_ Split.◁  utxo)

eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState.hs
```diff
@@ -909,9 +910,13 @@ consumed ::
...
   Core.Value era
-consumed pp u tx =
-  balance @era (eval (txins @era tx ◁ u)) <> Val.inject (refunds <+> withdrawals)
+consumed pp (UTxO u) tx =
+  {- balance (txins tx ◁ u) + wbalance (txwdrls tx) + keyRefunds pp tx -}
+  Set.foldl' lookupAddTxOut mempty (txins @era tx)
+    <> Val.inject (refunds <+> withdrawals)
   where
+    lookupAddTxOut acc txin = maybe acc (addTxOut acc) $ SplitMap.lookup txin u
+    addTxOut !b out = getField @"value" out <+> b
```
TimSheard, line 1032: Perhaps like this?

consumed pp u tx =
  balance @era (txins @era tx Split.◁ u) <> Val.inject (refunds <+> withdrawals)

eras/shelley/impl/src/Cardano/Ledger/Shelley/UTxO.hs
```diff
@@ -256,9 +231,15 @@ balance ::
...
-    addTxOuts !b out = getField @"value" out <+> b
+balance = sumTxOutValue . unUTxO
+
+-- | Sum all the value in TxOut
+sumTxOutValue ::
+  forall era f.
+  (Foldable f, Era era) =>
+  f (Core.TxOut era) ->
+  Core.Value era
+sumTxOutValue = foldMap (getField @"value")
```
TimSheard: This function is called a lot. It used to be defined in terms of a foldl' ,do we have a foldl' for SplitMap. if not perhaps we should write one. Not sure how foldMap behaves. foldl' is strict and very fast.

eras/shelley/test-suite/src/Test/Cardano/Ledger/Shelley/Generator/Core.hs
```diff
@@ -725,50 +700,13 @@ genesisAccountState =
...
-    }
-  where
-    dels = _delegationState ls
-    dst = _dstate dels
-    us = _utxoState ls
-    newAccounts =
-      reapRewards
-        ((_rewards . _dstate . _delegationState) ls)
-        (Map.mapKeys getRwdCred . unWdrl $ getField @"wdrls" tx)
+    SplitMap.fromList [(TxIn genesisTxId idx, out) | (idx, out) <- zip [0 ..] outs]
```
TimSheard, line 710: I don't understand what is happening here.  What did we replace the lines 733-771 with?

libs/compact-map/src/Data/Compact/KeyMap.hs
```diff
@@ -555,6 +565,23 @@ mapWithKey f (Full arr) = Full (fmap (mapWithKey f) arr)
 instance Functor KeyMap where
   fmap f x = mapWithKey (\_ v -> f v) x
 
+instance Foldable KeyMap where
+  length = size
+  foldr f = foldWithDescKey (const f)
+
```
TimSheard: We really should have a strict foldl' functiob

TimSheard (commented):
Every thing looks good. The set algebra function function  (◁) where the value on the right is a SpitMap is use so many times. We should make a version ins SpitMap, and then use

(set  Split.◁  splitmap) this maintains continuity with the spec and the older code.

libs/compact-map/src/Data/Compact/KeyMap.hs
lehins: Strict `foldl'` is given to us by `Foldable`, this was just a minimal implementation needed. In fact, because `foldWithDescKey` is strict even the lazy `foldr` is currently strict, so what we really need is a lazy foldr

I'd like to create a second PR with optimizations to `SplitMap`. Because it will require additions of bunch of benchmarks, I'd rather do it as a separate PR instead of bloating this one even more

eras/shelley/impl/src/Cardano/Ledger/Shelley/UTxO.hs
lehins: There is `foldMap'` too (which uses `foldl'` underneath), but all my prior benchmarks revealed that ghc is able to optimize `foldMap` just fine. In fact that is the default implementation for `sum` function in base :shrug: 

In any case, I'll change it to `foldMap'`, but at some point I'd like to write a benchmark for it, just to be 100% sure that we are using the optimal implementation, since it is a function that is used quite often.

eras/shelley/test-suite/src/Test/Cardano/Ledger/Shelley/Generator/Core.hs
lehins: `applyTxBody` is dead code. It is not used anywhere in the codebase, thus I removed it

JaredCorduan: great catch, thank you @lehins !

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Rules/Utxo.hs
lehins: We can't do that, because `dom utxo` is O(n) on a huge UTxO.

eras/alonzo/test-suite/src/Test/Cardano/Ledger/Alonzo/AlonzoEraGen.hs
lehins: This is a test suite, so we don't need to match any specs here. I'll leave it as is.

eras/alonzo/test-suite/src/Test/Cardano/Ledger/Alonzo/AlonzoEraGen.hs
lehins: Same here, I don't think we need to try and use same unicode operator in the test suite.

JaredCorduan (approved):
LGTM! I think it would be nice to make the special case domain restriction function like @TimSheard suggests.

eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState.hs
lehins: This operation `txins @era tx Split.◁ u)` will create an intermediate `Set`, while doing a fold on `txins` avoids that

SplitMap will already introduce a regression in performance, we need to hold on to any saving we can get.

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Rules/Utxow.hs
lehins: Done

libs/compact-map/src/Data/Compact/KeyMap.hs
lehins: Done

eras/shelley/impl/src/Cardano/Ledger/Shelley/UTxO.hs
lehins: I added some benchmarks  for this function.

You where right, for `SplitMap` it actually made a big difference. `foldMap` was 3 times slower than then the strict variant `foldMap'`. Interesting part is that it was slower only because of bad implementation of lazy `foldr`. Once I fixed it then the difference became marginal as it is for list and `Map`. This is what I was referring to about that being my prior experience with `foldMap`. 

(NB. Definition of strict `foldMap' f = foldl' (\ acc a -> acc <> f a) mempty`)

PR #2593, opened 2021-12-22
weaken `FromCBOR` for `TxOut` to `Annotator TxOut`

goolord:
required for https://github.com/input-output-hk/cardano-ledger/pull/2560

babbage's `TxOut` will only be an instance of `FromCBOR (Annotator TxOut)`, so this pr adds instances for `FromCBOR (Annotator..` as it bubbles up from `TxOut` and `FromCBOR` constraints where necessary in code that's reused in babbage

### DIFF
--- a/eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState.hs
+++ b/eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState.hs
@@ -244,10 +244,12 @@ import Control.Provenance (ProvM, liftProv, modifyM)
 import Control.SetAlgebra (Bimap, biMapEmpty, dom, eval, forwards, (∈), (∪+), (▷), (◁))
 import Control.State.Transition (STS (State))
 import Data.Coders
-  ( Decode (From, RecD),
+  ( Annotator (..),
+    Decode (Ann, From, RecD),
     decode,
     decodeRecordNamed,
     (<!),
+    (<*!),
   )
 import qualified Data.Compact.VMap as VMap
 import Data.Constraint (Constraint)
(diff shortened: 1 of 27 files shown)


lehins (changes requested):
I highly recommend avoiding all the copy pasting. Most of the code added in this PR can be avoided. For example:

```haskell
instance ... => FromCBOR (Annotator (EpochState era)) where
  fromCBOR = do
    es@EpochState {eslState} <- fromCBOR
    pure $ Annotator $ \fbs -> es {esLState = runAnnotator esLState fbs}
```

It is especially dangerous to copy serialization/deserialziation code like that, because it s possible that someone in the future will change one `fromCBOR`, but not the other other one.

Also it is very bad practice to copy paste code in general, it is almost always better to create an abstraction. A professor of mine said once: "If you are finding yourself copying a piece of code, it is most likely that you are doing something wrong." 

So please, keep your code as [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) as possible.

goolord:
I don't think the code in your example would typecheck. since the constraints are different the instances have to be duplicated, or we can remove the non-Annotator FromCBOR instances, but not both

goolord:
i think i have a couple of ideas wrt getting code reuse here actually, I'll try them out after Christmas ⛄🎄

goolord:
actually, I just realized that if `Ann` and `ApplyAnn` are generalized to any `Applicative`, all of these instances can share the same decoding function, so i'm going to do that ☃️

JaredCorduan (approved):
Adding those annotators do tend to infect a lot of code, thanks for plowing through this! LGTM

eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState.hs
```diff
@@ -506,6 +514,40 @@ instance (TransEpoch ToCBOR era) => ToCBOR (EpochState era) where
...
+  ) =>
+  FromCBOR (Annotator (EpochState era))
+  where
+  fromCBOR = decodeEpochState
+
+decodeEpochState ::
+  ( FromCBOR (Core.PParams era),
+    TransValue FromCBOR era,
+    FromSharedCBOR (f (LedgerState era)),
+    -- Share (f (Core.TxOut era)) ~ Interns (Credential 'Staking (Crypto era)),
```
lehins:
```suggestion
```

eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState.hs
```diff
@@ -662,6 +695,39 @@ instance TransUTxOState ToCBOR era => ToCBOR (UTxOState era) where
...
+    FromCBOR (State (Core.EraRule "PPUP" era)),
+    FromSharedCBOR (Annotator (Core.TxOut era)),
+    Share (Annotator (Core.TxOut era)) ~ Interns (Credential 'Staking (Crypto era)),
+    HashAnnotated (Core.TxBody era) EraIndependentTxBody (Crypto era)
+  ) =>
+  FromSharedCBOR (Annotator (UTxOState era))
+  where
+  type
+    Share (Annotator (UTxOState era)) =
+      Interns (Credential 'Staking (Crypto era))
```
lehins: I think it would be more appropriate to use UTXOState, but it would not be incorrect to leave it as it is right now too. So feel free to ignore this suggestion if you'd like
```suggestion
    Share (Annotator (UTxOState era)) = Share (UTxOState era)
```

lehins (approved):
This looks great. @goolord thank you for reducing duplication!

There are just a couple of minor suggestions that I have in this review.

PR #2602, opened 2022-01-05
Babbage: PParams

### LINKED ISSUE #2464: Babbage PParams
Implement the new Babbage `PParams` type and all necessary instances.

Babbage *removes* two protocol parameters:
  * d
  * extra entropy

goolord:
resolves https://github.com/input-output-hk/cardano-ledger/issues/2464
round tripping tests pass on https://github.com/input-output-hk/cardano-ledger/tree/zachc/babbage-big

### DIFF
--- a/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/PParams.hs
+++ b/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/PParams.hs
@@ -57,6 +57,7 @@ import Cardano.Ledger.BaseTypes
   )
 import qualified Cardano.Ledger.BaseTypes as BT (ProtVer (..))
 import Cardano.Ledger.Coin (Coin (..))
+import qualified Cardano.Ledger.Core as Core
 import Cardano.Ledger.Era
 import Cardano.Ledger.Serialization
   ( FromCBORGroup (..),
@@ -67,6 +68,7 @@ import Cardano.Ledger.Shelley.Orphans ()
 import Cardano.Ledger.Shelley.PParams (HKD)
 import qualified Cardano.Ledger.Shelley.PParams as Shelley (PParams' (..))
 import Cardano.Ledger.Slot (EpochNo (..))
+import Cardano.Prelude (HasField (getField))
 import Control.DeepSeq (NFData)
 import Data.ByteString (ByteString)
 import qualified Data.ByteString as BS
@@ -294,33 +296,33 @@ emptyPParams =
 -- IF THE ORDER OR TYPES OF THE FIELDS OF PParams changes, this instance may need adusting.
 instance Ord (PParams' StrictMaybe era) where
   compare x y =
-       compare (_minfeeA x) (_minfeeA y)
-    <> compare (_minfeeB x) (_minfeeB y)
-    <> compare (_maxBBSize x) (_maxBBSize y)
-    <> compare (_maxTxSize x) (_maxTxSize y)
-    <> compare (_maxBHSize x) (_maxBHSize y)
-    <> compare (_keyDeposit x) (_keyDeposit y)
-    <> compare (_poolDeposit x) (_poolDeposit y)
-    <> compare (_eMax x) (_eMax y)
-    <> compare (_nOpt x) (_nOpt y)
-    <> compare (_a0 x) (_a0 y)
-    <> compare (_rho x) (_rho y)
-    <> compare (_tau x) (_tau y)
-    <> compare (_d x) (_d y)
-    <> compare (_extraEntropy x) (_extraEntropy y)
-    <> compare (_protocolVersion x) (_protocolVersion y)
-    <> compare (_minPoolCost x) (_minPoolCost y)
-    <> compare (_coinsPerUTxOWord x) (_coinsPerUTxOWord y)
-    <> compare (_costmdls x) (_costmdls y)
-    <> compare (_prices x) (_prices y)
-    <> ( case compareEx (_maxTxExUnits x) (_maxTxExUnits y) of
-              LT -> LT
-              GT -> GT
-              EQ -> case compareEx (_maxBlockExUnits x) (_maxBlockExUnits y) of
-                LT -> LT
-                GT -> GT
-                EQ -> compare (_maxValSize x) (_maxValSize y)
-       )
+    compare (_minfeeA x) (_minfeeA y)
+      <> compare (_minfeeB x) (_minfeeB y)
+      <> compare (_maxBBSize x) (_maxBBSize y)
+      <> compare (_maxTxSize x) (_maxTxSize y)
+      <> compare (_maxBHSize x) (_maxBHSize y)
+      <> compare (_keyDeposit x) (_keyDeposit y)
+      <> compare (_poolDeposit x) (_poolDeposit y)
+      <> compare (_eMax x) (_eMax y)
+      <> compare (_nOpt x) (_nOpt y)
+      <> compare (_a0 x) (_a0 y)
+      <> compare (_rho x) (_rho y)
+      <> compare (_tau x) (_tau y)
+      <> compare (_d x) (_d y)
+      <> compare (_extraEntropy x) (_extraEntropy y)
+      <> compare (_protocolVersion x) (_protocolVersion y)
+      <> compare (_minPoolCost x) (_minPoolCost y)
+      <> compare (_coinsPerUTxOWord x) (_coinsPerUTxOWord y)
+      <> compare (_costmdls x) (_costmdls y)
+      <> compare (_prices x) (_prices y)
+      <> ( case compareEx (_maxTxExUnits x) (_maxTxExUnits y) of
+             LT -> LT
+             GT -> GT
+             EQ -> case compareEx (_maxBlockExUnits x) (_maxBlockExUnits y) of
+               LT -> LT
+               GT -> GT
+               EQ -> compare (_maxValSize x) (_maxValSize y)
+         )
 
 compareEx :: StrictMaybe ExUnits -> StrictMaybe ExUnits -> Ordering
 compareEx SNothing SNothing = EQ
@@ -491,7 +493,8 @@ data LangDepView = LangDepView {tag :: ByteString, params :: ByteString}
 
 getLanguageView ::
   forall era.
-  PParams era ->
+  (HasField "_costmdls" (Core.PParams era) (Map.Map Language CostModel)) =>
+  Core.PParams era ->
   Language ->
   LangDepView
 getLanguageView pp lang@PlutusV1 =
@@ -500,7 +503,7 @@ getLanguageView pp lang@PlutusV1 =
     ( serialize'
         ( serializeEncoding' $
             maybe encodeNull enc $
-              Map.lookup lang (_costmdls pp)
+              Map.lookup lang (getField @"_costmdls" pp)
         )
     )
   where
@@ -508,7 +511,7 @@ getLanguageView pp lang@PlutusV1 =
 getLanguageView pp lang@PlutusV2 =
   LangDepView
     (serialize' lang)
-    (serializeEncoding' $ maybe encodeNull toCBOR $ Map.lookup lang (_costmdls pp))
+    (serializeEncoding' $ maybe encodeNull toCBOR $ Map.lookup lang (getField @"_costmdls" pp))
 
 encodeLangViews :: Set LangDepView -> Encoding
 encodeLangViews views = encodeMapLen n <> foldMap encPair ascending
--- a/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/PParams.hs
+++ b/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/PParams.hs
@@ -488,7 +488,7 @@ data LangDepView = LangDepView {tag :: ByteString, params :: ByteString}
 
 getLanguageView ::
   forall era.
-  (HasField "_costmdls" (Core.PParams era) (Map.Map Language CostModel)) =>
+  (HasField "_costmdls" (Core.PParams era) (Map Language CostModel)) =>
   Core.PParams era ->
   Language ->
   LangDepView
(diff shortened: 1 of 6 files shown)


eras/babbage/impl/src/Cardano/Ledger/Babbage/PParams.hs
```diff
@@ -0,0 +1,533 @@
...
+  price
+  mxTx
+  mxBl
+  mxV
+  col
+  mxCol =
+    PParams ma mb mxBB mxT mxBH kd pd emx a n rho tau pv mnP ada cost price mxTx mxBl mxV col mxCol
+
+instance HasField "_d" (PParams era) UnitInterval where
+  getField _ = minBound
```
goolord, line 469: is this sensible? it just required the least change to satisfy constraints on https://github.com/input-output-hk/cardano-ledger/tree/zachc/babbage-big. i chose `minBound` because that's what was chose in `emptyPParams`

JaredCorduan: do you know what is requiring the `HasField` instance? The link just points to a branch. I suspect it is the reward calculation, it which case having this spurious instance might be the path of least resistance, saving us from coping a lot of reward calc code. and yea, minBound would be the best value.

goolord:
```
src/Cardano/Ledger/Babbage.hs:121:10: error:
    • Could not deduce (GHC.Records.HasField
                          "_d"
                          (PParams (BabbageEra c))
                          Cardano.Ledger.BaseTypes.UnitInterval)
        arising from the superclasses of an instance declaration
      from the context: API.ShelleyEraCrypto c
        bound by the instance declaration
        at src/Cardano/Ledger/Babbage.hs:121:10-64
    • In the instance declaration for 'API.ApplyBlock (BabbageEra c)'
    |
121 | instance API.ShelleyEraCrypto c => API.ApplyBlock (BabbageEra c)
    |          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
some superclass constraint of `ApplyBlock`, i can't really figure out where this constraint arises.

eras/babbage/impl/src/Cardano/Ledger/Babbage/PParams.hs
```diff
@@ -0,0 +1,533 @@
...
+import qualified Data.Set as Set
+import GHC.Generics (Generic)
+import GHC.Records (HasField (..))
+import NoThunks.Class (NoThunks (..))
+import Numeric.Natural (Natural)
+
+type PParamsUpdate era = PParams' StrictMaybe era
+
+-- | Protocol parameters.
+-- Shelley parameters + additional ones
```
JaredCorduan:
```suggestion
-- Alonzo parameters without d and extraEntropy
```

eras/babbage/impl/src/Cardano/Ledger/Babbage/PParams.hs
```diff
@@ -0,0 +1,533 @@
...
+              GT -> GT
+              EQ -> case compareEx (_maxBlockExUnits x) (_maxBlockExUnits y) of
+                LT -> LT
+                GT -> GT
+                EQ -> (_maxValSize x, _maxValSize y) <== EQ
+          )
+
+infixr 4 <==
+
+(<==) :: Ord a => (a, a) -> Ordering -> Ordering
```
JaredCorduan: I wonder if we can find a good home for this function, Alonzo uses it as well.

eras/babbage/impl/src/Cardano/Ledger/Babbage/PParams.hs
```diff
@@ -0,0 +1,533 @@
...
+      _prices = fromSMaybe (_prices pp) (_prices ppup),
+      _maxTxExUnits = fromSMaybe (_maxTxExUnits pp) (_maxTxExUnits ppup),
+      _maxBlockExUnits = fromSMaybe (_maxBlockExUnits pp) (_maxBlockExUnits ppup),
+      _maxValSize = fromSMaybe (_maxValSize pp) (_maxValSize ppup),
+      _collateralPercentage = fromSMaybe (_collateralPercentage pp) (_collateralPercentage ppup),
+      _maxCollateralInputs = fromSMaybe (_maxCollateralInputs pp) (_maxCollateralInputs ppup)
+    }
+
+-- ===================================================
+-- Figure 1: "Definitions Used in Protocol Parameters"
```
JaredCorduan: we can probably remove this comment, we don't know what figure it'l be in the babbage spec yet

eras/babbage/impl/src/Cardano/Ledger/Babbage/PParams.hs
```diff
@@ -0,0 +1,533 @@
...
+        )
+    )
+  where
+    enc (CostModel cm) = encodeFoldableAsIndefinite $ Map.elems cm
+getLanguageView pp lang@PlutusV2 =
+  LangDepView
+    (serialize' lang)
+    (serializeEncoding' $ maybe encodeNull toCBOR $ Map.lookup lang (_costmdls pp))
+
+encodeLangViews :: Set LangDepView -> Encoding
```
JaredCorduan: this is another function that is already defined in Alonzo, we should reuse it

eras/babbage/impl/src/Cardano/Ledger/Babbage/PParams.hs
```diff
@@ -0,0 +1,533 @@
...
+    ascending = sortBy (shortLex `on` tag) $ Set.toList views
+    encPair (LangDepView k v) = encodePreEncoded k <> encodePreEncoded v
+    shortLex :: ByteString -> ByteString -> Ordering
+    shortLex a b
+      | BS.length a < BS.length b = LT
+      | BS.length a > BS.length b = GT
+      | otherwise = compare a b
+
+-- | Turn an PParams' into a Shelley.Params'
+retractPP :: HKD f Coin -> HKD f UnitInterval -> HKD f Nonce -> PParams' f era2 -> Shelley.PParams' f era1
```
JaredCorduan: ditto

eras/babbage/impl/src/Cardano/Ledger/Babbage/PParams.hs
JaredCorduan: @TimSheard said he'd help investigate this. If it turns out it is needed (which is possible, the reward calc might be forcing it somehow) this this instance is good, but we should document why it is needed and the fact that it's a bit strange to have a HasField instance for a field it doesn't have...

JaredCorduan:
Overall looks great, I made a few minor suggestions, and I'd still like to figure out what is going on with the `HasField` for `d`.

eras/babbage/impl/src/Cardano/Ledger/Babbage/PParams.hs
lehins: We don'y really need this function, this is equivalent to `<>` on `Ordering`.

eras/babbage/impl/src/Cardano/Ledger/Babbage/PParams.hs
```diff
@@ -0,0 +1,533 @@
...
+      <== (_costmdls x, _costmdls y)
+      <== (_prices x, _prices y)
+      <== ( case compareEx (_maxTxExUnits x) (_maxTxExUnits y) of
+              LT -> LT
+              GT -> GT
+              EQ -> case compareEx (_maxBlockExUnits x) (_maxBlockExUnits y) of
+                LT -> LT
+                GT -> GT
+                EQ -> (_maxValSize x, _maxValSize y) <== EQ
+          )
```
lehins: It is more natural to use `Monoid` instance. Something along these lines
```suggestion
  compare x y = foldMap (uncurry compare) [
    (_minfeeA x, _minfeeA y)
    (_minfeeB x, _minfeeB y),
    (_maxBBSize x, _maxBBSize y),
    (_maxTxSize x, _maxTxSize y),
    (_maxBHSize x, _maxBHSize y),
    (_keyDeposit x, _keyDeposit y),
    (_poolDeposit x, _poolDeposit y),
    (_eMax x, _eMax y),
    (_nOpt x, _nOpt y),
    (_a0 x, _a0 y),
    (_rho x, _rho y),
    (_tau x, _tau y),
    (_protocolVersion x, _protocolVersion y),
    (_minPoolCost x, _minPoolCost y),
    (_coinsPerUTxOWord x, _coinsPerUTxOWord y),
    (_costmdls x, _costmdls y),
    (_prices x, _prices y)] <>
    ( compareEx (_maxTxExUnits x) (_maxTxExUnits y)
     <> compareEx (_maxBlockExUnits x) (_maxBlockExUnits y)
     <> compare (_maxValSize x) (maxValSize y)),
```

goolord: had to unfold this `foldMap` since `[]` is homogeneous, ty tho

eras/babbage/impl/src/Cardano/Ledger/Babbage/PParams.hs
TimSheard: Where is the file Babbage.hs,  I can't find it . Makes it hard to reproduce the problem

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/PParams.hs
```diff
@@ -294,38 +296,33 @@ emptyPParams =
...
+      <> compare (_costmdls x) (_costmdls y)
+      <> compare (_prices x) (_prices y)
+      <> ( case compareEx (_maxTxExUnits x) (_maxTxExUnits y) of
+             LT -> LT
+             GT -> GT
+             EQ -> case compareEx (_maxBlockExUnits x) (_maxBlockExUnits y) of
+               LT -> LT
+               GT -> GT
+               EQ -> compare (_maxValSize x) (_maxValSize y)
+         )
```
lehins: We can simplify here too:
```suggestion
      <> compareEx (_maxTxExUnits x) (_maxTxExUnits y)
      <> compareEx (_maxBlockExUnits x) (_maxBlockExUnits y)
      <> compare (_maxValSize x) (_maxValSize y)
```

> had to unfold this foldMap since [] is homogeneous

That's great, I used `foldMap` just to show the idea, using `<>` is even better

eras/babbage/impl/src/Cardano/Ledger/Babbage/PParams.hs
```diff
@@ -0,0 +1,474 @@
...
+import GHC.Generics (Generic)
+import GHC.Records (HasField (..))
+import NoThunks.Class (NoThunks (..))
+import Numeric.Natural (Natural)
+
+type PParamsUpdate era = PParams' StrictMaybe era
+
+-- | Protocol parameters.
+-- Alonzo parameters without d and extraEntropy
+data PParams' f era = PParams
```
lehins, line 92: @JaredCorduan, this might have been asked before, but why do we need `era` parameter here? It isn't used in any of the fields

eras/babbage/impl/src/Cardano/Ledger/Babbage/PParams.hs
goolord: this function is already as meaningfully abstract as it can be i think

eras/babbage/impl/src/Cardano/Ledger/Babbage/PParams.hs
goolord: it's on the branch i linked above https://github.com/input-output-hk/cardano-ledger/tree/zachc/babbage-big

eras/babbage/impl/src/Cardano/Ledger/Babbage/PParams.hs
JaredCorduan: ah right, you can not use the Alonzo one because of `d` and `extraEntropy`, makes sense now!

eras/babbage/impl/src/Cardano/Ledger/Babbage/PParams.hs
JaredCorduan: @lehins that's a good point, I don't see any reason for it (and similarly for the previous eras). I don't know if this is vestigial, or if we just originally era-parameterized all the types by convention.

lehins: Very good, I suggest we remove it, at least for Babbage for sure.

JaredCorduan: sounds good to me

eras/babbage/impl/src/Cardano/Ledger/Babbage/PParams.hs
JaredCorduan: @goolord I took a look at all the functions that need `_d`. It's going to be a lot of work to disentangle `_d`, and potentially add a lot of code duplication. I think what you've done, globally setting `d = 0` for Babbage, is the way to go. Let's document the `HasField "_d"` instance with a comment that there really is no such field, that we are just setting the value to zero since the reward calculation uses it. It is weird if I would rather set it to `0` than `minBound`? just seems more explicit...

eras/babbage/impl/src/Cardano/Ledger/Babbage/PParams.hs
nc6: I thought it got added via `CostModel` at some point

eras/babbage/impl/src/Cardano/Ledger/Babbage/PParams.hs
goolord: `UnitInterval` doesn't have its constructor exported and only provides `Bounded` and `BoundedRational` as smart constructors afaict

eras/babbage/impl/src/Cardano/Ledger/Babbage/PParams.hs
goolord: ah, this is why
image
> If a LHS type variable that is declared as injective is not mentioned on
injective position in the RHS we conclude that the type family is not
injective.  By "injective position" we mean argument to a type constructor or
argument to a type family on injective position.
```hs
type family PParams era = (r :: Type) | r -> era
```
i don't think the injectivity annotation is strictly necessary either

eras/babbage/impl/src/Cardano/Ledger/Babbage/PParams.hs
JaredCorduan: gotcha :+1:

JaredCorduan (approved):

PR #2605, opened 2022-01-06
Make header in `Block` a fully saturated type.

nc6:
This is mostly needed to allow injectivity when trying to vary the
header type with the protocol in consensus.

Almost all changes here are mechanical - the root of the change is at https://github.com/input-output-hk/cardano-ledger/compare/nc/bheader?expand=1#diff-85dc0c0efc44545250fd1786e3d95f10603f2ef8953361439adb99e908efb2ddL53


JaredCorduan (approved):
LGTM!

PR #2608, opened 2022-01-06
Moved the Coders tests to Test.Data.Coders

TimSheard:
Moved the Coders tests from shelley-ma-test to Test.Data,Coders .
We also extracted a few operators that deal with writing CBOR roundtrip tests in to the module  Data.Roundtrip. These are used in about 5 places, and are now accessible without requiring dependencies on things like Test.Cardano.Ledger.ShelleyMA


lehins (approved):
Ideally things like `Data.Roundtrip`, which are used in various test suites should be placed in some package that contains all the common stuff needed for testing. Unfortunately we currently don't have a package like that, so maybe in the future. I think cardano-ledger-core-test might have been intended for that, but that is not what it is.

In any case this change looks great!

PR #2609, opened 2022-01-10
Added domainRestrictedView

TimSheard:
Added the function domainRestrictedView.
Also made a change to unUnify, and this was able to remove some unnecessary constraints.


nc6 (approved):
Thanks for adding this!

PR #2611, opened 2022-01-11
Minor improvements

lehins:
This is a tiny PR that adds some bangs on data types that seem to have them missing.

It also simplifies recently added code just a tiny bit.

Please ignore changes to benchmark suite, they aren't important.

### DIFF
--- a/eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState.hs
+++ b/eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState.hs
@@ -76,10 +76,8 @@ module Cardano.Ledger.Shelley.LedgerState
     keyRefunds,
 
     -- * Epoch boundary
-    stakeDistr,
     incrementalStakeDistr,
     updateStakeDistribution,
-    aggregateUtxoCoinByCredential,
     applyRUpd,
     applyRUpd',
     createRUpd,
@@ -245,7 +243,7 @@ import Control.DeepSeq (NFData)
 import Control.Monad.State.Strict (evalStateT)
 import Control.Monad.Trans
 import Control.Provenance (ProvM, liftProv, modifyM)
-import Control.SetAlgebra (dom, eval, (∈), (▷), (◁))
+import Control.SetAlgebra (dom, eval, (∈), (◁))
 import Control.State.Transition (STS (State))
 import Data.Coders
   ( Decode (From, RecD),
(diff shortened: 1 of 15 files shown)


JaredCorduan (approved):
LGTM!

eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState.hs
```diff
@@ -1403,25 +1346,18 @@ resolveActiveIncrementalPtrs isActive ptrMap (IStake credStake ptrStake) =
...
-          _ -> Nothing
-    )
+    (\_k trip coin2 -> (<> coin2) <$> UM.tripDelegatedReward trip)
     -- what to do when a key appears just in 'tripmap', we only add the coin if the key is active
-    ( \mp ->
-        let p _key (Triple (SJust c) _ (SJust _)) = Just c
-            p _ _ = Nothing
-         in filterMaybe p mp
-    )
+    (Map.mapMaybeWithKey (\_ -> UM.tripDelegatedReward))
```
TimSheard: Maybe add a comment that tripDelegatedReward returns (Just coin) if there is a reward.
Or maybe change the name so that that fact is implicit in the name.

eras/shelley/test-suite/bench/Main.hs
```diff
@@ -207,31 +209,38 @@ profileCreateRegPools size = do
...
 action2m (dstate, pstate, utxo) = stakeDistr utxo dstate pstate
 
+action2im ::
+  ShelleyTest era =>
+  (DState (Crypto era), PState (Crypto era), UTxO era) ->
+  EB.SnapShot (Crypto era)
+action2im (dstate, pstate, utxo) =
+  let incStake = updateStakeDistribution mempty mempty utxo
+   in incrementalStakeDistr incStake dstate pstate
+
```
TimSheard, line 247: I am unsure what the semantics of "bench" is, is the cost of building  (nf action2im arg) counted? THat might account for the slight increase in  increase in time when using incrementalStakeDistr.  I f that is not the case then I think we have to dig deeper. Something does not look right.

TimSheard (approved):
Looks good, Might want to give more descriptive names to the pseudo-projction functions that test if one of the fields in a Triple is Just or Nothing.

eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState.hs
lehins: Will change to: `tripRewardActiveDelegation`

PR #2614, opened 2022-01-11
Revert #2593, use CBOR in CBOR encoding for Data

goolord:


JaredCorduan (approved):
thanks @goolord , looks good to me.

PR #2616, opened 2022-01-14
Performance of TxOut address

lehins:
This PR moves `CompactAddr` module out of `cardano-ledger-shelley` into `cardano-ledger-core`

Also It improves the performance of getting `Addr` and `CompactAddr` from `TxOut`. In particular for Shelley we store `Addr` in compact form, so not much will change there, but for `Alonzo` onwards most of `TxOut` will have `Addr` stored in a packed form of 4x `Word64`. It is wildly inefficient to serialize and then deserialize the full address (which is what happens now), it is much better to use those words with slight adjustment in order to construct uncompacted `Addr` directly, which is what this PR implements.

Moreover, in `getFilteredUTxO` it would be beneficial to know in which form a particular TxOut has the address stored in order to perform optimal filtering. This PR takes care of that too.


libs/cardano-ledger-core/src/Cardano/Ledger/Era.hs
```diff
@@ -69,6 +68,25 @@ class
   where
   type Crypto e :: Type
 
+  -- | Extract from TxOut either an address or its compact version by doing the
```
nc6, line 71: Doesn't seem the nicest place for these things?

libs/cardano-ledger-core/src/Cardano/Ledger/Era.hs
```diff
@@ -69,6 +68,25 @@ class
...
+  -- least amount of work. Default implementation relies on the "address" field.
```
nc6, line 72: I feel like this comment could be expanded on a little? It's not clear what a correct implementation should do, or why this method is helpful.

nc6 (commented):
Took a quick look (but didn't do a full review) - looks nice, left a couple of minor comments!

libs/cardano-ledger-core/src/Cardano/Ledger/Era.hs
lehins: I though about it carefully and I could not find a better place. We already have super constraint on the `Era` that conceptually requires the same functionality: `HasField "address" (Core.TxOut era) (Addr (Crypto era))`, so it only makes sense to have it here.

The only other place I though about placing it under was `UsesTxOut` class in shelley, but I would really like to eventually get rid of the `getField @"address"` hacks in favor of proper API. And for that to work it has to be in the place where `HasField "address"` is required, i.e. `Era` class.

libs/cardano-ledger-core/src/Cardano/Ledger/Era.hs
lehins: I've added more info explaining the motivation. Let me know if it is still not clear enough

JaredCorduan (approved):
reviewed in a call, looks good to me!

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxBody.hs
```diff
@@ -135,6 +138,14 @@ import GHC.TypeLits
 import NoThunks.Class (InspectHeapNamed (..), NoThunks)
 import Prelude hiding (lookup)
 
+data Addr28Extra
+  = Addr28Extra
+      {-# UNPACK #-} !Word64 -- Payment Addr
+      {-# UNPACK #-} !Word64 -- Payment Addr
+      {-# UNPACK #-} !Word64 -- Payment Addr
+      {-# UNPACK #-} !Word64 -- Payment Addr (32bits) + ... +  0/1 for Testnet/Mainnet + 0/1 Script/Pubkey
+  deriving (Eq)
```
DavidEichmann, line 147: A nice cleanup

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxBody.hs
```diff
@@ -135,6 +138,14 @@ import GHC.TypeLits
...
+
 data TxOut era
```
DavidEichmann, line 149: BTW, the Eq instance below is incorrect now that we export all constructors (i.e. we don't enforce a canonical form), correct? Though this is probably out of scope for this PR.

DavidEichmann (commented):
Looking good. Just one question: have you done some benchmarks on the `getFilteredUTxO`?

lehins:
> Just one question:  have you done some benchmarks on the getFilteredUTxO?

@DavidEichmann One answer: No I have not :smile: But with all honesty I would be shocked if this was not a speedup! I just don't know how much of a speedup.

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxBody.hs
lehins: Quite the opposite. This PR enforces just that by introducing this `Addr28Extra` data type that does not have the constructor exported, which means it can be only created by the `encodeAddress28` function, which is canonical.

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxBody.hs
lehins: Thank you :+1:

PR #2617, opened 2022-01-18
Plutus V2 to fail phase 1 validation on Byron addr

### LINKED ISSUE #2612: fail phase 1 validation on the presence of byron addresses
Byron addresses are current ignored by Plutus, in the sense that they are not a part of the context passed to Plutus. We would prefer to have the transaction fail phase 1 validation in the event that a transaction contains both a plutus script and any byron addresses (both inputs and outputs).

JaredCorduan:
In Plutus V1, we do not pass Byron addresses to the transaction context which is passed to Plutus, they are silently omitted. This includes both transaction inputs and outputs. Starting at major protocol version 7, we want to instead cause a phase 1 validation error instead. This PR does exactly that.

Additionally, this PR introduces a policy change for the `TxInfo` creation code: any time it sees something it can't deal with it forces a phase 1 validation error instead of ignoring it. In particular, two new logic errors are introduced, one for unknown transaction inputs (this is a logic error since the ledger already checks this) and one for unknown redeemer pointers (also something that the ledger already checks).

Note that with the introduction of the babbage, it will probably not make sense for alonzo to have logic for Plutus V2. This is because the Plutus V2 context will probably contain a new field, namely reference inputs (as opposed to just combining them with the usual inputs). We can adapt accordingly when the time comes.

closes #2612

### DIFF
--- a/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxInfo.hs
+++ b/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxInfo.hs
@@ -20,7 +20,7 @@ import Cardano.Ledger.Alonzo.Language (Language (..))
 import Cardano.Ledger.Alonzo.Scripts (CostModel (..), ExUnits (..), Script (..), decodeCostModel)
 import Cardano.Ledger.Alonzo.Tx
 import Cardano.Ledger.Alonzo.TxWitness (RdmrPtr, TxWitness (..), unRedeemers, unTxDats)
-import Cardano.Ledger.BaseTypes (ProtVer, StrictMaybe (..), certIxToInt, txIxToInt)
+import Cardano.Ledger.BaseTypes (ProtVer (..), StrictMaybe (..), certIxToInt, txIxToInt)
 import Cardano.Ledger.Coin (Coin (..))
 import Cardano.Ledger.Core as Core (PParams, TxBody, TxOut, Value)
 import Cardano.Ledger.Credential
@@ -35,8 +35,6 @@ import Cardano.Ledger.Keys (KeyHash (..), KeyRole (Witness), hashKey)
 import qualified Cardano.Ledger.Mary.Value as Mary (AssetName (..), PolicyID (..), Value (..))
 import Cardano.Ledger.SafeHash
 import qualified Cardano.Ledger.Shelley.HardForks as HardForks
-  ( translateTimeForPlutusScripts,
-  )
 import Cardano.Ledger.Shelley.Scripts (ScriptHash (..))
 import Cardano.Ledger.Shelley.TxBody
   ( DCert (..),
@@ -70,9 +68,9 @@ import Data.Coders
     (<!),
   )
 import qualified Data.Compact.SplitMap as SplitMap
+import Data.Either (rights)
 import Data.Fixed (HasResolution (resolution))
 import qualified Data.Map as Map
-import Data.Maybe (mapMaybe)
 import Data.Sequence.Strict (StrictSeq)
 import Data.Set (Set)
 import qualified Data.Set as Set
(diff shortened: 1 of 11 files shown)


eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxInfo.hs
```diff
@@ -195,13 +218,13 @@ txInfoIn ::
...
 txInfoIn (UTxO mp) txin =
   case SplitMap.lookup txin mp of
-    Nothing -> Nothing
-    Just txout -> case transTxOutAddr txout of
-      Just ad -> Just (PV1.TxInInfo (txInfoIn' txin) (PV1.TxOut ad valout dhash))
-      Nothing -> Nothing
+    Nothing -> Left TranslationLogicErrorInput
+    Just txout -> case transTxOutAddr txout ByronInputInContext of
+      Right ad -> Right (PV1.TxInInfo (txInfoIn' txin) (PV1.TxOut ad valout dhash))
+      Left err -> Left err
```
lehins: I think it would be clearer to leave `transTxOutAddr` function as is and make it return `Nothing`, while at call sites:
```suggestion
    Just txout -> case transTxOutAddr txout of
      Just ad -> Right (PV1.TxInInfo (txInfoIn' txin) (PV1.TxOut ad valout dhash))
      Nothing -> Left ByronInputInContext
```

Reason for this suggestion is `transTxOutAddr` can fail for one reason only, thus supplying an error to it is counterproductive and just adds complexity.

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxInfo.hs
```diff
@@ -127,15 +147,18 @@ transCred (KeyHashObj (KeyHash kh)) =
 transCred (ScriptHashObj (ScriptHash kh)) =
   PV1.ScriptCredential (PV1.ValidatorHash (PV1.toBuiltin (hashToBytes kh)))
 
-transAddr :: Addr crypto -> Maybe PV1.Address
-transAddr (Addr _net object stake) = Just (PV1.Address (transCred object) (transStakeReference stake))
-transAddr (AddrBootstrap _bootaddr) = Nothing
+transAddr :: Addr crypto -> TranslationError -> Either TranslationError PV1.Address
+transAddr (Addr _net object stake) _ = Right (PV1.Address (transCred object) (transStakeReference stake))
+transAddr (AddrBootstrap _bootaddr) err = Left err
```
lehins: As suggested in another comment, it is more idiomatic to return `Maybe` instead of passing an error as an argument

```suggestion
transAddr :: Addr crypto -> Maybe PV1.Address
transAddr (Addr _net object stake) = Just (PV1.Address (transCred object) (transStakeReference stake))
transAddr (AddrBootstrap _bootaddr) = Nothing
```

It is always possible to `maybe (Left err) Right . transAddr`

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxInfo.hs
```diff
@@ -127,15 +147,18 @@ transCred (KeyHashObj (KeyHash kh)) =
...
 
-transTxOutAddr :: Era era => TxOut era -> Maybe PV1.Address
-transTxOutAddr txOut = do
+transTxOutAddr :: Era era => TxOut era -> TranslationError -> Either TranslationError PV1.Address
+transTxOutAddr txOut err = do
```
lehins:
```suggestion
transTxOutAddr :: Era era => TxOut era -> Maybe PV1.Address
transTxOutAddr txOut = do
```

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxInfo.hs
```diff
@@ -127,15 +147,18 @@ transCred (KeyHashObj (KeyHash kh)) =
...
   -- filter out Byron addresses without uncompacting them
-  Nothing <- Just (getTxOutBootstrapAddress txOut)
-  transAddr $ getTxOutAddr txOut
+  case getTxOutBootstrapAddress txOut of
+    Just _ -> Left err
+    -- The Nothing case will not actually return Left, since the presence
+    -- of a Byron address is caught above in the Just case
+    Nothing -> transAddr (getTxOutAddr txOut) err
```
lehins:
```suggestion
  case getTxOutBootstrapAddress txOut of
    Just _ -> Nothing
    -- The presence of a Byron address is caught above in the Just case
    Nothing -> transAddr (getTxOutAddr txOut) err
```

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxInfo.hs
```diff
@@ -218,13 +241,13 @@ txInfoOut ::
...
+  Either TranslationError PV1.TxOut
 txInfoOut txout =
   let val = getField @"value" txout
       datahash = getField @"datahash" txout
-   in case transTxOutAddr txout of
-        Just ad -> Just (PV1.TxOut ad (transValue @(Crypto era) val) (transDataHash datahash))
-        Nothing -> Nothing
+   in case transTxOutAddr txout ByronOutputInContext of
+        Right ad -> Right (PV1.TxOut ad (transValue @(Crypto era) val) (transDataHash datahash))
+        Left err -> Left err
```
lehins:
```suggestion
   in case transTxOutAddr txout of
        Just ad -> Right (PV1.TxOut ad (transValue @(Crypto era) val) (transDataHash datahash))
        Nothing -> Left ByronOutputInContext
```

lehins (changes requested):
One recommendation, otherwise looks great.

JaredCorduan:
@lehins your suggestion was great, thanks!

lehins (approved):

PR #2618, opened 2022-01-18
Babbage: Era mapping

goolord:
this adds `Cardano.Ledger.Babbage` + some small modules with mostly just reexport Alonzo stuff, and it also makes some alonzo rules era polymorphic so we can reuse them for babbage

### DIFF
--- a/eras/babbage/impl/src/Cardano/Ledger/Babbage.hs
+++ b/eras/babbage/impl/src/Cardano/Ledger/Babbage.hs
@@ -1,0 +1,281 @@
+{-# LANGUAGE DataKinds #-}
+{-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE FlexibleInstances #-}
+{-# LANGUAGE MultiParamTypeClasses #-}
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeApplications #-}
+{-# LANGUAGE TypeFamilies #-}
+{-# LANGUAGE UndecidableInstances #-}
+
+module Cardano.Ledger.Babbage
+  ( BabbageEra,
+    Self,
+    TxOut,
+    Value,
+    TxBody,
+    Script,
+    AuxiliaryData,
+    PParams,
+    PParamsDelta,
+  )
+where
+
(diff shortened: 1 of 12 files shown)


eras/babbage/impl/src/Cardano/Ledger/Babbage.hs
```diff
@@ -0,0 +1,281 @@
...
+-- Era Mapping
+-------------------------------------------------------------------------------
+
+-- Rules inherited from Alonzo
+
+type instance Core.EraRule "UTXOS" (BabbageEra c) = Alonzo.UTXOS (BabbageEra c)
+
+type instance Core.EraRule "UTXO" (BabbageEra c) = Alonzo.AlonzoUTXO (BabbageEra c)
+
+type instance Core.EraRule "UTXOW" (BabbageEra c) = Alonzo.AlonzoUTXOW (BabbageEra c)
```
JaredCorduan, line 235: let's leave a TODO here to map the three UTXO rules to the new Babbage ones once they exist.

eras/babbage/impl/src/Cardano/Ledger/Babbage.hs
```diff
@@ -0,0 +1,281 @@
...
+      (PlutusScript PlutusV2 _) -> "\x02"
+  validateScript (TimelockScript script) tx = validateTimelock @(BabbageEra c) script tx
+  validateScript (PlutusScript _ _) _tx = True
+
+instance
+  ( CC.Crypto c
+  ) =>
+  API.CanStartFromGenesis (BabbageEra c)
+  where
+  type AdditionalGenesisConfig (BabbageEra c) = AlonzoGenesis
```
JaredCorduan, line 137: I need to remind myself of the details about how this works to convince myself that `AlonzoGenesis` makes sense here...

JaredCorduan (commented):
This looks really great @goolord , thanks! I just need to figure out how `AdditionalGenesisConfig` is being used.

eras/babbage/impl/src/Cardano/Ledger/Babbage.hs
goolord: it seems like `AlonzoGenesis` *only* exists so we can encode/decode the arguments to `extendPParams`, nothing alonzo-specific gets encoded so 🤷

JaredCorduan (approved):
LGTM!

PR #2620, opened 2022-01-19
Unspecialize test hash sizes

lehins:
Avoid mock crypto when possible in favor of StandardCrypto.

Stop using small blake hashes for testing.


eras/shelley/impl/src/Cardano/Ledger/Shelley/RewardUpdate.hs
```diff
@@ -72,9 +72,8 @@ import NoThunks.Class (NoThunks (..), allNoThunks)
 -- ===============================================================
 
 -- | The result of reward calculation is a pair of aggregate Maps.
-data RewardAns c
-  = RewardAns
-      !(Map (Credential 'Staking c) (Reward c))
+newtype RewardAns c
+  = RewardAns (Map (Credential 'Staking c) (Reward c))
```
JaredCorduan, line 76: why did you remove the strictness annotation?

JaredCorduan (approved):
If you can make CI happy I think it is good to go. You did a lot of clean-up, much appreciated!

eras/shelley/impl/src/Cardano/Ledger/Shelley/RewardUpdate.hs
lehins: Because I changed it to a `newtype` :)

lehins:
@JaredCorduan Awesome. Thank you for the review.

I am still battling with one test case, so it is almost done.

eras/shelley/impl/src/Cardano/Ledger/Shelley/RewardUpdate.hs
JaredCorduan: ah! :)

PR #2621, opened 2022-01-19
remove subSlotCount function

### LINKED ISSUE #1698: Remove`subSlotCount`
I grepped for occurrences of `subSlotCount` from the module `Cardano.Chain.Slotting.SlotNumber`. There are no occurrences in `ouroboros-network`. All of the occurrences in `cardano-ledger-specs` are at the head of an argument to an inequality operator and so could be replaced by an `addSlotCount` on the opposite side of the operator.

There are two reasons to avoid this function:

  (1) It's easy to mixup the order of the arguments: is it `a - b` or `b - a` (FYI, it's the latter).

  (2) It returns `0` instead of a negative/underflowed number. For example, in the current code, this means it's possible for a PPU to become a candidate proposal (which must be confirmed, **stably** confirmed, endorsed, and **stably** endorsed) all in slot 0! This is an edge case, sure, but not a necessary one.

cc: @edsko

JaredCorduan:
The `subSlotCount` function is dangerous and can be replaced by `addSlotCount` everywhere in the consensus and ledger repositories.

closes #1698

### DIFF
--- a/eras/byron/ledger/impl/src/Cardano/Chain/Slotting/SlotNumber.hs
+++ b/eras/byron/ledger/impl/src/Cardano/Chain/Slotting/SlotNumber.hs
@@ -10,6 +10,7 @@
 module Cardano.Chain.Slotting.SlotNumber
   ( SlotNumber (..),
     addSlotCount,
+    -- deprecated
     subSlotCount,
   )
 where
@@ -57,9 +58,12 @@ instance B.Buildable SlotNumber where
 
 -- | Increase a 'SlotNumber' by 'SlotCount'
 addSlotCount :: SlotCount -> SlotNumber -> SlotNumber
-addSlotCount (SlotCount a) (SlotNumber b) = SlotNumber $ a + b
+addSlotCount (SlotCount a) (SlotNumber b)
+  | a <= maxBound - b = SlotNumber $ a + b
+  | otherwise = SlotNumber maxBound
 
 -- | Decrease a 'SlotNumber' by 'SlotCount', going no lower than 0
+{-# DEPRECATED subSlotCount "this function is dangerous and can usually be replaced by addSlotCount" #-}
 subSlotCount :: SlotCount -> SlotNumber -> SlotNumber
 subSlotCount (SlotCount a) (SlotNumber b) =
   if a > b then SlotNumber 0 else SlotNumber (b - a)
--- a/eras/byron/ledger/impl/src/Cardano/Chain/Update/Validation/Endorsement.hs
+++ b/eras/byron/ledger/impl/src/Cardano/Chain/Update/Validation/Endorsement.hs
@@ -26,7 +26,7 @@ import Cardano.Binary
 import Cardano.Chain.Common (BlockCount, KeyHash)
 import qualified Cardano.Chain.Delegation as Delegation
 import Cardano.Chain.ProtocolConstants (kSlotSecurityParam)
-import Cardano.Chain.Slotting (SlotNumber, subSlotCount)
+import Cardano.Chain.Slotting (SlotNumber, addSlotCount)
 import Cardano.Chain.Update.Proposal (UpId)
 import Cardano.Chain.Update.ProtocolParameters (ProtocolParameters)
 import Cardano.Chain.Update.ProtocolVersion (ProtocolVersion)
@@ -172,8 +172,7 @@ register env st endorsement =
     isConfirmedAndStable upId = upId `M.member` scps
       where
         -- Stable and confirmed proposals.
-        scps = M.filter (<= stableAt) confirmedProposals
-        stableAt = subSlotCount (kSlotSecurityParam k) currentSlot
+        scps = M.filter (\x -> addSlotCount (kSlotSecurityParam k) x <= currentSlot) confirmedProposals
 
     numberOfEndorsements :: Int
     numberOfEndorsements =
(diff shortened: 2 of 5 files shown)


eras/byron/ledger/impl/src/Cardano/Chain/Slotting/SlotNumber.hs
```diff
@@ -60,6 +61,7 @@ addSlotCount :: SlotCount -> SlotNumber -> SlotNumber
 addSlotCount (SlotCount a) (SlotNumber b) = SlotNumber $ a + b
 
 -- | Decrease a 'SlotNumber' by 'SlotCount', going no lower than 0
+-- {-# DEPRECATED subSlotCount "this function is dangerous and can usually be replaced by addSlotCount" #-}
```
lehins: Is it commented out intentionally? Also replacement function can be linked.
```suggestion
{-# DEPRECATED subSlotCount "this function is dangerous and can usually be replaced by `addSlotCount`" #-}
```

eras/byron/ledger/impl/src/Cardano/Chain/Update/Validation/Endorsement.hs
```diff
@@ -172,8 +172,7 @@ register env st endorsement =
     isConfirmedAndStable upId = upId `M.member` scps
       where
         -- Stable and confirmed proposals.
-        scps = M.filter (<= stableAt) confirmedProposals
-        stableAt = subSlotCount (kSlotSecurityParam k) currentSlot
+        scps = M.filter (\x -> addSlotCount (kSlotSecurityParam k) x <= currentSlot) confirmedProposals
```
lehins, line 175: I don't know what the `confirmedProposals` is, but if they contain untrusted input then there is a danger of overflow, which before was guarded against with `subSlotCount`. Example:
```haskell
λ> let currentSlot = SlotNumber 10
λ> let k = SlotCount 5
λ> filter (\x -> x <= subSlotCount k currentSlot) (map SlotNumber [1,2,30, maxBound])
[SlotNumber {unSlotNumber = 1},SlotNumber {unSlotNumber = 2}]
λ> filter (\x -> addSlotCount k x <= currentSlot) (map SlotNumber [1,2,30, maxBound])
[SlotNumber {unSlotNumber = 1},SlotNumber {unSlotNumber = 2},SlotNumber {unSlotNumber = 18446744073709551615}]
```

So, I'd suggest we also fix `addSlotCount` just in case:

```haskell
addSlotCount :: SlotCount -> SlotNumber -> SlotNumber
addSlotCount (SlotCount a) (SlotNumber b)
  | a <= maxBound - b = SlotNumber $ a + b
  | otherwise = SlotNumber maxBound 
```

lehins (changes requested):
Overall looks great just couple of minor caveats

eras/byron/ledger/impl/src/Cardano/Chain/Slotting/SlotNumber.hs
JaredCorduan: oops, thanks for catching that!

eras/byron/ledger/impl/src/Cardano/Chain/Update/Validation/Endorsement.hs
JaredCorduan: it is not untrusted input, `currentSlot` is the current slot number, and `k` is fixed. plus this is Byron code that will never be run on new inputs on mainnet/testnet. But your code is indeed nicer in general. I'm happy to make the change.

lehins: > it is not untrusted input, currentSlot is the current slot number, and k is fixed.

And what about `confirmedProposals`?

> this is Byron code that will never be run on new inputs 

If this code doesn't run during syncing then it really doesn't matter.

JaredCorduan: yea, `confirmedProposals` is both bounded and only ever supplied by the genesis nodes.

> If this code doesn't run during syncing then it really doesn't matter.

it does runs during sync.so if there was an overflow lurking, a new node would not be able to sync. but there's no worry of a node that has synced of crashing from this.

lehins (approved):

JaredCorduan:
ha ha, fun failure :)

eras/byron/ledger/impl/test/Test/Cardano/Chain/Slotting/Properties.hs
```diff
@@ -66,16 +65,10 @@ ts_prop_addSlotCount = withTestsTS 100 . property $ do
...
-ts_prop_subSlotCount = withTestsTS 100 . property $ do
-  sc <- forAll genSlotCount
-  fs <- forAll genSlotNumber
-  let sc' = SlotNumber $ unSlotCount sc
-      subtracted = fs - sc'
-  subSlotCount sc fs === if fs > sc' then subtracted else SlotNumber 0
+  addSlotCount sc fs
+    === if unSlotNumber fs <= maxBound - (unSlotCount sc)
+      then added
+      else SlotNumber maxBound
```
JaredCorduan, line 71: this test feels pretty silly, it's the same logic as the function it is testing... and in general I feel like `addSlotCount` is only a bit less of a footgun than `subSlotCount` (the ordering on `subSlotCount` was pretty wacky).

PR #2622, opened 2022-01-19
computeStabilityWindow should not use to Double

### LINKED ISSUE #2073: `computeStabilityWindow` uses `Double` and so "rounds" counter-intuitively
I was surprised by this behavior in `computeStabilityWindow`. I looked for a preexisting Issue, but I didn't find it. Perhaps this Issue will just be archived as a public explanation/warning of intended behavior.

https://github.com/input-output-hk/cardano-ledger-specs/blob/823d06c4287f6ae51d0b17fe8e3b3697e1226e24/shelley/chain-and-ledger/executable-spec/src/Shelley/Spec/Ledger/StabilityWindow.hs#L21-L24

An example -- note that `60 /= 61`:

```
Prelude GHC.Real Data.Word> :set -XTypeApplications
Prelude GHC.Real Data.Word> let k = 7 :: Word64
Prelude GHC.Real Data.Word> let f = 7 / 20 :: Ratio Word64   -- this is ultimately the ActiveSlotCoeff type
Prelude GHC.Real Data.Word> ceiling $ fromIntegral @_ @Double (3 * k) / fromRational (toRational f)
61
Prelude GHC.Real Data.Word> ceiling $ fromIntegral @_ @Rational (3 * k) / fromRational (toRational f)
60
Prelude GHC.Real Data.Word> ceiling $ fromIntegral (3 * k) / toRational f
60
```

* Is this kind of "surprise" harmless as long as we're consistent?

* Even so, is it too confusing? I understand that `Double` will be a tighter bound on the computational cost, but are we computing the stability window often enough for that to matter?

* Perhaps the "Non-integer calculations specification" (which I haven't read carefully) forbids the use of `Rational`. I suppose it might be awkward/tedious to compute with as-needed precision in lower-level languages.

* I wonder if there's a clever manual fusion that is most efficient/bounded and exact.

My personal takeaway, regardless of how this issue is resolved: just import `computeStabilityWindow` in downstream libraries!

JaredCorduan:
There is no good reason to convert to a double before taking the ceiling in this function. We now use Rational instead. Even though this is actually a different function now, we can get away with this since both mainnet and testnet have only ever had a single value which happens to agree in both versions. In particular, mainnet and testnet have always
had `k = 2160` and `f = 1/20`.

Similarly for `computeRandomnessStabilisationWindow`.

closes #2073

### DIFF
--- a/eras/shelley/impl/src/Cardano/Ledger/Shelley/StabilityWindow.hs
+++ b/eras/shelley/impl/src/Cardano/Ledger/Shelley/StabilityWindow.hs
@@ -1,5 +1,3 @@
-{-# LANGUAGE TypeApplications #-}
-
 module Cardano.Ledger.Shelley.StabilityWindow
   ( computeStabilityWindow,
     computeRandomnessStabilisationWindow,
@@ -19,7 +17,7 @@ computeStabilityWindow ::
   ActiveSlotCoeff ->
   Word64
 computeStabilityWindow k asc =
-  ceiling $ fromIntegral @_ @Double (3 * k) / fromRational f
+  ceiling $ (3 * fromIntegral k) / f
   where
     f = unboundRational . activeSlotVal $ asc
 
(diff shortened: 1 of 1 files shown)


lehins (approved):
:+1:

PR #2623, opened 2022-01-20
Fix error in Shelley Spec regarding new pool count

### LINKED ISSUE #2517: POOLREAP rules has ambiguous rewardAcnts'
The value `rewardAcnts'` is not well defined in the `POOLREAP` rule, since there could be more than one reward account for multiple retiring pools. The implementation is correctly summing them.

### LINKED ISSUE #2587: Shelley spec - newPools in fig 17
In the Shelley spec, Figure 17, `newPools` is counting certificates, but should be counting unique pool ids. The implementation is doing the correct thing: 

https://github.com/input-output-hk/cardano-ledger/blob/1a9ec4ae9e0b09d54e49b2a40c4ead37edadcce5/eras/shelley/impl/src/Cardano/Ledger/Shelley/UTxO.hs#L284-L285

JaredCorduan:
Two small changes the Shelley spec (where the code was doing the correct thing).

1. In the Shelley spec, Figure 17, `newPools` is counting certificates, but should be counting unique pool ids. 
2. In the `POOLREAP` rule, we were not accounting for the fact that multiple retiring pools could have the same stake address. the old version used:
old-rewardAcnts
the new version uses:
new-rewardAcnts

closes #2517
closes #2587

### DIFF
--- a/eras/shelley/formal-spec/epoch.tex
+++ b/eras/shelley/formal-spec/epoch.tex
@@ -571,13 +571,10 @@ This transition has no preconditions and results in the following state change:
                  & \{\var{hk}\mapsto \fun{poolRAcnt}~\var{pool} \mid
                    \var{hk}\mapsto\var{pool} \in \var{retired}\restrictdom\var{poolParams} \} \\
         \var{rewardAcnts'} & \left\{
-                        a \mapsto c
-                        \mathrel{\Bigg|}
-                        \begin{array}{r@{~\in~}l}
-                          \var{hk} \mapsto c & \var{pr}, \\
-                          \var{hk}\mapsto\var{a} & \var{rewardAcnts} \\
-                        \end{array}
-                      \right\} \\
+                               \sum\limits_{\wcard\mapsto c\in\var{pr}\circ\var{rewardAcnts}^{-1}(a)} c
+                               \mathrel{\Bigg|}
+			       a\in\range{rewardAcnts}
+                             \right\} \\
         \var{refunds} & \dom{rewards}\restrictdom\var{rewardAcnts'} \\
         \var{mRefunds} & \dom{rewards}\subtractdom\var{rewardAcnts'} \\
         \var{refunded} & \sum\limits_{\wcard\mapsto c\in\var{refunds}} c \\
--- a/eras/shelley/formal-spec/epoch.tex
+++ b/eras/shelley/formal-spec/epoch.tex
@@ -571,9 +571,10 @@ This transition has no preconditions and results in the following state change:
                  & \{\var{hk}\mapsto \fun{poolRAcnt}~\var{pool} \mid
                    \var{hk}\mapsto\var{pool} \in \var{retired}\restrictdom\var{poolParams} \} \\
         \var{rewardAcnts'} & \left\{
-                               \sum\limits_{\wcard\mapsto c\in\var{pr}\circ\var{rewardAcnts}^{-1}(a)} c
-                               \mathrel{\Bigg|}
-			       a\in\range{rewardAcnts}
+                               a \mapsto
+                               \sum\var{pr}(\var{rewardAcnts}^{-1}(a))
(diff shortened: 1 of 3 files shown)


eras/shelley/formal-spec/utxo.tex
```diff
@@ -401,9 +401,9 @@ \subsection{Deposits and Refunds}
     & \text{total deposits for transaction} \\
     & \totalDeposits{pp}{poolParams}{certs} = \\
     &  ~~~ (\fun{keyDeposit}~pp)\cdot|\var{certs}\cap\DCertRegKey| \\
-    &  ~~~~~ + (\fun{poolDeposit}~pp)\cdot|\var{newPools}| \\
+    &  ~~~~~ + (\fun{poolDeposit}~pp)\cdot|\{\cwitness{c} ~\mid c\in~\var{newPools}\}| \\
     & ~~~\where \\
-    &  ~~~~~~~ \var{newPools} = \{\var{certs}\cap\DCertRegPool ~\mid~ \cwitness{c}\notin \var{poolParams}\}
+    &  ~~~~~~~ \var{newPools} = \{c ~\mid~ \var{certs}\cap\DCertRegPool,\cwitness{c}\notin \var{poolParams}\}
```
WhatisRT: I don't understand this line. Is `c` supposed to be an element of `certs ∩ DCertRegPool`? The old version doesn't make sense to me either btw.

eras/shelley/formal-spec/epoch.tex
```diff
@@ -571,13 +571,10 @@ \subsection{Pool Reaping Transition}
...
-                        \mathrel{\Bigg|}
-                        \begin{array}{r@{~\in~}l}
-                          \var{hk} \mapsto c & \var{pr}, \\
-                          \var{hk}\mapsto\var{a} & \var{rewardAcnts} \\
-                        \end{array}
-                      \right\} \\
+                               \sum\limits_{\wcard\mapsto c\in\var{pr}\circ\var{rewardAcnts}^{-1}(a)} c
+                               \mathrel{\Bigg|}
+			       a\in\range{rewardAcnts}
+                             \right\} \\
```
WhatisRT, line 578: I think what you want is `a \mapsto \sum\limits_{c\in\var{pr}(\var{rewardAcnts}^{-1}(a))} c`:
- currently, `rewardAcnts'` is just a set of numbers
- `\var{rewardAcnts}^{-1}(a)` is a set of hashes, so you can only apply it to `pr`, not compose it with, which will give you a set of numbers

Also, I think the `|` looks a bit better if it's just `Big`. Finally, I think `\sum\var{pr}(\var{rewardAcnts}^{-1}(a))` is a nice point-free version of the sum, but pick whichever one you like better.

WhatisRT (changes requested):

eras/shelley/formal-spec/utxo.tex
JaredCorduan: ah yes, it is supposed to be `c \in \var{certs}\cap\DCertRegPool`.

eras/shelley/formal-spec/epoch.tex
JaredCorduan: you're totally right! I had meant for `rewardAcnts` to be a mapping. thanks!

eras/shelley/formal-spec/utxo.tex
JaredCorduan: and the extra `c \in` totally breaks the table spacing/corners now :laughing: :sob: hopefully `tx` is a suitable abbreviation for `transaction`, that lets me reclaim the white space back

eras/shelley/formal-spec/epoch.tex
JaredCorduan: your point free version is very clean, I love it, I've changed to using it

WhatisRT (approved):
LGTM!

PR #2625, opened 2022-01-20
Improve performance of Value's size

lehins:


redxaxder (approved):

redxaxder (approved):

eras/shelley-ma/impl/src/Cardano/Ledger/Mary/Value.hs
```diff
@@ -668,6 +671,17 @@ from (CompactValueMultiAsset c numAssets rep) =
...
+nubOrd :: Ord a => [a] -> [a]
+nubOrd =
+  loop mempty
+  where
+    loop _ [] = []
+    loop s (a : as)
+      | a `Set.member` s = loop s as
+      | otherwise =
+        let s' = Set.insert a s in s' `seq` a : loop s' as
+
```
redxaxder, line 684: the ordered assumption could save on the set construction, since it suffices to compare an element to the previous one

lehins: Sorry, I am not quite sure what you mean. Are trying to say that `rawTriplets` are already sorted on the snd element? If that's the case then we don't need this  implementation of `nubOrd`.

redxaxder: There's a comment at the use site:
```
-- This assumes that the triples are ordered by nondecreasing asset name offset
```
It was based on my previous incorrect understanding of nub.

PR #2627, opened 2022-01-21
JSON instance for AlonzoGenesis

### LINKED ISSUE #2452: Move AlonzoGenesis From/ToJSON instances from cardano-node to cardano-ledger-specs
For a `ShelleyGenesis` ledger-specs provides `FromJSON` and `ToJSON` instances https://github.com/input-output-hk/cardano-ledger-specs/blob/master/shelley/chain-and-ledger/executable-spec/src/Shelley/Spec/Ledger/Genesis.hs#L149

But for `AlonzoGenesis` it does not provide those instances. 

Because of that cardano-node had to provided it on its own https://github.com/input-output-hk/cardano-node/blob/master/cardano-api/src/Cardano/Api/Orphans.hs#L354 and now db-analyser in consensus had to duplicate that code https://github.com/input-output-hk/ouroboros-network/pull/3337

For consistency I believe that those instances should be defined in ledger-specs (as they are defined for `ShelleyGenesis`

JaredCorduan:
taken from cardano-node

closes #2452

### DIFF
--- a/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Genesis.hs
+++ b/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Genesis.hs
@@ -29,7 +29,6 @@ import qualified Cardano.Ledger.Core as Core
 import Cardano.Ledger.Era (Era)
 import Cardano.Ledger.SafeHash (extractHash)
 import qualified Cardano.Ledger.Shelley.PParams as Shelley
-import Control.Applicative ((<|>))
 import Data.Aeson (FromJSON (..), ToJSON (..), Value, object, (.!=), (.:), (.:?), (.=))
 import qualified Data.Aeson as Aeson
 import Data.Aeson.Types (FromJSONKey (..), ToJSONKey (..), toJSONKeyText)
@@ -206,9 +205,7 @@ instance FromJSONKey Language where
 
 instance FromJSON AlonzoGenesis where
   parseJSON = Aeson.withObject "Alonzo Genesis" $ \o -> do
-    coinsPerUTxOWord <-
-      o .: "lovelacePerUTxOWord"
-        <|> o .: "adaPerUTxOWord" --TODO: deprecate
+    coinsPerUTxOWord <- o .: "lovelacePerUTxOWord"
     cModels <- o .:? "costModels"
     prices <- o .: "executionPrices"
     maxTxExUnits <- o .: "maxTxExUnits"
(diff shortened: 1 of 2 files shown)


eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Genesis.hs
```diff
@@ -100,3 +125,211 @@ instance ToCBOR AlonzoGenesis where
...
+            { coinsPerUTxOWord,
+              costmdls,
+              prices,
+              maxTxExUnits,
+              maxBlockExUnits,
+              maxValSize,
+              collateralPercentage,
+              maxCollateralInputs
+            }
+
```
JaredCorduan, line 243: @Jimbo4350 - in node there is a comment here stating that the cost model is rendered separately. but below I seed the cost models being rendered. can I delete this comment?

https://github.com/input-output-hk/cardano-node/blob/42582c828ac4c364839427bc5bd604a03c9ac6b3/cardano-api/src/Cardano/Api/Orphans.hs#L416-L418

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Genesis.hs
```diff
@@ -100,3 +125,211 @@ instance ToCBOR AlonzoGenesis where
...
+  toJSONKey = toJSONKeyText languageToText
+
+instance FromJSONKey Language where
+  fromJSONKey = Aeson.FromJSONKeyTextParser languageFromText
+
+instance FromJSON AlonzoGenesis where
+  parseJSON = Aeson.withObject "Alonzo Genesis" $ \o -> do
+    coinsPerUTxOWord <-
+      o .: "lovelacePerUTxOWord"
+        <|> o .: "adaPerUTxOWord" --TODO: deprecate
```
JaredCorduan: @Jimbo4350 - should I remove `adaPerUTxOWord`?

Jimbo4350: Yeah it looks like we can remove this, there is no trace of it in the node repo.

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Genesis.hs
Jimbo4350: Yes this comment is outdated and now incorrect, you can delete it :+1: .

Jimbo4350 (approved):
LGTM!

JaredCorduan:
oh embarrassing, I forgot to smash the commits :facepalm:

PR #2628, opened 2022-01-24
Add EraIndependentBlockHeader

nc6:
Now we have new protocols, the header may change, so introduce an era
(protocol) independent marker for this.


JaredCorduan (approved):
makes sense!

PR #2630, opened 2022-01-25
Rationalise 'chainChecks'.

nc6:
'chainChecks' is a bit of a weird function; it implements "envelope"
checks, but sits somewhat within the ledger code. The current
implementation requires full access to 'Globals', which contains a bunch
of ledger specific contents. It's also era-parametrised, despite having
nothing to do with the ledger era.

This PR therefore does two things:
- Change the API function to only require the max major PV, not the full
  globals.
- Remove the redundnant era parameter from 'ChainPredicateFailure'.


lehins (approved):
LGTM

JaredCorduan (approved):
very nice!

PR #2631, opened 2022-01-25
Ts reorganise pretty

TimSheard:
Reorganized Cardano.Ledger.Pretty

It is broken into two parts PrettyBase and Cardano.Ledger.Pretty

1) PrettyBase does not depend on any Era. it defines pretty printers for data structures widely used
    such as Map Set StrictMaybe StrictSequence, etc. 
2) Cardano.Ledger.Pretty  introduces the PrettyA class and provides scores of instances for many data types defined in Cardano.Core and Cardano.Ledger.Shelley


JaredCorduan (approved):
LGTM

JaredCorduan:
@TimSheard should we close this PR for now?

JaredCorduan:
we still want to re-organize the pretty printing code, but this PR is so old there's not enough here to salvage.

PR #2284, opened 2021-05-13
optimize the performance of isBootstrapRedeemer 

goolord:
resolves https://github.com/input-output-hk/cardano-ledger-specs/issues/1937

PR #2632, opened 2022-01-25
Proper fix for optimized version of `isBootstrapRedeemer` function.

lehins:
I ran into a test failure on CI, which can be reproduced with this command prior to this PR:
```shell
cabal test cardano-ledger-shelley-test --test-options='--quickcheck-replay=97935 -p "isBootstrapRedeemer is equivalent for CompactAddr and Addr"'
```

Original attempt to fix this was in #2284 which was invalid and resulted in the optimized version of the function not been used, however the test remained.

This is indeed a good optimization, since `returnRedeemAddrsToReserves` iterates over the whole UTxO and current unoptimized version of `isBootstrapRedeemer` will cause redundant address uncompacting of all Shelley addresses

### DIFF
--- a/eras/shelley/test-suite/src/Test/Cardano/Ledger/Shelley/Address/CompactAddr.hs
+++ b/eras/shelley/test-suite/src/Test/Cardano/Ledger/Shelley/Address/CompactAddr.hs
@@ -60,8 +60,7 @@ propDecompactShelleyLazyAddr = do
       keyHash1 = unsafeGetHash . CA.decompactAddr . mangle . CA.compactAddr $ addr
    in pure $ keyHash0 == keyHash1
 
--- | TODO This property test is failing to find a discrepancy that was found on mainnet.
-propIsBootstrapRedeemer :: Addr crypto -> Property
+propIsBootstrapRedeemer :: CC.Crypto crypto => Addr crypto -> Property
 propIsBootstrapRedeemer addr =
   Addr.isBootstrapRedeemer addr === CA.isBootstrapRedeemer (CA.compactAddr addr)
 
--- a/libs/cardano-ledger-core/src/Cardano/Ledger/CompactAddress.hs
+++ b/libs/cardano-ledger-core/src/Cardano/Ledger/CompactAddress.hs
@@ -32,6 +32,7 @@ import Cardano.Ledger.Address
     toWord7,
     word7sToWord64,
   )
+import qualified Cardano.Ledger.Address as Address (isBootstrapRedeemer)
 import Cardano.Ledger.BaseTypes (word8ToNetwork)
 import Cardano.Ledger.Credential
   ( Credential (KeyHashObj, ScriptHashObj),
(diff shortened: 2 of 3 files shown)


eras/shelley/test-suite/src/Test/Cardano/Ledger/Shelley/Address/CompactAddr.hs
```diff
@@ -60,8 +60,7 @@ propDecompactShelleyLazyAddr = do
       keyHash1 = unsafeGetHash . CA.decompactAddr . mangle . CA.compactAddr $ addr
    in pure $ keyHash0 == keyHash1
 
--- | TODO This property test is failing to find a discrepancy that was found on mainnet.
-propIsBootstrapRedeemer :: Addr crypto -> Property
+propIsBootstrapRedeemer :: CC.Crypto crypto => Addr crypto -> Property
```
goolord, line 63: does this constraint fix the test not covering a case that was on mainnet? is this some overlapping instance or something

libs/cardano-ledger-core/src/Cardano/Ledger/CompactAddress.hs
```diff
@@ -229,12 +230,10 @@ getPayCred header = case testBit header payCredIsScript of
...
--- with the one in Cardano.Ledger.Address
-isBootstrapRedeemer :: CompactAddr crypto -> Bool
-isBootstrapRedeemer (UnsafeCompactAddr bytes) =
-  testBit header byron -- AddrBootstrap
-    && addrType == 2 -- ATRedeem
+-- | Optimized version of `Address.isBootstrapRedeemer`, where we avoid
+-- decompacting of post Byron addresses.
+isBootstrapRedeemer :: forall crypto. CC.Crypto crypto => CompactAddr crypto -> Bool
+isBootstrapRedeemer cAddr@(UnsafeCompactAddr bytes) =
+  testBit header byron && Address.isBootstrapRedeemer (decompactAddr cAddr :: Addr crypto)
```
goolord, line 237: i guess this still avoids decompacting the addr more often then not

goolord (approved):

libs/cardano-ledger-core/src/Cardano/Ledger/CompactAddress.hs
lehins: Yep, that's the point. There aren't too many Byron addresses, when compared to the whole UTxO

eras/shelley/test-suite/src/Test/Cardano/Ledger/Shelley/Address/CompactAddr.hs
lehins: The `Crypto` constraint is necessary due to the fact that the `isBootstrapRedeemer` that accepts compact addresses now sometimes needs to decompact an address, which requires that constraint. This test checks two separate functions.

From what I've seen so far in the codebase we don't have any overlapping instances, otherwise I'd immediately advocate for their removal, it is a bad practice most of the time.

libs/cardano-ledger-core/src/Cardano/Ledger/CompactAddress.hs
JaredCorduan: this function is actually only ever called on the epoch boundary when Shelley transitions to Allegra. At that time, there actually were/are probably still a lot of Byron addresses.

JaredCorduan (approved):
thank you @lehins , I'm glad to see that TODO be resolved!

PR #2635, opened 2022-01-26
provide a reference for the cost model keys

### LINKED ISSUE #2634: important to define the key names for costModel
The cost model is used for computing script integrity hash, which is done by sorting the cost model values lexicographically by key names. The key names are unknown until we see find the key names as used by the cli when querying proto params. Since script integrity hash is also computed in other languages, it is required to define the key names in the CDDL spec.

https://github.com/input-output-hk/cardano-ledger/blob/01e362881823e30cf69907c65b038dc34f9bdd20/eras/alonzo/test-suite/cddl-files/alonzo.cddl#L312

JaredCorduan:
Provide a reference for the cost model keys so that the CDDL spec can reference it.

closes #2634


eras/alonzo/test-suite/cddl-files/alonzo.cddl
```diff
@@ -311,6 +311,11 @@ language = 0 ; Plutus v1
 
 costmdls = { * language => cost_model } ; New
 
+; The keys to the cost model map are not present in the serialization.
+; The values in the serialization are assumed to be ordered
+; lexicographically by their correpsonding key value.
+; The key values can be found in the Plutus repository:
+;   https://github.com/input-output-hk/plutus/blob/748a0c62de00f9944100b616367d9b6549346022/plutus-core/plutus-core/test/CostModelInterface/defaultCostModelParams.json
```
michaelpj: I'm less comfortable with referencing this externally. That's data that we use for testing, not anything that we're committing to be used as reference. If you're okay with that then fine.

Also there is a reference here: https://plutus.readthedocs.io/en/latest/reference/cost-model-parameters.html

JaredCorduan: I'm glad you said something, I must have misunderstood. Would you prefer if I add the text file back, the one which just lists the key names?

michaelpj: I think it's whichever you're more comfortable with the risks for:
- The text file we'll have to remember to keep up to date
- The reference to the test file might become wrong, or we might move it, or who knows

nc6 (approved):

PR #2636, opened 2022-01-27
Improve access to byron addresses

### LINKED ISSUE #1788: Avoid repeated viewCompactTxOut in the Utxo rule
I stumbled upon the following code: https://github.com/input-output-hk/cardano-ledger-specs/blob/d0f7eded325c4e4b36e44b4b49fd75b0c83f00f3/shelley/chain-and-ledger/executable-spec/src/Shelley/Spec/Ledger/STS/Utxo.hs#L310-L319

Note that the `TxOut`s of the transaction are traversed twice, which is fine, however, each traversal pattern matches on them, which requires the use of `viewCompactTxOut`, which is not _that_ cheap (it's being optimised, though).

*Disclaimer: this is not based on profiling results, it's a conjecture (some would call it the root of all evil :wink:). Since the number of `TxOut`s in a transaction is small, what I'm proposing may not have a measurable impact.* 

A potential improvement would be to first partition the `TxOut`s in Shelley `(Network, PaymentCredential crypto, StakeReference crypto, Coin)` (we actually only care about the `Coin`, so the first three elements can be dropped) and Byron `(BootStrapAddress crypto, coin)` (we don't care about the `Coin`) `TxOut`s so that `viewCompactTxOut` is only used once for each transaction. Each of the two traversals can now just the traverse the pre-decompacted `TxOut`s of the right constructor.

Alternatively, the `TxOut`s could be partitioned in Shelley and Byron `TxOut`s (but not pattern-matched) using the `testBit header byron` trick and then the two passes could do the pattern-matching like now (but the pattern matches would now never fail). This is a less invasive change.

Also: does the whole `addr` need to be deserialised to measure the size of the attrs (`bootstrapAddressAttrsSize addr > 64`)?

lehins:
Few other improvements that expand on the same idea from #2632 

Also fixes #1788 

There some places I was able to find that either:

* deserialize Byron addresses, but then ignore the result in the end.
* care only about Byron addresses, but result in deserialization of Shelley addresses as well, which get discarded.


JaredCorduan (approved):
reviewed in a call, looks good to me

PR #2639, opened 2022-01-31
Clarify which block's header hash the spec is referring to in the Tick Nonce Transition

Jimbo4350:


JaredCorduan (approved):
Thanks for cleaning this up @Jimbo4350 ! If you don't mind, I would like to add a commit to this PR that also adds a new item to the errata section about the fact that the consensus layer is actually using the penultimate block hash from the previous epoch. (I can make this second commit)

eras/shelley/formal-spec/frontmatter.tex
```diff
@@ -84,6 +85,7 @@ \section*{List of Contributors}
 Christian Lindgren,
 Yun Lu,
 Philipp Kant,
+Jordan Millar,
```
Jimbo4350, line 88: :partying_face:

Jimbo4350 (commented):
LGTM!

PR #2640, opened 2022-02-01
WIP  generalize Rules for Utxow, Utxo, Utxos, then use to make rules for Babbage

TimSheard:
We generalize rules for the Uxto* rules, We added
genericAlonzoUtxo, genericAlonzoUtxow, and genericShelleyUtxow, used them to make instances Utxow and Utxo instances for Shelley, Shelley-ma, and Alonzo. 

The goal is to use them to make Babbage instances.

### DIFF
--- a/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Rules/Utxo.hs
+++ b/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Rules/Utxo.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE ConstraintKinds #-}
 {-# LANGUAGE DataKinds #-}
 {-# LANGUAGE DeriveGeneric #-}
 {-# LANGUAGE DerivingStrategies #-}
@@ -26,7 +27,7 @@ import Cardano.Ledger.Alonzo.Scripts (ExUnits (..), Prices, pointWiseExUnits)
 import Cardano.Ledger.Alonzo.Tx (ValidatedTx (..), minfee, totExUnits)
 import qualified Cardano.Ledger.Alonzo.Tx as Alonzo (ValidatedTx)
 import qualified Cardano.Ledger.Alonzo.TxSeq as Alonzo (TxSeq)
-import Cardano.Ledger.Alonzo.TxWitness (TxWitness (txrdmrs'), nullRedeemers)
+import Cardano.Ledger.Alonzo.TxWitness (Redeemers, TxWitness (txrdmrs'), nullRedeemers)
 import Cardano.Ledger.BaseTypes
   ( Network,
     ShelleyBase,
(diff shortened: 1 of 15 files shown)


eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Rules/Utxo.hs
```diff
@@ -290,59 +292,63 @@ feesOK pp tx (UTxO utxo) = do
...
       vi@(ValidityInterval _ i_f) = getField @"vldt" txb
-      inputsAndCollateral =
-        Set.union
-          (getField @"inputs" txb)
-          (getField @"collateral" txb)
+      allInputs =
+        Set.unions
+          [ getField @"inputs" txb,
+            getField @"collateral" txb,
+            referenceInputs' delta txb
```
lehins, line 378: I don't think it is right to introduce concepts into past eras from the future eras. There was no reference inputs in Alonzo

TimSheard:
asadad

image

TimSheard:
image

strategy

TimSheard:
Names of transition functions in STS instances for some Rules across all Eras. Note that some transition functions are applicable to all Eras, some to only a subset of Eras. Some transition functions applicable to more than one Era need to be specialized, they appear in color and have an symbolic argument ( _ ).

image

These "generic" transition functions can be found here:
Cardano.Ledger.Shelley.Rules.Utxow(genericShelleyUtxow)
Cardano.Ledger.Shelley.Rules.Utxo(utxoInductive)
Cardano.Ledger.Alonzo.Rules.Utxow(genericAlonzoUtxow)
Cardano.Ledger.Alonzo.Rules.Utxos(genericUtxosTransition)
Cardano.Ledger.Alonzo.Rules.Utxo(genericAlonzoUtxo)

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Rules/Utxo.hs
TimSheard: In order to generalize, we need to anticipate changes that will occur in future eras. Of course we can't always do that, but we can by adding these things in later, and presenting one consistent story across all eras.  Readers are not interested in the path we took to get to the current state, but in a consistent story about the current state. Many of our "differences" are adding new things, that are consistent with the past, as long as they have a zero or null value in past eras, that did not originally include them. Examples are the minting field, and the collateral inputs. We tell the story that we aways had these in mind, but did not bother to store them in Shelley, and Allegra (for mint) and for Shelley, Allegra, and Mary for (collateral) because the were aways, 0 for mint, and emptyset for collateral.

eras/babbage/impl/src/Cardano/Ledger/Babbage/Collateral.hs
```diff
@@ -0,0 +1,92 @@
...
+import Cardano.Ledger.Val ((<->))
+import Data.Compact.SplitMap ((◁))
+import qualified Data.Compact.SplitMap as SplitMap
+import Data.Maybe.Strict (StrictMaybe (..))
+import GHC.Records (HasField (..))
+import Numeric.Natural (Natural)
+
+-- ============================================================
+
+isTwoPhaseScriptAddress ::
```
JaredCorduan, line 38: can we not reuse the version from Alonzo? https://github.com/input-output-hk/cardano-ledger/blob/2048cd4679ef3155d88a445e0f818abad09a0827/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Tx.hs#L268

JaredCorduan: nevermind! it's slightly different in babbage

JaredCorduan:
this we be replaced by a new version soon. thank you @TimSheard for all the work on this PR and the one coming up!!

PR #2643, opened 2022-02-03
Updated the generic mechanisms to build transactions to the Babbage era.

TimSheard:


libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/Updaters.hs
```diff
@@ -36,6 +37,11 @@ import Cardano.Ledger.Alonzo.Tx (hashScriptIntegrity)
 import qualified Cardano.Ledger.Alonzo.Tx as Alonzo
 import qualified Cardano.Ledger.Alonzo.TxBody as Alonzo (TxOut (..))
 import Cardano.Ledger.Alonzo.TxWitness (Redeemers (..), TxDats (..), TxWitness (..), unTxDats)
+--
+
```
JaredCorduan: I'm surprised ormolu allows this...

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/Updaters.hs
```diff
@@ -619,6 +689,27 @@ updatePParams (Alonzo _) pp dpp = case dpp of
   MaxBlockExUnits n -> pp {Alonzo._maxBlockExUnits = n}
   CollateralPercentage perc -> pp {Alonzo._collateralPercentage = perc}
   _ -> pp
+updatePParams (Babbage _) pp dpp = case dpp of
```
JaredCorduan, line 365: should we make a comment that we correctly and intentionally are dropping `d` and `extraEntropy` here?

JaredCorduan (approved):
Looks good to me

eras/babbage/impl/src/Cardano/Ledger/Babbage/TxBody.hs
```diff
@@ -245,6 +245,14 @@ data Datum era
   | Datum !(BinaryData era)
   deriving (Eq, Ord, Show)
 
+instance Semigroup (Datum era) where -- This instance is left biased
```
JaredCorduan: was this instance needed?

TimSheard: Yes it is used in merging in the generic stuff.

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/Updaters.hs
TimSheard: this is supposed to be the union over all Eras, so extraEntropy is still around in past Eras.

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/Updaters.hs
TimSheard: its gone now

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/Indexed.hs
```diff
@@ -307,4 +314,4 @@ instance PrettyA (SKey kr c) where
 instance PrettyA (MultiAsset era) where
   prettyA (MultiAsset v) = prettyA v
 
-type Good c = DSignable c (CH.Hash (CC.HASH c) EraIndependentTxBody)
+type Goode c = DSignable c (CH.Hash (CC.HASH c) EraIndependentTxBody)
```
JaredCorduan: goodie goodie gumdrops?

TimSheard: This was a leftover wart. It has been removed.

eras/babbage/impl/src/Cardano/Ledger/Babbage/TxBody.hs
TimSheard: Not true any more, with the simplified merging. So I removed it.

PR #2644, opened 2022-02-04
Refactor UTxO rules for better re-use in future Eras

lehins:
This approach relies on splitting large rules into smaller validation steps, which can then be reused individually in any era.

In this PR only UTXOW rule has been changed to show how this approach would work.

Any feedback is welcomed. 

(FYI this PR has a bit higher diff than strictly necessary because it moves `witsVKeyNeeded` and `verifiedWits` into `UTXOW`, since those functions do not need to be exported.)

### DIFF
--- a/eras/alonzo/formal-spec/utxo.tex
+++ b/eras/alonzo/formal-spec/utxo.tex
@@ -18,9 +18,9 @@
     & \fun{feesOK} : \PParams \to \Tx \to \UTxO \to \Bool  \\
     & \fun{feesOK}~\var{pp}~tx~utxo~= \\
     &~~      \minfee{pp}~{tx} \leq \txfee{txb} \wedge (\fun{txrdmrs}~tx \neq \emptyset \Rightarrow \\
-    &~~~~~~((\forall (a, \wcard, \_) \in \fun{range}~(\fun{collateral}~{txb} \restrictdom \var{utxo}), a \in \AddrVKey) \\
+    &~~~~~~((\forall (a, \wcard, \wcard) \in \fun{range}~(\fun{collateral}~{txb} \restrictdom \var{utxo}), a \in \AddrVKey) \\
     &~~~~~~\wedge~ \fun{adaOnly}~\var{balance} \\
-    &~~~~~~\wedge~ \var{balance}*100 \geq \txfee{txb} * (\fun{collateralPercent}~pp))) \\
+    &~~~~~~\wedge~ \var{balance}*100 \geq \txfee{txb} * (\fun{collateralPercent}~pp) \\
     &~~~~~~\wedge~ \fun{collateral}~{tx} \neq \emptyset) \\
     &~~      \where \\
     & ~~~~~~~ \var{txb} = \txbody{tx} \\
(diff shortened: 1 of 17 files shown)


nc6:
Taking a quick look, I think this approach is a lot clearer, without the need to introduce concepts from future eras!

TimSheard:
There are things I like about this, and things I don't.  Every rule requires a huge amount of constraints, so when we define a rule usually more than half of what we write is constraints, and some thing like this is required for every rule. A better approach would be to group rules together, so there is less "wasted" effort.

Then we should make rules an instance of Semigroup so we can combine them like this
(rule1 arg1 )<> (rule2 arg2 arg3) <> rule3  <> ...

I wanted to study what a Validation datatype looked like to see if I could do that, but I could not find the file. Did it not get added to the branch?

lehins:
> Taking a quick look, I think this approach is a lot clearer, without the need to introduce concepts from future eras!

@nc6 Good to hear that you like it. One of my main goals was to come up with a solution that would avoid introduction of changes to the older code as future eras come along. Overall this approach is ideal if we want to reduce mutation of older rules:

* we don't have to add new concepts to older rules, like it is done in #2640
* when we do need to fix or improve a predicate validation we can do it in one place where it is implemented, rather than having to go through all eras that use that particular check like it is now on master.


> Every rule requires a huge amount of constraints, so when we define a rule usually more than half of what we write is constraints, and some thing like this is required for every rule. A better approach would be to group rules together, so there is less "wasted" effort.

@TimSheard I think you misunderstood what I meant by reduction of duplication. The main goal is not to necessarily reduce amount of code, but to reduce duplication of the actual logic implemented in the rules. This approach solves precisely that at the cost of extra new functions with type signatures and various constraints. Type signatures are cheap to maintain, since they are forced upon us by the compiler (i.e . you can't really mess 'em up), while the logic is expensive to produce and maintain, because it requires extensive testing and is very sensitive to any modifications. 

> A better approach would be to group rules together, so there is less "wasted" effort.

The whole reason why this approach works is because it does the opposite, it splits rules into small steps that can be reused in subsequent eras as building blocks. This way unchanged sections can be invoked directly, while portions that have changed will get reimplemented. Moreover, in my opinion, placing each individual logical predicate validation into a separate function with a type signature and documentation is definitely not a wasted effort, but a way to clarify what a specific portion of a rule  does. Potentially we could even have a direct mapping into a spec, by naming those sections in the spec the same way they are named in the code.

> Then we should make rules an instance of Semigroup so we can combine them like this

Yes it is possible to combine smaller validations into bigger ones using either `Applicative` or `Semigroup` instances. In some case it might be sensible thing to do, but it is not critical in my opinion.

> I wanted to study what a Validation datatype looked like to see if I could do that, but I could not find the file. Did it not get added to the branch?

It comes from an external package [validation-selective](https://hackage.haskell.org/package/validation-selective-0.1.0.1/docs/Validation.html). I've use a similar package quite successfully on a few occasions, but this package is newer and I like even better then the one I used before.

JaredCorduan (approved):
I like this approach a lot. I hope you don't mind that I changed the title of the PR. :) Also thank you for the various clean-up also included in this PR!

PR #2645, opened 2022-02-05
Babbage: Reference scripts

### LINKED ISSUE #2465: Babbage TxOut

goolord:
resolves https://github.com/input-output-hk/cardano-ledger/issues/2465

### DIFF
--- a/eras/babbage/impl/src/Cardano/Ledger/Babbage/TxBody.hs
+++ b/eras/babbage/impl/src/Cardano/Ledger/Babbage/TxBody.hs
@@ -350,8 +350,8 @@ pattern TxOutCompactDH addr vl dh <-
 type ScriptIntegrityHash crypto = SafeHash crypto EraIndependentScriptIntegrity
 
 data TxBodyRaw era = TxBodyRaw
-  { _inputs :: !(Set (TxIn (Crypto era))),
-    _collateral :: !(Set (TxIn (Crypto era))),
+  { _spendInputs :: !(Set (TxIn (Crypto era))),
+    _collateralInputs :: !(Set (TxIn (Crypto era))),
     _referenceInputs :: !(Set (TxIn (Crypto era))),
     _outputs :: !(StrictSeq (TxOut era)),
     _collateralReturn :: !(StrictMaybe (TxOut era)),
@@ -495,8 +495,8 @@ pattern TxBody
   TxBodyConstr
     ( Memo
         TxBodyRaw
-          { _inputs = inputs,
-            _collateral = collateral,
+          { _spendInputs = inputs,
+            _collateralInputs = collateral,
             _referenceInputs = referenceInputs,
             _outputs = outputs,
             _collateralReturn = collateralReturn,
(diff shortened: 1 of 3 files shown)


eras/babbage/impl/src/Cardano/Ledger/Babbage/TxBody.hs
```diff
@@ -163,6 +164,11 @@ data TxOut era
       {-# UNPACK #-} !(CompactAddr (Crypto era))
       !(CompactForm (Core.Value era))
       {-# UNPACK #-} !(BinaryData era) -- Inline data
+  | TxOutCompactRefScript'
+      {-# UNPACK #-} !(CompactAddr (Crypto era))
+      !(CompactForm (Core.Value era))
+      !(Datum era)
```
goolord, line 170: this is kind of a compromise between performance and having a nice API + not polluting the CBOR with a bunch of extra constructors to represent every possible new permutation of `Addr × Value × (Datum | DataHash)? × Script?`

goolord:
https://github.com/input-output-hk/cardano-ledger/runs/5059981296?check_suite_focus=true#step:20:2261 this failure seems unrelated

eras/babbage/impl/src/Cardano/Ledger/Babbage/TxBody.hs
JaredCorduan: yep, I like this. we are indeed on a combinatorially explosive path that has to be dampened somehow.

JaredCorduan (approved):
Looks good to me! We are still trying to decide on the final serialization schema for the Babbage era transaction outputs, but I think this PR does not need to be blocked by that decision.

eras/babbage/impl/src/Cardano/Ledger/Babbage/TxBody.hs
lehins: It is not compromise between performance. It is compromise between memory overhead. `Datum era` takes up more than twice what the `BinaryData era` does!

I am afraid that in that we don't get UTxO HD in time we will need to save every byte possible. In my opinion, once we don't have to worry about memory overhead that much we can figure out how we can dampen this, but until then we should be careful!

lehins: My previous comment was wrong. This looks good, I like it too! :wink:

JaredCorduan:
I'm not sure what is going on with ormolu, I'll try to restart the github action

eras/babbage/impl/src/Cardano/Ledger/Babbage/TxBody.hs
yihuang: Does this mean that inline data and ref script can't be used in the same output?

lehins: On the contrary, it means that inline datum is optional with a ref script

goolord: `TxOutCompactRefScript' caddr value (Datum inlineDatum) refScript` is valid

PR #2615, opened 2022-01-12
Ts tell rewards

TimSheard:
We have added incremental telling of reward updates. On every pulse we tell the member rewards we computed in that pulse. We still compute the accumulated rewards so that we can apply them to the LedgerState. But the telling should help the DB-Synch folks.  

We maintain and test an invariant that the two ways of doing things compute the same result.
We develop a new testing framework to run property tests on Traces for a single epoch extracted from bigger traces.

PR #2647, opened 2022-02-07
add an epoch number to the TotalRewardEvent event

JaredCorduan:
Prior to #2615, the reward event included an epoch number. In #2615, we renamed the old reward event to `TotalRewardEvent` and created a new incremental reward event named `DeltaRewardEvent`. The event `TotalRewardEvent` does not currently, however, include an epoch number, though `DeltaRewardEvent` does. The PR add an epoch number to `TotalRewardEvent`.

### DIFF
--- a/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Tx.hs
+++ b/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Tx.hs
@@ -238,7 +238,6 @@ deriving instance Typeable era => NoThunks (ScriptIntegrity era)
 -- Instead, we must use a reproducable serialization
 instance Era era => SafeToHash (ScriptIntegrity era) where
   originalBytes (ScriptIntegrity m d l) =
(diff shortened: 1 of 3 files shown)


eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Tx.hs
```diff
@@ -238,7 +238,6 @@ deriving instance Typeable era => NoThunks (ScriptIntegrity era)
 -- Instead, we must use a reproducable serialization
 instance Era era => SafeToHash (ScriptIntegrity era) where
   originalBytes (ScriptIntegrity m d l) =
-    -- TODO: double check that canonical encodings are used for the langDepView (l)
```
JaredCorduan, line 241: We double checked this in #2589

nc6 (approved):

JaredCorduan:
the test that failed is the flaky `gen Always shrink` test. This is a test for the model-based testing generators, which is not critical and which we know is flaky and needs fixing.

PR #2650, opened 2022-02-08
fix two cbor list len mistakes

JaredCorduan:
In our ledger state serialization, two of the list lengths were incorrect. It passed the round trip tests since the broken encoding was consistent with the broken deserializer. This PR fixes the two mistakes. It also adds a golden test (the same one used by consensus) so that we can catch such changes and be aware of them before upgrading ledger in consensus.

I am hoping this makes its way into https://github.com/input-output-hk/ouroboros-network/pull/3599


lehins (approved):
LGTM

nc6 (approved):

PR #2654, opened 2022-02-11
Babbage era CDDL specification

JaredCorduan:
This is a first draft of the Babbage era CDDL specification

### DIFF
--- a/eras/babbage/test-suite/cddl-files/babbage.cddl
+++ b/eras/babbage/test-suite/cddl-files/babbage.cddl
@@ -1,0 +1,403 @@
+block =
+  [ header
+  , transaction_bodies         : [* transaction_body]
+  , transaction_witness_sets   : [* transaction_witness_set]
+  , auxiliary_data_set         : {* transaction_index => auxiliary_data }
+  , invalid_transactions       : [* transaction_index ]
+  ]; Valid blocks must also satisfy the following two constraints:
+   ; 1) the length of transaction_bodies and transaction_witness_sets
+   ;    must be the same
+   ; 2) every transaction_index must be strictly smaller than the
+   ;    length of transaction_bodies
+
+transaction =
+  [ transaction_body
+  , transaction_witness_set
+  , bool
+  , auxiliary_data / null
+  ]
+
+transaction_index = uint .size 2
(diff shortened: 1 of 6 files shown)


eras/babbage/test-suite/cddl-files/babbage.cddl
```diff
@@ -0,0 +1,404 @@
...
+; Note that in the post_alonzo_transaction_output, we do not allow
+; a transaction output to include both a datum hash and a datum.
+; In other words, keys 2 and 3 are mutually exclusive.
+post_alonzo_transaction_output =
+     { 0 => address
+     , 1 => value
+     , ? 2 => datum_hash
+     , ? 3 => datum      ; New; inline datum
+     , ? 4 => script     ; New; script reference
+     }
```
goolord, line 92: i like this 👍, how does deserializing the TxOut with cborg work though? do we decide how to branch by `peekTokenType`? cborg doesn't seem to have great machinery for deserializing choices

JaredCorduan: Unfortunately I don't think that there is a way to handle it besides branching on `peekTokenType`.

eras/babbage/test-suite/cddl-files/babbage.cddl
```diff
@@ -0,0 +1,403 @@
...
+
+ex_unit_prices =
+  [ mem_price: sub_coin, step_price: sub_coin ]
+
+language = 0 ; Plutus v1
+         / 1 ; Plutus v2
+
+costmdls = { * language => cost_model }
+
+cost_model = [ 166*166 int ]
```
JaredCorduan, line 327:
```suggestion
costmdls = { ? 0 : [ 166*166 int ]
           , ? 1 : [ (166+x)*(166+x) int ] ; where x accounts for new V2 cost model params
           }
```

JaredCorduan: @goolord - I'm going to remove this suggestion for now since we don't yet know what `x` is, but we can expect to add this back in pretty soon.

goolord (approved):

PR #2656, opened 2022-02-15
Added generic (Era agnostic tests) in Test.Cardano.Ledger.Generic 

TimSheard:
Test.Cardano.Ledger.Generic.Properties is meant to someday replace Test.Cardano.Ledger.Properties
Added many PrettyA instances for UTXOW UTXO UTXOS PPUP predicate failures.
Cleaned up the broken Policy type.


JaredCorduan (commented):
I'm excited to soon be able to use this code for making Babbage unit tests and examples! I think before we merge this we should make sure we only have one copy of the property tests (ie get rid of the alonzo specific ones in favor of the generic ones).

JaredCorduan (approved):
I'm approving, but I do hope we can reduce the duplication soon!

PR #2664, opened 2022-02-22
Extend using generic tests to the Babbage Era

TimSheard:
Added the Test type and runTest etc.
Added Babbage version of UTXOS, UTXO and UTXOW rules.


libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/Properties.hs
```diff
@@ -1239,11 +1240,12 @@ genTxAndLEDGERState proof = do
...
 applySTSByProof ::
-  (RuleTypeRep rtype, GoodCrypto (Crypto era)) =>
+  forall era.
+  (GoodCrypto (Crypto era)) =>
   Proof era ->
-  RuleContext rtype (Core.EraRule "LEDGER" era) ->
+  RuleContext 'Transition (Core.EraRule "LEDGER" era) ->
   (Either [PredicateFailure (Core.EraRule "LEDGER" era)] (State (Core.EraRule "LEDGER" era)))
-applySTSByProof (Babbage _) trc = runShelleyBase $ applySTS trc
+applySTSByProof (Babbage _) _trc = undefined -- runShelleyBase $ applySTS @(Core.EraRule "LEDGER" (BabbageEra (Crypto era))) _trc
```
lehins: Forgotten `undefined`

eras/babbage/impl/src/Cardano/Ledger/Babbage/Scripts.hs
```diff
@@ -0,0 +1,110 @@
...
+    txbody = getField @"body" tx
+    allInputs = Set.union (spendInputs' txbody) (referenceInputs' txbody)
+    witnessset = getField @"wits" tx
+    witnessScripts = txscripts' witnessset
+    referenceScripts = Set.foldl' accum Map.empty allInputs
+    accum ans txin =
+      case getScript txin of
+        Nothing -> ans
+        Just script -> Map.insert (hashScript @era script) script ans
+    getScript txin = do txout <- SplitMap.lookup txin m; extractScript txout
```
lehins: Usage of semicolons are rather unusual in Haskell. For the sake of the future readers:
```suggestion
    getScript txin = SplitMap.lookup txin m >>= extractScript
```

eras/babbage/impl/src/Cardano/Ledger/Babbage/Scripts.hs
```diff
@@ -0,0 +1,110 @@
...
+        case datum of
+          NoDatum -> []
+          Datum d -> [d]
+          DatumHash hash ->
+            case Map.lookup hash (unTxDats $ txdats' (getField @"wits" tx)) of
+              Nothing -> []
+              Just dat -> [dataToBinaryData dat]
+
+extractScript :: Core.TxOut era -> Maybe (Core.Script era)
+extractScript _txout = undefined
```
lehins: Another `undefined`

eras/babbage/impl/src/Cardano/Ledger/Babbage/Scripts.hs
```diff
@@ -0,0 +1,110 @@
...
+    case SplitMap.lookup txin m of
+      Nothing -> []
+      Just (TxOut _ _ datum _refScript) ->
+        case datum of
+          NoDatum -> []
+          Datum d -> [d]
+          DatumHash hash ->
+            case Map.lookup hash (unTxDats $ txdats' (getField @"wits" tx)) of
+              Nothing -> []
+              Just dat -> [dataToBinaryData dat]
```
lehins: Why is `getDatum` returning a list. It can only return either a n empty list or a singleton. So it should be a `Maybe (BinaryData era)`, right? Also it doesn't look like `
```suggestion
  Just (BinaryData era)
getDatum tx (UTxO m) sp = case  of
  txin <- getTxIn sp
  TxOut _ _ datum _refScript <- SplitMap.lookup txin m
  case datum of
     NoDatum -> Nothing
     Datum d -> Just d
     DatumHash hash ->
       dataToBinaryData <$> Map.lookup hash (unTxDats $ txdats' (getField @"wits" tx))
```

eras/babbage/impl/src/Cardano/Ledger/Babbage/Scripts.hs
```diff
@@ -0,0 +1,110 @@
...
+import GHC.Records (HasField (..))
+
+-- ====================================================================
+
+hasTxIn :: ScriptPurpose crypto -> Maybe (TxIn crypto)
+hasTxIn (Spending txin) = Just txin
+-- Only the Spending ScriptPurpose contains TxIn
+hasTxIn (Minting _policyid) = Nothing
+hasTxIn (Rewarding _rewaccnt) = Nothing
+hasTxIn (Certifying _dcert) = Nothing
```
lehins: Has usually implies a function returns a boolean, maybe `get` instead?

```suggestion
getTxIn :: ScriptPurpose crypto -> Maybe (TxIn crypto)
getTxIn (Spending txin) = Just txin
-- Only the Spending ScriptPurpose contains TxIn
getTxIn (Minting _policyid) = Nothing
getTxIn (Rewarding _rewaccnt) = Nothing
getTxIn (Certifying _dcert) = Nothing
```

lehins (commented):
I skimmed through a PR overall looks pretty good. I left few comments. I'll do a proper review once the PR is finished

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/Properties.hs
TimSheard: It is now gone.

eras/babbage/impl/src/Cardano/Ledger/Babbage/Scripts.hs
TimSheard: Semi colons allow short do statements to exist on one line so they have their uses. But I like your suggestion too.

eras/babbage/impl/src/Cardano/Ledger/Babbage/Scripts.hs
TimSheard: I started some thing and never finished. But it is fixed now.

eras/babbage/impl/src/Cardano/Ledger/Babbage/Scripts.hs
TimSheard: This is what the Spec says, It returns a list.

eras/babbage/impl/src/Cardano/Ledger/Babbage/Scripts.hs
TimSheard: sounds good to me.

eras/babbage/impl/src/Cardano/Ledger/Babbage/Scripts.hs
lehins: @TimSheard I think the spec should be fixed, but @WhatisRT said that it would be inconvenient and we can use `Maybe` in the implementation instead:

> I think what you should really do instead is to have a Maybe in the implementation, and keep it in the spec as it is

TimSheard: I am not so sure.  In the rules this is used to build a list of BinaryData which is then concateted with other lists, and then serialized to pass to the TxInfo. Something like this
serialize (getDatum x ++ getAnotherKindOfData y ++ getYetAnother z)

lehins: Yeah as far as the spec is concerned, I understand that it can be more convenient to return a list, but then in our implementation it should be:
```
serialize (maybeToList (getDatum x) ++ getAnotherKindOfData y ++ getYetAnother z)
```

TimSheard:
Added missing arbitrary and tripping tests for BabbabgeUtxoPred

eras/babbage/impl/src/Cardano/Ledger/Babbage/Rules/UtxoSave.hs
```diff
@@ -0,0 +1,225 @@
+{-# LANGUAGE DataKinds #-}
```
JaredCorduan: did you mean to check this file in?

eras/babbage/impl/src/Cardano/Ledger/Babbage/Rules/UtxosSave.hs
```diff
@@ -0,0 +1,134 @@
+{-# LANGUAGE BangPatterns #-}
```
JaredCorduan: did you meant to check in this file?

eras/babbage/impl/src/Cardano/Ledger/Babbage/Rules/UtxowSave.hs
```diff
@@ -0,0 +1,196 @@
+{-# LANGUAGE DataKinds #-}
```
JaredCorduan: did you meant to check in this file?

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Rules/Utxos.hs
```diff
@@ -82,42 +88,50 @@ import NoThunks.Class (NoThunks)
 -- The UTXOS transition system
 --------------------------------------------------------------------------------
 
+{- These are the Field constraints needed. Since this STS instance is only going to
+   be used once, We can make all these go away by fixing the Core.XX types to the
+   concrete types used in this instance.
+type UtxosFieldNeeds era =
```
JaredCorduan: can we delete this comment?

eras/babbage/impl/src/Cardano/Ledger/Babbage/Rules/UtxoSave.hs
TimSheard: No, I will delete it.

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Rules/Utxos.hs
TimSheard: sure. That was really just notes for when comparing the old and new versions

JaredCorduan (approved):
Looks great to me! Thank you for this huge effort!!

eras/babbage/impl/src/Cardano/Ledger/Babbage/Rules/UtxosSave.hs
TimSheard: deleted

eras/babbage/impl/src/Cardano/Ledger/Babbage/Rules/UtxowSave.hs
TimSheard: deleted

PR #2665, opened 2022-02-23
bump index state to 2022-02-18T00:00:00Z

JaredCorduan:
and also specify the GHC version in the cabal.project


nc6:
I think we don't need to specify it in `cabal.project`

lehins:
Looks like `cardano-crypto-class` needs to account for new release of `aeson`

nc6:
This is fine on hydra, it looks like the github controlled builds are using 8.10.4

lehins (approved):

PR #2666, opened 2022-02-23
Babbage serialization tests

goolord:


eras/babbage/test-suite/test/Test/Cardano/Ledger/Babbage/Serialisation/Tripping.hs
```diff
@@ -0,0 +1,78 @@
...
+import qualified Data.ByteString.Lazy.Char8 as BSL
+import Data.Roundtrip (roundTrip, roundTrip', roundTripAnn)
+import Test.Cardano.Ledger.Babbage.Serialisation.Generators ()
+import Test.Cardano.Ledger.Shelley.ConcreteCryptoTypes
+import Test.Cardano.Ledger.ShelleyMA.Serialisation.Generators ()
+import Test.Tasty
+import Test.Tasty.QuickCheck
+import Text.Show.Pretty (ppShow)
+
+trippingF ::
```
goolord, line 30: it would be nice to have a place to put common testing code like this

lehins (commented):
~I skimmed through a PR overall looks pretty good. I left few comments. I'll do a proper review once the PR is finished~

Sorry I commented at a wrong PR. I was reviewing another one

eras/babbage/test-suite/src/Test/Cardano/Ledger/Babbage/Serialisation/Generators.hs
```diff
@@ -0,0 +1,348 @@
...
+      <*> arbitrary
+      <*> arbitrary
+      <*> arbitrary
+      <*> arbitrary
+      <*> genMintValues
+      <*> arbitrary
+      <*> arbitrary
+      <*> arbitrary
+
+instance Arbitrary (Data era) where
```
JaredCorduan: I think we should be able to re-use a lot of the alonzo generators, such at the Data one.

https://github.com/input-output-hk/cardano-ledger/blob/7aae0d97b87fa508de8799510b49739b65e28335/eras/alonzo/test-suite/src/Test/Cardano/Ledger/Alonzo/Serialisation/Generators.hs#L55-L56

eras/babbage/test-suite/src/Test/Cardano/Ledger/Babbage/Serialisation/Generators.hs
```diff
@@ -0,0 +1,348 @@
...
+  arbitrary = Data <$> arbitrary
+
+genPair :: Gen a -> Gen b -> Gen (a, b)
+genPair = liftA2 (,)
+
+instance Arbitrary Plutus.Data where
```
JaredCorduan: reuse alonzo?

eras/babbage/test-suite/src/Test/Cardano/Ledger/Babbage/Serialisation/Generators.hs
```diff
@@ -0,0 +1,348 @@
...
+            ]
+      gendata _ = oneof [Plutus.I <$> arbitrary, Plutus.B <$> arbitrary]
+
+instance
+  ( UsesScript era,
+    Ord (Core.Script era),
+    Core.Script era ~ Script era,
+    Arbitrary (Core.Script era)
+  ) =>
+  Arbitrary (AuxiliaryData era)
```
JaredCorduan: reuse alonzo? I'll stop making this comment now, but I think it applies to a lot of the generators below as well.

eras/babbage/test-suite/src/Test/Cardano/Ledger/Babbage/Trace.hs
```diff
@@ -0,0 +1,19 @@
...
+
+-- | Special tracing code usefull for debugging tests
+module Test.Cardano.Ledger.Babbage.Trace (trace, ptrace, occaisionally) where
+
+import Cardano.Ledger.Pretty (PrettyA (..), ppMap)
+import Data.Hashable (Hashable (..))
+import Data.Map (Map)
+import Debug.Trace (trace)
+
+ptrace :: PrettyA t => [Char] -> t -> a -> a
```
JaredCorduan: should these maybe live in cardano-ledger-pretty?

eras/babbage/test-suite/test/Test/Cardano/Ledger/Babbage/Serialisation/Tripping.hs
JaredCorduan: I agree! Let's make this a talking point next time the team meets. `cardano-ledger-test` is nice for things that live "at the end of all eras", but we don't have a good place for testing support that lives "before all eras".

JaredCorduan (commented):
This is looking really great, thanks! I made some comments about trying to re-use generators from Alonzo, but besides that I think this is great.

JaredCorduan (approved):
looks great, thanks @goolord !

PR #2567, opened 2021-11-29
Adding a new datatype to replace UTxO currently a Data.Map

TimSheard:
Introduces the SplitMap. Added basic operations for SplitMap that
every Map should supply (insert, delete, union, intersection etc).
Added Basic and Iter instances for Control.SetAlgebra.

PR #2667, opened 2022-02-24
Swap splitmap for map

lehins:
This PR fixes major regression by reverting usage of the custom key value data structure that we've used for UTxO, which  was designed in order to reduce memory overhead in #2567 and #2592

It has not been decided yet if we want to remove that approach completely or do we want to give it a little more effort in order to attempt and improve the performance of SplitMap. It would be preferable to get it to perform well, since we can confirm about ~15% in memory reduction overhead for the whole ledger state. Because of that uncertainty the original SplitMap module is backed up in `SplitMapOriginal` module with its whole git history, however it is not included in the project, so it does not contribute to the build times.

If we do decide to remove SplitMap, we'll need to go through the whole code base and remove its usage, so dumping SplitMap would require some effort as well. 

For now this PR removes the regression, while putting minimal efforts on developers regardless of the future decision.


JaredCorduan (approved):
Thank you for the swiftness on this @lehins !

PR #2668, opened 2022-02-24
Improve performance txout with benchmarks

lehins:
This PR improves performance of TxOut and adds benchmarks. In particular serialization and TxOutCompact+TxOutCompactDH patterns

There is also a fix for couple of Babbage's TxOut accessor functions

### DIFF
--- a/libs/cardano-ledger-core/src/Cardano/Ledger/Val.hs
+++ b/libs/cardano-ledger-core/src/Cardano/Ledger/Val.hs
@@ -21,8 +21,9 @@ module Cardano.Ledger.Val
 where
 
 import Cardano.Binary (Decoder, Encoding, decodeWord64, toCBOR)
-import Cardano.Ledger.Coin (Coin (..), DeltaCoin (..))
+import Cardano.Ledger.Coin (Coin (..), CompactForm (..), DeltaCoin (..))
 import Cardano.Ledger.Compactible (Compactible (..))
+import Data.Coerce
 import Data.Foldable (foldl')
 import Data.Group (Abelian)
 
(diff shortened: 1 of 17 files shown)


libs/cardano-ledger-core/src/Cardano/Ledger/Val.hs
```diff
@@ -66,6 +67,12 @@ class
   -- | If a quantity is stored in only one of 'v1' or 'v2', we use 0 for the missing quantity.
   pointwise :: (Integer -> Integer -> Bool) -> t -> t -> Bool
 
+  isAdaOnly :: t -> Bool
```
JaredCorduan, line 73: the only trouble with renaming this from `adaOnly` to `isAdaOnly` is that both the ShelleyMA and the Alonzo spec are using `adaOnly`.

JaredCorduan (approved):
looks good to me! I only had one comment regarding changing a variable name :grimacing: which creates a diversion from two of the specs...

libs/cardano-ledger-core/src/Cardano/Ledger/Val.hs
lehins: Would it be problematic if we renamed it in the spec as well? `isAdaOnly` is a lot more descriptive name. It would be much better if we could improve the spec as well as the code. So, if it is possible I can go through and change it in the spec, but if it is a problem for some reason, I can definitely just move the `adaOnly` function into the `Val` class.

JaredCorduan: nope, renaming it in the spec is my favorite option (I agree that `isAdaOnly` is more descriptive).

lehins: Renamed.

PR #2669, opened 2022-02-25
Fix benchmarks

lehins:
Fixing benchmarks requires bunch of NFData instance.

Many orphans have been moved to cardano-base, which means we need to get https://github.com/input-output-hk/cardano-base/pull/260 be merged for this PR

Also removed old symlinks.

### DIFF
--- a/.github/workflows/haskell.yml
+++ b/.github/workflows/haskell.yml
@@ -41,7 +41,7 @@ jobs:
       if: matrix.os == 'ubuntu-latest'
       run: |
         sudo apt-get update
-        sudo apt-get -y install libsodium23 libsodium-dev
+        sudo apt-get -y install libsodium23 libsodium-dev libsecp256k1-dev
         sudo apt-get -y remove --purge software-properties-common
         sudo apt-get -y autoremove
 
--- a/eras/alonzo/test
+++ b/eras/alonzo/test
@@ -1,1 +1,0 @@
-test-suite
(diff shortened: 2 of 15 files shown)


.github/workflows/haskell.yml
```diff
@@ -41,7 +41,7 @@ jobs:
       if: matrix.os == 'ubuntu-latest'
       run: |
         sudo apt-get update
-        sudo apt-get -y install libsodium23 libsodium-dev
+        sudo apt-get -y install libsodium23 libsodium-dev libsecp256k1-dev
```
JaredCorduan, line 44: so easy a fix now that I now what to look for :laughing: thanks!

eras/alonzo/test
```diff
@@ -1 +0,0 @@
-test-suite
```
JaredCorduan, line 1: :tada: good riddance!

libs/cardano-ledger-test/bench/Bench/Cardano/Ledger/ApplyTx.hs
```diff
@@ -177,8 +182,8 @@ applyTxBenchmarks =
         "Deserialise Shelley Tx"
         [ deserialiseTxEra (Proxy @ShelleyBench),
           deserialiseTxEra (Proxy @AllegraBench),
-          deserialiseTxEra (Proxy @MaryBench),
-          deserialiseTxEra (Proxy @AlonzoBench)
+          deserialiseTxEra (Proxy @MaryBench)
+          --deserialiseTxEra (Proxy @AlonzoBench)
```
JaredCorduan, line 186: why are we disabling the tests for Alonzo?

JaredCorduan (commented):
looks good to me, provided there is a good reason to disable the benchmarks on the alonzo tx deserialization

libs/cardano-ledger-test/bench/Bench/Cardano/Ledger/ApplyTx.hs
lehins: It's not a test, but a benchmark, we can afford to disable a benchmark. Which is important :smile: 

It has to do with problems with generator for `Alonzo`. See a few lines above `benchApplyTx` is also disabled for `Alonzo`. It looks like it will take more effort to get it implemented for Alonzo.

JaredCorduan (approved):

PR #2670, opened 2022-02-25
Add a ledger event for successful Plutus scripts

### LINKED ISSUE #2659: Plutus script ledger event
When a Plutus script fails, the predicate failure includes enough information to re-run the script with all the arguments passed to it. See [here](https://github.com/input-output-hk/cardano-ledger/blob/2048cd4679ef3155d88a445e0f818abad09a0827/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxInfo.hs#L608).

It would also be very helpful to be able to re-run _succeeding_ Plutus scripts.

Let's create a new ledger event which provides the same `PlutusDebug` information for successful scripts. Probably somewhere around here: https://github.com/input-output-hk/cardano-ledger/blob/2048cd4679ef3155d88a445e0f818abad09a0827/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/PlutusScriptApi.hs#L264

JaredCorduan:
This PR adds a ledger event for successful Plutus scripts, providing all the information needed to re-run the scripts.

# Redux Notes for reviewers

There are two new events:
* `SuccessfulPlutusScriptsEvent [PlutusDebug]`
* `FailedPlutusScriptsEvent [PlutusDebug]`

In the case of no failures, with the transaction marked as `IsValid == True`, all the information needed to re-run all the scripts in a given transaction are a part of a `SuccessfulPlutusScriptsEvent` event. In the case of failures, with the transaction marked as `IsValid == False`, there are two events, one for the successes and one for the failures. No event is transmitted otherwise, since this triggers a predicate failure.

## Other notable changes

The type returned by `runPLCScript` (the ledger function which wraps the Plutus evaluator) is named `ScriptResult` and has been changed from

```
data ScriptResult = Passes | Fails ![FailureDescription]
```
to
```
data ScriptResult  = Passes [PlutusDebug]  | Fails [PlutusDebug] (NonEmpty ScriptFailure)
```

In other words, `ScriptResult` now collects the successful script information as well.  There is only a small difference between `FailureDescription` and `ScriptFailure`. The former contains an encoded bytestring with the needed information to re-run scripts, the later contains the un-encoded data. We cannot change the serialization of the predicates failures without breaking the node-to-client protocol, and since `FailureDescription` is used for a predicate failure we keep this type around (though it is moved to `Cardano.Ledger.Alonzo.Rules.Utxos`). The `ScriptFailure` type allows use to transmit the ledger event without encoding the data.

Lastly, the type `PlutusDebugInfo` has one small change in the `DebugInfo` constructor. Instead of the less helpful `String`, it now contains the actual Plutus error. It also now contains the `PlutusDebug`.

closes #2659

### DIFF
--- a/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Rules/Utxos.hs
+++ b/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Rules/Utxos.hs
@@ -231,9 +231,9 @@ scriptsNotValidateTransition = do
       whenFailureFree $
         case evalScripts @era (getField @"_protocolVersion" pp) tx sLst of
           Passes _ps -> False ?!## ValidationTagMismatch (getField @"isValid" tx) PassedUnexpectedly
-          Fails ps fs ->
+          Fails ps fs -> do
             tellEvent (SuccessfulPlutusScriptsEvent ps)
-              >> tellEvent (FailedPlutusScriptsEvent (scriptFailuresToPlutusDebug fs))
+            tellEvent (FailedPlutusScriptsEvent (scriptFailuresToPlutusDebug fs))
     Left info -> failBecause (CollectErrors info)
 
   let !_ = traceEvent invalidEnd ()
@@ -280,7 +280,7 @@ instance FromCBOR FailureDescription where
 scriptFailureToFailureDescription :: ScriptFailure -> FailureDescription
 scriptFailureToFailureDescription (OnePhaseSF t) = OnePhaseFailure t
 scriptFailureToFailureDescription (PlutusSF t pd) =
-  PlutusFailure t (B64.encode . serialize' $ pd)
+  PlutusFailure t (B64.encode $ serialize' pd)
 
 scriptFailuresToPredicateFailure :: NonEmpty ScriptFailure -> NonEmpty FailureDescription
 scriptFailuresToPredicateFailure = fmap scriptFailureToFailureDescription
--- a/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Rules/Utxos.hs
+++ b/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Rules/Utxos.hs
@@ -132,7 +132,7 @@ instance
 data UtxosEvent era
   = AlonzoPpupToUtxosEvent (Event (Core.EraRule "PPUP" era))
   | SuccessfulPlutusScriptsEvent [PlutusDebug]
-  | FailedPlutusScriptsEvent [PlutusDebug]
+  | FailedPlutusScriptsEvent (NonEmpty PlutusDebug)
 
 instance
   ( Era era,
@@ -278,15 +278,14 @@ instance FromCBOR FailureDescription where
       dec n = Invalid n
 
 scriptFailureToFailureDescription :: ScriptFailure -> FailureDescription
-scriptFailureToFailureDescription (OnePhaseSF t) = OnePhaseFailure t
 scriptFailureToFailureDescription (PlutusSF t pd) =
   PlutusFailure t (B64.encode $ serialize' pd)
 
 scriptFailuresToPredicateFailure :: NonEmpty ScriptFailure -> NonEmpty FailureDescription
 scriptFailuresToPredicateFailure = fmap scriptFailureToFailureDescription
 
-scriptFailuresToPlutusDebug :: NonEmpty ScriptFailure -> [PlutusDebug]
-scriptFailuresToPlutusDebug sfs = [pdb | PlutusSF _ pdb <- toList sfs]
+scriptFailuresToPlutusDebug :: NonEmpty ScriptFailure -> NonEmpty PlutusDebug
+scriptFailuresToPlutusDebug = fmap (\(PlutusSF _ pdb) -> pdb)
 
 data TagMismatchDescription
   = PassedUnexpectedly
--- a/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Rules/Utxos.hs
+++ b/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Rules/Utxos.hs
@@ -26,10 +26,13 @@ module Cardano.Ledger.Alonzo.Rules.Utxos
     UtxosEvent (..),
     (?!##),
     ConcreteAlonzo,
+    FailureDescription (..),
+    scriptFailuresToPredicateFailure,
+    scriptFailuresToPlutusDebug,
   )
 where
 
-import Cardano.Binary (FromCBOR (..), ToCBOR (..))
+import Cardano.Binary (FromCBOR (..), ToCBOR (..), serialize')
 import qualified Cardano.Ledger.Alonzo.PParams as Alonzo
 import Cardano.Ledger.Alonzo.PlutusScriptApi
   ( CollectError,
@@ -43,7 +46,7 @@ import Cardano.Ledger.Alonzo.Tx
     ValidatedTx (..),
   )
 import qualified Cardano.Ledger.Alonzo.TxBody as Alonzo
-import Cardano.Ledger.Alonzo.TxInfo (ExtendedUTxO (..), FailureDescription (..), ScriptResult (..))
+import Cardano.Ledger.Alonzo.TxInfo (ExtendedUTxO (..), PlutusDebug, ScriptFailure (..), ScriptResult (..))
 import qualified Cardano.Ledger.Alonzo.TxWitness as Alonzo
 import Cardano.Ledger.BaseTypes
   ( Globals,
@@ -71,13 +74,17 @@ import Cardano.Ledger.Val as Val
 import Control.Monad.Except (MonadError (throwError))
 import Control.Monad.Trans.Reader (asks)
 import Control.State.Transition.Extended
+import Data.ByteString as BS (ByteString)
+import qualified Data.ByteString.Base64 as B64
 import Data.Coders
 import qualified Data.Compact.SplitMap as SplitMap
 import Data.Foldable (toList)
 import Data.List (intercalate)
+import Data.List.NonEmpty (NonEmpty (..))
 import qualified Data.Map.Strict as Map
 import Data.Sequence.Strict (StrictSeq)
 import Data.Set (Set)
+import Data.Text (Text)
 import Debug.Trace (traceEvent)
 import GHC.Generics (Generic)
 import GHC.Records (HasField (..))
@@ -122,8 +129,10 @@ instance
   type Event (UTXOS era) = UtxosEvent era
   transitionRules = [utxosTransition]
 
-newtype UtxosEvent era
-  = UpdateEvent (Event (Core.EraRule "PPUP" era))
+data UtxosEvent era
+  = AlonzoPpupToUtxosEvent (Event (Core.EraRule "PPUP" era))
+  | SuccessfulPlutusScriptsEvent [PlutusDebug]
+  | FailedPlutusScriptsEvent [PlutusDebug]
 
 instance
   ( Era era,
@@ -134,7 +143,7 @@ instance
   Embed (PPUP era) (UTXOS era)
   where
   wrapFailed = UpdateFailure
-  wrapEvent = UpdateEvent
+  wrapEvent = AlonzoPpupToUtxosEvent
 
 utxosTransition ::
   forall era.
@@ -184,12 +193,12 @@ scriptsValidateTransition = do
   case collectTwoPhaseScriptInputs ei sysSt pp tx utxo of
     Right sLst ->
       case evalScripts @era (getField @"_protocolVersion" pp) tx sLst of
-        Fails sss ->
+        Fails _ps fs ->
           False
             ?!## ValidationTagMismatch
               (getField @"isValid" tx)
-              (FailedUnexpectedly sss)
-        Passes -> pure ()
+              (FailedUnexpectedly (scriptFailuresToPredicateFailure fs))
+        Passes ps -> tellEvent (SuccessfulPlutusScriptsEvent ps)
     Left info -> failBecause (CollectErrors info)
 
   () <- pure $! traceEvent validEnd ()
@@ -221,8 +230,10 @@ scriptsNotValidateTransition = do
     Right sLst ->
       whenFailureFree $
         case evalScripts @era (getField @"_protocolVersion" pp) tx sLst of
-          Passes -> False ?!## ValidationTagMismatch (getField @"isValid" tx) PassedUnexpectedly
-          Fails _sss -> pure ()
+          Passes _ps -> False ?!## ValidationTagMismatch (getField @"isValid" tx) PassedUnexpectedly
+          Fails ps fs ->
+            tellEvent (SuccessfulPlutusScriptsEvent ps)
+              >> tellEvent (FailedPlutusScriptsEvent (scriptFailuresToPlutusDebug fs))
     Left info -> failBecause (CollectErrors info)
 
   let !_ = traceEvent invalidEnd ()
@@ -250,9 +261,36 @@ invalidEnd = intercalate "," ["[LEDGER][SCRIPTS_NOT_VALIDATE_TRANSITION]", "END"
 -- =============================================
 -- PredicateFailure data type for UTXOS
 
+data FailureDescription
+  = OnePhaseFailure Text
+  | PlutusFailure Text BS.ByteString
+  deriving (Show, Eq, Generic, NoThunks)
+
+instance ToCBOR FailureDescription where
+  toCBOR (OnePhaseFailure s) = encode $ Sum OnePhaseFailure 0 !> To s
+  toCBOR (PlutusFailure s b) = encode $ Sum PlutusFailure 1 !> To s !> To b
+
+instance FromCBOR FailureDescription where
+  fromCBOR = decode (Summands "FailureDescription" dec)
+    where
+      dec 0 = SumD OnePhaseFailure <! From
+      dec 1 = SumD PlutusFailure <! From <! From
+      dec n = Invalid n
+
+scriptFailureToFailureDescription :: ScriptFailure -> FailureDescription
+scriptFailureToFailureDescription (OnePhaseSF t) = OnePhaseFailure t
+scriptFailureToFailureDescription (PlutusSF t pd) =
+  PlutusFailure t (B64.encode . serialize' $ pd)
+
+scriptFailuresToPredicateFailure :: NonEmpty ScriptFailure -> NonEmpty FailureDescription
+scriptFailuresToPredicateFailure = fmap scriptFailureToFailureDescription
+
+scriptFailuresToPlutusDebug :: NonEmpty ScriptFailure -> [PlutusDebug]
+scriptFailuresToPlutusDebug sfs = [pdb | PlutusSF _ pdb <- toList sfs]
+
 data TagMismatchDescription
   = PassedUnexpectedly
-  | FailedUnexpectedly [FailureDescription]
+  | FailedUnexpectedly (NonEmpty FailureDescription)
   deriving (Show, Eq, Generic, NoThunks)
 
 instance ToCBOR TagMismatchDescription where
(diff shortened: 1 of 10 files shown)


eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxInfo.hs
```diff
@@ -444,28 +444,28 @@ instance FromCBOR FailureDescription where
...
 instance ToCBOR ScriptResult where
-  toCBOR Passes = encode $ Sum Passes 0
+  toCBOR (Passes ps) = encode $ Sum Passes 0 !> To ps
   toCBOR (Fails fs) = encode $ Sum Fails 1 !> To fs
 
 instance FromCBOR ScriptResult where
   fromCBOR = decode (Summands "ScriptResult" dec)
     where
-      dec 0 = SumD Passes
+      dec 0 = SumD Passes <! From
```
JaredCorduan: Note that the `Passes` constructor (unlike `Fails` below) is not a part of any predicate failure. Therefore this change to the serialization does _not_ break the node-to-client protocol.

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxInfo.hs
```diff
@@ -444,28 +444,28 @@ instance FromCBOR FailureDescription where
...
       dec 1 = SumD Fails <! From
       dec n = Invalid n
 
-andResult :: ScriptResult -> ScriptResult -> ScriptResult
-andResult Passes Passes = Passes
-andResult Passes ans = ans
-andResult ans Passes = ans
-andResult (Fails xs) (Fails ys) = Fails (xs ++ ys)
-
+-- | The ScriptResult operation is biased towards failure.
```
michaelpj: I guess the upshot of  this is that if I have a transaction which has some passing and some failing scripts, then I won't get any events for the succeeding scripts, only the failing ones? That's okay (and it's a weird case), just checking.

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxInfo.hs
```diff
@@ -444,28 +444,28 @@ instance FromCBOR FailureDescription where
...
+--  with failures, but a failure always overrides successes.
 instance Semigroup ScriptResult where
-  (<>) = andResult
+  (Passes xs) <> (Passes ys) = Passes (xs <> ys)
+  (Passes _) <> fs@(Fails _) = fs
+  fs@(Fails _) <> (Passes _) = fs
+  (Fails xs) <> (Fails ys) = Fails (xs <> ys)
 
 data PlutusDebug = PlutusDebug
   { debugCostModel :: CostModel,
```
michaelpj: Should this be a `CostModelParams`? Not sure.

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxInfo.hs
```diff
@@ -444,28 +444,28 @@ instance FromCBOR FailureDescription where
...
+--  Successes are combined with successes, failures are combined
+--  with failures, but a failure always overrides successes.
 instance Semigroup ScriptResult where
-  (<>) = andResult
+  (Passes xs) <> (Passes ys) = Passes (xs <> ys)
+  (Passes _) <> fs@(Fails _) = fs
+  fs@(Fails _) <> (Passes _) = fs
+  (Fails xs) <> (Fails ys) = Fails (xs <> ys)
 
 data PlutusDebug = PlutusDebug
```
michaelpj: Out of the diff, but maybe `debugExUnits` should be `debugEvaluationBudget` so it says what it is rather than just the type.

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxInfo.hs
```diff
@@ -474,7 +474,7 @@ data PlutusDebug = PlutusDebug
     debugData :: [PV1.Data],
```
michaelpj: `debugScriptArguments`?

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxInfo.hs
```diff
@@ -474,7 +474,7 @@ data PlutusDebug = PlutusDebug
...
     debugVersion :: Language
```
michaelpj: Is this actually sufficient? If things change substantially between versions, maybe we'll need to change the structure for different languages. So we might need `PlutusV1Debug`, `PlutusV2Debug` etc. But I guess we can do that later.

michaelpj: Speaking of other languages: if we're doing this, should we be consistent and emit events for success/failures of native scripts (or whatever we're calling them now)?

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxInfo.hs
```diff
@@ -542,7 +542,7 @@ runPLCScript proxy lang (CostModel cost) scriptbytestring units ds =
     scriptbytestring
     ds of
     (_, Left e) -> explainPlutusFailure proxy lang scriptbytestring e ds (CostModel cost) units
```
michaelpj: How is this actually going to look in the events? From the look of this function, here we're going to get a predicate failure (?) with the _serialized_ version of `PlutusDebug`, whereas in the success case we'll just get the structure itself? It seems like that might be a bit of a pain to analyze consistently.

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxInfo.hs
michaelpj: Worth documenting perhaps.

JaredCorduan: Thanks for bringing this up, it's a bit tricky and I don't know how best to handle it.

If a transaction contains one or more failing scripts, then:
* if the transaction is marked as `IsValid = True`, then the transaction will be rejected. The predicate failure will collect all the failures, and the node logs will display all the needed information to re-run the failed scripts. but indeed there is no mention of the succeeding scripts in this case.
* if the transaction is marked as `IsValid = False`, then as of right now neither the logs nor the events record this information. because of this semigroup instance, it would be easy to make an event for the failures, but an event for all the scripts would need to either re-run everything or we could redo how `evalScripts` works altogether (ie not use `<>`). I'll point out where in this PR we could put the event for the failures that are marked as failures.

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxInfo.hs
JaredCorduan: sounds great!

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxInfo.hs
JaredCorduan: `CostModel` is what we currently pass to the evaluator, so that's why it is captured here.

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxInfo.hs
JaredCorduan: I like it!

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxInfo.hs
JaredCorduan: yea, are probably right to think now about future proofing this. So I guess you are proposing using multiple constructors? that's reasonable. I'm not sure if we should do this now or when it is needed...

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Rules/Utxos.hs
```diff
@@ -221,7 +222,7 @@ scriptsNotValidateTransition = do
   case collectTwoPhaseScriptInputs ei sysSt pp tx utxo of
     Right sLst ->
       case evalScripts @era tx sLst of
-        Passes -> False ?!## ValidationTagMismatch (getField @"isValid" tx) PassedUnexpectedly
+        (Passes _) -> False ?!## ValidationTagMismatch (getField @"isValid" tx) PassedUnexpectedly
         Fails _sss -> pure ()
```
JaredCorduan: This is where we could place an event for those transactions that are successfully submitted to the chain which have a failing script and are marked as `IsValid = False`. `_sss` will collect only the failures, given the `<>` in this PR. we could redo how it all works to collect successes and failures.

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxInfo.hs
michaelpj: As a complete outsider, what I would have sort of expected would be that the events would be emitted as a side effect of doing the validation, and orthogonally to how the actual results of validation are reported. But I don't know if that makes any sense here :shrug:

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxInfo.hs
michaelpj: Is that because both `CostModel` and `CostModelParams` are synonyms for `Map Text Integer` and so they're equivalent? :thinking:

JaredCorduan: Sorry, I was confused. Yes, it is essentially a `CostModelParams`. But wrapped in a newtype, and with a custom serialization.

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxInfo.hs
michaelpj: I guess? In the spirit of "Plutus V1 and Plutus V2 are separate languages as far as the ledger is concerned", it seems reasonable to have a `ScriptEvaluationEvent` constructor per language. Not confident though.

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Rules/Utxos.hs
michaelpj: I do wonder whether this is a symptom of what I mentioned above, which is that it looks like the events and the result of checking are communicated in the same band...

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxInfo.hs
michaelpj: I guess I was expecting this to be more-or-less the arguments to the evaluation functions in the Plutus ledger api module.

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxInfo.hs
JaredCorduan: This is all related to https://github.com/input-output-hk/cardano-ledger/pull/2670/files#r816904931

`evalScripts` is used both in the case that `IsValid` is true and false. Currently, in this PR, I've avoided the `IsValid = False` case (which is maybe a mistake) with respect to the events. This left here is currently only feeding into a predicate failure, not the events.

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxInfo.hs
JaredCorduan: it's not quite that simple (at least not without a big refactor), but I can probably make it look more like that from the outside.

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Rules/Utxos.hs
JaredCorduan: indeed it is. but treating the events as side effects was a more dramatic change than I had wanted to make.

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxInfo.hs
JaredCorduan: it's no longer a record, but a sum type, so bye bye accessor names :)

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxInfo.hs
JaredCorduan: it's now a sum type, should be more future proof now.

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Rules/Utxos.hs
```diff
@@ -221,8 +230,10 @@ scriptsNotValidateTransition = do
     Right sLst ->
       whenFailureFree $
         case evalScripts @era (getField @"_protocolVersion" pp) tx sLst of
-          Passes -> False ?!## ValidationTagMismatch (getField @"isValid" tx) PassedUnexpectedly
-          Fails _sss -> pure ()
+          Passes _ps -> False ?!## ValidationTagMismatch (getField @"isValid" tx) PassedUnexpectedly
+          Fails ps fs ->
+            tellEvent (SuccessfulPlutusScriptsEvent ps)
+              >> tellEvent (FailedPlutusScriptsEvent (scriptFailuresToPlutusDebug fs))
```
lehins: Minor syntactic suggestion
```suggestion
          Fails ps fs -> do
            tellEvent (SuccessfulPlutusScriptsEvent ps)
            tellEvent (FailedPlutusScriptsEvent (scriptFailuresToPlutusDebug fs))
```

eras/babbage/impl/src/Cardano/Ledger/Babbage/Rules/Utxos.hs
```diff
@@ -199,8 +201,10 @@ scriptsNo = do
       {- sLst := collectTwoPhaseScriptInputs pp tx utxo -}
       {- isValid tx = evalScripts tx sLst = False -}
       case evalScripts @era (getField @"_protocolVersion" pp) tx sLst of
-        Passes -> False ?!## ValidationTagMismatch (getField @"isValid" tx) PassedUnexpectedly
-        Fails _sss -> pure ()
+        Passes _ -> False ?!## ValidationTagMismatch (getField @"isValid" tx) PassedUnexpectedly
+        Fails ps fs ->
+          tellEvent (SuccessfulPlutusScriptsEvent ps)
+            >> tellEvent (FailedPlutusScriptsEvent (scriptFailuresToPlutusDebug fs))
```
lehins: Same here:
```suggestion
        Fails ps fs -> do
          tellEvent (SuccessfulPlutusScriptsEvent ps)
          tellEvent (FailedPlutusScriptsEvent (scriptFailuresToPlutusDebug fs))
```

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Rules/Utxos.hs
```diff
@@ -250,9 +261,36 @@ invalidEnd = intercalate "," ["[LEDGER][SCRIPTS_NOT_VALIDATE_TRANSITION]", "END"
...
+  fromCBOR = decode (Summands "FailureDescription" dec)
+    where
+      dec 0 = SumD OnePhaseFailure <! From
+      dec 1 = SumD PlutusFailure <! From <! From
+      dec n = Invalid n
+
+scriptFailureToFailureDescription :: ScriptFailure -> FailureDescription
+scriptFailureToFailureDescription (OnePhaseSF t) = OnePhaseFailure t
+scriptFailureToFailureDescription (PlutusSF t pd) =
+  PlutusFailure t (B64.encode . serialize' $ pd)
```
lehins:
```suggestion
  PlutusFailure t (B64.encode $ serialize' pd)
```

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Rules/Utxos.hs
```diff
@@ -122,8 +129,10 @@ instance
   type Event (UTXOS era) = UtxosEvent era
   transitionRules = [utxosTransition]
 
-newtype UtxosEvent era
-  = UpdateEvent (Event (Core.EraRule "PPUP" era))
+data UtxosEvent era
+  = AlonzoPpupToUtxosEvent (Event (Core.EraRule "PPUP" era))
+  | SuccessfulPlutusScriptsEvent [PlutusDebug]
+  | FailedPlutusScriptsEvent [PlutusDebug]
```
lehins: I'd suggest using `NonEmpty` here as well
```suggestion
  | FailedPlutusScriptsEvent (NonEmpty PlutusDebug)
```

This will be useful for downstream consumers of this event.

lehins (approved):
Looks great. One suggestions for `FailedPlutusScriptsEvent` I think is worth considering. Others feel free to ignore, since they are just a matter of taste.

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Rules/Utxos.hs
JaredCorduan: I keep doing that, no clue why :)

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Rules/Utxos.hs
JaredCorduan: I actually tried to use a `NonEmpty` list here, but ran into an unfortunate lack of "parse, don't validate". I'd love to fix it, I'm not sure how easy it will be. Maybe a follow up PR?

The function `collectTwoPhaseScriptInputs` only returns Plutus scripts (together will the collected data that it needs to run them) in the happy path, but unfortunately the returns is `AlonzoScript.Script`, thus forgetting the fact that only Plutus scripts are present.  When `evalScripts` get the results from `collectTwoPhaseScriptInputs`, it has to do something with the case of Timelocks, so it creates the possibility of a 1-phase failure, namely the constructor `OnePhaseSF Text` of `ScriptFailure`.

This leads to having to filter the non-empty list of `ScriptFailure`s to just those relating to plutus. See the function `scriptFailuresToPlutusDebug` introduced in this PR. And this is why I could not easily use a non-empty list inside `FailedPlutusScriptsEvent`.

This is a silly situation, of course, and I do want to fix it.

JaredCorduan: it's worth noting that timelocks (aka "1-phase scripts") are validated in Alonzo in:

https://github.com/input-output-hk/cardano-ledger/blob/70dbfc9f12283666e29f8492aff5da9b1c8c5a7f/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Rules/Utxow.hs#L361

This check uses the `validateScript` class method of `ValidateScript`. One of the first bits of tech debt that I would like to address post Babbage is an overhaul in the handling of 1-phase vs 2-phase scripts. The `validateScript` methods look generic, as though it runs all scripts, but does not.

lehins: It is silly indeed. I don't think it make sense to have a possibility of emitting a `FailedPlutusScriptsEvent` without any "Plutus" failure. Correct me if I am wrong, but I think there is an easy solution to this, would this work?
https://github.com/input-output-hk/cardano-ledger/pull/2724

JaredCorduan: Yep, that's great. I'll use it if I can't quickly come up with a decent solution to the root of the problem.

JaredCorduan: I've just pushed another commit which addresses the root of the problem, namely the unneeded checking of timelocks inside collectTwoPhaseInputs.

lehins: Nice!!!! :100: 
I think this is one of those examples where a stricter type encourages a better implementation

JaredCorduan: yep, absolutely. I've gotten so much mileage out of [Parse, don’t validate](https://lexi-lambda.github.io/blog/2019/11/05/parse-don-t-validate/).

PR #2671, opened 2022-02-25
Make sure `getDatum` returns corect type. Rename `isAdaOnly`

lehins:


TimSheard (approved):
Looks good to me.  Why did we rename adaOnly, is it just to use a consistent naming scheme for predicates? This is going to break Jared's PR where that function is used. But it should be easy to fix. Be sure all the tests pass, they are still running as I type.

lehins:
Renaming `adaOnly` -> `isAdaOnly` was done to convey the intent of the function a bit better, i.e. make it consistent with predicates.

Renaming was done in the recently merged PR #2668 that's why CI didn't fail for the PR that introduced `getDatum` #2664

JaredCorduan (approved):

PR #2672, opened 2022-02-28
Add missing HasField "collateral" instances for Shelley, Allegra and Mary

Jimbo4350:


JaredCorduan:
sorry @Jimbo4350 , what you have added is totally consistent with what we have done in the past, but we now think that adding these spurious accessors was a mistake and we would like to eventually remove them.

lehins:
Yeah, we try to avoid adding concepts from newer eras into older ones

PR #2673, opened 2022-02-28
Always generate a TotalRewardEvent even if empty

erikd:


TimSheard (approved):
This seems like a reasonable request. One wouldn't want to write incrmental changes (since most of those are empty) but the final epoch total should be fine. I believe there is a test that compares the sum of all the incremental events with the final event (assumes zero if no final epoch event arrives) So with the change even the tests should pass. If the don't et me know.

PR #2675, opened 2022-03-02
Improve SplitMap deserialization. Make interning stricter

lehins:
Current deserialization for Map is a bit dumb: first it deserializes into a list, which is then reversed and only after that it is converted to a `Map`. It makes sense only for preservation of semantics of deserialization in presence of duplicate keys. For UTxO we know for sure there will never be any duplicate TxIns, which means that we can construct a Map as it is being deserialized thus avoiding intermediate list allocations, which saves memory and about 20% in deserialization time.

This PR also reduces memory overhead during deserialization by making internin a little stricter


JaredCorduan (approved):
LGTM!

PR #2676, opened 2022-03-02
Use the correct pulseSize for rewards

kderme:
On mainnet this causes a very small number of rewards to be computed for most blocks (less than 10) and most of them to be computed after the 8k/f slot on a single block.


JaredCorduan (approved):
fantastic catch @kderme ! there are `(10k/f)`-many _slots_ per epoch, but the odds of a block happening in a given slot is `f`. And of course the events happen once per block, not once per slot.

PR #2677, opened 2022-03-03
Update nix index

lehins:
Updates all sources: `hackage.nix`, `haskell.nix`, `iohk-nix` and `ormolu`


eras/byron/ledger/impl/test/Test/Cardano/Chain/Delegation/Model.hs
```diff
@@ -74,8 +74,11 @@ newtype SignalSDELEG (v :: Type -> Type)
   = SignalSDELEG (STS.Signal SDELEG)
   deriving (Show)
 
-instance HTraversable SignalSDELEG where
-  htraverse _ s = pure (coerce s)
+instance FunctorB SignalSDELEG where
```
nc6, line 77: From `barbies`? I didn't even know we used that...

nc6 (approved):
LGTM

PR #2681, opened 2022-03-08
babbage unit tests

JaredCorduan:
This PR adds two new units tests for some Babbage era features, namely:

* using an inline datum
* using a reference script

I also fixed some problems along the way. It might be easier to read this PR commit by commit. The fixes are:

* ~I reverted the bump to the index-state, we can add it back when we resolve the problems with https://github.com/input-output-hk/cardano-ledger/pull/2677~
* I :cherries:-picked the commit from @TimSheard 's PR https://github.com/input-output-hk/cardano-ledger/pull/2680 that added the `ExtendedUTxO` class.
* I added another method to the `ExtendedUTxO` class, `txscripts`, so that we could reuse Alonzo functions and predicates.
* I made it so that `txOutDataHash` no longer returns hashed inline datums. This function is fed into the `HasField "datahash"` instance, and I found it confusing for this "accessor" to retrieve the inline datums.

### DIFF
--- a/eras/babbage/impl/src/Cardano/Ledger/Babbage/Scripts.hs
+++ b/eras/babbage/impl/src/Cardano/Ledger/Babbage/Scripts.hs
@@ -10,9 +10,6 @@
 module Cardano.Ledger.Babbage.Scripts where
 
 import Cardano.Ledger.Alonzo.Data (BinaryData, dataToBinaryData)
-import Cardano.Ledger.Alonzo.Language (Language (..))
-import Cardano.Ledger.Alonzo.PlutusScriptApi (language)
-import Cardano.Ledger.Alonzo.Scripts (Script)
 import Cardano.Ledger.Alonzo.Tx
   ( ScriptPurpose (..),
     ValidatedTx (..),
@@ -28,7 +25,7 @@ import Cardano.Ledger.Babbage.TxBody
     spendInputs',
   )
 import qualified Cardano.Ledger.Core as Core
-import Cardano.Ledger.Era (Crypto, Era, ValidateScript (hashScript, isNativeScript))
+import Cardano.Ledger.Era (Crypto, Era, ValidateScript (hashScript))
 import Cardano.Ledger.Hashes (DataHash)
 import Cardano.Ledger.Shelley.Scripts (ScriptHash (..))
 import Cardano.Ledger.Shelley.UTxO (UTxO (..))
@@ -72,7 +69,7 @@ getDatum tx (UTxO m) sp = do
 {- txscripts tx utxo = txwitscripts tx ∪ {hash s ↦ s | ( , , , s) ∈ utxo (spendInputs tx ∪ refInputs tx)} -}
 
 -- | Compute a Map of all the Scripts in a ValidatedTx.
-txscripts ::
+babbageTxScripts ::
   forall era.
   ( Core.TxBody era ~ TxBody era,
     Core.TxOut era ~ TxOut era,
@@ -81,7 +78,7 @@ txscripts ::
   UTxO era ->
   Core.Tx era ->
   Map.Map (ScriptHash (Crypto era)) (Core.Script era)
-txscripts (UTxO mp) tx = SplitMap.foldl' accum (getField @"scriptWits" tx) smallUtxo
+babbageTxScripts (UTxO mp) tx = SplitMap.foldl' accum (getField @"scriptWits" tx) smallUtxo
   where
     txbody = getField @"body" tx
     scriptinputs = referenceInputs' txbody `Set.union` spendInputs' txbody
@@ -90,24 +87,6 @@ txscripts (UTxO mp) tx = SplitMap.foldl' accum (getField @"scriptWits" tx) small
       case txout of
         (TxOut _ _ _ SNothing) -> ans
         (TxOut _ _ _ (SJust script)) -> Map.insert (hashScript @era script) script ans
-
-languages ::
-  forall era.
-  ( ValidateScript era,
-    Core.TxOut era ~ TxOut era,
-    Core.TxBody era ~ TxBody era,
-    Core.Script era ~ Script era
-  ) =>
-  UTxO era ->
-  Core.Tx era ->
-  Set Language
-languages utxo tx =
-  Set.fromList
-    [ lang
-      | (_, script) <- Map.toList (txscripts utxo tx),
-        (not . isNativeScript @era) script,
-        Just lang <- [language @era script]
-    ]
 
 -- Compute two sets for all TwoPhase scripts in a Tx.
 -- set 1) DataHashes for each Two phase Script in a TxIn that has a DataHash
--- a/eras/babbage/impl/src/Cardano/Ledger/Babbage/Scripts.hs
+++ b/eras/babbage/impl/src/Cardano/Ledger/Babbage/Scripts.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE BangPatterns #-}
 {-# LANGUAGE DataKinds #-}
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE RankNTypes #-}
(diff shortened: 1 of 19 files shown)


eras/babbage/impl/src/Cardano/Ledger/Babbage/Scripts.hs
```diff
@@ -67,39 +69,60 @@ getDatum tx (UTxO m) sp = do
...
+            else ans
+        (TxOut addr _ (DatumHash dhash) _) ->
+          if isTwoPhaseScriptAddressFromMap @era hashScriptMap addr
+            then (Set.insert dhash hashSet, inputSet)
+            else ans
+        (TxOut addr _ (Datum _) _) ->
+          if isTwoPhaseScriptAddressFromMap @era hashScriptMap addr
+            then ans -- An a TwoPhaseScript with Explict Datum and does not need a DataHash
+            else (hashSet, Set.insert txin inputSet)
+
```
TimSheard: What did we decide here?  Is a phase 1 script allowed to have an explicit datum.  If that is so then it should be
`   (TxOut addr _ (Datum _) _) -> ans`

As both of the branches of the if, need to do nothing.

eras/babbage/impl/src/Cardano/Ledger/Babbage/TxBody.hs
```diff
@@ -981,12 +981,12 @@ txOutDataHash :: Era era => TxOut era -> Maybe (DataHash (Crypto era))
...
-  TxOutCompactDatum _ _ binaryData -> Just $! hashBinaryData binaryData
+  TxOutCompactDatum _ _ _ -> Nothing
   TxOutCompactRefScript _ _ datum _ ->
     case datum of
       NoDatum -> Nothing
       DatumHash dh -> Just dh
-      Datum binaryData -> Just $! hashBinaryData binaryData
+      Datum _ -> Nothing
   TxOut_AddrHash28_AdaOnly {} -> Nothing
   TxOut_AddrHash28_AdaOnly_DataHash32 _ _ _ dataHash32 -> decodeDataHash32 dataHash32
```
TimSheard, line 995: We might want to add a comment to this function, that we are looking for actual DataHashes, not evidence of Data use. If we wanted evidence of data use we would do something on the BinaryData fields. But as we can see here that was a mistake. So lets make it clear what are legitimate uses of this function.

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Examples/BabbageFeatures.hs
```diff
@@ -0,0 +1,303 @@
...
+-- =======================
+-- Setup the initial state
+-- =======================
+
+scriptAddr :: forall era. (Scriptic era) => Core.Script era -> Proof era -> Addr (Crypto era)
+scriptAddr s _pf = Addr Testnet pCred sCred
+  where
+    pCred = ScriptHashObj . hashScript @era $ s
+    (_ssk, svk) = mkKeyPair @(Crypto era) (RawSeed 0 0 0 0 0)
+    sCred = StakeRefBase . KeyHashObj . hashKey $ svk
```
TimSheard: I think it is good practice to put the (Proof era) as the first argument.

TimSheard (approved):
Just a few small changes. Looks good to me.

eras/babbage/impl/src/Cardano/Ledger/Babbage/TxBody.hs
JaredCorduan: great idea!

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Examples/BabbageFeatures.hs
JaredCorduan: yes, I agree, thanks for catching that!

eras/babbage/impl/src/Cardano/Ledger/Babbage/Scripts.hs
JaredCorduan: ah yes, good point. I think we wish to allow a phase 1 script allowed to have an explicit datum. I will change it and make a ticket/todo to confirm the choice.

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Examples/BabbageFeatures.hs
```diff
@@ -0,0 +1,303 @@
...
+        ++ map (\i -> (mkGenesisTxIn i, somePlainOutput pf)) [3 .. 8]
+        ++ map (\i -> (mkGenesisTxIn i, collateralOutput pf)) [11 .. 18]
+
+defaultPPs :: [PParamsField era]
+defaultPPs =
+  [ Costmdls $ Map.fromList [(PlutusV1, freeCostModel), (PlutusV2, freeCostModel)],
+    MaxValSize 1000000000,
+    MaxTxExUnits $ ExUnits 1000000 1000000,
+    MaxBlockExUnits $ ExUnits 1000000 1000000,
+    ProtocolVersion $ ProtVer 6 0,
```
JaredCorduan:
```suggestion
    ProtocolVersion $ ProtVer 7 0,
```

cabal.project
```diff
@@ -1,4 +1,4 @@
-index-state: 2022-02-18T00:00:00Z
+index-state: 2021-12-03T06:07:23Z
```
lehins: Make sure to rebase on master, cause that change is no longer correct.

eras/babbage/impl/src/Cardano/Ledger/Babbage/Scripts.hs
```diff
@@ -67,39 +69,60 @@ getDatum tx (UTxO m) sp = do
...
+  (Set (DataHash (Crypto era)), Set (TxIn (Crypto era)))
+babbageInputDataHashes hashScriptMap tx (UTxO mp) =
+  SplitMap.foldlWithKey' accum (Set.empty, Set.empty) smallUtxo
+  where
+    txbody = body tx
+    spendinputs = getField @"inputs" txbody :: (Set (TxIn (Crypto era)))
+    smallUtxo = spendinputs SplitMap.◁ mp
+    accum ans@(hashSet, inputSet) txin txout =
+      case txout of
+        (TxOut addr _ NoDatum _) ->
```
lehins: Redundant parenthesis. Pattern matching in case statements doesn't need those. This comment applies to all pattern matches below too 
```suggestion
        TxOut addr _ NoDatum _ ->
```

eras/babbage/impl/src/Cardano/Ledger/Babbage/Scripts.hs
```diff
@@ -67,39 +69,60 @@ getDatum tx (UTxO m) sp = do
...
-        Just lang <- [language @era script]
-    ]
+  (Set (DataHash (Crypto era)), Set (TxIn (Crypto era)))
+babbageInputDataHashes hashScriptMap tx (UTxO mp) =
+  SplitMap.foldlWithKey' accum (Set.empty, Set.empty) smallUtxo
+  where
+    txbody = body tx
+    spendinputs = getField @"inputs" txbody :: (Set (TxIn (Crypto era)))
+    smallUtxo = spendinputs SplitMap.◁ mp
+    accum ans@(hashSet, inputSet) txin txout =
```
lehins: Tuples are lazy, in order to avoid memory leak we need to force its elements:
```suggestion
    accum ans@(!hashSet, !inputSet) txin txout =
```

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Examples/TwoPhaseValidation.hs
```diff
@@ -1955,25 +1966,38 @@ testUTXOWsubset,
...
     Either [(PredicateFailure (Core.EraRule "UTXOW" era))] (State (Core.EraRule "UTXOW" era)) ->
     Assertion
 
 -- | Use an equality test on the expected and computed [PredicateFailure]
-testUTXOW wit@(UTXOW (Alonzo _)) = testUTXOWwith wit genericCont
-testUTXOW wit@(UTXOW (Babbage _)) = testUTXOWwith wit genericCont
-testUTXOW (UTXOW other) = error ("Cannot use testUTXOW in era " ++ show other)
+testUTXOW wit@(UTXOW (Alonzo _)) utxo = testUTXOWwith wit genericCont utxo
+testUTXOW wit@(UTXOW (Babbage _)) utxo = testUTXOWwith wit genericCont utxo
+testUTXOW (UTXOW other) _ = error ("Cannot use testUTXOW in era " ++ show other)
```
lehins: Why did you eta expanded it? Seems unnecessary.

lehins (approved):
Looks great. Left a few suggestions

cabal.project
lehins: > I reverted the bump to the index-state, we can add it back when we resolve the problems with https://github.com/input-output-hk/cardano-ledger/pull/2677

That PR has been merged, so an alternative would be to just remove the commit [ca11870](https://github.com/input-output-hk/cardano-ledger/pull/2681/commits/ca11870d351c63105a9b4ccac81242fba03a3feb)

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Examples/TwoPhaseValidation.hs
JaredCorduan: good point! (I assume you mean we can omit the `utxo`)

JaredCorduan: I have no idea why I did that, must be vestigial, good catch!

PR #2687, opened 2022-03-10
Update ormolu in nix sources to 0.4.0.0

lehins:
I think it is better to use a version tag, rather than just a git sha. After all sha256 will ensure that underlying archive doesn't change.


JaredCorduan (approved):
thanks! :raised_hands:

PR #2689, opened 2022-03-14
add Babbage CDDL tests

goolord:
this PR contains 

- bug fixes for some `From/ToCBOR` instances
- uses nested cbor functions from https://input-output-hk.github.io/ouroboros-network/cardano-binary/Cardano-Binary-Serialize.html#g:1 which use the proper `encoded-cbor` tag https://datatracker.ietf.org/doc/html/rfc8610#appendix-D for cbor-in-cbor
- changes to the babbage cddl that let us use coders and help make cbor-in-cbor types more well defined
- more details in the serialization tests' error messages that were helpful to me while debugging failing tests


JaredCorduan (approved):
Looks great!

PR #2690, opened 2022-03-15
Revert `TotalRewardEvent` changes

kderme:
This reverts changes to the `TotalRewardEvent` done in https://github.com/input-output-hk/cardano-ledger/pull/2615/files#diff-e38c0cafe1544253b0aa549f1fdaeb9b824089696fa5320f0c1dd5b3fa7d7f58R159-R161 and fixes the related test.

### DIFF
--- a/eras/shelley/test-suite/test/Test/Cardano/Ledger/Shelley/Rewards.hs
+++ b/eras/shelley/test-suite/test/Test/Cardano/Ledger/Shelley/Rewards.hs
@@ -104,7 +104,7 @@ import Cardano.Ledger.Shelley.Rewards
     mkApparentPerformance,
     mkPoolRewardInfo,
   )
-import Cardano.Ledger.Shelley.Rules.NewEpoch (NewEpochEvent (TotalRewardEvent))
+import Cardano.Ledger.Shelley.Rules.NewEpoch (NewEpochEvent (DeltaRewardEvent, TotalRewardEvent))
 import Cardano.Ledger.Shelley.Rules.Rupd (RupdEvent (..))
 import qualified Cardano.Ledger.Shelley.Rules.Tick as Tick
 import Cardano.Ledger.Shelley.TxBody (PoolParams (..), RewardAcnt (..))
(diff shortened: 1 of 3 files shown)


JaredCorduan (approved):
looks great, thank you very much @kderme !

eras/shelley/test-suite/test/Test/Cardano/Ledger/Shelley/Rewards.hs
```diff
@@ -754,6 +754,7 @@ aggDeltaRewardEvents :: [ChainEvent C] -> Map (Credential 'Staking (Crypto C)) (
 aggDeltaRewardEvents events = foldl' accum Map.empty events
   where
     accum ans (TickEvent (Tick.RupdEvent (RupdEvent _ m))) = Map.unionWith Set.union m ans
+    accum ans (TickEvent (Tick.NewEpochEvent (DeltaRewardEvent (RupdEvent _ m)))) = Map.unionWith Set.union m ans
```
kderme, line 758: Do I understand this event correctly? UNfortunately there is no description about it.

kderme: Refering to my commit message 
```
This is an issue that the generators are not strong enough to discover. The DeltaRewardEvent appears in rare 
cases when there are no blocks after the 80% of an epoch. For these cases it has to be taken into account while
 aggregating incremental rewards
```

JaredCorduan: We definitely want to redo the names and structure of the ledger events, they are a mess. And at the same time produce a document.

The `Tick.NewEpochEvent (DeltaRewardEvent (RupdEvent _ m)))` should be rare, it's an edge case for when the timing of the reward pulser does not complete in time. The other event, `Tick.RupdEvent (RupdEvent _ m))` is the usual one for the pulsing rewards. And indeed it makes sense that you would see that if no blocks happened in the last fifth of the epoch. What you have written looks correct to me.

JaredCorduan:
I am tempted to merge this PR even though one of the non-required CI jobs is yellow. It looks to me like the job actually did successfully complete.

PR #2692, opened 2022-03-15
update Plutus

JaredCorduan:
* The cost model now contains the EvaluationContext (from Plutus)
* There is no more default cost model
* The plutus check that scripts are well-formed now takes a protocol version
* The collection of cost models is now wrapped in a newtype, in order to control the JSON and CBOR (de)serialization.
* The Plutus evaluator now takes the protocol version


eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Scripts.hs
```diff
@@ -189,36 +212,40 @@ instance SafeToHash CostModel where
...
 instance HashWithCrypto CostModel CostModel
 
-instance NoThunks CostModel
+deriving via InspectHeapNamed "PV1.EvaluationContext" PV1.EvaluationContext instance NoThunks PV1.EvaluationContext
+
+deriving via InspectHeapNamed "CostModel" CostModel instance NoThunks CostModel
 
-instance NFData CostModel
+instance NFData PV1.EvaluationContext where
+  rnf = undefined
```
JaredCorduan: I think we will need Plutus to provide this instance.

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/PlutusScriptApi.hs
```diff
@@ -198,7 +198,7 @@ collectTwoPhaseScriptInputs ei sysS pp tx utxo =
       case txinfo lang of
         Right inf -> Right (script, getData tx utxo sp ++ (d : [valContext inf sp]), eu, costs Map.! lang)
         Left te -> Left $ BadTranslation te
-    apply _ (_, _, eu) script = Right (script, [], eu, CostModel mempty)
+    apply _ (_, _, eu) script = Right (script, [], eu, CostModelV1 mempty undefined)
```
JaredCorduan: It is no longer easy to make a `mempty` like cost model, since `mkEvaluationContext` can return Nothing. We'll need a plan for this...

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Scripts.hs
JaredCorduan: I'll update to the tip of `release/1.0.0` (in Plutus) once https://github.com/input-output-hk/plutus/pull/4471 lands.

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/PlutusScriptApi.hs
JaredCorduan: this branch is entirely unneeded with a bit of rearrangement.

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Examples/TwoPhaseValidation.hs
```diff
@@ -2171,11 +2177,7 @@ alonzoUTXOWexamplesB pf =
               pf
               (trustMeP pf True $ missing1phaseScriptWitnessTx pf)
               ( Left
-                  [ fromUtxos @era . CollectErrors $
-                      [ NoRedeemer (Spending (mkGenesisTxIn 100)),
-                        NoWitness (timelockHash 0 pf)
-                      ],
```
JaredCorduan, line 2177: I'm really happy that the cleanup that I did to `collectTwoPhaseScriptInputs` also cleaned up these cascading predicate failures. This was "parse, don't validate" at its best!

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Examples/TwoPhaseValidation.hs
```diff
@@ -2184,12 +2186,7 @@ alonzoUTXOWexamplesB pf =
               pf
               (trustMeP pf True $ missing2phaseScriptWitnessTx pf)
               ( Left
-                  [ fromUtxos @era . CollectErrors $
-                      [ NoWitness (alwaysSucceedsHash 2 pf),
-                        NoWitness (alwaysSucceedsHash 2 pf),
-                        NoWitness (alwaysSucceedsHash 2 pf)
-                      ],
```
JaredCorduan, line 2191: also from the `collectTwoPhaseScriptInputs` cleanup

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Genesis.hs
```diff
@@ -34,20 +34,22 @@ import qualified Data.Aeson as Aeson
...
-import Plutus.V1.Ledger.Api (defaultCostModelParams)
+import Plutus.V1.Ledger.Api as PV1 hiding (Map, Script, TxOut, Value)
+import Plutus.V2.Ledger.Api as PV2 hiding (Map, Script, TxOut, Value)
 import Prelude
 
 data AlonzoGenesis = AlonzoGenesis
   { coinsPerUTxOWord :: !Coin,
-    costmdls :: !(Map Language CostModel),
+    costmdls :: !CostModels,
     prices :: !Prices,
```
TimSheard, line 53: Please explicitly import CostModel and CostModels from somewhere. They come from some blanket import. and that is too bad.  Maybe they come from Aonzo.Scripts(Costmodel,Costmodels)

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Scripts.hs
```diff
@@ -189,36 +212,37 @@ instance SafeToHash CostModel where
...
 
-instance NFData CostModel
+deriving via InspectHeapNamed "CostModel" CostModel instance NoThunks CostModel
+
+instance NFData CostModel where
+  rnf (CostModelV1 cm ectx) = seq (rnf cm) (rnf ectx)
+  rnf (CostModelV2 cm ectx) = seq (rnf cm) (rnf ectx)
 
 checkCostModel :: Language -> Map Text Integer -> Either String CostModel
 checkCostModel PlutusV1 cm =
```
TimSheard: Add a comment about what this function does.  Is it something like parsing it to see that it is not bottom, and that the parser actually succeeds, so we know that it is well formed?

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Scripts.hs
```diff
@@ -244,6 +268,20 @@ hashCostModel ::
   SafeHash (Crypto e) CostModel
 hashCostModel _proxy = hashWithCrypto (Proxy @(Crypto e))
 
+getEvaluationContext :: CostModel -> PV1.EvaluationContext
+getEvaluationContext (CostModelV1 _ ec) = ec
+getEvaluationContext (CostModelV2 _ ec) = ec
+
+newtype CostModels = CostModels {unCostModels :: Map Language CostModel}
+  deriving (Eq, Show, Ord, Generic, NFData, NoThunks)
+  deriving newtype (Monoid, Semigroup)
```
TimSheard: Do we actually use the Semigroup instance. Any Semigroup instance based on Map is suspect as (x <> y) is not always the same as (y <> x).

eras/alonzo/test-suite/test/Test/Cardano/Ledger/Alonzo/Examples.hs
```diff
@@ -187,12 +189,20 @@ alonzo = Proxy
...
+           PlutusV1
+           testingCostModelV1
+           bytes
+           (ExUnits 100000000 10000000)
+           ds
+       ) of
+    (ShouldSucceed, Passes) -> assertBool "" True
+    (ShouldSucceed, Fails xs) -> assertBool (show xs) (trace (show (head xs)) False)
+    (ShouldFail, Passes) -> assertBool ("Test that should fail, passes: " ++ show script) False
+    (ShouldFail, Fails _) -> assertBool "" True
```
TimSheard: Do we really want this trace here?

TimSheard (approved):
Who would have thought changing 1 type, would cascade to so many files. 
Any pending PRs when rebased on this are going to have a lot of changes.
Otherwise, other than the few comments, it looks good.

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Genesis.hs
JaredCorduan: I totally agree, open imports can be confusing!

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Scripts.hs
JaredCorduan: great point, I will try to remove it!

eras/alonzo/test-suite/test/Test/Cardano/Ledger/Alonzo/Examples.hs
JaredCorduan: indeed not, great catch

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Scripts.hs
JaredCorduan: great point, in fact the name is no longer great

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Scripts.hs
JaredCorduan: I was able to remove it

PR #2694, opened 2022-03-16
Babbage examples for Consensus

### LINKED ISSUE #2691: Babbage Consensus Examples
Consensus uses examples from the ledger for testing. In particular, the type `ShelleyLedgerExamples` captures all the examples that are needed for each era. We need Babbage examples.

For example, the Alonzo examples are here: https://github.com/input-output-hk/cardano-ledger/blob/73f9c089981e587a0b2698ef6ad65d75fb41a85f/eras/alonzo/test-suite/src/Test/Cardano/Ledger/Alonzo/Examples/Consensus.hs

JaredCorduan:
This add the requisite Babbage examples needed by consensus. I've used all the new features: inline datums, collateral return, reference inputs, and reference scripts.

I also change the Babbage `TranslationContext` to unit, which is all I think we need.

closes #2691

### DIFF
--- a/eras/babbage/test-suite/src/Test/Cardano/Ledger/Babbage/Examples/Consensus.hs
+++ b/eras/babbage/test-suite/src/Test/Cardano/Ledger/Babbage/Examples/Consensus.hs
@@ -1,0 +1,183 @@
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeApplications #-}
+
+module Test.Cardano.Ledger.Babbage.Examples.Consensus where
+
+import Cardano.Ledger.Alonzo.Data (AuxiliaryData (..), AuxiliaryDataHash (..), Data (..), dataToBinaryData, hashData)
+import Cardano.Ledger.Alonzo.Language (Language (..))
+import Cardano.Ledger.Alonzo.Scripts (ExUnits (..), Script (..))
+import qualified Cardano.Ledger.Alonzo.Scripts as Tag (Tag (..))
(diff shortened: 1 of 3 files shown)


eras/babbage/test-suite/src/Test/Cardano/Ledger/Babbage/Examples/Consensus.hs
```diff
@@ -0,0 +1,183 @@
...
+import Cardano.Slotting.Slot (EpochNo (..), SlotNo (..))
+import Data.Default.Class (def)
+import qualified Data.Map.Strict as Map
+import Data.Proxy (Proxy (..))
+import qualified Data.Sequence.Strict as StrictSeq
+import qualified Data.Set as Set
+import qualified PlutusTx as Plutus
+import Test.Cardano.Ledger.Alonzo.Scripts (alwaysFails, alwaysSucceeds)
+import qualified Test.Cardano.Ledger.Mary.Examples.Consensus as SLE
+import qualified Test.Cardano.Ledger.Shelley.Examples.Consensus as SLE
```
TimSheard, line 52: Might is be better to use two different qualified names, like  ShelleySLE and MarySLE then it is clear what parts of the Babbage examples come from what earlier examples.

TimSheard (approved):
Looks clear to me.

eras/babbage/test-suite/src/Test/Cardano/Ledger/Babbage/Examples/Consensus.hs
JaredCorduan: good idea!

PR #2697, opened 2022-03-22
Stuff

newhoggy:


JaredCorduan:
@newhoggy are you done with this experiment?

lehins:
I am guessing from the name of the branch that intent was to fix GithubActions. Considering that it is not currently broken and we don't even have GA running for Mac, I suspect that branch was intended for cardano-node.

Closing. Feel free to reopen if I was wrong about all theses assumptions

PR #2699, opened 2022-03-24
Ledger benchmarks and fix for regression related to static checks

lehins:


JaredCorduan (approved):
LGTM, thanks @lehins !

nc6:
What would it take to have this benchmark included in our standard benchmark run?

lehins:
> What would it take to have this benchmark included in our standard benchmark run?

@nc6 All we need is a place where we can easily upload 1.5gb ledger state for downloading on CI.

Issue #2471, opened 2021-09-15
Babbage unit tests


JaredCorduan:
Write units tests, examples, and golden tests for all the new Babbage features. Note that the Alonzo unit tests / examples are already running in the Babbage era, so this list only includes tests that only make sense in the Babbage era.

# Valid :heavy_check_mark: 

- [x] spend a plutus utxo which contains an inline datum
- [x] spend a plutus utxo which contains a reference script
- [x] spend a plutus utxo which contains a reference script and an inline datum
- [x] reference inputs are not consumed
- [x] collateral outputs work. (if you submit a failing plutus script and mark the tx as IsValid=false, the over-payment goes to the collateral output)
- [x] a tx with a plutus script which is both needed and supplied in two ways: 1) via reference script, 2) in the script witnesses
- [x] using a reference input whose output contains a data hash, but the datum **is not** in the witness set.
- [x] using a reference input whose output contains a data hash, and the datum **is** in the witness set.
- [x] using a reference script in order to authorize a plutus locked delegation cert / minting policy / withdrawal

# Invalid :no_good:

- [x] collateral total must be correct
- [x] a tx with an inline datum and the same datum in the witness set
- [x] a tx with a script/datum which is not well-formed
- [x] a tx with a plutus script and and byron outputs
- [x] a tx with a plutus V1 script and reference inputs (or inline datums).

PR #2700, opened 2022-03-24
more babbage unit tests, two babbage bug fixes

JaredCorduan:
This PR fixes two bugs in the Babbage rules (each fix is in its own commit):

* In the case of expected failing Plutus scripts, collateral outputs must be a part of the incremental stake distribution computation.
* The collection of tests named `missingRequiredDatums` now uses a new class method in `ExtendedUTxO` to determine which data witnesses are allowed which are not strictly needed. In the Babbage era, data witnesses are allowed for reference inputs. Additionally, `danglingWitnessDataHashes` was removed since it is redundant with `missingRequiredDatums`.

This PR adds unit tests for most of the remaining ones in  #2471.

### DIFF
--- a/eras/alonzo/impl/src/Cardano/Ledger/Alonzo.hs
+++ b/eras/alonzo/impl/src/Cardano/Ledger/Alonzo.hs
@@ -89,6 +89,7 @@ import Control.Monad.Except (liftEither)
 import Control.Monad.Reader (runReader)
 import Control.State.Transition.Extended (TRC (TRC))
 import Data.Default (def)
+import Data.Foldable (toList)
 import qualified Data.Map.Strict as Map
 import Data.Maybe.Strict
 import qualified Data.Set as Set
@@ -225,6 +226,12 @@ instance CC.Crypto c => ExtendedUTxO (AlonzoEra c) where
   txInfo = alonzoTxInfo
   inputDataHashes = alonzoInputHashes
   txscripts _ = txscripts' . getField @"wits"
+  getAllowedSupplimentalDataHashes txbody _ =
+    Set.fromList
+      [ dh
+        | out <- toList (getField @"outputs" txbody),
+          SJust dh <- [getField @"datahash" out]
+      ]
 
 -------------------------------------------------------------------------------
 -- Era Mapping
--- a/eras/babbage/impl/src/Cardano/Ledger/Babbage/Rules/Utxow.hs
+++ b/eras/babbage/impl/src/Cardano/Ledger/Babbage/Rules/Utxow.hs
@@ -22,9 +22,9 @@ import Cardano.Ledger.Alonzo.Rules.Utxow
     witsVKeyNeeded,
   )
 import Cardano.Ledger.Alonzo.Scripts (Script)
-import Cardano.Ledger.Alonzo.Tx (ValidatedTx (..), wits)
+import Cardano.Ledger.Alonzo.Tx (ValidatedTx (..))
 import Cardano.Ledger.Alonzo.TxInfo (ExtendedUTxO (..))
-import qualified Cardano.Ledger.Alonzo.TxWitness as Alonzo (TxDats (..), TxWitness (..), txdats')
+import qualified Cardano.Ledger.Alonzo.TxWitness as Alonzo (TxDats (..))
 import Cardano.Ledger.AuxiliaryData (ValidateAuxiliaryData)
 import Cardano.Ledger.Babbage.PParams (PParams' (..))
 import Cardano.Ledger.Babbage.Rules.Utxo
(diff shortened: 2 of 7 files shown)


eras/babbage/impl/src/Cardano/Ledger/Babbage/Rules/Utxow.hs
```diff
@@ -22,22 +22,17 @@ import Cardano.Ledger.Alonzo.Rules.Utxow
     witsVKeyNeeded,
   )
 import Cardano.Ledger.Alonzo.Scripts (Script)
-import Cardano.Ledger.Alonzo.Tx (ValidatedTx (..), wits)
+import Cardano.Ledger.Alonzo.Tx (ValidatedTx (..)) -- , wits)
```
lehins: Forgotten comment
```suggestion
import Cardano.Ledger.Alonzo.Tx (ValidatedTx (..))
```

eras/babbage/impl/src/Cardano/Ledger/Babbage/Rules/Utxow.hs
```diff
@@ -22,22 +22,17 @@ import Cardano.Ledger.Alonzo.Rules.Utxow
...
 import Cardano.Ledger.Alonzo.TxInfo (ExtendedUTxO (..))
-import qualified Cardano.Ledger.Alonzo.TxWitness as Alonzo (TxDats (..), TxWitness (..), txdats')
+import qualified Cardano.Ledger.Alonzo.TxWitness as Alonzo (TxDats (..)) -- , TxWitness (..), txdats')
```
lehins: Same here
```suggestion
import qualified Cardano.Ledger.Alonzo.TxWitness as Alonzo (TxDats (..))
```

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Examples/BabbageFeatures.hs
```diff
@@ -212,22 +307,22 @@ utxoStEx1 ::
...
 
 -- =========================================================================
---  Example 2: Use a reference script
+--  Example 2: Use a reference script.
 -- =========================================================================
 
-outEx2 :: Scriptic era => Proof era -> Core.TxOut era
-outEx2 pf = newTxOut pf [Address (plainAddr pf), Amount (inject $ Coin 4995)]
+txDatsExample2 :: Era era => TxDats era
+txDatsExample2 = TxDats $ keyBy hashData $ [datumExample1]
```
lehins:
```suggestion
txDatsExample2 = TxDats $ keyBy hashData [datumExample1]
```

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Examples/BabbageFeatures.hs
```diff
@@ -292,7 +778,62 @@ genericBabbageFeatures pf =
             testU
               pf
               (trustMeP pf True $ referenceScriptTx pf)
-              (Right . utxoStEx2 $ pf)
+              (Right . utxoStEx2 $ pf),
```
lehins: Just a syntactic observation that I've seen in this codebase many times, feel free to ignore it.

When there is only one composition it takes less characters to do this: 
`(Right $ utxoStEx2 pf)` vs `(Right . utxoStEx2 $ pf)`

eras/alonzo/impl/src/Cardano/Ledger/Alonzo.hs
```diff
@@ -225,6 +226,8 @@ instance CC.Crypto c => ExtendedUTxO (AlonzoEra c) where
   txInfo = alonzoTxInfo
   inputDataHashes = alonzoInputHashes
   txscripts _ = txscripts' . getField @"wits"
+  getAllowedSupplimentalDataHashes txbody _ =
+    Set.fromList [dh | out <- toList (getField @"outputs" txbody), SJust dh <- [getField @"datahash" out]]
```
lehins: Maybe split across multiple lines, its a bit long?

lehins (approved):
Looks great. Just a couple minor syntactic suggestions.

eras/babbage/impl/src/Cardano/Ledger/Babbage/Rules/Utxow.hs
JaredCorduan: good catch, that's embarrassing :grimacing:

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Examples/BabbageFeatures.hs
JaredCorduan: :+1:

PR #2701, opened 2022-03-24
implement feesOK for babbage

goolord:


goolord:
huh, it looks like `feesOK` was defined in babbage twice and i missed the other one https://github.com/input-output-hk/cardano-ledger/blob/0ad3ccf55dd0c479d30ba2e1b33d7444a3a4282f/eras/babbage/impl/src/Cardano/Ledger/Babbage/Rules/Utxo.hs#L158 , but i think this one is closer to the spec, so i'm going to swap that one out for this one

eras/babbage/impl/src/Cardano/Ledger/Babbage/Collateral.hs
```diff
@@ -59,13 +59,21 @@ minCollateral txb pp = Coin ((fee * percent) `divideCeiling` 100)
...
+  ( Era era,
+    HasField "collateralReturn" (Core.TxBody era) (StrictMaybe (TxOut era)),
+    HasField "collateral" (Core.TxBody era) (Set (TxIn (Crypto era)))
+  ) =>
+  Core.TxBody era ->
+  UTxO era ->
+  Core.Value era
 collBalance txb (UTxO m) =
-  case collateralReturn' txb of
+  case getField @"collateralReturn" txb of
```
JaredCorduan, line 72: thanks for syncing this up much better with the spec!

JaredCorduan (approved):
looks great!

PR #2706, opened 2022-03-28
correctly label figures in Alonzo.Tx

### LINKED ISSUE #2388: Mislabelling figures in Tx.hs
# Mislabelling figures in Tx.hs

Many of the exported functions are labelled incorrectly at the top of [Tx.hs](https://github.com/input-output-hk/cardano-ledger-specs/blob/114266336dc336792ee7782ac60ca492c886fefb/alonzo/impl/src/Cardano/Ledger/Alonzo/Tx.hs).

## Example

Formal spec specifies that Script purpose comes from figure 5
image

Implementation labels `ScriptPurpose` (from figure 5 above) labelled as coming from figure 4
https://github.com/input-output-hk/cardano-ledger-specs/blob/114266336dc336792ee7782ac60ca492c886fefb/alonzo/impl/src/Cardano/Ledger/Alonzo/Tx.hs#L49-L51

This minor inconsistency has appeared multiple times in the Alonzo implementation, which could be due to figures being renamed after the implementation was written.

goolord:
resolves https://github.com/input-output-hk/cardano-ledger/issues/2388


JaredCorduan (approved):
LGTM, thanks!

PR #2707, opened 2022-03-29
Syntactic improvements as well as avoid lazy tuple for State of LEDGER

lehins:
1. Change `STS`'s `State` of `LEDGER` and `LEDGERS` to a more appropriate and a perfectly named `LedgerState` type. Reasons for doing that:

* Avoiding lazy tuples, which could lead to memory leaks when ledger state is an accumulator in a folding operation
* Perfectly matched type `LedgerState` was already available, it is a bit strange that it was not used to begin with.
* Avoid conversion between `LedgerState` and a tuple in many places

2. Also switched `MempoolState` to `LedgerState`

3. Remove useless type constraints synonyms (eg. `TransUTxOState`). Benefits:
* Instead of doing a weird parent constraint `TransUTxOState Show era` it is better to use an already existing concept: `Show (UTxOState era)` that achieves the same thing
* constraint synonyms can introduce redundant constraints, which will not be noticed by the compiler and will propagate everywhere. This was also the case for `TransUTxOState` and other `Trans*`  synonyms

4. Rename record fields for `LedgerState` and `DPState` (many more are still left using this pattern, we'll fix them in later PRs): 
* avoid underscore prefix `_`, since in Haskell its meaning is "don't care", which is wrong to use if we actually do care about those accessors.
* use the most common pattern of abbreviating the type as a prefix, which is already used in `NewEpochState` and `EpochState`: `nes*` and `es*` respectively.


eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Rules/Utxo.hs
```diff
@@ -206,16 +206,21 @@ deriving stock instance
...
+    Eq (Core.Value era),
     Eq (Core.TxOut era),
-    CC.Crypto (Crypto era),
     Eq (PredicateFailure (Core.EraRule "UTXOS" era))
   ) =>
   Eq (UtxoPredicateFailure era)
 
 instance
   ( Era era,
-    Shelley.TransUTxOState NoThunks era,
```
JaredCorduan, line 218: I believe that the `TransFoo` classes were originally added to make constraints like this more compact. I see in the PR notes that you noticed this added superfluous constraints. It's a shame we can't find some middle ground...

JaredCorduan (approved):
Thank you for this much needed clean up!

* I have no idea why `LedgerState` wasn't being used everywhere, it's much nicer now.
* Removing underscore names has been long overdue, that was an old wart from template haskell
* I have mixed feelings about the `TransFoo` classes. They were nice for making the constraints more compact, but I guess if it comes at the cost of superfluous constraints it's not worth it.

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Rules/Utxo.hs
TimSheard: The trans constraint have slowly been disappearing. I rip them out when I can. So it is good we are getting rid of them all finally.

eras/shelley/impl/src/Cardano/Ledger/Shelley/API/Validation.hs
```diff
@@ -107,14 +100,6 @@ class
...
     m (EventReturnType ep (Core.EraRule "BBODY" era) (NewEpochState era))
-  default applyBlockOpts ::
-    forall ep m.
-    (EventReturnTypeRep ep, MonadError (BlockTransitionError era) m) =>
-    ApplySTSOpts ep ->
-    Globals ->
-    NewEpochState era ->
-    Block (BHeaderView (Crypto era)) era ->
-    m (EventReturnType ep (Core.EraRule "BBODY" era) (NewEpochState era))
   applyBlockOpts opts globals state blk =
```
TimSheard, line 103: Just to make sure I understand, we are simply getting rid of the typing of the default method here?

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Ledgers.hs
```diff
@@ -128,24 +122,19 @@ ledgersTransition ::
...
-        toList txwits
-
-  pure $ LedgerState u'' dp''
+  foldM
+    ( \ !ls' (ix, tx) ->
+        trans @(Core.EraRule "LEDGER" era) $
+          TRC (LedgerEnv slot ix pp account, ls', tx)
+    )
+    ls
+    $ zip [minBound ..] $ toList txwits
```
TimSheard, line 137: some times it is better to use a named function as it brings the other arguments of foldM closer to the call

```
foldM accum ls (zip [minBound ..] $ toList txwits) where
   accum !ls' (ix,tx) = trans @(Core.EraRule "LEDGER" era) $ TRC (LedgerEnv slot ix pp account, ls', tx)
```

TimSheard (approved):
It is a brave man that touches 70 files to get rid of (,) and _
The sooner we get this merged the better.

eras/shelley/impl/src/Cardano/Ledger/Shelley/API/Validation.hs
lehins: Yeah, it's just a `default` signature removal, because it is redundant. `DefaultSignatures` are useful when default implementation has some constraints that aren't present in the actual class function signature. In these cases signatures are identical, thus redundant.

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Ledgers.hs
lehins: This is really a matter of preference. I personally don't mind lambdas at all. In any case it was a lambda before, so let's keep it as such. There are enough changes for one PR already.

> It is a brave man that touches 70 files to get rid of (,) and _

FYI, this is one of the places where a memory leak is possible because `(,)` is used instead of a type with strict elements like `LedgerState`.  Note that despite that `foldM` is strict and accumulator is pattern matched on, it is only strict to whnf, so `u'` and `dp'` in the accumulator might not get evaluated. That is one of the main reasons why I am willing to put the effort into replacing tuples with strict types. With regards to `_`, their usage in our codebase is just plain wrong. It normally indicates to the reader that binding or function is unused, which is furthers from the truth in our codebase. Also, I like compiler telling me if I forgot some unused bindings, `_` masks that.

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Rules/Utxo.hs
lehins: @JaredCorduan I don't think removing a few extra lines in a list of constraints are worth the trouble of adding a new concept (namely `Trans*`) and they are definitely not worth redundant constraints that the concept introduces. It takes almost no effort to add these constraints manually, moreover compiler tells us exactly which constraints are missing, which is why all theses changes in this PR where extremely trivial and easy for me to do. 

Also note that for example for functions that use `show` on `UTxOState`, it is enough to add `Show (UTxOState era)` instead of `TransUTxOState Show era` or listing all the constraints manually.

PR #2709, opened 2022-03-29
`cardano-data` cleanup

goolord:
- removes some usages of `fromSJust`
- move `encodeKeyedStrictMaybe` to `cardano-data`
- Introduce `ofield` and `invalidField`. I'm not totally sold on these actually, but they're both common patterns and it would be nice if the api nudged you towards writing the `SparseKeyed` fields in the way it wants you to. A really easy mistake I'm trying to prevent is `field (\x foo -> foo { fooX = x }) (From :: Decoder ('Closed d) (StrictMaybe X))` where `x` is actually omittable
- Introduce `MapE`
- Split up the definite and indefinite length branches of `getSparseBlock`, into separate functions should very mildly reduce allocations and help branch prediction


libs/cardano-data/src/Data/Coders.hs
```diff
@@ -740,9 +767,11 @@ decodeSparse ::
   Decoder s a
 decodeSparse name initial pick required = do
   lenOrIndef <- decodeMapLenOrIndef
-  (!v, used) <- getSparseBlock lenOrIndef initial pick (Set.empty) name
+  (!v, used) <- case lenOrIndef of
+    Just len -> getSparseBlock len initial pick (Set.empty) name
+    Nothing -> getSparseBlockIndef initial pick (Set.empty) name
   if all (\(key, _name) -> member key used) required
-    then pure v
+    then pure $! v
```
lehins: `v` will already be evaluated thanks to `(!v, used)`, so `$!` is redundant

lehins (commented):
I don't have enough familiarity with coders package to say anything definitely about the correctness of the PR, but at a high level it looks great. I've requested @TimSheard to review it, since he'll probably be the best judge in this area.

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxBody.hs
```diff
@@ -781,24 +773,24 @@ instance
           (D (decodeStrictSeq fromCBOR))
       bodyFields 2 = field (\x tx -> tx {_txfee = x}) From
       bodyFields 3 =
-        field
+        ofield
```
JaredCorduan, line 776: `o` for optional?

JaredCorduan, line 776: `o` for optional?

JaredCorduan (approved):
LGTM (but always happy to have another review, if Tim still wants to have a look)

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxBody.hs
goolord: 'omittable' or 'optional', ya

PR #2711, opened 2022-03-30
Define 'GetLedgerView' for Babbage.

nc6:
This instance is a little (well, a lot) suspicious, since we don't
actually have extra entropy in Babbage. We rely on the fact that this
type is only used to construct the `Praos` ledger view, which is much
simler (it looks like this:

```
data LedgerView crypto = LedgerView
  { -- | Stake distribution
    lvPoolDistr     :: SL.PoolDistr crypto,
    -- | Maximum header size
    lvMaxHeaderSize :: !Natural,
    -- | Maximum block body size
    lvMaxBodySize   :: !Natural
  }
  deriving (Show)
```
). So the `error` field will be thrown away without being evaluated.

At some point we should make this situation better, but not before the
Vasil HF.

### DIFF
--- a/libs/cardano-protocol-tpraos/src/Cardano/Protocol/TPraos/API.hs
+++ b/libs/cardano-protocol-tpraos/src/Cardano/Protocol/TPraos/API.hs
@@ -12,7 +12,6 @@
 {-# LANGUAGE StrictData #-}
 {-# LANGUAGE TypeApplications #-}
 {-# LANGUAGE TypeFamilies #-}
-{-# LANGUAGE TypeSynonymInstances #-}
 {-# LANGUAGE UndecidableInstances #-}
 {-# LANGUAGE UndecidableSuperClasses #-}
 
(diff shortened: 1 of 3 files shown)


libs/cardano-protocol-tpraos/src/Cardano/Protocol/TPraos/API.hs
```diff
@@ -186,6 +192,36 @@ instance CC.Crypto c => GetLedgerView (MaryEra c)
...
+-- because it makes it simpler to get the ledger view for Praos.
+instance CC.Crypto c => GetLedgerView (BabbageEra c) where
+  currentLedgerView
+    NewEpochState
+      { nesPd,
+        nesEs
+      } =
+      LedgerView
+        { lvD = getField @"_d" . esPp $ nesEs,
+          lvExtraEntropy = error "Extra entropy is not set in the Babbage era",
```
JaredCorduan, line 205: maybe it would be less scary to use the neutral nonce? though perhaps failing hard is better? :thinking:

JaredCorduan (approved):

libs/cardano-protocol-tpraos/src/Cardano/Protocol/TPraos/API.hs
nc6: I'm inclined to go for failing hard, though I could be argued the other way!

JaredCorduan: Let's fail hard!

```
Do not go gentle into that good night,
Old age should burn and rave at close of day;
Rage, rage against the dying of the light.
```

PR #2715, opened 2022-03-30
neededTxInsForBlock now accounts for all inputs

### LINKED ISSUE #2714: neededTxInsForBlock
The function `neededTxInsForBlock` is intended to be used by the consensus layer for the UTxO on-disk project. Specifically, consensus will use this function to determine what subset of the full UTxO is needed by the ledger for a given block.

The current version of this function has a couple problems:
* it ignores collateral inputs (relevant to Alonzo and onward)
* it ignores reference inputs (relevant to Babbage and onward)

https://github.com/input-output-hk/cardano-ledger/blob/b0944fc213deb5ca5eee805922154640fab5e18b/libs/cardano-ledger-core/src/Cardano/Ledger/Block.hs#L171

JaredCorduan:
I've add a new class method to `Era`, `getAllTxInputs`, which collects all the various inputs in the transaction body in each era.

* For Shelley through Mary this is just the spending inputs.
* Alonzo adds collateral inputs.
* Babbage adds reference inputs.

closes #2714

### DIFF
--- a/eras/alonzo/impl/src/Cardano/Ledger/Alonzo.hs
+++ b/eras/alonzo/impl/src/Cardano/Ledger/Alonzo.hs
@@ -104,6 +104,11 @@ instance CC.Crypto c => EraModule.Era (AlonzoEra c) where
   type Crypto (AlonzoEra c) = c
   getTxOutEitherAddr = getAlonzoTxOutEitherAddr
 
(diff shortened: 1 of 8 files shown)


lehins (approved):
Perfect! :100:

PR #2716, opened 2022-03-31
The total collateral field should be a Maybe Coin

JaredCorduan:
The Babbage spec uses a `Maybe Coin` for the total collateral field of the transaction body, not a `Coin`, as is in the current implementation.

Using `Nothing` instead of zero to indicate that the total collateral field is not present in the transaction body is clearer and avoids edge cases.


goolord (approved):
lgtm

lehins:
@JaredCorduan Is there (will there be) a predicate check that when `_totalCollateral` is `SNothing`, then both `_collateralInputs` and `_collateralReturn` must be `mempty`?

lehins (approved):

PR #2717, opened 2022-04-01
babbage script well-formedness check

### LINKED ISSUE #2713: Babbage isWellFormed check
There is a `isWellFormed` check in the Babbage formal spec, Figure 6: UTxO with witnesses inference rules for Tx, which is not clearly  being checked in the implementation. The situation is a bit confusing due to the fact that these checks are spread out in the ledger code base. We need to make sure that we are properly checking everything.

The current state of things seems to be:

**scripts**

* The function [isScriptWellFormed](https://github.com/input-output-hk/plutus/blob/e59ade14367261a68b156e4ec71e7f65de627957/plutus-ledger-api/src/Plutus/V1/Ledger/Api.hs#L174) checks if Plutus scripts are well-formed. This check is done implicitly by scripts which are run.
* The `isScriptWellFormed` check is wrapped in the ledger code by [validScript](https://github.com/input-output-hk/cardano-ledger/blob/b0944fc213deb5ca5eee805922154640fab5e18b/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxInfo.hs#L719).
* `validScript` is only called to validate the auxiliary data. (for example, see [here](yhttps://github.com/input-output-hk/cardano-ledger/blob/b0944fc213deb5ca5eee805922154640fab5e18b/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxInfo.hs#L719) in Alonzo).
* We are _not_ checking that script references are well-formed.
* We are _not_ checking that failed scripts (which are marked as thus in the transaction) are well-formed. This _should_ be a phase 1 failure.

**datums**

* Datums are completely guarded by the wire format, including the new inline datums in babbage.

Action items:
* check the details above
* fix the two missing checks above
* find a coherent way to document that we are doing this check. it probably needs at least a comment in the `UTXOW` rule.

goolord:
i think this small check is all that's necessary

- datums well-formedness is an invariant maintained by the wire format
- reference scripts are part of the `txscripts` set for babbage
- failedScripts are part of the `txscripts` set
- auxilliary data is checked by validateMetadata

resolves https://github.com/input-output-hk/cardano-ledger/issues/2713

### DIFF
--- a/eras/babbage/impl/src/Cardano/Ledger/Babbage/Rules/Utxow.hs
+++ b/eras/babbage/impl/src/Cardano/Ledger/Babbage/Rules/Utxow.hs
@@ -23,7 +23,7 @@ import Cardano.Ledger.Alonzo.Rules.Utxow
   )
 import Cardano.Ledger.Alonzo.Scripts (Script)
 import Cardano.Ledger.Alonzo.Tx (ValidatedTx (..))
-import Cardano.Ledger.Alonzo.TxInfo (ExtendedUTxO (..))
+import Cardano.Ledger.Alonzo.TxInfo (ExtendedUTxO (..), validScript)
 import qualified Cardano.Ledger.Alonzo.TxWitness as Alonzo (TxDats (..))
 import Cardano.Ledger.AuxiliaryData (ValidateAuxiliaryData)
 import Cardano.Ledger.Babbage.PParams (PParams' (..))
@@ -120,7 +120,7 @@ validateFailedBabbageScripts ::
   Core.Tx era ->
   UTxO era ->
   Test (Shelley.UtxowPredicateFailure era)
-validateFailedBabbageScripts tx utxo = do
+validateFailedBabbageScripts tx utxo =
   let failedScripts =
         Map.filterWithKey
           ( \hs script ->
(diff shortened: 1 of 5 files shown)


eras/babbage/impl/src/Cardano/Ledger/Babbage/Rules/Utxow.hs
```diff
@@ -130,9 +130,29 @@ validateFailedBabbageScripts tx utxo = do
                   answer = one && (two || three)
                in answer
           )
-          (txscripts utxo tx)
-  failureUnless (Map.null failedScripts) $
-    Shelley.ScriptWitnessNotValidatingUTXOW (Map.keysSet failedScripts)
+          (txscripts utxo tx) -- txscripts = babbageTxScripts which includes reference scripts.
+   in failureUnless
+        (Map.null failedScripts)
+        (Shelley.ScriptWitnessNotValidatingUTXOW $ Map.keysSet failedScripts)
```
goolord, line 136: is this still an accurate name? this test also validates reference scripts

JaredCorduan: good point. I also think that the `Not` is weird in that name.

JaredCorduan (approved):
This looks great, thanks @goolord ! Thanks for adding that test! So this PR closes both #2713 and #2471

I'm not sure why the old Alonzo examples are now getting `MalformedScripts` predicate failures. Once we figure that out, this PR is good to merge.

goolord:
should be all premo now :ok_person:

PR #2724, opened 2022-04-06
Plutus event suggestion

lehins:


JaredCorduan (commented):
This is clever! I'll use it if I can't quickly come up with a decent solution to the root of the problem...  but thanks!

PR #2726, opened 2022-04-07
Introduce RestrainedRewards event

kderme:
This gives us a way to clean up rewards on the epoch boundary to addresses that are deregistered or rewards subject to era specific issues.


JaredCorduan (approved):
looks great @kderme ! :raised_hands:

Issue #2684, opened 2022-03-09
Audit/Compare the Babbage spec to the implementation


JaredCorduan:
We need to line for line compare the Babbage specification with the implementation.

PR #2727, opened 2022-04-07
several small babbage audit tweaks

JaredCorduan:
During the internal Babbage audit [#2684](https://github.com/input-output-hk/cardano-ledger/issues/2684) we found a handful of small things that we wanted to clean up. The PR addresses them, each of the six commits is one small fix.

### DIFF
--- a/eras/babbage/impl/src/Cardano/Ledger/Babbage/Collateral.hs
+++ b/eras/babbage/impl/src/Cardano/Ledger/Babbage/Collateral.hs
@@ -19,7 +19,7 @@ import Cardano.Ledger.Babbage.TxBody
     outputs',
     txfee',
   )
-import Cardano.Ledger.BaseTypes (txIxFromIntegral)
+import Cardano.Ledger.BaseTypes (TxIx (..), txIxFromIntegral)
 import Cardano.Ledger.Coin (Coin (..))
 import qualified Cardano.Ledger.Core as Core
 import Cardano.Ledger.Era (Era (Crypto), ValidateScript (..))
@@ -30,6 +30,7 @@ import Data.Compact.SplitMap ((◁))
 import qualified Data.Compact.SplitMap as SplitMap
 import Data.Maybe.Strict (StrictMaybe (..))
 import Data.Set (Set)
+import Data.Word (Word16)
 import GHC.Records (HasField (..))
 import Numeric.Natural (Natural)
 
@@ -90,4 +91,7 @@ collOuts txb =
       where
         index = case txIxFromIntegral (length (outputs' txb)) of
           Just i -> i
-          Nothing -> error ("length outputs, should always fit in a TxIx")
+          -- In the impossible event that there are more transaction outputs
+          -- in the transaction than will fit into a Word16 (which backs the TxIx),
+          -- we give the collateral return output an index of maxBound.
+          Nothing -> TxIx (maxBound :: Word16)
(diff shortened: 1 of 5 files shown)


eras/babbage/impl/src/Cardano/Ledger/Babbage/Collateral.hs
```diff
@@ -90,4 +90,7 @@ collOuts txb =
       where
         index = case txIxFromIntegral (length (outputs' txb)) of
           Just i -> i
-          Nothing -> error ("length outputs, should always fit in a TxIx")
+          -- In the improbable event that there are more transaction outputs
+          -- in the transaction than will fit into a Word16 (which backs the TxIx),
+          -- we give the collateral return output the zero index.
+          Nothing -> TxIx 0
```
lehins: I am not sure it makes sense. It is not improbable, but an impossible event. So using an arbitrary `0`  is worse choice than `error` IMHO.

One way or another `improbable`  has to be changed to `impossible` in the comment, otherwise it leaves room for doubt.

eras/babbage/impl/src/Cardano/Ledger/Babbage/Rules/Utxos.hs
```diff
@@ -153,6 +153,14 @@ scriptsYes = do
   sysSt <- liftSTS $ asks systemStart
   ei <- liftSTS $ asks epochInfo
 
+  -- We intentionally run the PPUP rule before evaluating any Plutus scripts.
+  -- We do not want to waste computation running plutus scripts if the
+  -- transaction will fail for other reasons.
```
```suggestion
  -- transaction will fail due to `PPUP`
```

lehins (approved):
Non critical suggestions

eras/babbage/impl/src/Cardano/Ledger/Babbage/Collateral.hs
JaredCorduan: I said "improbable" instead of "impossible" just because a bug could cause the `Nothing` code path to be exercised. And were this to happen, I figured we would prefer using a zero index instead of crashing the node. This is admittedly all pretty far fetched. :shrug:

lehins: That's the point, bugs should be impossible! If we do take bugs into consideration then anything is both possible and probable. We do assume the state of the system with no bugs when we write documentation and code.

I guess my real issue is with wording "improbable". It might be my interpretation of English meaning of the word, but in my mind it leaves out some chance for that event to occur, no matter how minuscule it is. English is not my native language, so please correct me if I am wrong on this.

With respect to choice of `error` or `0`, both are are equally bad, but that case is indeed impossible (forbidden by deserialization too) so how we deal with it is another important discussion. I agree that failing node is bad, but leaving a security whole of defaulting to `0` is also bad, so maybe `(maxBound :: Word16)` would be a safer alternative for the worst case scenario, since that is the least likely case possible.

JaredCorduan: You're correct about the meaning of "improbable". I'll change it to "impossible", that does make our expectations clearer.

I like using `maxBound :: Word16`. In the event of a bug, `0` could go unnoticed. but `maxBound :: Word16` would stand out. But I'm curious what security problem you see with using `0`.

lehins: I usually assume any bug as a potential security concern right away, since it is hard to predict how bugs can be used in order to break the system. In this case however, I looked at how index is being used and I can't think of any potential problem in ledger itself, regardless of the value used for the index, because there can only be one return collateral output.

That being said if for some reason index were to overflow and set to `0`, then anyone looking at the the TxBody and the value of TxIn  in isolation would think it refers to the first TxOut in the transaction instead of a collateral return TxOut, which could lead to bugs in other programs written that try to deal with ledger directly. It is all far fetched speculation though. So I don't think we need to dwell on it.

One way or another, I am glad we agree that using ceiling for overflow, rather than wrap around is better as fallback. :+1:

JaredCorduan: :raised_hands:

PR #2728, opened 2022-04-13
CAD-4157: Stash AVVM to allow on-disk UTxO.

nc6:
This commit provides ledger support for the transition of UTxO to
on-disk storage. In particular, the translation between Shelley and
Allegra is currently problematic, owing to the need to remove AVVM
addresses (dormant since Byron) from the UTxO. This requires a linear
scan of the UTxO, which is now stored on disk and hence cannot/should
not be accessed in such a way.

The solution to this is a bit of a hack; since the Byron UTxO are not
stored on disk, and since AVVM addresses remain untouched during the
Shelley era, we instead perform the scan at the Byron/Shelley boundary,
store the set of UTxO in the ledger state, only to yield them to the
consensus layer at the Shelley/Allegra boundary. Thus provided with the
set of addresses, consensus passes them back as the UTxO, allowing the
translation function to delete them.

This change is written such that, with no on-disk UTxO, the ledger
functionality remains identical, other than a slight bump in memory
usage during the Shelley era.

As far as possible, this commit minimises ledger changes. Thus, a
pattern is introduced replicating the existing `NewEpochState`
constructor. Unfortunately, there are two knock-on effects related to
weaknesses in pattern synonyms:

- A couple of patterns matching on `NewEpochState` now seem to be
  failable. Since they cannot fail, we instead make them irrefutable.
- Record fields associated with pattern synonyms do not play nicely with
  `NamedFieldPuns`. As such, we replace a couple of puns with explicit
  bindings.

A single additional function `shelleyToAllegraAVVMsToDelete` is exported
for the use of the consensus layer. Otherwise no API changes are made.

This does entail a change in the ledger state serialisation format.

### DIFF
--- a/eras/shelley-ma/impl/src/Cardano/Ledger/Allegra/Translation.hs
+++ b/eras/shelley-ma/impl/src/Cardano/Ledger/Allegra/Translation.hs
@@ -44,6 +44,14 @@ import qualified Data.Map.Strict as Map
 -- being total. Do not change it!
 --------------------------------------------------------------------------------
 
+-- | Return the subset of UTxO corresponding to Byron-era AVVM addresses, which
+-- are to be removed on the Shelley/Allegra boundary. This set will be passed
+-- _back_ to the translation functions as the UTxO, allowing these addresses to
+-- be removed. This is needed because we cannot do a full scan on the UTxO at
+-- this point, since it has been persisted to disk.
+shelleyToAllegraAVVMsToDelete :: NewEpochState era -> StrictMaybe (UTxO era)
+shelleyToAllegraAVVMsToDelete = stashedAVVMAddresses
+
 type instance PreviousEra (AllegraEra c) = ShelleyEra c
 
 -- | Currently no context is needed to translate from Shelley to Allegra.
@@ -62,6 +70,11 @@ instance Crypto c => TranslateEra (AllegraEra c) NewEpochState where
           nesEs = translateEra' ctxt $ LS.returnRedeemAddrsToReserves . nesEs $ nes,
           nesRu = nesRu nes,
           nesPd = nesPd nes
+        }
+        { -- At this point, the consensus layer has passed in our stashed AVVM
+          -- addresses as our UTxO, and we have deleted them above (with
+          -- 'returnRedeemAddrsToReserves'), so we may safely discard this map.
+          stashedAVVMAddresses = SNothing
         }
 
 instance forall c. Crypto c => TranslateEra (AllegraEra c) Tx where
--- a/eras/shelley-ma/impl/src/Cardano/Ledger/Allegra/Translation.hs
+++ b/eras/shelley-ma/impl/src/Cardano/Ledger/Allegra/Translation.hs
@@ -49,7 +49,7 @@ import qualified Data.Map.Strict as Map
 -- _back_ to the translation functions as the UTxO, allowing these addresses to
 -- be removed. This is needed because we cannot do a full scan on the UTxO at
 -- this point, since it has been persisted to disk.
-shelleyToAllegraAVVMsToDelete :: NewEpochState era -> StrictMaybe (UTxO era)
+shelleyToAllegraAVVMsToDelete :: NewEpochState (ShelleyEra c) -> UTxO (ShelleyEra c)
 shelleyToAllegraAVVMsToDelete = stashedAVVMAddresses
 
 type instance PreviousEra (AllegraEra c) = ShelleyEra c
@@ -69,12 +69,11 @@ instance Crypto c => TranslateEra (AllegraEra c) NewEpochState where
           nesBcur = nesBcur nes,
           nesEs = translateEra' ctxt $ LS.returnRedeemAddrsToReserves . nesEs $ nes,
           nesRu = nesRu nes,
-          nesPd = nesPd nes
-        }
-        { -- At this point, the consensus layer has passed in our stashed AVVM
+          nesPd = nesPd nes,
+          -- At this point, the consensus layer has passed in our stashed AVVM
           -- addresses as our UTxO, and we have deleted them above (with
           -- 'returnRedeemAddrsToReserves'), so we may safely discard this map.
-          stashedAVVMAddresses = SNothing
+          stashedAVVMAddresses = ()
         }
 
 instance forall c. Crypto c => TranslateEra (AllegraEra c) Tx where
--- a/eras/shelley/impl/src/Cardano/Ledger/Shelley/API/ByronTranslation.hs
+++ b/eras/shelley/impl/src/Cardano/Ledger/Shelley/API/ByronTranslation.hs
@@ -101,16 +101,15 @@ translateToShelleyLedgerState genesisShelley epochNo cvs =
       nesBcur = BlocksMade Map.empty,
       nesEs = epochState,
       nesRu = SNothing,
-      nesPd = PoolDistr Map.empty
-    }
-    { -- At this point, we compute the stashed AVVM addresses, while we are able
+      nesPd = PoolDistr Map.empty,
+      -- At this point, we compute the stashed AVVM addresses, while we are able
       -- to do a linear scan of the UTxO, and stash them away for use at the
       -- Shelley/Allegra boundary.
       stashedAVVMAddresses =
         let UTxO utxo = _utxo . lsUTxOState . esLState $ epochState
             redeemers =
               SplitMap.filter (maybe False isBootstrapRedeemer . getTxOutBootstrapAddress) utxo
-         in SJust $ UTxO redeemers
+         in UTxO redeemers
     }
   where
     pparams :: PParams (ShelleyEra c)
--- a/eras/shelley/impl/src/Cardano/Ledger/Shelley/API/ByronTranslation.hs
+++ b/eras/shelley/impl/src/Cardano/Ledger/Shelley/API/ByronTranslation.hs
@@ -17,10 +17,12 @@ import qualified Cardano.Chain.Common as Byron
 import qualified Cardano.Chain.UTxO as Byron
 import qualified Cardano.Crypto.Hash as Crypto
 import qualified Cardano.Crypto.Hashing as Hashing
+import Cardano.Ledger.Address (isBootstrapRedeemer)
 import Cardano.Ledger.BaseTypes (BlocksMade (..), TxIx (..))
 import Cardano.Ledger.Coin (CompactForm (CompactCoin))
 import Cardano.Ledger.CompactAddress (CompactAddr (UnsafeCompactAddr))
 import qualified Cardano.Ledger.Crypto as CC
+import Cardano.Ledger.Era (getTxOutBootstrapAddress)
 import Cardano.Ledger.SafeHash (unsafeMakeSafeHash)
 import Cardano.Ledger.Shelley (ShelleyEra)
 import Cardano.Ledger.Shelley.API.Types
@@ -100,6 +102,15 @@ translateToShelleyLedgerState genesisShelley epochNo cvs =
       nesEs = epochState,
       nesRu = SNothing,
       nesPd = PoolDistr Map.empty
+    }
+    { -- At this point, we compute the stashed AVVM addresses, while we are able
+      -- to do a linear scan of the UTxO, and stash them away for use at the
+      -- Shelley/Allegra boundary.
+      stashedAVVMAddresses =
+        let UTxO utxo = _utxo . lsUTxOState . esLState $ epochState
+            redeemers =
+              SplitMap.filter (maybe False isBootstrapRedeemer . getTxOutBootstrapAddress) utxo
+         in SJust $ UTxO redeemers
     }
   where
     pparams :: PParams (ShelleyEra c)
--- a/eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState.hs
+++ b/eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState.hs
@@ -13,11 +13,12 @@
 {-# LANGUAGE MultiParamTypeClasses #-}
 {-# LANGUAGE NamedFieldPuns #-}
 {-# LANGUAGE OverloadedStrings #-}
-{-# LANGUAGE PatternSynonyms #-}
+{-# LANGUAGE QuantifiedConstraints #-}
 {-# LANGUAGE ScopedTypeVariables #-}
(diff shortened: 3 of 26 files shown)


eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/NewEpoch.hs
```diff
@@ -147,7 +147,7 @@ newEpochTransition ::
 newEpochTransition = do
   TRC
     ( _,
-      src@(NewEpochState (EpochNo eL) _ bcur es ru _pd),
+      ~src@(NewEpochState (EpochNo eL) _ bcur es ru _pd),
```
nfrisby: I suspect the matching occurrence of `NewEpochState` in the returned expression about 20 lines below means that `newEpochTransition` always forgets the map. That's problematic, right?

Maybe make `NewEpochState` a unidirectional pattern? Thereby forcing folks to use record update expressions to update the incoming `nes`, which would avoid this kind of forgetting.

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Tick.hs
```diff
@@ -179,7 +179,7 @@ bheadTransition ::
   ) =>
   TransitionRule (TICK era)
 bheadTransition = do
-  TRC ((), nes@(NewEpochState _ bprev _ es _ _), slot) <-
+  TRC ((), ~nes@(NewEpochState _ bprev _ es _ _), slot) <-
```
nfrisby: These lazy patterns are needed even with the `COMPLETE` pragma? Harumph.

nfrisby (dismissed):
I'm Requesting Changes, because I _think_ I noticed and commented on a bug.

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/NewEpoch.hs
nc6: Ah, darn, you're complete right. Good catch, and I like the suggested fix

nc6: Unfortunately, you can't use unidirectional patterns in record updates :-( But I'll disable the pattern _temporarily_ and use that to guide updates.

eras/shelley/impl/src/Cardano/Ledger/Shelley/API/ByronTranslation.hs
```diff
@@ -101,6 +103,15 @@ translateToShelleyLedgerState genesisShelley epochNo cvs =
...
       nesPd = PoolDistr Map.empty
     }
+    { -- At this point, we compute the stashed AVVM addresses, while we are able
+      -- to do a linear scan of the UTxO, and stash them away for use at the
+      -- Shelley/Allegra boundary.
+      stashedAVVMAddresses =
+        let UTxO utxo = _utxo . lsUTxOState . esLState $ epochState
+            (redeemers, _) =
+              SplitMap.partition (maybe False isBootstrapRedeemer . getTxOutBootstrapAddress) utxo
+         in SJust $! UTxO redeemers
```
lehins: SJust is already strict:
```suggestion
         in SJust $ UTxO redeemers
```

eras/shelley-ma/impl/src/Cardano/Ledger/Allegra/Translation.hs
```diff
@@ -63,6 +71,11 @@ instance Crypto c => TranslateEra (AllegraEra c) NewEpochState where
           nesRu = nesRu nes,
           nesPd = nesPd nes
         }
+        { -- At this point, the consensus layer has passed in our stashed AVVM
```
lehins: This syntax looks weird chained like that :smile: Never seen it used before and it took me a second to realize what's going on

eras/shelley/impl/src/Cardano/Ledger/Shelley/API/ByronTranslation.hs
```diff
@@ -101,6 +103,15 @@ translateToShelleyLedgerState genesisShelley epochNo cvs =
       nesRu = SNothing,
       nesPd = PoolDistr Map.empty
     }
+    { -- At this point, we compute the stashed AVVM addresses, while we are able
+      -- to do a linear scan of the UTxO, and stash them away for use at the
+      -- Shelley/Allegra boundary.
+      stashedAVVMAddresses =
+        let UTxO utxo = _utxo . lsUTxOState . esLState $ epochState
+            (redeemers, _) =
+              SplitMap.partition (maybe False isBootstrapRedeemer . getTxOutBootstrapAddress) utxo
```
lehins: This is just a filter
```suggestion
            redeemers =
              SplitMap.filter (maybe False isBootstrapRedeemer . getTxOutBootstrapAddress) utxo
```

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/NewEpoch.hs
lehins: Providing unidirctional pattern synonym without record fields (thus also not renaming original fields) and not exporting the constructor would force folks to use record update syntax.

lehins (approved):
I am not too excited about another pattern synonym, but it does get the job done.

Other than that there are couple minor suggestions.

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/NewEpoch.hs
nfrisby: There are a handful of design choices here. The only thing I'm relatively confident/strongly opinionated about is that if we export a constructor (via the bidirectional pattern) that _implicitly_ sets the field to `SNothing`, then that's Chekhov's footgun.

My other suggestion is to use a `type` or `data` family for the new field along these lines.

```
data NewEpochState era = ... { ... stashedAVVMAddresses :: StashedAVVMAddresses era }

type family StashedAVVMAddresses era where
  StashedAVVMAddresses (ShelleyEra c) = UTxO era
  StashedAVVMAddresses era = ()
```

Edit: I'm leaving my review as Request Changes until we all agree we're confident about the `NewEpochState` constructor expression not being misused in this code or downstream. (Feel free to object to that choice -- I'm struggling to find the words here and don't feel as strongly as that previous sentence might sound.)

nc6: > Providing unidirctional pattern synonym without record fields (thus also not renaming original fields) and not exporting the constructor would force folks to use record update syntax.

It would, but it would also force anywhere that used the `NewEpochState` as a record pattern to have to change, since the record selectors cannot share names.

eras/shelley-ma/impl/src/Cardano/Ledger/Allegra/Translation.hs
nc6: I agree it's a weird usage, but I kind of liked it here to enforce the distinction between the original code and the hackery!

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/NewEpoch.hs
lehins: @nfrisby I really like the close type family solution!

nc6:
So, I like both of these things in theory, but having tried to put them into practice they both cause sufficient enough disruption that I don't want to introduce them.

- Having a closed type family is really nice, but it introduces the issue that there can be no default constructed value. We have a couple of places where this is a problem:
  - In the initial rules, which we still use for `NEWEPOCH`, and
  - In the bidirectional pattern, meaning that choosing the closed type family implies dropping the bidirectional pattern, as below.

- Dropping the bidirectional pattern forces us to either have the record syntax for pattern matching or updates, but not both. This to me is too high a price to pay. I'm not overly concerned by the calibre of the footgun here - the worst result a mistake here can cause is that we fall over when validating the Shelley/Allegra boundary, something we are forced to catch well before release.

I think the only palatable alternative would be Alexey's original suggestion, to drop the pattern altogether and force everyone to deal with the new field. For discussion, I'll push that as a separate commit.

eras/shelley-ma/impl/src/Cardano/Ledger/Allegra/Translation.hs
```diff
@@ -61,7 +69,11 @@ instance Crypto c => TranslateEra (AllegraEra c) NewEpochState where
           nesBcur = nesBcur nes,
           nesEs = translateEra' ctxt $ LS.returnRedeemAddrsToReserves . nesEs $ nes,
```
JaredCorduan, line 70: The use of `returnRedeemAddrsToReserves` here might be confusing to future readers. This function both removes the AVVM addresses from the UTxO and also increases the reserves balance by the corresponding amount. Here, however, the first operation is a no-op since we (the informed) know that we have already removed the AVVM, and so `returnRedeemAddrsToReserves` is only being used to increase the reserves. At the very least, maybe we could add a comment?

JaredCorduan:
@nc6 I think we should do something to account for the fact that we've just created a `Seventh Pot` of ADA. I bet that the property tests did not notice this since we never generate AVVM addresses in testing.

I can think of two reasonable options to fix `totalAdaPotsES`.

https://github.com/input-output-hk/cardano-ledger/blob/1db68a3ec0a2dcb5751004beb22b906162474f23/eras/shelley/impl/src/Cardano/Ledger/Shelley/API/Wallet.hs#L561-L589

One option is to add a seventh pot. The other option is to include `stashedAVVMAddresses` in the UTxO pot.

We might also want to add some prose about this in the errata in the Shelley spec, specifically in the section about this particular wart.

eras/shelley-ma/impl/src/Cardano/Ledger/Allegra/Translation.hs
nc6: We haven't removed the AVVM addresses! On the Byron/Shelley border, we don't remove them, we just take note of them. They are removed, and the pots adjusted, on the Shelley/Allegra border, exactly the same as before.

nc6:
@JaredCorduan I'm afraid I don't understand your comments (or fear that they are based on a misconception of the work here). Whilst we did discuss a "seventh pot" approach, that is _not_ what is implemented here. Consider in particular https://github.com/input-output-hk/cardano-ledger/pull/2728/files#diff-b403883c359c9c3f41b927c964599024c8c6f3a10cf70027835ec3eb44b1a4f6L100 - you will see that the UTxO is _not_ modified on the Byron/Shelley boundary. The only thing that happens is that a new field - `stashedAVVMAddresses` is populated. This field is never read or modified until the Shelley/Allegra translation, where it is deleted. From a ledger semantics perspective, this PR is a complete no-op - we put something somewhere, never look at it, and then delete it. UTxO, reserves etc remain entirely as they were before.

It is only by the action of consensus that this PR has any effect, and even then, it is simply to ensure the fundamental property of UTxO HD - that the ledger should behave identically on the full UTxO and on the _view_ of the UTxO presented by the consensus layer. In this case, by ensuring that on the Shelley/Allegra boundary, the UTxO contains the AVVM addresses such that they can be deleted.

Ideally, we could emphasize this by only storing the TxId in `stashedAVVMAddresses`, but unfortunately the consensus layer (currently) cannot perform lookup of the UTxO when doing translation, so this isn't possible.

eras/shelley-ma/impl/src/Cardano/Ledger/Allegra/Translation.hs
JaredCorduan: ah!!

JaredCorduan (approved):
LGTM!

eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState.hs
```diff
@@ -767,22 +792,29 @@ instance
...
     ToCBOR (Core.PParams era),
-    ToCBOR (State (Core.EraRule "PPUP" era))
+    ToCBOR (State (Core.EraRule "PPUP" era)),
+    ToCBOR (StashedAVVMAddresses era)
   ) =>
   ToCBOR (NewEpochState era)
   where
-  toCBOR (NewEpochState e bp bc es ru pd) =
-    encodeListLen 6 <> toCBOR e <> toCBOR bp <> toCBOR bc <> toCBOR es
+  toCBOR (NewEpochState e bp bc es ru pd av) =
```
nfrisby, line 800: If this is going to be released before UTxO HD, then we don't _need_ to make the serialization backwards incompatible. We could just deserialize a `ListLen 6` as having an empty map, and a hypothetical non-yet-UTxO-HD Consensus layer downstream wouldn't notice.

nfrisby: Alternatively, we could only change the serialization for eras for which this is not `()`. That way today's users would never have to reapply from the genesis block. (My goal is just to avoid an unnecessary long replay for users, but that's ultimately for a different decision maker to balance.)

eras/shelley/test-suite/src/Test/Cardano/Ledger/Shelley/Examples/Consensus.hs
```diff
@@ -311,7 +313,8 @@ exampleNewEpochState value ppp pp =
       nesBcur = BlocksMade (Map.singleton (mkKeyHash 2) 3),
       nesEs = epochState,
       nesRu = SJust rewardUpdate,
-      nesPd = examplePoolDistr
+      nesPd = examplePoolDistr,
+      stashedAVVMAddresses = def
```
nfrisby, line 317: Worthwhile to make this non-empty for the Shelley era?

nfrisby (commented):
Looks right to me! Thank much.

The PR description is out of date: eg I don't see any new pattern synonym in this latest diff.

I also made a couple minor comments below.

nfrisby:
> I think the only palatable alternative would be Alexey's original suggestion, to drop the pattern altogether and force everyone to deal with the new field. For discussion, I'll push that as a separate commit.

Looks like this is what you all went with? It's what I see in the diff.

eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState.hs
nc6: I considered that, but there are already serialisation changes in the ledger which will be visible in the next release, so there's not much point.

PR #2703, opened 2022-03-25
hide CostModel constructor

JaredCorduan:
The `CostModel` constructor should have been hidden when the `EvaluationContext` was added to it, `costModelParamsToCostModel` is the appropriate smart constructor that everyone should use. The evaluation
context is completely dependent on the cost model parameters.

PR #2730, opened 2022-04-18
preserve Alonzo script integrity hash bug

JaredCorduan:
In the Alonzo era, the map of languages to cost models was mistakenly encoded using an indefinite CBOR map (contrary to canonical CBOR, as intended) when computing the script integrity hash.

For this reason, PlutusV1 remains with this encoding. Future versions of Plutus, starting with PlutusV2 in the Babbage era, will
use the intended definite length encoding.

PR #2703 accidentally reversed this Alonzo bug, this commit restores it. See https://github.com/input-output-hk/cardano-ledger/pull/2703/files#diff-600ae5ce75fa519690c6b948aa3c3453878235d279c5b800efe49e79d2927363L494-L500


lehins (approved):
I had a feeling this issue had something to do with CostModels serialization. 
:100:

PR #2731, opened 2022-04-18
do not serialize the block header to get the size

JaredCorduan:
The block header already stores its serialized bytes, it is a big waste to recompute it in order to get the size.

Many thanks to @RyanGlScott for spotting this!! :raised_hands:

### DIFF
--- a/libs/cardano-protocol-tpraos/src/Cardano/Protocol/TPraos/BHeader.hs
+++ b/libs/cardano-protocol-tpraos/src/Cardano/Protocol/TPraos/BHeader.hs
@@ -103,7 +103,6 @@ import Cardano.Ledger.Slot (BlockNo (..), SlotNo (..))
 import Cardano.Protocol.TPraos.OCert (OCert (..))
 import Cardano.Slotting.Slot (WithOrigin (..))
 import Control.DeepSeq (NFData)
-import qualified Data.ByteString as BS
 import qualified Data.ByteString.Builder as BS
 import qualified Data.ByteString.Builder.Extra as BS
 import qualified Data.ByteString.Lazy as BSL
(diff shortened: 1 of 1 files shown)


libs/cardano-protocol-tpraos/src/Cardano/Protocol/TPraos/BHeader.hs
```diff
@@ -358,12 +357,8 @@ prevHashToNonce = \case
...
 issuerIDfromBHBody = hashKey . bheaderVk
 
-bHeaderSize ::
-  forall crypto.
-  (CC.Crypto crypto) =>
-  BHeader crypto ->
-  Int
-bHeaderSize = BS.length . serialize'
+bHeaderSize :: forall crypto. BHeader crypto -> Int
+bHeaderSize = fromIntegral . BSL.length . bHeaderBytes
```
lehins, line 361: `fromIntegral` is redundant
```suggestion
bHeaderSize = BSL.length . bHeaderBytes
```

lehins (approved):
One comment, but otherwise this is a great catch

libs/cardano-protocol-tpraos/src/Cardano/Protocol/TPraos/BHeader.hs
JaredCorduan: @lehins and I chatted about this, turns out `BSL.length` returns an `Int64`. We are working under the assumption that we do not support 32 bit architecture so that this conversion make sense.

lehins: This can only be a problem if a Block header serializes to a ByteString larger than 2Gb on a 32bit machine, which is definitely not a realistic scenario. Moreover previous implementation had the same semantics, except it first converted the lazy ByteString into a strict one.

This comment was because I totally forgot that length on lazy ByteString returns `Int64` and not `Int`

PR #2732, opened 2022-04-19
update plutus to latest in release/1.0.0

JaredCorduan:
Updating Plutus to get https://github.com/input-output-hk/plutus/pull/4540

and also https://github.com/input-output-hk/plutus/pull/4560


nc6 (approved):

JaredCorduan:
@nc6 - fyi, I've pushed a commit (after you approved) which intentionally breaks the tests. The PV2 cost model serialization is supposed to be different.

eras/alonzo/test-suite/test/Test/Cardano/Ledger/Alonzo/Golden.hs
```diff
@@ -207,7 +207,7 @@ exampleLangDepViewPV1 = LangDepView b1 b2
     b2 =
       fromRight (error "invalid hex encoding of the cost model inside exampleLangDepViewPV1") $
         B16.decode $
-          "5901d59f1a000302590001011a00060bc719026d00011a000249f01903e80001"
+          "5901e19f1a000302590001011a00060bc719026d00011a000249f01903e80001"
```
JaredCorduan, line 210: The two golden tests for the script integrity hash need to be updated since the value of `costModelParamsForTesting` changed. If @bezirg can confirm that we expected `costModelParamsForTesting` to change, then updating these two tests is expected.

JaredCorduan: I chatted with @bezirg, all is good. We might consider making our own cost model for this golden test next time `costModelParamsForTesting` changes, though.

PR #2733, opened 2022-04-19
Added the allOuts function, made the script witness check stricter

Soupstraw:


eras/babbage/impl/src/Cardano/Ledger/Babbage/Rules/Utxo.hs
```diff
@@ -242,6 +242,80 @@ validateCollateralEqBalance bal txcoll =
...
+validateOutputTooSmallUTxO pp outs = failureUnless (null outputsTooSmall) $ OutputTooSmallUTxO outputsTooSmall
+  where
+    Coin coinsPerUTxOWord = getField @"_coinsPerUTxOWord" pp
+    serSize = fromIntegral . BSL.length . serialize . getField @"value"
+    outputsTooSmall =
+      filter
+        ( \out ->
+            let v = getField @"value" out
+             in -- pointwise is used because non-ada amounts must be >= 0 too
+                not $ Val.pointwise (>=) v (Val.inject $ Coin (serSize out * coinsPerUTxOWord `div` 8))
```
Soupstraw: multiply first, division rounded up

eras/babbage/impl/src/Cardano/Ledger/Babbage/Rules/Utxo.hs
```diff
@@ -242,6 +242,80 @@ validateCollateralEqBalance bal txcoll =
     SNothing -> pure ()
     SJust tc -> failureUnless (bal == tc) (UnequalCollateralReturn bal tc)
 
+-- > getValue txout ≥ inject ( ⌈ serSize txout ∗ coinsPerUTxOWord pp / 8 ⌉ )
```
Soupstraw, line 244: rename `coinsPerUTxOWord` to `coinsPerUTxOByte`

eras/babbage/impl/src/Cardano/Ledger/Babbage/Rules/Utxow.hs
```diff
@@ -63,52 +69,57 @@ import Control.State.Transition.Extended
...
-    else
-      failureUnless (sNeeded == sReceived) $
-        Shelley.MissingScriptWitnessesUTXOW (sNeeded `Set.difference` sReceived)
+babbageMissingScripts _ sNeeded sRefs sReceived =
+  sequenceA_
+    [ failureUnless (Set.null extra) $ Shelley.ExtraneousScriptWitnessesUTXOW extra,
+      failureUnless (Set.null missing) $ Shelley.MissingScriptWitnessesUTXOW missing
+    ]
+  where
+    -- FIXME what about the hard forks?
```
Soupstraw: Remove this

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Examples/BabbageFeatures.hs
```diff
@@ -344,7 +344,8 @@ referenceScriptTx pf =
       WitnessesI
         [ AddrWits' [makeWitnessVKey (hashAnnotated (referenceScriptTxBody pf)) (someKeys pf)],
           DataWits' [datumExample1],
-          RdmrWits validatingRedeemersEx1
```
Soupstraw: Have Jared look over these

eras/babbage/impl/src/Cardano/Ledger/Babbage/Rules/Utxo.hs
JaredCorduan: we can do that in a follow up PR

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Examples/BabbageFeatures.hs
JaredCorduan: I pushed a commit, we can removed these added script witnesses, which I've done after fixing a problem with how script references were being gather.

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Examples/BabbageFeatures.hs
```diff
@@ -904,6 +892,17 @@ genericBabbageFeatures pf =
               pf
               (trustMeP pf True $ malformedScriptsTx pf)
               (Left [fromUtxoB @era (MalformedScripts (Set.fromList [hashScript @era $ malformedScript pf "rs"]))]),
+          testCase "inline datum and ref script and redundant script witness" $
```
JaredCorduan, line 895: Note that this went from a valid transaction to an invalid transaction. I'm happy about this! We no longer allow this extraneous script.

JaredCorduan (approved):
@Soupstraw this PR looks great! Let's clean up the commit history a bit before we merge, but the changes are good. We do obviously need to figure out why the Babbage generators are supplying extraneous script witnesses, but we can do that in a follow up PR.

eras/babbage/impl/src/Cardano/Ledger/Babbage/Rules/Utxo.hs
```diff
@@ -242,6 +242,80 @@ validateCollateralEqBalance bal txcoll =
...
+      filter
+        ( \txOut ->
+            case getTxOutBootstrapAddress txOut of
+              Just addr -> bootstrapAddressAttrsSize addr > 64
+              _ -> False
+        )
+        outs
+
+-- > ∀(_ → (a, _)) ∈ allOuts txb, netId a = NetworkId
+validateWrongNetwork ::
```
JaredCorduan: I think we could alternatively change the old version of this function to take outputs instead of a tx body, and then re-use it.

eras/babbage/impl/src/Cardano/Ledger/Babbage/Rules/Utxo.hs
```diff
@@ -242,6 +242,80 @@ validateCollateralEqBalance bal txcoll =
...
+    ToCBOR (Core.Value era),
+    HasField "_coinsPerUTxOWord" (Core.PParams era) Coin
+  ) =>
+  Core.PParams era ->
+  [Core.TxOut era] ->
+  Test (UtxoPredicateFailure era)
+validateOutputTooSmallUTxO pp outs = failureUnless (null outputsTooSmall) $ OutputTooSmallUTxO outputsTooSmall
+  where
+    Coin coinsPerUTxOWord = getField @"_coinsPerUTxOWord" pp
+    serSize = fromIntegral . BSL.length . serialize . getField @"value"
```
JaredCorduan, line 258: let's add a comment here that we probably do not want to re-serialize the output, but rather use the length of the actual supplied bytes. this is a much trickier change, we should do it in a follow-up PR.

PR #2734, opened 2022-04-19
base Babbage min UTxO value on serialized bytes

### LINKED ISSUE #2682: New babbage min-utxo calculation
Figure 4 of the Babbage spec introduces a new calculation for the minimum ada needed inside a given utxo.

JaredCorduan:
In the Babbage era, the minimum lovelace required for an output is based on the serialized bytes.

To accomplish this, I've added a `Word64` to each of the `TxOut` constructors. The `TxOut` pattern automatically fills in this value. Note also that this value is not a part of the wire specification, but is added on deserialization.

closes #2682


JaredCorduan:
closing this PR for two reasons:
* #2733 also solves this same issue, but looks better if we are reserializing the output
* I should have used the bytes on the wire, not reserialized.

PR #2735, opened 2022-04-20
do not allow encoding NoDatum, instead omit it

JaredCorduan:
The `NoDatum` constructor of the `Datum` type is really an implementation detail and not something that we should be exposing on the wire format. If there is no datum present, the entire field (inside the transaction output) should be omitted, which is what our serializers do anyway.

~It is unfortunate that the `Datum` serializer is now partial, but I think it is still worth it.~ `OmitC` to the rescue, thanks for the suggestion @lehins !

### DIFF
--- a/eras/babbage/impl/src/Cardano/Ledger/Babbage/TxBody.hs
+++ b/eras/babbage/impl/src/Cardano/Ledger/Babbage/TxBody.hs
@@ -270,16 +270,15 @@ data Datum era
 
 instance Era era => ToCBOR (Datum era) where
   toCBOR d = encode $ case d of
-    NoDatum -> Sum NoDatum 0
-    DatumHash dh -> Sum DatumHash 1 !> To dh
-    Datum d' -> Sum Datum 2 !> To d'
+    DatumHash dh -> Sum DatumHash 0 !> To dh
+    Datum d' -> Sum Datum 1 !> To d'
+    NoDatum -> OmitC NoDatum
 
 instance Era era => FromCBOR (Datum era) where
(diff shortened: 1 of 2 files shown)


eras/babbage/impl/src/Cardano/Ledger/Babbage/TxBody.hs
```diff
@@ -270,16 +270,15 @@ data Datum era
 
 instance Era era => ToCBOR (Datum era) where
   toCBOR d = encode $ case d of
-    NoDatum -> Sum NoDatum 0
-    DatumHash dh -> Sum DatumHash 1 !> To dh
-    Datum d' -> Sum Datum 2 !> To d'
+    DatumHash dh -> Sum DatumHash 0 !> To dh
+    Datum d' -> Sum Datum 1 !> To d'
+    NoDatum -> error "NoDatum should always be omitted, not encoded"
```
lehins: Why error? Would something like `OmitC` work?

It is a noop to serialize `NoDatum`, right? In this kind of place there is gotta be a way to avoid a partial function

lehins (changes requested):
Totally agree with not serializing NoDatum, but there is gotta be a way to avoid `error`.

eras/babbage/impl/src/Cardano/Ledger/Babbage/TxBody.hs
JaredCorduan: `OmitC` is a much better choice, thanks!

lehins (approved):
:100:

PR #2747, opened 2022-04-21
Whitenoise generators without EraGen

TimSheard:
The module Test.Cardano.Ledger.Shelley.Serialisation.EraIndepGenerators had several white noise Arbitrary instances that had EraGen as preconditions (super classes). It was possible to remove all of these.  So now the tests in the consensus layer should not need EraGen instances. 

I also point out that a few Babbage specific Arbitrary instances can be found here
src/Test/Cardano/Ledger/Babbage/Serialisation/Generators.hs


JaredCorduan (approved):
beautiful :clap:

nc6:
Awesome, thanks Tim!

PR #2748, opened 2022-04-22
Add a function to check the VRF natural value directly.

nc6:
This is helpful in order to allow different VRF uses (in Praos) to be
type-differentiated.


lehins (approved):
:+1:

PR #2750, opened 2022-04-24
Address deserialization

### LINKED ISSUE #2729: Handling of broken Alonzo address missing from the spec
# Background

On August 28th, cardano-node 1.29.0 was released. Shortly after, on Sept 5th, [this transaction](https://cardanoscan.io/transaction/369826b2f60b443d27966d222fa5bedc6b900326b77a09f15473d77f5ef2871c) which contains an [invalid address](https://cardanoscan.io/address/015bad085057ac10ecc7060f7ac41edd6f63068d8963ef7d86ca58669e5ecf2d283418a60be5a848a2380eb721000da1e0bbf39733134beca4) was created.

Information about this address:
1. Its bech32 representation is `addr1q9d66zzs27kppmx8qc8h43q7m4hkxp5d39377lvxefvxd8j7eukjsdqc5c97t2zg5guqadepqqx6rc9m7wtnxy6tajjvk4a0kze4ljyuvvrpexg5up2sqxj33363v35gtew`
2. Its binary representation is `015bad085057ac10ecc7060f7ac41edd6f63068d8963ef7d86ca58669e5ecf2d283418a60be5a848a2380eb721000da1e0bbf39733134beca4cb57afb0b35fc89c63061c9914e055001a518c7516`
4. It's a base address address of type <keyhash, keyhash> according to its header
5. The address fails to be parsed by [cardano-address](https://input-output-hk.github.io/cardano-addresses/demo/)
6. The complete txbody this tx is contained in is as follows: `a40083825820bf4f8f6287993e17f785c949ef287416ba784198bb0ee12b2c7720cbffecd26f0082582052cf971bee4ba1b4e3b32b762e0bc3f7af83700bc8897eddbd77bd425dd28e8300825820bf4f8f6287993e17f785c949ef287416ba784198bb0ee12b2c7720cbffecd26f01018282584e015bad085057ac10ecc7060f7ac41edd6f63068d8963ef7d86ca58669e5ecf2d283418a60be5a848a2380eb721000da1e0bbf39733134beca4cb57afb0b35fc89c63061c9914e055001a518c75161a035377d082583901f4e9e89cc628b9204fd6e2f4ae3e875aa0591fc2eb45b721520d2d22f4e9e89cc628b9204fd6e2f4ae3e875aa0591fc2eb45b721520d2d221a2c651d70021a00031570031a05f5e100` but it may be good to verify this tx body with some different tool to make sure this isn't an chain indexer issue (it could be that Oura and CardanoScan's indexer both happen to have the same bug and mangle the address info in the same way)
7. If you remove the extra bytes from this address, you end up with the address `addr1q9d66zzs27kppmx8qc8h43q7m4hkxp5d39377lvxefvxd8j7eukjsdqc5c97t2zg5guqadepqqx6rc9m7wtnxy6tajjq6r54x9`

The next node version (1.30.1) doesn't mention this bug being resolved, so I'm not sure if it sill exists or if it was just left out of the changelogs

# What to do?

- Update the ledger spec to document how this address should be handled
- Update cardano-address to parse this address

lehins:
This PR improves performance of Address and RewardAcnt types, but most importantly it makes deserialization more stringent starting with Babbage. Fixes #2729 

Requires https://github.com/input-output-hk/cardano-base/pull/274


eras/babbage/impl/src/Cardano/Ledger/Babbage/TxBody.hs
```diff
@@ -303,29 +310,41 @@ pattern TxOut ::
...
-            SNothing -> case d of
-              NoDatum -> TxOutCompact' a v
-              DatumHash dh -> TxOutCompactDH' a v dh
-              Datum binaryData -> TxOutCompactDatum a v binaryData
-            SJust rs' -> TxOutCompactRefScript a v d rs'
+    TxOut addr vl datum refScript = mkTxOut addr (compactAddr addr) vl datum refScript
 
 {-# COMPLETE TxOut #-}
 
+mkTxOut ::
```
lehins, line 318: Add haddock!

JaredCorduan (approved):
The bulk of the review was done in a call, thanks for all the context @lehins. looks good to me!

shell.nix
```diff
@@ -34,6 +34,7 @@ let
       pkg-config
       hlint
       ormolu.ormolu
+      haskell-language-server
```
JaredCorduan: I need to start making use of this now! :grin:

nc6: This isn't needed! `haskell-language-server` is specified in the tools section below, and is already available!

nc6 (dismissed):
Just the hls thing!

shell.nix
lehins: How come it didn't work for me until I added it?

lehins: I have no idea what I did with my emacs :smile: , but it works now just fine. I must have fixed my config and added this line at the same time. You are right this line is not necessary, will remove it.

PR #2751, opened 2022-04-25
align babbage spec to impl regarding overflow

JaredCorduan:
In the impossible event that there are more transaction outputs in the in the transaction than will fit into a Word16 (which backs the TxIn), we give the collateral return output the index 2^16 - 1.

This behavior is already present in the implementation, see
c601deaccb3d283573dd754c52c6cc0e4e22bad7

This change makes the corresponding change to the formal spec.


WhatisRT (approved):
LGTM

PR #2752, opened 2022-04-25
Removed erroneous prose in the shelley spec

JaredCorduan:
Removed erroneous prose in the Shelley spec regarding fees


WhatisRT (approved):
LGTM!

PR #2757, opened 2022-04-27
Datum validation happens only on deserialization

JaredCorduan:
This PR removes the ledger validation check done on `Datum`s. The correct check is done in the Plutus library, during deserialization.

### DIFF
--- a/libs/cardano-ledger-test/src/Test/Cardano/Ledger/Examples/BabbageFeatures.hs
+++ b/libs/cardano-ledger-test/src/Test/Cardano/Ledger/Examples/BabbageFeatures.hs
@@ -54,6 +54,7 @@ import Cardano.Ledger.Shelley.UTxO (makeWitnessVKey)
 import Cardano.Ledger.TxIn (TxIn (..), txid)
 import Cardano.Ledger.Val (inject)
 import Control.State.Transition.Extended hiding (Assertion)
+import qualified Data.ByteString as BS
 import Data.ByteString.Short (ShortByteString)
 import qualified Data.Compact.SplitMap as SplitMap
 import Data.Default.Class (Default (..))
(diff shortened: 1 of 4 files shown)


libs/cardano-ledger-test/src/Test/Cardano/Ledger/Examples/BabbageFeatures.hs
```diff
@@ -121,8 +122,21 @@ collateralOutput :: Scriptic era => Proof era -> Core.TxOut era
...
+-- the wire is done during deserialization using the Plutus library.
+sixtyFiveBytes :: ByteString
+sixtyFiveBytes =
+  "0123456789"
+    <> "0123456789"
+    <> "0123456789"
+    <> "0123456789"
+    <> "0123456789"
+    <> "0123456789"
+    <> "012345"
```
lehins: It is best to do something like that, because then it is easier to see that there are indeed 65 bytes, to support this claim your implementation has 66 bytes :stuck_out_tongue_winking_eye:
```suggestion
sixtyFiveBytes = BS.pack [1..65]
```

lehins (approved):
One suggestion. Otherwise looks good

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Examples/BabbageFeatures.hs
JaredCorduan: yep, that's a great suggestion! I know that haskell does what a mathematician would expect with that notation, it's a closed interval. Other languages can trick you, I know in python `[0, 1][0:1] == [0]`. I've been bitten by this before, but need to remind myself that I'm in a safe place here :laughing:

JaredCorduan: also thanks for sparing me for committing an off by one error while trying to prevent an off by one error :laughing:

PR #2760, opened 2022-04-29
Fixed MockChain trace failures, Added strictness ann to some Data types.

TimSheard:
Fixed the annoying MockChain trace failures. Paradoxically this was done be adding a 'seq' call to
genRecipientsFrom.  Also added strictness annotations (!) to GenSize, GenState, and MockChainState as well as the predicate faiulres for MOCKCHAIN. Reabased on master and ormolised.

### DIFF
--- a/libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/TxGen.hs
+++ b/libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/TxGen.hs
@@ -79,7 +79,6 @@ import Data.Word (Word16)
 import GHC.Stack
 import Test.Cardano.Ledger.Alonzo.Serialisation.Generators ()
 import Test.Cardano.Ledger.Babbage.Serialisation.Generators ()
-import Test.Cardano.Ledger.Generic.ApplyTx (applyTx)
 import Test.Cardano.Ledger.Generic.Fields hiding (Mint)
 import qualified Test.Cardano.Ledger.Generic.Fields as Generic (TxBodyField (Mint))
 import Test.Cardano.Ledger.Generic.Functions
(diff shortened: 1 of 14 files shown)


libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/TxGen.hs
```diff
@@ -677,27 +677,33 @@ genRecipientsFrom txOuts = do
           v = getTxOutVal reify txout
       genWithChange s txout rs = do
         let !(!addr, !v, !ds) = txoutFields reify txout
-        c <- Coin <$> lift (choose (1, unCoin(coin v)))  -- v could be 0
```
JaredCorduan: I guess this was the bug? if `v` is zero, are we accidentally "partitioning" zero into `{ 1 }`?

lehins: Yes, my original assumption was that there could never be a TxOut with `0` ada in it, which is the case for Cardano mainnet, but turns out in theory with specific PParams, it is a legitimate state, which indeed was triggering this bug in the generator.

So when such TxOut is present in the inputs, this would result in creation a TxOut in a transaction of one more lovelace that there was in the inputs. That is because `choose (1, 0) === choose (0, 1)`.

JaredCorduan: It's such a great feeling for everything to make sense again, great catch! :)

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/TxGen.hs
```diff
@@ -677,15 +677,34 @@ genRecipientsFrom txOuts = do
...
+  if old /= 0 && new == old
+     then pure ans
+     else error ("Recipients have different totals "++show old++"  "++show new++"\nOld\n"++
+                 show(ppList (pcTxOut reify) txOuts)++"\nNew\n"++show(ppList (pcTxOut reify) ans)++
+                 "\noutCount="++show(outCount)++" approxCount="++show(approxCount)++" extra="++show extra++
+                 " avgExtra="++show(avgExtra))
+-}
+
+--  else seq new (pure ans)
+
```
JaredCorduan: dead code?

JaredCorduan (approved):
Looks good to me! There's a bit of dead code, I think. And if you rebase on master, you won't need to remove the field updater pattern matches, I've merged that into master already.

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/TxGen.hs
TimSheard: gone

PR #2761, opened 2022-04-29
Fix inlining for CompactAddr in the FromCBOR instance

lehins:
There was not enough inlining prior to this PR which resulted in a slight but noticeable regression, when comparing to the implementation prior to the fix to address deserialization. TLDR, nothing crazy, just some inlining.

### DIFF
--- a/eras/shelley/test-suite/src/Test/Cardano/Ledger/Shelley/Address/CompactAddr.hs
+++ b/eras/shelley/test-suite/src/Test/Cardano/Ledger/Shelley/Address/CompactAddr.hs
@@ -10,7 +10,13 @@ module Test.Cardano.Ledger.Shelley.Address.CompactAddr where
 
 import Cardano.Binary (serialize')
 import qualified Cardano.Crypto.Hash.Class as Hash
-import Cardano.Ledger.Address (Addr (..), RewardAcnt (..), putVariableLengthWord64, serialiseAddr)
+import Cardano.Ledger.Address
+  ( Addr (..),
+    RewardAcnt (..),
+    deserialiseAddr,
+    putVariableLengthWord64,
+    serialiseAddr,
+  )
 import qualified Cardano.Ledger.CompactAddress as CA
 import Cardano.Ledger.Credential
 import qualified Cardano.Ledger.Crypto as CC (Crypto (ADDRHASH))
@@ -30,8 +36,9 @@ import Test.QuickCheck.Gen (chooseWord64)
 
 propValidateNewDecompact :: forall crypto. CC.Crypto crypto => Addr crypto -> Property
 propValidateNewDecompact addr =
-  let compact = SBS.toShort $ serialiseAddr addr
-      decompactedOld = CA.deserializeShortAddr @crypto compact
+  let bs = serialiseAddr addr
+      compact = SBS.toShort bs
+      decompactedOld = deserialiseAddr @crypto bs
       decompactedNew = CA.decodeAddrShort @crypto compact
    in isJust decompactedOld .&&. decompactedOld === decompactedNew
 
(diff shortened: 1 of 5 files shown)


eras/shelley/test-suite/src/Test/Cardano/Ledger/Shelley/Address/CompactAddr.hs
```diff
@@ -30,8 +36,9 @@ import Test.QuickCheck.Gen (chooseWord64)
 
 propValidateNewDecompact :: forall crypto. CC.Crypto crypto => Addr crypto -> Property
 propValidateNewDecompact addr =
-  let compact = SBS.toShort $ serialiseAddr addr
-      decompactedOld = CA.deserializeShortAddr @crypto compact
```
nc6, line 34: Can you explain the change here? This seems like we're comparing to a different method than before. I am generally worried that we have so many different deserialisers now, it's very difficult to know which to use.

libs/cardano-ledger-core/src/Cardano/Ledger/CompactAddress.hs
```diff
@@ -9,8 +9,6 @@
 module Cardano.Ledger.CompactAddress
   ( compactAddr,
     decompactAddr,
-    decompactAddrLazy,
-    deserializeShortAddr,
```
nc6, line 13: :+1:

libs/cardano-ledger-core/src/Cardano/Ledger/CompactAddress.hs
```diff
@@ -590,12 +619,10 @@ instance CC.Crypto crypto => ToCBOR (CompactAddr crypto) where
...
 instance CC.Crypto crypto => FromCBOR (CompactAddr crypto) where
-  fromCBOR = snd <$> fromCborBackwardsBothAddr
-
--- sbs <- fromCBOR
--- case deserializeShortAddr @crypto sbs of
---   Just _ -> pure $ UnsafeCompactAddr sbs
---   Nothing -> cborError $ DecoderErrorCustom "Addr" "invalid address"
+  fromCBOR = do
+    (_addr, cAddr) <- fromCborBackwardsBothAddr
+    pure cAddr
```
nc6, line 624: Why the change here?

nc6 (commented):
No major objections, just some questions.

eras/shelley/test-suite/src/Test/Cardano/Ledger/Shelley/Address/CompactAddr.hs
lehins: Very good question and a very good observation: there are indeed too many deserializers now. In fact all this recent work paves the way for us remove all the old deserializers, but only after we are past the hardfork.

With regards to what has changed here is `deserialiseAddr` and `deserializeShortAddr` are almost identical, except latter works on `ShortByteString`s, while the former works on `ByteString`s. So, considering that `deserializeShortAddr` was used only for this test, we can just get rid of that function now and replace it with semantically equivalent one

libs/cardano-ledger-core/src/Cardano/Ledger/CompactAddress.hs
lehins: This was mainly an experiment to check if for some reason `fmap snd` is too lazy for the `Decoder` and forcing the tuple would make an improvement. Honestly I don't know if it did, because I was also adding INLINE pragmas as well. As soon as I saw a regression gone, I didn't bother going back and changing it back to `snd <$> fromCborBackwardsBothAddr`, since that would mean at least another hour being wasted. However, if you are really curious I can make an effort and do a db-analyser run with this change reverted.

lehins:
@nc6 Hopefully I've answered your questions. I've restarted the hydra build, since there was one of those recurring false positive test failures, however I have no idea why hydra is reporting CI as failing. Maybe you know you know how to fix it, if not, I'll force push a commit tomorrow.

JaredCorduan (approved):
LGTM!

PR #2763, opened 2022-05-02
Make `lvExtraEntropy` lazy.

nc6:
This field isn't needed in Babbage, and indeed has no sensible value.
But we're in a `StrictData` module, so it's being evaluated. Making it
lazy should solve this.

Where the extra entropy is set, we force it first in order to not
introduce unnecessary thunks.


JaredCorduan (approved):
LGTM! "Lazy Entropy" sounds like a good band name...

PR #2764, opened 2022-05-02
Make `lvExtraEntropy` lazy.

newhoggy:
This field isn't needed in Babbage, and indeed has no sensible value.
    But we're in a `StrictData` module, so it's being evaluated. Making it
    lazy should solve this.


JaredCorduan:
@newhoggy does #2763 not fix the problem?

PR #2766, opened 2022-05-02
fix babbage min utxo calculation

### LINKED ISSUE #2765: Min ada requirement - Babbage
I'm trying to undertand the min ada calculation for a utxo in Babbage.
According to the specs it's:  `(serializedSize of output * coinsPerUTxOWord + 7) / 8`

I see the fomula being used here to validate utxos: https://github.com/input-output-hk/cardano-ledger/blob/88c0fe663720c5ef40f6fd00b9d331de81a6a7cb/eras/babbage/impl/src/Cardano/Ledger/Babbage/Rules/Utxo.hs#L268

However a few lines above, where the output should get serialized, I see only the value field is actually serialized:
https://github.com/input-output-hk/cardano-ledger/blob/88c0fe663720c5ef40f6fd00b9d331de81a6a7cb/eras/babbage/impl/src/Cardano/Ledger/Babbage/Rules/Utxo.hs#L257

I'm not sure if I'm missing something, but this would imply to me, that adding inline datums, datum hashes or script refs, wouldn't cost anything.

Any help appreciated, thanks!

JaredCorduan:
Fix the bug in the Babbage minimum utxo calculation. closes #2765

This commit can be review commit by commit:
* The first commit removes the wildcard pattern matching in updaters. We were silently ignoring updates to the `AdaPerUTxOWord` field.
* The next commit creates a unit test which exhibits the bug described in #2765.
* The next commit fixes the bug. We now correctly count the size of all the bytes in the serialized outputs.
* The last commit swaps out the Alonzo predicate failure for violations of the min-utxo check for a more descriptive failure. In particular, the failure now tells you what the minimum value was for the given offending outputs.

In addition to the unit test added in this PR, we will soon change the value of `AdaPerUTxOWord` in the property tests to something non-zero.

### DIFF
--- a/eras/babbage/impl/src/Cardano/Ledger/Babbage/Rules/Utxo.hs
+++ b/eras/babbage/impl/src/Cardano/Ledger/Babbage/Rules/Utxo.hs
@@ -108,11 +108,16 @@ data BabbageUtxoPred era
   | UnequalCollateralReturn !Coin !Coin
   | DanglingWitnessDataHash !(Set.Set (DataHash (Crypto era)))
   | MalformedScripts !(Set (ScriptHash (Crypto era)))
+  | -- | list of supplied transaction outputs that are too small,
+    -- together with the minimum value for the given output.
+    BabbageOutputTooSmallUTxO
+      ![(Core.TxOut era, Integer)]
 
 deriving instance
   ( Era era,
     Show (UtxoPredicateFailure era),
     Show (PredicateFailure (Core.EraRule "UTXO" era)),
+    Show (Core.TxOut era),
     Show (Core.Script era)
   ) =>
   Show (BabbageUtxoPred era)
@@ -121,6 +126,7 @@ deriving instance
   ( Era era,
     Eq (UtxoPredicateFailure era),
     Eq (PredicateFailure (Core.EraRule "UTXO" era)),
+    Eq (Core.TxOut era),
     Eq (Core.Script era)
   ) =>
   Eq (BabbageUtxoPred era)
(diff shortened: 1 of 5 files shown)


eras/babbage/impl/src/Cardano/Ledger/Babbage/Rules/Utxo.hs
```diff
@@ -108,11 +108,16 @@ data BabbageUtxoPred era
   | UnequalCollateralReturn !Coin !Coin
   | DanglingWitnessDataHash !(Set.Set (DataHash (Crypto era)))
   | MalformedScripts !(Set (ScriptHash (Crypto era)))
+  | -- | list of supplied transaction outputs that are too small,
+    -- together with the minimum value for the given output.
+    BabbageOutputTooSmallUTxO
+      ![(Core.TxOut era, Integer)]
```
TimSheard, line 115: Could we please include the Hash of the TxBody here as well.  It is often impossible to determine which Tx an predicate failure arise from. This would be very useful in may paces, but at least we have the opportunity to do it here.

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Examples/BabbageFeatures.hs
```diff
@@ -258,7 +258,8 @@ defaultPPs =
     MaxTxExUnits $ ExUnits 1000000 1000000,
     MaxBlockExUnits $ ExUnits 1000000 1000000,
     ProtocolVersion $ ProtVer 7 0,
-    CollateralPercentage 100
+    CollateralPercentage 1,
+    AdaPerUTxOWord (Coin 5)
```
TimSheard, line 262: Is it a percentage,or is it a rational (or float) between 0 and 1?

TimSheard (approved):
Looks good to me. I am serious about the adding the hash. We should do that for every new predicate failure we add from now on. Is it too late for us to that all over Babbage?

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Examples/BabbageFeatures.hs
JaredCorduan: it is a number of lovelace

eras/babbage/impl/src/Cardano/Ledger/Babbage/Rules/Utxo.hs
JaredCorduan: that's a bigger refactor than I was looking to do in this PR... at least if we want any kind of consistency... maybe we can make that a stretch goal in a follow up PR? this is helpful for us when testing, but is probably not actually helpful for most users, who get the predicate failure after locally submitting a single Tx.

PR #2767, opened 2022-05-03
Only evaluate Plutus scripts in static checks.

nc6:
Initially, Plutus scripts were evaluated using the following:
```
      evalScripts @era tx sLst
        ?!## ValidationTagMismatch (getField @"isValid" tx)
```

However, as of https://github.com/input-output-hk/cardano-ledger/pull/2386, additional reporting was added to Plutus script
failure, and this was changed to the following:
```
      case evalScripts @era tx sLst of
        Fails sss -> False ?!## ValidationTagMismatch (getField @"isValidating" tx) (pack (Prelude.unlines sss))
        Passes -> pure ()
```
The problem here: `evalScripts` is no longer gated by the `?!##`
operator; it must be evaulated at least to WHNF in order to match the
`Fails` constructor. This means that when reapplying transactions in the
mempool (as well as when replaying blocks), we are always running all
Plutus scripts.

The current semantics for using "labeled" predicates is insufficient to
solve this, since we cannot carry out additional actions (such as
emitting events) inside the predicate. As such, we introduce additional
functionality in the STS system to allow gating sections of rules (which
do not result in a return value) with labels. For the sake of
consistency, the existing `labeledPred` function et al are updated to
make use of this new feature.

The entire call to `evalScripts` is now gated by a `nonStatic` label,
and hence will not be evaulated in any `reapply` scenario.

### DIFF
--- a/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Rules/Utxos.hs
+++ b/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Rules/Utxos.hs
@@ -5,6 +5,7 @@
 {-# LANGUAGE DeriveGeneric #-}
 {-# LANGUAGE DerivingStrategies #-}
 {-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE MultiParamTypeClasses #-}
 {-# LANGUAGE RankNTypes #-}
 {-# LANGUAGE ScopedTypeVariables #-}
@@ -24,7 +25,7 @@ module Cardano.Ledger.Alonzo.Rules.Utxos
     invalidBegin,
     invalidEnd,
     UtxosEvent (..),
-    (?!##),
+    when2Phase,
     ConcreteAlonzo,
     FailureDescription (..),
     scriptFailuresToPredicateFailure,
@@ -81,6 +82,7 @@ import qualified Data.Compact.SplitMap as SplitMap
 import Data.Foldable (toList)
 import Data.List (intercalate)
 import Data.List.NonEmpty (NonEmpty (..))
+import qualified Data.List.NonEmpty as NE
 import qualified Data.Map.Strict as Map
 import Data.Sequence.Strict (StrictSeq)
 import Data.Set (Set)
@@ -192,10 +194,10 @@ scriptsValidateTransition = do
 
   case collectTwoPhaseScriptInputs ei sysSt pp tx utxo of
     Right sLst ->
-      case evalScripts @era (getField @"_protocolVersion" pp) tx sLst of
+      when2Phase $ case evalScripts @era (getField @"_protocolVersion" pp) tx sLst of
         Fails _ps fs ->
-          False
-            ?!## ValidationTagMismatch
+          failBecause $
+            ValidationTagMismatch
               (getField @"isValid" tx)
               (FailedUnexpectedly (scriptFailuresToPredicateFailure fs))
         Passes ps -> tellEvent (SuccessfulPlutusScriptsEvent ps)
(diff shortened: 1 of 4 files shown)


libs/small-steps/src/Control/State/Transition/Extended.hs
```diff
@@ -341,7 +348,7 @@ validateTrans ::
   (e -> PredicateFailure sts) ->
   Validation (NonEmpty e) () ->
   Rule sts ctx ()
-validateTrans t v = liftF $ Predicate [] v t ()
+validateTrans t v = liftF $ Label [] (liftF $ Predicate v t ()) ()
```
lehins: I think this becomes just this:
```suggestion
validateTrans t v = liftF $ Predicate v t ()
```

libs/small-steps/src/Control/State/Transition/Extended.hs
```diff
@@ -314,12 +315,18 @@ data Clause sts (rtype :: RuleType) a where
...
-    [Label] ->
     Validation (NonEmpty e) a ->
     -- Type of failure to return if the predicate fails
     (e -> PredicateFailure sts) ->
     a ->
     Clause sts rtype a
+  -- | Label part of a rule. The interpreter may be configured to only run parts
+  -- of rules governed by (or by the lack of) certain labels.
+  Label ::
+    [Label] ->
```
lehins: Let's change this to `NonEmpty Label`, since a `[Label]` with no labels has no meaning and only adds overhead. Should be an easy change, a la "just follow the type checker".
```suggestion
    NonEmpty Label ->
```

lehins (changes requested):
@nc6 Really good catch! :100:  I suspect this will have tremendous impact on the forger!

I also like the approach, it removes the need for most predicates to have an empty label. Couple of suggestions that should also improve that common case.

libs/small-steps/src/Control/State/Transition/Extended.hs
nc6: Yes, of course, much nicer!

nc6:
@lehins Nice suggestions, I've applied them both :-)

libs/small-steps/src/Control/State/Transition/Extended.hs
```diff
@@ -341,26 +348,28 @@ validateTrans ::
...
 
 -- | Oh noes!
 --
 --   This takes a condition (a boolean expression) and a failure and results in
 --   a clause which will throw that failure if the condition fails.
 (?!) :: Bool -> PredicateFailure sts -> Rule sts ctx ()
-(?!) = labeledPred []
+(?!) cond onFail =
+  liftF $
+    Predicate (if cond then Success () else Failure (() :| [])) (const onFail) ()
```
lehins, line 372: Just one minor optional suggestion, which doesn't change the semantics, but might make it easier to read the code:
```suggestion
    Predicate (if cond then Success () else Failure (onFail :| [])) id ()
```

lehins (approved):
Looks great!

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Rules/Utxos.hs
```diff
@@ -437,10 +442,8 @@ lbl2Phase = "2phase"
 -- | Construct a 2-phase predicate check.
 --
 --   Note that 2-phase predicate checks are by definition static.
-(?!##) :: Bool -> PredicateFailure sts -> Rule sts ctx ()
-(?!##) = labeledPred [lblStatic, lbl2Phase]
-
-infix 1 ?!##
+when2Phase :: Rule sts ctx () -> Rule sts ctx ()
+when2Phase = labeled $ lblStatic NE.:| [lbl2Phase]
```
TimSheard, line 447: How about some comment here, explaining exactly what is going on here, and some discussion about the correct use of this function.   One might also explain what NE.:| does.   It looks like a constructor function?

TimSheard (commented):
Some discussion about how the features are intended to be used, would be useful.

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Rules/Utxos.hs
lehins: @TimSheard It is a type constructor for `NonEmpty`, it comes from `base`: https://downloads.haskell.org/~ghc/8.10.7/docs/html/libraries/base-4.14.3.0/Data-List-NonEmpty.html#t:NonEmpty

PR #2768, opened 2022-05-05
Add missing haddock comments for some Babbage and Alonzo rules

teodanciu:
and turn some existing plain comments into haddock ones


JaredCorduan (approved):
This all looks great, thank you @teodanciu ! I'll note here that this PR is progress toward #2754

teodanciu:
Forced push to add gpg signature

PR #2769, opened 2022-05-05
Refactor old repo url

lehins:
This PR renames all uses of `cardano-ledger-specs` to `cardano-ledger`, since the name of the repository has been renamed a while back.

Also removes unused `plistf` function from `Cardano.Ledger.Babbage.Scripts` module


JaredCorduan (approved):
much appreciated! :raised_hands:

PR #2771, opened 2022-05-06
Fix `ValidateNone` policy.

nc6:
PR #2767 moved to using `labeled` sections to gate validation for more
than just predicates. However, it totally failed to take account of the
fact that `ValidateNone` should turn off all predicates, and hence
accidentally turned them all back on. *facepalm*.

This PR fixes that, and adds a set of tests to ensure that it remains
fixed.


kderme (approved):
I don't fully understand the logic, but this fixes db-sync tests

lehins (approved):
Totally makes sense. Thank you for adding tests for this as well.

PR #2772, opened 2022-05-06
Remove Show for PlutusDebug and disable traceEvent on PlutusScriptResult

lehins:
@TimSheard and I discovered that it is really expensive to `traceEvent` that involves `PlutusDebug`, because showing that type is way too inefficient (naturally due to the `String` type and the size of the type).

This PR simply disables showing the `ScriptResult` in the event log tracing. If this is an unacceptable solution we need discuss a better one. However, adding the huge debug trace for every script even if it is only for profiling is unacceptable, because it affects our ability to argue about performance of running a script, since showing the `PlutusDebug` will dominate the overhead. 

Also this PR intentionally removes the Show instance, to prevent future temptation of relying on it.


JaredCorduan (approved):
fantastic catch @lehins ! I don't think anyone depends on showing the script result in the trace, I think it is just used for gathering statistics.

I'd love to know how we missed this in our earlier profiles (using db-analyze). :thinking:

PR #2773, opened 2022-05-09
Remove `DanglingWitnessDataHash` 

teodanciu:
as it was replaced by  [`missingRequiredDatums` ](https://github.com/input-output-hk/cardano-ledger/blob/1ee17ad1dab70b2a96c32bf3883ed0e3994c5465/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Rules/Utxow.hs#L231) in Alonzo 

I don't know whether it's acceptable to remove the data constructor (how to check if it's actually used), so there are two commits: the first one is just dead code, so it should be safe. The second commit is questionable, but easy to remove if it's problematic.

### DIFF
--- a/eras/babbage/impl/src/Cardano/Ledger/Babbage/Rules/Utxo.hs
+++ b/eras/babbage/impl/src/Cardano/Ledger/Babbage/Rules/Utxo.hs
@@ -14,7 +14,6 @@ module Cardano.Ledger.Babbage.Rules.Utxo where
 
 import Cardano.Binary (FromCBOR (..), ToCBOR (..), serialize)
 import Cardano.Ledger.Address (bootstrapAddressAttrsSize)
-import Cardano.Ledger.Alonzo.Data (DataHash)
 import Cardano.Ledger.Alonzo.Rules.Utxo
   ( UtxoEvent (..),
     UtxoPredicateFailure (..),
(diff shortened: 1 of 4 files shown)


eras/babbage/impl/src/Cardano/Ledger/Babbage/Rules/Utxo.hs
```diff
@@ -471,7 +465,6 @@ instance
       work (FromAlonzoUtxoFail x) = Sum FromAlonzoUtxoFail 1 !> To x
       work (FromAlonzoUtxowFail x) = Sum FromAlonzoUtxowFail 2 !> To x
       work (UnequalCollateralReturn c1 c2) = Sum UnequalCollateralReturn 3 !> To c1 !> To c2
-      work (DanglingWitnessDataHash x) = Sum DanglingWitnessDataHash 4 !> To x
```
JaredCorduan, line 474: :thinking: I'm wondering if it makes sense to leave a comment explaining why we've mysteriously skipped `4`...

JaredCorduan (approved):
Thanks for removing this dead code! LGTM.

(for posterity, note that the code that is being removed here was rendered obsolete by https://github.com/input-output-hk/cardano-ledger/pull/2733)

eras/babbage/impl/src/Cardano/Ledger/Babbage/Rules/Utxo.hs
teodanciu: Yes absolutely , lest someone assumes it's an accident

JaredCorduan: thank you @lehins for pointing out to me that this is Babbage code, so it's totally fine to change the serialization (ie not skip 4)! sorry for the confusion

PR #2775, opened 2022-05-10
Restrict TTL to lie within the stability windown.

nc6:
The `EpochInfo` provided to the ledger by consensus can throw an error
(or return the Left of an Either) when the slot it is asked to translate
lies outside of its predictable window. However, when consensus doesn't
know about any future epochs, it considers its predictable interval to
be unbounded. This makes sense in the context of its original reason for
creation, which is validating blocks, but has the unfortunate
consequence that relying on it in the ledger will allow unbounded
translation of TTL in the final era. This may then change when a new era
is added to the codebase.

This PR adds additional bounds into the ledger to ensure that, in any
case, the TTL should be bounded by the stability window.

Note that we cannot entirely avoid relying on the `EpochInfo` (e.g.
turning the `Left` hand side into an `error`), since it is possible for
the `EpochInfo` to _restrict_ the valid translation interval (e.g. if it
knows about a new epoch where time translation is only possible for half
the number of slots). This potential nondeterminism should hopefully not
be an issue, since it should only depend on the _slot number_, which is
fixed for a given transaction.

Unfortunately, since some transactions with TTLs outside of the usual
window have been allowed on-chain, we have to disable this check in the
Alonzo era. In the future (once we have transitioned to Babbage), we may
wish to replace this check with specific exclusions for the transactions
which exhibit this.


JaredCorduan (approved):
Thanks @nc6 , this looks good to me

JaredCorduan:
@nc6 - after we merge https://github.com/input-output-hk/cardano-ledger/pull/2760 I will rebase this branch on master, I think we've fixed the issue in Tim's branch. The problem, I think, was that the property tests did not always have the correct major protocol version for each era.

JaredCorduan:
ok, we are down to only property test failures in the babbage era. we just need to constrain the generators to not make validity intervals too far out into the future. I can look into that tomorrow.

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Rules/Utxo.hs
```diff
@@ -353,19 +357,31 @@ validateCollateralContainsNonADA bal =
...
+validateOutsideForecast pp ei slotNo sysSt sw tx =
   {-   (_,i_f) := txvldt tx   -}
   case getField @"vldt" (body tx) of
     ValidityInterval _ (SJust ifj)
       | not (nullRedeemers (txrdmrs' $ wits tx)) ->
-          -- ◇ ∉ { txrdmrs tx, i_f } ⇒
-          failureUnless (isRight (epochInfoSlotToUTCTime ei sysSt ifj)) $ OutsideForecast ifj
+          unless (allowOutsideForecastTTL pp) $
+            -- ◇ ∉ { txrdmrs tx, i_f } ⇒
+            if ifj < addSlots sw (succ slotNo)
```
nfrisby, line 379: I don't think `slotNo` is correct here. Specifically, the sound forecast range doesn't begin at the current slot, it begins at the slot of the last applied block. As far as I know, "the slot of the last applied block" is not easily available inside `Test` etc. I have a local prototype that eg adds it to `Globals`, but it's a nuisance.

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/GenState.hs
```diff
@@ -506,6 +506,15 @@ genCredential tag =
         Just (h, _) -> pure h
         Nothing -> genScript reify tag
 
+-- | Generate a transaction body validity interval which is close in proximity
+--  (less than a stability window) from the current slot.
```
nfrisby, line 558: `current slot` is wrong here too, I think.

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Rules/Utxo.hs
```diff
@@ -353,19 +357,31 @@ validateCollateralContainsNonADA bal =
...
   EpochInfo (Either a) ->
+  -- | Current slot number
+  SlotNo ->
   SystemStart ->
+  -- | Stability window, in terms of number of slots
+  Word64 ->
   ValidatedTx era ->
   Test (UtxoPredicateFailure era)
-validateOutsideForecast ei sysSt tx =
+validateOutsideForecast pp ei slotNo sysSt sw tx =
```
nfrisby, line 372: Also: when we disable the check of the validity interval, Plutus still requires we actually provide POSIX times. This currently happens via `collectTwoPhaseScriptInputs` et al, and it uses the `EpochInfo Identity`. And it crashes my custom node when using this patch.

There's a few options that come to mind.

- we either need a Consensus patch that lifts enables arbitrary far ahead translations when appropriate and use that here, when the check has been disabled

- or eg would it be sound to just use a constant, extremely wide/generous "translated" POSIX time? Is this sound for Plutus semantics?

- or maybe we always "translate" the upper bound to the end of the stability window (perhaps unless the upper bound is already within that)? Is this sound for Plutus semantics?

nfrisby (changes requested):
I'm requesting changes because 1) I think "current slot" is the wrong anchor and 2) we need a backup translation

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Rules/Utxo.hs
JaredCorduan: That's good to know, thanks! It would definitely be a more invasive change to get the slot of last applied to this rule (and divergence from the spec, etc). Having it in `Globals` is a bit odd, but certainly easier for the ledger code.

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Rules/Utxo.hs
JaredCorduan: The first option seems the best to me, it is consistent with the current semantics.

Unless I am misunderstanding the second two options, I don't think that they are an option, I do think that they break the Plutus semantics. We have no control what users are doing with the POSIX time inside of a Plutus script, we need to continue to translate the slots to the POSIX time that they are currently being translated to. (eg, imagine a script that succeeds iff the ttl is exactly some particular value).

nfrisby: I tried `min slot maxStableSlot` and that got me farther but then failed. I'm guessing the script had some redundancy checks?

Ah ha! I just realized something: such redundancy checks shouldn't don't risk interfering with the Babbage transition because Babbage doesn't change the forecasts (ie slot or epoch length). That's a bit of a relief.

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Rules/Utxo.hs
nfrisby: The `eraEnd` logic here is what we're trying to emulate, for reference. I still need to do some careful off-by-one checking.

https://github.com/input-output-hk/ouroboros-network/blob/d3eaedd4e4326fd3252b1be660898ef9c1bf5257/ouroboros-consensus/src/Ouroboros/Consensus/HardFork/Combinator/State/Infra.hs#L199-L207

nc6: This feels like something that _should_ be in the ledger rules (and spec), since it does have an effect on transaction validity. Putting something mutable like this in `Globals` seems wrong...

nc6:
I think we no longer want to merge this PR

nc6:
Closing this as superseded. @TimSheard Maybe the test changes you've made here are already in other PRs, otherwise suggest you open new PRs for them?

PR #2776, opened 2022-05-10
Require the caller of `checkLeaderValue` to provide bounds.

nc6:
The input to `checkLeaderValue` is now not guaranteed to be the output
of a VRF evaluation. Consequently, we require the input to explicitly
specify its bounds.

We only provide one way to construct a 'BoundedNatural', which will
throw an error if the bound is violated. This is most appropriate for
our use case, since we "know" the bounds statically, and a violation
indicates a programmer error.


JaredCorduan (approved):
Looks great to me, thanks @nc6 !

PR #2777, opened 2022-05-10
removing dead code

JaredCorduan:
I found two bits of dead code:

* The function `constructValidated` is not used anywhere in ledger, ouroboros-consensus, or cardano-node. It appears to be a helper function that we created for Alonzo before we redid how the `IsValid` tag works with consensus.
* The `UPDN` rule was moved to `cardano-protocol-tpraos`, but somehow the old version still exists in the Shelley directory, and needs to be deleted.


lehins (approved):
Feels great removing dead code :smiley:

nc6 (approved):

PR #2782, opened 2022-05-12
Append non-zero exit code to nix-shell prompt

teodanciu:
to make command failures easier to see

Not sure if anyone else finds this useful - sometimes I ignored by mistake a failed command because the failure was not very obvious in the output. 
This adds the error code to the prompt in case the last command failed, for example: 

If this is not an annoyance for anyone else, we can ignore the PR

### DIFF
--- a/shell.nix
+++ b/shell.nix
@@ -48,6 +48,18 @@ let
     exactDeps = false;
 
     inherit withHoogle;
+    shellHook = ''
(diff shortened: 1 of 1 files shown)


JaredCorduan (approved):
wow, thanks @teodanciu !

nc6:
Looks cool!

PR #2784, opened 2022-05-13
Provide tip slot in the ledger rules

nc6:
This is built atop #2775, and gives an alternate approach to that in nfrisby/CAD-4264-Babbage-crash


nc6:
I think we no longer wish to merge this PR

PR #2785, opened 2022-05-16
Extend the EpochInfo window in the Alonzo era.

### LINKED ISSUE #2788: Alonzo ledger to support arbitrarily far out time translations.
There was a problem with the HFC's support for the time translations (slot to POSIX, for Plutus scripts) for the last known era. Prior to there being a Babbage era, the translation was allowing slots past a stability window. This is being fixed in the consensus layer (ie even the last era will now be bound by one stability window), but we will change the ledger in order to stay compatible with the bug in Alonzo.

nc6:
The `EpochInfo` provided to the ledger by consensus can throw an error
(or return the Left of an Either) when the slot it is asked to translate
lies outside of its predictable window. However, at present this
function has a bug: when consensus doesn't know about any future epochs,
it considers its predictable interval to be unbounded. This may then
change when a new era is added to the codebase, resulting in old
transactions not validating.

The intention is to remove this bug in consensus, such that the
`EpochInfo` is appropriately bounded in all eras. Unfortunately, since
some transactions with TTLs outside of the usual window have been
allowed on-chain, we have to disable this check in the Alonzo era. In
order to do this, we linearly extend the `EpochInfo` based on the
current and previous slot, which we are guaranteed to be able to
translate, since otherwise we could not be processing this slot.

closes #2788

### DIFF
--- a/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Rules/Utxos.hs
+++ b/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Rules/Utxos.hs
@@ -61,6 +61,7 @@ import Cardano.Ledger.Shelley.Rules.Ppup (PPUP, PPUPEnv (..), PpupPredicateFailu
 import Cardano.Ledger.Shelley.Rules.Utxo (UtxoEnv (..), updateUTxOState)
 import Cardano.Ledger.Shelley.UTxO (UTxO (..), balance, totalDeposits)
 import Cardano.Ledger.Val as Val
+import Cardano.Slotting.EpochInfo.Extend (unsafeLinearExtendEpochInfo)
 import Control.Monad.Trans.Reader (asks)
 import Control.State.Transition.Extended
 import Data.ByteString as BS (ByteString)
@@ -178,7 +179,7 @@ scriptsValidateTransition = do
 
   () <- pure $! traceEvent validBegin ()
 
-  case collectTwoPhaseScriptInputs ei sysSt pp tx utxo of
+  case collectTwoPhaseScriptInputs (unsafeLinearExtendEpochInfo slot ei) sysSt pp tx utxo of
     Right sLst ->
       when2Phase $ case evalScripts @era (getField @"_protocolVersion" pp) tx sLst of
         Fails _ps fs ->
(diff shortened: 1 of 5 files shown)


JaredCorduan (approved):
looks good to me!

nfrisby (approved):
Approved. I agree this implements what we discussed 👍

nfrisby (changes requested):
Ach. We need to extend the `EpochInfo` in the other place too. This PR currently only updates the checking logic; the actual translation happens elsewhere when `collectTwoPhaseScriptInputs` calls `txInfo`.

JaredCorduan:
> We need to extend the...

fantastic catch, thanks :pray:

nfrisby (approved):
My previous point was addressed by the force push. Approved again.

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Rules/Utxos.hs
```diff
@@ -178,7 +179,7 @@ scriptsValidateTransition = do
 
   () <- pure $! traceEvent validBegin ()
 
-  case collectTwoPhaseScriptInputs ei sysSt pp tx utxo of
+  case collectTwoPhaseScriptInputs (unsafeLinearExtendEpochInfo slot ei) sysSt pp tx utxo of
```
JaredCorduan, line 182: I think we need to switch on `allowOutsideForecastTTL pp` here as well, so that `unsafeLinearExtendEpochInfo` is only used before major version 7

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Rules/Utxos.hs
```diff
@@ -207,14 +208,14 @@ scriptsNotValidateTransition ::
...
-  TRC (UtxoEnv _ pp _ _, us@(UTxOState utxo _ fees _ _), tx) <- judgmentContext
+  TRC (UtxoEnv slot pp _ _, us@(UTxOState utxo _ fees _ _), tx) <- judgmentContext
   let txb = body tx
   sysSt <- liftSTS $ asks systemStart
   ei <- liftSTS $ asks epochInfo
 
   let !_ = traceEvent invalidBegin ()
 
-  case collectTwoPhaseScriptInputs ei sysSt pp tx utxo of
+  case collectTwoPhaseScriptInputs (unsafeLinearExtendEpochInfo slot ei) sysSt pp tx utxo of
```
JaredCorduan, line 218: same issues as unsafeLinearExtendEpochInfo

PR #2789, opened 2022-05-16
use coinsPerUTxOByte in babbage

JaredCorduan:
Use `coinsPerUTxOByte` in babbage, instead of `coinsPerUTxOWord` from Alonzo.

### DIFF
--- a/eras/babbage/impl/cardano-ledger-babbage.cabal
+++ b/eras/babbage/impl/cardano-ledger-babbage.cabal
@@ -65,7 +65,6 @@ library
     nothunks,
     plutus-ledger-api ^>= 1.0,
     plutus-tx ^>= 1.0,
-    set-algebra,
     small-steps,
     strict-containers,
     text,
--- a/eras/babbage/impl/src/Cardano/Ledger/Babbage/PParams.hs
+++ b/eras/babbage/impl/src/Cardano/Ledger/Babbage/PParams.hs
@@ -114,8 +114,8 @@ data PParams' f era = PParams
     _protocolVersion :: !(HKD f BT.ProtVer),
     -- | Minimum Stake Pool Cost
     _minPoolCost :: !(HKD f Coin),
-    -- | Cost in lovelace per word (8 bytes) of UTxO storage (instead of _minUTxOValue)
-    _coinsPerUTxOWord :: !(HKD f Coin),
+    -- | Cost in lovelace per byte of UTxO storage (instead of _coinsPerUTxOByte)
+    _coinsPerUTxOByte :: !(HKD f Coin),
     -- | Cost models for non-native script languages
     _costmdls :: !(HKD f CostModels),
     -- | Prices of execution units (for non-native script languages)
@@ -161,7 +161,7 @@ instance Era era => ToCBOR (PParams era) where
         _tau = tau',
         _protocolVersion = protocolVersion',
         _minPoolCost = minPoolCost',
-        _coinsPerUTxOWord = coinsPerUTxOWord',
+        _coinsPerUTxOByte = coinsPerUTxOByte',
         _costmdls = costmdls',
         _prices = prices',
         _maxTxExUnits = maxTxExUnits',
@@ -186,7 +186,7 @@ instance Era era => ToCBOR (PParams era) where
             !> To tau'
             !> E toCBORGroup protocolVersion'
             !> To minPoolCost'
-            !> To coinsPerUTxOWord'
+            !> To coinsPerUTxOByte'
             !> To costmdls'
             !> To prices'
             !> To maxTxExUnits'
@@ -214,7 +214,7 @@ instance Era era => FromCBOR (PParams era) where
         <! From -- _tau             :: UnitInterval
         <! D fromCBORGroup -- _protocolVersion :: ProtVer
         <! From -- _minPoolCost     :: Natural
-        <! From -- _coinsPerUTxOWord  :: Coin
+        <! From -- _coinsPerUTxOByte  :: Coin
         <! From -- _costmdls :: CostModel
         <! From -- _prices = prices',
         <! From -- _maxTxExUnits = maxTxExUnits',
@@ -241,7 +241,7 @@ emptyPParams =
       _tau = minBound,
       _protocolVersion = BT.ProtVer 0 0,
       _minPoolCost = mempty,
-      _coinsPerUTxOWord = Coin 0,
+      _coinsPerUTxOByte = Coin 0,
       _costmdls = CostModels mempty,
       _prices = Prices minBound minBound,
       _maxTxExUnits = ExUnits 0 0,
@@ -269,7 +269,7 @@ instance Ord (PParams' StrictMaybe era) where
       <> compare (_tau x) (_tau y)
       <> compare (_protocolVersion x) (_protocolVersion y)
       <> compare (_minPoolCost x) (_minPoolCost y)
-      <> compare (_coinsPerUTxOWord x) (_coinsPerUTxOWord y)
+      <> compare (_coinsPerUTxOByte x) (_coinsPerUTxOByte y)
       <> compare (_costmdls x) (_costmdls y)
       <> compare (_prices x) (_prices y)
       <> compareEx (_maxTxExUnits x) (_maxTxExUnits y)
@@ -317,7 +317,7 @@ encodePParamsUpdate ppup =
     !> omitStrictMaybe 11 (_tau ppup) toCBOR
     !> omitStrictMaybe 14 (_protocolVersion ppup) toCBOR
     !> omitStrictMaybe 16 (_minPoolCost ppup) toCBOR
-    !> omitStrictMaybe 17 (_coinsPerUTxOWord ppup) toCBOR
+    !> omitStrictMaybe 17 (_coinsPerUTxOByte ppup) toCBOR
     !> omitStrictMaybe 18 (_costmdls ppup) toCBOR
     !> omitStrictMaybe 19 (_prices ppup) toCBOR
     !> omitStrictMaybe 20 (_maxTxExUnits ppup) toCBOR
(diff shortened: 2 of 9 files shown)


eras/babbage/impl/src/Cardano/Ledger/Babbage/Translation.hs
```diff
@@ -180,6 +181,19 @@ translateTxOut (Alonzo.TxOut addr value dh) = TxOut addr value d SNothing
...
 translatePParams ::
-  Alonzo.PParams' f (AlonzoEra c) -> PParams' f (BabbageEra c)
-translatePParams (Alonzo.PParams {..}) = PParams {..}
+  Alonzo.PParams (AlonzoEra c) -> PParams (BabbageEra c)
+translatePParams (Alonzo.PParams {_coinsPerUTxOWord = cpuw, ..}) = PParams {_coinsPerUTxOByte = cpub, ..}
+  where
+    cpub = coinsPerUTxOWordToCoinsPerUTxOByte cpuw
+
+translatePParamsUpdate ::
+  Alonzo.PParamsUpdate (AlonzoEra c) -> PParamsUpdate (BabbageEra c)
```
JaredCorduan: I couldn't find a nice way to keep `translatePParams` parametric in `HKD f`, (where `f` is either `Identity` or `StrictMaybe`). Maybe someone has a suggestion?

eras/babbage/impl/cardano-ledger-babbage.cabal
```diff
@@ -65,7 +65,6 @@ library
     nothunks,
     plutus-ledger-api ^>= 1.0,
     plutus-tx ^>= 1.0,
-    set-algebra,
```
teodanciu, line 68: I thought it was only me having problems with this, so I started to comment out ` -Wunused-packages` locally instead, oops :see_no_evil:

eras/babbage/impl/src/Cardano/Ledger/Babbage/Translation.hs
```diff
@@ -168,7 +169,7 @@ instance Crypto c => TranslateEra (BabbageEra c) API.PPUPState where
 
 instance Crypto c => TranslateEra (BabbageEra c) API.ProposedPPUpdates where
   translateEra _ctxt (API.ProposedPPUpdates ppup) =
-    pure $ API.ProposedPPUpdates $ fmap translatePParams ppup
+    pure $ API.ProposedPPUpdates $ fmap translatePParamsUpdate ppup
```
teodanciu: So changing from Word to Byte requires changing from `PParams` to `PParamsUpdate` ? 
Just curious, but If it's too complicated to explain, don't worry, I'm missing tons of context!

teodanciu: Ah maybe it's to indicate the change itself ? :thinking:

teodanciu (commented):
Just curious if these references in test code matter at all? 

`  AdaPerUTxOWord n -> pp {Babbage._coinsPerUTxOWord = n}` in
https://github.com/input-output-hk/cardano-ledger/blob/jc/babbage-use-coins-per-utxo-byte/libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/Updaters.hs#L388

And 
`    (emptyPParams {_coinsPerUTxOWord = Coin 1})`  in 
https://github.com/input-output-hk/cardano-ledger/blob/jc/babbage-use-coins-per-utxo-byte/eras/babbage/test-suite/src/Test/Cardano/Ledger/Babbage/Examples/Consensus.hs#L184

I assume it wouldn't compile if they mattered, but just thought I ask out of curiosity

eras/babbage/impl/cardano-ledger-babbage.cabal
JaredCorduan: yea, I think it's all of us :) your new nix prompt is really nice, btw, highlighting the errors :)

eras/babbage/impl/src/Cardano/Ledger/Babbage/Translation.hs
JaredCorduan: great question!

First notice that `PParams`  and `PParamsUpdate` are type aliases:
* https://github.com/input-output-hk/cardano-ledger/blob/2acff66e84d63a81de904e1c0de70208ff1819ea/eras/babbage/impl/src/Cardano/Ledger/Babbage/PParams.hs#L137
* https://github.com/input-output-hk/cardano-ledger/blob/2acff66e84d63a81de904e1c0de70208ff1819ea/eras/babbage/impl/src/Cardano/Ledger/Babbage/PParams.hs#L83

So this function `translatePParams` was previously handling both `HKD Identity Coin` and `HKD StrictMaybe Coin`. I wasn't able to come up with a clean way of adapting the original function to handle both while trying to manipulate the `Coin`.

JaredCorduan:
> Just curious if these references in test code matter at all?
> 
> ...
 
They probably do, thanks! I'm guilty of not building everything :)

eras/babbage/impl/src/Cardano/Ledger/Babbage/Translation.hs
lehins: Not sure I am a fan of this solution, but it is one that come to mind first. Adding this next to `HKD` definition:
```haskell
class HKDFunctor f where
  hkdMap :: Proxy f -> (a -> b) -> HKD f a -> HKD f b

instance HKDFunctor Identity where
  hkdMap _ f a = f a

instance HKDFunctor Maybe where
  hkdMap _ f = fmap f

instance HKDFunctor StrictMaybe where
  hkdMap _ f = fmap f
```

Then I think implementation can be:
```haskell
translatePParams ::
  forall f c. HKDFunctor f => Alonzo.PParams f (AlonzoEra c) -> PParams' f (BabbageEra c)
translatePParams Alonzo.PParams {_coinsPerUTxOWord = cpuw, ..} = 
  PParams {_coinsPerUTxOByte = cpub, ..}
  where
    cpub = hkdMap (Proxy :: Proxy f) coinsPerUTxOWordToCoinsPerUTxOByte cpuw
```

eras/babbage/impl/src/Cardano/Ledger/Babbage/PParams.hs
```diff
@@ -114,8 +114,8 @@ data PParams' f era = PParams
     _protocolVersion :: !(HKD f BT.ProtVer),
     -- | Minimum Stake Pool Cost
     _minPoolCost :: !(HKD f Coin),
-    -- | Cost in lovelace per word (8 bytes) of UTxO storage (instead of _minUTxOValue)
-    _coinsPerUTxOWord :: !(HKD f Coin),
+    -- | Cost in lovelace per bytes of UTxO storage (instead of _coinsPerUTxOByte)
```
lehins:
```suggestion
    -- | Cost in lovelace per byte of UTxO storage (instead of _coinsPerUTxOByte)
```

eras/babbage/impl/src/Cardano/Ledger/Babbage/Translation.hs
```diff
@@ -180,6 +181,19 @@ translateTxOut (Alonzo.TxOut addr value dh) = TxOut addr value d SNothing
       SNothing -> NoDatum
       SJust d' -> DatumHash d'
 
+-- | A word is 8 bytes, so to convert from coinsPerUTxOWord to coinsPerUTxOByte,
+-- we divide by 8, after first adding 7 so that we are rounding up.
+coinsPerUTxOWordToCoinsPerUTxOByte :: Coin -> Coin
+coinsPerUTxOWordToCoinsPerUTxOByte (Coin c) = Coin $ c + 7 `div` 8
```
lehins: :wink:  For this to be correct, we need some parenthesis. `div` has a custom fixity defined for it, to match the regular division:
```haskell
λ> :i `div`
type Integral :: * -> Constraint
class (Real a, Enum a) => Integral a where
  ...
  div :: a -> a -> a
  ...
  	-- Defined in ‘GHC.Real’
infixl 7 `div`
λ> :i (/)
type Fractional :: * -> Constraint
class Num a => Fractional a where
  (/) :: a -> a -> a
  ...
  	-- Defined in ‘GHC.Real’
infixl 7 /
λ> :i (+)
type Num :: * -> Constraint
class Num a where
  (+) :: a -> a -> a
  ...
  	-- Defined in ‘GHC.Num’
infixl 6 +
```

So, in your implementation division will happen first. Instead it should be:
```suggestion
coinsPerUTxOWordToCoinsPerUTxOByte (Coin c) = Coin $ (c + 7) `div` 8
```

lehins (changes requested):
Overall looks good, except one subtle, but crucial bug :wink:

eras/babbage/impl/src/Cardano/Ledger/Babbage/Translation.hs
JaredCorduan: yes indeed, thank you for catching this!

eras/babbage/impl/src/Cardano/Ledger/Babbage/Translation.hs
JaredCorduan: ah cool! thinks it's worth it? :thinking:

lehins: It might be worth it. It's up to you. It could prove handy again in the future eras.

JaredCorduan: it worked great, I'm happy with it, thanks!

eras/babbage/impl/src/Cardano/Ledger/Babbage/Translation.hs
teodanciu: Thanks  the explanation, makes sense!

JaredCorduan: @lehins solution is cool, though, I've used it now :)

eras/babbage/impl/cardano-ledger-babbage.cabal
teodanciu: Glad it's useful! :tada:

lehins (approved):
:+1:

PR #2790, opened 2022-05-16
Generate each Tx with its own ValidityInterval.

TimSheard:
Updates the Generic generator of transaction used in Property.hs to give each Tx its own ValidityInterval tied to the actual SlotNo the Tx is validated in.


JaredCorduan (approved):
Looks great to me, Tim, thank you!

lehins (approved):
:+1:

PR #2791, opened 2022-05-16
Emit new `TotalAdaPotEvent` after epoch transition

teodanciu:
in order to remove the dependency of `API.Wallet` on `NewEpoch`. This is
necessary in order reverse the dependency and use functions from
`API.Wallet` within `NewEpoch`

### DIFF
--- a/eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/NewEpoch.hs
+++ b/eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/NewEpoch.hs
@@ -16,18 +16,17 @@ module Cardano.Ledger.Shelley.Rules.NewEpoch
     NewEpochPredicateFailure (..),
     NewEpochEvent (..),
     PredicateFailure,
-    calculatePoolDistr,
   )
 where
 
 import Cardano.Ledger.BaseTypes
 import Cardano.Ledger.Coin
-import Cardano.Ledger.Compactible (fromCompact)
 import qualified Cardano.Ledger.Core as Core
 import Cardano.Ledger.Credential (Credential)
 import Cardano.Ledger.Era (Crypto, Era)
 import Cardano.Ledger.Keys (KeyRole (Staking))
-import Cardano.Ledger.PoolDistr (IndividualPoolStake (..), PoolDistr (..))
+import Cardano.Ledger.PoolDistr (PoolDistr (..))
+import Cardano.Ledger.Shelley.API.Wallet (calculatePoolDistr)
 import Cardano.Ledger.Shelley.Constraints (UsesTxOut, UsesValue)
 import Cardano.Ledger.Shelley.EpochBoundary
 import Cardano.Ledger.Shelley.LedgerState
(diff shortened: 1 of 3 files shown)


JaredCorduan (commented):
I think it makes sense, but we can wait and see what the others think.

nc6:
I'm fine with the move, but not so much with reversing the dependency - the API is the public facing component, we shouldn't be using it from the internals.

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/NewEpoch.hs
```diff
@@ -173,9 +170,11 @@ newEpochTransition = do
           tellReward (DeltaRewardEvent (RupdEvent e event))
           updateRewards ans
         SJust (Complete ru') -> updateRewards ru'
+      let adaPots = totalAdaPotsES es'
```
teodanciu, line 173: @JaredCorduan  - I'm guessing here is where the subtlety should kick in but didn't!  I choose this EpochState kinda arbitrarily.
The ticket says "after the epoch transition, but before the application of the first block of the epoch.", but I can't tell which bit of the code is doing that.

JaredCorduan: This is a great spot! This rule is only called by the `TICK` rule, which happens before the transactions in the block are applied.

I'm pretty sure, though, that we should use `es` instead of `es'`, so that we are emitting the ada pots events _before_ the reward update is applied. Do you agree, @kderme ?

kderme: Actually, if not mistaken, I think we want es'''. We want the fees pot to include exactly the fees payed in the previous epoch (and not the fees payed in the new epoch block). The reward pots will be incremented by all rewards payed in the new epoch, by instant rewards and by the deposits refunds, but not affected by withdrawals from the new epoch block.

kderme: Hoepfully what I'm describing exists :)

JaredCorduan: All four of these epoch states, `es`, `es'`, `es''`, and `es'''` are not affected by withdrawals.

* `es` represents the epoch state at the end of the epoch which is ending
* `es'` represents the epoch state after applying the reward update (so it zeros out the fee pot, manipulates the treasury and the reserves, and pays out the rewards)
* `es''` represents the epoch state after applying the MIR certs.
* `es'''` represents the epoch state after reaping the retiring pools and potentially adjusting the protocol parameters (a change in the deposit amounts would affect the reserves, etc).

Does is still sound to you like you want `es'''`? you can think of it as "the state of the world at the beginning of the next epoch, just before applying transactions".

kderme: > `es'` represents the epoch state after applying the reward update (so it zeros out the fee pot, ..

Hm is this the case? I though it subtracts the fees payed on epoch - 2, so the fees pot still has the fees payed on the just passed epoch.

Also `es` is the initial state before the block application right? We already have this ledger state so an event would be unnecessary.

Yes, I think I'd like `es'''`

JaredCorduan: > Hm is this the case? I though it subtracts the fees payed on epoch - 2, so the fees pot still has the fees payed on the just passed epoch.

yep, you are correct, good catch. it does not "zero them out" as I said, it just removes the amount listed in the reward update.

>Yes, I think I'd like es'''

great, thanks!

teodanciu:
Closing, as it's superseded by https://github.com/input-output-hk/cardano-ledger/pull/2797

Issue #2786, opened 2022-05-16
Fix FromJSON instance for CostModels


JaredCorduan:
We made an agreement with the Plutus team that the ledger code would only depend on the _ordering_ of the cost models keys, and that they were free to change the names of the keys, provided that the ordering didn't change. The `FromCBOR` instance of `CostModels`, however, _does_ depending on the keys names.

We need to either translate the old format to the new one (ie translate the changed key names) or change the original Alonzo Genesis JSON file to use a list of integers (assumed to be in the correct order). The first option is unfortunate in that it is potentially confusing for folks to understand the Alonzo Genesis file, and the second option is unfortunate in that it involves changing the Alonzo Genesis.

JaredCorduan:
closed by #2792

PR #2792, opened 2022-05-16
translate legacy PV1 cost model param names

JaredCorduan:
Previously, I told the Plutus team that the ledger does not care about the names of the cost model parameters, only the ordering. I was only considering the CBOR serialization, and completely forgot about JSON.

This PR translates legacy PV1 cost model parameter names in the Alonzo genesis file to the new names. Moreover, the parameter names can change again, provided that the number and the ordering of them does not change.

This is one possible fix, which has the nice property of not needing to update the Alonzo Genesis file. The downside is that maybe it could cause confusion by being quite a silent and invisible translation..

See #2786

### DIFF
--- a/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Genesis.hs
+++ b/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Genesis.hs
@@ -221,7 +221,7 @@ validateCostModel (lang, cmps) = case mkCostModel lang cmps of
   Left err -> fail err
   Right cm -> pure (lang, cm)
 
--- | The keys the Plutus V1 cost models have changed since the Alonzo genesis file was created.
+-- | The keys of the Plutus V1 cost models have changed since the Alonzo genesis file was created.
 -- The number of keys, and the ordering of the keys, however, has not changed.
 -- Therefore we just replace (in order) the new keys for the old ones.
 translateLegacyV1paramNames :: Map Text Integer -> Map Text Integer
--- a/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Genesis.hs
+++ b/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Genesis.hs
@@ -46,11 +46,13 @@ import Data.Map (Map)
 import qualified Data.Map as Map
 import Data.Maybe (mapMaybe)
 import Data.Scientific (fromRationalRepetendLimited)
+import qualified Data.Set as Set
 import Data.Text (Text)
 import Data.Word (Word64)
 import GHC.Generics (Generic)
 import NoThunks.Class (NoThunks)
 import Numeric.Natural (Natural)
+import Plutus.V1.Ledger.Api as PV1 (costModelParamNames)
 import Prelude
 
 data AlonzoGenesis = AlonzoGenesis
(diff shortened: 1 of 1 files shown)


eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Genesis.hs
```diff
@@ -219,11 +221,20 @@ validateCostModel (lang, cmps) = case mkCostModel lang cmps of
   Left err -> fail err
   Right cm -> pure (lang, cm)
 
+-- | The keys the Plutus V1 cost models have changed since the Alonzo genesis file was created.
```
ch1bo:
```suggestion
-- | The keys of the Plutus V1 cost models have changed since the Alonzo genesis file was created.
```

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Genesis.hs
```diff
@@ -219,11 +221,20 @@ validateCostModel (lang, cmps) = case mkCostModel lang cmps of
...
+-- The number of keys, and the ordering of the keys, however, has not changed.
```
ch1bo, line 225: Why are you so sure the ordering of the keys is the same if "the keys of Plutus V1 cost models have changed"? I'm not saying you are wrong here, but this seems to be DANGEROUS! Would spelling out the mapping (of the differences) be an option, albeit verbose?

ch1bo (commented):
Sorry to be pedantic, but the code is not convincing me that the order of the keys did not change, i.e. it's not guaranteed the `PV1.costModelParamNames :: Set Text` results in the same `[Text]` as the keys of `cmps :: Map Text Integer`!?

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Genesis.hs
JaredCorduan: you are right, there is no guarantee, this is entirely an unenforced and unchecked agreement with the Plutus team. (any change in the ordering would almost certainly cause us to not be able to sync the chain, but this is not very satisfactory at all).  The names themselves are a leaked abstraction that we now regret. The idea with this PR was that the Plutus team could continue to change the names as often as they like, provided they did not change the number or the ordering. If we provided the translation, we would have to maintain the translation, and the abstraction is still leaked.

JaredCorduan: ~I can, at the very least, add a unit test that checks that the number and the ordering do not change.~ Actually, this is harder than I realized. The size of the list of names is already checked by our CBOR tests. Checking that the "ordering hasn't changed" isn't possible without understanding what the Plutus evaluator does with parameters.

michaelpj: We're going to replace the textual keys with an opaque enumerable key type with a tightly controlled ordering to make this more explicit.

We didn't expect to have this ordering constraint, hence why it's haphazard currently.

nc6 (approved):
I think this seems reasonable, as a workaround to a problem that lacks a particularly nice solution.

PR #2793, opened 2022-05-16
remove cabal-docspec from github action

JaredCorduan:
Our doc tests should not depend on `oleg.fi`, so I am disabling them to unblock all the current PRs.  We can enable them another way.


lehins (approved):

JaredCorduan:
I am going to merge without seeing the `ci/hydra-eval` go green. If you follow the link, all the tests actually passed, the notification seems broken. And every attempt I've made to restart it is ineffectual.

PR #2795, opened 2022-05-17
Refactor compact map

### LINKED ISSUE #2774: `compact-map` already exists on Hackage
https://hackage.haskell.org/package/compact-map

Found while doing stuff on https://github.com/input-output-hk/haskell.nix/issues/1349

lehins:
This PR takes care of a few related issues:

* Removes no longer utilized `SplitMap` and its underlying implementation `KeyMap` and `HashMap`. It turned out to be great for memory, but not so much for performance.
* Renames `compact-map` to `vector-map`, since that package is only left with compact map implementation backed by key value vectors. Also it turned out `compact-map` is already a taken name on hackage. Fixes #2774 
* Switch all usage of lazy Map to strict: `Data.Map` -> `Data.Map.Strict`. (except in byron codebase)


libs/cardano-data/src/Data/MapExtras.hs
```diff
@@ -126,11 +111,29 @@ disjointMapSetFold accum (Bin _ k v l1 l2) set !ans =
...
+extractKeys sm = Set.foldl' f (sm, Map.empty)
+  where
+    f acc@(without, restrict) k =
+      case Map.lookup k without of
+        Nothing -> acc
+        Just v ->
+          let !without' = Map.delete k without
+              !restrict' = Map.insert k v restrict
+           in (without', restrict')
+{-# INLINEABLE extractKeys #-}
```
TimSheard, line 139: This may have pretty bad complexity, since it visits every node in 'sm' .  Can't we use the property that the set and map are balanced using the same Ord instance, and split them both. If one of the splittings has a left or right side that is empty, we don't have to explore the (left ir right side) of the other?

TimSheard (approved):
95% of this PR is formulaic (replace SplitMap with Map). Everything looks good except the one place where I left a comment. I think we should address this.

libs/cardano-data/src/Data/MapExtras.hs
lehins: Yes we could! Thank you for pointing it out, @TimSheard 

I will go ahead an implement it, but I think I'll better do it in a separate PR, since I would prefer such change to be more visible. Also adding a couple of tests for this would be a good idea.

As far as this PR is concerned this function just moved from one module to another.

TimSheard: OK, just don't forget.

lehins: @TimSheard As promised: https://github.com/input-output-hk/cardano-ledger/pull/2801

Issue #2780, opened 2022-05-11
New Ada Pots event


kderme:
We would like a new epoch boundary event, which exposes the `AdaPots` after the epoch transition, but before the application of the first block of the epoch.

JaredCorduan:
closed by https://github.com/input-output-hk/cardano-ledger/pull/2797

PR #2797, opened 2022-05-17
Emit new TotalAdaPotEvent after epoch transition (gh #2780)

teodanciu:
This is an alternative to https://github.com/input-output-hk/cardano-ledger/pull/2791

It is a solution for: https://github.com/input-output-hk/cardano-ledger/issues/2780

The difference is in how the code is refactored. This PR extracts `AdaPots` in a separate module (outside the API) and uses it from `NewEpoch`, rather than move `calculatePoolDistr` inside `Wallet.API`.

But the question is: do these functions have to be in the API? From what I can tell, there is nothing using them at the moment, but I know little!  

Are there better alternatives to this refactoring?

### DIFF
--- a/eras/shelley/impl/src/Cardano/Ledger/Shelley/API/Wallet.hs
+++ b/eras/shelley/impl/src/Cardano/Ledger/Shelley/API/Wallet.hs
@@ -32,6 +32,10 @@ module Cardano.Ledger.Shelley.API.Wallet
     -- * Transaction helpers
     CLI (..),
     addShelleyKeyWitnesses,
+    -- -- * Ada pots
+    AdaPots (..),
+    totalAdaES,
+    totalAdaPotsES,
   )
 where
 
@@ -68,6 +72,11 @@ import Cardano.Ledger.PoolDistr
     PoolDistr (..),
   )
 import Cardano.Ledger.Shelley (ShelleyEra)
+import Cardano.Ledger.Shelley.AdaPots
+  ( AdaPots (..),
+    totalAdaES,
+    totalAdaPotsES,
+  )
 import qualified Cardano.Ledger.Shelley.EpochBoundary as EB
 import Cardano.Ledger.Shelley.LedgerState
   ( DPState (..),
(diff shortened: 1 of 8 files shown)


eras/shelley/impl/src/Cardano/Ledger/Shelley/API/Wallet.hs
```diff
@@ -32,11 +32,6 @@ module Cardano.Ledger.Shelley.API.Wallet
     -- * Transaction helpers
     CLI (..),
     addShelleyKeyWitnesses,
-
-    -- * Ada Pots
-    AdaPots (..),
-    totalAdaES,
-    totalAdaPotsES,
```
JaredCorduan, line 39: I think this module should re-export these three items.

JaredCorduan, line 39: I think this module should re-export these three items.

JaredCorduan (commented):
Besides re-export the ada pots support in the wallet API, this PR looks good to me!

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/NewEpoch.hs
```diff
@@ -166,6 +169,8 @@ newEpochTransition = do
             -- map is empty (db-sync depends on it).
             tellEvent $ TotalRewardEvent e regRU
             pure es'
+      let adaPots = totalAdaPotsES es
```
kderme: As discussed in https://github.com/input-output-hk/cardano-ledger/pull/2791#discussion_r874619228 I think `es'''` pots better fits what we need.

teodanciu: Yes! Thank you for all the explanation, I'm just about to push a new version with this. 
Thanks!

teodanciu:
Forced push to use `es'''` EpochState. Also exporting AdaPots from Wallet.

teodanciu:
Force-pushed to retrigger build :crossed_fingers:

JaredCorduan (approved):
LGTM!

teodanciu:
Tests are passing on my machine (even after deleting dist-newstyle), so I don't get the failure on CI.  Retrying

PR #2802, opened 2022-05-18
pin cardano-base to current master

JaredCorduan:
Our version of `cardano-base dep` is currently set to [this branch](https://github.com/input-output-hk/cardano-base/pull/277),  instead of master after the merge.

This PR sets `cardano-base` to the current master.

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -41,7 +41,7 @@ test-show-details: streaming
 source-repository-package
   type: git
   location: https://github.com/input-output-hk/cardano-base
-  tag: b1eb0e678b834ef5a6eaea84e75b41a14123331a
+  tag: 631cb6cf1fa01ab346233b610a38b3b4cba6e6ab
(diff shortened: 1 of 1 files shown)


cabal.project
```diff
@@ -41,7 +41,7 @@ test-show-details: streaming
 source-repository-package
   type: git
   location: https://github.com/input-output-hk/cardano-base
-  tag: b1eb0e678b834ef5a6eaea84e75b41a14123331a
+  tag: 631cb6cf1fa01ab346233b610a38b3b4cba6e6ab
```
lehins, line 44: I am not sure if such change is really worth the effort, since `b1eb0e678b834ef5a6eaea84e75b41a14123331a` is part of `cardano-base` `master` branch as much as the merge commit is.

lehins (approved):

cabal.project
JaredCorduan: The effort was minimal, and someone else said that others might be confused. I'm happy either way :shrug:

PR #2803, opened 2022-05-19
Use `Data.Relation` for tests.

lehins:
Module `Data.Relation` is dead code, however instead of removing it completely, I thought parts of it would make a nice small addition to the `set-algebra` test suite. This PR also includes cleanups to the existing `set-algebra` test suite: tackle disabled compilation warnings and apply hlint hints

### DIFF
--- a/libs/cardano-data/cardano-data.cabal
+++ b/libs/cardano-data/cardano-data.cabal
@@ -33,7 +33,6 @@ library
   import:             base, project-config
 
   exposed-modules:     Data.AbstractSize
-                     , Data.Relation
                      , Data.CanonicalMaps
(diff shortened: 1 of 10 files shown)


libs/cardano-data/cardano-data.cabal
```diff
@@ -33,7 +33,6 @@ library
   import:             base, project-config
 
   exposed-modules:     Data.AbstractSize
-                     , Data.Relation
```
lehins, line 36: Doing this, was the main goal of this PR. The rest stuff is extra :smiley:

JaredCorduan (approved):
LGTM!

PR #2801, opened 2022-05-18
Improve implementation of `extractKeys`. Now ~30% faster.

lehins:
* Add tests and benchmarks for `extractKeys`
* Slight improvement to `splitMember` and addition of a better version for `Set`

Some benchmark results for the record:

* `extractKeysFoldSet` is the previous implementation
* `extractKeysNaive m s === (withoutKeys m s, restrictKeys m s)`

```
benchmarking DeleteTxOuts/extractKeys
time                 121.3 ms   (120.6 ms .. 121.9 ms)
                     1.000 R²   (1.000 R² .. 1.000 R²)
mean                 122.1 ms   (121.6 ms .. 123.4 ms)
std dev              1.186 ms   (328.2 μs .. 1.897 ms)
variance introduced by outliers: 11% (moderately inflated)

benchmarking DeleteTxOuts/extractKeysFoldSet
time                 172.7 ms   (172.1 ms .. 173.3 ms)
                     1.000 R²   (1.000 R² .. 1.000 R²)
mean                 173.2 ms   (172.8 ms .. 174.5 ms)
std dev              1.079 ms   (101.6 μs .. 1.573 ms)
variance introduced by outliers: 12% (moderately inflated)

benchmarking DeleteTxOuts/extractKeysNaive
time                 224.5 ms   (221.4 ms .. 228.1 ms)
                     1.000 R²   (1.000 R² .. 1.000 R²)
mean                 223.2 ms   (221.6 ms .. 224.8 ms)
std dev              2.305 ms   (1.383 ms .. 3.325 ms)
variance introduced by outliers: 14% (moderately inflated)
```

PR #2804, opened 2022-05-20
Fix performance of `extractKeys` for very small sets.

lehins:
It turned out that #2801 introduced a slight regression instead of an expected improvement in performance, which was discovered with profiling a db-analyser run.

After looking at profiling report it was immediately obvious to me why the regression actually happened: `extractKeys` is called a lot, but it is always called with a large Map (UTxO) and a very small Set (number of `TxIn`s). Implementation added in #2801 was benchmarked for Sets of size 100 and 100000, which performed quite well. However, majority of times `extractKeys` is called with the Set that has number of elements in single digits. Comparing performance of old and new implementations, confirmed my suspicion: we already had very fast implementation before #2801 

That being said, there is no reason to discard a function that performs well, therefore this PR adopts both implementations for the `extractKeys` functions, making it fast for both large and small inputs alike. This can prove useful in the future if we start getting many transactions with large number of outputs.


JaredCorduan (approved):
Thanks for checking with db-analyser! LGTM

PR #2806, opened 2022-05-20
Switch `AssetName` to be backed by `ShortByteString` and `Show` it in Hex

lehins:
AssetName is limited to 32 bytes, there is no reason to use `ByteString` for it's representation. Also the value is not ascii or not even utf8, so hex is now used for the Show instance

### DIFF
--- a/libs/cardano-ledger-pretty/src/Cardano/Ledger/Pretty/Mary.hs
+++ b/libs/cardano-ledger-pretty/src/Cardano/Ledger/Pretty/Mary.hs
@@ -14,7 +14,7 @@ import Cardano.Ledger.Pretty hiding (ppTxBody)
 import Cardano.Ledger.ShelleyMA.AuxiliaryData
 import Cardano.Ledger.ShelleyMA.Timelocks
 import Cardano.Ledger.ShelleyMA.TxBody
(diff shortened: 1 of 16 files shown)


libs/cardano-ledger-pretty/src/Cardano/Ledger/Pretty/Mary.hs
```diff
@@ -27,7 +27,7 @@ ppPolicyID :: PolicyID crypto -> PDoc
 ppPolicyID (PolicyID sh) = ppScriptHash sh
 
 ppAssetName :: AssetName -> PDoc
-ppAssetName (AssetName bs) = ppLong bs
+ppAssetName = viaShow
```
JaredCorduan, line 30: :clap:

JaredCorduan (approved):
This is great, thanks @lehins !

lehins:
@JaredCorduan There was some more changes added since you last reviewed the PR, but all of those where to make test suites build.

JaredCorduan:
I'm going to re-run the github CI action that fail, it was a classifier for the generators:
```
*** Gave up! Passed only 99 tests; 2 discarded tests:
```

PR #2807, opened 2022-05-21
cosmetic changes to babbage CDDL

JaredCorduan:
I made two small, cosmetic changes to the Babbage CDDL:

* I removed an outdated comment
* I renamed `datum` to `datum_option` (since it can be a hash or a datum)


lehins (approved):
:+1:

PR #2808, opened 2022-05-23
Remove usages of `Cardano.Prelude`

teodanciu:
The only Cardano.Prelude things left are now: 
1) in Byron
2) HeapWords 
3) `panic` from Protolude: https://github.com/protolude/protolude/blob/master/src/Protolude/Panic.hs#L26

`HeapWords` will be addressed with PR for cardano-base (https://github.com/input-output-hk/cardano-base/pull/278) , but I wonder what to do about this `Panic`.  How about redefining it, maybe as well in cardano-base?


eras/shelley/test-suite/src/Test/Cardano/Ledger/Shelley/Serialisation/GoldenUtils.hs
```diff
@@ -25,11 +25,11 @@ import Cardano.Binary
     toCBOR,
   )
 import Cardano.Ledger.Serialization (ToCBORGroup (..))
-import Cardano.Prelude (LByteString)
 import Codec.CBOR.Encoding (Encoding (..), Tokens (..))
 import Codec.CBOR.Term (decodeTerm)
 import Control.Exception (throwIO)
 import Control.Monad (unless)
+import qualified Data.ByteString.Lazy as Lazy (ByteString)
```
lehins: For consistency sake let's use `as BSL`. In fact if you can grep for `as Lazy` and change the few occurrences there to `as BSL` too that would be great.
```suggestion
import qualified Data.ByteString.Lazy as BSL (ByteString)
```

eras/shelley/test-suite/test/Test/Cardano/Ledger/Shelley/Serialisation/Golden/Encoding.hs
```diff
@@ -181,6 +180,7 @@ import Codec.CBOR.Encoding (Encoding (..), Tokens (..))
 import Data.ByteString (ByteString)
 import qualified Data.ByteString.Base16 as B16
 import qualified Data.ByteString.Char8 as BS (pack)
+import qualified Data.ByteString.Lazy as Lazy (ByteString)
```
lehins: Same thing here
```suggestion
import qualified Data.ByteString.Lazy as BSL (ByteString)
```

eras/shelley/test-suite/src/Test/Cardano/Ledger/Shelley/Serialisation/CDDLUtils.hs
```diff
@@ -30,18 +30,20 @@ import Cardano.Ledger.Serialization
...
-    forM_,
-    throwIO,
-  )
 import Codec.CBOR.Read (deserialiseFromBytes)
 import Codec.CBOR.Term (decodeTerm)
+import qualified Control.Exception as PException
+import Control.Monad.Trans as Trans
+  ( MonadIO,
+    liftIO,
+  )
```
lehins: `MonadIO` is part of `base`:
```suggestion
import Control.Monad.IO.Class (MonadIO (liftIO))
```

eras/shelley/test-suite/src/Test/Cardano/Ledger/Shelley/Serialisation/CDDLUtils.hs
```diff
@@ -30,18 +30,20 @@ import Cardano.Ledger.Serialization
...
 import qualified Data.ByteString.Base16.Lazy as Base16
 import qualified Data.ByteString.Lazy as BSL
 import Data.ByteString.Lazy.Char8 as Char8 (lines, unpack)
+import Data.Foldable (forM_)
 import Data.Typeable
+import GHC.Exception
+import GHC.IO.Exception (ExitCode (..))
```
lehins: It's usually better to avoid `GHC.*` imports, unless specific internals unexported otherwise are needed

```suggestion
import Control..Exception
import System.Exit (ExitCode (..))
```

eras/shelley/test-suite/src/Test/Cardano/Ledger/Shelley/Serialisation/CDDLUtils.hs
```diff
@@ -189,3 +191,6 @@ verifyConforming value cddl =
     >>= \case
       Right _ -> pure True
       Left _ -> pure False
+
+throwIO :: (Trans.MonadIO m, Exception e) => e -> m a
+throwIO = liftIO . PException.throwIO
```
lehins: We already depend on a library `unliftio` that provides such lifted function: [`throwIO`](https://www.stackage.org/haddock/nightly-2022-05-20/unliftio-0.2.22.0/UnliftIO-Exception.html#v:throwIO), so there is no need to redefine, we can just import it.

eras/shelley/test-suite/src/Test/TestScenario.hs
```diff
@@ -3,7 +3,11 @@
 
 module Test.TestScenario where
 
-import Cardano.Prelude hiding (Option)
+import Cardano.Prelude
+  ( putTextLn,
+    show,
+  )
```
lehins: There is no real reason to go through `Text` here. Just switch it to regular `putStrLn` and `show` from regular `Prelude`

lehins (changes requested):
I personally don't particularly care about HeapWords, but I'd suggest we don't remove any instances unless we are really sure we want to completely get rid of relying on the type class.

If we still do want to rely on HeapWords, then we will need to split it into a separate package and probably put into `cardano-base`. I suggest we all discuss it in a live call later today.

teodanciu:
Thanks for the great suggestions @lehins. I made a few additional changes and reorganized a little the commits, if you would like to have another look. Thanks!

eras/shelley/test-suite/cardano-ledger-shelley-test.cabal
```diff
@@ -228,7 +225,6 @@ benchmark mainbench
     cardano-crypto-praos,
     cardano-data,
     cardano-ledger-core,
-    cardano-prelude,
```
JaredCorduan, line 231: buh bye! :wave:

JaredCorduan (approved):
great work, thank you for taking this on @teodanciu !

lehins (approved):
:+1:

PR #2814, opened 2022-05-23
Fix Babbage era evaluateMinLovelaceOutput function

### LINKED ISSUE #2812: Babbage era `evaluateMinLovelaceOutput` class method is wrong
The `evaluateMinLovelaceOutput` class method is incorrect in the Babbage era.

It is currently computed as:

https://github.com/input-output-hk/cardano-ledger/blob/ac1c074393074c00c854aa8827aca5bde0f07a23/eras/babbage/impl/src/Cardano/Ledger/Babbage.hs#L195-L196

But `utxoEntrySize out` is incorrect. In Babbage, we base the calculation on the serialized size:

https://github.com/input-output-hk/cardano-ledger/blob/ac1c074393074c00c854aa8827aca5bde0f07a23/eras/babbage/impl/src/Cardano/Ledger/Babbage/Rules/Utxo.hs#L267

JaredCorduan:
The babbage era `evaluateMinLovelaceOutput` should use the serialized bytes, not `utxoEntrySize`.

closes #2812


lehins (approved):
LGTM

PR #2815, opened 2022-05-23
Implement `Sized` abstraction and use it for `OutputTooSmallUTxO` 

lehins:
Current `OutputTooSmallUTxO` is suboptimal for two reasons:

* each TxOu has to be unnecessarily serialized upon validation
* TxOut size computed might be different from size of the TxOut submitted over the wire due to non-canonical serialization, which can prove to be problem.

This PR fixes those deficiencies by recording the size each TxOut occupies in the TxOut and using that value for validation.


JaredCorduan (approved):
This is really slick, I love it. I am **so** glad that we are not re-serializing (and without drastic consequences). I tried an experiment using `ByteSpan`, but didn't think to swap out the type inside the transaction body, and ended up with a ton of mess.

:100:

PR #2816, opened 2022-05-24
update plutus

JaredCorduan:
This brings more expressive errors to cost model parameter validation.

### DIFF
--- a/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Scripts.hs
+++ b/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Scripts.hs
@@ -35,13 +35,15 @@ module Cardano.Ledger.Alonzo.Scripts
     ExUnits',
     Prices (..),
     hashCostModel,
-    isCostModelParamsWellFormed,
+    assertWellFormedCostModelParams,
     decodeCostModelMap,
     decodeCostModel,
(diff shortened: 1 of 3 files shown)


eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Scripts.hs
```diff
@@ -35,13 +35,15 @@ module Cardano.Ledger.Alonzo.Scripts
     ExUnits',
     Prices (..),
     hashCostModel,
-    isCostModelParamsWellFormed,
+    assertWellFormedCostModelParams,
     decodeCostModelMap,
     decodeCostModel,
     CostModels (..),
+    CostModelApplyError (..),
```
JaredCorduan, line 42: this re-export is nice, yea?

lehins: Totally!

lehins (approved):
:+1:

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Scripts.hs
JaredCorduan: great, thanks for confirming :)

PR #2818, opened 2022-05-25
Ensure pure `EpochInfo` is not overused.

nc6:
The `EpochInfo` in `Globals` sits in `Either Text`, but we allow it to
be exposed in a pure `Identity` monad for cases where the information
solicited is guaranteed to be available. However, since this is the
"default" option, at some points it has been misused. As such, we make
three changes:

1. The names have been reversed: `epochInfo` will now give the `Either
   Text` version, whereas `epochInfoPure` will give the `Identity`
   version. The latter has some additional notes describing when it's
   valid to use these.
2. Not all places that previously used `epochInfo` are switched to
   `epochInfoPure`. In particular, we ensure that places that use time
   translation must use the full `EpochInfo (Either Text)`.
3. We rationalise the logic around `txInfo`. We now take the `EpochInfo
   (Either Text)` explicitly and fail inside the same `Left` case as all
   other failures.

Note that this is necessary to avoid `collectTwoPhaseScriptInputs` throwing a pure error, even when the earlier check has failed, since a failing check does not terminate processing and a pure error will be thrown anyway.

### DIFF
--- a/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Tools.hs
+++ b/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Tools.hs
@@ -125,7 +125,7 @@ languagesUsed scripts = Set.fromList $ mapMaybe getLanguage scripts
 --  The execution budgets in the supplied transaction are completely ignored.
 --  The results of 'evaluateTransactionExecutionUnits' are intended to replace them.
 evaluateTransactionExecutionUnits ::
-  forall era m.
+  forall era.
   ( Era era,
     ExtendedUTxO era,
     HasField "inputs" (Core.TxBody era) (Set (TxIn (Crypto era))),
@@ -137,8 +137,7 @@ evaluateTransactionExecutionUnits ::
     HasField "_protocolVersion" (Core.PParams era) ProtVer,
     HasField "datahash" (Core.TxOut era) (StrictMaybe (DataHash (Crypto era))),
     HasField "datum" (Core.TxOut era) (StrictMaybe (Data era)),
-    Core.Script era ~ Script era,
-    Monad m
+    Core.Script era ~ Script era
   ) =>
   Core.PParams era ->
   -- | The transaction.
@@ -146,7 +145,7 @@ evaluateTransactionExecutionUnits ::
   -- | The current UTxO set (or the relevant portion for the transaction).
   UTxO era ->
   -- | The epoch info, used to translate slots to POSIX time for plutus.
-  EpochInfo m ->
+  EpochInfo (Either Text) ->
   -- | The start time of the given block chain.
   SystemStart ->
   -- | The array of cost models, indexed by the supported languages.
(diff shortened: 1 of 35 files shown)


eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Tools.hs
```diff
@@ -155,21 +154,21 @@ evaluateTransactionExecutionUnits ::
...
-        Left transEr -> pure . Left $ BadTranslation transEr
-        Right ctx ->
-          pure . Right $
-            Map.mapWithKey
-              (findAndCount pp (array (PlutusV1, PlutusV2) ctx))
-              (unRedeemers $ getField @"txrdmrs" ws)
-    Just e -> pure . Left $ e
+    Nothing ->
+      let getInfo :: Language -> Either TranslationError (Language, VersionedTxInfo)
+          getInfo lang = ((,) lang <$>) $ txInfo pp lang ei sysS utxo tx
```
lehins: Redundant stuff
```suggestion
          getInfo lang = (,) lang <$> txInfo pp lang ei sysS utxo tx
```

eras/alonzo/test-suite/test/Test/Cardano/Ledger/Alonzo/TxInfo.hs
```diff
@@ -37,8 +37,8 @@ byronAddr = AddrBootstrap (BootstrapAddress aliceByronAddr)
 shelleyAddr :: Addr StandardCrypto
 shelleyAddr = Addr Testnet alicePHK StakeRefNull
 
-ei :: EpochInfo Identity
-ei = fixedEpochInfo (EpochSize 100) (mkSlotLength 1)
+ei :: EpochInfo (Either a)
+ei = hoistEpochInfo (Right . runIdentity) $ fixedEpochInfo (EpochSize 100) (mkSlotLength 1)
```
lehins: No need to hoist, this will unify with `Either a`
```suggestion
ei = fixedEpochInfo (EpochSize 100) (mkSlotLength 1)
```

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Alonzo/Tools.hs
```diff
@@ -265,17 +264,17 @@ updateTxExUnits ::
...
-  case res of
-    Left e -> err (show e)
-    -- rdmrs :: Map RdmrPtr ExUnits
-    Right rdmrs -> (replaceRdmrs proof tx) <$> traverse (failLeft err) rdmrs
+updateTxExUnits proof tx utxo ei ss costmdls err =
+  let res = evaluateTransactionExecutionUnits (pparams proof) tx utxo ei ss costmdls
+   in case res of
+        Left e -> err (show e)
+        -- rdmrs :: Map RdmrPtr ExUnits
+        Right rdmrs -> (replaceRdmrs proof tx) <$> traverse (failLeft err) rdmrs
```
lehins: Might as well fix this comment:
```suggestion
        Right (rdmrs :: Map RdmrPtr ExUnits) -> 
          replaceRdmrs proof tx <$> traverse (failLeft err) rdmrs
```

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Examples/TwoPhaseValidation.hs
```diff
@@ -157,8 +158,10 @@ import Test.Tasty.HUnit (Assertion, testCase, (@?=))
 -- Setup the initial state
 -- =======================
 
-testEpochInfo :: EpochInfo Identity
-testEpochInfo = fixedEpochInfo (EpochSize 100) (mkSlotLength 1)
+testEpochInfo :: EpochInfo (Either Text)
+testEpochInfo =
+  hoistEpochInfo (Right . runIdentity) $
+    fixedEpochInfo (EpochSize 100) (mkSlotLength 1)
```
lehins: Same here:
```suggestion
    fixedEpochInfo (EpochSize 100) (mkSlotLength 1)
```

eras/babbage/test-suite/test/Test/Cardano/Ledger/Babbage/TxInfo.hs
```diff
@@ -44,8 +44,8 @@ byronAddr = AddrBootstrap (BootstrapAddress aliceByronAddr)
 shelleyAddr :: Addr StandardCrypto
 shelleyAddr = Addr Testnet alicePHK StakeRefNull
 
-ei :: EpochInfo Identity
-ei = fixedEpochInfo (EpochSize 100) (mkSlotLength 1)
+ei :: EpochInfo (Either a)
+ei = hoistEpochInfo (Right . runIdentity) $ fixedEpochInfo (EpochSize 100) (mkSlotLength 1)
```
lehins: One more:
```suggestion
ei = fixedEpochInfo (EpochSize 100) (mkSlotLength 1)
```

lehins (approved):

eras/babbage/test-suite/test/Test/Cardano/Ledger/Babbage/TxInfo.hs
```diff
@@ -18,12 +18,12 @@ import Cardano.Ledger.Shelley.UTxO (UTxO (..))
 import Cardano.Ledger.ShelleyMA.Timelocks (ValidityInterval (..))
 import Cardano.Ledger.TxIn (TxIn (..), mkTxInPartial)
 import qualified Cardano.Ledger.Val as Val
-import Cardano.Slotting.EpochInfo (EpochInfo, fixedEpochInfo)
+import Cardano.Slotting.EpochInfo (EpochInfo, fixedEpochInfo, hoistEpochInfo)
```
nc6:
```suggestion
import Cardano.Slotting.EpochInfo (EpochInfo, fixedEpochInfo)
```

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Examples/TwoPhaseValidation.hs
```diff
@@ -109,22 +109,23 @@ import Cardano.Ledger.Shelley.UTxO (makeWitnessVKey)
 import Cardano.Ledger.ShelleyMA.Timelocks (ValidityInterval (..))
 import Cardano.Ledger.TxIn (TxIn (..), txid)
 import Cardano.Ledger.Val (inject, (<+>))
-import Cardano.Slotting.EpochInfo (EpochInfo, fixedEpochInfo)
+import Cardano.Slotting.EpochInfo (EpochInfo, fixedEpochInfo, hoistEpochInfo)
```
nc6:
```suggestion
import Cardano.Slotting.EpochInfo (EpochInfo, fixedEpochInfo)
```

PR #2819, opened 2022-05-25
Add a constant for the estimation of size in the UTxO

WhatisRT:


JaredCorduan (approved):
LGTM!

JaredCorduan:
Once this change is in the implementation as well, it should resolve https://github.com/input-output-hk/cardano-node/issues/3913

WhatisRT:
Alexey just found that it should be `160` instead of `168`, since our estimation was based on an output with a single lovelace, and one that actually contains one Ada would actually have required more than one Ada to be valid.

eras/babbage/formal-spec/utxo.tex
```diff
@@ -173,6 +173,8 @@ \section{UTxO}
   \label{fig:rules:utxo-state-upd}
 \end{figure}
 
+In the UTXO rule, we switch from a manual estimation of the size consumed by $\UTxO$ entries to an estimation using the serialization. However, since the $\TxIn$ used as a key in the $\UTxO$ map is not part of the serialization, we need to account for it manually. By itself it is $40$ bytes, but we add another $128$ bytes of overhead for the in-memory representation of Haskell data.
```
lehins: It should add up to 160, right? :wink: 
```suggestion
In the UTXO rule, we switch from a manual estimation of the size consumed by $\UTxO$ entries to an estimation using the serialization. However, since the $\TxIn$ used as a key in the $\UTxO$ map is not part of the serialization, we need to account for it manually. By itself it is $40$ bytes, but we add another $120$ bytes of overhead for the in-memory representation of Haskell data.
```

eras/babbage/formal-spec/utxo.tex
```diff
@@ -189,7 +191,7 @@ \section{UTxO}
       \\~\\
       \mathsf{adaID}\notin \supp {\fun{mint}~tx} \\~\\
       \forall txout \in \hldiff{\fun{allOuts}~txb}, \\
-      \fun{getValue}~txout \geq \fun{inject}~(\hldiff{\lceil\fun{serSize}~txout * \fun{coinsPerUTxOWord}~pp/8\rceil}) \\~
+      \fun{getValue}~txout \geq \fun{inject}~(\hldiff{\lceil(\fun{serSize}~txout + 160) * \fun{coinsPerUTxOByte}~pp\rceil}) \\~
```
lehins: We no longer need to take the ceiling.
```suggestion
      \fun{getValue}~txout \geq \fun{inject}~(\hldiff{(\fun{serSize}~txout + 160) * \fun{coinsPerUTxOByte}~pp}) \\~
```

lehins (changes requested):
A couple of comments

eras/babbage/formal-spec/utxo.tex
WhatisRT: Ah, I only changed it in one place, thanks!

eras/babbage/formal-spec/utxo.tex
WhatisRT: Right, it's finally in integer :)

lehins (approved):
:+1:

PR #2821, opened 2022-05-25
Add a constant to `babbageMinUTxOValue` computation

lehins:
When using TxOut serialization size for minUTxO value, we did not account for the overhead that comes from TxIn and Map data structure that is used for keeping TxOut in memeory.

This PR fixes this problem by adding a constant of 160 bytes. See comment in the PR for more info.


JaredCorduan (approved):
LGTM!

JaredCorduan:
We expect this PR to resolve https://github.com/input-output-hk/cardano-node/issues/3913

PR #2822, opened 2022-05-25
refactor hashScriptIntegrity

JaredCorduan:
Refactor `hashScriptIntegrity` to take what is needs, `Set LangDepView`, instead of `PParams` and `Set Language`. This is more convenient for re-use in downstream tools, such as the CLI. cc @Jimbo4350 

@WhatisRT - this is a change to the spec, we'll probably want to update it as well


WhatisRT:
The whole script integrity thing is super messy in the spec. AFAICT this is only a syntactic change, so we don't _need_ to change the spec here. I'd rather just sit this one out, and wait until we have the time to figure out how to clean up the mess properly.

nc6 (approved):
LGTM

JaredCorduan:
`ci/hydra:Cardano:cardano-ledger-specs:required` finished, all passed, but isn't reporting somehow. I'm going to merge.

PR #2823, opened 2022-05-26
Fix benchmarks and cleanup

lehins:
There are two major problems with benchmarks that are fixed in this PR:

* We couldn't benchmark applyTx for Alonzo era before because HasTrace instance for `AlonzoLEDGER` was in tests, and thus not in scope. Moved it into the test library, while dumping a whole bunch of debugging dead code
* Wrong PParams were used in the benchmarks, which resulted in STS failures. We didn't see that before because we got lucky with the seed and generated tx actually worked with `emptyPParams`. But a slight change to the seed resulted in all sorts of predicate failures.


JaredCorduan (approved):

JaredCorduan:
`ci/hydra:Cardano:cardano-ledger-specs:required` has actually finished, and all tests have passed. I am going to merge.

PR #2824, opened 2022-05-26
round down for hfc conversion to coinsPerUTxOBytes

JaredCorduan:
Very small PR.

When converting the Alonzo protocol parameter `coinsPerUTxOWord` to the Babbage `coinsPerUTxOBytes` at the time of the hard fork, we have decided to round down instead of up, in hopes that this causes less disruption, for example, to folks who have hard-coded values for token bundles, etc.


JaredCorduan:
The mainnet value is `34488`, which is divisible by 8 (rendering this change useless). I was confused since another network was using `34482`.

lehins (approved):

lehins:
Thanks to @kevinhammond it actually looks like the opposite. `Mainnet` value is `34482` and the `Testnet` value `34488`. I don't understand what was the motivation behind using very slightly different values for `utxoCostPerWord`, but that is what seems to be the case.

Merging the PR as it does take care of the initial idea.

PR #2825, opened 2022-05-30
Alonzo tools cleanup

KtorZ:
- :round_pushpin: **Remove unused 'CorruptCostModel' constructor.**
  
- :round_pushpin: **Remove 'BasicFailure', made redundant by TranslationError/TranslationLogicErrorInput**
    The 'BasicFailure' was initially introduced as an extra sanity check to make
  sure that inputs referenced in the transaction were also present in the
  provided UTxO.

  Now however, this is done regardless in the `txInfoIn` function which is used
  to construct the 'PV1.TxInInfo' for both Alonzo & Babbage transactions:

  ```hs
  case Map.lookup txin mp of
    Nothing -> Left TranslationLogicErrorInput
    Just txout -> case transTxOutAddr txout of
  ```

  Thus, the `TranslationLogicErrorInput` branch from `TranslationError` makes
  the `UnknownInputs` branch redundant, and consequently, `BasicFailure`
  redundant completely. To be 100% equivalent, `TranslationLogicErrorInput`
  should be modified to also contain the missing inputs. I didn't do to that
  extend here because `TranslationError` seems to be use as a kind of _enum_,
  with only 0-ary constructors.

### DIFF
--- a/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Tools.hs
+++ b/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Tools.hs
@@ -6,7 +6,6 @@
 
 module Cardano.Ledger.Alonzo.Tools
   ( evaluateTransactionExecutionUnits,
-    BasicFailure (..),
     ScriptFailure (..),
   )
 where
@@ -52,14 +51,6 @@ import GHC.Records (HasField (..))
 import qualified Plutus.V1.Ledger.Api as PV1
 import qualified Plutus.V2.Ledger.Api as PV2
 
--- | Basic validtion failures that can be returned by 'evaluateTransactionExecutionUnits'.
-data BasicFailure c
-  = -- | The transaction contains inputs that are not present in the UTxO.
-    UnknownTxIns (Set (TxIn c))
-  | -- | The transaction context translation failed
-    BadTranslation TranslationError
-  deriving (Show)
-
 -- | Script failures that can be returned by 'evaluateTransactionExecutionUnits'.
 data ScriptFailure c
   = -- | A redeemer was supplied that does not point to a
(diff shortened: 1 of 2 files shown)


KtorZ:
(note: haven't compiled or run ormolu _yet_; waiting for the automated CI to do it for me).

JaredCorduan (approved):
This is a wonderful clean up, thank you @KtorZ !

The `UnknownTxIns (Set (TxIn c))` was nice since it was more expressive, but we can't change the Alonzo predicate failure serialization now, so perhaps it's worth the loss to get rid of `BasicFailure`.

I think as a "first time contributor" I need to approve you to run ormolu, I'll look into that. There are a few test failures that look pretty straight-forward, let me know if I can help.

KtorZ:
Thanks @JaredCorduan , I'll have another pass on this PR once I am done with other stuff (e.g. upgrading Ogmios' internals).

lehins:
I like that this PR removes a redundant check. However, there is problem. We are not giving the user any information about which TxIn is in fact missing from the UTxO. In `basicValidation` we return `UnknownTxIns` which gives the user an idea what is actually causing the issue. I would normally recommend we add `TxIn` as an argument to the error `TranslationLogicErrorInput TxIn`, but this would break current serialization. If we'd be OK with breaking this serialization, I would go even further and improve others as well, by including the offending values into the errors. I don't think we can change the serialization at this particular moment.

Problem is we are using `TxInfo` translation for both:
* cardano-api, which doesn't care about CBOR and 
* ledger rules, which require CBOR for `TranslationError` for predicate failures.

A potential solution here is to create an extra error type `TxInfoTranslationError` that has all the failure information so cli can report to the user when transaction is being build, while having an extra function of:
```
stripErrorInfo :: TxInfoTranslationError -> TranslationError
```
That converts it back to the old error type.

To sum it up, I am totally fine with merging this PR, but we need to have a task that improves error reporting in the way I described above. I thought I post my thoughts here right now and convert it to a ticket later.

JaredCorduan:
@KtorZ if you rebase onto master, you'll get #2829, which we expect to solve the problem with ormolu not running on this PR (thanks goes to @lehins for helping me out).

JaredCorduan:
The commits became a part of https://github.com/input-output-hk/cardano-ledger/pull/2848

Thanks again for this excellent clean-up @KtorZ !

PR #2827, opened 2022-05-30
Only run phase 1 scripts that are required by `scriptsNeeded`

WhatisRT:
Found by Jamie Harper


JaredCorduan (approved):
Great catch, thank you @jmhrpr for finding this and thank you @WhatisRT for addressing it so quickly!

PR #2832, opened 2022-06-01
Made some progress towards generating cross-epoch traces

Soupstraw:
This PR fixes some problems that appeared when crossing the epoch boundary.
Changes:
 * At the start of a trace we generate a set of pools and delegators that cannot be de-registered or retired (honest pools/delegators). This makes it impossible that the total stake in all of the staking pools at an epoch boundary is zero.
 * Generators now generate pool registration and retirement certificates.

### DIFF
--- a/libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/ApplyTx.hs
+++ b/libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/ApplyTx.hs
@@ -46,7 +46,7 @@ import Test.Cardano.Ledger.Generic.ModelState
     mNewEpochStateZero,
     pcModelNewEpochState,
   )
-import Test.Cardano.Ledger.Generic.PrettyCore (pcTx, pcKeyHash, pcCredential)
+import Test.Cardano.Ledger.Generic.PrettyCore (pcCredential, pcTx)
 import Test.Cardano.Ledger.Generic.Proof hiding (lift)
 
 -- ========================================================================
--- a/libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/ApplyTx.hs
+++ b/libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/ApplyTx.hs
@@ -46,7 +46,7 @@ import Test.Cardano.Ledger.Generic.ModelState
     mNewEpochStateZero,
     pcModelNewEpochState,
   )
-import Test.Cardano.Ledger.Generic.PrettyCore (pcTx)
+import Test.Cardano.Ledger.Generic.PrettyCore (pcTx, pcKeyHash, pcCredential)
 import Test.Cardano.Ledger.Generic.Proof hiding (lift)
 
 -- ========================================================================
@@ -108,7 +108,7 @@ applyCert proof model dcert = case dcert of
       pp = mPParams model
       (keydeposit, _) = keyPoolDeposits proof pp
   (DCertDeleg (DeRegKey x)) -> case Map.lookup x (mRewards model) of
-    Nothing -> error "DeRegKey not in rewards"
+    Nothing -> error . show $ "DeRegKey not in rewards: " <> pcCredential x
     Just (Coin 0) ->
       model
         { mRewards = Map.delete x (mRewards model),
--- a/libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/ApplyTx.hs
+++ b/libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/ApplyTx.hs
@@ -108,7 +108,7 @@ applyCert proof model dcert = case dcert of
       pp = mPParams model
       (keydeposit, _) = keyPoolDeposits proof pp
   (DCertDeleg (DeRegKey x)) -> case Map.lookup x (mRewards model) of
-    Nothing -> error . show $ "DeRegKey not in rewards: " <> pcCredential x
+    Nothing -> error $ "DeRegKey not in rewards: " <> show (pcCredential x)
     Just (Coin 0) ->
       model
         { mRewards = Map.delete x (mRewards model),
--- a/libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/GenState.hs
+++ b/libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/GenState.hs
@@ -13,7 +13,38 @@
 -- | Strategy for Generic Tests
 --   Make the GenState include a Mode of the NewEpochState, modify
 --   the ModelNewEpochState to reflect what we generated.
-module Test.Cardano.Ledger.Generic.GenState where
+module Test.Cardano.Ledger.Generic.GenState 
+  ( GenEnv(..)
+  , GenRS
+  , GenState(..)
+  , GenSize
+  , elementsT  -- TODO move to a utilities module
+  , frequencyT -- TODO move to a utilities module
+  , genCredential
+  , genDatumWithHash
+  , genFreshCredential
+  , genFreshRegCred
+  , genKeyHash
+  , genPool
+  , genPositiveVal
+  , genRewards
+  , genScript
+  , getBlocksizeMax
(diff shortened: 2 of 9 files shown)


libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/GenState.hs
```diff
@@ -350,11 +410,58 @@ getNewPoolTest = do
...
-runGenRS proof gsize action = do
-  genenv <- genGenEnv proof gsize
-  (ans, state, ()) <- runRWST action genenv (emptyGenState proof genenv)
-  pure (ans, state)
+runGenRS proof gsize action =
+  do
+    genenv <- genGenEnv proof gsize
+    let action' = initHonestFields proof >> action
+    (ans, state, ()) <- runRWST action' genenv (emptyGenState proof genenv)
+    pure (ans, state)
```
lehins: For consistency sake and diff reduction
```suggestion
runGenRS proof gsize action = do
  genenv <- genGenEnv proof gsize
  let action' = initHonestFields proof >> action
  (ans, state, ()) <- runRWST action' genenv (emptyGenState proof genenv)
  pure (ans, state)
```

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/GenState.hs
```diff
@@ -142,9 +186,14 @@ data GenState era = GenState
     gsInitialRewards :: !(Map (Credential 'Staking (Crypto era)) Coin),
     gsInitialPoolParams :: !(Map (KeyHash 'StakePool (Crypto era)) (PoolParams (Crypto era))),
     gsInitialPoolDistr :: !(Map (KeyHash 'StakePool (Crypto era)) (IndividualPoolStake (Crypto era))),
-    gsRegKey :: !(Set (Credential 'Staking (Crypto era))),
+    -- Honest fields are stable from initialization to the end of the generation process
```
JaredCorduan: maybe `gsStablePools`, etc, is a better name?

JaredCorduan (approved):
Looks good to me! Thanks @Soupstraw !

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/ApplyTx.hs
```diff
@@ -108,7 +108,7 @@ applyCert proof model dcert = case dcert of
       pp = mPParams model
       (keydeposit, _) = keyPoolDeposits proof pp
   (DCertDeleg (DeRegKey x)) -> case Map.lookup x (mRewards model) of
-    Nothing -> error "DeRegKey not in rewards"
+    Nothing -> error . show $ "DeRegKey not in rewards: " <> pcCredential x
```
lehins:
```suggestion
    Nothing -> error $ "DeRegKey not in rewards: " <> show (pcCredential x)
```

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/GenState.hs
```diff
@@ -430,14 +536,31 @@ genRewards = do
...
+genRetirementHash :: forall era. Reflect era => GenRS era (KeyHash 'StakePool (Crypto era))
+genRetirementHash = do
+  m <- gets $ mPoolParams . gsModel
+  honestKhs <- gets gsHonestPools
+  avoidKey <- gets gsAvoidKey
+  res <- lift . genMapElemWhere m 10 $ \kh _ ->
+    kh `Set.notMember` honestKhs && kh `Set.notMember` avoidKey
+  case res of
+    Just x -> do
+      return $ fst x
```
lehins:
```suggestion
    Just x -> return $ fst x
```

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/GenState.hs
```diff
@@ -475,66 +598,128 @@ genFreshCredential tries0 tag old = go tries0
...
+
+genFreshKeyHash :: Reflect era => Int -> GenRS era (KeyHash kr (Crypto era))
+genFreshKeyHash n
+  | n <= 0 = error "Something very unlikely happened"
+  | otherwise = do
+      avoidKeys <- gets gsAvoidKey
+      kh <- genKeyHash
+      if coerceKeyRole kh `Set.member` avoidKeys
+        then genFreshKeyHash $ n - 1
+        else return kh
```
lehins: It seems that the argument `n` is only to ensure that this function terminates, cause genKeyHash should produce uniformly distributed keys, but there is an impossibly tiny chance of a collision. If my understanding is correct it is better to bind the max number of iterations right into the function:

```suggestion
genFreshKeyHash :: Reflect era => GenRS era (KeyHash kr (Crypto era))
genFreshKeyHash = go 100 -- avoid unlikely chance of generated hash collisions.
  where
    go n
      | n <= 0 = error "Something very unlikely happened"
      | otherwise = do
          avoidKeys <- gets gsAvoidKey
          kh <- genKeyHash
          if coerceKeyRole kh `Set.member` avoidKeys
            then go $ n - 1
            else return kh
```

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/GenState.hs
```diff
@@ -475,66 +598,128 @@ genFreshCredential tries0 tag old = go tries0
...
+    genKeyHash' = do
+      kh <- genKeyHash
+      case tag of
+        Rewrd -> modify $ \st ->
+          st {gsInitialRewards = Map.insert (KeyHashObj kh) (Coin 0) $ gsInitialRewards st}
+        _ -> pure ()
+      return $ coerceKeyRole kh
+    genScript' n
+      | n <= 0 = error "Failed to generate a fresh script hash"
+      | otherwise = do
```
lehins: Same thing here, we don't need the `n` argument. I understand that this is already a local binding, but accepting an argument makes the reader question what does this argument mean and why an arbitrary number of 100 was used.

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/GenState.hs
```diff
@@ -475,66 +598,128 @@ genFreshCredential tries0 tag old = go tries0
...
+
+-- | Use this function to get a new pool that should not be used in the future transactions
+genNewPool :: forall era. Reflect era => GenRS era (KeyHash 'StakePool (Crypto era), PoolParams (Crypto era), IndividualPoolStake (Crypto era))
+genNewPool = do
+  poolId <- genFreshKeyHash 100
+  poolparam <- genPoolParams poolId
+  percent <- lift $ choose (0, 1 :: Float)
+  let stake = IndividualPoolStake @(Crypto era) (toRational percent) (_poolVrf poolparam)
+  modify (\s -> s {gsAvoidKey = Set.insert (coerceKeyRole poolId) $ gsAvoidKey s})
+  pure (poolId, poolparam, stake)
```
lehins: I know you didn't write this, but I can't help but suggest the came case ;) 
```suggestion
  poolParam <- genPoolParams poolId
  percent <- lift $ choose (0, 1 :: Float)
  let stake = IndividualPoolStake @(Crypto era) (toRational percent) (_poolVrf poolParam)
  modify (\s -> s {gsAvoidKey = Set.insert (coerceKeyRole poolId) $ gsAvoidKey s})
  pure (poolId, poolParam, stake)
```

lehins (commented):
Just a few minor comments.

I can't vouch for correctness, since I didn't dive too deep into semantics of the PR, but overall this is looking good.

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/GenState.hs
Soupstraw: Perhaps yes, honest pools makes it sound like there are pools that don't adhere to the protocol, but that's not really the case.

PR #2833, opened 2022-06-01
Update Utxow to only validate phase 1 scripts that are required

teodanciu:
A fix for https://github.com/input-output-hk/cardano-ledger/issues/2826, for which the spec was fixed in https://github.com/input-output-hk/cardano-ledger/pull/2827

### DIFF
--- a/eras/babbage/impl/src/Cardano/Ledger/Babbage/Rules/Utxow.hs
+++ b/eras/babbage/impl/src/Cardano/Ledger/Babbage/Rules/Utxow.hs
@@ -102,15 +102,17 @@ validateFailedBabbageScripts ::
   ) =>
   Core.Tx era ->
   UTxO era ->
+  Set (ScriptHash (Crypto era)) ->
   Test (Shelley.UtxowPredicateFailure era)
-validateFailedBabbageScripts tx utxo =
+validateFailedBabbageScripts tx utxo neededHashes =
   let failedScripts =
         Map.filterWithKey
           ( \hs script ->
-              let one = isNativeScript @era script
+              let zero = hs `Set.member` neededHashes
+                  one = isNativeScript @era script
                   two = hashScript @era script /= hs -- TODO this is probably not needed. Only the script is transmitted on the wire, we compute the hash
                   three = not (validateScript @era script tx)
-                  answer = one && (two || three)
+                  answer = zero && one && (two || three)
                in answer
           )
           (txscripts utxo tx)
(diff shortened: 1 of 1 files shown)


eras/babbage/impl/src/Cardano/Ledger/Babbage/Rules/Utxow.hs
```diff
@@ -102,15 +102,17 @@ validateFailedBabbageScripts ::
...
   UTxO era ->
+  Set (ScriptHash (Crypto era)) ->
   Test (Shelley.UtxowPredicateFailure era)
-validateFailedBabbageScripts tx utxo =
+validateFailedBabbageScripts tx utxo neededHashes =
   let failedScripts =
         Map.filterWithKey
           ( \hs script ->
-              let one = isNativeScript @era script
+              let zero = hs `elem` neededHashes
```
teodanciu: Not sure if it would be better to compute the intersection of the `neededHashes` with the keys instead of this check

WhatisRT:
LGTM! I don't know about whether the test for this should be included in this PR or not though.

eras/babbage/impl/src/Cardano/Ledger/Babbage/Rules/Utxow.hs
WhatisRT: Doesn't make a difference to me, but maybe someone else has an opinion

WhatisRT (approved):

eras/babbage/impl/src/Cardano/Ledger/Babbage/Rules/Utxow.hs
lehins: This is fine, no need to use `intrersection`, however, I'd suggest switching to `Set.member`, instead of `elem`. This way we'll get a compiler error in an unlikely situation where `neededHashes` switched a type from `Set` to something less efficient, eg. list or Sequence, or even worse a `Map (ScriptHash (Crypto era)) (ScriptHash (Crypto era))`, which would silently change semantics of this function.

lehins (approved):
One suggestion, other than that looks good

eras/babbage/impl/src/Cardano/Ledger/Babbage/Rules/Utxow.hs
teodanciu: Sounds good, thanks! will change

teodanciu:
Force-pushed to use `isMember`

JaredCorduan:
I'm late to the party, but it looks great to me! Thank you @teodanciu :raised_hands:

PR #2836, opened 2022-06-01
better error messages

JaredCorduan:
This PR improves two error messages:

* The `MissingScript` error (only used by the CLI helper function) now contains the pointers that it _could_ have resolved.
* The `TranslationLogicErrorRedeemer` predicate failure has been renamed to `RdmrPtrPointsToNothing` and now includes the pointer that could not be resolved. We are only able to get away with this (breaking the serialization) since this failure is only ever called in the Babbage era.


lehins (approved):
:+1:

PR #2841, opened 2022-06-02
clarify cost model comment in CDDL files

JaredCorduan:
Clarify cost model comment in the Alonzo and Babbage CDDL files. The Babbage comment was misleading/wrong.


lehins (approved):

JaredCorduan:
`ci/hydra:Cardano:cardano-ledger:required` finished successfully (with no failures), and is not reporting properly. This seems to be happening more and more frequently... I am going to merge.

PR #2844, opened 2022-06-03
Fix possibility of recursive BabbageUtxo predicate failure

### LINKED ISSUE #2843: Weird recursive type in Babbage's rules
Hi, I just wanted to point out that I found this particular type _awkward_ and potentially dangerous / error-prone:

https://github.com/input-output-hk/cardano-ledger/blob/95d9f20887495a3a6f8d61a675016b888bf5d186/eras/babbage/impl/src/Cardano/Ledger/Babbage/Rules/Utxo.hs#L105-L120

I get that this is embedding predicate failures from Alonzo for re-use, but the hierarchy seems quite wrong / recursive. We can see that the predicate failures for both `"UTXO"` and `"UTXOW"` rules point to the same type:

https://github.com/input-output-hk/cardano-ledger/blob/95d9f20887495a3a6f8d61a675016b888bf5d186/eras/babbage/impl/src/Cardano/Ledger/Babbage/Rules/Utxo.hs#L456

https://github.com/input-output-hk/cardano-ledger/blob/95d9f20887495a3a6f8d61a675016b888bf5d186/eras/babbage/impl/src/Cardano/Ledger/Babbage/Rules/Utxow.hs#L271

---

This makes the `BabbageUtxoPred` a recursive type. Indeed, the second branch contains a `UtxowPredicateFail era` coming from the Alonzo's rules:

https://github.com/input-output-hk/cardano-ledger/blob/95d9f20887495a3a6f8d61a675016b888bf5d186/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Rules/Utxow.hs#L106-L135

This one also wraps era failures from the Shelley era as:

```hs
WrappedShelleyEraFailure !(UtxowPredicateFailure era)
```

Now, looking into [Shelley's UtxowPredicateFailure](https://github.com/input-output-hk/cardano-ledger/blob/1db68a3ec0a2dcb5751004beb22b906162474f23/eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Utxow.hs#L156-L179), we can see that it is also a bigger type itself embedding the following:

https://github.com/input-output-hk/cardano-ledger/blob/1db68a3ec0a2dcb5751004beb22b906162474f23/eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Utxow.hs#L166

Which is where it gets interesting, because we do have the following:

```hs
PredicateFailure (Core.EraRule "UTXO" (BabbageEra crypto)) ~ BabbageUtxoPred (BabbageEra crypto)
```

This makes the `BabbageUtxoPred` a recursive data-structure containing itself. I found it quite weird / dangerous because it is far from obvious at first read and, I am flagging this because that's the first time I see this kind of construction in the ledger rules. I guess it all stems from the fact that the type used as predicate failures for the `UTXO` and `UTXOW` rules are the same. It wouldn't be the case if those types were different. So my intuition is that the first branch of `BabbageUtxoPred` should not be there, but instead, we should have:

```hs
type PredicateFailure (BabbageUTXO era) = UtxoPredicateFailure era 
```

in the era rules machinery for Babbage (where this `UtxoPredicateFailure` corresponds to the Alonzo one).

cc @JaredCorduan

lehins:
This PR fixes #2843 

Problem was in  definition of BabbageUtxoPred. It was an erroneous attempt at reducing duplication by having the same predicate failure type for both UTXO and UTXOW rules. This had two reprocussions:

* Possibility of recursive predicate failure as described in #2843
* Possibility of producing UTXOW related predicate failures from UTXO rules, which makes no sense since UTXO rule is called from UTXOW


JaredCorduan (approved):
Thank you @lehins !

PR #2847, opened 2022-06-06
Ensure Plutus scripts are not executed when other failures already exist

### LINKED ISSUE #2838: Well-formed script failure should prevent running phase 2 validation
We have at least one example where a malformed script both fails the well-formed check, but is still run through the plutus evaluator:

https://github.com/input-output-hk/cardano-ledger/pull/2837/files#diff-e87fb00b6caa153e54d2003b22f4cb8f9d8bfe0e5173b3ebc53bd8031c1842a2R997

lehins:
Fixes #2838 

Expectation was that #2679 has fixed this issue, however there was a tricky oversight that did not fix it for nested rules, but only for the immediate ones. In other words if UTXOS rule `whenFailureFree` it will only react to predicate failure produced by that rule alone and not the one that invoked it, i.e. the UTXO rule. This PR fixes that oversight and fixes the test that discovered this issue.

It also enables this conditional execution for Babbage UTXOS rule only and not in the Alonzo, since this fix does not affect node syncing.


JaredCorduan (approved):
Thank you for resolving this @lehins , it looks great to me!

PR #2852, opened 2022-06-07
reference scripts not always executed

### LINKED ISSUE #2851: Faulty assumption that every reference script is executed
Both `collectTwoPhaseScriptInputs` and `evaluateTransactionExecutionUnits` make the assumption that every reference script inside of every transaction input (both regular and reference) will be executed. The former assumes it for checking for cost models, the latter for gathering transaction contexts.

JaredCorduan:
Two functions in the ledger code were making the faulty assumption that every reference script is executed:

* The function `collectTwoPhaseScriptInputs` was making this faulty assumption when collecting cost models. We would not be able to get away with fixing this were it not for the fact that Alonzo has always had a V1 cost model, and we can depend on it always having one.
* The function `evaluateTransactionExecutionUnits` (which is used by the CLI for building transactions) was making this faulty assumption when collecting needed transaction contexts. This resulted in outputs containing Plutus V1 reference scripts being unspendable by the CLI (even as a regular input, ie not just reference inputs). To fix this problem, I factored out `knownToNotBe1Phase` out of `collectTwoPhaseScriptInputs`.

resolves #2851

### DIFF
--- a/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/PlutusScriptApi.hs
+++ b/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/PlutusScriptApi.hs
@@ -156,21 +156,21 @@ collectTwoPhaseScriptInputs ::
   UTxO era ->
   Either [CollectError (Crypto era)] [(ShortByteString, Language, [Data era], ExUnits, CostModel)]
 collectTwoPhaseScriptInputs ei sysS pp tx utxo =
-  let usedLanguages = [lang | (AlonzoScript.PlutusScript lang _) <- Map.elems scriptsUsed]
+  let usedLanguages = [lang | (_, lang, _) <- needed]
       costModels = unCostModels $ getField @"_costmdls" pp
       missingCMs = [lang | lang <- usedLanguages, lang `Map.notMember` costModels]
    in case missingCMs of
         l : _ -> Left [NoCostModel l]
         _ -> merge (apply costModels) (map redeemer needed) (map getscript needed) (Right [])
   where
-    scriptsUsed = txscripts utxo tx
+    scriptsAvailable = txscripts utxo tx
     txinfo lang = txInfo pp lang ei sysS utxo tx
     needed = mapMaybe knownToNotBe1Phase $ scriptsNeeded utxo tx
     -- The formal spec achieves the same filtering as knownToNotBe1Phase
     -- by use of the (partial) language function, which is not defined
     -- on 1-phase scripts.
     knownToNotBe1Phase (sp, sh) =
-      case sh `Map.lookup` scriptsUsed of
+      case sh `Map.lookup` scriptsAvailable of
         Just (AlonzoScript.PlutusScript lang script) -> Just (sp, lang, script)
         Just (AlonzoScript.TimelockScript _) -> Nothing
         Nothing -> Nothing
--- a/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/PlutusScriptApi.hs
+++ b/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/PlutusScriptApi.hs
@@ -20,6 +20,7 @@ module Cardano.Ledger.Alonzo.PlutusScriptApi
     language,
     CollectError (..),
     collectTwoPhaseScriptInputs,
+    knownToNotBe1Phase,
   )
 where
 
@@ -132,6 +133,22 @@ instance (CC.Crypto crypto) => FromCBOR (CollectError crypto) where
       dec 3 = SumD BadTranslation <! From
       dec n = Invalid n
 
+-- Given a script purpose and a script hash, determine if it is *not*
+-- a simple 1-phase script by looking up the script hash in a mapping
+-- of script hashes to labeled scripts.
+-- If the script is determined to not be a 1-phase script, we return
+-- a triple: the script purpose, the language, and the script bytes.
+--
+-- The formal spec achieves the same filtering as knownToNotBe1Phase
+-- by use of the (partial) language function, which is not defined on 1-phase scripts.
+knownToNotBe1Phase ::
+  Map.Map (ScriptHash (Crypto era)) (AlonzoScript.Script era) ->
+  (ScriptPurpose (Crypto era), ScriptHash (Crypto era)) ->
+  Maybe (ScriptPurpose (Crypto era), Language, ShortByteString)
+knownToNotBe1Phase scriptsAvailable (sp, sh) = do
+  AlonzoScript.PlutusScript lang script <- sh `Map.lookup` scriptsAvailable
+  Just (sp, lang, script)
+
 -- | Collect the inputs for twophase scripts. If any script can't find ist data return
 --     a list of CollectError, if all goes well return a list of quadruples with the inputs.
 --     Previous PredicateFailure tests should ensure we find Data for every script, BUT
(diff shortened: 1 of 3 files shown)


eras/alonzo/impl/src/Cardano/Ledger/Alonzo/PlutusScriptApi.hs
```diff
@@ -156,24 +175,16 @@ collectTwoPhaseScriptInputs ::
...
-    needed = mapMaybe knownToNotBe1Phase $ scriptsNeeded utxo tx
-    -- The formal spec achieves the same filtering as knownToNotBe1Phase
-    -- by use of the (partial) language function, which is not defined
-    -- on 1-phase scripts.
-    knownToNotBe1Phase (sp, sh) =
-      case sh `Map.lookup` scriptsUsed of
-        Just (AlonzoScript.PlutusScript lang script) -> Just (sp, lang, script)
-        Just (AlonzoScript.TimelockScript _) -> Nothing
-        Nothing -> Nothing
+    needed = mapMaybe (knownToNotBe1Phase scriptsAvailable) $ scriptsNeeded utxo tx
```
Jimbo4350: Instead of `needed` what about `onlyPlutusScripts`? For me `needed` does not immediately tell me about the kind of scripts needed or what `needed` is. I would have to dig further into how `needed` was created.

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Tools.hs
```diff
@@ -175,15 +169,16 @@ evaluateTransactionExecutionUnits pp tx utxo ei sysS costModels = do
     txb = getField @"body" tx
     ws = getField @"wits" tx
     dats = unTxDats $ getField @"txdats" ws
-    scripts = txscripts utxo tx
+    scriptsAvailable = txscripts utxo tx
     needed = scriptsNeeded utxo tx
+    nonPhase1Needed = mapMaybe (knownToNotBe1Phase scriptsAvailable) needed
```
Jimbo4350: Same thing here `onlyPlutusScripts`. `nonPhase1Needed` is confusing to me.

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/PlutusScriptApi.hs
JaredCorduan: yea, that's better. is `neededAndConfirmedToBePlutus` too long? :grimacing:

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Tools.hs
JaredCorduan: I was trying to stress that we might not have _all_ the plutus scripts. we've just gathered the ones we can, based on what we know. `onlyPlutusScripts` is mathematically correct, but maybe elides the presence of sneaky scripts we can't identify?

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/PlutusScriptApi.hs
Jimbo4350: That's fine as well

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Tools.hs
Jimbo4350: > sneaky scripts we can't identify

Which sneaky scripts? Doesn't `txscripts` get all the possible scripts our transaction is using?

JaredCorduan: yep, `txscripts` gathers all the scripts that we have available for executing (both simple and Plutus). Some of which we don't want to execute, because they are not actually needed. The ones we need are identified by a collection of hashes, namely `needed`. In order to figure out which of the needed ones are Plutus scripts, we look them up in `scriptsAvailable`.  The sneaky part is that there could be a script hash that corresponds to a script that we need, but that we cannot identify as being a Plutus script since it is not in `scriptsAvailable`.

JaredCorduan: I've gone with `neededAndConfirmedToBePlutus` for this variable and the one above that we talked about.

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/PlutusScriptApi.hs
```diff
@@ -132,6 +133,24 @@ instance (CC.Crypto crypto) => FromCBOR (CollectError crypto) where
       dec 3 = SumD BadTranslation <! From
       dec n = Invalid n
 
+-- Given an script purpose and a script hash, determine if it is *not*
```
lehins:
```suggestion
-- Given a script purpose and a script hash, determine if it is *not*
```

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/PlutusScriptApi.hs
```diff
@@ -132,6 +133,24 @@ instance (CC.Crypto crypto) => FromCBOR (CollectError crypto) where
...
+knownToNotBe1Phase ::
+  Map.Map (ScriptHash (Crypto era)) (AlonzoScript.Script era) ->
+  (ScriptPurpose (Crypto era), ScriptHash (Crypto era)) ->
+  Maybe (ScriptPurpose (Crypto era), Language, ShortByteString)
+knownToNotBe1Phase scriptsAvailable (sp, sh) =
+  case sh `Map.lookup` scriptsAvailable of
+    Just (AlonzoScript.PlutusScript lang script) -> Just (sp, lang, script)
+    Just (AlonzoScript.TimelockScript _) -> Nothing
+    Nothing -> Nothing
+
```
lehins: Just a suggestion:
```suggestion
knownToNotBe1Phase scriptsAvailable (sp, sh) = do
  AlonzoScript.PlutusScript lang script <- sh `Map.lookup` scriptsAvailable
  Just (sp, lang, script)

```

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/PlutusScriptApi.hs
```diff
@@ -156,24 +175,22 @@ collectTwoPhaseScriptInputs ::
...
   Either [CollectError (Crypto era)] [(ShortByteString, Language, [Data era], ExUnits, CostModel)]
 collectTwoPhaseScriptInputs ei sysS pp tx utxo =
-  let usedLanguages = [lang | (AlonzoScript.PlutusScript lang _) <- Map.elems scriptsUsed]
+  let usedLanguages = [lang | (_, lang, _) <- neededAndConfirmedToBePlutus]
       costModels = unCostModels $ getField @"_costmdls" pp
       missingCMs = [lang | lang <- usedLanguages, lang `Map.notMember` costModels]
    in case missingCMs of
         l : _ -> Left [NoCostModel l]
-        _ -> merge (apply costModels) (map redeemer needed) (map getscript needed) (Right [])
+        _ ->
```
lehins: I just realized the whole thing should be made more efficient:
```haskell
  let usedLanguages = Set.fromList [lang | (_, lang, _) <- neededAndConfirmedToBePlutus]
      costModels = unCostModels $ getField @"_costmdls" pp
      missingCMs = Set.filter (`Map.notMember` costModels) usedLanguages
   in case Set.lookupMin missingCMs of
        Just l -> Left [NoCostModel l]
        Nothing -> ...
```

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Tools.hs
```diff
@@ -162,7 +156,7 @@ evaluateTransactionExecutionUnits pp tx utxo ei sysS costModels = do
     Nothing ->
       let getInfo :: Language -> Either TranslationError (Language, VersionedTxInfo)
           getInfo lang = (,) lang <$> txInfo pp lang ei sysS utxo tx
-          txInfos = map getInfo (Set.toList $ languagesUsed (Map.elems scripts))
+          txInfos = [getInfo lang | (_, lang, _) <- neededAndConfirmedToBePlutus]
```
lehins: This will force `getInfo` to be called as many times as there are scripts, instead of as many times as there are languages. In order to fix this we need to nub it:
```suggestion
          txInfos = map getInfo $ Set.toList $
                      Set.fromList [lang | (_, lang, _) <- neededAndConfirmedToBePlutus]
```

lehins (changes requested):
There are couple of suggestions that fix performance. Otherwise looks good.

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/PlutusScriptApi.hs
JaredCorduan: much nicer!

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Tools.hs
JaredCorduan: fantastic!

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/PlutusScriptApi.hs
JaredCorduan: thanks!

JaredCorduan: done

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Tools.hs
```diff
@@ -129,24 +123,26 @@ evaluateTransactionExecutionUnits pp tx utxo ei sysS costModels =
   let getInfo :: Language -> Either (TranslationError (Crypto era)) (Language, VersionedTxInfo)
       getInfo lang = (,) lang <$> txInfo pp lang ei sysS utxo tx
    in do
-        ctx <- mapM getInfo (Set.toList $ languagesUsed (Map.elems scripts))
+        ctx <- mapM getInfo languagesUsed
         pure $
           Map.mapWithKey
-            (findAndCount pp (array (PlutusV1, PlutusV2) ctx))
+            (findAndCount pp (array (minBound :: Language, maxBound) ctx))
```
lehins: I like that. I thought about suggesting it too ;)

lehins (approved):
Looks great!

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Tools.hs
lehins: This actually won't work, because if `ctx` is missing a `Language` this will be bottom.

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Tools.hs
```diff
@@ -54,40 +54,40 @@ import qualified Plutus.V1.Ledger.Api as PV1
 import qualified Plutus.V2.Ledger.Api as PV2
 
 -- | Script failures that can be returned by 'evaluateTransactionExecutionUnits'.
-data ScriptFailure c
+data TransactionScriptFailure c
```
JaredCorduan, line 57: :tada:

PR #2855, opened 2022-06-08
Make Cost Model parameters strict

### LINKED ISSUE #2853: CostModel data definition has non-strict fields and wrong NoThunks instance
I ran the `db-analyser` `noThunks` check and found one thunk, here:

https://github.com/input-output-hk/cardano-ledger/blob/51d97a7faeec32f3825a21d775ca70b6a05fd092/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Scripts.hs#L206

I don't think this is cause for alarm, since it's not involved in any iteration or risks retaining much, as far as I anticipate. However, we should fix it, if only so the `noThunks` test passes.

1) Would it be fine to mark the parameters `Map` and the `EvaluationContext` as strict?

2) This instance need not use `InspectHeap`: you should instead `deriving (Generic)` on `CostModel` and just write `instance NoThunks CostModel`.

https://github.com/input-output-hk/cardano-ledger/blob/51d97a7faeec32f3825a21d775ca70b6a05fd092/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Scripts.hs#L241

3) Moreover, if the `Map` and/or `EvaluationContext` thunks are intentional, then we'll need to handwrite the `wNoThunks` in that instance to not check those arguments for thunks.

JaredCorduan:
resolves  #2853

This PR makes all the `CostModel` fields strict. It also changes the `NoThunks` instance: instead of deriving it by hand, we add a `Generic` instance to `CostModel` and derive `NoThunks` automatically.

Many thanks to @nfrisby for spotting this and doing all the hard work to spell out how to fix it!

### DIFF
--- a/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Scripts.hs
+++ b/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Scripts.hs
@@ -203,7 +203,8 @@ pointWiseExUnits oper (ExUnits m1 s1) (ExUnits m2 s2) = (m1 `oper` m2) && (s1 `o
 -- cost model parameters (ie the `Map` `Text` `Integer`) and that
 -- this type uses the smart constructor `mkCostModel`
 -- to hide the evaluation context.
-data CostModel = CostModel !Language (Map Text Integer) PV1.EvaluationContext
+data CostModel = CostModel !Language !(Map Text Integer) !PV1.EvaluationContext
+  deriving (Generic)
 
 -- | Note that this Eq instance ignores the evaluation context, which is
 -- entirely dependent on the cost model parameters and is guarded by the
@@ -233,12 +234,16 @@ instance SafeToHash CostModel where
 
 instance HashWithCrypto CostModel CostModel
 
+-- | It would be preferable to have a proper NoThunks instance for
+-- EvaluationContext in the Plutus repository, but we can use this
+-- workaround in the meantime.
+-- See https://github.com/input-output-hk/plutus/issues/4687
 deriving via
   InspectHeapNamed "PV1.EvaluationContext" PV1.EvaluationContext
   instance
     NoThunks PV1.EvaluationContext
 
-deriving via InspectHeapNamed "CostModel" CostModel instance NoThunks CostModel
+instance NoThunks CostModel
 
 instance NFData CostModel where
   rnf (CostModel lang cm ectx) = lang `deepseq` cm `deepseq` rnf ectx


eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Scripts.hs
```diff
@@ -238,7 +239,7 @@ deriving via
   instance
     NoThunks PV1.EvaluationContext
```
JaredCorduan, line 244: @nfrisby Is this hand-written instance problematic?

nfrisby: `EvaluationContext` is declared in the `plutus` repo. It does have a `Generic` instance, so you could again drop the `InspectHeap` usage here.

The overall intuition is that `InspectHeap` is an escape hatch for use when the `NoThunks` instances get rough/inconvenient. See https://hackage.haskell.org/package/nothunks-0.1.3/docs/NoThunks-Class.html#t:InspectHeap

I'm not familiar with all the types that occur within the `EvaluationContext` -- maybe one of them deep inside it does require `InspectHeap`? (Perhaps `BuiltinsRuntime`?)

Ideally, the `plutus` repo would be declaring the `NoThunks` instances for their data types. Using `InspectHeap` as a stop-gap measure until the instances are upstreamed doesn't seem crazy to me.

JaredCorduan: Many thanks for the explanation. When I first added the `EvaluationContext` to the cost models, I was not able to automatically derive these instances, but I've since forgotten the details (" should have left a trail of comments!). I'll go back with renewed confidence!

JaredCorduan: ok, here is the error that I encounter when trying to make a `NoThunks` instance automatically with `instance NoThunks PV1.EvaluationContext`:

```
src/Cardano/Ledger/Alonzo/Scripts.hs:237:10: error:
    * No instance for (NoThunks
                         (PlutusCore.Evaluation.Machine.MachineParameters.MachineParameters
                            plutus-core-1.0.0.0:UntypedPlutusCore.Evaluation.Machine.Cek.CekMachineCosts.CekMachineCosts
                            UntypedPlutusCore.Evaluation.Machine.Cek.Internal.CekValue
                            plutus-core-1.0.0.0:PlutusCore.Default.Universe.DefaultUni
                            plutus-core-1.0.0.0:PlutusCore.Default.Builtins.DefaultFun))
        arising from a use of `NoThunks.Class.$dmwNoThunks'
    * In the expression:
        NoThunks.Class.$dmwNoThunks @(EvaluationContext)
      In an equation for `wNoThunks':
          wNoThunks = NoThunks.Class.$dmwNoThunks @(EvaluationContext)
      In the instance declaration for `NoThunks EvaluationContext'
    |
237 | instance NoThunks PV1.EvaluationContext
    |          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

So it looks like `MachineParameters` is the culprit. I'll make an issue in the Plutus repo, and leave a TODO in the ledger code.

JaredCorduan: I'll go with the workaround for `EvaluationContext` for now, and we can remove it after https://github.com/input-output-hk/plutus/issues/4687 is resolved.

lehins (approved):

jasagredo:
I think the problem is not completely solved. I was checking thunks in a separate branch, came across the same Thunk as @nfrisby , fixed it in the same way this PR does, and now the thunk moved to `PV1.EvaluationContext`:
```
cardano-node: Invariant violation: ThunkInfo {thunkContext = ["...","PV1.EvaluationContext","CostModel","Map","AlonzoGenesis","ShelleyLedgerConfig","ShelleyPartialLedgerConfig","fst","NP","snd","NP","snd","NP","snd","NP","snd","PerEraLedgerConfig [ByronBlock,ShelleyBlock (TPraos StandardCrypto) (ShelleyEra StandardCrypto),ShelleyBlock (TPraos StandardCrypto) (ShelleyMAEra 'Allegra StandardCrypto),ShelleyBlock (TPraos StandardCrypto) (ShelleyMAEra 'Mary StandardCrypto),ShelleyBlock (TPraos StandardCrypto) (AlonzoEra StandardCrypto),ShelleyBlock (Praos StandardCrypto) (BabbageEra StandardCrypto)]","HardForkLedgerConfig","TopLevelConfig","LgrDB","ChainDbEnv m HardForkBlock (': * ByronBlock (': * (ShelleyBlock (TPraos StandardCrypto) (ShelleyEra StandardCrypto)) (': * (ShelleyBlock (TPraos StandardCrypto) (ShelleyMAEra 'Allegra StandardCrypto)) (': * (ShelleyBlock (TPraos StandardCrypto) (ShelleyMAEra 'Mary StandardCrypto)) (': * (ShelleyBlock (TPraos StandardCrypto) (AlonzoEra StandardCrypto)) (': * (ShelleyBlock (Praos StandardCrypto) (BabbageEra StandardCrypto)) ('[] *)))))))","ChainDbState"]}
```

PR #2857, opened 2022-06-09
Add testcase to check that reference scripts in output won't be run

teodanciu:
for validation

A test for: https://github.com/input-output-hk/cardano-ledger/issues/2826


JaredCorduan (approved):
This looks great to me, exactly the test I was hoping for!

PR #2858, opened 2022-06-10
Unit test for spending a simple script locked UTxO

JaredCorduan:
Unit test for spending a simple script locked UTxO using a reference script.


teodanciu (approved):
Looks good to me

PR #2860, opened 2022-06-13
removed the unused function minCollateral

JaredCorduan:
The function `minCollateral` is not used anywhere, and is delete by this PR. Great catch by @jmhrpr!

The babbage UTxO rule calls `validateTotalCollateral`, which uses `validateTotalCollateral` from the Alonzo era, to check that enough collateral was provided. The spec does not used a named function either.

For reference:

https://github.com/input-output-hk/cardano-ledger/blob/0b0d066bf35f4062545bfaeeaf2cca9a1a87a2c3/eras/babbage/impl/src/Cardano/Ledger/Babbage/Rules/Utxo.hs#L219

https://github.com/input-output-hk/cardano-ledger/blob/0b0d066bf35f4062545bfaeeaf2cca9a1a87a2c3/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Rules/Utxo.hs#L337-L344

### DIFF
--- a/eras/babbage/impl/src/Cardano/Ledger/Babbage/Collateral.hs
+++ b/eras/babbage/impl/src/Cardano/Ledger/Babbage/Collateral.hs
@@ -16,10 +16,8 @@ import Cardano.Ledger.Babbage.TxBody
     TxOut (..),
     collateralReturn',
     outputs',
-    txfee',
   )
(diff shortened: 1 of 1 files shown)


lehins (approved):
:+1:

PR #2861, opened 2022-06-13
Remove empty SuccessfulPlutusScriptsEvent events

### LINKED ISSUE #2859: SuccessfulPlutusScriptsEvent emitted too often
The `SuccessfulPlutusScriptsEvent` event is emitted even when a transaction contains _no_ Plutus scripts:

https://github.com/input-output-hk/cardano-ledger/blob/0b0d066bf35f4062545bfaeeaf2cca9a1a87a2c3/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Rules/Utxos.hs#L243

The path to `Passes ps` contains the case that `ps` is empty.

We should not emit this event when `ps` is empty. We should use `NonEmpty` instead of a List inside `SuccessfulPlutusScriptsEvent` (this also makes sense in the branch where a failed Plutus script exists).

JaredCorduan:
The `SuccessfulPlutusScriptsEvent` event should only be emitted when the list of successful Plutus scripts is nonempty. I've made `SuccessfulPlutusScriptsEvent` use `NonEmpty` instead of list.

Additionally, the `SuccessfulPlutusScriptsEvent` events were not being emitted at all in the Babbage era, in the no-failures path.

resolves #2859

### DIFF
--- a/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Rules/Utxos.hs
+++ b/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Rules/Utxos.hs
@@ -85,7 +85,7 @@ import qualified Data.ByteString.Base64 as B64
 import Data.Coders
 import Data.Foldable (toList)
 import Data.List (intercalate)
-import Data.List.NonEmpty (NonEmpty (..))
+import Data.List.NonEmpty (NonEmpty (..), nonEmpty)
 import qualified Data.List.NonEmpty as NE
 import qualified Data.Map.Strict as Map
 import Data.MapExtras (extractKeys)
@@ -138,7 +138,7 @@ instance
 
 data UtxosEvent era
   = AlonzoPpupToUtxosEvent (Event (Core.EraRule "PPUP" era))
-  | SuccessfulPlutusScriptsEvent [PlutusDebug]
+  | SuccessfulPlutusScriptsEvent (NonEmpty PlutusDebug)
   | FailedPlutusScriptsEvent (NonEmpty PlutusDebug)
 
 instance
@@ -240,7 +240,7 @@ scriptsValidateTransition = do
         ValidationTagMismatch
           (getField @"isValid" tx)
           (FailedUnexpectedly (scriptFailuresToPredicateFailure fs))
-    Passes ps -> tellEvent (SuccessfulPlutusScriptsEvent ps)
+    Passes ps -> mapM_ (tellEvent . SuccessfulPlutusScriptsEvent) (nonEmpty ps)
 
   () <- pure $! traceEvent validEnd ()
 
@@ -270,7 +270,7 @@ scriptsNotValidateTransition = do
       failBecause $
         ValidationTagMismatch (getField @"isValid" tx) PassedUnexpectedly
     Fails ps fs -> do
-      tellEvent (SuccessfulPlutusScriptsEvent ps)
+      mapM_ (tellEvent . SuccessfulPlutusScriptsEvent) (nonEmpty ps)
       tellEvent (FailedPlutusScriptsEvent (scriptFailuresToPlutusDebug fs))
 
   let !_ = traceEvent invalidEnd ()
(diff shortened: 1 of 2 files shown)


teodanciu (approved):
Looks good to me.

I have a vague feeling  that there is some abstraction to express `maybe (pure ()) f someMaybe`, but I don't know which and playing a bit with it lead to no results! It was more a curiosity, this is perfectly fine  :+1:

JaredCorduan:
> Looks good to me.
> 
> I have a vague feeling that there is some abstraction to express `maybe (pure ()) f someMaybe`, but I don't know which and playing a bit with it lead to no results! It was more a curiosity, this is perfectly fine +1

Thanks @teodanciu ! I'm happy to leave the PR up for a while and see if anyone else knows a better way to write it. I had the same vague feeling...

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Rules/Utxos.hs
```diff
@@ -240,7 +240,7 @@ scriptsValidateTransition = do
         ValidationTagMismatch
           (getField @"isValid" tx)
           (FailedUnexpectedly (scriptFailuresToPredicateFailure fs))
-    Passes ps -> tellEvent (SuccessfulPlutusScriptsEvent ps)
+    Passes ps -> maybe (pure ()) (tellEvent . SuccessfulPlutusScriptsEvent) (nonEmpty ps)
```
lehins:
```suggestion
    Passes ps -> mapM_ (tellEvent . SuccessfulPlutusScriptsEvent) (nonEmpty ps)
```

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Rules/Utxos.hs
```diff
@@ -270,7 +270,7 @@ scriptsNotValidateTransition = do
       failBecause $
         ValidationTagMismatch (getField @"isValid" tx) PassedUnexpectedly
     Fails ps fs -> do
-      tellEvent (SuccessfulPlutusScriptsEvent ps)
+      maybe (pure ()) (tellEvent . SuccessfulPlutusScriptsEvent) (nonEmpty ps)
```
lehins:
```suggestion
      mapM_ (tellEvent . SuccessfulPlutusScriptsEvent) (nonEmpty ps)
```

eras/babbage/impl/src/Cardano/Ledger/Babbage/Rules/Utxos.hs
```diff
@@ -208,7 +209,7 @@ scriptsNo = do
         when2Phase $ case evalScripts @era (getField @"_protocolVersion" pp) tx sLst of
           Passes _ -> failBecause $ ValidationTagMismatch (getField @"isValid" tx) PassedUnexpectedly
           Fails ps fs -> do
-            tellEvent (SuccessfulPlutusScriptsEvent ps)
+            maybe (pure ()) (tellEvent . SuccessfulPlutusScriptsEvent) (nonEmpty ps)
```
```suggestion
            mapM_ (tellEvent . SuccessfulPlutusScriptsEvent) (nonEmpty ps)
```

lehins (approved):
@teodanciu and @JaredCorduan abstraction you are looking for called `mapM_`/`traverse_` :wink: 

This looks great. It always makes me happy when a switch to NonEmpty happens :smile:

eras/babbage/impl/src/Cardano/Ledger/Babbage/Rules/Utxos.hs
```diff
@@ -175,7 +176,7 @@ scriptsYes = do
                 ValidationTagMismatch
                   (getField @"isValid" tx)
                   (FailedUnexpectedly (scriptFailuresToPredicateFailure fs))
-            Passes _ -> pure ()
+            Passes ps -> maybe (pure ()) (tellEvent . SuccessfulPlutusScriptsEvent) (nonEmpty ps)
```
JaredCorduan: I'll snag this one too! :)

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Rules/Utxos.hs
Soupstraw: Dammit, I actually wanted to suggest the same, but I forgot to submit the review after I made the suggestions :laughing:

Soupstraw (approved):

PR #2863, opened 2022-06-14
Do not serialize empty txbody fields

### LINKED ISSUE #2862: Alonzo and Babbage TxBody serializers sometimes produce empty containers.
Two optional transaction body fields from Alonzo and Babbage, and one Babbage-only optional field, are being serialized as empty containers when not used. We should not include them at all!

Alonzo:
* collateral inputs: https://github.com/input-output-hk/cardano-ledger/blob/0b0d066bf35f4062545bfaeeaf2cca9a1a87a2c3/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxBody.hs#L724
* required signers: https://github.com/input-output-hk/cardano-ledger/blob/0b0d066bf35f4062545bfaeeaf2cca9a1a87a2c3/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxBody.hs#L732

Babbage:
* collateral inputs: https://github.com/input-output-hk/cardano-ledger/blob/0b0d066bf35f4062545bfaeeaf2cca9a1a87a2c3/eras/babbage/impl/src/Cardano/Ledger/Babbage/TxBody.hs#L826
* required signers: https://github.com/input-output-hk/cardano-ledger/blob/0b0d066bf35f4062545bfaeeaf2cca9a1a87a2c3/eras/babbage/impl/src/Cardano/Ledger/Babbage/TxBody.hs#L837
* reference inputs: https://github.com/input-output-hk/cardano-ledger/blob/0b0d066bf35f4062545bfaeeaf2cca9a1a87a2c3/eras/babbage/impl/src/Cardano/Ledger/Babbage/TxBody.hs#L827

JaredCorduan:
Two optional transaction body fields from Alonzo and Babbage, and one Babbage-only optional field, are being serialized as empty containers when not used. They should be omitted when not used, which is what this PR changes.

The Alonzo/Babbage fields are:
* collateral inputs (key 13 our CBOR encoding of the txbody)
* required signers (key 14 our CBOR encoding of the txbody)

The Babbage-only field is:
* reference inputs (key 18 our CBOR encoding of the txbody)

Note that this is not a breaking change, our deserializers are flexible enough to handle empty containers.

resolves #2862

### DIFF
--- a/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxBody.hs
+++ b/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxBody.hs
@@ -721,7 +721,7 @@ encodeTxBodyRaw
           TxBodyRaw i ifee o c w f (ValidityInterval b t) u rsh mi sh ah ni
       )
       !> Key 0 (E encodeFoldable _inputs)
-      !> Key 13 (E encodeFoldable _collateral)
+      !> Omit null (Key 13 (E encodeFoldable _collateral))
       !> Key 1 (E encodeFoldable _outputs)
       !> Key 2 (To _txfee)
       !> encodeKeyedStrictMaybe 3 top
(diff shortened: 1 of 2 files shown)


TimSheard (approved):
Data.Coders used as it was meant to be.
good catch.

PR #2864, opened 2022-06-14
correct the use of VRF extension in babbage spec

JaredCorduan:
The babbage ledger spec was not correctly describing the way to perform the VRF extension from the [paper](https://iohk.io/en/research/library/papers/on-uc-secure-range-extension-and-batch-verification-for-ecvrf/). The implementation, however, is doing the [correct thing](https://github.com/input-output-hk/ouroboros-network/blob/d78f61c4d720c1eab65eb3edc1a313f58b5134e8/ouroboros-consensus-protocol/src/Ouroboros/Consensus/Protocol/Praos/VRF.hs#L95-L99).

### DIFF
--- a/eras/babbage/formal-spec/references.bib
+++ b/eras/babbage/formal-spec/references.bib
@@ -126,3 +126,10 @@ url = {https://github.com/input-output-hk/cardano-ledger/blob/master/eras/shelle
   year  = {2021},
   url = {https://hydra.iohk.io/build/7244320/download/1/plutus-core-specification.pdf}
 }
+
+@misc{vrf-range-extension,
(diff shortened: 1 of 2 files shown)


iquerejeta (approved):
Looks good 👍

SebastienGllmt:
I spent at least two days re-reading through the VRF stuff wondering how any of this made sense so this PR definitely helps a lot. Still kind of confused because of https://github.com/input-output-hk/cardano-ledger/issues/2868

JaredCorduan:
I'm sorry this caused you confusion and frustration @SebastienGllmt .

PR #2865, opened 2022-06-14
WIP- refactor Babbage Features test

teodanciu:
My idea with this refactoring is to avoid  having to keep "manual" track of what variable to  use/reuse, so that the test case is consistent. So now the testcase is constructing this TestCaseData object and calls a function, and then inside the function, we are using the data to construct circumstances that make sense for the scenario. 

I'm sure it's too specific/specialized in some cases, because my understanding of these concepts are quite shallow atm, so I doubt that these abstractions I introduced are sound.. Suggestions welcome!

A few test cases for valid tx missing,  plus the the invalid cases - but the approach seems to work so far :crossed_fingers:

### DIFF
--- a/libs/cardano-ledger-test/cardano-ledger-test.cabal
+++ b/libs/cardano-ledger-test/cardano-ledger-test.cabal
@@ -46,6 +46,7 @@ library
     Test.Cardano.Ledger.Alonzo.Tools
     Test.Cardano.Ledger.BaseTypes
     Test.Cardano.Ledger.Examples.BabbageFeatures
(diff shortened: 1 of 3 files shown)


lehins:
I haven't looked at the changes yet, but I'd recommend modifying the `Test.Cardano.Ledger.Examples.BabbageFeatures` directly. It will be easier to see what is being changed in the github diff.

teodanciu:
Superseded by https://github.com/input-output-hk/cardano-ledger/pull/2872

PR #2871, opened 2022-06-17
Added ListMap, changed some fields in ShelleyGenesis to use ListMap

### LINKED ISSUE #2867: Replace maps with associations lists in ShelleyGenesis
We want to replace the maps inside the [ShelleyGenesis](https://github.com/input-output-hk/cardano-ledger/blob/9d5fe3e539605e3bf9d98d73ea892aa467d55058/eras/shelley/impl/src/Cardano/Ledger/Shelley/Genesis.hs#L126) type with associations lists (ie `[(k, v)]`). We do **not**, however, want to change to the JSON serialized format.

The context is that for bench-marking we sometimes want to make and serialize to JSON genesis files with huge initial utxos and delegation maps. Therefore the JSON output should use a stream friendly `toEncoding` implementation.

Soupstraw:
Created a `ListMap` wrapper around association maps. The wrapper makes the association list serialize like a dictionary/object, but otherwise behaves like a list. It should take up less memory when serializing than a regular `Map` since it can be streamed.

Changed `GenDeleg` so it uses a `ListMap` instead of `Map`. 

`ToJSON` implementation for `ListMap` was taken from here: https://github.com/input-output-hk/cardano-node/pull/4021/files

resolves #2867

### DIFF
--- a/eras/shelley/test-suite/src/Test/Cardano/Ledger/Shelley/Generator/Block.hs
+++ b/eras/shelley/test-suite/src/Test/Cardano/Ledger/Shelley/Generator/Block.hs
@@ -69,6 +69,7 @@ import Test.Cardano.Ledger.Shelley.Utils
   )
 import Test.QuickCheck (Gen)
 import qualified Test.QuickCheck as QC (choose)
+import qualified Data.ListMap as LM
 
 -- ======================================================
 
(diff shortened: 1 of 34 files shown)


eras/shelley/test-suite/src/Test/Cardano/Ledger/Shelley/Generator/Block.hs
```diff
@@ -213,17 +214,17 @@ selectNextSlotWithLeader
...
                 <$> List.find
-                  ( \(AllIssuerKeys {vrf, hk}) ->
+                  ( \AllIssuerKeys {vrf, hk} ->
                       isLeader hk (fst vrf)
                   )
                   ksStakePools
             Just (ActiveSlot x) ->
               fmap coerce $
-                Map.lookup x cores
+                LM.lookup x cores
```
Soupstraw: Lookups are now linear instead of logarithmic. Are we fine with that?

libs/cardano-data/src/Data/ListMap.hs
```diff
@@ -0,0 +1,136 @@
+{-# LANGUAGE DeriveFoldable #-}
+{-# LANGUAGE DeriveFunctor #-}
+{-# LANGUAGE DeriveGeneric #-}
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE RankNTypes #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+
+module Data.ListMap
+  ( ListMap (..),
```
Soupstraw, line 9: It might be a good idea to hide the constructor and implement `fromList` so it works similarly to `Map` wrt lookups. Right now I think the constructor adds the list elements in the opposite order to `Map.toList`.

lehins: Simulating what `Map` does and hiding constructor will defeat the purpose of this ListMap, at least what cardano-node is trying to use it for

JaredCorduan (commented):
This looks good to me!

JaredCorduan (approved):
LGTM!

PR #2872, opened 2022-06-17
Refactor  `BabbageFeatures`

teodanciu:
The idea behind this refactoring is this:
 - instead of having a shared initial utxo and each test create both: 1) transaction to validate and 2) expected utxo,  in this new version the test case will specify only  the transaction and the outputs that have to exist in the initialUtxo when validating. The responsibility of creating the initialUtxo and the expectedState falls on the three functions: testExpectSuccessValid, testExpectSuccessInvalid and testExpectFailure, and it's doing that based on the data provided by the testcase.

The things I like about this in comparison with the existing version:
 - It seems easier to get a quick understanding of the transaction whose validity we are checking, because the data that composes it is in the same place
 - No more/fewer awkwardly-named variables (like: simpleV2EUTxOInput, collateralInput17) - they were necessary because each testcase was using different bits of data to both create the transaction and the expectated state
 - No more shared `initUtxo` map, which means when a testcase fails it's way easier to understand the output
 - For me the biggest problem in the current implementation is the non-explicit dependency between the transaction being validated and the output corresponding to the input referenced by it (this output has to be in the initUtxo map).  In the new version, this dependency is still not great (in the form of `InitOutputs`) because it's not enforced in any way, however, because they are part of the same value, it's at least easier to correlate them, in my opinion
 - I think the logic of which inputs should be in the resulting utxo (that is implemented in testExpectSuccess(In)Valid) is easier to understand than the alternative `expectedUtxo`, mostly because the value of the transaction index has no meaning (so the +10 for the collateral logic is gone)

What I don't like:
 - the relationship between the txBody and the corresponding outputs is still loose and unenforced. The only benefit is that they are closer to each other in the code, so maybe easier to mentally correlate them
 - specifying the WitnessesI of the transaction is unpleasant, because it has to be done with two different fields.. Ideally they would all go in a list, but we need the txBody in order to determine the AddrWits' - so for this reason, instead of passing the WitnessesField, it's passing just the keys, and computing the AddressWits' inside testExpectSuccessValid. The rest of the WitnessesFiels are passed as a list via the field `otherWitsFields`  (terrible name)

### DIFF
--- a/libs/cardano-ledger-test/src/Test/Cardano/Ledger/Examples/BabbageFeatures.hs
+++ b/libs/cardano-ledger-test/src/Test/Cardano/Ledger/Examples/BabbageFeatures.hs
@@ -4,6 +4,7 @@
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE FlexibleInstances #-}
 {-# LANGUAGE GADTs #-}
+{-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE MultiParamTypeClasses #-}
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE ScopedTypeVariables #-}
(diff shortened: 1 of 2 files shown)


JaredCorduan (approved):
This is fantastic!

This re-organization makes it much easier to understand what data goes into each test. It also makes it much clearer how each test touches the UTxO accounting. Future tests readers and authors will thank you!

PR #2875, opened 2022-06-17
Disable staking Ptr optimization

lehins:
Unfortunately we can't make Ptr unpacking optimization until Babbage has been properly released, so this PR reverts the functionality to the 1.33/1.34 version. This indirectly relies on Address deserialization, which has been fixed in #2750, but it needs to be on mainnet, before we can unpack the Ptr.

### DIFF
--- a/eras/babbage/test-suite/test/Test/Cardano/Ledger/Babbage/Serialisation/Tripping.hs
+++ b/eras/babbage/test-suite/test/Test/Cardano/Ledger/Babbage/Serialisation/Tripping.hs
@@ -46,7 +46,7 @@ trippingF f x =
             <> case deserialiseFromBytes decodeTerm (serialize x) of
               Left e -> ppShow e
               Right (_, terms) -> ppShow terms
-            <> "\nreamining term: "
+            <> "\nremaining term: "
             <> case deserialiseFromBytes decodeTerm (serialize y) of
               Left e -> ppShow e
               Right (_, terms) -> ppShow terms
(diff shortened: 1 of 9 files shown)


libs/cardano-ledger-core/src/Cardano/Ledger/Credential.hs
```diff
@@ -115,11 +115,11 @@ data StakeReference crypto
...
 
--- | Pointer to a slot number, transaction index and an index in certificate
--- list. We expect that `SlotNo` will fit into `Word32` for a very long time,
+-- TODO: implement this optimization:
+-- We expect that `SlotNo` will fit into `Word32` for a very long time,
 -- because we can assume that the rate at which it is incremented isn't going to
 -- increase in the near future. Therefore with current rate we should be fine for
--- about a 150 years. I suggest to remove this optimization in about a
+-- another 134 years. I suggest to remove this optimization in about a
 -- hundred years or thereabouts, so around a year 2122 would be good.
```
JaredCorduan, line 123: I forgot about this gem :laughing:

JaredCorduan (approved):
Great catch @lehins ! Looks good to me.

eras/babbage/test-suite/test/Test/Cardano/Ledger/Babbage/Serialisation/Tripping.hs
```diff
@@ -46,7 +46,7 @@ trippingF f x =
             <> case deserialiseFromBytes decodeTerm (serialize x) of
               Left e -> ppShow e
               Right (_, terms) -> ppShow terms
-            <> "\nreamining term: "
```
JaredCorduan, line 49: you don't want to ream the term n times? :laughing:

JaredCorduan:
looks like the `ci/hydra:Cardano:cardano-ledger-specs:required` job timed out (on one of the mac-minis). seeing as the specs weren't even touched here, I'm going to merge.

PR #2876, opened 2022-06-18
Non-semantic cleanup of some of the rules modules

lehins:


eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Pool.hs
```diff
@@ -201,13 +201,11 @@ poolDelegationTransition = do
...
 
-      when (SoftForks.restrictPoolMetadataHash pp) $ do
-        case _poolMD poolParam of
-          SNothing -> pure ()
-          SJust pmd ->
-            let s = BS.length (_poolMDHash pmd)
-             in s <= fromIntegral (sizeHash ([] @(CC.HASH (Crypto era))))
-                  ?! PoolMedataHashTooBig (_poolId poolParam) s
+      when (SoftForks.restrictPoolMetadataHash pp) $
+        sfor_ (_poolMD poolParam) $ \pmd ->
```
JaredCorduan: this version is a bit more compact, but (at least for me) a bit harder to understand (since we don't usually use `_sfor`. I'm not opposed to the change, I'm just not sure what benefit you see.

JaredCorduan (approved):
great stuff! I made one small comment, which you can ignore if you want. thanks for the clean up!

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Pool.hs
lehins: You know, I agree, `sfor_` is our custom version of `for_` for `Traversable`, I'll got ahead and switch it to `forM_` from base. It is just a variant of `traverse_` and it is the class of functions that we should embrace. Coincidentally it came up just a few days ago: https://github.com/input-output-hk/cardano-ledger/pull/2861#pullrequestreview-1007397998

JaredCorduan: great, `forM_` is less of a roadbump for me, thanks!

PR #2878, opened 2022-06-20
integrity hash not needed for all ref scripts

JaredCorduan:
This fixes a bug found by @james-iohk. See https://github.com/input-output-hk/cardano-node/issues/4057#issuecomment-1158011896

If a reference script is not actually going to be run, we do not need to demand the corresponding language be a part of the script integrity hash. The problem stems from the `txscripts` function, which is almost never the function one wants (it gives you all the scripts associated with a tx, not the ones that are going to be evaluated).

This commit fixes the implementation, we must also fix the spec (cc @WhatisRT , when we fix this, we should definitely add @james-iohk to the list of contributors!).

Note that I renamed an era-generic helper function used in the property tests. I renamed `scriptsNeeded'` to `scriptWitsNeeded'`, since that function is used for supplying the script witnesses that are needed. I also added a new function called `scriptsNeeded'` which returns the scripts that are needed for evaluation (this name matches the spec and the implementation now).

### DIFF
--- a/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxInfo.hs
+++ b/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxInfo.hs
@@ -783,9 +783,11 @@ languages ::
   ) =>
   Core.Tx era ->
   UTxO era ->
+  Set (ScriptHash (Crypto era)) ->
   Set Language
-languages tx utxo = Map.foldl' accum Set.empty allscripts
+languages tx utxo sNeeded = Map.foldl' accum Set.empty allscripts
   where
-    allscripts = txscripts @era utxo tx
+    isNeeded scriptHash _script = scriptHash `Set.member` sNeeded
+    allscripts = Map.filterWithKey isNeeded $ txscripts @era utxo tx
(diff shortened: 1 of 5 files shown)


eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxInfo.hs
```diff
@@ -783,9 +783,11 @@ languages ::
...
   Core.Tx era ->
   UTxO era ->
+  Set (ScriptHash (Crypto era)) ->
   Set Language
-languages tx utxo = Map.foldl' accum Set.empty allscripts
+languages tx utxo sNeeded = Map.foldl' accum Set.empty allscripts
   where
-    allscripts = txscripts @era utxo tx
+    isNeeded scriptHash _script = scriptHash `Set.member` sNeeded
+    allscripts = Map.filterWithKey isNeeded $ txscripts @era utxo tx
```
lehins:
```suggestion
    allscripts = Map.restrictKeys (txscripts @era utxo tx) sNeeded
```

lehins (approved):
Looks good. Just one suggestion.

PR #2879, opened 2022-06-21
Removed provenance from the reward calculation

### LINKED ISSUE #2744: Remove reward provenance
The ledger events are replacing the reward provenance, and hence we should remove reward provenance.

TimSheard:
Removed provenance from the reward calculation, and also all the tests. This involved all the reward-pulsing tests (that used to involve provenance, and now don't) and also the tests that provenance didn't affect the underlying computation.

closes #2744


eras/shelley/impl/src/Cardano/Ledger/Shelley/API/Wallet.hs
```diff
@@ -430,12 +427,10 @@ getRewardProvenance ::
   ) =>
   Globals ->
   NewEpochState era ->
-  (RewardUpdate (Crypto era), RewardProvenance (Crypto era))
+  (RewardUpdate (Crypto era))
 getRewardProvenance globals newepochstate =
```
JaredCorduan, line 433: This is great @TimSheard , thanks for doing this clean up!  Before we merge this, we should make sure we coordinate with the consensus team about how best to deprecate their [query](https://github.com/input-output-hk/ouroboros-network/blob/a10ba425a5db0b7a009d6f16f066bb70b2bd3707/ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Ledger/Query.hs#L250-L251). We've already gutted most of the information from that query, so it's really just about deprecating the query in the most clear and helpful way we can.

eras/shelley/impl/cardano-ledger-shelley.cabal
```diff
@@ -55,7 +55,7 @@ library
     Cardano.Ledger.Shelley.PoolRank
     Cardano.Ledger.Shelley.PParams
     Cardano.Ledger.Shelley.Rewards
-    Cardano.Ledger.Shelley.RewardProvenance
+    Cardano.Ledger.Shelley.RewardProvenance
```
JaredCorduan, line 58: trailing whitespace (I _think_ ormolu will notice as well)

JaredCorduan (approved):
looks great, thank you!

eras/shelley/impl/cardano-ledger-shelley.cabal
lehins: Ormolu does not touch cabal files

JaredCorduan:
`ExtraneousScriptWitnessesUTXOW` in the generation strikes again. https://github.com/input-output-hk/cardano-ledger/issues/2811

I'm going to merge this PR, as this failure is a known in the generators and is unrelated.

PR #2880, opened 2022-06-22
Cleanup `Cardano.Ledger.Shelley.TxBody` module.

lehins:
* Move `WitVKey` into its own module in `cardano-ledger-core`. Make `wvkKeyHash` lazy.
* Moved `HKD` into `cardano-ledger-core`
* Cleanup dead code: remove unused `StakeCreds`
* Move `PoolParams` into their own module
* Move all `DCert` related functionality from `TxBody` module

See individual commit messages for more info.


JaredCorduan (approved):
Fantastic clean up, happy to see all of it! (I think much of this can move to `cardano-ledger-core` at some point in the future as well, but this is an excellent first step that is less disruptive right before the HF).

PR #2888, opened 2022-06-28
fix plutus V2 conversion in error report

JaredCorduan:
We were attempting to convert all script contexts to the V1 context when reporting a Plutus script failure.

I've also refactored the function that produces the error report.

### DIFF
--- a/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxInfo.hs
+++ b/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxInfo.hs
@@ -669,91 +669,74 @@ explainPlutusFailure ::
   CostModel ->
   ExUnits ->
   ScriptResult
-explainPlutusFailure _proxy pv lang scriptbytestring e ds@[dat, redeemer, info] cm eu =
-  -- A three data argument script.
-  let ss :: Script era
-      ss = PlutusScript lang scriptbytestring
-      name :: String
-      name = show ss
-   in case PV1.fromData info of
-        Nothing -> scriptFail $ PlutusSF line db
-          where
-            line =
-              pack $
-                unlines
-                  [ "\nThe 3 arg plutus script (" ++ name ++ ") fails.",
-                    show e,
-                    "The protocol version is: " ++ show pv,
-                    "The data is: " ++ show dat,
-                    "The redeemer is: " ++ show redeemer,
-                    "The third data argument, does not decode to a context\n" ++ show info
-                  ]
-            db = PlutusDebugV1 cm eu scriptbytestring ds pv
-        Just info2 -> scriptFail $ PlutusSF line db
-          where
-            info3 = show (pretty (info2 :: PV1.ScriptContext))
-            line =
-              pack $
-                unlines
-                  [ "\nThe 3 arg plutus script (" ++ name ++ ") fails.",
-                    show e,
-                    "The protocol version is: " ++ show pv,
-                    "The data is: " ++ show dat,
-                    "The redeemer is: " ++ show redeemer,
-                    "The context is:\n" ++ info3
-                  ]
-            db = PlutusDebugV1 cm eu scriptbytestring ds pv
-explainPlutusFailure _proxy pv lang scriptbytestring e ds@[redeemer, info] cm eu =
-  -- A two data argument script.
-  let ss :: Script era
-      ss = PlutusScript lang scriptbytestring
-      name :: String
-      name = show ss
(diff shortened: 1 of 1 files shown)


eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxInfo.hs
```diff
@@ -669,91 +669,75 @@ explainPlutusFailure ::
...
+                      "The redeemer is: " ++ show redeemer,
+                      "The third data argument, does not translate to a V1 script context\n" ++ show info
+                    ]
+                  Just info2 ->
+                    [ "The data is: " ++ show dat,
+                      "The redeemer is: " ++ show redeemer,
+                      "The script context is:\n" ++ show (pretty (info2 :: PV1.ScriptContext))
+                    ]
+              PlutusV2 ->
+                case PV2.fromData info of
```
JaredCorduan, line 697: This line (and the one like it below for two-arg scripts) is the bug fix. We need to use `PV2.fromData` for Plutus V2.

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxInfo.hs
```diff
@@ -669,91 +669,75 @@ explainPlutusFailure ::
   CostModel ->
   ExUnits ->
   ScriptResult
-explainPlutusFailure _proxy pv lang scriptbytestring e ds@[dat, redeemer, info] cm eu =
+explainPlutusFailure _proxy pv lang scriptbytestring e ds cm eu =
   -- A three data argument script.
```
teodanciu: This comment is perhaps outdated at this place  now  ?

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxInfo.hs
```diff
@@ -669,91 +669,75 @@ explainPlutusFailure ::
...
+                      "The redeemer is: " ++ show redeemer,
+                      "The script context is:\n" ++ show (pretty (info2 :: PV2.ScriptContext))
+                    ]
+          [redeemer, info] ->
+            case lang of
+              PlutusV1 ->
+                case PV1.fromData info of
+                  Nothing ->
+                    [ "The redeemer is: " ++ show redeemer,
+                      "The third data argument, does not translate to a V1 script context\n" ++ show info
```
teodanciu: Hm, should this be maybe "The second data argument" ?

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxInfo.hs
JaredCorduan: yes indeed, great catch, thank you!

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxInfo.hs
```diff
@@ -669,91 +669,75 @@ explainPlutusFailure ::
...
+                    ]
+                  Just info2 ->
+                    [ "The redeemer is: " ++ show redeemer,
+                      "The script context is:\n" ++ show (pretty (info2 :: PV1.ScriptContext))
+                    ]
+              PlutusV2 ->
+                case PV2.fromData info of
+                  Nothing ->
+                    [ "The redeemer is: " ++ show redeemer,
+                      "The third data argument, does not translate to a V2 script context\n" ++ show info
```
teodanciu: Same question as above, wonder if it should say "The second argument"  instead of  third

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxInfo.hs
JaredCorduan: :facepalm: thanks

teodanciu (approved):
Nice refactoring, I like the separation of  building the messages and then in a single place putting them together and using the result, much easier to understand the idea of the function in this way. 

Just a curiosity: how did this problem manifest itself? Was it throwing errors for transactions with failing V2 scripts? Was it causing problems beyond inadequate reporting of what went wrong?

JaredCorduan:
> Just a curiosity: how did this problem manifest itself? Was it throwing errors for transactions with failing V2 scripts? Was it causing problems beyond inadequate reporting of what went wrong?

yep, exactly! the translation from `Data` originating from a V2 script context to a V2 script context failed `fromData`, and returned the "does not decode to a context" message.

Issue #2811, opened 2022-05-23
Alonzo test failure


JaredCorduan:
Investigate the following alonzo test failure 

```
$ cabal test cardano-ledger-alonzo-test --test-options='--quickcheck-replay=535047 -p "Fast Alonzo Property Tests"'

...
Test suite cardano-ledger-alonzo-test: RUNNING...

Running in scenario: Development
Alonzo tests
  Fast Alonzo Property Tests
    total amount of Ada is preserved (Chain): 
Discarded trace: MinFeee violation: genEraDne: AlonzoEraGen.hs

Discarded trace: TotExUnits violation: genEraTweakBlock: AlonzoEraGen.hs

Discarded trace: No inputs left. Utxo.hs
FAIL (21.81s)
      *** Failed! (after 15 tests):
      Exception while generating shrink-list:
        LEDGER sigGen: [UtxowFailure (WrappedShelleyEraFailure (ExtraneousScriptWitnessesUTXOW (fromList [ScriptHash "42c7a014a4cd5537f64e5ae8ec7349db3d8603e16765dc37f8fb6e67"])))]
        CallStack (from HasCallStack):
          error, called at src/Test/Cardano/Ledger/Shelley/Generator/Trace/Ledger.hs:179:24 in cardano-ledger-shelley-test-0.1.0.0-inplace:Test.Cardano.Ledger.Shelley.Generator.Trace.Ledger
          genAndApplyTx, called at src/Test/Cardano/Ledger/Shelley/Generator/Trace/Ledger.hs:159:11 in cardano-ledger-shelley-test-0.1.0.0-inplace:Test.Cardano.Ledger.Shelley.Generator.Trace.Ledger
          sigGen, called at src/Test/Cardano/Ledger/Shelley/Generator/Block.hs:101:16 in cardano-ledger-shelley-test-0.1.0.0-inplace:Test.Cardano.Ledger.Shelley.Generator.Block
      Use --quickcheck-replay=535047 to reproduce.

1 out of 1 tests failed (21.82s)
Test suite cardano-ledger-alonzo-test: FAIL
```

wocaoluna:
[...]

PR #2890, opened 2022-06-30
Extra ListMap functions

newhoggy:
These additional functions will be useful for downstream projects.

### DIFF
--- a/libs/cardano-data/src/Data/ListMap.hs
+++ b/libs/cardano-data/src/Data/ListMap.hs
@@ -14,6 +14,12 @@ module Data.ListMap
     lookup,
     filter,
     toMap,
+    fromMap,
+    mapKeys,
+    map,
+    empty,
+    fromList,
+    toList,
   )
 where
 
@@ -39,6 +45,7 @@ import qualified Data.Aeson.Encoding as E
 import qualified Data.Aeson.Key as Key
 import qualified Data.Aeson.KeyMap as KM
 import Data.Aeson.Types (listValue)
+import Data.Bifunctor (Bifunctor (..))
 import Data.Coerce (coerce)
 import qualified Data.List as L
 import qualified Data.Map.Strict as Map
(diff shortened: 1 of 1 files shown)


JaredCorduan (approved):
LGTM!

newhoggy:
I also added `empty`, `fromList` and `toList` to make the API more like `Data.Map`.

libs/cardano-data/src/Data/ListMap.hs
```diff
@@ -143,3 +150,21 @@ filter f (ListMap xs) = ListMap $ Pre.filter (uncurry f) xs
...
 toMap (ListMap xs) = Map.fromList xs
+
+fromMap :: Map.Map k v -> ListMap k v
+fromMap = ListMap . Map.toList
+
+mapKeys :: (k1 -> k2) -> ListMap k1 a -> ListMap k2 a
+mapKeys f = ListMap . fmap (first f) . unListMap
+
+map :: Ord k => (a -> v) -> ListMap k a -> ListMap k v
+map f = ListMap . Map.toList . Map.map f . Map.fromList . unListMap
```
Soupstraw, line 161: Wouldn't it be better to just map over the list instead of converting to a Map?

JaredCorduan:
Looks like #2811 hit this PR as well (and is unrelated to the changes here). I'll re-run CI.

libs/cardano-data/src/Data/ListMap.hs
lehins: @Soupstraw I saw that thing as well, do you wanna submit a PR with this fix, considering this one has already been merged. Using list comprehensions would very a clean solution.

lehins: In fact, as defined here, it strictly speaking is not even a correct `map`, since it also removes duplicates and affects the order of elements. Which is not an invariant we maintain in this implementation, as far as I am aware

PR #2891, opened 2022-07-01
Change sgGenDelegs to use ListMap

newhoggy:


JaredCorduan (commented):
@newhoggy is there a reason you wanted to change the underlying type of the genesis key mapping? Is it to match the types in the Shelley Genisis? The genesis key mapping only has seven elements on mainnet, and even in the benchmarks I believe it is small.

lehins:
We have specifically not converted `sgGenDelegs` to a `ListMap`, despite it being requested, because that data structure is an unsafe representation of a `Map`, which preserves neither the same complexities nor invariant. The only reason why it was acceptable for `sgInitialFunds` and `sgStaking` is because those fields are not used by ledger at all and their sole purpose is to adjust the state for benchmarking, which is what the ListMap change was also designed.

@newhoggy Sorry for closing this PR, but that is the reason behind it being closed. If you feel that you really need this change then feel free to reopen the PR or open another issue or reach out to me on slack and we can discuss how we can accommodate what you need in a safer manner.

PR #2892, opened 2022-07-01
Add toEncoding function to ToJSON instance of ShelleyGenesis and Shel…

newhoggy:
…leyGenesisStaking and use strict evaluation to help with lazy streaming

### DIFF
--- a/eras/shelley/impl/src/Cardano/Ledger/Shelley/Genesis.hs
+++ b/eras/shelley/impl/src/Cardano/Ledger/Shelley/Genesis.hs
@@ -168,6 +168,45 @@ instance Era era => ToJSON (ShelleyGenesis era) where
         "staking" .= sgStaking sg
       ]
 
+  toEncoding
+    ShelleyGenesis
+      { sgSystemStart,
+        sgNetworkMagic,
+        sgNetworkId,
+        sgActiveSlotsCoeff,
+        sgSecurityParam,
+        sgEpochLength,
+        sgSlotsPerKESPeriod,
+        sgMaxKESEvolutions,
+        sgSlotLength,
+        sgUpdateQuorum,
+        sgMaxLovelaceSupply,
+        sgProtocolParams,
+        sgGenDelegs,
+        sgInitialFunds,
+        sgStaking
+      } =
+      Aeson.pairs $
+        let !strictSgInitialFunds = sgInitialFunds
+            !strictSgStaking = sgStaking
+         in mconcat
+              [ "systemStart" .= sgSystemStart,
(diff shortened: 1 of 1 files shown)


eras/shelley/impl/src/Cardano/Ledger/Shelley/Genesis.hs
```diff
@@ -167,6 +167,43 @@ instance Era era => ToJSON (ShelleyGenesis era) where
...
+        !epochLength = sgEpochLength sg
+        !slotsPerKESPeriod = sgSlotsPerKESPeriod sg
+        !maxKESEvolutions = sgMaxKESEvolutions sg
+        !slotLength = sgSlotLength sg
+        !updateQuorum = sgUpdateQuorum sg
+        !maxLovelaceSupply = sgMaxLovelaceSupply sg
+        !protocolParams = sgProtocolParams sg
+        !genDelegs = sgGenDelegs sg
+        !initialFunds = sgInitialFunds sg
+        !staking = sgStaking sg
```
lehins: If you want to ensure `sgInitialFunds` and `sgStaking` are strict then it would make sense to manually bind them with a `!`, but for other fields that are already strict this is just unnecessary boiler plate.  Can you please switch all others to `NamedFieldPuns`

eras/shelley/impl/src/Cardano/Ledger/Shelley/Genesis.hs
```diff
@@ -199,6 +236,17 @@ instance CC.Crypto crypto => ToJSON (ShelleyGenesisStaking crypto) where
       [ "pools" .= sgsPools sgs,
         "stake" .= sgsStake sgs
       ]
+  toEncoding sgs =
+    -- Forced evaluation of each field allows the parent object to no longer be
+    -- referenced, which helps prevent retention of any fields that have already
+    -- been serialised.
+    let !pools = sgsPools sgs
+        !stake = sgsStake sgs
```
lehins: Comment and strict let bindings are unnecessary, both of these fields are strict. Can you switch them to NamedFieldPuns?

lehins (dismissed):
Looks good overall, but I think it will be nicer if we can reduce boiler plate and redundant strictness annotations.

lehins:
@newhoggy Also I'd like to ask you to improve the commit message a little. We are trying to get better at it in ledger. Here is short guide about it: https://github.com/input-output-hk/cardano-ledger/blob/master/CONTRIBUTING.md#commit-messages

lehins (approved):
LGTM

Issue #1675, opened 2020-07-16
Re-organise code


nc6:
Code organisation. We have some giant dumping grounds of "stuff". It's not in a position where it can really be released to the community.

- [x] Rename `TxData` to `TxBody` (#1837)
- [x] Split apart `LedgerState` (#2894)
- [x] Split apart and document the `SetAlgebraInternals` (#2564)

PR #2894, opened 2022-07-01
Refactor - break LedgerState into smaller modules

Soupstraw:
This PR breaks the `Cardano.Ledger.Shelley.LedgerState` into smaller submodules.

Removed `WitHashes` and replaced all references to it with `Set (KeyHash 'Witness (Crypto era))`. 
~~**^ This might break some things upstream**~~ 
(Added deprecated functions and types to make everything backwards-compatible)

Before:
Before
After:
modules png-1


eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState.hs
```diff
@@ -28,272 +20,78 @@
 -- as state transformations on a ledger state ('LedgerState'),
 -- as specified in /A Simplified Formal Specification of a UTxO Ledger/.
 module Cardano.Ledger.Shelley.LedgerState
-  ( AccountState (..),
```
lehins: In order to reduce breakage downstream we still need to make sure we export everything from the LedgerState module that we exported before.

eras/shelley/impl/cardano-ledger-shelley.cabal
```diff
@@ -50,6 +50,10 @@ library
     Cardano.Ledger.Shelley.Genesis
     Cardano.Ledger.Shelley.HardForks
     Cardano.Ledger.Shelley.LedgerState
+    Cardano.Ledger.Shelley.LedgerState.Types
+    Cardano.Ledger.Shelley.LedgerState.DPState
+    Cardano.Ledger.Shelley.LedgerState.PulsingReward
+    Cardano.Ledger.Shelley.LedgerState.IncrementalStake
```
lehins: Maybe it is better to make those modules internal and let everything LedgerState related be exported from `Cardano.Ledger.Shelley.LedgerState`, just as it was before.

Less modules for public to deal with is better.

lehins (changes requested):
I did a quick skim through the PR and it overall looks pretty good. I made couple of comments on the re-exporting.

<strike> Also I do like having some visualizations and stuff, however they are bound to get outdated over time. So I don't think it is a good idea to add this image into the repo: `libs/cardano-ledger-core/modules.png-1.png` The only way we could add such graphs to the repo if we have a CI action that validates they did not go out of sync together with a clear readme section on how to regenerate them when they do.</strike> I saw that you removed it :smile:, but it was only after I submitted the review

One more important point. I am sure you know this, but just to be safe I'd like to mention it anyways, make sure you don't add any semantic changes in this PR, since in refactorings like that it is hard to notice any changes to the logic implemented in functions.

JaredCorduan:
Note that this PR closes  #1675

eras/shelley/impl/cardano-ledger-shelley.cabal
Soupstraw: Done, moved the submodules to the `other-modules` section.

JaredCorduan (approved):
I really like all these changes!

eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState.hs
Soupstraw: Done, and I added deprecated functions and a type synonym for WitHashes.

lehins (approved):
Looks awesome. Love this cleanup

PR #2897, opened 2022-07-06
Ensure Babbage TxOut decoder can't fail due to malformed Ptr

lehins:
It was only intended for TxBody to be causing a deserialization error
on malformed Ptr, not the TxOut in the LedgerState or TxBody in eras prior
to Babbage.


JaredCorduan (approved):
ah yes, great catch!

PR #2898, opened 2022-07-06
adopt a changelog and trunk based development

JaredCorduan:
This is long overdue, but from now on we will keep a changelog. The commit used by cardano-node for version `1.35.0` is `ce3057e0863304ccb3f79d78c77136219dc786c6`. I am considering that commit to be ledger version `1.0.0` (but for now just by using release branches), and the changelog begins after that commit.

Reviewers, feel free to look over the commits since `ce3057e0863304ccb3f79d78c77136219dc786c6` and see if you agree with which ones I have omitted (which should be exactly those that are invisible to anything downstream).

We will also adopt [trunk based development](https://trunkbaseddevelopment.com) (I've noted this in the CONTRIBUTING doc).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,46 @@
+# Changelog
+All notable changes to this repository will be documented in this file.
+At at later date we may adopt per-Haskell-package changelogs.
+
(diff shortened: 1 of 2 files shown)


lehins (approved):
:fireworks:

PR #2904, opened 2022-07-10
Avoid redundant copy of hashes to unpinned memory.

lehins:
It is more efficient to construct a `ByteString` `Builder` from smaller
`ShortByteString`s and then convert them into one strict `ByteString`
in one swoop for further hashing, since latter operation requires
pinned memory anyways.


JaredCorduan (approved):
:+1:

PR #2906, opened 2022-07-11
Ts haddock setalgebra

TimSheard:
Added Haddock to SetAlgebra, also to Data.BiMap, and Cardano.Ledger.UnifiedMap which have set algebra class instances.


JaredCorduan:
@TimSheard strangely CI thinks there is a redundant import, even though it seems you only touched comments...

JaredCorduan (approved):
:+1:

PR #2887, opened 2022-06-28
Remove dependency on cardano-prelude (except byron) 

teodanciu:
This removes the last remaining references to cardano-prelude (excluding byron), namely: 
  * HeapWords (usage from cardano-prelude replaced with that from cardano-base)
  * panic  

Because of the depedency on cardano-base, this PR depends on the merging of this one: https://github.com/input-output-hk/cardano-base/pull/280.  After the latter is merged, we have to update [cabal.project ](https://github.com/input-output-hk/cardano-ledger/blob/td/remove-cardano-prelude/cabal.project#L44)  with the corresponding tag and sha256  on this branch, before merging this PR. 

closes #1676

PR #2907, opened 2022-07-12
New FromCBOR for DPState

newhoggy:


eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState/DPState.hs
```diff
@@ -208,6 +209,12 @@ instance CC.Crypto crypto => FromSharedCBOR (DPState crypto) where
     dpsDState <- fromSharedPlusCBOR
     pure DPState {dpsPState, dpsDState}
 
+instance
+  ( CC.Crypto crypto
+  , FromSharedCBOR (DPState crypto)
+  ) => FromCBOR (DPState crypto) where
+  fromCBOR = evalStateT fromSharedPlusCBOR mempty
```
lehins, line 216: There is a specialized function for this already in `Data.Sharing`
```suggestion
  fromCBOR = fromNotSharedCBOR
```

cabal.project
```diff
@@ -110,6 +110,8 @@ constraints:
   -- ones they reuse the one from 'some', but there isn't e.g. a proper version
   -- constraint from dependent-sum-template (which is the library we actually use).
   , dependent-sum > 0.6.2.0
+  , algebraic-graphs < 0.7
```
lehins, line 113: I can't figure out why this upper bound is needed, which package depends on it? None of the packages in cardano-ledger directly depend on it, neither does the cardano-prelude. I wouldn't care if it was a lower bound, but upper bound can be problematic in the future.

lehins (changes requested):
We are trying to get rid of dependency on cardano-prelude (see #2887), but until we actually do get rid of it is good to stay compatible with newer versions. Thank you for the update

I assume separate `FromCBOR` instance for `DPState` is to make it possible to decode it CLI without decoding the whole of `LedgerState`. I am curious how it can be done, so ping me in a cardano-node PR once you have it ready.

cabal.project
newhoggy: It's a `plutus` dependency.  The bound is needed because `plutus` is missing the upper bound and won't compile with the latest package version.

newhoggy:
Closing.  This is superceded by https://github.com/input-output-hk/cardano-ledger/pull/2908

cabal.project
lehins: I see, plutus repo hasn't bumped the index yet, that's why they don't have that upper bound set. 

Makes sense, thanks.

PR #2909, opened 2022-07-12
Added tests for thunks in MockChainState

### LINKED ISSUE #2856: noThunks tests
The `ouroboros-network` repository contains [noThunks](https://hackage.haskell.org/package/nothunks) tests which sometime catch memory leaks in the ledger. It would be much better if there were tests in `cardano-ledger` which could catch memory leaks before they make their way to `ouroboros-network`.

The `ouroboros-network` noThunks tests take a while to run, and are probably better suited for a nightly test than CI tests.

I'll provide instructions here for running the `ouroboros-network` noThunks tests, so that we can use them as an example.

First make sure you have the following flags set in the `cabal` configuration (eg in `cabal.project`):

```
package io-sim-classes
  flags: +asserts +checktvarinvariant
```

Then start a nix shell and run the tests:

```
~/io/ouroboros-network$ nix-shell
[nix-shell:~/io/ouroboros-network]$ cabal v2-build -j4 ouroboros-consensus-shelley-test:test:test
./dist-newstyle/build/x86_64-linux/ghc-8.10.4/ouroboros-consensus-shelley-test-0.1.0.0/t/test/build/test/test --quickcheck-tests $N --iohk-enable-nightly-tests -p convergence
```
Setting `N=40` seems to work well when expecting failures, and `N=400` seems to work well when expecting no failures.

Soupstraw:
This PR adds tests that make sure that there are no unexpected thunks in the MockChainState after running a long trace. To run the nightly tests, use `cabal v2-test cardano-ledger-test --test-options="--scenario Nightly"`.

Currently we ignore thunks in the Plutus `EvaluationContext`, since we haven't yet bumped `plutus-core` to a version that adds a `NoThunks` instance to it (see #2853, [ plutus/#4687](https://github.com/input-output-hk/plutus/issues/4687)).

resolves #2856

### DIFF
--- a/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Rules/Utxo.hs
+++ b/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Rules/Utxo.hs
@@ -225,13 +225,10 @@ deriving stock instance
   Eq (UtxoPredicateFailure era)
 
 instance
-  ( Era era,
-    ToCBOR (Core.Value era),
-    ToCBOR (Core.TxOut era),
-    ToCBOR (Core.TxBody era),
-    NoThunks (Core.Value era),
-    NoThunks (Core.TxOut era),
-    NoThunks (PredicateFailure (Core.EraRule "UTXOS" era))
+  ( NoThunks (Core.Value era),
+    NoThunks (UTxO era),
+    NoThunks (PredicateFailure (Core.EraRule "UTXOS" era)),
+    NoThunks (Core.TxOut era)
   ) =>
   NoThunks (UtxoPredicateFailure era)
 
--- a/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Scripts.hs
+++ b/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Scripts.hs
@@ -87,7 +87,7 @@ import Data.Text (Text)
 import Data.Typeable (Proxy (..), Typeable)
 import Data.Word (Word64, Word8)
 import GHC.Generics (Generic)
-import NoThunks.Class (InspectHeapNamed (..), NoThunks (..), AllowThunksIn(..))
+import NoThunks.Class (AllowThunksIn (..), InspectHeapNamed (..), NoThunks (..))
 import Numeric.Natural (Natural)
 import Plutus.V1.Ledger.Api as PV1 hiding (Map, Script)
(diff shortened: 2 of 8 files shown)


libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/MockChain.hs
```diff
@@ -113,6 +116,14 @@ instance Show (MockBlock era) where
 instance Reflect era => TotalAda (MockChainState era) where
   totalAda (MockChainState nes _ _) = totalAda nes
 
+deriving instance Generic (MockChainState era)
+
+instance
+  ( Era era,
+    NoThunks (NewEpochState era)
+  ) =>
+  NoThunks (MockChainState era)
```
lehins: I don't think there is a reason to span over five lines of code something that can fit into one. Longer files are harder to read, we should save that real state if we can.

```suggestion
instance (Era era, NoThunks (NewEpochState era)) => NoThunks (MockChainState era)
```

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Rules/Utxo.hs
```diff
@@ -225,13 +225,10 @@ deriving stock instance
   Eq (UtxoPredicateFailure era)
 
 instance
-  ( Era era,
-    ToCBOR (Core.Value era),
-    ToCBOR (Core.TxOut era),
-    ToCBOR (Core.TxBody era),
```
lehins, line 231: Nice catch! Thank you!

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Scripts.hs
```diff
@@ -243,7 +247,10 @@ deriving via
   instance
     NoThunks PV1.EvaluationContext
 
-instance NoThunks CostModel
+-- instance NoThunks CostModel
```
```suggestion
```

lehins (commented):
Just a preliminary scan. It's looking good :wink:

JaredCorduan (commented):
this looks great @Soupstraw , thanks! We add the instructions for how to run these somewhere. Maybe in the `Performance` section of the the CONTRIBUTING doc?

Edit: I see that they are just a part of the Nightly test, maybe my comment doesn't make sense :thinking:

JaredCorduan (approved):
:+1:

PR #2911, opened 2022-07-12
WIP: Decouple `mint` from `Value`

teodanciu:
https://github.com/input-output-hk/cardano-ledger/issues/2805

in preparation for decoupling the usage of `Value` as a type for the
mint field in `TxBody`


JaredCorduan:
This is looking great! It will be interesting to see how easy it is to move the new `MultiAsset` into the transaction body. thanks!

PR #2914, opened 2022-07-13
Update nix packages

teodanciu:
required after the index state update.

This should fix running nix commands, inxluding `nix-shell`


JaredCorduan (approved):
thank you @teodanciu ! I can get a nix shell again :relieved:

lehins (approved):
One of the commits have already landed in master, so if you can rebase it we can get the instructions merged as well

PR #2916, opened 2022-07-14
Include missing files into cabal

lehins:
Also uses the commit from #2914 to update nix definitions


JaredCorduan (approved):
Many thanks for taking on this un-fun task! looks great to me

PR #2918, opened 2022-07-15
Bumped plutus version, enabled NoThunks tests for EvaluationState

### LINKED ISSUE #2853: CostModel data definition has non-strict fields and wrong NoThunks instance

Soupstraw:
resolves #2853

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -77,8 +77,8 @@ source-repository-package
 source-repository-package
   type: git
   location: https://github.com/input-output-hk/plutus
-  tag: f680ac6979e069fcc013e4389ee607ff5fa6672f
-  --sha256: 180jq8hd0jlg48ya7b5yw3bnd2d5czy0b1agy9ng3mgnzpyq747i
+  tag: 8ab4c3355c5fdf67dcf6acc1f5a14668d5e6f0a9
+  --sha256: coD/Kpl7tutwXb6ukQCH5XojBjquYkW7ob0BWZtdpok=
   subdir:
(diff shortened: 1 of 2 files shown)


JaredCorduan (approved):
Looks good!

It would be good if we could consolidate the commits before merging. maybe just put "Removed an out-of-date comment" in with the plutus bump? And give a bit more description to "Cleanup" (like "update nix sources" or some such).

PR #2921, opened 2022-07-18
WIP ValidateScript can only be applied to phase 1 scripts, enforced by types.

TimSheard:
Changed ValidateScript, so that validateScript has type
PhasedScript 'One era -> Core.Tx era -> Bool
Type ensure only scriptes from Phase1 can be validated.
Introduces types Phase, PhasedScript, and type family SomeScript

### DIFF
--- a/eras/alonzo/impl/src/Cardano/Ledger/Alonzo.hs
+++ b/eras/alonzo/impl/src/Cardano/Ledger/Alonzo.hs
@@ -37,7 +37,7 @@ import Cardano.Ledger.Alonzo.Rules.Utxo (utxoEntrySize)
 import qualified Cardano.Ledger.Alonzo.Rules.Utxo as Alonzo (AlonzoUTXO)
 import qualified Cardano.Ledger.Alonzo.Rules.Utxos as Alonzo (UTXOS)
 import qualified Cardano.Ledger.Alonzo.Rules.Utxow as Alonzo (AlonzoUTXOW)
-import Cardano.Ledger.Alonzo.Scripts (Script (..), isPlutusScript)
+import Cardano.Ledger.Alonzo.Scripts (Script (..))
 import Cardano.Ledger.Alonzo.Tx (ValidatedTx (..), alonzoInputHashes, minfee)
 import Cardano.Ledger.Alonzo.TxBody (TxBody, TxOut (TxOut), getAlonzoTxOutEitherAddr)
 import Cardano.Ledger.Alonzo.TxInfo (ExtendedUTxO (..), alonzoTxInfo, validScript)
@@ -48,6 +48,7 @@ import Cardano.Ledger.BaseTypes (BlocksMade (..))
 import Cardano.Ledger.Coin
 import qualified Cardano.Ledger.Core as Core
 import qualified Cardano.Ledger.Crypto as CC
+import Cardano.Ledger.Era (Phase (..), PhaseRep (..), PhasedScript (..), SomeScript)
 import qualified Cardano.Ledger.Era as EraModule
 import Cardano.Ledger.Keys (GenDelegs (GenDelegs))
 import qualified Cardano.Ledger.Mary.Value as V (Value)
@@ -81,15 +82,15 @@ import qualified Cardano.Ledger.Shelley.Rules.Rupd as Shelley
 import qualified Cardano.Ledger.Shelley.Rules.Snap as Shelley
 import qualified Cardano.Ledger.Shelley.Rules.Tick as Shelley
 import qualified Cardano.Ledger.Shelley.Rules.Upec as Shelley
-import qualified Cardano.Ledger.Shelley.Tx as Shelley
 import Cardano.Ledger.Shelley.UTxO (balance)
 import Cardano.Ledger.ShelleyMA.Rules.Utxo (consumed)
-import Cardano.Ledger.ShelleyMA.Timelocks (validateTimelock)
+import Cardano.Ledger.ShelleyMA.Timelocks (Timelock, validateTimelock)
 import Cardano.Ledger.Val (Val (inject), coin, (<->))
 import Control.Arrow (left)
 import Control.Monad.Except (liftEither)
 import Control.Monad.Reader (runReader)
 import Control.State.Transition.Extended (TRC (TRC))
+import Data.ByteString.Short (ShortByteString)
 import Data.Default (def)
 import Data.Foldable (toList)
 import qualified Data.Map.Strict as Map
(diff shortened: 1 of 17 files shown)


libs/cardano-ledger-core/src/Cardano/Ledger/Language.hs
```diff
@@ -0,0 +1,40 @@
+{-# LANGUAGE DeriveGeneric #-}
```
lehins, line 1: It would have been best if this module was moved with `git mv`, so the git history was preserved.

libs/cardano-ledger-core/src/Cardano/Ledger/Era.hs
```diff
@@ -328,3 +342,44 @@ others where the concrete type (Update and WitnessSet) will have to be made into
...
    -- HasField "exUnits" (Core.Tx era) ExUnits,
 -}
+-- ====================================================
+-- Reflecting the phase of scripts into types
+-- ====================================================
+
+-- | There are only two Phases
+data Phase
+  = One -- simple scripts run in phase 1
+  | Two -- smart scripts (Plutus) run in phase 2
```
lehins: Just to avoid any potential clashes with such common names and make it more specific I'd like to suggest a common prefix:
```suggestion
  = PhaseOne -- simple scripts run in phase 1
  | PhaseTwo -- smart scripts (Plutus) run in phase 2
```

Alternatively to reduce on typing and promote consistency we can use numbers everywhere:

```haskell
  = Phase1 -- simple scripts run in phase 1
  | Phase2 -- smart scripts (Plutus) run in phase 2
```
and
```haskell
data PhaseRep t where
  Phase1Rep :: PhaseRep 'Phase1
  Phase2Rep :: PhaseRep 'Phase2
```

libs/cardano-ledger-core/src/Cardano/Ledger/Era.hs
```diff
@@ -328,3 +342,44 @@ others where the concrete type (Update and WitnessSet) will have to be made into
...
+
+-- | A GADT that witnesses the two Phases
+data PhaseRep t where
+  Phase1 :: PhaseRep 'One
+  Phase2 :: PhaseRep 'Two
```
lehins:
```suggestion
  PhaseOneRep :: PhaseRep 'PhaseOne
  PhaseTwoRep :: PhaseRep 'PhaseTwo
```

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/PlutusScriptApi.hs
```diff
@@ -185,15 +168,22 @@ collectTwoPhaseScriptInputs ei sysS pp tx utxo =
...
+              Just (Phase2Script lang bytes) -> Just (neededPurpose, lang, bytes)
+              Nothing -> Nothing
+        )
+        (scriptsNeeded utxo tx)
     redeemer (sp, lang, _) =
       case indexedRdmrs tx sp of
         Just (d, eu) -> Right (lang, sp, d, eu)
         Nothing -> Left (NoRedeemer sp)
-    getscript (_, _, script) = script
+    getscript (_, _, scriptbytes) = scriptbytes
```
lehins: I like the name change, but lets stick to conventions
```suggestion
    getscript (_, _, scriptBytes) = scriptBytes
```

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Tools.hs
```diff
@@ -135,17 +134,25 @@ evaluateTransactionExecutionUnits pp tx utxo ei sysS costModels = do
     txb = getField @"body" tx
     ws = getField @"wits" tx
     dats = unTxDats $ getField @"txdats" ws
-    scriptsAvailable = txscripts utxo tx
+    scriptsAvailable = getPhase2 @era (txscripts utxo tx)
     needed = scriptsNeeded utxo tx
-    neededAndConfirmedToBePlutus = mapMaybe (knownToNotBe1Phase scriptsAvailable) needed
+    neededAndConfirmedToBePlutus =
+      -- mapMaybe (knownToBe2Phase scriptsAvailable) needed
```
lehins: Dead code
```suggestion
```

lehins (commented):
Overall this looks very nice, just a few minor naming suggestions. 

I am not sure if that is what @JaredCorduan had in mind, so I'd like for him to review this PR before we merge it.

eras/alonzo/impl/src/Cardano/Ledger/Alonzo.hs
```diff
@@ -122,15 +123,20 @@ instance API.ShelleyEraCrypto c => API.ApplyTx (AlonzoEra c) where
...
-instance (CC.Crypto c) => Shelley.ValidateScript (AlonzoEra c) where
-  isNativeScript x = not (isPlutusScript x)
+type instance SomeScript 'One (AlonzoEra c) = Timelock c
+
+type instance SomeScript 'Two (AlonzoEra c) = (Language, ShortByteString)
+
+instance (CC.Crypto c) => EraModule.ValidateScript (AlonzoEra c) where
+  phaseScript Phase1 (TimelockScript s) = Just (Phase1Script s)
+  phaseScript Phase2 (PlutusScript lang bytes) = Just (Phase2Script lang bytes)
+  phaseScript _ _ = Nothing
```
JaredCorduan, line 133: I think all these changes are a huge improvement, thanks! Here is one ponderable, something perhaps for another PR:

Can we parse the scripts at an earlier phase, and preferably in only one spot, so that this `phaseScript _ _ = Nothing` is not possible? ie have the type checker rule this scenario out for us.

JaredCorduan (approved):
Looks great, this is a big improvement, thanks @TimSheard ! I'll approve, but I agree with all of @lehins 's comments. I also left a ponderable in a comment, something to think about next.

eras/alonzo/impl/src/Cardano/Ledger/Alonzo.hs
TimSheard: I am not sure this is possible without changing the type of witnesses, as they keep all the scripts together in one map. Changing that would be a very invasive change.

libs/cardano-ledger-core/src/Cardano/Ledger/Language.hs
TimSheard: Is there any way to undo that?

JaredCorduan:
replaced by #2953

PR #2924, opened 2022-07-20
Replace deprecated testProperty with our own version.

TimSheard:
Re[aces al calls to Hedghog testProperty with our own testProperty written in terms of testPropertyNamed.

Since most of these were in 3 different modules in the byron era, and none of them have a common module they all import from. I had to write our version of testProperty several times.


lehins (approved):
Looks great, thank you. Can you also remove this ghc flag, it was added unnecessarily:

https://github.com/input-output-hk/cardano-ledger/blob/2e9f8b4e81612886c05328a923ec3a4545fa69c3/libs/small-steps-test/test/examples/Main.hs#L2-L4

JaredCorduan (approved):
thanks for cleaning this up!

PR #2925, opened 2022-07-21
Reduce duplication in the ToJSON instance by factoring out the common part

lehins:
We need to avoid copy pasting code as much as possible, especially when it involves things that can't be checked statically, for example json field names. Here we extract key value pairs construction into a separate function and reuse it in both `toJSON` and in `toEncoding` functions.

### DIFF
--- a/eras/shelley/impl/src/Cardano/Ledger/Shelley/Genesis.hs
+++ b/eras/shelley/impl/src/Cardano/Ledger/Shelley/Genesis.hs
@@ -149,63 +149,49 @@ sgActiveSlotCoeff :: ShelleyGenesis era -> ActiveSlotCoeff
 sgActiveSlotCoeff = mkActiveSlotCoeff . sgActiveSlotsCoeff
 
 instance Era era => ToJSON (ShelleyGenesis era) where
-  toJSON sg =
-    Aeson.object
-      [ "systemStart" .= sgSystemStart sg,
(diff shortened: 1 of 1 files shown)


eras/shelley/impl/src/Cardano/Ledger/Shelley/Genesis.hs
```diff
@@ -149,63 +149,49 @@ sgActiveSlotCoeff :: ShelleyGenesis era -> ActiveSlotCoeff
...
+      sgSlotLength,
+      sgUpdateQuorum,
+      sgMaxLovelaceSupply,
+      sgProtocolParams,
+      sgGenDelegs,
+      sgInitialFunds,
+      sgStaking
+    } =
+    let !strictSgInitialFunds = sgInitialFunds
+        !strictSgStaking = sgStaking
```
teodanciu, line 178: The fact that these two fields will be strict in `toJSON` after this change has no impact, correct?

teodanciu (approved):
Looks good to me, just a small question

eras/shelley/impl/src/Cardano/Ledger/Shelley/Genesis.hs
lehins: Yeah, it has no impact on the `toEncoding`, but for the `toJSON` it actually makes it is better.

JaredCorduan:
great catch, thanks @lehins !

PR #2935, opened 2022-07-26
fix partiality in FromCBOR instance of Language

JaredCorduan:
Thanks to @lehins for the clever cake-and-eat-it-too solution to having the type checker notice new language versions in the `FromCBOR` instance.

### DIFF
--- a/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Language.hs
+++ b/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Language.hs
@@ -5,8 +5,9 @@
 --     versions of old languages) will be added here.
 module Cardano.Ledger.Alonzo.Language where
 
-import Cardano.Binary (FromCBOR (..), ToCBOR (..), decodeWord64)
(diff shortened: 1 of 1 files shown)


eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Language.hs
```diff
@@ -31,10 +32,21 @@ instance NoThunks Language
...
-  toCBOR = toCBOR . fromEnum
+  toCBOR PlutusV1 = toCBOR (0 :: Int)
+  toCBOR PlutusV2 = toCBOR (1 :: Int)
 
 instance FromCBOR Language where
-  fromCBOR = toEnum . fromIntegral <$> decodeWord64
+  fromCBOR = do
+    n <- decodeInt
+    lang <-
+      if n <= fromEnum (maxBound :: Language) && n >= fromEnum (maxBound :: Language)
```
lehins:
```suggestion
      if n >= fromEnum (minBound :: Language) && n <= fromEnum (maxBound :: Language)
```

lehins (changes requested):

lehins (approved):

PR #2936, opened 2022-07-27
the alonzo UTxO rule to use alonzo minfee function

### LINKED ISSUE #4228: [BUG] - 1.35.* v7 nodes on Mainnet are picking up the wrong parent-block - Vasil HF Blocker?
### Strange Issue

So, i will post a bunch of pics here, screenshots taken from pooltool.io. Blockflow is from botton->top. These are only a few examples, there are many more!

**This is happening on mainnet right now** running 1.35.* nodes. There are many occasions with double & tripple height battles where **newer v7 1.35.\* nodes** are **picking up the wrong parent-block** and try to build on another v7 block. So v6 1.34.* nodes are winning those all the time. 

I personally had the issue loosing 10 height-battles within a 1-2 day window against v6 nodes. That was 100% of all my height-battles i had. 

Its always the same pattern: There is a height battle that a v7 node looses against a v6 node. If there are also two nodes scheduled for the next block and one of them is a v7 node, it picks up the wrong lost block hash from the previous v7 node and builds on it. Of course it looses against the other v6 node which is building on the correct block. But as you can see in the example below, this can span multiple slotheights/blocks ⚠️ 

This is a Vasil-HF blocker IMO, because it would lead to the situation that SPOs are only upgrading to 1.35.* at the last possible moment before the HF, giving the ones staying on 1.34.1 an advantage. Not a good idea, it must be sorted out before. Q&A team please start an investigation on that asap, thx! 🙏 

image
-----
image
-----
Here is a nightmare one, v7 built ontop of other v7 node (green path):
image
-----
image
-----
image
-----
image
-----
image
-----
image

JaredCorduan:
A bug was introduced, causing the Alonzo UTxO rule to use the Shelley `minfee` function instead of the Alonzo version. This PR reverts that.

Additionally, a new golden test for the Alonzo fee calculation has been added, using the block from:
https://github.com/input-output-hk/cardano-node/issues/4228#issuecomment-1195707491

I will cherry pick the commit in this PR onto the release branch `release/1.0.0` after this PR is merged.

resolves https://github.com/input-output-hk/cardano-node/issues/4228


eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Rules/Utxo.hs
```diff
@@ -295,9 +294,11 @@ feesOK pp tx (UTxO utxo) =
       -- restrict Utxo to those inputs we use to pay fees.
       utxoCollateral = eval (collateral ◁ utxo)
       bal = balance @era (UTxO utxoCollateral)
+      theFee = txBody ^. feeTxBodyL
```
teodanciu, line 297: The first usage of our new lens friends  :grinning:

teodanciu (approved):
Looks good to me. 

I was going through the commit that introduced the problem, and was wondering about two other cases that are reusing a Shelley function  (and also using `pp`, which from my understanding was the problem - I might be wrong!) 

- [ShelleyMA.validateValueNotConservedUTxO](https://github.com/input-output-hk/cardano-ledger/commit/08021e0241da05caac6e625de8fb2d26b583b3be#diff-3e5cc03f4161411e588f84cc28f7b6d7242d20bba77c3ed52f017866868edb76R408) .    But the original version is still using the `Shelley.produced` , so maybe not a problem? 
- [Shelley.validateMaxTxSizeUTxO](https://github.com/input-output-hk/cardano-ledger/commit/08021e0241da05caac6e625de8fb2d26b583b3be#diff-3e5cc03f4161411e588f84cc28f7b6d7242d20bba77c3ed52f017866868edb76R440)  - is this using the maxTxSize from Shelley? 

Anyway, just some questions  based on perhaps vague understanding- sorry if I'm barking up the wrong tree  here!

JaredCorduan:
Thank you so much @teodanciu for checking out the rest of the refactor!

* The use of `ShelleyMA.validateValueNotConservedUTxO` looks fine to me, for the reason you stated, it uses `Shelley.produced` as before.
* The use of `Shelley.validateMaxTxSizeUTxO` also looks fine to me, `maxTxSize` is a protocol parameter that has not changed since Shelley.

Thanks again for looking out for more trouble!

eras/alonzo/test-suite/test/Test/Cardano/Ledger/Alonzo/Golden.hs
```diff
@@ -180,6 +190,36 @@ goldenSerialization =
...
+        hex <- readDataFile "golden/hex-block-node-issue-4228.cbor"
+        let cborBlock = fromRight mempty (B16L.decode hex)
+            (_leftover, Annotator f) =
+              fromRight (error "bad golden block 4228") $
+                deserialiseFromBytes fromCBOR cborBlock
+            _block :: Block (BHeader StandardCrypto) (AlonzoEra StandardCrypto)
+            _block@(Block _header txs) = f (Full cborBlock)
+            txs' = fromTxSeq @(AlonzoEra StandardCrypto) txs
+            firstTx = head $ toList txs'
+
```
TimSheard: we should say whether we expect this test to succeed or fail.

TimSheard (commented):
Looks good to me.

PR #2901, opened 2022-07-08
Restructure the Core for consistency, clarity and new functionality

lehins:
# Changes in this humongous PR:

### New classes:

```
EraSegWits --> EraTx --> EraTxBody --> EraTxOut --> Era
                     \              `--> EraPParams --> Era
                       `--> EraWitnesses --> EraScript --> Era
                        `--> EraAuxiliaryData --> Era
```

Note that all of the `Era*` classes have a few properties:

* depend either directly or transitively on the `Era` type class.
* define an associated type family (which was previously standalone).
* add all constraints that the associated type must implement.

This means that any function that uses an associated type family can safely use
the corresponding type class. For example a function that accepts `TxBody` as
argument can always set `EraTx` as a constraint, which will bring all of the
superclasses as constraints as well.

Loose corespondance of new classes to the old removed concepts:

* `Era` class is the same as before with the addition of type level protocol
  version and deprecation of all of the functions
* `EraTxOut` has  all superclasses as in `UsesTxOut` type  synonym constraint as
  well  as defines  lenses that  supercede all  functions previously  defined in
  `Era` type class
* `EraTxBody` has all superclasses as in `UsesTxBody` type synonym constraint
* `EraAuxiliaryData` has all superclasses as in `UsesAuxiliaryData` type synonym
  constraint and consolidates `ValidateAuxiliaryData` into it
* `EraPParams` replaces `UsesPParams` type class now with an associated type family
* `EraScript` has replaced `UsesScript` type synonym constraint and the
  `ValidateScripts` type class, with an exception of `validateScript` function,
  which was moved into `EraTx` type class, since it depends on the `Tx`

### Lenses

This PR expands usage of
[`microlens`](https://hackage.haskell.org/package/microlens) package.

Each class defines lenses for associated types, all of which are getters and
potentially setter. They also define a way to construct a minimal value of an
asscoiated type. For example if we want to construct a function that will
construct a `Tx` with a new `TxBody`, which uses `AuxiliaryData` from another
`Tx`:

```haskell
newTxWithOtherAuxData :: EraTx era => TxBody era -> Tx era -> Tx era
newTxWithOtherAuxData txBody tx =
  mkBasicTx txBody & auxDataTxL .~ (tx ^. auxDataTxL)
```

Above function is guaranteed to work with transactions of all eras.

Some are just getters, which simply relay some information from one or more
fields of a concrete type, altohugh they might do some minimal computation. For
example `allInputsTxBodyF` will return all of the inputs present in the
[...]

PR #2938, opened 2022-07-28
the alonzo UTxO rule to use alonzo minfee function

JaredCorduan:
This PR cherry picks the commit from #2936 for the release branch `release/1.0.0`. The cherry pick required a bit of massaging due to the glorious recent cleanup in #2901, but can be compared line by line with #2936 to check that is is semantically the same.


eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Rules/Utxo.hs
```diff
@@ -272,8 +271,10 @@ feesOK ::
     Core.Tx era ~ Alonzo.ValidatedTx era,
     -- "collateral" to get inputs to pay the fees
     HasField "collateral" (Core.TxBody era) (Set (TxIn (Crypto era))),
+    HasField "txrdmrs" (Core.Witnesses era) (Redeemers era),
```
teodanciu, line 274: This seems to be missing from the original PR..but probably not relevant

teodanciu (approved):
Diffing the diffs, I noticed that this PR adds  `HasField "txrdmrs" (Core.Witnesses era) (Redeemers era)` to `feesOK` function, whereas #2936 didn't. 
Also, in #2936,  `HasField "_prices" ` was added as constraints to `utxoTransition` and `  STS (AlonzoUTXO era)` instance, whereas in this PR we're only adding it to `feesOK`.   Perhaps these are necessary because of the other changes that were left behind on this branch!  
The rest looks the same to me :+1:

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Rules/Utxo.hs
JaredCorduan: yea, this was made obsolete by https://github.com/input-output-hk/cardano-ledger/pull/2901

JaredCorduan:
I am going to merge, even though `ci/hydra:Cardano:cardano-ledger-specs:required` is reporting `Pending`. If you follow the link, all the tests have actually passed.

PR #2942, opened 2022-07-28
better comment for the PVCannotFollowPPUP failure

### LINKED ISSUE #2939: Improve comment on PVCannotFollowPPUP
```
  | PVCannotFollowPPUP
      !ProtVer -- the first bad protocol version
```

"the first bad protocol version` is pretty  opaque to those that don't work directly on the ledger. This comment could be improved to explain what the error constructor means.

JaredCorduan:
better comment for the PVCannotFollowPPUP failure

resolves #2939


teodanciu (approved):
Looks like an undeniably better comment to me!

JaredCorduan:
> Looks like an undeniably better comment to me!

@newhoggy noticed that we could make it even better by using a haddock. I will do that shortly...

teodanciu:
Great point

PR #2945, opened 2022-08-02
Placeholder package for the new era

### LINKED ISSUE #2946: Barebones conway era
- [x] Create a new cabal package for the conway ledger era.
- [x] Create the rules mapping for the conway era, re-using all the babbage rules
- [x] Create all the class instance required for the conway era, re-using all the babbage logic
- [x] Create the babbage to conway translation, which will initially be entirely the identity map

lehins:
After some discussion the choice of the name was settled on `Conway`, after a great mathematician John Horton Conway.

This package just creates a bare bone `cardano-ledger-conway` package, that can be further filled with features, but as it stands right now it is identical in functionality to Babbage.

There was some minor adjustments done to previous eras in order to reduce code copy for the new era.

Fixes: #2946

### DIFF
--- a/eras/babbage/impl/src/Cardano/Ledger/Babbage/Scripts.hs
+++ b/eras/babbage/impl/src/Cardano/Ledger/Babbage/Scripts.hs
@@ -1,4 +1,4 @@
-{-# LANGUAGE BangPatterns #-}
+{-# LANGUAGE AllowAmbiguousTypes #-}
 {-# LANGUAGE DataKinds #-}
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE OverloadedStrings #-}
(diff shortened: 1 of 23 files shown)


eras/babbage/impl/src/Cardano/Ledger/Babbage/Scripts.hs
```diff
@@ -1,4 +1,4 @@
-{-# LANGUAGE BangPatterns #-}
+{-# LANGUAGE AllowAmbiguousTypes #-}
```
JaredCorduan, line 1: I'm curious why `babbageScriptPrefixTag` needed this extension...

JaredCorduan (approved):
Looks great!

lehins:
That is because Script type family is not injective, yet, which means era parameter is ambiguous

PR #2949, opened 2022-08-02
Era diagrams

teodanciu:
Added some diagrams showing relationships between types from specific eras and how they relate to the ones from Core. 
It started as a small thing to help me understand better, and I think things got a bit out of hand. 

The degree of precision is rather arbitrary  - a combination of what I thought is important and what fitted on a human-readable picture. 
Not sure if useful/maintainable enough to be merged.  The diagrams have "source code" in text format (created with PlantUML), so checking those in could also be an option, but of course, that brings another set of problems with it. 

Please feel free to suggest a better place for them,  or even close this PR if it's too unmaintainable..


JaredCorduan:
The diagrams are beautiful @teodanciu ! I do have a knee-jerk reaction to commit binary data, and would be inclined to commit the source, but I can imagine it maybe being trouble to maintain. :thinking:

PR #2950, opened 2022-08-02
Add `FromSharedCBOR` constraint to the `EraTxOut`

lehins:
In order to not propagate this `FromSharedCBOR` constraint all the way to consensus it should be as a superclass of `EraTxOut` (before it was superclass of `ShelleyBasedEra` synonym, which was removed in #2901)


JaredCorduan (approved):
:+1:

lehins:
Seems like this PR has discovered a bug in GHC: https://gitlab.haskell.org/ghc/ghc/-/issues/21973 A work around has been applied

PR #2951, opened 2022-08-02
New keysSet function for Data.VMap

newhoggy:


libs/vector-map/src/Data/VMap.hs
```diff
@@ -244,6 +247,10 @@ keys :: VG.Vector kv k => VMap kv vv k v -> [k]
 keys = VG.toList . keysVector . unVMap
 {-# INLINE keys #-}
 
+keysSet :: (Ord k, VG.Vector kv k) => VMap kv vv k v -> Set k
+keysSet = Set.fromList . keys
```
lehins, line 251: You can use `Set.fromDistinctAscList` here, because keys are unique and sorted.

newhoggy:
Closing.  I ended up not using this new function.

lehins (changes requested):
This seems like a useful function. One comment that will make it faster.

There are a couple of things I'd like to ask:

* A roundtrip test in libs/vector-map/test/Test/VMap.hs
* an entry in changelog

lehins:
@newhoggy I left my review just in case you will change your mind on this PR :wink:

PR #2952, opened 2022-08-02
Added a hash field to MemoBytes

### LINKED ISSUE #2882: Show and Hash for MemoBytes
It was brought to our attention by @sevanspowell that it is possible for two transactions to not be equal, while the string produced by the `show` function are identical. 

> i.e. `show txA == show txB` but `txA /= txB` 

This is actually true due to the serialization, which does not participate in the Show instance. In fact this is totally legit since there is no law in Haskell that regulates that `show txA == show txB <=> txA == txB` property holds for all types. Nevertheless, I do agree that this is inconvenient, at least for the purpose of testing/debugging.

It is done this way for a very important reason: we must preserve the binary serialization in the original form to ensure cryptographic integrity. The bit of context that is important that there is no normal/canonical form for CBOR serialization. Whenever we get a valid transaction over the wire, it could have been constructed using some other CBOR implementation and thus can have slightly different binary representation from what it would be if we where to construct it and serialize it using ledger code. In other words: it is absolutely possible and valid to have to transactions be "sematically" the same but not equal. Whenever we compute the hashes of such transactions they will be different, thus different transaction, despite having exactly the same values in a Haskell Tx type.

The best solution I think is to also display the hash of the thing that is being memoized in hex format. This will help us with:

* Avoiding hash recomputation whenever we need it more than once
* Reduce number of bytes copied (each hash computation requires an extra buffer copy to pinned memory)
* Account for difference in binary representation when displaying the thing that is being memoized
* It would have also helped us fixing the `Ord` instance, however we decided to simply remove it in #2881 

So the proposed solution is to:

__Disclaimer__ - this was in browser programming, so it is bound to be incomplete or even incorrect:

* Move `MemoBytes` into `cardano-ledger-core` 
* Add hash into the type next to the bytes (must be lazy):
```haskell
data MemoBytes t era = Memo
  { memoType :: !t era,
    memoBytes :: ShortByteString,
    memoHash :: SafeHash (Crypto era) (MemoHashIndex t)
  }

type family MemoHashIndex (t :: Type -> Type) :: Type

-- | Smart constructor
mkMemoBytes :: forall era t. (SafeToHash bs, Era era) => t era -> BSL.ByteString -> MemoBytes t era
mkMemoBytes t bsl = 
  MemoBytes t (toShort bs) (makeHashWithExplicitProxys (Proxy @(Crypto era)) (Proxy @(MemoHashIndex t)) bs)
  where
    bs = toStrict bsl
```

* Move `hashAnnotated` function into `HashAnnotated` class, so it can be customized

Then instances would look something like this:
```haskell
data TxBodyRaw era = TxBodyRaw
  { _inputs :: !(Set (TxIn (Crypto era))),
  ...
  }
newtype TxBody era = TxBodyConstr (MemoBytes TxBodyRaw era)
  deriving newtype (SafeToHash)

type instance MemoHashIndex TxBodyRaw = EraIndependentTxBody

instance (c ~ Crypto era) => HashAnnotated (TxBody era) EraIndependentTxBody c where
  hashAnnotated (TxBodyConstr mb) = memoHash mb
```

Soupstraw:
This PR adds a hash of the byte string to the MemoBytes data structure. This will make it easier to spot if the byte string representations of two MemoBytes values are different by calling `show` on MemoBytes.

I also changed the type signature of `MemoBytes t` to `MemoBytes t era` and made all the necessary changes in other files. Some other type signatures also had to be changed as a consequence:
 * `Mem t` -> `Mem t era`
 * `Timelock crypto` -> `Timelock era`
 * `MultiSig crypto` -> `MultiSig era`

The `MemoBytes` constructor and destructor are now behind a bidirectional pattern `Memo`. Since we have to calculate the hash based on the current era and crypto, the pattern has a constraint `Era era`, which had to be added to a lot of function signatures across the codebase.

closes #2882


lehins:
@Soupstraw looks like you [are hitting this bug as well](https://gitlab.haskell.org/ghc/ghc/-/issues/21973). Can't quite pinpoint where exactly is the problem, but the tests indicate that it is in the `Data` type.

Soupstraw:
@lehins It turned out to be a bug in my implementation of `FromCBOR` for the data wrapper for `Plutus.Data` 😅

libs/cardano-ledger-core/src/Cardano/Ledger/MemoBytes.hs
```diff
@@ -60,48 +65,85 @@ import Prelude hiding (span)
...
-
-deriving instance Generic (MemoBytes t)
-
-instance (Typeable t) => ToCBOR (MemoBytes t) where
-  toCBOR (Memo _ bytes) = encodePreEncoded (fromShort bytes)
-
-instance (Typeable t, FromCBOR (Annotator t)) => FromCBOR (Annotator (MemoBytes t)) where
+data MemoBytes t era = Memo'
+  { memoType :: !(t era),
+    memoByteString :: ShortByteString,
```
lehins: `memoByteString` is misleading, since it is not really a `ByteString`.
```suggestion
    memoBytes :: ShortByteString,
```

Alternatively you can call it `memoShortByteString`, which would be accurate, but a bit long for my taste. It's up to you which one you wanna choose, but `memoByteString` is not good IMHO

eras/shelley-ma/test-suite/src/Test/Cardano/Ledger/AllegraEraGen.hs
```diff
@@ -132,22 +132,22 @@ instance Mock c => MinGenTxout (AllegraEra c) where
...
-unQuantifyTL :: CryptoClass.Crypto crypto => Quantifier (Timelock crypto) -> Timelock crypto
+unQuantifyTL :: Era era => Quantifier (Timelock era) -> Timelock era
 unQuantifyTL (AllOf xs) = RequireAllOf (fromList xs)
 unQuantifyTL (AnyOf xs) = RequireAnyOf (fromList xs)
 unQuantifyTL (MOf n xs) = RequireMOf n (fromList xs)
 unQuantifyTL (Leaf t) = t
 
 genAuxiliaryData ::
-  Mock crypto =>
+  (Arbitrary (Core.AuxiliaryData era)) =>
```
lehins: Redundant parenthesis
```suggestion
  Arbitrary (Core.AuxiliaryData era) =>
```

libs/cardano-ledger-core/src/Cardano/Ledger/MemoBytes.hs
```diff
@@ -60,48 +65,85 @@ import Prelude hiding (span)
...
+  where
   fromCBOR = do
     (Annotator getT, Annotator getBytes) <- withSlice fromCBOR
-    pure (Annotator (\fullbytes -> Memo (getT fullbytes) (toShort (toStrict (getBytes fullbytes)))))
+    pure (Annotator (\fullbytes -> mkMemoBytes (getT fullbytes) (getBytes fullbytes)))
+
+instance Eq (MemoBytes t era) where (Memo' _ x _) == (Memo' _ y _) = x == y
 
-instance Eq (MemoBytes t) where (Memo _ x) == (Memo _ y) = x == y
+instance Show (t era) => Show (MemoBytes t era) where show (Memo' y _ h) = show y <> "(hash " <> show h <> ")"
```
lehins: Make it multiline and tell the world what type of hash it is instead of just `hash`:
```suggestion
instance Show (t era) => Show (MemoBytes t era) where
  show (Memo' y _ h) = show y <> " (" <> hashAlgorithmName (Just h) <> ": " <> show h <> ")"
```

libs/cardano-ledger-core/src/Cardano/Ledger/MemoBytes.hs
```diff
@@ -60,48 +65,85 @@ import Prelude hiding (span)
...
-memoBytes :: Encode w t -> MemoBytes t
-memoBytes t = Memo (runE t) (shorten (toLazyByteString (encode t)))
+memoBytes :: Era era => Encode w (t era) -> MemoBytes t era
+memoBytes t = mkMemoBytes (runE t) (toLazyByteString (encode t))
 
 -- | Try and deserialize a MemoBytes, and then reconstruct it. Useful in tests.
-roundTripMemo :: (FromCBOR t) => MemoBytes t -> Either Codec.CBOR.Read.DeserialiseFailure (Lazy.ByteString, MemoBytes t)
-roundTripMemo (Memo _t bytes) =
-  case deserialiseFromBytes fromCBOR (fromStrict (fromShort bytes)) of
+roundTripMemo :: (FromCBOR (t era), Era era) => MemoBytes t era -> Either Codec.CBOR.Read.DeserialiseFailure (Lazy.ByteString, MemoBytes t era)
```
lehins: Let's fix this ugly qualification too by explicitly importing it:
```suggestion
roundTripMemo :: (FromCBOR (t era), Era era) => MemoBytes t era -> Either DeserialiseFailure (Lazy.ByteString, MemoBytes t era)
```

lehins (changes requested):
This looks great!  Besides suggestions in the review there are two main comments:

### One last part of the original ticket #2882 is still missing

If you'd rather do it as a follow up PR that is fine too, but I'd recommend do all of those related changes in one swoop"

* Move `hashAnnotated` function into `HashAnnotated` class, so it can be customized

Then instances would look something like this:
```haskell
data TxBodyRaw era = TxBodyRaw
  { _inputs :: !(Set (TxIn (Crypto era))),
  ...
  }
newtype TxBody era = TxBodyConstr (MemoBytes TxBodyRaw era)
  deriving newtype (SafeToHash)

type instance MemoHashIndex TxBodyRaw = EraIndependentTxBody

instance (c ~ Crypto era) => HashAnnotated (TxBody era) EraIndependentTxBody c where
  hashAnnotated (TxBodyConstr mb) = memoHash mb
```

### We can make script type family injective

Thanks to switching of all concrete script types to be parametrized by `era` instead of `crypto` we can switch: 
https://github.com/input-output-hk/cardano-ledger/blob/b44a3b8959c00e8243c3004c01adf412d1b2d317/libs/cardano-ledger-core/src/Cardano/Ledger/Core.hs#L385

to
```haskell
  type Script era = (r :: Type) | r -> era
```

which will improve type inference

libs/cardano-ledger-core/src/Cardano/Ledger/MemoBytes.hs
Soupstraw: The issue here is that there is also a function named `memoBytes` in the same file and the field was previously named `memobytes` in lower caps, which is inconsistent with the way we usually name fields. If I renamed the field `memoBytes`, I'll also have to rename the other function and then change all the references to that function as well. I even tried that, but there were a lot of files that were using that while the field was not really used anywhere else AFAIK.

lehins: I see, that definitely makes sense. Since the accessor is  only used in one place then let's call it `memoShortByteString`. It is more appropriate, since there is big difference between `ByteString` and `ShortByteString`

Soupstraw: I changed all fields to use `mb*` prefix instead of `memo*` before I saw your reply. Since the field accessors aren't used in many places, it shouldn't really matter either way.

lehins: > I changed all fields to use mb* prefix

That's an even better solution :wink: :+1:

eras/shelley/impl/src/Cardano/Ledger/Shelley/Tx.hs
```diff
@@ -455,7 +457,7 @@ segwitTx
           body'
           witnessSet
           (maybeToStrictMaybe metadata)
-          (SBS.toShort . BSL.toStrict $ fullBytes)
+          fullBytes
```
lehins, line 501: Very nice! :+1:

lehins (approved):
Awesome work!!! Thank you.

PR #2953, opened 2022-08-02
ValidateScript can only be applied to phase 1 scripts, enforced by types. Take 2

TimSheard:
This is a replacement for https://github.com/input-output-hk/cardano-ledger/pull/2921

Changed ValidateScript, so that validateScript has type
PhasedScript 'One era -> Core.Tx era -> Bool
Type ensure only scriptes from Phase1 can be validated.
Introduces types Phase, PhasedScript, and type family SomeScript

All this built on top of Alexey's major rewrite

### DIFF
--- a/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Language.hs
+++ b/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Language.hs
@@ -1,53 +1,8 @@
 {-# LANGUAGE DeriveGeneric #-}
-{-# LANGUAGE FlexibleContexts #-}
 
--- | This module provides data structures and operations for talking about
---     Non-native Script languages. It is expected that new languages (or new
---     versions of old languages) will be added here.
-module Cardano.Ledger.Alonzo.Language where
+-- | This module exports the module Cardano.Ledger.Language. That module exports data
+--   structures and operations for talking about Non-native Script languages. It is expected that new languages (or new
+--   versions of old languages) will be added there, and just imported here.
+module Cardano.Ledger.Alonzo.Language (module Cardano.Ledger.Language) where
 
-import Cardano.Binary (FromCBOR (..), ToCBOR (..), decodeInt)
-import Control.DeepSeq (NFData (..))
-import Data.Coders (invalidKey)
-import Data.Ix (Ix)
-import GHC.Generics (Generic)
-import NoThunks.Class (NoThunks)
-
--- | Non-Native Script language. This is an Enumerated type.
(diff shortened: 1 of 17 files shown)


eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Language.hs
```diff
@@ -1,53 +1,8 @@
...
-{-# LANGUAGE FlexibleContexts #-}
 
--- | This module provides data structures and operations for talking about
---     Non-native Script languages. It is expected that new languages (or new
---     versions of old languages) will be added here.
-module Cardano.Ledger.Alonzo.Language where
+-- | This module exports the module Cardano.Ledger.Language. That module exports data
+--   structures and operations for talking about Non-native Script languages. It is expected that new languages (or new
+--   versions of old languages) will be added there, and just imported here.
+module Cardano.Ledger.Alonzo.Language (module Cardano.Ledger.Language) where
```
JaredCorduan, line 6: is there a reason to keep `Cardano.Ledger.Alonzo.Language`?

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Tx.hs
```diff
@@ -188,16 +186,7 @@ instance CC.Crypto c => EraTx (AlonzoEra c) where
...
-
-validateAlonzoNativeScript ::
-  forall era.
-  (EraTx era, ShelleyMAEraTxBody era, Script era ~ AlonzoScript era) =>
-  Script era ->
-  Tx era ->
-  Bool
-validateAlonzoNativeScript (TimelockScript script) tx = validateTimelock @era script tx
-validateAlonzoNativeScript (PlutusScript _ _) _tx = True
+  validateScript (Phase1Script script) tx = validateTimelock @(AlonzoEra c) script tx
```
JaredCorduan, line 189: beautiful

JaredCorduan (approved):
looks great! I'm so happy to see this added type safety!

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Language.hs
TimSheard: Only because there are about 40 imports. I didn't want to touch 40 files, and I don't know if consensus or api import these things.

TimSheard: We will fix this in a follow on PR, that only changes those 40 imports.

PR #2954, opened 2022-08-02
Decouple mint from Value 

teodanciu:
Resolves #2911

* Changes the `mint` field in `TxBody` from `Value` to `MultiAsset`, representing a map of assets 
* Removes now redundant rule checking that ada  is not being forged


eras/shelley-ma/impl/src/Cardano/Ledger/ShelleyMA/Rules/Utxo.hs
```diff
@@ -256,7 +255,7 @@ utxoTransition = do
   ppup' <-
     trans @(EraRule "PPUP" era) $ TRC (PPUPEnv slot pp genDelegs, ppup, txup tx)
 
-  runTest $ validateTriesToForgeADA txb
+  {- adaPolicy ∉ supp mint tx  - check not needed because mint field of type MultiAsset cannot contain ada -}
```
JaredCorduan, line 259: beautiful :clap:

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Examples/TwoPhaseValidation.hs
```diff
@@ -783,11 +784,11 @@ utxoStEx6 pf = smartUTxOState (utxoEx6 pf) (Coin 0) (Coin 5) def
 --  Example 7: Process a MINT transaction with a succeeding Plutus script.
 -- =============================================================================
 
-mintEx7 :: forall era. (Scriptic era, HasTokens era) => Proof era -> Value era
+mintEx7 :: forall era. (Scriptic era, HasTokens era) => Proof era -> MultiAsset (Crypto era)
 mintEx7 pf = forge @era 1 (always 2 pf)
 
-outEx7 :: (HasTokens era, EraTxOut era, Scriptic era) => Proof era -> TxOut era
-outEx7 pf = newTxOut pf [Address (someAddr pf), Amount (mintEx7 pf <+> inject (Coin 995))]
+outEx7 :: (HasTokens era, EraTxOut era, Scriptic era, Value era ~ MaryValue (Crypto era)) => Proof era -> TxOut era
```
JaredCorduan, line 790: do you know why we need `Value era ~ MaryValue (Crypto era)` now?

JaredCorduan (commented):
@teodanciu this PR looks great, thank you! We should clean up the commit history before merging, though.

My plan for merging is: lets get all the [conway](https://github.com/input-output-hk/cardano-ledger/projects/4) issuses merged first, cut a new release branch `release/1.1.0`, and then we can merge this.

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Examples/TwoPhaseValidation.hs
teodanciu: That's a good question.  I accepted this sort of mechanically (because the `Amount`  field defined for tests takes a Value), but I haven't questioned this. Looking to see if it's indeed necessary, thanks for the suggestion

JaredCorduan: thanks! I was mostly curious...

CHANGELOG.md
```diff
@@ -83,6 +83,8 @@ BabbageEraTxBody --> AlonzoEraTxBody --> ....
   - `TxBody` to `BabbageTxBody` (kept type synonym with a deprecation message)
   - `PParams` to `BabbagePParams` (kept type synonym with a deprecation message)
   - `PParamsUpdate` to `BabbagePParamsUpdate` (kept type synonym with a deprecation message)
+- Changed `mint` field from type `Value era`/`Value (Crypto era)` to type `MultiAsset (Crypto era)` in `MATxBody`, `AlonzoTxBody`, `BabbageTxBody`
+  # 2974
```
TimSheard: Changed `mint` field from type `Value era`/`Mary.Value (Crypto era)` to type `MultiAsset (Crypto era)` in `MATxBody`, `AlonzoTxBody`, `BabbageTxBody`

(Value era) is the Core type familiy and  Mary.Value(Crypto era) is the concrete type. Be good to make this clear

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxBody.hs
```diff
@@ -410,7 +409,7 @@ data TxBodyRaw era = TxBodyRaw
     _vldt :: !ValidityInterval,
     _update :: !(StrictMaybe (Update era)),
     _reqSignerHashes :: Set (KeyHash 'Witness (Crypto era)),
-    _mint :: !(MaryValue (Crypto era)),
+    _mint :: !(MultiAsset (Crypto era)),
     -- The spec makes it clear that the mint field is a
     -- Cardano.Ledger.Mary.Value.MaryValue, not a Cardano.Ledger.Core.Value.
     -- Operations on the TxBody in the AlonzoEra depend upon this.
```
TimSheard, line 422: The comment is a bit out of date. Maybe something like

   -- The spec makes it clear that the mint field is a
    -- Cardano.Ledger.Mary.Value.MaryValue, not a Cardano.Ledger.Core.Value. (the type family)
    -- Operations on the TxBody in the AlonzoEra depend upon this. We now store only the MultiAsset part of a Mary.Value.

eras/alonzo/test-suite/test/Test/Cardano/Ledger/Alonzo/Golden.hs
```diff
@@ -80,36 +80,38 @@ goldenUTxOEntryMinAda =
         calcMinUTxO
           ( AlonzoTxOut
               carlAddr
-              (valueFromList 1407406 [(pid1, smallestName, 1)])
+              (MaryValue 1407406 (multiAssetFromList [(pid1, smallestName, 1)]))
```
TimSheard: I wonder if it wouldn't have been better to change the definition of valueFromList (just once), then we woudn't need to change any of the call sites?

There are more than a dozen in this file alone. Could we revert this file and just change the definition?

eras/babbage/impl/src/Cardano/Ledger/Babbage/TxBody.hs
```diff
@@ -496,7 +496,7 @@ data TxBodyRaw era = TxBodyRaw
     _vldt :: !ValidityInterval,
     _update :: !(StrictMaybe (Update era)),
     _reqSignerHashes :: !(Set (KeyHash 'Witness (Crypto era))),
-    _mint :: !(MaryValue (Crypto era)),
+    _mint :: !(MultiAsset (Crypto era)),
     -- The spec makes it clear that the mint field is a
     -- Cardano.Ledger.Mary.Value.MaryValue, not a Value.
     -- Operations on the TxBody in the BabbageEra depend upon this.
```
TimSheard, line 524: Same as the comment in the Alonzo era.  Clear up the comments.
-- The spec makes it clear that the mint field is a
-- Cardano.Ledger.Mary.Value.MaryValue, not a Cardano.Ledger.Core.Value. (the type family)
-- Operations on the TxBody in the AlonzoEra depend upon this. We now store only the MultiAsset part of a Mary.Value.

eras/shelley-ma/impl/src/Cardano/Ledger/Mary/Value.hs
```diff
@@ -192,7 +223,7 @@ instance CC.Crypto crypto => Val (MaryValue crypto) where
               + repOverhead
           )
 
-  isAdaOnly (MaryValue _ v) = Map.null v
+  isAdaOnly (MaryValue _ (MultiAsset m)) = m == mempty
```
TimSheard: why not
isAdaOnly (MaryValue _ (MultiAsset v)) = Map.null v

There is quite a bit over overhead involved using the EQ class and the Monoid class instances we can avoid.
This might appear else where, as well. Might want to search for this pattern.

eras/shelley-ma/impl/src/Cardano/Ledger/Mary/Value.hs
```diff
@@ -509,8 +540,8 @@ to ::
   -- The Nothing case of the return value corresponds to a quantity that is outside
   -- the bounds of a Word64. x < 0 or x > (2^64 - 1)
   Maybe (CompactValue crypto)
-to (MaryValue ada ma)
-  | Map.null ma =
+to (MaryValue ada m)
+  | m == mempty =
       CompactValueAdaOnly . CompactCoin <$> integerToWord64 ada
```
TimSheard: Like this maybe

to (MaryValue ada (MultiAsset m))
  | Map.null m =
      CompactValueAdaOnly . CompactCoin <$> integerToWord64 ada

eras/shelley-ma/impl/src/Cardano/Ledger/Mary/Value.hs
```diff
@@ -793,12 +822,9 @@ prune assets =
   Map.filter (not . null) $ Map.filter (/= 0) <$> assets
 
 -- | Rather than using prune to remove 0 assets, when can avoid adding them in the
---   first place by using valueFromList to construct a MaryValue.
-valueFromList :: Integer -> [(PolicyID era, AssetName, Integer)] -> MaryValue era
-valueFromList ada =
-  foldr
-    (\(p, n, i) ans -> insert (+) p n i ans)
-    (MaryValue ada Map.empty)
```
TimSheard: I think there is value keeping this function, re-defined in terms of multiAssetFromList, because it is used so many places.

eras/shelley-ma/test-suite/src/Test/Cardano/Ledger/Mary/Golden.hs
```diff
@@ -84,107 +84,116 @@ goldenScaledMinDeposit =
...
       testCase "one policy, one (small) name" $
         scaledMinDeposit
           ( MaryValue 1444443 $
-              Map.singleton
-                pid1
-                (Map.fromList [(smallName 1, 1)])
+              MultiAsset $
+                Map.singleton
+                  pid1
+                  (Map.fromList [(smallName 1, 1)])
```
TimSheard, line 98: Couldn't  we use  valueFromList here? (and elsewhere)
valueFromList 1444443 [(pid1,smallName 1,1)]

eras/shelley-ma/test-suite/src/Test/Cardano/Ledger/Mary/Golden.hs
```diff
@@ -84,107 +84,116 @@ goldenScaledMinDeposit =
...
-                )
+              MultiAsset $
+                Map.singleton
+                  pid1
+                  ( Map.fromList
+                      [ (largestName 65, 1),
+                        (largestName 66, 1),
+                        (largestName 67, 1)
+                      ]
+                  )
```
TimSheard, line 148: mapFromValue 1962961
  [ (pid1,65,1),
    (pid1,66,1),
   (pid1,67 1)]

eras/shelley-ma/test-suite/test/Test/Cardano/Ledger/ShelleyMA/Serialisation/Golden/Encoding.hs
```diff
@@ -245,7 +245,7 @@ goldenEncodingTestsAllegra =
                 (ValidityInterval SNothing SNothing)
                 SNothing
                 SNothing
-                (Coin 0)
+                mempty
```
TimSheard: Did we need to change this. In the Allegra Era the value is a Coin?

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/Fields.hs
```diff
@@ -254,7 +254,7 @@ initialTxBody (Alonzo _) =
     initVI
     SNothing
     Set.empty
-    initValue
+    mempty
```
TimSheard: There is something not quite right about using mempty here. All the other types use the initXXX, so why change it to mempty for Value?  There is a sort of self documentation that is being broken when we do this.

TimSheard (commented):
A nice job. I think we added lots of changes we didn't have to by getting rid of valueFromList.

eras/alonzo/test-suite/test/Test/Cardano/Ledger/Alonzo/Golden.hs
teodanciu: Sound sensible. I think my idea was not so much to get rid of `valueFromList` as to expose `multiAssetFromList`, which is needed in places other the pattern you're pointing out, so I replaced one with the other. But there is no reason for which both can't exist.  Will make the change, thanks for the suggestion.

CHANGELOG.md
teodanciu: For ShelleyMA era, the mint type as defined by TxBody seems to be the  Value type  family defined in core, that's why I wrote it like this - but I see why it's confusing and maybe an unnecessary detail in this context.  
I think I will leave that part out and just keep: 
```
 Changed `mint` field to type `MultiAsset (Crypto era)` in `MATxBody`, `AlonzoTxBody`, `BabbageTxBody`
```

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxBody.hs
teodanciu: Indeed, this comment makes things more confusing. Will update

eras/shelley-ma/impl/src/Cardano/Ledger/Mary/Value.hs
teodanciu: Sure.  I made an almost arbitrary choice to change it to this because I was thinking to use the higher level abstraction - but really just as a way to choose one. 
The performance overhead is a real reason to prefer the other one - I'm happy to change it and glad to have learnt this, thanks!

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/Fields.hs
teodanciu: Indeed, just laziness on my side.  Will change

eras/shelley-ma/test-suite/test/Test/Cardano/Ledger/ShelleyMA/Serialisation/Golden/Encoding.hs
teodanciu: I think `Coin` worked when mint was defined to be of type `Core.Value`, and the whole MaryOrAllegra stuff was kicking in and accepting `Coin` for Allegra and `MaryValue` for Mary.  But with mint as MultiAsset this doesn't apply anymore, so Coin can't be passed as a mint , because the `MATxBody` constructor changed to take a `MultiAsset` instead of a `Value`. 
However, per your previous comments, I can change this to (MultiAsset Map.empty) instead -  maybe it makes it clearer (and more efficient).

teodanciu:
Force-pushed to incorporate feedback from Tim (except `valueFromList`)

teodanciu:
Added a commit that restores `valueFromList`  and uses it to remove duplication, following Tim's  apt suggestion

CHANGELOG.md
```diff
@@ -83,6 +83,8 @@ BabbageEraTxBody --> AlonzoEraTxBody --> ....
   - `TxBody` to `BabbageTxBody` (kept type synonym with a deprecation message)
   - `PParams` to `BabbagePParams` (kept type synonym with a deprecation message)
   - `PParamsUpdate` to `BabbagePParamsUpdate` (kept type synonym with a deprecation message)
+- Changed `mint` field type to `MultiAsset (Crypto era)` in `MATxBody`, `AlonzoTxBody`, `BabbageTxBody`
+  # 2974
```
lehins: Current PR number is already known :wink:
```suggestion
  #2954
```

eras/shelley-ma/impl/src/Cardano/Ledger/ShelleyMA/Rules/Utxo.hs
```diff
@@ -289,17 +288,6 @@ validateOutsideValidityIntervalUTxO slot txb =
...
--- | Check that the mint field does not try to mint ADA. This is equivalent to
--- the check:
---
--- > adaPolicy ∉ supp mint tx
-validateTriesToForgeADA ::
-  ShelleyMAEraTxBody era =>
-  TxBody era ->
-  Test (UtxoPredicateFailure era)
-validateTriesToForgeADA txb =
-  failureUnless (Val.coin (txb ^. mintTxBodyL) == Val.zero) TriesToForgeADA
```
lehins, line 301: It would be a good idea to also add a note on the `TriesToForgeADA` predicate failure that it is no longer possible and is kept around for backwards compatibility only. This way we can clean it up in some future node version when we can break node-to-client serialization

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Tx.hs
```diff
@@ -429,7 +429,7 @@ rdptrInv txBody (RdmrPtr Cert idx) =
   Certifying <$> fromIndex idx (txBody ^. certsTxBodyL)
 
 getMapFromValue :: MaryValue crypto -> Map.Map (PolicyID crypto) (Map.Map AssetName Integer)
-getMapFromValue (MaryValue _ m) = m
+getMapFromValue (MaryValue _ (MultiAsset m)) = m
```
lehins, line 457: Let's move this function into `Cardano.Ledger.Shelley.Mary.Value`. We should of course continue reexporting it from this module for backwards compatibility

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxInfo.hs
```diff
@@ -535,7 +535,7 @@ alonzoTxInfo pp lang ei sysS utxo tx = do
     txWits = tx ^. witsTxL
     txOuts = txBody ^. outputsTxBodyL
     fee = txBody ^. feeTxBodyL
-    mint = txBody ^. mintTxBodyL
+    mint = txBody ^. mintValueTxBodyF
```
lehins: Let's continue using `mintTxBodyL` here. However, it order for this to work we'll need to extract portion of `transValue` into a separate function `transMultiAsset`, which can then be applied to `mint`:

```haskell
transMultiAsset :: MultiAsset c -> PV1.Value
transMultiAsset (MultiAsset mp) = Map.foldlWithKey' accum1 mempty mp
  where
    accum1 ans sym mp2 = Map.foldlWithKey' accum2 ans mp2
      where
        accum2 ans2 tok quantity =
          PV1.unionWith
            (+)
            ans2
            (PV1.singleton (transPolicyID sym) (transAssetName tok) quantity)

transValue :: MaryValue c -> PV1.Value
transValue (MaryValue n ma) = justAda <> transMultiAsset mp
  where
    justAda = PV1.singleton PV1.adaSymbol PV1.adaToken n
```

eras/babbage/impl/src/Cardano/Ledger/Babbage/TxBody.hs
```diff
@@ -584,11 +585,16 @@ vldtBabbageTxBodyL =
   lensTxBodyRaw _vldt (\txBodyRaw vldt_ -> txBodyRaw {_vldt = vldt_})
 
 mintBabbageTxBodyL ::
-  (BabbageEraTxBody era, Value era ~ MaryValue (Crypto era)) =>
-  Lens' (BabbageTxBody era) (Value era)
+  (BabbageEraTxBody era) =>
+  Lens' (BabbageTxBody era) (MultiAsset (Crypto era))
```
lehins: Redundant parenthesis and combine into one line
```suggestion
  BabbageEraTxBody era => Lens' (BabbageTxBody era) (MultiAsset (Crypto era))
```

eras/babbage/impl/src/Cardano/Ledger/Babbage/TxInfo.hs
```diff
@@ -234,7 +234,7 @@ babbageTxInfo pp lang ei sysS utxo tx = do
     witnesses = tx ^. witsTxL
     outs = txBody ^. outputsTxBodyL
     fee = txBody ^. feeTxBodyL
-    forge = txBody ^. mintTxBodyL
+    forge = txBody ^. mintValueTxBodyF
```
lehins: Same thing here as in Alonzo, we can continue using `mintTxBodyL`, just this line will need to change:
```diff
-            PV2.txInfoMint = Alonzo.transValue forge,
+            PV2.txInfoMint = Alonzo.transMultiAsset forge,
```

eras/shelley-ma/impl/src/Cardano/Ledger/Mary/Value.hs
```diff
@@ -815,15 +845,16 @@ showValue v = show c ++ "\n" ++ unlines (map trans ts)
...
-    bad = Map.keys (Map.filter Map.null m1) -- This is a malformed value, not in cannonical form.
+    (c, triples) = gettriples v
 
 gettriples :: MaryValue crypto -> (Integer, [(PolicyID crypto, AssetName, Integer)])
-gettriples v = case gettriples' v of
-  (a, b, _) -> (a, b)
+gettriples (MaryValue c ma) = (c, flattenMultiAssets ma)
+
+flattenMultiAssets :: MultiAsset crypto -> [(PolicyID crypto, AssetName, Integer)]
+flattenMultiAssets (MultiAsset m) =
```
lehins: For consistency with name of the type it should be singular
```suggestion
gettriples (MaryValue c ma) = (c, flattenMultiAsset ma)

flattenMultiAsset :: MultiAsset crypto -> [(PolicyID crypto, AssetName, Integer)]
flattenMultiAsset (MultiAsset m) =
```

eras/shelley-ma/test-suite/test/Test/Cardano/Ledger/Mary/Value.hs
```diff
@@ -49,29 +50,41 @@ pickNew, pickOld :: a -> a -> a
...
 pickOld o _n = o
 
+insert' ::
+  (Integer -> Integer -> Integer) ->
+  PolicyID crypto ->
+  AssetName ->
+  Integer ->
+  MaryValue crypto ->
+  MaryValue crypto
+insert' combine pid aid new (MaryValue c m) = MaryValue c $ insert combine pid aid new m
```
lehins: I know it is only tests, but this tick naming scheme is horrible, let's rename it to `insertValue` or something.

eras/shelley-ma/impl/src/Cardano/Ledger/Mary/Value.hs
```diff
@@ -815,15 +845,16 @@ showValue v = show c ++ "\n" ++ unlines (map trans ts)
 -- | Turn the nested 'MaryValue' map-of-maps representation into a flat sequence
 -- of policyID, asset name and quantity, plus separately the ada quantity.
 gettriples' :: MaryValue crypto -> (Integer, [(PolicyID crypto, AssetName, Integer)], [PolicyID crypto])
-gettriples' (MaryValue c m1) = (c, triples, bad)
+gettriples' v@(MaryValue _ (MultiAsset m)) = (c, triples, Map.keys (Map.filter Map.null m))
```
lehins: I suggest we remove this function, this is only used in pretty printing. What we actually need for pretty printer is:
```haskell
-- | If MutliAsset is a malformed value, not in canonical form, return the violators:
getBadMultiAsset (MultiAsset m) = Map.keys (Map.filter Map.null m)
```

See my next comment in `libs/cardano-ledger-pretty/src/Cardano/Ledger/Pretty/Mary.hs` why this works

libs/cardano-ledger-pretty/src/Cardano/Ledger/Pretty/Mary.hs
```diff
@@ -17,24 +17,31 @@ import Cardano.Ledger.ShelleyMA.Timelocks
...
 
-instance PrettyA AssetName where prettyA x = ppSexp "AssetName" [ppAssetName x]
+instance CC.Crypto crypto => PrettyA (MultiAsset crypto) where prettyA x = ppSexp "MultiAsset" [ppMultiAsset x]
+
+ppValue :: MaryValue crypto -> PDoc
+ppValue v = case gettriples' v of
+  (n, triples, []) -> ppSexp "Value" [ppCoin (Coin n), ppList pptriple triples]
+  (n, triples, bad) -> ppSexp "Value" [ppCoin (Coin n), ppList pptriple triples, ppString "Bad " <> ppList ppPolicyID bad]
+  where
+    pptriple (i, asset, num) = hsep [ppPolicyID i, ppAssetName asset, ppInteger num]
```
lehins: We can simplify this. See my comment on `gettripples'` in [eras/shelley-ma/impl/src/Cardano/Ledger/Mary/Value.hs](https://github.com/input-output-hk/cardano-ledger/pull/2954/files#diff-ec516a1c5b807ba71f5b0a126cd17bf9d406d6d91e48ebf056ff80e5b73634ed)
```suggestion
ppValue (MaryValue n m) = ppSexp "Value" $ [ppCoin (Coin n), ppMultiAsset m] ++ ppBad 
  where
    ppBad = case getBadMultiAsset m of
       [] -> []
       bad -> [ppString "Bad " <> ppList ppPolicyID bad]
```

lehins (changes requested):
This is great stuff! It could already be merged, but I just noticed a few further improvements we could do as part of this PR.

CHANGELOG.md
teodanciu: For unknown reasons, I thought that this number should be the PR that adds it to the release branch, but indeed, looking at the other PR references and git history, I was mistaken. Thanks!

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Tx.hs
teodanciu: Actually it doesn't seem to be used at all.. But we still need to keep it around, because someone using us might be calling it?

lehins: You are right, in fact, it is not used anywhere in IOG: https://github.com/search?q=org%3Ainput-output-hk+getMapFromValue&type=code

Just in case someone outside organization is using it slap `DEPRECATED` pragma on it, so we can phase it out.
```haskell
{-# DEPRECATED getMapFromValue "No longer used" #-}
```

teodanciu:
Reorganized commits to reduce noise and squashed them such that each one leaves the code in a valid state

CHANGELOG.md
lehins: Note also that there should be no space between the number and the `#`

```diff
- # 2954
+ #2954
```

eras/shelley-ma/impl/src/Cardano/Ledger/Mary/Value.hs
teodanciu: I would write this new function where it's needed in pretty and remove `gettriples'`  from here completely, what do you think?

teodanciu: And actually, the whole `gettriples` function: 

```
gettriples :: MaryValue crypto -> (Integer, [(PolicyID crypto, AssetName, Integer)])
gettriples (MaryValue c ma) = (c, flattenMultiAsset ma)
```

doesn't seem to bring a lot of value- I would remove it completely and instead expose only `flattenMultiAsset`

teodanciu: Actually nevermind, I see that this `gettriples` doesn't get exposed from the module, I will leave it alone.  I have however removed `gettriples'` from this module in the new version - hope that's ok

teodanciu:
I think I addressed all comments and change requests - would be great to merge if anyone wants to take another look and/or approve!

CHANGELOG.md.orig
```diff
@@ -0,0 +1,220 @@
+# Changelog
```
lehins: backup file was added to the repo. I think it could be useful for you to add `*.orig` pattern to you `.global_gitignore` file

CHANGELOG.md
```diff
@@ -156,7 +156,8 @@ BabbageEraTxBody --> AlonzoEraTxBody --> ....
       - `FromAlonzoUtxoFail` -> `AlonzoInBabbageUtxoPredFailure`
     - `BabbageUtxowPred` -> `BabbageUtxowPredFailure` and constructor:
       - `FromAlonzoUtxowFail` -> `AlonzoInBabbageUtxowPredFailure`
-
+- Changed `mint` field type to `MultiAsset (Crypto era)` in `MATxBody`, `AlonzoTxBody`, `BabbageTxBody`
+  # 2954
```
```suggestion
  #2954
```

lehins (changes requested):
Couple more comments :wink:

lehins (approved):
Awesome work!

PR #2955, opened 2022-08-03
Make VMap.VU an Unboxed Vector

LudvikGalois:
VMap.VU was erroneously a Storable Vector.


JaredCorduan (approved):
thank you @LudvikGalois !

PR #2957, opened 2022-08-04
New calculatePoolDistr' function

newhoggy:
It is the same as `calculatePoolDistr`, but additionally takes a filter function to constrain returned results

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,9 @@ in the naming of release branches.
 ## [Unreleased]
 
 ### Added
+- New `calculatePoolDistr'` function which is similar to `calculatePoolDistr` but has a new
+  filter argument to only include the stake pool ids (stake pool key hashes) that are needed.
+  #2957
 - New package `cardano-ledger-conway`, which defines new era: the `ConwayEra`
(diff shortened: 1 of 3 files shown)


JaredCorduan (approved):
looks great @newhoggy , thanks!

Could you please update the changelog for this nice new addition?

CHANGELOG.md
```diff
@@ -10,6 +10,8 @@ in the naming of release branches.
 ## [Unreleased]
 
 ### Added
+- New `calculatePoolDistr'` function which is similar to `calculatePoolDistr` but has a new
+  filter argument to only include the stake pool ids (stake pool key hashes) that are needed.
```
JaredCorduan, line 14: we've been adding in the PR number here to, which is quite awkward at the moment (since you have to put up the PR, see what number you get, then push). the hope is to get some hook that'll do this for us. apologies for forgetting to tell you this.

JaredCorduan, line 14:
```suggestion
  filter argument to only include the stake pool ids (stake pool key hashes) that are needed.
  #2957 
```

newhoggy: Done!

PR #2960, opened 2022-08-04
set LC_ALL="C.UTF-8" inside nix shell

JaredCorduan:
Micro PR :microscope:  to set `LC_ALL="C.UTF-8"` in the nix shell. This is needed if you want to build our haddocks, since we use unicode characters. Thanks goes to @teodanciu who showed me how to do this :)


teodanciu (approved):
Looks good to me

PR #2962, opened 2022-08-08
Remove unused data-default-class library from shelley-ma dependencies

teodanciu:
Getting intermittent `unused-packages` error.  It seems that this library is not used


lehins (approved):
Yeah, its usage has been removed in #2945 but the dependency was not removed from the cabal file

PR #2970, opened 2022-08-09
Added precision to the Babbage era specification concerning the use of new features with PlutusV1 scripts

koslambrou:
This PR adds a precision to the Babbage era specification which specifies which features can't be used in a transaction which try to execute PlutusV1 scripts.

Don't hesitate to change the formulation if it's not clear enough.

### DIFF
--- a/eras/babbage/formal-spec/utxo.tex
+++ b/eras/babbage/formal-spec/utxo.tex
@@ -271,8 +271,10 @@ In the UTXO rule, we switch from a manual estimation of the size consumed by $\U
 
 To the UTXOW rule, in addition to the changes required by the new
 features, we add a check that all scripts and datums involved in the
-transaction are well-formed. Also, we forbid transactions that use the
-new features and try to execute $\PlutusVI$ scripts.
+transaction are well-formed. Also, we forbid transactions that try to
+execute $\PlutusVI$ scripts and use the new features which can't be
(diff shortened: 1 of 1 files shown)


eras/babbage/formal-spec/utxo.tex
```diff
@@ -271,8 +271,10 @@ \section{UTxO}
 
 To the UTXOW rule, in addition to the changes required by the new
 features, we add a check that all scripts and datums involved in the
-transaction are well-formed. Also, we forbid transactions that use the
-new features and try to execute $\PlutusVI$ scripts.
+transaction are well-formed. Also, we forbid transactions that try to
+execute $\PlutusVI$ scripts and use the new features which can't be
+translated to $\PlutusVI$'s $\TxInfo$ (reference inputs, inline datums
+and reference scripts).
```
JaredCorduan: Maybe we should also mention that we reject Byron addresses. How about:

```suggestion
transaction are well-formed. We forbid transactions that try to
execute $\PlutusVI$ scripts and use the new features which can't be
translated to $\PlutusVI$'s $\TxInfo$ (reference inputs, inline datums
and reference scripts). We also forbid transactions that involve Byron 
addresses.
```

JaredCorduan:
Thank you @koslambrou ! This looks great, I made just one small comment.

JaredCorduan (approved):

PR #2971, opened 2022-08-10
Add two invalid Babbage feature tests

catch-21:
For PlutusV1 phase-1 checks: `ReferenceInputsNotSupported` and `ReferenceScriptsNotSupported`


JaredCorduan (approved):
Thank you @james-iohk !

I'll drop a bit of context here (that James and I already talked through). We do have a suite of tests already for the translation errors ([here](https://github.com/input-output-hk/cardano-ledger/blob/d4743d896ef615e265af2d9637c27e831c82868d/eras/babbage/test-suite/test/Test/Cardano/Ledger/Babbage/TxInfo.hs#L191-L195) are the babbage ones), but it's nice to have more visibility about the translation errors, and some folks think to look first at these examples.

PR #2976, opened 2022-08-15
Unify witness types naming across eras - issue #2931 

### LINKED ISSUE #2931: Rename Witnesses consistently across all eras
Current naming is quite inconsistent and should be fixed.

The consensus throughout the team is to name the core type family `TxWits`, which means:

* The class will look like this:
```haskell
class
  ( EraScript era,
    Eq (TxWits era),
    Show (TxWits era),
    Monoid (TxWits era),
    NoThunks (TxWits era),
    ToCBOR (TxWits era),
    FromCBOR (Annotator (TxWits era))
  ) =>
  EraTxWits era
  where
  type TxWits era = (r :: Type) | r -> era

  mkBasicTxWits :: TxWits era
  mkBasicTxWits = mempty

  addrTxWitsL :: Lens' (TxWits era) (Set (WitVKey 'Witness (Crypto era)))

  bootAddrTxWitsL :: Lens' (TxWits era) (Set (BootstrapWitness (Crypto era)))

  scriptTxWitsL :: Lens' (TxWits era) (Map (ScriptHash (Crypto era)) (Script era))
```

And the concrete types per era will be placed into `Cardano.Ledger.[Era].TxWits`, which means for Shelley they'll have to move from the `Tx.hs` module. Also important for Shelley is to define a `newtype` around the `ShelleyTxWitsHKD`, rather than using a `type` synonym`. This would be an example on Alonzo

```haskell
instance (EraScript (AlonzoEra c), CC.Crypto c) => EraTxWits (AlonzoEra c) where
  type TxWits (AlonzoEra c) = AlonzoTxWits (AlonzoEra c)

  mkBasicTxWits = mempty
  addrTxWitsL = addrAlonzoTxWitsL
  bootAddrTxWitsL = bootAddrAlonzoTxWitsL
  scriptTxWitsL = scriptAlonzoTxWitsL

class EraTxWits era => AlonzoEraTxWits era where
  datsTxWitsL :: Lens' (TxWits era) (TxDats era)

  rdmrsTxWitsL :: Lens' (TxWits era) (Redeemers era)

instance (EraScript (AlonzoEra c), CC.Crypto c) => AlonzoEraTxWits (AlonzoEra c) where
  datsTxWitsL = datsAlonzoTxWitsL
  rdmrsTxWitsL = rdmrsAlonzoTxWitsL
```

Record field names could also be fixed during this task.

teodanciu:
Closes #2931


JaredCorduan:
I love it!

eras/shelley/impl/src/Cardano/Ledger/Shelley/TxWits.hs
```diff
@@ -119,27 +120,56 @@ deriving via
...
+
+instance
+  ( Era era,
+    NFData (Script era),
+    NFData (WitVKey 'Witness (EraCrypto era)),
+    NFData (BootstrapWitness (EraCrypto era))
+  ) =>
+  NFData (ShelleyTxWits era)
+
+instance (Era era, NoThunks (Script era)) => (NoThunks (ShelleyTxWits era))
```
lehins: Redundant parenthesis
```suggestion
instance (Era era, NoThunks (Script era)) => NoThunks (ShelleyTxWits era)
```

eras/shelley/impl/src/Cardano/Ledger/Shelley/TxWits.hs
```diff
@@ -255,12 +286,14 @@ decodeWits = do
...
-      <$> addrWits' witSet
-      <*> scriptWits' witSet
-      <*> bootWits' witSet
-      <*> annBytes
+  let witSetAnnotator =
+        do
+          aw <- addrWits' witSet
+          sw <- scriptWits' witSet
+          bw <- bootWits' witSet
+          WitnessSet' aw sw bw <$> annBytes
```
lehins: What is the motivation behind this transformation to the do syntax? Applicative syntax like there is currently is usually preferred

eras/shelley/impl/src/Cardano/Ledger/Shelley/TxWits.hs
```diff
@@ -119,27 +120,56 @@ deriving via
...
+    (\(ShelleyTxWitsConstr w) -> getter w)
+    (\(ShelleyTxWitsConstr w) val -> ShelleyTxWitsConstr $ setter w val)
 
 -- | Addresses witness setter and getter for `ShelleyTxWits`. The
 -- setter does update memoized binary representation.
 addrShelleyWitsL ::
-  EraTxWits era => Lens' (ShelleyTxWits era) (Set (WitVKey 'Witness (EraCrypto era)))
-addrShelleyWitsL = lens addrWits' (\w s -> w {addrWits = s})
+  Lens' (ShelleyTxWits era) (Set (WitVKey 'Witness (EraCrypto era)))
+addrShelleyWitsL = lensWitsRaw addrWits' (\witsRaw a -> witsRaw {addrWits' = a})
```
lehins: It might still be in your plans, but just in case you overlooked that part of the ticket since it is a bit subtle. We also need to adjust lens names for consistency `addrShelleyWitsL` -> `addrShelleyTxWitsL` and the same for class functions: `addrWitsL` -> `addrTxWitsL`

This can be easily achieved with a regular search ad replace in all files, so it should be pretty easy. Let me know if you need help with it.

eras/shelley/impl/src/Cardano/Ledger/Shelley/TxWits.hs
teodanciu: I was just gonna add a comment here on the PR to ask - for some reason I couldn't manage to get it to compile using applicative syntax. 
However, while trying to get the compilation error to add it to the comment,  I got the the right combination of <$> and <*> to make it work.   Will be changed in the next push

eras/shelley/impl/src/Cardano/Ledger/Shelley/TxWits.hs
teodanciu: Yes I definitely overlooked that, thanks for the reminder. Will add a commit

JaredCorduan (approved):
Looks great to me! The great re-organization of 2022 :raised_hands: continues!

eras/shelley/impl/src/Cardano/Ledger/Shelley/TxWits.hs
```diff
@@ -119,27 +120,58 @@ deriving via
...
 -- setter does update memoized binary representation.
 bootAddrShelleyWitsL ::
-  EraTxWits era => Lens' (ShelleyTxWits era) (Set (BootstrapWitness (EraCrypto era)))
-bootAddrShelleyWitsL = lens bootWits' (\w s -> w {bootWits = s})
+  EraScript era =>
+  Lens' (ShelleyTxWits era) (Set (BootstrapWitness (EraCrypto era)))
+bootAddrShelleyWitsL =
+  lens
+    (\(ShelleyTxWitsConstr w) -> bootWits' w)
+    (\(ShelleyTxWitsConstr w) s -> ShelleyTxWits {addrWits = addrWits' w, scriptWits = scriptWits' w, bootWits = s})
```
teodanciu: @lehins In the end, the line with the setter that you showed me  (` ShelleyTxWits (w {addrWits = s}))` ) didn't compile for me exactly like that, I think because the definition of the pattern was different.  Or maybe I messed something else up! 

If you know of a better way let me know please

lehins: This should work:
```haskell
bootAddrShelleyWitsL =
  lens
    (\(ShelleyTxWitsConstr w) -> bootWits' w)
    (\(ShelleyTxWitsConstr w) s -> ShelleyTxWitsConstr (w {bootWits = s}))
```

teodanciu: That's what I tried, but with the pattern definition like this: 

```
pattern ShelleyTxWits ::
  EraScript era =>
  Set (WitVKey 'Witness (EraCrypto era)) ->
  Map (ScriptHash (EraCrypto era)) (Script era) ->
  Set (BootstrapWitness (EraCrypto era)) ->
  ShelleyTxWits era
pattern ShelleyTxWits {addrWits, scriptWits, bootWits} <- 
...
```

I'm getting 

```
    • Couldn't match expected type ‘WitnessSetHKD Identity era’
                  with actual type ‘ShelleyTxWits era0’
    • In the first argument of ‘ShelleyTxWitsConstr’, namely
        ‘(w {bootWits = s})’
```

Maybe the pattern should be defined in another way? :thinking:

lehins: Oh yeah, you've changed already the type of `ShelleTxWits` pattern synonym, so it should just be:
```haskell
bootAddrShelleyWitsL =
  lens
    (\(ShelleyTxWitsConstr w) -> bootWits' w)
    (\w s -> w {bootWits = s})
```

teodanciu: Oh! :facepalm:

lehins (approved):
Beautiful!

PR #2977, opened 2022-08-15
Improve performance of access to coin in TxOut

lehins:
Performance of `areAllAdaOnly` and `coinBalance` are over 100x faster than the one of `balance` (see benchmark below). This is because whenever we only need `Coin` we don't have to deserialize the `MaryValue` and we can operate on the compact form of the coin directly.

Therefore this PR attempts to switch all places where this pattern happens

* `coin $ balance utxo` -> `coinBalance utxo`
* `isAdaOnly $ balance utxo` -> `areAllAdaOnly utxo`
* `coin (txOut ^. valueTxOutL)` -> `txOut ^. coinTxOutL`

Another big part of this PR is the explicit implementation of the fact that it is possible to have MultiAsset in the collateral input set, but only if it is fully being send back in the returnCollateral. Most importantly it changes the predicate failure `CollateralContainsNonADA` to not return negative MultiAsset values, when returnCollateral contains mismatched type and/or amount of concrete MultiAsset.

---
These are the benchmarks of the newly added functions on the full UTxO, except `areAllAdaOnly`, which is applied to the part of the UTxO that is actually all ada only.
```
benchmarking UTxO/balance
time                 15.16 s    (13.32 s .. 16.53 s)
                     0.998 R²   (0.994 R² .. 1.000 R²)
mean                 14.82 s    (14.31 s .. 15.14 s)
std dev              507.2 ms   (195.3 ms .. 675.9 ms)
variance introduced by outliers: 19% (moderately inflated)

benchmarking UTxO/coinBalance
time                 147.4 ms   (146.8 ms .. 147.7 ms)
                     1.000 R²   (1.000 R² .. 1.000 R²)
mean                 147.2 ms   (146.9 ms .. 147.4 ms)
std dev              317.9 μs   (177.1 μs .. 449.6 μs)
variance introduced by outliers: 12% (moderately inflated)

benchmarking UTxO/areAllAdaOnly
time                 114.6 ms   (111.9 ms .. 115.9 ms)
                     1.000 R²   (0.999 R² .. 1.000 R²)
mean                 115.6 ms   (114.7 ms .. 116.2 ms)
std dev              1.003 ms   (529.9 μs .. 1.439 ms)
variance introduced by outliers: 11% (moderately inflated)
```

### DIFF
--- a/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Rules/Utxos.hs
+++ b/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Rules/Utxos.hs
@@ -70,8 +70,8 @@ import qualified Cardano.Ledger.Shelley.LedgerState as Shelley
 import Cardano.Ledger.Shelley.PParams (Update)
 import Cardano.Ledger.Shelley.Rules.Ppup (PpupEnv (..), ShelleyPPUP, ShelleyPpupPredFailure)
 import Cardano.Ledger.Shelley.Rules.Utxo (UtxoEnv (..), updateUTxOState)
-import Cardano.Ledger.Shelley.UTxO (UTxO (..), balance, totalDeposits)
-import Cardano.Ledger.Val as Val (Val (coin, (<->)))
+import Cardano.Ledger.Shelley.UTxO (UTxO (..), coinBalance, totalDeposits)
+import Cardano.Ledger.Val as Val (Val ((<->)))
 import Cardano.Slotting.EpochInfo.Extend (unsafeLinearExtendEpochInfo)
 import Cardano.Slotting.Slot (SlotNo)
 import Control.Monad.Trans.Reader (ReaderT, asks)
@@ -281,7 +281,7 @@ scriptsNotValidateTransition = do
   pure
     $! us
       { _utxo = UTxO utxoKeep,
-        _fees = fees <> Val.coin (balance (UTxO utxoDel)),
+        _fees = fees <> coinBalance (UTxO utxoDel),
         _stakeDistro = updateStakeDistribution (_stakeDistro us) (UTxO utxoDel) mempty
       }
 
--- a/eras/babbage/impl/src/Cardano/Ledger/Babbage/Collateral.hs
+++ b/eras/babbage/impl/src/Cardano/Ledger/Babbage/Collateral.hs
@@ -6,20 +6,24 @@
 
 -- | Figure 2: Functions related to fees and collateral
 --   Babbage Specification
-module Cardano.Ledger.Babbage.Collateral where
+module Cardano.Ledger.Babbage.Collateral
+  ( isTwoPhaseScriptAddress,
+    collAdaBalance,
+    collOuts,
+  )
+where
 
 import Cardano.Ledger.Address (Addr (..))
 import Cardano.Ledger.Alonzo.Tx (isTwoPhaseScriptAddressFromMap)
-import Cardano.Ledger.Alonzo.TxBody (AlonzoEraTxBody (..))
 import Cardano.Ledger.Alonzo.TxInfo (ExtendedUTxO (txscripts))
 import Cardano.Ledger.Babbage.TxBody (BabbageEraTxBody (..), BabbageTxBody, BabbageTxOut (..))
 import Cardano.Ledger.BaseTypes (TxIx (..), txIxFromIntegral)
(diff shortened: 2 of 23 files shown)


eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Rules/Utxos.hs
```diff
@@ -281,7 +281,7 @@ scriptsNotValidateTransition = do
   pure
     $! us
       { _utxo = UTxO utxoKeep,
-        _fees = fees <> Val.coin (balance (UTxO utxoDel)),
+        _fees = fees <> coinBalance (UTxO utxoDel),
```
JaredCorduan, line 284: I love this new, clearer language.

shell.nix
```diff
@@ -59,7 +59,7 @@ let
         fi
       }
       PROMPT_COMMAND=prompt
-      export LC_ALL="C.UTF-8"
+      export LC_ALL=en_US.utf-8
```
JaredCorduan: can you explain the difference to me?

eras/babbage/impl/src/Cardano/Ledger/Babbage/Collateral.hs
```diff
@@ -36,19 +40,17 @@ isTwoPhaseScriptAddress ::
...
 
-collBalance ::
-  forall era.
+collAdaBalance ::
   (BabbageEraTxBody era, TxOut era ~ BabbageTxOut era) =>
   TxBody era ->
-  UTxO era ->
-  Value era
-collBalance txBody (UTxO m) =
+  Map.Map (TxIn (Crypto era)) (TxOut era) ->
```
JaredCorduan, line 46: In a few places I see we've shed the constructor and work directly with the `Map`. Would it be clearer if we kept it as `UTxO`? Or maybe there's a reason for using a map that I haven't understood... :thinking:

JaredCorduan:
looks great to me, thanks @lehins !

shell.nix
lehins: That change is wrong. I'll delete it.

On NixOS:
* this `export LC_ALL="C.UTF-8"` gives a warning, but everything still works
* `export LC_ALL=en_US.utf-8` this works just fine, no warnings

On other OSes in `nix-shell` this doesn 't work `LC_ALL=en_US.utf-8`

I haven't tried to figure out all the reasoning behind this, because waning on NixOS can just be ignored. As log as everything works I don't care much :smile:

eras/babbage/impl/src/Cardano/Ledger/Babbage/Collateral.hs
lehins: I did it on purpose, because before `callBalance` accepted the whole of `UTxO`, but now it only suppose to accept `eval ((txBody ^. collateralInputsTxBodyL) ◁ utxo)`. So I wanted to get a type error instead of it happily accepting either one of those values. Also it is more consistent with Alonzo:
https://github.com/input-output-hk/cardano-ledger/blob/2eafa6e55723c4cd2ef04911674e8f66830661d5/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Rules/Utxo.hs#L315

shell.nix
JaredCorduan: gotcha, thanks for the explanation! It would obviously be great to have no warnings, but indeed warnings are better than errors :) 

I'll make an issue so we can solve this later...

eras/babbage/impl/src/Cardano/Ledger/Babbage/Collateral.hs
JaredCorduan: great :+1:

shell.nix
JaredCorduan: Issue -> #2980

lehins: Commit removed

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Rules/Utxos.hs
lehins: It is not only clearer, but it is also 100x faster :wink:

JaredCorduan: :metal:

WhatisRT (approved):
LGTM

PR #2978, opened 2022-08-16
Minimize constraints to avoid ghc bug

lehins:
Consensus tools tests [stumbled upon the same GHC bug](https://hydra.iohk.io/build/17830322) that was mentioned in this https://github.com/input-output-hk/cardano-ledger/pull/2950#issuecomment-1207298011

I went through all of the sites where bug could potentially be triggered and removed this possibility by removing redundant constraints. 

There are no semantic changes in this PR, just expansion of class instances' constraints to the minimal possible set, but only for those instances that can trigger the bug.


JaredCorduan (approved):
Looks great to me!

You have`Catrdano` :cat: in the commit messages.

lehins:
Fixed that :joy_cat: rdano and squashed into one commit with a better message. Will merge as soon as PR goes green.

PR #2979, opened 2022-08-16
Add a whole bunch of `INLINE` and `SPECIALIZE` pragmas

lehins:
Despite that this is a lot of boiler plate it guarantees that the
new interface has no overhead


JaredCorduan (approved):
:+1:

PR #2984, opened 2022-08-19
Split and Refactor TwoPhaseValidation

teodanciu:
The main change is the introduction of module `STSTestUtils` - which is perhaps dubious in both its name and existence. 
It's an attempt to aggregate things (functions, values, classes, instances) that  _want to_  mean something - so that they are not simply shortcuts to avoid copy - pasting, but have some sort of significance that can be reused across different scenarios. There's arbitrariness to this with regards to what makes it in this module and what is too context-specific  - this is my best attempt at finding a balance between gross duplication and accidental abstraction/reuse . Both of these extremes make the code difficult to navigate and comprehend.  

Using the building blocks from STSTestUtils, I split `TwoPhaseValidation` module into 5 smaller modules.  By removing some of the data reuse that seemed accidental, I introduced some duplication between these modules, but I have tried to keep each of them understandable on their own. 
I also did some refactoring/renaming of the code, rather than merely copying the code over, in an attempt to make the code easier to follow. Mostly : 
  - renamed the main components of each testcase uniformly across the test cases
  - moved top-level functions that were not reused to a narrower scope (`where` clause), in order to keep the main  namespace less polluted and hopefully easier to navigate 
  - tried to name things in a meaningful way - in particular the ones indexed with numbers (eg : redeemersEx6 ) : If it seemed to "mean something "  (for example, it was being used in both constructing the transaction AND reused for checking the result afterwards), then I gave it a name related to the test case that used it. Otherwise, I replaced the name with the definition or reduced it's scope and gave it a more generic name. 

I'm sure there's many ways this could be done better - happy to hear other ideas, I got a bit fed up with my own thoughts about this in these past days !


libs/cardano-ledger-test/cardano-ledger-test.cabal
```diff
@@ -46,7 +46,13 @@ library
     Test.Cardano.Ledger.Alonzo.Tools
     Test.Cardano.Ledger.BaseTypes
     Test.Cardano.Ledger.Examples.BabbageFeatures
-    Test.Cardano.Ledger.Examples.TwoPhaseValidation
+    Test.Cardano.Ledger.Examples.AlonzoUTXOW
+    Test.Cardano.Ledger.Examples.AlonzoUTXOWValidTx
+    Test.Cardano.Ledger.Examples.AlonzoUTXOWInvalidTx
```
lehins: Minor optional suggestion: I think it would read a little easier this way:
```suggestion
    Test.Cardano.Ledger.Examples.AlonzoValidTxUTXOW
    Test.Cardano.Ledger.Examples.AlonzoInvalidTxUTXOW
```

teodanciu: I agree,  your suggestions seem more pleasant for some reason, will change

lehins (approved):

PR #2985, opened 2022-08-19
Remove an exact same validation in Babbage in favor of the one in Shelley

lehins:
Remove the version of `validateOutputBootAddrAttrsTooBig` that is
defined in Babbage. All eras use the version from Shelley. The only
thing special about Babbage version was that it accepted a list instead
of a `UTxO`. Converting to a list with `elems` at the call site instead
of inside the function itself is a small price to pay for removing an
extra copy of the validation function.


JaredCorduan (approved):
:+1:

PR #2989, opened 2022-08-22
Make Arbitrary instances Era agnostic

lehins:
There is no need to redefine duplicate Arbitrary instances for the same type in different era. This PR solves this problem


JaredCorduan (approved):
Looks great to me!

Issue #2988, opened 2022-08-22
Instances needed for proper master key rotation


JaredCorduan:
In input-output-hk/ouroboros-network#2963, an the `TranslateEra LedgerState` instance was used to rotate the master keys. This should suffice for mainnet, and other networks that begin in the Byron era, but questions remain about Shelley-only networks. In particular, do we need instances for `CanStartFromGenesis` and `AdditionalGenesisConfig`, and do we need to translate `ShelleyGenesis` via `TranslateEra ShelleyGenesis`.

JaredCorduan:
The ledger does not need to use `TranslateEra ShelleyGenesis` to rotate the master keys, we have a plan for consensus that will let us get rid of these instances in general (see https://github.com/input-output-hk/ouroboros-consensus/issues/409).

We still want to add a `CanStartFromGenesis ConwayEra` instance, though, where `AdditionalGenesisConfig` is something isomorphic to `(AlonzoGenesisConfig, ConwayGenesisConfig)`. The instance is responsible for converting a Shelley ledger state directly into a Conway ledger state.

Soupstraw:
There already seems to be a `CanStartFromGenesis` instance for `ConwayEra` ([added here](https://github.com/input-output-hk/cardano-ledger/commit/c3a46e04244404261bb7fddc2629ddb0729c7927#diff-e106c7b5b8a971b311fc101be676dcc1a5b13d1111f36a070d7699c6e68b1ab7R51))

JaredCorduan:
Ah, you are right @Soupstraw . The question still remains, though, if we have the correct behavior.

The current `CanStartFromGenesis` for conway is potentially a source for confusion. We are using the conway genesis to override the master keys in the shelley genesis. So if you are starting in conway (which is what the `CanStartFromGenesis` class is used for), the shelley master keys are entirely un-used. Alternatively, `CanStartFromGenesis` for conway could forgo the conway genesis altogether, and just use the shelley genesis to set the values for conway. I'm not sure which is less confusing. Maybe we could assert that precisely one of the two is non-empty (or that they agree)?

JaredCorduan:
A lot has changed since I made this issue, it is not really relevant anymore. We will still be doing master key rotation, but in a much different way.

PR #2990, opened 2022-08-22
Add missing instances for Conway era

lehins:
This are all instances that were needed for consensus integration of ConwayEra and are implemented by @nfrisby

### DIFF
--- a/eras/conway/impl/src/Cardano/Ledger/Conway/Translation.hs
+++ b/eras/conway/impl/src/Cardano/Ledger/Conway/Translation.hs
@@ -46,7 +46,7 @@ import Cardano.Ledger.ShelleyMA.Timelocks (translateTimelock)
 import Data.Coerce
 
 --------------------------------------------------------------------------------
--- Translation from Alonzo to Babbage
+-- Translation from Babbage to Conway
 --
 -- The instances below are needed by the consensus layer. Do not remove any of
 -- them without coordinating with consensus.
--- a/eras/conway/impl/cardano-ledger-conway.cabal
+++ b/eras/conway/impl/cardano-ledger-conway.cabal
@@ -56,6 +56,7 @@ library
                 cardano-ledger-shelley-ma,
                 containers,
                 microlens,
+                nothunks,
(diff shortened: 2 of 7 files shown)


eras/conway/impl/src/Cardano/Ledger/Conway/Translation.hs
```diff
@@ -93,7 +93,7 @@ instance Crypto c => TranslateEra (ConwayEra c) ShelleyGenesis where
           API.sgUpdateQuorum = API.sgUpdateQuorum genesis,
           API.sgMaxLovelaceSupply = API.sgMaxLovelaceSupply genesis,
           API.sgProtocolParams = translateEra' ctxt (API.sgProtocolParams genesis),
-          API.sgGenDelegs = API.sgGenDelegs genesis,
+          API.sgGenDelegs = API.sgGenDelegs genesis,    -- TODO Is this correct?
```
lehins: @JaredCorduan Could you please comment on this. I haven't fully groked what ConwayEra introduces yet.

JaredCorduan: The conway era will probably only provide the master key rotation. For mainnet, and other networks that start with the Byron era, I do not know of any need to modify the master keys inside the Shelley genesis, but I am not an expert on everything that the consensus layer uses the various ledger state for. It's probably true that shelley-only networks do need to modify `sgGenDelegs`. I made #2988 for exactly these questions.

JaredCorduan (approved):
Looks good to me! (but I am happy to give @nfrisby a chance to look this over before we merge it)

eras/conway/impl/src/Cardano/Ledger/Conway/Translation.hs
lehins: Ok, I removed the TODO. If we need to do anything about it, we can do it in a follow up PR

JaredCorduan: perfect, thanks

PR #2991, opened 2022-08-22
Switch larger functions from `INLINE` to `INLINEABLE` 

lehins:
Memory usage during compilation  has gone through the roof due to recent INLINE pragma additions: #2979 

This PR switches usage of INLINE in favor of INLINEABLE for obviously large functions

### DIFF
--- a/eras/babbage/impl/src/Cardano/Ledger/Babbage/TxBody.hs
+++ b/eras/babbage/impl/src/Cardano/Ledger/Babbage/TxBody.hs
@@ -262,13 +262,13 @@ dataHashBabbageTxOutL =
         SNothing -> BabbageTxOut addr cv NoDatum s
         SJust dh -> BabbageTxOut addr cv (DatumHash dh) s
     )
-{-# INLINE dataHashBabbageTxOutL #-}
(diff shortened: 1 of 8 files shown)


JaredCorduan (approved):
:+1:

eras/babbage/impl/src/Cardano/Ledger/Babbage/TxBody.hs
```diff
@@ -755,6 +756,7 @@ instance CC.Crypto c => ShelleyMAEraTxBody (BabbageEra c) where
   {-# INLINE mintTxBodyL #-}
 
   mintValueTxBodyF = mintValueBabbageTxBodyF
+  {-# INLINE mintValueTxBodyF #-}
```
JaredCorduan, line 759: you meant `INLINE` and not `INLINEABLE`, right?

lehins: Yes. This is a very small function, it is good to inline it :)
`mintValueTxBodyF = mintValueBabbageTxBodyF`

PR #2992, opened 2022-08-23
Remove dependency on cardano-prelude

### LINKED ISSUE #1676: Remove use of `cardano-prelude`
Depends on https://github.com/input-output-hk/cardano-prelude/issues/115

teodanciu:
This PR also updates `plutus` and `cardano-base` dependencies.

The initial purpose of this was to remove the dependency on cardano-prelude. It turns out a good amount of yak-shaving was necessary to achieve this, in the form of updating plutus and adapting to the API changes.  Also a change in nix-shell, without which cabal from within the nix-shell could not resolve plutus dependencies  (for nix reasons). 

You might need to delete ~/.cabal/store/ghc-8.10.7  in order for cabal to get the new plutus version properly. Before deleting this I was getting errors like this:

```
 Could not load module 'Plutus.V2.Ledger.Api'
    It is a member of the package 'plutus-ledger-api-1.0.0.0-b674ddb5eb0946c34d87898eef5ad3899a783b9697a81e84375d1682c1e7629f' 
    which is unusable due to missing dependencies:
barbies-2.0.3.1-82eGrD9c7V217T7RY4npks base16-bytestring-1.0.2.0-
  ... (and many others) 
```

So if this happens to you, you might want to 1) delete this folder 2) brace yourself with patience 

resolves  https://github.com/input-output-hk/cardano-ledger/issues/1676


eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Scripts.hs
```diff
@@ -295,8 +305,8 @@ decodeCostModel lang = do
     Right cm -> pure cm
   where
     keys = case lang of
-      PlutusV1 -> PV1.costModelParamNames
-      PlutusV2 -> PV2.costModelParamNames
+      PlutusV1 -> Set.fromList $ Text.pack . showParamName <$> enumerate @PV1.ParamName
```
teodanciu: This line gets repeated a few times in this commit, but I couldn't find a good place for it that doesn't feel completely arbitrary. I'm happy to extract it somewhere if anyone has suggestions on where it would best fit..

libs/plutus-preprocessor/src/Main.hs
```diff
@@ -152,9 +152,10 @@ main = do
       "import Cardano.Ledger.Alonzo.Language (Language (..))",
```
teodanciu: I don't know how this `Main` gets called, or how to test it..  I had to "compile in my head"  and I have no confidence it will work. Any idea how to check ?

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Genesis.hs
```diff
@@ -224,7 +226,7 @@ validateCostModel (lang, cmps) = case mkCostModel lang cmps of
 translateLegacyV1paramNames :: Map Text Integer -> Map Text Integer
 translateLegacyV1paramNames cmps =
   Map.fromList $
-    zipWith (\(_, v) k2 -> (k2, v)) (Map.toList cmps) (Set.toList PV1.costModelParamNames)
+    zipWith (\(_, v) k2 -> (k2, v)) (Map.toList cmps) (Text.pack . showParamName <$> enumerate @PV1.ParamName)
```
lehins:
```suggestion
    zip (Text.pack . showParamName <$> enumerate @PV1.ParamName) (Map.elems cmps)
```

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Scripts.hs
lehins: This is indeed a very common pattern in this PR. What I can recommend is this:

Add functions to `Cardano.Ledger.Language` in the `cardano-ledger-core`. It is as good of a place as any, especially since it is directly related to Plutus language version.

Something like this:
```haskell
costParamsNames :: Language -> [Text]
costParamsNames = \case
  PlutusV1 -> Text.pack . showParamName <$> enumerate @PV1.ParamName
  PlutusV2 -> Text.pack . showParamName <$> enumerate @PV2.ParamName

costParamsNamesSet :: Language -> [Text]
costParamsNamesSet = Set.fromList . costParamNames
```


Then `keys` just becomes:

```haskell
  where
    keys = costParamsNamesSetPlutusV1 lang
```

eras/alonzo/test-suite/src/Test/Cardano/Ledger/Alonzo/Serialisation/Generators.hs
```diff
@@ -227,8 +229,8 @@ genCM lang costModelParamNames = do
   either (error "Corrupt cost model") pure $ mkCostModel lang newCMPs
 
 genCostModel :: Language -> Gen (Language, CostModel)
-genCostModel PlutusV1 = (PlutusV1,) <$> genCM PlutusV1 PV1.costModelParamNames
-genCostModel PlutusV2 = (PlutusV2,) <$> genCM PlutusV2 PV2.costModelParamNames
+genCostModel PlutusV1 = (PlutusV1,) <$> genCM PlutusV1 (Set.fromList (pack . showParamName <$> enumerate @PV1.ParamName))
+genCostModel PlutusV2 = (PlutusV2,) <$> genCM PlutusV2 (Set.fromList (pack . showParamName <$> enumerate @PV2.ParamName))
```
lehins: With my suggestion of `costParamsNamesSet`  this function becomes:
```suggestion
genCostModel lang = (,) lang <$> genCM lang (costParamsNamesSet lang)
```

~~Note that `TupleSections` extension can also be removed~~

eras/alonzo/test-suite/test/Test/Cardano/Ledger/Alonzo/Golden.hs
```diff
@@ -233,8 +237,8 @@ freeCostModel :: HasCallStack => Language -> CostModel
 freeCostModel lang =
   fromRightError "freeCostModel is not well-formed" $ mkCostModel lang (cmps lang)
   where
-    names PlutusV1 = PV1.costModelParamNames
-    names PlutusV2 = PV2.costModelParamNames
+    names PlutusV1 = Set.fromList $ Text.pack . showParamName <$> enumerate @PV1.ParamName
+    names PlutusV2 = Set.fromList $ Text.pack . showParamName <$> enumerate @PV2.ParamName
     cmps = Map.fromSet (const 0) . names
```
lehins: Same here
```suggestion
    cmps = Map.fromSet (const 0) . costParamsNamesSet
```

libs/plutus-preprocessor/src/Main.hs
lehins: See here for explanation of how to do it (last line) https://github.com/input-output-hk/cardano-ledger/pull/2328

lehins (changes requested):
Looks great. As you did notice we do need to remove some duplication though

teodanciu:
@lehins Thanks for the review and suggestions - I think I have to redo the Plutus update slightly and use better their changes. Concretely, [CostModel](https://github.com/input-output-hk/cardano-ledger/blob/master/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Scripts.hs#L231)  in our code could use the enum for the param names instead of the text .  I will make the change and see which of your suggestions still apply

L-as:
I'm trying to do a similar thing, and I've had to change a lot of `.`s on [`Format`](https://hackage.haskell.org/package/formatting-7.1.3/docs/Formatting-Internal.html#t:Format) into `%`s for some reason. I'm curious why you didn't need the same thing here. Maybe I'm using a formatting that's too new accidentally?

lehins:
> I'm trying to do a similar thing, and I've had to change a lot of .s on [Format](https://hackage.haskell.org/package/formatting-7.1.3/docs/Formatting-Internal.html#t:Format) into %s for some reason.

@L-as I think you need to use `.` from `Control.Category` for it to work

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Scripts.hs
lehins: I've addressed this comment

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Genesis.hs
lehins: This one too.

eras/alonzo/test-suite/src/Test/Cardano/Ledger/Alonzo/Serialisation/Generators.hs
lehins: Addressed this one too

eras/alonzo/test-suite/test/Test/Cardano/Ledger/Alonzo/Golden.hs
lehins: And this one

lehins:
This PR brings ledger up to date with latest plutus and cardano-base as well as removes dependency on cardano-prelude. We need to unblock @erikd and others on their effort to update ghc-9.2 Becuase @teodanciu is off on Monday I took it upon myself to address all the review comments and incorporated #3030 into this PR as well. If there is anything else needs to be done for this PR we can tackle it in a subsequent one at a later point.

lehins (approved):
Looks great. Thank you for doing all the work @teodanciu If there were any other plans for this PR we can postpone it to a later PR.

@JaredCorduan Once you back, it would be nice if you could take a second look at this PR, even though it will merged. It will be valuable, since you are the most familiar person with the Plutus Cost Models in ledger

JaredCorduan:
> @JaredCorduan Once you back, it would be nice if you could take a second look at this PR, even though it will merged. It will be valuable, since you are the most familiar person with the Plutus Cost Models in ledger

Most definitely, I am very keen to review this one carefully. thanks for the heads up.

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Scripts.hs
```diff
@@ -270,15 +287,18 @@ instance NFData CostModel where
...
+  case eCostModel of
+    Right (evalCtx, _) -> Right (CostModel lang cm evalCtx)
     Left e -> Left e
+  where
+    mkEvaluationContext =
+      case lang of
+        PlutusV1 -> PV1.mkEvaluationContext
+        PlutusV2 -> PV2.mkEvaluationContext
+    eCostModel :: Either PV1.CostModelApplyError (PV1.EvaluationContext, [CostModelApplyWarn])
+    eCostModel = runWriterT (mkEvaluationContext (Map.elems cm))
```
JaredCorduan, line 301: wonderful! it's awesome to see the plan put into action! https://github.com/input-output-hk/cardano-ledger/issues/2902#issuecomment-1183644628

There is actually now no longer a need to keep the keys in the cost model. In other words, we can replace the `Map Text Integer` with `[Integer]`.

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Scripts.hs
```diff
@@ -291,14 +311,19 @@ decodeCostModelMap = decodeMapByKey fromCBOR decodeCostModel
 
 decodeCostModel :: Language -> Decoder s CostModel
 decodeCostModel lang = do
+  let keys = costModelParamsNamesSet lang
```
JaredCorduan, line 314: I'll make a follow up issues for this, but as a mentioned above, we can now change the "visible" cost model from `Map Text Integer` to `[Integer]`. Which will make this decoder even simpler. :) I suspect we'll have no need for the `costModelParamsNames` anywhere.

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxInfo.hs
```diff
@@ -227,9 +226,6 @@ transDataHash' safe = PV1.DatumHash (transSafeHash safe)
 transKeyHash :: KeyHash d c -> PV1.PubKeyHash
 transKeyHash (KeyHash h) = PV1.PubKeyHash (PV1.toBuiltin (hashToBytes h))
 
-transScriptHash :: ScriptHash c -> PV1.ValidatorHash
-transScriptHash (ScriptHash h) = PV1.ValidatorHash (PV1.toBuiltin (hashToBytes h))
```
JaredCorduan, line 231: good catch!

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxInfo.hs
```diff
@@ -240,8 +236,8 @@ txInfoId :: TxId c -> PV1.TxId
 txInfoId (TxId safe) = PV1.TxId (transSafeHash safe)
 
 transStakeCred :: Credential kr c -> PV1.Credential
-transStakeCred (ScriptHashObj (ScriptHash kh)) =
-  PV1.ScriptCredential (PV1.ValidatorHash (PV1.toBuiltin (hashToBytes kh)))
+transStakeCred (ScriptHashObj (ScriptHash sh)) =
+  PV1.ScriptCredential (PV1.ScriptHash (PV1.toBuiltin (hashToBytes sh)))
```
JaredCorduan, line 240: thank you for fixing that silly variable name!

JaredCorduan (approved):
Thank you so much for doing these two items @teodanciu , it looks great! (and you too @lehins for getting it across the line!)

It's going to be wonderful having these changes to the evaluation context (I reviewed it all very carefully).

I made one small followup issue: #3042

lehins:
Thank you @JaredCorduan

PR #2996, opened 2022-08-25
Organize Shelley rules modules:

lehins:
In order to battle orphan instances escaping the cardano-ledger-shelley
package we need to remove `EraMappings` module. This commit implements
rules module structure in the same way as it is done in all subsequent
eras, thus solving the orphan problem and keeping the structure
consistent. Individual rules are defined in `Cardano.Ledger.Shelley.Era`
and individual rules modules are no longer exposed. All shelley rules
are available through `Cadano.Ledger.Shelley.Rules` module.


JaredCorduan (approved):
looks great!

PR #2997, opened 2022-08-25
Fix typo

eyeinsky:
Fix typo and resolve redirect in vector-map home page.


JaredCorduan (approved):
good catch, thank you @eyeinsky !

PR #3002, opened 2022-08-25
specify txid in shelley spec appendix

JaredCorduan:
The Shelley ledger spec was missing a description of the `txid` function in the appendix (it's fine for it to be abstract in the spec iteself).

I also used the opportunity to warn people about re-serializing data.

### DIFF
--- a/eras/shelley/formal-spec/cddl.tex
+++ b/eras/shelley/formal-spec/cddl.tex
@@ -10,3 +10,8 @@ The CDDL specification is located at
 
 % TODO - Include the CDDL spec inline?
 % \lstinputlisting[backgroundcolor = \color{lightgray}]{../impl/cddl-files/shelley.cddl}
+
+Note that the ledger does
+\textbf{not support a canonical representation}.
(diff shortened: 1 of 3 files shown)


WhatisRT (approved):
I don't know whether it makes sense to also mention that this means that `txid` isn't really a function in `TxBody`, otherwise LGTM

JaredCorduan:
> I don't know whether it makes sense to also mention that this means that `txid` isn't really a function in `TxBody`, otherwise LGTM

Can you say more?

https://github.com/input-output-hk/cardano-ledger/blob/ce96a41602d48959974f6964485980e272e4ebb9/libs/cardano-ledger-core/src/Cardano/Ledger/Block.hs#L164-L166

JaredCorduan:
@WhatisRT and I chatted privately, the issue is the subtlety around the serialization not being unique. I think it's worth making a note of this!

JaredCorduan:
CI failure was in one of the generator classifiers, totally unrelated.

PR #3008, opened 2022-08-29
Improve interface of minutxo calculation

lehins:
This PR essentially moves the implementation of `evaluateMinLovelaceOutput` into `EraTxOurt` as a newly named function(s) `getMinCoinTxOut`/`getMinCoinSizedTxOut` and makes use of it both in UTXO rules and in the external API.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,9 @@ in the naming of release branches.
 
 ### Added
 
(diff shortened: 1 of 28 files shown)


JaredCorduan (approved):
excellent work!

JaredCorduan:
@lehins - changelog reminder

lehins:
> changelog reminder

Updated the changelog and added couple more tests

PR #3009, opened 2022-08-30
Rename Crypto type to EraCrypto

### LINKED ISSUE #2933: Fix conflicting name `type Crypto` vs `class Crypto`
This is an omnipresent issue throughout the repo, which causes us to `import qualified Cardano.Ledger.Class as CC`. This is not that big of a deal, but if there is an agreement throughout the team we could rename one to avoid this collision. It does lead to more typing and confusion for the newcomers, so I think it would be an improvement. In consensus, in fact, a type synonym `EraCrypto` is created for it for it, which makes me think we could adopt the same name:

https://github.com/input-output-hk/ouroboros-network/blob/470ee82d9dfb1e07f55e6600d50de6aa60cf8a5c/ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Eras.hs#L108-L112

Soupstraw:
This PR renames the `Crypto` dependent type to `EraCrypto` to avoid name clashes with the `Crypto` class.

closes #2933


JaredCorduan (approved):
LGTM! don't forget to update the changelog.

PR #3010, opened 2022-08-30
Add the module Test.Cardano.Ledger.Generic.Consensus, which combines

TimSheard:
Adds the module Test.Cardano.Ledger.Generic.Consensus, which combines all the Consensus examples into one file, and uses the Generic framework to parameterize the examples over a (Proof era).
Also adds the Test.Cardano.Ledger.Generic.Same module that one uses similar to the class Eq, but which on failure
returns a "path" to points in the values that differ.
Adds a test old style consensus examples agree with the generic examples.


eras/babbage/impl/src/Cardano/Ledger/Babbage.hs
```diff
@@ -60,6 +61,8 @@ import Data.Maybe.Strict
 import qualified Data.Set as Set
 import Lens.Micro
 
+-- Bring "instance TranslationContext (BabbageEra c)" into scope.
```
lehins: I think this comment is out of place. In fact it is not really necessary since the whole package relies on this structure

eras/shelley-ma/impl/src/Cardano/Ledger/Allegra/Translation.hs
```diff
@@ -25,10 +24,15 @@ import qualified Cardano.Ledger.Shelley.LedgerState as LS
...
 import Cardano.Ledger.Shelley.TxWits (decodeWits)
+import Cardano.Ledger.ShelleyMA (MaryOrAllegra (Allegra), ShelleyMAEra)
 import Control.Monad.Except (throwError)
 import Data.Coerce (coerce)
 import qualified Data.Map.Strict as Map
 
+--------------------------------------------------------------------------------
+
+type AllegraEra = ShelleyMAEra 'Allegra
+
```
lehins: Why did you move the type synonym into this module? It is totally out of place. The only acceptable places for this would be in either one of these modules, so can you please put it back

* `Cardano.Ledger.ShelleyMA.Era`
* `Cardano.Ledger.Allegra`
* or `Cardano.Ledger.ShelleyMA`

eras/shelley-ma/impl/src/Cardano/Ledger/Mary/Translation.hs
```diff
@@ -47,6 +47,8 @@ import Data.Maybe (fromMaybe)
 -- being total. Do not change it!
 --------------------------------------------------------------------------------
 
+type MaryEra = ShelleyMAEra 'Mary
+
```
lehins: Same here, it is defined out of place

eras/shelley/impl/cardano-ledger-shelley.cabal
```diff
@@ -61,6 +61,8 @@ library
     Cardano.Ledger.Shelley.SoftForks
     Cardano.Ledger.Shelley.StabilityWindow
     Cardano.Ledger.Shelley.Rules
+    Cardano.Ledger.Shelley.Rules.Ledger
+    Cardano.Ledger.Shelley.Rules.Delegs
```
lehins: Please revert this change. These modules are hidden on purpose. All imports should go through `Cardano.Ledger.Shelley.Rules`, since all rules modules are re-exported from there.

I guess at some point I'll need to write up a document explaining our solution to orphan instances. In fact here is  blog post that I ran into two days ago that briefly touches on this approach: https://www.parsonsmatt.org/2019/11/27/keeping_compilation_fast.html#hidden-orphans

eras/shelley/impl/src/Cardano/Ledger/Shelley.hs
```diff
@@ -25,7 +25,7 @@ import Cardano.Ledger.Core
     Value,
   )
 import Cardano.Ledger.Crypto (StandardCrypto)
-import Cardano.Ledger.Shelley.Era (ShelleyEra)
+import Cardano.Ledger.Shelley.Era (ShelleyEra) -- Bring "instance TranslationContext (ShelleyEra c)" into scope.
```
lehins: This comment is also out of place, since that import does not bring that instance into scope and also it doesn't make sense to make those comments all over the place since  all of our packages depend on this approach.

```suggestion
import Cardano.Ledger.Shelley.Era (ShelleyEra)
```

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Examples/Consensus.hs
```diff
@@ -0,0 +1,59 @@
...
+    new = ledgerExamples proof
+    equal =
+      ( case proof of
+          Shelley _ -> old == new
+          Allegra _ -> old == new
+          Mary _ -> old == new
+          Alonzo _ -> old == new
+          Babbage _ -> old == new
+          Conway _ -> old == new
+      )
```
lehins: Could you use `specialize` here? Also redundant parenthesis:
```suggestion
        case proof of
          Shelley _ -> old == new
          Allegra _ -> old == new
          Mary _ -> old == new
          Alonzo _ -> old == new
          Babbage _ -> old == new
          Conway _ -> old == new
```

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/Same.hs
```diff
@@ -0,0 +1,535 @@
...
+eqVia pcf x y = if x == y then Nothing else Just (notEq (pcf x) (pcf y))
+
+-- ==========================================
+
+-- | The Same class is similar to Eq but returns descriptions (paths and
+--   non-matching components) for each difference in a large structure.
+class Same era t where
+  same :: Proof era -> t -> t -> [(String, Maybe PDoc)]
+
+instance (EraCrypto era ~ crypto) => Same era (DPState crypto) where
```
lehins: Could you please switch all uses of `crypto` type parameter to `c` in this PR. Joosep just went through a lot of trouble to rename all such uses through out the repo to make it consistent in: #3027 
```suggestion
instance (EraCrypto era ~ crypto) => Same era (DPState crypto) where
```

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/Same.hs
```diff
@@ -0,0 +1,535 @@
...
+      ++ extendLabel "LedgerState " (same proof ls1 ls2)
+
+instance (Era era) => Same era (NewEpochState era) where
+  same proof (NewEpochState en1 bmA1 bmB1 es1 pr1 pd1 _st1) (NewEpochState en2 bmA2 bmB2 es2 pr2 pd2 _st2) =
+    [ ("nesEL", eqByShow en1 en2),
+      ("nesBprev", eqByShow bmA1 bmA2),
+      ("nesBcur", eqByShow bmB1 bmB2),
+      ("nesRU", eqByShow pr1 pr2),
+      ("nesPd", eqByShow pd1 pd2)
+      -- , ("nesStashAVVM", eqByShow st1 st2)
```
lehins: Any reason this is commented out? Seems wrong to omit fields when printing things that aren't equal.

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/Same.hs
```diff
@@ -0,0 +1,535 @@
...
+sameWithDependency (SomeM labx actx x1 x2 : more) =
+  case actx x1 x2 of
+    [] -> sameWithDependency more
+    ansx -> extendLabel (labx ++ " ") ansx ++ sameWithDependency more
+
+instance Reflect era => Same era (ShelleyLedgerExamples era) where
+  same
+    proof
+    (ShelleyLedgerExamples (Block' h1 x1 _) hh1 tx1 ap1 rc1 re1 ne1 cd1 ctx1)
+    (ShelleyLedgerExamples (Block' h2 x2 _) hh2 tx2 ap2 rc2 re2 ne2 cd2 ctx2) =
```
lehins: This is a general suggestion which applies in many places. I just wanna raise it once because this is a good example for this suggestion.

I would like to suggest using record style access for types like this instead of pattern matching with positional arguments. I understand that it would be more verbose, but it would not only make it clear what all those bindings mean (`rc`, `re`, etc. carry almost no information about the binding), but most importantly it would make it less dangerous to change those types in the future. For example right now changing the order of fields or simply adding and removing a field could very well make it compile, but the output of the function would be wrong.

For example if we made this two line change:

```diff
data ShelleyLedgerExamples era = ShelleyLedgerExamples
  { sleBlock :: Block (BHeader (EraCrypto era)) era,
    sleHashHeader :: HashHeader (EraCrypto era),
    sleTx :: Core.Tx era,
    sleApplyTxError :: ApplyTxError era,
    sleRewardsCredentials :: Set (Either Coin (Credential 'Staking (EraCrypto era))),
-    sleResultExamples :: ShelleyResultExamples era,
-    sleNewEpochState :: NewEpochState era,
+    sleNewEpochState :: NewEpochState era,
+    sleResultExamples :: ShelleyResultExamples era,
    sleChainDepState :: ChainDepState (EraCrypto era),
    sleTranslationContext :: TranslationContext era
  }
```

The `same`` function would happily compile but would produce an incorrect output.

Of course, this functionality is used for testing, so I am not worried much about, but I did wanna make this recommendation nonetheless.

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/Same.hs
```diff
@@ -0,0 +1,535 @@
...
+
+instance Same era (ShelleyResultExamples era) where
+  same proof (ShelleyResultExamples pp1 ppup1 pdist1 myop1 gen1) (ShelleyResultExamples pp2 ppup2 pdist2 myop2 gen2) =
+    [ ("PParams", samePParams proof pp1 pp2),
+      ( "ProposedPPUpdates",
+        case proof of
+          Shelley _ ->
+            case mapdiffs (getmap ppup1) (getmap ppup2) of
+              [] -> Nothing
+              xs -> Just (ppString (show xs))
```
lehins: Can you use `specialize` in all such places, instead of pattern matching on all proofs?
```suggestion
        specialize proof $
            case mapdiffs (getmap ppup1) (getmap ppup2) of
              [] -> Nothing
              xs -> Just (ppString (show xs))
```

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/Same.hs
```diff
@@ -0,0 +1,535 @@
...
+sameLedgerFail (Conway _) x y = eqByShow x y
+{-# NOINLINE sameLedgerFail #-}
+
+sameTransCtx :: Proof era -> Core.TranslationContext era -> Core.TranslationContext era -> Maybe PDoc
+sameTransCtx (Shelley _) x y = eqByShow x y
+sameTransCtx (Allegra _) x y = eqByShow x y
+sameTransCtx (Mary _) x y = eqByShow x y
+sameTransCtx (Alonzo _) x y = eqByShow x y
+sameTransCtx (Babbage _) x y = eqByShow x y
+sameTransCtx (Conway _) x y = if x == y then Nothing else Just ("Conway Genesis")
```
lehins: How come `eqByShow` isn't used here? Also there are redundant parenthesis.

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/Same.hs
```diff
@@ -0,0 +1,535 @@
...
+    ("ValidityInterval", eqVia ppValidityInterval v1 v2),
+    ("PPupdate", eqVia (\_ -> ppString "Update") pu1 pu2),
+    ("AuxDataHash", eqVia (ppStrictMaybe (\(AuxiliaryDataHash h) -> trim (ppSafeHash h))) d1 d2),
+    ("Mint", eqVia multiAssetSummary m1 m2)
+  ]
+
+sameAlonzoTxBody ::
+  ( Reflect era,
+    Core.TxOut era ~ AlonzoTxOut era
+  ) =>
```
lehins: This constraint is probably not needed:
```suggestion
  Reflect era =>
```

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/Same.hs
```diff
@@ -0,0 +1,535 @@
...
+sameTxBody proof@(Mary _) x y = sameMATxBody proof x y
+sameTxBody proof@(Alonzo _) x y = sameAlonzoTxBody proof x y
+sameTxBody proof@(Babbage _) x y = sameBabbageTxBody proof x y
+sameTxBody proof@(Conway _) x y = sameBabbageTxBody proof x y
+
+-- =======================
+-- Comparing Tx for Sameness
+
+sameShelleyTx ::
+  (Reflect era, Core.TxWits era ~ ShelleyTxWits era) =>
```
lehins, line 462: This one too:
```suggestion
  Reflect era =>
```

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/Same.hs
```diff
@@ -0,0 +1,535 @@
...
+  [(String, Maybe PDoc)]
+sameShelleyTxSeq proof (ShelleyTxSeq ss1) (ShelleyTxSeq ss2) =
+  sameWithDependency (zipWith3 f ints (toList ss1) (toList ss2))
+  where
+    f n t1 t2 = SomeM (show n) (sameTx proof) t1 t2
+
+sameAlonzoTxSeq ::
+  ( Reflect era,
+    Core.Tx era ~ AlonzoTx era
+  ) =>
```
lehins: Same here:
```suggestion
  Reflect era =>
```

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/Same.hs
```diff
@@ -0,0 +1,535 @@
...
+    ]
+    where
+      getmap (ProposedPPUpdates x) = x
+
+-- =========================================================================
+-- Functions like 'same' from the 'Same' class, but which apply to type families
+-- We cannot make them Same instances because they are type families.
+-- We also can avoid all extra constraints by pattern matching against all current Proofs.
+
+samePParams :: Proof era -> Core.PParams era -> Core.PParams era -> Maybe PDoc
```
lehins, line 265: All these `same*` function could use `specialize`, right?

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/Consensus.hs
```diff
@@ -0,0 +1,1166 @@
...
+-- genericTxBody proof
+--  [Just field1
+--  , if (Some proof) >= Mary Mock then Just field2 else Nothing
+--  , Just field3
+--  ]
+genericTxBody :: Era era => Proof era -> [Maybe (TxBodyField era)] -> Core.TxBody era
+genericTxBody proof xs = specialize @Core.EraTxBody proof (newTxBody proof (foldl' accum [] xs))
+  where
+    accum ans Nothing = ans
+    accum ans (Just field) = field : ans
```
lehins:
```suggestion
genericTxBody proof xs = specialize @Core.EraTxBody proof (newTxBody proof (catMaybes xs))
```

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/Consensus.hs
```diff
@@ -0,0 +1,1166 @@
...
+genericTxOut proof xs = newTxOut proof (foldl' accum [] xs)
+  where
+    accum ans Nothing = ans
+    accum ans (Just field) = field : ans
+
+genericWits :: Era era => Proof era -> [Maybe (WitnessesField era)] -> Core.TxWits era
+genericWits proof xs = newWitnesses merge proof (foldl' accum [] xs)
+  where
+    accum ans Nothing = ans
+    accum ans (Just field) = field : ans
```
lehins:
```suggestion
genericWits proof xs = newWitnesses merge proof (catMaybes xs)
```

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/Consensus.hs
```diff
@@ -0,0 +1,1166 @@
...
+genericTxBody proof xs = specialize @Core.EraTxBody proof (newTxBody proof (foldl' accum [] xs))
+  where
+    accum ans Nothing = ans
+    accum ans (Just field) = field : ans
+
+genericTxOut :: Era era => Proof era -> [Maybe (TxOutField era)] -> Core.TxOut era
+genericTxOut proof xs = newTxOut proof (foldl' accum [] xs)
+  where
+    accum ans Nothing = ans
+    accum ans (Just field) = field : ans
```
lehins:
```suggestion
genericTxOut proof xs = newTxOut proof (catMaybes xs)
```

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/Consensus.hs
```diff
@@ -0,0 +1,1166 @@
...
+    step :: UsesPP era => PulsingRewUpdate (EraCrypto era)
+    step =
+      ( startStep @era
+          (EpochSize 432000)
+          (BlocksMade (Map.singleton (mkKeyHash 1) 10))
+          epochState
+          (Coin 45)
+          (activeSlotCoeff testGlobals)
+          10
+      )
```
lehins: Redundant parenthesis
```suggestion
        startStep @era
          (EpochSize 432000)
          (BlocksMade (Map.singleton (mkKeyHash 1) 10))
          epochState
          (Coin 45)
          (activeSlotCoeff testGlobals)
          10
```

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/Consensus.hs
```diff
@@ -0,0 +1,1166 @@
...
+          byte
+
+    sk = DSIGN.genKeyDSIGN seed
+
+exampleHashHeader ::
+  forall era.
+  Reflect era =>
+  Proof era ->
+  HashHeader (EraCrypto era)
+exampleHashHeader _proof = coerce $ mkDummyHash (Proxy @(HASH (EraCrypto era))) 0
```
lehins, line 320: Why supply proof if it is not needed? We just need the proxy, not proof.
```suggestion
exampleHashHeader ::
  forall era.
  Reflect era =>
  proxy era ->
  HashHeader (EraCrypto era)
exampleHashHeader _proxy = coerce $ mkDummyHash (Proxy @(HASH (EraCrypto era))) 0
```

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/Consensus.hs
```diff
@@ -0,0 +1,1166 @@
...
+    sleHashHeader :: HashHeader (EraCrypto era),
+    sleTx :: Core.Tx era,
+    sleApplyTxError :: ApplyTxError era,
+    sleRewardsCredentials :: Set (Either Coin (Credential 'Staking (EraCrypto era))),
+    sleResultExamples :: ResultExamples era,
+    sleNewEpochState :: NewEpochState era,
+    sleChainDepState :: ChainDepState (EraCrypto era),
+    sleTranslationContext :: TranslationContext era
+  }
+-}
```
lehins: This is dead code. No need to copy and paste type definitions here, it is bound to get out of sync with the actual code. It is always better to reference the original types.
```suggestion
```

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/Consensus.hs
```diff
@@ -0,0 +1,1166 @@
...
+{-# LANGUAGE TypeFamilies #-}
+{-# OPTIONS_GHC -Wno-orphans #-}
+
+-- | Generic Consensus examples
+--   This module is meant to replace
+--     Test.Cardano.Ledger.Shelley.Examples.Consensus
+--     Test.Cardano.Ledger.Allegra.Examples.Consensus
+--     Test.Cardano.Ledger.Mary.Examples.Consensus
+--     Test.Cardano.Ledger.Alonzo.Examples.Consensus
+--     Test.Cardano.Ledger.Babbage.Examples.Consensus
```
lehins: It is better to add such comment either to the PR, to the commit message or create an issue about the ultimate goal of this module. Such comments in the code do get out of date eventually.
```suggestion
--   This module is meant to replace
--     Test.Cardano.Ledger.Shelley.Examples.Consensus
--     Test.Cardano.Ledger.Allegra.Examples.Consensus
--     Test.Cardano.Ledger.Mary.Examples.Consensus
--     Test.Cardano.Ledger.Alonzo.Examples.Consensus
--     Test.Cardano.Ledger.Babbage.Examples.Consensus
```

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/Consensus.hs
```diff
@@ -0,0 +1,1166 @@
+{-# LANGUAGE AllowAmbiguousTypes #-}
+{-# LANGUAGE ConstraintKinds #-}
+{-# LANGUAGE DataKinds #-}
+{-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE RankNTypes #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeApplications #-}
+{-# LANGUAGE TypeFamilies #-}
+{-# OPTIONS_GHC -Wno-orphans #-}
```
lehins: There are no instances in this module
```suggestion
```

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Examples/Consensus.hs
```diff
@@ -0,0 +1,59 @@
...
+    -- A result is bad, if the snd component is a Just (i.e. contains a description of an inequality).
+    bad (_, Nothing) = False
+    bad (_, Just _) = True
+    results = filter bad (if equal then [] else same proof old new)
+    displayResults :: [(String, Maybe PDoc)] -> String
+    displayResults xs = (show (vsep (map f xs)))
```
lehins: Redundant parenthesis and potential for eta reduction:
```suggestion
    displayResults = show . vsep . map f
```

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Examples/Consensus.hs
```diff
@@ -0,0 +1,59 @@
...
+          Allegra _ -> old == new
+          Mary _ -> old == new
+          Alonzo _ -> old == new
+          Babbage _ -> old == new
+          Conway _ -> old == new
+      )
+    -- A result is bad, if the snd component is a Just (i.e. contains a description of an inequality).
+    bad (_, Nothing) = False
+    bad (_, Just _) = True
+    results = filter bad (if equal then [] else same proof old new)
```
lehins: No need to filter empty list
```suggestion
    results = if equal then [] else filter bad (same proof old new)
```

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/Consensus.hs
```diff
@@ -0,0 +1,1166 @@
...
+genericWits proof xs = newWitnesses merge proof (foldl' accum [] xs)
+  where
+    accum ans Nothing = ans
+    accum ans (Just field) = field : ans
+
+-- =========================================================
+-- Individual examples for each Era
+-- =========================================================
+
+type StandardShelley = ShelleyEra StandardCrypto
```
lehins: There are already type synonyms like that, they where added in #3017 
```suggestion
```

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/Consensus.hs
```diff
@@ -0,0 +1,1166 @@
...
+          StrictSeq.fromList
+            [ RequireSignature (mkKeyHash 0),
+              RequireSignature (mkKeyHash 1)
+            ],
+        RequireSignature (mkKeyHash 100)
+      ]
+
+-- ==============================================
+
+type StandardMary = MaryEra CC.StandardCrypto
```
lehins: Same here, just use `Mary` from `Cardano.Ledger.Mary`

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/Consensus.hs
```diff
@@ -0,0 +1,1166 @@
...
+    }
+  where
+    boundRational' :: HasCallStack => Rational -> NonNegativeInterval
+    boundRational' x = case boundRational x of
+      Nothing -> error $ "Expected non-negative value but got: " <> show x
+      Just x' -> x'
+
+-- ============================================================
+
+type StandardBabbage = BabbageEra StandardCrypto
```
lehins: Same here, use `Babbage` from `Cardano.Ledger.Babbage`
```suggestion
```

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/Consensus.hs
```diff
@@ -0,0 +1,1166 @@
...
+
+    couttsCoin :: AssetName
+    couttsCoin = AssetName "couttsCoin"
+
+multiAssetOf :: forall c. MaryValue c -> MultiAsset c
+multiAssetOf (MaryValue _ ma) = ma
+
+-- ============================================================
+
+type StandardAlonzo = AlonzoEra StandardCrypto
```
lehins: Same here, use `Alonzo` from `Cardano.Ledger.Alonzo`
```suggestion
```

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/Consensus.hs
```diff
@@ -0,0 +1,1166 @@
...
+    ( exampleTxBody
+        (Babbage Standard)
+        (exampleMultiAssetValue @StandardCrypto 2)
+    )
+    exampleAuxiliaryDataAlonzo
+    exampleAlonzoGenesis
+
+-- ============================================================
+
+type StandardConway = ConwayEra StandardCrypto
```
lehins: Same here, use `Conway` from `Cardano.Ledger.Conway`
```suggestion
type StandardConway = ConwayEra StandardCrypto
```

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Examples/Consensus.hs
```diff
@@ -0,0 +1,59 @@
...
+genericConsensusTest :: TestTree
+genericConsensusTest =
+  testGroup
+    "Generic Consensus examples agree with non-generic ones."
+    [ testCase "Shelley" (oneTest (Shelley Standard)),
+      testCase "Allegra" (oneTest (Allegra Standard)),
+      testCase "Mary" (oneTest (Mary Standard)),
+      testCase "Alonzo" (oneTest (Alonzo Standard)),
+      testCase "Babbage" (oneTest (Babbage Standard)),
+      testCase "Conway" (oneTest (Conway Standard))
```
lehins, line 57: Duplicate type synonyms should be removed and appropriate existing type synonyms should be used instead:
```suggestion
    [ testCase "Shelley" (oneTest Shelley),
      testCase "Allegra" (oneTest Allegra),
      testCase "Mary" (oneTest Mary),
      testCase "Alonzo" (oneTest Alonzo),
      testCase "Babbage" (oneTest Babbage),
      testCase "Conway" (oneTest Conway)
```

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/Consensus.hs
```diff
@@ -0,0 +1,1166 @@
...
+    Shelley _ -> ShelleyTxOut (mkAddr (examplePayKey, exampleStakeKey)) spendval -- (Coin 100000)
+    Allegra _ -> ShelleyTxOut (mkAddr (examplePayKey, exampleStakeKey)) spendval -- exampleCoin
+    Mary _ ->
+      ShelleyTxOut
+        (mkAddr (examplePayKey, exampleStakeKey))
+        spendval -- (exampleMultiAssetValue 1)
+    Alonzo _ ->
+      AlonzoTxOut
+        (mkAddr (SLE.examplePayKey, SLE.exampleStakeKey))
+        spendval -- (exampleMultiAssetValue 2)
```
lehins: Dead code
```suggestion
        spendval
```

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/Consensus.hs
```diff
@@ -0,0 +1,1166 @@
...
+
+exampleTxOut :: Proof era -> Core.Value era -> Core.TxOut era
+exampleTxOut proof spendval =
+  case proof of
+    Shelley _ -> ShelleyTxOut (mkAddr (examplePayKey, exampleStakeKey)) spendval -- (Coin 100000)
+    Allegra _ -> ShelleyTxOut (mkAddr (examplePayKey, exampleStakeKey)) spendval -- exampleCoin
+    Mary _ ->
+      ShelleyTxOut
+        (mkAddr (examplePayKey, exampleStakeKey))
+        spendval -- (exampleMultiAssetValue 1)
```
lehins: Dead code
```suggestion
        spendval
```

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/Consensus.hs
```diff
@@ -0,0 +1,1166 @@
...
+
+exampleTotalCollateral :: Proof era -> Maybe Coin
+exampleTotalCollateral proof =
+  case proof of
+    Shelley _ -> Nothing
+    Allegra _ -> Nothing
+    Mary _ -> Nothing
+    Alonzo _ -> Nothing
+    Babbage _ -> Just $ (Coin 8675309)
+    Conway _ -> Just $ (Coin 8675309)
```
lehins: Redundant `$`
```suggestion
    Babbage _ -> Just (Coin 8675309)
    Conway _ -> Just (Coin 8675309)
```

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/Consensus.hs
```diff
@@ -0,0 +1,1166 @@
...
+      AuxiliaryDataHash $ mkDummySafeHash (Proxy @(EraCrypto era)) 30
+{-# NOINLINE exampleTxBody #-}
+
+exampleValidity :: Proof era -> Maybe (TxBodyField era)
+exampleValidity proof =
+  case proof of
+    Shelley _ -> Just (TTL (SlotNo 10))
+    Allegra _ -> Just (Vldt (ValidityInterval (SJust (SlotNo 2)) (SJust (SlotNo 4))))
+    Mary _ -> Just (Vldt (ValidityInterval (SJust (SlotNo 2)) (SJust (SlotNo 4))))
+    Alonzo _ -> Just (Vldt (ValidityInterval (SJust (SlotNo 2)) (SJust (SlotNo 4)))) -- txvldt
```
lehins:
```suggestion
    Alonzo _ -> Just (Vldt (ValidityInterval (SJust (SlotNo 2)) (SJust (SlotNo 4))))
```

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/Consensus.hs
```diff
@@ -0,0 +1,1166 @@
...
+    Mary _ -> Nothing
+    Alonzo _ -> Nothing
+    Babbage _ -> Just (Set.fromList [mkTxInPartial (TxId (mkDummySafeHash Proxy 1)) 3])
+    Conway _ -> Just (Set.fromList [mkTxInPartial (TxId (mkDummySafeHash Proxy 1)) 3])
+
+exampleTxOut :: Proof era -> Core.Value era -> Core.TxOut era
+exampleTxOut proof spendval =
+  case proof of
+    Shelley _ -> ShelleyTxOut (mkAddr (examplePayKey, exampleStakeKey)) spendval -- (Coin 100000)
+    Allegra _ -> ShelleyTxOut (mkAddr (examplePayKey, exampleStakeKey)) spendval -- exampleCoin
```
lehins: Dead code
```suggestion
    Shelley _ -> ShelleyTxOut (mkAddr (examplePayKey, exampleStakeKey)) spendval
    Allegra _ -> ShelleyTxOut (mkAddr (examplePayKey, exampleStakeKey)) spendval
```

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/Consensus.hs
```diff
@@ -0,0 +1,1166 @@
...
+
+-- ================================================================================
+
+-- | Build a Tx from a list of maybe fields. The maybe is usfull because sometimes we
+--   only want to add a field in a particular era, so we might say something like this
+-- genericTx proof
+--  [Just field1
+--  , if (Some proof) >= Mary Mock then Just field2 else Nothing
+--  , Just field3
+--  ]
```
lehins: Spelling and haddock formatting
```suggestion
-- | Build a Tx from a list of maybe fields. The maybe is useful because sometimes we
--   only want to add a field in a particular era, so we might say something like this
--
-- @
--   genericTx proof
--     [ Just field1
--     , if Some proof >= Mary Mock then Just field2 else Nothing
--     , Just field3
--     ]
-- @
```

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/Consensus.hs
```diff
@@ -0,0 +1,1166 @@
...
+genericTx :: Proof era -> [Maybe (TxField era)] -> Core.Tx era
+genericTx proof xs = newTx proof (foldl' accum [] xs)
+  where
+    accum ans Nothing = ans
+    accum ans (Just field) = field : ans
+
+-- | Build a Tx from a list of maybe fields. The maube is usfull because sometimes we
```
lehins:
```suggestion
-- | Build a Tx from a list of maybe fields. The maybe is useful because sometimes we
```

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/Consensus.hs
```diff
@@ -0,0 +1,1166 @@
...
+--   only want to add a field in a particular era, so we might say something like this
+-- genericTxBody proof
+--  [Just field1
+--  , if (Some proof) >= Mary Mock then Just field2 else Nothing
+--  , Just field3
+--  ]
```
lehins, line 900:
```suggestion
--
-- @
--   genericTxBody proof
--     [ Just field1
--     , if Some proof >= Mary Mock then Just field2 else Nothing
--     , Just field3
--     ]
-- @
```

eras/shelley-ma/impl/src/Cardano/Ledger/Allegra.hs
```diff
@@ -20,6 +20,7 @@ module Cardano.Ledger.Allegra
   )
 where
 
+import Cardano.Ledger.Allegra.Translation (AllegraEra) -- Bring "instance TranslationContext (AllegraEra c)" into scope.
```
lehins: Another occurrence we don't need to comment about it:
```suggestion
import Cardano.Ledger.Allegra.Translation ()
```

eras/conway/impl/src/Cardano/Ledger/Conway.hs
```diff
@@ -43,6 +44,8 @@ import Data.Maybe.Strict
 import qualified Data.Set as Set
 import Lens.Micro
 
+-- Bring "instance TranslationContext (ConwayEra c)" into scope.
```
lehins: This comment is not needed
```suggestion
```

eras/babbage/impl/src/Cardano/Ledger/Babbage.hs
lehins: Same here
```suggestion
```

eras/alonzo/impl/src/Cardano/Ledger/Alonzo.hs
```diff
@@ -70,6 +71,8 @@ import qualified Data.Set as Set
 import Lens.Micro
 import Lens.Micro.Extras (view)
 
+-- Bring "instance TranslationContext (AlonzoEra c)" into scope.
```
lehins: Same here
```suggestion
```

eras/shelley-ma/impl/src/Cardano/Ledger/Mary.hs
```diff
@@ -32,6 +32,7 @@ import Cardano.Ledger.Core (EraCrypto)
 import Cardano.Ledger.Crypto (Crypto, StandardCrypto)
 import Cardano.Ledger.Hashes (EraIndependentTxBody)
 import Cardano.Ledger.Keys (DSignable)
+import Cardano.Ledger.Mary.Translation (MaryEra) -- Bring "instance TranslationContext (MaryEra c)" into scope.
```
lehins: Same here
```suggestion
import Cardano.Ledger.Mary.Translation ()
```

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/Consensus.hs
```diff
@@ -0,0 +1,1166 @@
...
+      (3, List [I 1, I 2]),
+      (4, Map [(I 3, B "b")])
+    ]
+
+exampleAuxiliaryDataShelley :: Core.AuxiliaryData StandardShelley
+exampleAuxiliaryDataShelley = Metadata exampleMetadataMap
+
+-- ======================
+
+type StandardAllegra = AllegraEra StandardCrypto
```
lehins:
```suggestion
```

lehins (changes requested):
This looks great for the most part. Also needs a rebase on master.

eras/babbage/impl/src/Cardano/Ledger/Babbage.hs
TimSheard: done

eras/shelley-ma/impl/src/Cardano/Ledger/Mary/Translation.hs
TimSheard: MaryEra  is used 20 times in this file.  But because of the refatoring, it cannot be imported from Cardano.Ledger.Mary (where it was previously defined) without causing cyclic imports. So I now define it in Cardano.Ledger.ShelleyMA.Era, And it is imported from there (both here and inside of Cardano.Ledger.Shelley)

TimSheard: I used the same trick for both MaryEra and AllegraEra.

eras/shelley/impl/src/Cardano/Ledger/Shelley.hs
TimSheard: done

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/Same.hs
TimSheard: done

eras/shelley-ma/impl/src/Cardano/Ledger/Mary/Translation.hs
lehins: Perfect!

eras/shelley/impl/cardano-ledger-shelley.cabal
JaredCorduan: Sounds like more great information for our contributing doc! :grin:

eras/shelley-ma/impl/src/Cardano/Ledger/Allegra/Translation.hs
TimSheard: Because of the 19 uses of AllegraEra in this file, and the fact that one could not import it due to circular imports. Like for Mary, I moved the type synonym of AllegraEra to   Cardano.Ledger.ShelleyMA.Era, and import it from there, and have Cardano.Ledger.Allegra reexport it.

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/Same.hs
TimSheard: StashedAVVMAddresses  is a type family, so it does not have an Eq or Show instance. So I punted when I put this together. I have now written 
sameStashedAVVMAddresses ::
  Proof era -> StashedAVVMAddresses era -> StashedAVVMAddresses era -> Maybe PDoc
which uses a case analysis over the proof.

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/Same.hs
TimSheard: I partly agree with you, but there is something about the beauty of pattern matching that is lost when you do this. So I applied this suggestion in 6 places, but did not use it for TxWits, TxBody, and Tx.

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Examples/Consensus.hs
TimSheard: Unfortunately, this does not work to use specialize because old,new :: ShelleyLedgerExamples era
and ShelleyLedgerExamples  is a type family that cannot be partially applied. 
But we can remove the parenthesis.

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/Same.hs
TimSheard: Doesn't quite work. Specialize does not apply very often. But I did factor out some stuff.
"
instance Same era (ShelleyResultExamples era) where
  same proof r1 r2 =
    [ ("PParams", samePParams proof (srePParams r1) (srePParams r2)),
      ( "ProposedPPUpdates",  
        case proof of
          Shelley _ -> sameProposedPPUpdates (sreProposedPPUpdates r1) (sreProposedPPUpdates r2)
          Allegra _ -> sameProposedPPUpdates (sreProposedPPUpdates r1) (sreProposedPPUpdates r2)
          Mary _ ->sameProposedPPUpdates (sreProposedPPUpdates r1) (sreProposedPPUpdates r2)
          Alonzo _ -> sameProposedPPUpdates (sreProposedPPUpdates r1) (sreProposedPPUpdates r2)
          Babbage _ -> sameProposedPPUpdates (sreProposedPPUpdates r1) (sreProposedPPUpdates r2)
          Conway _ -> sameProposedPPUpdates (sreProposedPPUpdates r1) (sreProposedPPUpdates r2) ),
      ("poolDistr", eqByShow (srePoolDistr r1) (srePoolDistr r2)),
      ("NonMyopicRewards", eqByShow (sreNonMyopicRewards r1) (sreNonMyopicRewards r2)),
      ("ShelleyGenesis", eqByShow (sreShelleyGenesis r1) (sreShelleyGenesis r2))
    ]
    where
      getmap (ProposedPPUpdates x) = x
      sameProposedPPUpdates x y =  ppDiff $ mapdiffs (getmap x) (getmap y)
        where ppDif [] = Nothing
              ppDiff xs = Just (ppString (show xs))"

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/Same.hs
TimSheard: At one time it did not have a Show instances. But it does now.

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/Same.hs
TimSheard: No, that is not the case. Specialize only works if the context needed in every case is a real class (like Eq or Show)
In order to get it to typecheck one usually has to type apply the call, to be concrete about what that class is.
For example we may say  "specialize @Eq proof x y"  Many times the constraints differ from Era to Era,  or the constraint is some compound constrant like  \ z -> (Eq z, Show z)  which we can't really express. 

I wish we had a better more flexible approach, but I have yet to find it.

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/Consensus.hs
TimSheard: done

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/Consensus.hs
TimSheard: done

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/Consensus.hs
TimSheard: done

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/Consensus.hs
TimSheard: done

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/Consensus.hs
TimSheard: done

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/Consensus.hs
TimSheard: done

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Examples/Consensus.hs
TimSheard: done

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/Same.hs
lehins: I see, the `constraint` type parameter in `specialize` needs to be instantiated to a concrete class. I wish there were anonymous lambdas at the type level. At this point it can only be done with named type synonym constraints, eg.
```
type  Combine z = (Eq z, Show z)
```

Even if it would work with specialize, I don't think it is worth littering the code with bunch of random type synonyms.

Don't worry about this, I was just curious.

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Examples/Consensus.hs
TimSheard: done

eras/conway/impl/src/Cardano/Ledger/Conway.hs
TimSheard: done

eras/babbage/impl/src/Cardano/Ledger/Babbage.hs
TimSheard: done

eras/shelley-ma/impl/src/Cardano/Ledger/Mary.hs
TimSheard: done

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/Consensus.hs
TimSheard: This mimics the code we are trying to replace. For each Era they make a type synonym. This is done in this file for all the other Eras. So I will leave it for consistency.

eras/alonzo/impl/src/Cardano/Ledger/Alonzo.hs
TimSheard: done

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/Same.hs
TimSheard: done

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/Same.hs
TimSheard: done

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/Same.hs
TimSheard: done

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/Consensus.hs
TimSheard: Well,  lets see.  Ah, A proof is a proxy, one with a witness, it keeps the type of the function consistent with all the others, makes it easier to follow.

lehins: That's cool

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/Consensus.hs
TimSheard: done

TimSheard: done

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/Consensus.hs
TimSheard: If we use the type synonyms, they clash with the proofs. So I am leaving this alone. Think of this type synonym as a local self documenting comment.

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/Consensus.hs
TimSheard: same here

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/Consensus.hs
TimSheard: same here, using Alonzo Clashes.

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/Consensus.hs
TimSheard: clashes

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Examples/Consensus.hs
TimSheard: clashes cause more confusion. Lets leave it alone.

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/Consensus.hs
TimSheard: The comments were meant to be a reminder what the original examples returned. No longer needed.

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/Consensus.hs
lehins: They don't clash with Proofs, Proofs are at the value level, while type synonyms are at the type level.

We don't want to have every single package redefining theses type synonyms. I removed 5 different copies already throughout tests suites, all with different names.

I created those on purpose, because we need to be consistent. They are also exported in the API.

If you not gonna delete them, I'll have to do that later. Please Tim, delete these duplicate type synonyms

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/Consensus.hs
TimSheard: done

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/Consensus.hs
TimSheard: done

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/Consensus.hs
TimSheard: done

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/Consensus.hs
TimSheard: done

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/Consensus.hs
TimSheard: done

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/Consensus.hs
TimSheard: done

eras/shelley-ma/impl/src/Cardano/Ledger/Allegra.hs
TimSheard: done

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/Consensus.hs
TimSheard: done

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/Consensus.hs
TimSheard: Using Mary causes a clash. think of this as local self documenting code.

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/Consensus.hs
TimSheard: I got rid of this.

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/Consensus.hs
TimSheard: I inlined the type synonyms. They each had just one use.

lehins (approved):

JaredCorduan:
:tada:

PR #3011, opened 2022-08-31
Consolidate and deprecate minfee and allOuts

lehins:
We have these classes `CLI` and `ExtendedUTxO` that bunch together somewhat related functions. However now we have type classes that could consume functionality from those classes in a more related fashion.

Also we have multiple functions with the same name like `minfee` which are used in different eras, which has caused bugs on multiple occasions. Here we adopt a type class approach to avoid such bugs in the future

- Deprecated `minfee` and `CLI.evaluateMinfee` in favor of new `EraTx.getMinFeeTx`
- Deprecated `ExtendedUTxO.getTxOutDatum` in favor of new `AlonzoEraTxOut.datumTxOutF`
- Removed `ExtendedUTxO.allOuts` and `ExtendedUTxO.allSizedOuts` in favor of
  `BabbageEraTxBody.allInputsTxBodyF`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,8 +13,9 @@ in the naming of release branches.
 
 - Start on the `cardano-ledger-api` package and implement
   `setMinCoinTxOut`+`setMinCoinSizedTxOut`: #2995
-- Added `getMinCoinTxOut`/`getMinCoinSizedTxOut`: #3008
-- Added `getMinFeeTx`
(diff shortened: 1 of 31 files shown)


JaredCorduan (approved):
Looks great to me! (the CI failure is spurious).

lehins:
> CI failure is spurious

Yeah I had to rerun hydra due to recurring failing test. Looks like it just didn't report correctly that the re-run has passed

PR #3016, opened 2022-08-31
Fix overloaded function name `consumed`

lehins:
Deprecated `consumed` and `evaluateConsumed` in favor of new `EraUTxO.getConsumedValue`.

Create a `EraUTxO` class that will have more functions added to it in subsequent PRs

### DIFF
--- a/libs/cardano-ledger-api/src/Cardano/Ledger/Api/UTxO.hs
+++ b/libs/cardano-ledger-api/src/Cardano/Ledger/Api/UTxO.hs
@@ -1,0 +1,7 @@
+module Cardano.Ledger.Api.UTxO
+  ( UTxO (..),
+    EraUTxO (..),
+  )
+where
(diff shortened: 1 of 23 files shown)


libs/cardano-ledger-api/src/Cardano/Ledger/Api/UTxO.hs
```diff
@@ -0,0 +1,7 @@
+module Cardano.Ledger.Api.UTxO
+  ( UTxO (..),
+    EraUTxO (..),
+  )
+where
+
+import Cardano.Ledger.Shelley.UTxO (EraUTxO (..), UTxO (..))
```
JaredCorduan, line 7: :tada: the API grows! :seedling:

JaredCorduan (approved):
this now look much more organized!

PR #3018, opened 2022-08-31
Reduce duplication remove input data hashes

lehins:
The recently added to Alonzo era lens `datumTxOutF` allows us to remove duplicate implementations for `inputDataHashes`. That is exactly what this PR does.


teodanciu (approved):
Looks good to me

PR #3019, opened 2022-09-02
Make scripts needed part of utxo class

lehins:
Add `getScriptsNeeded` to the `EraUTxO`:

This approach allows us to move `mintedTxBodyF` into ShelleyMA era, where it belongs. It also unifies the spec and code, where spec uses the same function `scriptsNeeded` in all eras. In order to achieve this we need a special type family `ScriptsNeeded`, since eras do not agree in the return type of the function.

Model testing framework is conflicting with this change and will require significant effort to make it work. At the team level we've already noticed that that framework is starting to bitrot and we agreed to remove it. This PR does that as well.


eras/alonzo/impl/src/Cardano/Ledger/Alonzo/PlutusScriptApi.hs
```diff
@@ -147,8 +136,10 @@ knownToNotBe1Phase scriptsAvailable (sp, sh) = do
 collectTwoPhaseScriptInputs ::
   forall era.
   ( EraTx era,
-    ShelleyEraTxBody era,
+    ShelleyMAEraTxBody era,
     AlonzoEraTxWits era,
+    EraUTxO era,
+    ScriptsNeeded era ~ [(ScriptPurpose (EraCrypto era), ScriptHash (EraCrypto era))],
```
JaredCorduan: Would it be clearer, and more consistent with the rest of the code, if we used a new type for `[(ScriptPurpose (EraCrypto era), ScriptHash (EraCrypto era))]` and called it something like `AlonzoScriptsNeeded`?

JaredCorduan: ha ha, I should have read the whole PR before commenting. It seem you've already done exactly this elsewhere. So maybe just this spot and a couple others need the newtype instead?

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/UTxO.hs
```diff
@@ -53,3 +74,76 @@ getInputDataHashesTxBody (UTxO mp) txBody hashScriptMap =
...
+--    and reference inputs are never authorized. That might not always be the case.
+-- 2) Rewarding (Withdrawals),
+-- 3) Minting (minted field), and
+-- 4) Certifying (Delegating) scripts.
+--
+-- 'getAlonzoScriptsNeeded' is an aggregation of the needed Credentials referring to
+-- Scripts used in Utxow rule.  The flip side of 'getAlonzoScriptsNeeded' (which collects
+-- script hashes) is 'txscripts' which finds the actual scripts. We maintain an invariant
+-- that every script credential refers to some actual script.  This is tested in the test
+-- function 'validateMissingScripts' in the Utxow rule.
```
JaredCorduan, line 122: thanks for documenting this so well!

eras/shelley/impl/src/Cardano/Ledger/Shelley/UTxO.hs
```diff
@@ -395,7 +404,19 @@ keyRefunds pp tx = length deregistrations <×> getField @"_keyDeposit" pp
...
+
+  -- | Calculate all the value that is being consumed by the transaction.
   getConsumedValue :: PParams era -> UTxO era -> TxBody era -> Value era
 
+  -- | Produce all the information required for figuring out which scripts are required
+  -- for the transaction to be valid, once those scripts are evaluated
+  getScriptsNeeded :: UTxO era -> TxBody era -> ScriptsNeeded era
+
 instance Crypto c => EraUTxO (ShelleyEra c) where
+  type ScriptsNeeded (ShelleyEra c) = Set (ScriptHash c)
```
JaredCorduan: should we use a newtype to name this better, ie `ShelleyScriptsNeeded`?

JaredCorduan (approved):
looks wonderful to me! I had just a couple questions/comments about when you chose to give newtype wrappers to things.

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/UTxO.hs
lehins: Oh, no. That's not me. I just merely moved the haddock over from another module :smile:

eras/shelley/impl/src/Cardano/Ledger/Shelley/UTxO.hs
lehins: Yeah, sure, why not. I'll make both ShelleyScriptsNeeded and AlonzoScriptsNeeded into the newtypes.

JaredCorduan: thanks!

lehins: Switched to `newtypes` and added `ShelleyScriptsNeeded`

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/PlutusScriptApi.hs
lehins: Done.

PR #3020, opened 2022-09-02
Split GenState module into three

lehins:
It looks like on hydra there is a timeout limit that kills a CI job when there is no output for a certain period of time. It so happens that when GenState module is build on a mac-mini it takes just long enough for the CI job to get terminated during the build phase

A simple solution to this problem is to split the module into two smaller ones, that is what this PR does.


Soupstraw (approved):
`GenState.hs` seems a bit redundant, since `Generators.hs` already re-exports `Types.hs`. Other than that, looks fine to me.

PR #3024, opened 2022-09-07
Added TxOut modules for each era

### LINKED ISSUE #2930: Split `TxOut` into its own module
Currently all `TxOut` definitions in all eras are placed into the `TxBody.hs` module

In some eras, like Alonzo and Babbage `TxOut` implementation is fairly involved, so it makes sense to move those into their own module. This should be done consistently across all eras, even for those that do not define the TxOut, but only create the instances.

Soupstraw:
This PR creates separate modules in each era for `TxOut` and all the related instances/functions.

closes #2930


lehins (approved):
Great stuff!. Thank you!

PR #3029, opened 2022-09-08
Polymorphise Alonzo types

lehins:
This PR makes Alonzo types to use type families from core rather than concrete types. This simplifies the constraints and improves future compatibility.

### DIFF
--- a/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Tx.hs
+++ b/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Tx.hs
@@ -69,7 +69,7 @@ module Cardano.Ledger.Alonzo.Tx
     rdptrInv,
     getMapFromValue,
     -- Segwit
-    segwitTx,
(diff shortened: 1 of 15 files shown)


eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Tx.hs
```diff
@@ -168,7 +162,7 @@ newtype IsValid = IsValid Bool
 
 data AlonzoTx era = AlonzoTx
   { body :: !(Core.TxBody era),
-    wits :: !(AlonzoTxWits era),
+    wits :: !(Core.TxWits era),
     isValid :: !IsValid,
```
TimSheard, line 166: 15 files are changed because of this tiny change. Who knew?

TimSheard (approved):
Lots of changes, but they are all very straightforward. Looks good to me.

JaredCorduan:
:raised_hands:

PR #3030, opened 2022-09-08
Update plutus version

teodanciu:
This is an attempt to isolate changes from https://github.com/input-output-hk/cardano-ledger/pull/2992, to see what's causing the build error in plutus-core that `cabal build all` is attempting to build. 

Turns out the problem is happening because the `hashable` library was upgraded in plutus to 1.4.10, but in our project, the 
 dependency on `Unique` is restricting the version of `hashable` to 1.4.  It worked after adding the `Unique:hashable` to allow-newer  in cabal.project. 

~Now that the dependency problems are solved, the remaining issue is adapting to the new code. I have tried  (last commit) and might have partially succeeded, but I got stuck at  the `ValidatorHash` type that seems to have been removed from the API and ran out of time investigating what it was replaced with.~  EDIT: Sorted this out now

I think it would be good to sort this PR out and merge it, and then resume the original prelude-removal one  https://github.com/input-output-hk/cardano-ledger/pull/2992, because the latter depends on this PR.


Unisay:
FYI: https://github.com/kapralVV/Unique/pull/10

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Scripts.hs
```diff
@@ -428,8 +442,8 @@ instance Era era => FromCBOR (Annotator (Script era)) where
 validScript :: ProtVer -> Script era -> Bool
 validScript pv scrip = case scrip of
   TimelockScript sc -> deepseq sc True
-  PlutusScript PlutusV1 bytes -> PV1.isScriptWellFormed (transProtocolVersion pv) bytes
-  PlutusScript PlutusV2 bytes -> PV2.isScriptWellFormed (transProtocolVersion pv) bytes
+  PlutusScript PlutusV1 bytes -> (not . null) (PV1.assertScriptWellFormed (transProtocolVersion pv) bytes :: Either ScriptDecodeError ())
+  PlutusScript PlutusV2 bytes -> (not . null) (PV2.assertScriptWellFormed (transProtocolVersion pv) bytes :: Either ScriptDecodeError ())
```
lehins, line 446: It is better to use specialized function `isRight` instead of relying on Foldable instance for Either:

lehins (changes requested):
I'll close this PR, since I've incorporated it into #2992

lehins:
Closed it in favor of #2992

teodanciu:
Nice, thank you for making something usable from two partial things!

Issue #3001, opened 2022-08-25
Rename AuxiliaryData consistently across all eras


lehins:
Current naming is quite inconsistent and should be fixed.

The consensus throughout the team is to name the core type family `TxAuxData`, which means:

* The class will look like this:
```haskell
-- | TxAuxData which may be attached to a transaction
class
  ( Era era,
    Eq (TxAuxData era),
    Show (TxAuxData era),
    NoThunks (TxAuxData era),
    ToCBOR (TxAuxData era),
    FromCBOR (Annotator (TxAuxData era)),
    HashAnnotated (TxAuxData era) EraIndependentTxAuxData (Crypto era)
  ) =>
  EraTxAuxData era
  where
  type TxAuxData era = (r :: Type) | r -> era
  hashTxAuxData :: TxAuxData era -> TxAuxDataHash (Crypto era)
  validateTxAuxData :: ProtVer -> TxAuxData era -> Bool
```

* And the concrete types per era will be placed into `Cardano.Ledger.[Era].TxAuxData`. This would be an example on Alonzo

```haskell

newtype AlonzoTxAuxData era = TxAuxDataConstr (MemoBytes TxAuxDataRaw era)
  deriving newtype (ToCBOR, SafeToHash)

type TxAuxData era = AlonzoTxAuxData era

{-# DEPRECATED TxAuxData "Use `AlonzoTxAuxData` instead" #-}

instance CC.Crypto c => EraTxAuxData (AlonzoEra c) where
  type TxAuxData (AlonzoEra c) = AlonzoTxAuxData (AlonzoEra c)
  hashTxAuxData = hashAlonzoTxAuxData
  validateTxAuxData = validateAlonzoTxAuxData
```

TimSheard:
This issue was closed by PR #3032

PR #3032, opened 2022-09-12
Adresses issue #3001, Renames AuxiliaryData consistently across eras.

TimSheard:
Addresses issue https://github.com/input-output-hk/cardano-ledger/issues/3001
Renames AuxiliaryData type families consistently across eras to TxAuxData. Actual instances are called
ShelleyTxAuxData
AllegraTxAuxData
AlonzoTxAuxData

### DIFF
--- a/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Rules/Utxow.hs
+++ b/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Rules/Utxow.hs
@@ -168,7 +168,7 @@ instance
 instance
   ( Era era,
     ToCBOR (PredicateFailure (EraRule "UTXO" era)),
-    Typeable (AuxiliaryData era),
+    Typeable (TxAuxData era),
     Typeable (Script era),
     ToCBOR (Script era)
   ) =>
@@ -183,7 +183,7 @@ encodePredFail ::
   ( Era era,
     ToCBOR (PredicateFailure (EraRule "UTXO" era)),
     Typeable (Script era),
-    Typeable (AuxiliaryData era)
+    Typeable (TxAuxData era)
   ) =>
   AlonzoUtxowPredFailure era ->
   Encode 'Open (AlonzoUtxowPredFailure era)
@@ -200,7 +200,7 @@ instance
   ( Era era,
     FromCBOR (PredicateFailure (EraRule "UTXO" era)),
     Typeable (Script era),
-    Typeable (AuxiliaryData era)
+    Typeable (TxAuxData era)
   ) =>
   FromCBOR (AlonzoUtxowPredFailure era)
   where
@@ -210,7 +210,7 @@ decodePredFail ::
   ( Era era,
     FromCBOR (PredicateFailure (EraRule "UTXO" era)), -- TODO, we should be able to get rid of this constraint
     Typeable (Script era),
-    Typeable (AuxiliaryData era)
+    Typeable (TxAuxData era)
   ) =>
   Word ->
   Decode 'Open (AlonzoUtxowPredFailure era)
@@ -500,7 +500,7 @@ extSymmetricDifference as fa bs fb = (extraA, extraB)
 instance
   forall era.
   ( AlonzoEraTx era,
-    EraAuxiliaryData era,
+    EraTxAuxData era,
     ExtendedUTxO era,
     Signable (DSIGN (EraCrypto era)) (Hash (HASH (EraCrypto era)) EraIndependentTxBody),
     Script era ~ AlonzoScript era,
--- a/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Tx.hs
+++ b/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Tx.hs
@@ -164,7 +164,7 @@ data AlonzoTx era = AlonzoTx
   { body :: !(Core.TxBody era),
     wits :: !(Core.TxWits era),
     isValid :: !IsValid,
-    auxiliaryData :: !(StrictMaybe (AuxiliaryData era))
+    auxiliaryData :: !(StrictMaybe (TxAuxData era))
   }
   deriving (Generic)
 
@@ -218,9 +218,9 @@ witsAlonzoTxL :: Lens' (AlonzoTx era) (TxWits era)
 witsAlonzoTxL = lens wits (\tx txWits -> tx {wits = txWits})
 {-# INLINEABLE witsAlonzoTxL #-}
 
--- | `AuxiliaryData` setter and getter for `AlonzoTx`.
-auxDataAlonzoTxL :: Lens' (AlonzoTx era) (StrictMaybe (AuxiliaryData era))
-auxDataAlonzoTxL = lens auxiliaryData (\tx txAuxiliaryData -> tx {auxiliaryData = txAuxiliaryData})
+-- | `TxAuxData` setter and getter for `AlonzoTx`.
+auxDataAlonzoTxL :: Lens' (AlonzoTx era) (StrictMaybe (TxAuxData era))
+auxDataAlonzoTxL = lens auxiliaryData (\tx txTxAuxData -> tx {auxiliaryData = txTxAuxData})
 {-# INLINEABLE auxDataAlonzoTxL #-}
 
 -- | txsize computes the length of the serialised bytes
@@ -233,25 +233,24 @@ isValidAlonzoTxL = lens isValid (\tx valid -> tx {isValid = valid})
 {-# INLINEABLE isValidAlonzoTxL #-}
 
 deriving instance
-  (Era era, Eq (Core.TxBody era), Eq (TxWits era), Eq (AuxiliaryData era)) =>
-  Eq (AlonzoTx era)
+  (Era era, Eq (Core.TxBody era), Eq (TxWits era), Eq (TxAuxData era)) => Eq (AlonzoTx era)
 
 deriving instance
-  (Era era, Show (Core.TxBody era), Show (TxWits era), Show (AuxiliaryData era)) =>
+  (Era era, Show (Core.TxBody era), Show (TxAuxData era), Show (Script era), Show (TxWits era)) =>
   Show (AlonzoTx era)
 
 instance
   ( Era era,
-    NoThunks (AuxiliaryData era),
     NoThunks (TxWits era),
+    NoThunks (TxAuxData era),
     NoThunks (Core.TxBody era)
   ) =>
   NoThunks (AlonzoTx era)
 
 instance
   ( Era era,
-    NFData (AuxiliaryData era),
     NFData (TxWits era),
+    NFData (TxAuxData era),
     NFData (Core.TxBody era)
   ) =>
   NFData (AlonzoTx era)
@@ -320,7 +319,7 @@ isTwoPhaseScriptAddress tx =
 toCBORForSizeComputation ::
   ( ToCBOR (Core.TxBody era),
     ToCBOR (TxWits era),
-    ToCBOR (AuxiliaryData era)
+    ToCBOR (TxAuxData era)
   ) =>
   AlonzoTx era ->
   Encoding
@@ -484,7 +483,7 @@ alonzoSegwitTx ::
   Annotator (Core.TxBody era) ->
   Annotator (TxWits era) ->
   IsValid ->
-  Maybe (Annotator (AuxiliaryData era)) ->
+  Maybe (Annotator (TxAuxData era)) ->
   Annotator (Tx era)
 alonzoSegwitTx txBodyAnn txWitsAnn isValid auxDataAnn = Annotator $ \bytes ->
   let txBody = runAnnotator txBodyAnn bytes
@@ -517,7 +516,7 @@ alonzoSegwitTx txBodyAnn txWitsAnn isValid auxDataAnn = Annotator $ \bytes ->
 toCBORForMempoolSubmission ::
   ( ToCBOR (Core.TxBody era),
     ToCBOR (TxWits era),
-    ToCBOR (AuxiliaryData era)
+    ToCBOR (TxAuxData era)
   ) =>
   AlonzoTx era ->
   Encoding
@@ -533,8 +532,8 @@ toCBORForMempoolSubmission
 instance
   ( Era era,
     ToCBOR (Core.TxBody era),
-    ToCBOR (TxWits era),
-    ToCBOR (AuxiliaryData era)
+    ToCBOR (TxAuxData era),
+    ToCBOR (TxWits era)
   ) =>
   ToCBOR (AlonzoTx era)
   where
@@ -544,7 +543,7 @@ instance
   ( Typeable era,
     FromCBOR (Annotator (Core.TxBody era)),
     FromCBOR (Annotator (Core.TxWits era)),
-    FromCBOR (Annotator (AuxiliaryData era))
+    FromCBOR (Annotator (TxAuxData era))
   ) =>
   FromCBOR (Annotator (AlonzoTx era))
   where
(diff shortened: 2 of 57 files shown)


eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Rules/Utxow.hs
```diff
@@ -500,7 +500,9 @@ extSymmetricDifference as fa bs fb = (extraA, extraB)
 instance
   forall era.
   ( AlonzoEraTx era,
-    EraAuxiliaryData era,
+    EraTx era,
+    AlonzoEraTxBody era,
```
lehins: These two constraints are redundant, they are implied by `AlonzoEraTx`
```suggestion
```

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Tx.hs
```diff
@@ -233,26 +234,45 @@ isValidAlonzoTxL = lens isValid (\tx valid -> tx {isValid = valid})
 {-# INLINEABLE isValidAlonzoTxL #-}
 
 deriving instance
-  (Era era, Eq (Core.TxBody era), Eq (TxWits era), Eq (AuxiliaryData era)) =>
+  ( Era era,
+    Eq (Core.TxBody era),
+    Eq (TxAuxData era),
+    Eq (TxWits era)
+  ) =>
```
lehins: There is no need to make this multiline
```suggestion
  (Era era, Eq (Core.TxBody era), Eq (TxWits era), Eq (TxAuxData era)) =>
```

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Tx.hs
```diff
@@ -233,26 +234,45 @@ isValidAlonzoTxL = lens isValid (\tx valid -> tx {isValid = valid})
...
   Eq (AlonzoTx era)
 
 deriving instance
-  (Era era, Show (Core.TxBody era), Show (TxWits era), Show (AuxiliaryData era)) =>
+  ( Era era,
+    Show (Core.TxBody era),
+    Show (TxAuxData era),
+    Show (Script era),
+    Show (TxWits era)
+  ) =>
```
lehins: Same here:
```suggestion
  (Era era, Show (Core.TxBody era), Show (TxWits era), Show (TxAuxData era)) =>
```

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Tx.hs
```diff
@@ -233,26 +234,45 @@ isValidAlonzoTxL = lens isValid (\tx valid -> tx {isValid = valid})
...
 instance
   ( Era era,
-    NoThunks (AuxiliaryData era),
     NoThunks (TxWits era),
-    NoThunks (Core.TxBody era)
+    NoThunks (TxAuxData era),
+    NoThunks (Script era),
+    NoThunks (Core.TxBody era),
+    NoThunks (Value era),
+    NoThunks (PParamsUpdate era)
```
lehins: Redundant constraints:
```suggestion
    NoThunks (Core.TxBody era)
```

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Tx.hs
```diff
@@ -233,26 +234,45 @@ isValidAlonzoTxL = lens isValid (\tx valid -> tx {isValid = valid})
...
-    NFData (Core.TxBody era)
+    Script era ~ AlonzoScript era,
+    c ~ EraCrypto era,
+    NFData (TxAuxData era),
+    NFData (Script era),
+    NFData (Core.TxBody era),
+    NFData (Value era),
+    NFData (PParamsUpdate era),
+    NFData (TxDats era),
+    NFData (Redeemers era)
```
lehins: Redundant constraints:
```suggestion
    NFData (TxAuxData era),
    NFData (Core.TxBody era)
```

eras/shelley-ma/impl/src/Cardano/Ledger/ShelleyMA/AuxiliaryData.hs
```diff
@@ -84,53 +87,54 @@ deriving instance NoThunks (Script era) => NoThunks (AuxiliaryDataRaw era)
 
 instance NFData (Script era) => NFData (AuxiliaryDataRaw era)
 
-newtype MAAuxiliaryData era = AuxiliaryDataWithBytes (MemoBytes AuxiliaryDataRaw era)
+-- We call it AllegraTxAuxData because it is first used in the Allegra Era
```
lehins: I don't think this comment is needed, this is a common pattern throughout the codebase.
```suggestion
```

Solid reasoning on choosing the name though!!!

libs/cardano-ledger-pretty/src/Cardano/Ledger/Pretty/Alonzo.hs
```diff
@@ -58,7 +58,8 @@ ppTxSeq (AlonzoTxSeq xs) =
 instance
   ( AlonzoEraTx era,
     SafeToHash (TxWits era),
-    PrettyA (Tx era)
+    PrettyA (Tx era),
+    EraTx era
```
lehins: Redundant constraint, implied by: `AlonzoEraTx`
```suggestion
    PrettyA (Tx era)
```

lehins (changes requested):
This looks great! There are a few lines that were reverted when compared to master during a rebase, but I marked all of them in the comments with concrete diffs that can be applied directly from the PR review.

One more thing is worth doing is a changelog entry, since that will affect downstream users.

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Tx.hs
teodanciu: I never understood whether ormulu has opinions on this -  and this comment prompted me to try to understand.
Looks like if you split it on multilines, ormulu respects that . But on the other hand, it doesn't seem to know to split the line itself beyond a certain number of characters, which is what I  would have expected! 

So for example this has been "ormollized" successfully: 



So I wonder then what is our convention for line width?  Is there an implicit one we're having?

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Tx.hs
teodanciu: Interesting that the compiler didn't  complain about this ..Compiler 0 - @lehins 1 

And actually not even haskell -language-server does (assuming it works correctly)

eras/shelley/test-suite/src/Test/Cardano/Ledger/Shelley/Examples/Consensus.hs
```diff
@@ -442,17 +442,17 @@ exampleTxBodyShelley =
     auxiliaryDataHash =
       AuxiliaryDataHash $ mkDummySafeHash (Proxy @StandardCrypto) 30
 
-exampleMetadataMap :: Map Word64 Metadatum
-exampleMetadataMap =
+exampleAuxDataMap :: Map Word64 Metadatum
```
teodanciu, line 445: Wondering why the rename, since the type is  still `Metadatum` ?

eras/shelley/test-suite/test/Test/Cardano/Ledger/Shelley/Serialisation/Tripping/CBOR.hs
```diff
@@ -273,7 +273,7 @@ tests =
       testProperty "roundtrip Epoch State" prop_roundtrip_EpochState,
       testProperty "roundtrip NewEpoch State" prop_roundtrip_NewEpochState,
       testProperty "roundtrip MultiSig" prop_roundtrip_MultiSig,
-      testProperty "roundtrip Metadata" prop_roundtrip_metadata,
+      testProperty "roundtrip Metadata" prop_roundtrip_auxData,
```
teodanciu: Should we also change the description  in the first argument  to `roundtrip AuxData` or it's been left for a reason ?

teodanciu (commented):
Looks good to me, added a few small questions and wonderings.  Will leave it to @lehins to approve.

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Tx.hs
lehins: That's a good question. However, ormolu is only tangentially related to this question. Ormolu formats the code, but it does not enforce the line length. That is an impossible problem, because it would require customization, which ormolu does not support.

I'll give you my personal guideline and personal motivation behind it:

* Using 80 character limit was imposed in the early days when terminals ability was limited to displaying only 80 characters in one line, so this archaic guideline is no longer relevant
* What is relevant, however, is that many people, including myself, use split screen. Eg. two emacs buffers side-by-side, or an editor and a terminal. And using 80 characters works pretty good. I also found that using something like 90 to a 100 character lines work even better and causes no wraparound while using all default OS settings, eg. font size etc. So I'd say around 90 character limit is really a sweet spot.
* Using lines much longer than 100 characters, like in the example above, not only makes it impossible to read in some editors, but also it makes it impossible to review a PR with such a change, because Github does not wraparound when showing the diff, so it will just cut off right half of that diff
* Using lines that are too short on the other hand where the whole thing could fit into one slightly longer line utilizes the space least efficiently and makes you scroll through many more lines in order to be able to read the code. And as I am sure we all know, more lines of code does not mean better code :smile: 

All of those point are not a hard guideline, of course, code readability and conventions must be the priority. 

For instance, using semicolons is absolutely legal in Haskell, however it is unconventional, thus it is extremely rare when you'll see someone using them, consequently making code that uses them less readable. I can best demonstrate my point on an example, perhaps one that I was currently working on. Which one do you think is more readable?

1. The original:

```haskell
instance (FromCBOR a, FromCBOR b, FromCBOR c) => FromCBOR (a, b, c) where
  fromCBOR =
    decodeListLenOf 3
    !x <- fromCBOR
    !y <- fromCBOR
    !z <- fromCBOR
    return (x, y, z)
```

2. or this obscure variant?
```haskell
instance (FromCBOR a, FromCBOR b, FromCBOR c) => FromCBOR (a, b, c) where
  fromCBOR = do decodeListLenOf 3; !x <- fromCBOR; !y <- fromCBOR; !z <- fromCBOR; return (x, y, z)
```
3. or even better
```haskell
instance (FromCBOR a, FromCBOR b, FromCBOR c) => FromCBOR (a, b, c) where fromCBOR = do decodeListLenOf 3; !x <- fromCBOR; !y <- fromCBOR; !z <- fromCBOR; return (x, y, z)
```

4. Another extreme, which is suggested in this PR, is to use more lines. Problem with this approach is that it does not utilize available screen space efficiently, thus making the whole module less readable, especially when it is all written in such style:
```haskell
instance
  ( FromCBOR a,
    FromCBOR b,
    FromCBOR c
  ) =>
  FromCBOR (a, b, c)
  where
  fromCBOR = do
    decodeListLenOf 3
    !x <- fromCBOR
    !y <- fromCBOR
    !z <- fromCBOR
    return (x, y, z)
```

5. In my opinion the optimal would be to use `Applicative` syntax, thus making it less verbose, more readable and using less space:

```haskell
instance (FromCBOR a, FromCBOR b, FromCBOR c) => FromCBOR (a, b, c) where
  fromCBOR = do
    decodeListLenOf 3
    (,,) <$> fromCBOR <*> fromCBOR <*> fromCBOR
```

6. Below is the best variant, in my opinion, which is quite readable for anyone who is used to Applicative style parsers, but that is obviously quite subjective:
```haskell
instance (FromVCBOR a, FromVCBOR b, FromVCBOR c) => FromVCBOR (a, b, c) where
  fromVCBOR = decodeListLenOf 3 *> ((,,) <$> fromVCBOR <*> fromVCBOR <*> fromVCBOR)
```

Although, IMHO, the original is good enough.

Point I am trying to make is that I was able to put on two lines and keep it readable the same functionality, which would go onto 13 lines if it was formatted in a multi-line hungry fashion.

So the guideline is: the more code you can see at once while keeping it readable for most readers and following the Haskell conventions and not upsetting ormolu, the better.

@JaredCorduan if you agree with above, should I put this into `CONTRIBUTING.md`  doc or start a new code styling guideline doc? I can certainly place more similar stuff in such document over time.

eras/shelley/test-suite/src/Test/Cardano/Ledger/Shelley/Examples/Consensus.hs
lehins: Another good question. `Metadatum` is one small piece of `ShelleyTxAuxData`. Maybe we could rename it too, but I suggest we do it in a separate PR, since it is only an implementation detail of the actual type that needed renaming.

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Tx.hs
JaredCorduan: I agree :100: with you, @lehins ,  on all of the above. The only thing I would add is: when in doubt, favor vertical space since it plays better with source control. I think this is great material for the `CONTRIBUTING.md` doc.

@teodanciu - ormolu gives the programmer the option between "all on one line" or "on separate lines in the style ormolu chooses".

eras/shelley/test-suite/test/Test/Cardano/Ledger/Shelley/Serialisation/Tripping/CBOR.hs
JaredCorduan: I think we should change it too :)

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Rules/Utxow.hs
TimSheard: done

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Tx.hs
TimSheard: I m pretty sure this was done by ormolou. But I changed it back. We will see what happens.

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Tx.hs
TimSheard: done (blame ormolou)

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Tx.hs
TimSheard: done

libs/cardano-ledger-pretty/src/Cardano/Ledger/Pretty/Alonzo.hs
TimSheard: done

eras/shelley/test-suite/src/Test/Cardano/Ledger/Shelley/Examples/Consensus.hs
TimSheard: Because it is used to make a TxAuxData value?

eras/shelley/test-suite/test/Test/Cardano/Ledger/Shelley/Serialisation/Tripping/CBOR.hs
TimSheard: No you are right. I changed it.

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Tx.hs
TimSheard: done

eras/shelley-ma/impl/src/Cardano/Ledger/ShelleyMA/AuxiliaryData.hs
TimSheard: done

eras/shelley/test-suite/src/Test/Cardano/Ledger/Shelley/Examples/Consensus.hs
teodanciu: Makes sense!  My check was sort of mechanical, but of course there's no point to have the type of the variable encoded in the name  :sweat_smile:    :+1:

lehins (approved):
Thank you!

PR #3036, opened 2022-09-15
Added Core.hs to each era that has a TxBody class

### LINKED ISSUE #2932: Classes defined in each era should go into `Core` module
It was discussed that classes are spread around through out the repo and it would be best if each era defined type classes in the `Cardano.Ledger.[Era].Core` module. Unless of course the type class is used as an internal implementation in some module.

For example newly defined type classes in #2901, eg `AlonzoEraTxBody` is defined in `Cardano.Ledger.Alonzo.TxBody` module, but it needs to move into `Cardano.Ledger.Alonzo.Core`, while instances should still live in `Cardano.Ledger.Alonzo.TxBody`.

For convenience I also suggest we re-export the whole of `Cardno.Ledger.Core` module and `Cardano.Ledger.[Era].Core` from a previous era from it as well. Eg.

```haskell
module Cardano.Ledger.Shelley.Core
  ( module Cardano.Ledger.Core
  , ShelleyTxBody(..)
  ...
  ) where
```

```haskell
module Cardano.Ledger.ShelleyMA.Core
  ( module Cardano.Ledger.Shelley.Core
  , ShelleyMATxBody(..)
  ...
  ) where
```

Soupstraw:
This PR moves the `[Era]EraTxBody` classes from each era to the `Core` module of that era.

closes #2932

### DIFF
--- a/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Core.hs
+++ b/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Core.hs
@@ -1,0 +1,30 @@
+{-# LANGUAGE DataKinds #-}
+
+module Cardano.Ledger.Alonzo.Core
+  ( AlonzoEraTxBody (..),
+    ScriptIntegrityHash,
+    module Cardano.Ledger.ShelleyMA.Core,
+  )
+where
+
+import Cardano.Ledger.Alonzo.TxOut (AlonzoEraTxOut)
+import Cardano.Ledger.BaseTypes (Network)
+import Cardano.Ledger.Keys (KeyHash (..), KeyRole (..))
+import Cardano.Ledger.SafeHash (SafeHash)
+import Cardano.Ledger.ShelleyMA.Core
+import Cardano.Ledger.TxIn (TxIn (..))
+import Data.Maybe.Strict (StrictMaybe)
+import Data.Set (Set)
+import Lens.Micro (Lens')
+
+type ScriptIntegrityHash c = SafeHash c EraIndependentScriptIntegrity
+
+class (ShelleyMAEraTxBody era, AlonzoEraTxOut era) => AlonzoEraTxBody era where
+  collateralInputsTxBodyL :: Lens' (TxBody era) (Set (TxIn (EraCrypto era)))
+
+  reqSignerHashesTxBodyL :: Lens' (TxBody era) (Set (KeyHash 'Witness (EraCrypto era)))
+
+  scriptIntegrityHashTxBodyL ::
+    Lens' (TxBody era) (StrictMaybe (ScriptIntegrityHash (EraCrypto era)))
+
+  networkIdTxBodyL :: Lens' (TxBody era) (StrictMaybe Network)
--- a/eras/babbage/impl/src/Cardano/Ledger/Babbage/Core.hs
+++ b/eras/babbage/impl/src/Cardano/Ledger/Babbage/Core.hs
@@ -1,0 +1,28 @@
+module Cardano.Ledger.Babbage.Core
+  ( BabbageEraTxBody (..),
+    module Cardano.Ledger.Alonzo.Core,
+  )
+where
+
+import Cardano.Ledger.Alonzo.Core
+import Cardano.Ledger.Babbage.TxOut (BabbageEraTxOut)
+import Cardano.Ledger.Coin (Coin (..))
+import Cardano.Ledger.Serialization (Sized (..))
+import Cardano.Ledger.TxIn (TxIn (..))
+import Data.Maybe.Strict (StrictMaybe)
+import Data.Sequence.Strict (StrictSeq)
+import Data.Set (Set)
+import Lens.Micro (Lens', SimpleGetter)
+
+class (AlonzoEraTxBody era, BabbageEraTxOut era) => BabbageEraTxBody era where
+  sizedOutputsTxBodyL :: Lens' (TxBody era) (StrictSeq (Sized (TxOut era)))
+
+  referenceInputsTxBodyL :: Lens' (TxBody era) (Set (TxIn (EraCrypto era)))
+
+  totalCollateralTxBodyL :: Lens' (TxBody era) (StrictMaybe Coin)
+
+  collateralReturnTxBodyL :: Lens' (TxBody era) (StrictMaybe (TxOut era))
+
+  sizedCollateralReturnTxBodyL :: Lens' (TxBody era) (StrictMaybe (Sized (TxOut era)))
+
+  allSizedOutputsTxBodyF :: SimpleGetter (TxBody era) (StrictSeq (Sized (TxOut era)))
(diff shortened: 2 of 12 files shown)


eras/conway/impl/src/Cardano/Ledger/Conway/Core.hs
```diff
@@ -0,0 +1,4 @@
+module Cardano.Ledger.Conway.Core
```
teodanciu: Not sure about this, does it bring any value at this point, or better to add it when needed? I don't necessarily mind, just wondering

teodanciu (commented):
Looks good to me. 
I noticed we didn't do the suggestion from the ticket to 
> re-export the whole of Cardno.Ledger.Core module and Cardano.Ledger.[Era].Core from a previous era from it as well

- by choice or an omission?

eras/conway/impl/src/Cardano/Ledger/Conway/Core.hs
Soupstraw: I actually didn't mean to push this file

Soupstraw:
> Looks good to me. I noticed we didn't do the suggestion from the ticket to
> 
> > re-export the whole of Cardno.Ledger.Core module and Cardano.Ledger.[Era].Core from a previous era from it as well
> 
>     * by choice or an omission?

Fixed it now

eras/shelley/impl/src/Cardano/Ledger/Shelley/Core.hs
```diff
@@ -0,0 +1,55 @@
...
+  ( ShelleyEraTxBody (..),
+    Wdrl (..),
+    module Cardano.Ledger.Core,
+  )
+where
+
+import Cardano.Binary (FromCBOR (..), ToCBOR (..))
+import Cardano.Ledger.Address (RewardAcnt (..))
+import Cardano.Ledger.Coin (Coin (..))
+import Cardano.Ledger.Core (Era (..), EraTxBody (..), ExactEra)
```
lehins: The idea was to re-export the whole of `Core`, so in all other modules we can import `import Cardano.Ledger.Shelley.Core` and not bother with `import Cardano.Ledger.Core`. For this we need to import that module as a whole
```suggestion
import Cardano.Ledger.Core
```

eras/shelley-ma/impl/src/Cardano/Ledger/ShelleyMA/Core.hs
```diff
@@ -0,0 +1,24 @@
...
+
+module Cardano.Ledger.ShelleyMA.Core
+  ( ShelleyMAEraTxBody (..),
+    module Cardano.Ledger.Shelley.Core,
+  )
+where
+
+import Cardano.Ledger.Core (Era (..), EraTxBody (..), Value)
+import Cardano.Ledger.Mary.Value (MultiAsset (..))
+import Cardano.Ledger.Shelley.Core (ShelleyEraTxBody)
```
lehins: So, here we no longer need to import `Cardano.Ledger.Core`, but we re-export the whole of `Cardano.Ledger.Shelley.Core`:
```suggestion
```
```suggestion
import Cardano.Ledger.Mary.Value (MultiAsset (..))
import Cardano.Ledger.Shelley.Core
```

eras/babbage/impl/src/Cardano/Ledger/Babbage/Core.hs
```diff
@@ -0,0 +1,29 @@
+module Cardano.Ledger.Babbage.Core
+  ( BabbageEraTxBody (..),
+    module Cardano.Ledger.Alonzo.Core,
+  )
+where
+
+import qualified Cardano.Ledger.Alonzo.Core as X
+import Cardano.Ledger.Babbage.TxOut (BabbageEraTxOut)
+import Cardano.Ledger.Coin (Coin (..))
+import Cardano.Ledger.Core (Era (..), EraTxBody (..), EraTxOut (..))
```
lehins: Same here
```suggestion
import Cardano.Ledger.Alonzo.Core
import Cardano.Ledger.Babbage.TxOut (BabbageEraTxOut)
import Cardano.Ledger.Coin (Coin (..))
```

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Core.hs
```diff
@@ -0,0 +1,31 @@
...
+    module Cardano.Ledger.ShelleyMA.Core,
+  )
+where
+
+import Cardano.Ledger.Alonzo.TxOut (AlonzoEraTxOut)
+import Cardano.Ledger.BaseTypes (Network)
+import Cardano.Ledger.Core (Era (..), EraIndependentScriptIntegrity, EraTxBody (..))
+import Cardano.Ledger.Keys (KeyHash (..), KeyRole (..))
+import Cardano.Ledger.SafeHash (SafeHash)
+import Cardano.Ledger.ShelleyMA.Core (ShelleyMAEraTxBody)
```
lehins: Same here
```suggestion
import Cardano.Ledger.Keys (KeyHash (..), KeyRole (..))
import Cardano.Ledger.SafeHash (SafeHash)
import Cardano.Ledger.ShelleyMA.Core
```

lehins (changes requested):
Just one conceptual change need, otherwise looks great.

eras/shelley/impl/src/Cardano/Ledger/Shelley/Core.hs
```diff
@@ -0,0 +1,55 @@
...
+{-# LANGUAGE DeriveGeneric #-}
+{-# LANGUAGE DerivingVia #-}
+{-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE FlexibleInstances #-}
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE MultiParamTypeClasses #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeFamilies #-}
+{-# LANGUAGE UndecidableInstances #-}
+{-# OPTIONS_GHC -Wno-orphans #-}
```
lehins: No orphans here
```suggestion
```

eras/shelley/impl/src/Cardano/Ledger/Shelley/Core.hs
```diff
@@ -0,0 +1,55 @@
+{-# LANGUAGE DataKinds #-}
+{-# LANGUAGE DeriveGeneric #-}
+{-# LANGUAGE DerivingVia #-}
+{-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE FlexibleInstances #-}
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE MultiParamTypeClasses #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeFamilies #-}
+{-# LANGUAGE UndecidableInstances #-}
```
lehins: Unused extensions
```suggestion
```

eras/shelley/impl/src/Cardano/Ledger/Shelley/Core.hs
```diff
@@ -0,0 +1,55 @@
+{-# LANGUAGE DataKinds #-}
```
lehins: Unused extension
```suggestion
```

lehins (approved):
A few redundant things, but otherwise looks great

Issue #2839, opened 2022-06-02
Alonzo test failure: Missing script witnesses


Soupstraw:
On commit dd13fada:
```
Build profile: -w ghc-8.10.7 -O1
In order, the following will be built (use -v for more details):
 - cardano-ledger-alonzo-test-0.1.0.0 (test:cardano-ledger-alonzo-test) (first run)
Preprocessing test suite 'cardano-ledger-alonzo-test' for cardano-ledger-alonzo-test-0.1.0.0..
Building test suite 'cardano-ledger-alonzo-test' for cardano-ledger-alonzo-test-0.1.0.0..
Running 1 test suites...
Test suite cardano-ledger-alonzo-test: RUNNING...

Running in scenario: Development
Alonzo tests
  Fast Alonzo Property Tests
    total amount of Ada is preserved (Chain): 
Discarded trace: TotExUnits violation: genEraTweakBlock: AlonzoEraGen.hs

Discarded trace: MinFeee violation: genEraDne: AlonzoEraGen.hs

Discarded trace: No inputs left. Utxo.hs

Discarded trace: NoMoneyleft Utxo.hs
FAIL (74.43s)
      *** Failed! (after 36 tests):
      Exception while generating shrink-list:
        LEDGER sigGen: [UtxowFailure (WrappedShelleyEraFailure (MissingScriptWitnessesUTXOW (fromList [])))]
        CallStack (from HasCallStack):
          error, called at src/Test/Cardano/Ledger/Shelley/Generator/Trace/Ledger.hs:179:24 in cardano-ledger-shelley-test-0.1.0.0-inplace:Test.Cardano.Ledger.Shelley.Generator.Trace.Ledger
          genAndApplyTx, called at src/Test/Cardano/Ledger/Shelley/Generator/Trace/Ledger.hs:159:11 in cardano-ledger-shelley-test-0.1.0.0-inplace:Test.Cardano.Ledger.Shelley.Generator.Trace.Ledger
          sigGen, called at src/Test/Cardano/Ledger/Shelley/Generator/Block.hs:101:16 in cardano-ledger-shelley-test-0.1.0.0-inplace:Test.Cardano.Ledger.Shelley.Generator.Block
      Use --quickcheck-replay=970901 to reproduce.

1 out of 1 tests failed (74.46s)
Test suite cardano-ledger-alonzo-test: FAIL
Test suite logged to:
[...]

PR #3039, opened 2022-09-15
Fix random failure: MissingScriptWitnessesUTXOW (fromList [])

TimSheard:
For a long time we get random, in frequent failures in the property tests labeled with
MissingScriptWitnessesUTXOW (fromList [])
This is annoying, and actually very misleading because the actual failure is that there is an extra script witness that is not actually needed. This misleading error message is caused by a long known bug that was improved by code like this
```haskell
validateMissingScripts pp sNeeded sReceived =
  if HardForks.missingScriptsSymmetricDifference pp
    then
      sequenceA_
        [ failureUnless (sNeeded `Set.isSubsetOf` sReceived) $
            MissingScriptWitnessesUTXOW (sNeeded `Set.difference` sReceived),
          failureUnless (sReceived `Set.isSubsetOf` sNeeded) $
            ExtraneousScriptWitnessesUTXOW (sReceived `Set.difference` sNeeded)
        ]
    else
      failureUnless (sNeeded == sReceived) $
        MissingScriptWitnessesUTXOW (sNeeded `Set.difference` sReceived) 
```
Note when (HardForks.missingScriptsSymmetricDifference pp) is False, and sRecieved is a supersetof sNeeded, we get the misleading error.

So the real problem is that the EraGen process of generating random traces, occasionally (and randomly) generates an extra script witness. It is hard to track this down.   There is some evidence that this arises because the script hash appears as the PolicyId in a Value. But that is not been verified. Regardless of the cause, we fix this by
discarding the trace. Note that this is failure to generate a "random" but valid transaction. Discarding the trace adjust for this inadequacy in the generation process. This only appears in the Alonzo era, so this "fix" is applied here, in the genEraDone 
method of the EraGen class in the (AlonzoEra c) instance.

### DIFF
--- a/eras/alonzo/test-suite/src/Test/Cardano/Ledger/Alonzo/AlonzoEraGen.hs
+++ b/eras/alonzo/test-suite/src/Test/Cardano/Ledger/Alonzo/AlonzoEraGen.hs
@@ -34,15 +34,7 @@ import Cardano.Ledger.Alonzo.Scripts as Alonzo
     Prices (..),
     mkCostModel,
   )
-import Cardano.Ledger.Alonzo.Tx
-  ( AlonzoEraTx (..),
-    AlonzoTx (..),
-    IsValid (..),
-    ScriptPurpose (..),
-    hashScriptIntegrity,
-    rdptr,
-    totExUnits,
-  )
+import Cardano.Ledger.Alonzo.Tx (AlonzoEraTx (..), AlonzoTx (..), IsValid (..), ScriptPurpose (..), hashScriptIntegrity, rdptr, totExUnits)
 import Cardano.Ledger.Alonzo.TxBody
   ( AlonzoEraTxBody (..),
     AlonzoTxBody (..),
@@ -76,7 +68,13 @@ import Cardano.Ledger.Mary.Value
 import Cardano.Ledger.Pretty.Alonzo ()
 import Cardano.Ledger.Shelley.PParams (Update)
 import Cardano.Ledger.Shelley.TxBody (DCert, Wdrl)
-import Cardano.Ledger.Shelley.UTxO (EraUTxO (..), UTxO (..), coinBalance)
+import Cardano.Ledger.Shelley.UTxO
+  ( EraUTxO (..),
+    UTxO (..),
+    coinBalance,
+    getScriptsHashesNeeded,
+    getScriptsNeeded,
+  )
 import Cardano.Ledger.ShelleyMA.AuxiliaryData (AllegraTxAuxData (..))
 import Cardano.Ledger.ShelleyMA.Era ()
 import Cardano.Ledger.ShelleyMA.Timelocks (Timelock (..), translateTimelock)
@@ -439,12 +437,25 @@ instance Mock c => EraGen (AlonzoEra c) where
         Nothing -> storageCost 0 pp script
       else storageCost 0 pp script
 
-  genEraDone pp tx =
+  -- For some reason, the EraGen generators occasionally generate an extra script witness.
+  --    There is some evidence that this arises because the script hash appears as the PolicyId
+  --    in a Value. But that is not been verified. Regardless of the cause, we can fix this by
+  --    discarding the trace. Note that this is failure to generate a "random" but valid
+  --    transaction. Discarding the trace adjust for this inadequacy in the generation process.
+  --    This only appears in the Alonzo era, so this "fix" is applied here, in the genEraDone
+  --    method of the EraGen class in the (AlonzoEra c) instance.
+  genEraDone utxo pp tx =
     let theFee = tx ^. bodyTxL . feeTxBodyL -- Coin supplied to pay fees
         minimumFee = getMinFeeTx @(AlonzoEra c) pp tx
+        neededHashes = getScriptsHashesNeeded (getScriptsNeeded utxo (tx ^. bodyTxL))
+        oldScriptWits = tx ^. witsTxL . scriptTxWitsL
+        newWits = oldScriptWits `Map.restrictKeys` neededHashes
      in if minimumFee <= theFee
-          then pure tx
-          else myDiscard "MinFeee violation: genEraDne: AlonzoEraGen.hs"
+          then
+            if oldScriptWits == newWits
+              then pure tx
+              else myDiscard "Random extra scriptwitness: genEraDone: AlonzoEraGen.hs"
+          else myDiscard "MinFeee violation: genEraDone: AlonzoEraGen.hs"
 
   genEraTweakBlock pp txns =
     let txTotal, ppMax :: ExUnits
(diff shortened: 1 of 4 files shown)


JaredCorduan:
I believe this PR will resolve #2839 and #2811 :tada:

eras/alonzo/test-suite/src/Test/Cardano/Ledger/Alonzo/AlonzoEraGen.hs
```diff
@@ -35,15 +35,7 @@ import Cardano.Ledger.Alonzo.Scripts as Alonzo
...
-import Cardano.Ledger.Alonzo.Tx
-  ( AlonzoEraTx (..),
-    AlonzoTx (..),
-    IsValid (..),
-    ScriptPurpose (..),
-    hashScriptIntegrity,
-    rdptr,
-    totExUnits,
-  )
+import Cardano.Ledger.Alonzo.Tx (AlonzoEraTx (..), AlonzoTx (..), IsValid (..), ScriptPurpose (..), hashScriptIntegrity, rdptr, totExUnits, witsAlonzoTxL)
```
JaredCorduan: did you meant to collapse these?

eras/alonzo/test-suite/src/Test/Cardano/Ledger/Alonzo/AlonzoEraGen.hs
```diff
@@ -433,12 +426,26 @@ instance Mock c => EraGen (AlonzoEra c) where
...
       else storageCost 0 pp script
 
-  genEraDone pp tx =
+  -- For some reason, the EraGen generators occasionally generate an extra script witness.
+  --    There is some evidence that this arises because the script hash appears as the PolicyId
+  --    in a Value. But that is not been verified. Regardless of the cause, we can fix this by
+  --    discarding the trace. Note that this is failure to generate a "random" but valid
+  --    transaction. Discarding the trace adjust for this inadequacy in the generation process.
+  --    This only appears in the Alonzo era, so this "fix" is applied here, in the genEraDone
+  --    method of the EraGen class in the (AlonzoEra c) instance.
```
JaredCorduan, line 446: maybe we could leave a `FIXME` or `TODO` to address the root cause? I'm happy having a discoverable discard over constant flaky tests breaking CI, but if/when we get into discard madness it's good to be able to find them... (though maybe searching for `myDiscard` will lead us to them).

JaredCorduan (approved):

eras/alonzo/test-suite/src/Test/Cardano/Ledger/Alonzo/AlonzoEraGen.hs
```diff
@@ -433,12 +426,26 @@ instance Mock c => EraGen (AlonzoEra c) where
...
+  genEraDone utxo pp tx =
     let theFee = tx ^. bodyTxL . feeTxBodyL -- Coin supplied to pay fees
         minimumFee = getMinFeeTx @(AlonzoEra c) pp tx
+        neededHashes = snd <$> (scriptsNeeded utxo tx)
+        oldScriptWits = (tx ^. witsAlonzoTxL . scriptAlonzoTxWitsL)
+        newWits = Map.filterWithKey (\k _ -> List.elem k neededHashes) oldScriptWits
```
lehins: You need to rebase on `master`. Latest PR #3019 deprecated `scriptsNeeded`. Also lenses for concrete types are available only to define era agnostic lenses and should be avoided otherwise. This will become:
```suggestion
        neededHashes = getScriptsHashesNeeded (getScriptsNeeded utxo (tx ^. bodyTxL))
        oldScriptWits = tx ^. witsxL . scriptTxWitsL
        newWits = oldScriptWits `Map.restrictKeys`  neededHashes
```

eras/alonzo/test-suite/src/Test/Cardano/Ledger/Alonzo/AlonzoEraGen.hs
```diff
@@ -433,12 +426,26 @@ instance Mock c => EraGen (AlonzoEra c) where
...
      in if minimumFee <= theFee
-          then pure tx
-          else myDiscard "MinFeee violation: genEraDne: AlonzoEraGen.hs"
+          then
+            ( if not (oldScriptWits == newWits)
+                then myDiscard "Random extra scriptwitness: genEraDone: AlonzoEraGen.hs"
+                else pure tx
+            )
```
lehins: Redundant parenthesis and unnecessary negation
```suggestion
          then
              if oldScriptWits /= newWits
                then myDiscard "Random extra scriptwitness: genEraDone: AlonzoEraGen.hs"
                else pure tx
```

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/Consensus.hs
```diff
@@ -60,7 +60,7 @@ import qualified Cardano.Ledger.Shelley.PParams as ShelleyPP
 import Cardano.Ledger.Shelley.Rules
 import Cardano.Ledger.Shelley.UTxO (makeWitnessesVKey)
 import Cardano.Ledger.ShelleyMA (ShelleyMAEra)
-import Cardano.Ledger.ShelleyMA.AuxiliaryData (MAAuxiliaryData (..))
+import Cardano.Ledger.ShelleyMA.AuxiliaryData (AllegraTxAuxData (..))
```
lehins: Once you rebase on master this should dissapear, since it has already been fixed on master

lehins (changes requested):
Thank you for finally fixing this annoyance

eras/alonzo/test-suite/src/Test/Cardano/Ledger/Alonzo/AlonzoEraGen.hs
TimSheard: I will make an issue, referencing this PR and some notes.

eras/alonzo/test-suite/src/Test/Cardano/Ledger/Alonzo/AlonzoEraGen.hs
TimSheard: done

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/Consensus.hs
TimSheard: that was true

lehins (approved):
:+1:

eras/alonzo/test-suite/src/Test/Cardano/Ledger/Alonzo/AlonzoEraGen.hs
TimSheard: done

eras/alonzo/test-suite/src/Test/Cardano/Ledger/Alonzo/AlonzoEraGen.hs
TimSheard: Here is the issue:  https://github.com/input-output-hk/cardano-ledger/issues/3040
I created

PR #3038, opened 2022-09-15
Node/ghc 9.2.4

newhoggy:
This PR is branched off f49879a79098d9372d63baa13b94a941a56eda34 which is the commit that `cardano-node` on master currently uses.

Please don't rebase on `origin/master`.

PR #3033, opened 2022-09-12
Fixes to compile on ghc-9.2.4

newhoggy:

PR #3041, opened 2022-09-16
9.2.4 compatibility

angerman:
VERY WIP

Depends on https://github.com/input-output-hk/cardano-ledger/pull/3098.

This is mostly here to get Hydra building the shell in the background for easier iteration. Then we need to actually push the changes to this branch to get things compiling with 9.2.

----

This moves the ledger to build primarily with GHC 9.2. I say "primarily", because there annoyingly remains one package that doesn't build with 9.2, namely `plutus-preprocessor`. This is because we haven't fixed up `plutus-tx-plugin` to work with 9.2 yet. 

The compromise position I've taken is to move the default build environment to 9.2, but leave instructions for how to get an 8.10 shell if you need to use `plutus-preprocessor`. This requires us to not break 8.10 support for a while, but hopefully not too long. And it still lets us progress towards a 9.2 node.

The alternative would be to just not try to keep 8.10 support at all, and not try to keep `plutus-preprocessor` building. Then we would resurrect it once we fix `plutus-tx-plugin`. I think that's worth considering, since it's not used that frequently and it's not that much code. This approach would allow you to start using 9.2 features everywhere (and give you a simpler CI), but at the cost of losing `plutus-preprocessor` for a bit. Up to you.


JaredCorduan:
@angerman @erikd can we close the following PRs in favor of this one?

* #3081 
* #3038
* #3033

cabal.project
```diff
@@ -94,3 +97,16 @@ source-repository-package
...
+  location: https://github.com/input-output-hk/plutus
+  tag: cec1eaca9ade859797acec66d9741c9aaac48b45
+  --sha256: 084yvxcx56i5nync597029v23idgvcxysvk7zghsiqfvv4qr911x
+  subdir:
+    plutus-ledger-api
+    plutus-tx
+    plutus-tx-plugin
+    plutus-core
+    prettyprinter-configurable
+    word-array
```
angerman: This one should be gone, after @michaelpj's commit, I think

eras/alonzo/impl/cardano-ledger-alonzo.cabal
```diff
@@ -80,9 +80,8 @@ library
     mtl,
     microlens,
     nothunks,
-    plutus-ledger-api ^>= 1.1,
-    plutus-tx ^>= 1.1,
-    plutus-core ^>= 1.1,
+    plutus-ledger-api,
+    plutus-core ^>= 1.0,
```
angerman: This one too. I think you'll be pulling in the wrong plutus-core now.

eras/alonzo/test-suite/cardano-ledger-alonzo-test.cabal
```diff
@@ -64,9 +64,8 @@ library
     containers,
     data-default-class,
     hashable >= 1.4.1.0,
-    plutus-core,
     plutus-tx,
-    plutus-ledger-api:{plutus-ledger-api, plutus-ledger-api-testlib} ^>=1.1,
+    plutus-ledger-api,
```
angerman: I'm not sure this is correct?

eras/alonzo/test-suite/cardano-ledger-alonzo-test.cabal
```diff
@@ -111,8 +110,7 @@ test-suite cardano-ledger-alonzo-test
     cborg,
     containers,
     data-default-class,
-    plutus-core,
-    plutus-ledger-api:{plutus-ledger-api, plutus-ledger-api-testlib} ^>=1.1,
+    plutus-ledger-api,
```
angerman: neither here

eras/babbage/impl/cardano-ledger-babbage.cabal
```diff
@@ -68,8 +68,7 @@ library
                 deepseq,
                 microlens,
                 nothunks,
-                plutus-ledger-api ^>= 1.1,
-                plutus-tx ^>= 1.1,
+                plutus-ledger-api,
```
angerman: This also seems to revert Michaels commit.

libs/cardano-ledger-test/cardano-ledger-test.cabal
```diff
@@ -93,7 +93,7 @@ library
     microlens,
     mtl,
     nothunks,
-    plutus-ledger-api:{plutus-ledger-api, plutus-ledger-api-testlib} ^>=1.1,
+    plutus-ledger-api,
```
angerman: This one seems to revert Michaels changes as well.

angerman:
I added back @michaelpj's original commentary.

eras/babbage/impl/cardano-ledger-babbage.cabal
```diff
@@ -69,7 +69,7 @@ library
                 microlens,
                 nothunks,
                 plutus-ledger-api ^>= 1.1,
-                plutus-tx ^>= 1.1,
+                -- plutus-tx ^>= 1.1,
```
michaelpj: commented-out

eras/byron/crypto/cardano-crypto-wrapper.cabal
```diff
@@ -134,7 +132,7 @@ test-suite test
                      , filepath
                      , hedgehog >= 1.0.4
                      , memory
-                     , text
+                     -- , text
```
michaelpj: commented-out

michaelpj (approved):
Looks good!

angerman:
Thanks @michaelpj!

angerman:
The next iteration on this repo, should cut the nix code aggressively for a stock flake if possible.

michaelpj:
Some test failures, looks like it's looking for test files in the wrong place.

michaelpj:
I pushed the haskell.nix bump that fixes the datadir problem.

michaelpj:
The cicero job is broken due to https://input-output.atlassian.net/browse/CIC-246, but Hydra is green. In fact we're about to release a 9.2-compatible plugin, so if we wait a short time then we'll be able to avoid the conditionals an everything.

michaelpj:
Went green!

michaelpj:
@JaredCorduan I think this is good to go now!

lehins (approved):
:+1: 

Github Actions CI is still failing due to constraint resolution, other than that it looks good

JaredCorduan (approved):
very exciting, thank you so much to everyone who helped with this!

michaelpj:
Ah indeed, we should mark `plutus-preprocessor` as not buildable on older GHC versions, so cabal doesn't think it should be.

michaelpj:
:tada:

PR #3052, opened 2022-09-26
applyRuleInternal: descend into subrule in the Label case upon ValidateNone

zliu41:
This is probably not the proper fix, but it does fix the problem I'm having.

As far as I can tell, `ValidateAll / ValidateNone / ValidateSuchThat` were originally used to decide whether to evaluate a `Predicate`. But the current behavior is: in the `Label` case, upon `ValidateNone`, it would stop descending into the subrule even if the subrule is not a `Predicate`.

This causes a problem for me: `SuccessfulPlutusScriptsEvent` and `FailedPlutusScriptsEvent` are never emitted, because the action that emits them [is hidden behind a Label added by when2Phase](https://github.com/input-output-hk/cardano-ledger/blob/b60ecae5a68230c60921fcc57e4929a1d235dfbe/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Rules/Utxos.hs#L204)!

The proper fix, if we do want to stop descending whenever we see a `Label`, is probably to create a separate type similar to `ValidationPolicy`, and leave the original `ValidationPolicy` for the `Predicate` case.

@JaredCorduan @lehins

### DIFF
--- a/libs/small-steps/src/Control/State/Transition/Extended.hs
+++ b/libs/small-steps/src/Control/State/Transition/Extended.hs
@@ -629,7 +629,7 @@ applyRuleInternal isAlreadyFailing ep vp goSTS jc r = do
       EPDiscard -> pure a
     validateIf lbls = case vp of
       ValidateAll -> True
-      ValidateNone -> False
+      ValidateNone -> True
       ValidateSuchThat f -> f lbls
 
 applySTSInternal ::


JaredCorduan:
thanks @zliu41 , we'll try to take a look as soon as we can and come up with a proper plan.

libs/small-steps/src/Control/State/Transition/Extended.hs
```diff
@@ -629,7 +629,7 @@ applyRuleInternal isAlreadyFailing ep vp goSTS jc r = do
       EPDiscard -> pure a
     validateIf lbls = case vp of
       ValidateAll -> True
-      ValidateNone -> False
+      ValidateNone -> True
```
lehins, line 632: `ValidateNone` disables validation irrespective of any labels, so this change is definitely wrong

lehins (changes requested):
@zliu41 Could you please describe exactly what you are trying to achieve with a concrete example of how STS rule is being executed. I struggle to understand why would you expect SuccessfulPlutusScriptsEvent and FailedPlutusScriptsEvent events to be fired when actual predicates aren't evaluated

zliu41:
@lehins I'm trying to collect Plutus evaluation events on mainnet, from LedgerEvents.

I'm using the [streaming library in plutus-apps](https://github.com/input-output-hk/plutus-apps/tree/e882efc2aabea0a3e80186cbed2d1025b9714af3/cardano-streaming). It is similar to `foldBlocks`, and I think `foldBlocks` should have the same problem, which is:  the streamer [sees all other LedgerEvents](https://github.com/zliu41/cardano-node/blob/ledgerevent/script/cardano-api/src/Cardano/Api/LedgerEvent.hs#L71-L82), but never sees `SuccessfulPlutusScript` or `FailedPlutusScript`, unless I change that Boolean flag.

If you are interested in seeing the code I'm running, it's [here](https://github.com/input-output-hk/plutus-apps/blob/fc79df7ee697a3cee572b56abc21add5d2e4b8ce/plutus-script-evaluation-test/testlib/Plutus/Script/Evaluation/Dump.hs#L207-L210).

zliu41:
And this has been a problem ever since [this line](https://github.com/input-output-hk/cardano-ledger/blob/b60ecae5a68230c60921fcc57e4929a1d235dfbe/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Rules/Utxos.hs#L204) was modified to add the `when2Phase`. It was working before.

lehins:
@zliu41 If I had to guess you are replaying the chain without validation, which means that scripts do not get evaluated and therefore script events do not get triggered either, which is expected.

I suspect what you need is events being produced regardless of validation flag. We need to see what can we do about this.

JaredCorduan:
> I suspect what you need is events being produced regardless of validation flag. We need to see what can we do about this.

indeed, that's exactly the problem. but it was not clear to me how to do it (and we certainly don't want to set
`ValidateNone` to true).

JaredCorduan:
I've made issue #3082 to track this.

PR #3053, opened 2022-09-27
Add context for translating ShelleyLedgerConfig

### LINKED ISSUE #409: Remove ShelleyGenesis from ShelleyLedgerConfig
The difference between Babbage and Conway was intentionally as minimal as possible. This has a couple benefits.

- Clarke had orchestrated the integration of new eras. So it's nice that the first one we did without him is so minimal.

- The minimality helped us (especially in Clarke's absence) notice the aspects that we found confusing.

This Issue is about one such aspect: I don't think `ShelleyGenesis` (as `CompactGenesis`) should be in the `ShelleyLedgerConfig`. In other words, I'm proposing removing the following function.

```
Ouroboros.Consensus.Shelley.Ledger.Ledger.shelleyLedgerGenesis :: ShelleyLedgerConfig era -> SL.ShelleyGenesis era
```

I've catalogued its uses as follows:

- It's used in `Ouroboros.Consensus.Cardano.CanHardFork` for translating/forecasting from Byron to the first Shelley era. This is a necessary use.
- It's used in `Ouroboros.Consensus.Shelley.Ledger.Inspect`, `Ouroboros.Consensus.Shelley.Ledger.Ledger`, and `Ouroboros.Consensus.Shelley.ShelleyHFC` to get the value of `k`, `f` (only used to compute the stability window), `quorum`, `slotLength`, and `epochLength`. I think all of these could and should instead use the (already-existing!) `shelleyLedgerGlobals :: SL.Globals` field.

The second bullet point is for enabling the hard fork combinator, which every era needs to do. The first bullet point is instead for translating from Byron to Shelley, which only the first Shelley-based era needs to do. Moreover, I don't think we want to require that _any_ Shelley-based era could be the successor Byron, do we? So I propose:

```
data ShelleyLedgerConfig era = ShelleyLedgerConfig {
      shelleyLedgerGlobals                     :: !SL.Globals
    , shelleyLedgerTranslationContext          :: !(Core.TranslationContext era)
    , shelleyLedgerTranslationContextFromByron :: !(StaticMaybe (SL.ByronTranslationContext era))
    }

data StaticMaybe :: Maybe Type -> Type where
  StaticJust :: !a -> StaticMaybe (Just a)
  StaticNothing :: StaticMaybe Nothing
```

(see https://github.com/input-output-hk/cardano-ledger/pull/3053#discussion_r1015751613 for motivating the `StaticMaybe` layer)

where in the ledger we have:

```
type family ByronTranslationContext era

type instance ByronTranslationContext ShelleyEra = SL.ParedDownShelleyGenesis
type instance ByronTranslationContext AllegraEra = Void -- or TypeError?         (I originally wrote () here but then I objected to it on one of Bart's PRs :face_palm)
type instance ByronTranslationContext MaryEra    = Void -- or TypeError?
type instance ByronTranslationContext AlonzoEra  = Void -- or TypeError?
type instance ByronTranslationContext BabbageEra = Void -- or TypeError?
type instance ByronTranslationContext ConwayEra  = Void -- or TypeError?
```

As a result of this, we should be able to remove all of the oxymoronic `TranslateEra ShelleyGenesis` instances. We'll have to replace them with `TranslateEra Globals` instances (or some translate-able data that determines the `Globals`). For example, these new translation instances won't even have a "genesis delegates" field, and so it'll avoid the existing confusing fact that the `TranslateEra ShelleyGenesis ConwayEra` instance does _not_ need to update the `sgGenDelegs` field!

bartfrenk:
Work-in-progress: Still needs to be branched off from head of master.

These changes branch of from the commit labeled ledger/1.1.0, which, at the time of writing, is the version that `master` of ouroboros-network depends on.

The aim is to close https://github.com/input-output-hk/ouroboros-consensus/issues/409, which is done by https://github.com/input-output-hk/ouroboros-network/pull/4091. That PR in ouroboros-network depends on changes in this PR.

### DIFF
--- a/eras/shelley/impl/src/Cardano/Ledger/Shelley/Era.hs
+++ b/eras/shelley/impl/src/Cardano/Ledger/Shelley/Era.hs
@@ -4,7 +4,7 @@
 module Cardano.Ledger.Shelley.Era (ShelleyEra) where
 
 import Cardano.Ledger.Coin (Coin)
-import Cardano.Ledger.Core (Era (..), TranslationContext, Value)
+import Cardano.Ledger.Core (Era (..), Value)
 import Cardano.Ledger.Crypto as CC (Crypto)
 
 data ShelleyEra crypto
@@ -14,5 +14,3 @@ instance CC.Crypto c => Era (ShelleyEra c) where
   type ProtVerLow (ShelleyEra c) = 2
 
 type instance Value (ShelleyEra _c) = Coin
-
-type instance TranslationContext (ShelleyEra c) = ()
--- a/eras/shelley/impl/src/Cardano/Ledger/Shelley/Genesis.hs
+++ b/eras/shelley/impl/src/Cardano/Ledger/Shelley/Genesis.hs
@@ -24,6 +24,7 @@ module Cardano.Ledger.Shelley.Genesis
     validateGenesis,
     describeValidationErr,
     mkShelleyGlobals,
+    translateShelleyGenesis
   )
 where
 
@@ -142,6 +143,54 @@ data ShelleyGenesis era = ShelleyGenesis
   deriving stock (Eq, Show, Generic)
 
 deriving instance Era era => NoThunks (ShelleyGenesis era)
+
+-- | Although there could be (and was) a `TranslateEra` instance for ShelleyGenesis, it is morally
+-- more correct to have a separate function fulfill that task. The reason is that conceptually there
+-- is no need to translate ShelleyGenesis values when moving to the next era. A `TranslateEra`
+-- instance would be a convenience that is ultimately more confusing.
+translateShelleyGenesis :: Crypto era1 ~ Crypto era2 => ShelleyGenesis era1 -> ShelleyGenesis era2
+translateShelleyGenesis genesis =
+      ShelleyGenesis
+        { sgSystemStart = sgSystemStart genesis,
+          sgNetworkMagic = sgNetworkMagic genesis,
+          sgNetworkId = sgNetworkId genesis,
+          sgActiveSlotsCoeff = sgActiveSlotsCoeff genesis,
+          sgSecurityParam = sgSecurityParam genesis,
+          sgEpochLength = sgEpochLength genesis,
+          sgSlotsPerKESPeriod = sgSlotsPerKESPeriod genesis,
+          sgMaxKESEvolutions = sgMaxKESEvolutions genesis,
+          sgSlotLength = sgSlotLength genesis,
+          sgUpdateQuorum = sgUpdateQuorum genesis,
+          sgMaxLovelaceSupply = sgMaxLovelaceSupply genesis,
+          sgProtocolParams = translatePParams $ sgProtocolParams genesis,
+          sgGenDelegs = sgGenDelegs genesis,
+          sgInitialFunds = sgInitialFunds genesis,
+          sgStaking = sgStaking genesis
+        }
+  where
+    translatePParams :: ShelleyPParams era1 -> ShelleyPParams era2
+    translatePParams pp = ShelleyPParams
+        { _minfeeA = _minfeeA pp,
+          _minfeeB = _minfeeB pp,
+          _maxBBSize = _maxBBSize pp,
+          _maxTxSize = _maxTxSize pp,
+          _maxBHSize = _maxBHSize pp,
+          _keyDeposit = _keyDeposit pp,
+          _poolDeposit = _poolDeposit pp,
+          _eMax = _eMax pp,
+          _nOpt = _nOpt pp,
+          _a0 = _a0 pp,
+          _rho = _rho pp,
+          _tau = _tau pp,
+          _d = _d pp,
+          _extraEntropy = _extraEntropy pp,
+          _protocolVersion = _protocolVersion pp,
+          _minUTxOValue = _minUTxOValue pp,
+          _minPoolCost = _minPoolCost pp
+        }
+
+
+
 
 sgActiveSlotCoeff :: ShelleyGenesis era -> ActiveSlotCoeff
 sgActiveSlotCoeff = mkActiveSlotCoeff . sgActiveSlotsCoeff
--- a/.github/workflows/haskell.yml
+++ b/.github/workflows/haskell.yml
@@ -4,7 +4,7 @@ on:
   push:
     branches: [ '*' ]
   pull_request:
-    branches: [ master ]
+    branches: [ "master", "release/*" ]
 
 jobs:
   build:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ in the naming of release branches.
 
 ## [Unreleased]
 
+## Release branch 1.1.x
+
 ### Added
 - New `calculatePoolDistr'` function which is similar to `calculatePoolDistr` but has a new
   filter argument to only include the stake pool ids (stake pool key hashes) that are needed.
@@ -208,7 +210,7 @@ BabbageEraTxBody --> AlonzoEraTxBody --> ....
   It now uses the Alonze minfee function.
   #2936
 
-## Release branch 1.0.0
+## Release branch 1.0.x
 
 The first release branch in the cardano-ledger repository,
 namely `release/1.0.0`, branches from the
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -4,9 +4,13 @@
 
 We use [trunk based developement](https://trunkbaseddevelopment.com/).
 In particular, releases will be handled by release branches,
-starting with `release/1.0.0`.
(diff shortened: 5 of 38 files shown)


eras/shelley/impl/src/Cardano/Ledger/Shelley/Genesis.hs
```diff
@@ -116,6 +119,16 @@ emptyGenesisStaking =
       sgsStake = mempty
     }
 
+data ParedDownShelleyGenesis era = ParedDownShelleyGenesis
+  { pdsgSlotLength     :: !NominalDiffTime
+  , pdsgEpochLength    :: !EpochSize
+  , pdsgProtocolParams :: !(ShelleyPParams era)
+  } deriving (Eq, Show, Generic)
```
JaredCorduan: this is great to see, thanks for taking this on @bartfrenk !

JaredCorduan:
@bartfrenk should we close this for the time being?

bartfrenk:
I have put this on hold until @nfrisby was back from holidays, since this PR was part of resolving this issue https://github.com/input-output-hk/ouroboros-consensus/issues/409, which would involve changing the external API of consensus. It seems like something we would still want to do, and confirming that is next on my priority list. @JaredCorduan

cabal.project
```diff
@@ -30,6 +30,9 @@ packages:
   libs/set-algebra
   libs/vector-map
 
+program-options
```
nfrisby: My understanding is that the typical policy among IOG repos is for each developer to add `-Werror` to their `cabal.project.local` file, if they prefer, instead of unilaterally putting it in the repository. I don't know for sure whether that's the case for `cardano-ledger`, but I would guess so.

eras/shelley-ma/impl/cardano-ledger-shelley-ma.cabal
```diff
@@ -52,6 +52,9 @@ library
   other-modules:
     Cardano.Ledger.ShelleyMA.Era
     Cardano.Ledger.ShelleyMA.Rules.Utxo
+    Cardano.Ledger.ShelleyMA.Rules.Utxo
```
nfrisby: This looks like an accidental duplicate line?

eras/shelley/impl/src/Cardano/Ledger/Shelley/Era.hs
```diff
@@ -1,4 +1,4 @@
-{-# LANGUAGE DataKinds #-}
+{-# LANGUAGE DataKinds    #-}
```
nfrisby: I think `cardano-ledger` has different style rules than we do in Consensus. (If I recall correctly, they use `ormolu`.) So tweaks like this are probably out-of-bounds for this PR.

libs/cardano-ledger-core/src/Cardano/Ledger/Core.hs
```diff
@@ -463,6 +464,8 @@ class
 -- > type instance PreviousEra (AllegraEra c) = ShelleyEra c
 type family PreviousEra era = (r :: Type) | r -> era
 
+type family ByronTranslationContext era :: Type
```
nfrisby: I see that you're instantiating this for eras other than `ShelleyEra`. That seems wrong to me. As far as I know, `ShelleyEra` is the only one that is designed to be capable of following `Byron`.

What do you think, @JaredCorduan ?

nfrisby (commented):
Just passing by, had these thoughts to share.

cabal.project
lehins: It is the opposite in ledger. We have `-Werror` enabled for all projects. If a developer wants to turn them back into warnings then he/she can do so in the `cabal.project.local`

libs/cardano-ledger-core/src/Cardano/Ledger/Core.hs
JaredCorduan: That is correct @nfrisby , only `ShelleyEra` should be capable of following `Byron`, and is not a "core" feature for all eras.

nfrisby: Ah, I realize now that @bartfrenk copied these instances from my sketch in the Issue input-output-hk/ouroboros-network#3976. I suppose me disagreeing with my past self shows the sketch needs more thought!

In the Consensus layer, we're reusing one data type for all the Shelley-based eras. For the Shelley era itself, that data type should contain whatever it needs to translate the final Byron ledger state into the first Shelley ledger state. But we want any other era to have `()` for that field, since no other era needs to translate the final Byron ledger state.

Another way to think about it: the `TranslationContext` for `ShelleyEra` should include the necessary bits for the Byron ledger state translation, whereas all the era Shelley-based eras wouldn't need those bits. However, `TranslationContext` is only for translating from Shelley-based era to Shelley-based era, and thus there is no `TranslationContext` for `ShelleyEra` (since it's the first Shelley-based era).

So: `type instance ByronTranslationContext (BabbageEra _) = ()` doesn't mean "`BabbageEra` can translate from Byron without needing any addition data." It instead means "`BabbageEra` doesn't need any such context data since it never translates from Byron".

Maybe a renaming would help clarify?

Alternatively, we could have this data type:

```
data ShelleyLedgerConfig era = ShelleyLedgerConfig {
      shelleyLedgerGlobals                     :: !SL.Globals
    , shelleyLedgerTranslationContext          :: !(Core.TranslationContext era)
    , shelleyLedgerTranslationContextFromByron :: !(Strict.Maybe (SL.ByronTranslationContext era))
    }

type instance ByronTranslationContext ShelleyEra = SL.ParedDownShelleyGenesis
type instance ByronTranslationContext AllegraEra = TypeError ...
type instance ByronTranslationContext AlonzoEra  = TypeError ...
type instance ByronTranslationContext BabbageEra = TypeError ...
type instance ByronTranslationContext ConwayEra  = TypeError ...
```

Thus the person constructing the `ShelleyLedgerConfig ` could `Just` only with the `ShelleyEra`, and must use `Nothing` with all the others, as intended. I _think_ this would also effectively prevent any of those non-`ShelleyEra` era's code from accessing this field: the type of the argument when checking a `Just x` pattern would raise a typeerror, wouldn't it?

Edit: Ah, but want the data type to definitely contain the `ByronTranslationContext` for the `ShelleyEra`--`Nothing` would be very problematic there. So I think we simply need a richer type for the family:

```
type family ByronTranslationContextIfAny era :: Maybe Type

data StaticMaybe :: Maybe Type -> Type where
  StaticJust    :: !a -> StaticMaybe (Just a)
  StaticNothing :: StaticMaybe Nothing
```

nfrisby: Actually, I think the following is the much simpler and ideal option. Just redefine the already existing `TranslationContext (ShelleyEra c)` type family instance from its current `()` right-hand side to the following:

```
-- | There is no @TranslateEra (ShelleyEra c)@ instance, since 'TranslateEra' is for translating from one Shelley-based era to the next, and 'ShelleyEra' is the first such.
-- However, we do need extra data when translating from Byron to 'ShelleyEra' itself. Hence this instance.
type instance TranslationContext (ShelleyEra c) = TranslationContextByronToShelley c

data TranslationContextByronToShelley c = TranslationContextByronToShelley {
  byronToShelleyProtocolParams    :: ShelleyPParams (ShelleyEra c)
, byronToShelleyGenDelegs         :: GenDelegs c
, byronToShelleyMaxLovelaceSupply :: Word64
}
```

cc @amesgen

amesgen: That indeed seems like the most minimal change that would do the job. :+1:

Obviously, this would special-case Byron as a predecessor of the first Shelley era, as we store the `TranslationContext` directly in the `ShelleyLedgerConfig` in Consensus. I am not sure how keen Ledger is on organizing the code in such a way that the Cardano chain (which includes Byron) is not given special treatment compared to a Shelley-only chain (where `TranslationContext (ShelleyEra c) ~ ()` would be natural) or even a chain where Shelley is preceded by a protocol other than Byron (on which `TranslationContext (ShelleyEra c)` might depend).

At least e.g. the docs of `ShelleyGenesis` are alluding to the possibility of a Shelley-only chain:
https://github.com/input-output-hk/cardano-ledger/blob/836452c1ff7b38b16ca5d2d3b0c58383ef59f512/eras/shelley/impl/src/Cardano/Ledger/Shelley/Genesis.hs#L121-L125

JaredCorduan: @amesgen I believe that there are tests from the QA team, and benchmarks from the benchmarking, that make use of a Shelley-only chain. We should check with them.

> I am not sure how keen Ledger is on organizing the code in such a way that the Cardano chain (which includes Byron) is not given special treatment ...

I can't say that I understand the implications of this, sorry.

PR #3057, opened 2022-09-28
regenerate requirements.txt for readthedocs

JaredCorduan:
Regenerate `requirements.txt` in order to fix readthedocs.


teodanciu (approved):
I can't understand the pattern of these failures - though it seems like a version problem,  I find it odd that it only seems to happen in some cases.  
But the change,  whether it solves the problem or not, can't be bad, so I think it's good we merge it and then we see if we need to do more. 

Just checking: the changes in this commit, are they generated based on the existing version of `requirements.in` (not an updated one), correct?

JaredCorduan:
> I can't understand the pattern of these failures - though it seems like a version problem, I find it odd that it only seems to happen in some cases. But the change, whether it solves the problem or not, can't be bad, so I think it's good we merge it and then we see if we need to do more.

I was confused as well. The error had to do with some function signature changing (expecting a `conf`).
 
> Just checking: the changes in this commit, are they generated based on the existing version of `requirements.in` (not an updated one), correct?

Correct! All I did was follow your [wonderful instructions](https://github.com/input-output-hk/cardano-ledger/blob/master/doc/README.md#updating-sphinx-dependencies).

PR #3059, opened 2022-09-29
move RewardType and Reward to cardano-ledger-core package

### LINKED ISSUE #3048: Please move RewardType to cardano-ledger-core
Currently the `RewardType` is defined in `cardano-ledger-shelley` which eventually depends on `plutus-tx`. Since the `RewardType` is needed in the `cardano-db` package, having to provide the Plutus dependencies is a bit of a pain in the neck.

If the definition of `RewardType` was moved to `cardano-ledger-core` things would be improved vastly up the stack.

JaredCorduan:
I moved the `RewardType` and `Reward` types from the `Cardano.Ledger.Shelley.Reward` module in the `cardano-ledger-shelley` package into a new module `Cardano.Ledger.Reward` in the `cardano-ledger-core` package. And added a wee bit of documentation while I was at it.

resolves #3048

### DIFF
--- a/eras/shelley/impl/src/Cardano/Ledger/Shelley/Rewards.hs
+++ b/eras/shelley/impl/src/Cardano/Ledger/Shelley/Rewards.hs
@@ -12,9 +12,9 @@ module Cardano.Ledger.Shelley.Rewards
     PoolRewardInfo (..),
     mkApparentPerformance,
     RewardType (..),
+    Reward (..),
     LeaderOnlyReward (..),
     leaderRewardToGeneral,
-    Reward (..),
     leaderRew,
     memberRew,
     aggregateRewards,
@@ -28,8 +28,6 @@ where
 import Cardano.Binary
   ( FromCBOR (..),
     ToCBOR (..),
-    decodeWord,
-    encodeWord,
   )
 import Cardano.Ledger.BaseTypes
   ( BlocksMade (..),
@@ -37,7 +35,6 @@ import Cardano.Ledger.BaseTypes
     NonNegativeInterval,
     ProtVer,
     UnitInterval,
-    invalidKey,
   )
 import Cardano.Ledger.Coin
   ( Coin (..),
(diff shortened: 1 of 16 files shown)


eras/shelley/impl/src/Cardano/Ledger/Shelley/Rewards.hs
```diff
@@ -11,41 +11,40 @@ module Cardano.Ledger.Shelley.Rewards
...
     leaderRew,
     memberRew,
     aggregateRewards,
     filterRewards,
     sumRewards,
     rewardOnePoolMember,
     mkPoolRewardInfo,
+    -- deprecated
+    RewardType,
+    Reward,
```
lehins: This will only introduce unnecessary breakage downstream it is better to just re-export those types from here.

libs/cardano-ledger-core/src/Cardano/Ledger/Rewards.hs
```diff
@@ -0,0 +1,93 @@
...
+  deriving (Eq, Show, Generic)
+
+-- | Note that this Ord instance is chosen to align precisely
+--  with the Allegra reward aggregation, as given by the
+--  function 'aggregateRewards' so that 'Set.findMax' returns
+--  the expected value.
+instance Ord (Reward c) where
+  compare (Reward MemberReward _ _) (Reward LeaderReward _ _) = GT
+  compare (Reward LeaderReward _ _) (Reward MemberReward _ _) = LT
+  compare (Reward _ pool1 _) (Reward _ pool2 _) = compare pool1 pool2
```
lehins, line 81: Oh crap. Not another one. I created a ticket for it: https://github.com/input-output-hk/cardano-ledger/issues/3060

lehins (changes requested):
I am not really sure what is the motivation behind moving Reward into Core, but I don't think it hurts anyone either. We definitely don't need to add any deprecations. There is nothing wrong with re-exporting types from another module

libs/cardano-ledger-core/src/Cardano/Ledger/Rewards.hs
```diff
@@ -0,0 +1,93 @@
...
+import Data.Coders (Decode (..), Encode (..), decode, encode, (!>), (<!))
+import GHC.Generics (Generic)
+import NoThunks.Class (NoThunks (..))
+
+-- | The staking rewards in Cardano are all either:
+--
+-- * member rewards - rewards given to a registered stake credential which has delegated
+-- to a stake pool, or
+--
+-- * leader rewards - rewards given to a registered stake pool (in particular, given to to the
```
lehins:
```suggestion
-- * leader rewards - rewards given to a registered stake pool (in particular, given to the
```

lehins (approved):
LGTM

libs/cardano-ledger-core/src/Cardano/Ledger/Rewards.hs
JaredCorduan: given :tutu:

Issue #3014, opened 2022-08-31
Versioned serialization schemes for types


JaredCorduan:
# The problem

When we encounter a problem with our deserializers, such as in #3003, #2965 and  #2444, it can be very difficult to implement a fix. It is difficult because we can only fix such issues during a hard fork, and leading up to the hard fork we must maintain two serializations for the same type in order to not cause unintended network splitting (the problematic version must be used before the hard fork, and the fixed version is used afterwards). This can be especially tricky with the `FromCBOR` typeclass, since it is not always easy to search for where all the problematic uses are located.

# A solution

Create a new typeclass to replace `ToCBOR` and `FromCBOR` which depends not only on the type but also on the protocol version. The first instances will all be implemented identically to our current `ToCBOR` and `FromCBOR` instances for all protocol versions up to the current one.

Note that this change that will effect the downstream components. We will need to coordinate with consensus and node, and many other teams.

michaelpj:
My $0.02. 

We have a similar problem in `plutus` with our serialization for scripts. I think the considerations are the same although t they're slightly different libraries. What we do at the moment is we don't use the typeclass that the library provides, but rather we just use explicit decoder values. For CBOR this would be something like:
```
-- don't do this
instance Serialise Foo where
  decode :: Decoder s Foo

-- do this
decodeFoo :: ProtocolVersion -> Decoder s Foo
```

It's a bit clunky, but it works okay. You can also then take a mixed approach: if you need to decode something that *isn't* version-sensitive (an integer, say), you can just call `decode` and use the typeclass version.

I guess we could define a typeclass for this, but that does force you to pick an interface for all the decoders to share. And in fact I lied above: we don't just pass the protocol version, in some cases we also pass a predicate that indicates whether particular builtins are allowed. So we do make use of the variability. Something to consider, anyway.

HeinrichApfelmus:
I second @michaelpj 's suggestion — for a slightly different reason.

In cardano-wallet, we always have trouble finding the piece of code that is responsible for serializing/deserializing a specific type, because it's hard to figure out which instance of which type class is selected (this information is inferred from the code, but not written explicitly). So, usually we end up with "Hm, I think it's ok to use `toCBOR` here, but I'm not quite sure". The "not quite sure part" makes me feel uneasy — it's acceptable for the `Show` type, but not if I want to explicitly control the format.

[...]

PR #3063, opened 2022-10-05
Versioned CBOR

lehins:
Started new package `cardano-ledger-binary`:
    
This package introduces new `Encoding` and `Decoder` wrappers that can change their behavior depending on a version, which corresponds to major protocol version. This package introduces new conceptws and consolidates all the old ones that were spread around different packages:
    
* Introduced `Version` in its own module and make it safe to construct
* Switch from `ReaderT` to manual Version passing to recover      representational role
* All decoders and encoders as well as `FromCBOR`/`ToCBOR` classes      from `cardano-binary` package
* Re-exports `Annotator` functionality from `cardano-binary` package,      with plans to migrate it completely in the future
* Moved dropper from `cardano-binary`
* Test suites from `cardano-binary` and `cardano-data` as vintage test      suites
* `FromSharedCBOR` from `cardano-data` package
* `Data.Coders` module from `cardano-data` package
* `Data.Twiddler` module from `cardano-data` package into a `test-lib`
* Improved Roundtripping tests and sophisticated way to display CBOR      encoded blobs that was transferred from `ouroboros-consensus` repo.
* Custom decoders/encoders from `Cardano.Ledger.Serialization` module      from `cardano-ledger-core` package
* Implement round trip test for all To/FromCBOR instances. Discovered      major bug with `SlicedByteArray` and fix it in `cborg` package:      https://github.com/well-typed/cborg/pull/301

### DIFF
--- a/libs/cardano-ledger-binary/.ghcid
+++ b/libs/cardano-ledger-binary/.ghcid
@@ -1,0 +1,1 @@
+-c "nix-shell ../../shell.nix --run \"cabal repl cardano-ledger-binary\"" -o ghcid.txt
--- a/libs/cardano-ledger-binary/src/Cardano/Ledger/Binary.hs
+++ b/libs/cardano-ledger-binary/src/Cardano/Ledger/Binary.hs
@@ -1,0 +1,35 @@
+{-# LANGUAGE FlexibleContexts #-}
+
+module Cardano.Ledger.Binary
+  ( module Cardano.Ledger.Binary.Decoding,
+    module Cardano.Ledger.Binary.Encoding,
+    module Cardano.Ledger.Binary.Version,
+    Term (..),
+    C.DeserialiseFailure (..),
+    translateViaCBORAnnotator,
+  )
+where
+
+import Cardano.Ledger.Binary.Decoding
+import Cardano.Ledger.Binary.Encoding
+import Cardano.Ledger.Binary.Version
+import qualified Codec.CBOR.Read as C (DeserialiseFailure (..))
+import Codec.CBOR.Term (Term (..))
+import Control.Monad.Except (Except, MonadError (throwError))
+import Data.Text (Text)
+
+-- | Translate one type into another through their common binary format. This is
+-- useful for types that build upon one another and are "upgradeable". It is
+-- important to note that the deserialization will happen with `Annotator`,
+-- since that is usually the way to deserialize upgradeable types that live on
+-- chain.
+translateViaCBORAnnotator ::
+  (ToCBOR a, FromCBOR (Annotator b)) =>
+  Version ->
+  Text ->
+  a ->
+  Except DecoderError b
+translateViaCBORAnnotator version name x =
+  case decodeFullAnnotator version name fromCBOR (serialize version x) of
+    Right newx -> pure newx
+    Left decoderError -> throwError decoderError
--- a/libs/cardano-ledger-binary/src/Cardano/Ledger/Binary/Decoding.hs
+++ b/libs/cardano-ledger-binary/src/Cardano/Ledger/Binary/Decoding.hs
@@ -1,0 +1,228 @@
+{-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE RankNTypes #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeApplications #-}
+
+module Cardano.Ledger.Binary.Decoding
+  ( -- * Running decoders
+    decodeFull,
+    decodeFull',
+    decodeFullDecoder,
+    decodeFullDecoder',
+    decodeFullAnnotator,
+    decodeFullAnnotatedBytes,
+    module Cardano.Ledger.Binary.Version,
+    module Cardano.Ledger.Binary.Decoding.FromCBOR,
+    module Cardano.Ledger.Binary.Decoding.Sharing,
+    module Cardano.Ledger.Binary.Decoding.Decoder,
+    module Cardano.Ledger.Binary.Decoding.Annotated,
+    module Cardano.Ledger.Binary.Decoding.Sized,
+    module Cardano.Ledger.Binary.Decoding.Drop,
+
+    -- * Nested CBOR in CBOR
+    decodeNestedCbor,
+    decodeNestedCborBytes,
+
+    -- * Unsafe deserialization
+    unsafeDeserialize,
+    unsafeDeserialize',
+
+    -- * Helpers
+    toStrictByteString,
+
+    -- * Deprecated
+    decodeAnnotator,
+    fromCBORMaybe,
+  )
+where
+
+import Cardano.Ledger.Binary.Decoding.Annotated
+import Cardano.Ledger.Binary.Decoding.Decoder
+import Cardano.Ledger.Binary.Decoding.Drop
+import Cardano.Ledger.Binary.Decoding.FromCBOR
+import Cardano.Ledger.Binary.Decoding.Sharing
+import Cardano.Ledger.Binary.Decoding.Sized
+import Cardano.Ledger.Binary.Version
+import Codec.CBOR.Read as Read (DeserialiseFailure, IDecode (..), deserialiseIncremental)
+import Codec.CBOR.Write (toStrictByteString)
+import Control.Exception (displayException)
+import Control.Monad (when)
+import Control.Monad.ST (ST, runST)
+import Data.Bifunctor (bimap)
+import qualified Data.ByteString as BS
+import qualified Data.ByteString.Lazy as BSL
+import qualified Data.ByteString.Lazy.Internal as BSL
+import Data.Proxy (Proxy (Proxy))
+import Data.Text (Text)
+
+-- | Deserialize a Haskell value from the external binary representation, which
+-- have been made using 'serialize' or a matching serialization functionilty in
+-- another language that uses CBOR format. Accepts lazy `BSL.ByteString` as
+-- input, for strict variant use `unsafeDeserialize'` instead.
+--
+-- This deserializaer is not safe for these reasons:
+--
+-- *  /Throws/: @'Read.DeserialiseFailure'@ if the given external
+--   representation is invalid or does not correspond to a value of the
+--   expected type.
+--
+-- * Decoding will not fail when the binary input is not consumed in full.
+unsafeDeserialize ::
+  FromCBOR a =>
+  Version ->
+  BSL.ByteString ->
+  a
+unsafeDeserialize version =
+  either (error . displayException) id . bimap fst fst . deserialiseDecoder version fromCBOR
+
+-- | Variant of 'unsafeDeserialize' that accepts a strict `BS.ByteString` as
+-- input.
+unsafeDeserialize' :: FromCBOR a => Version -> BS.ByteString -> a
+unsafeDeserialize' version = unsafeDeserialize version . BSL.fromStrict
+
+-- | Deserialize a Haskell value from a binary CBOR representation, failing if
+--   there are leftovers. In other words, the __Full__ here implies the contract
+--   on this function that the input must be consumed in its entirety by the
+--   decoder specified in `FromCBOR`.
+decodeFull :: forall a. FromCBOR a => Version -> BSL.ByteString -> Either DecoderError a
+decodeFull version = decodeFullDecoder version (label $ Proxy @a) fromCBOR
+
+-- | Same as `decodeFull`, except accepts a strict `BS.ByteString` as input
+-- instead of the lazy one.
+decodeFull' :: forall a. FromCBOR a => Version -> BS.ByteString -> Either DecoderError a
+decodeFull' version = decodeFull version . BSL.fromStrict
+
+-- | Same as `decodeFull`, except instead of relying on the `FromCBOR` instance
+-- the `Decoder` must be suplied manually.
+decodeFullDecoder ::
+  -- | Protocol version to be used during decoding.
+  Version ->
+  -- | Label for error reporting
+  Text ->
+  -- | The parser for the @ByteString@ to decode. It should decode the given
+  -- @ByteString@ into a value of type @a@
+  (forall s. Decoder s a) ->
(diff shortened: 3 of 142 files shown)


libs/cardano-ledger-binary/.ghcid
```diff
@@ -0,0 +1 @@
+-c "nix-shell ../../shell.nix --run \"cabal repl cardano-ledegr-binary\"" -o ghcid.txt
```
TimSheard: Should this
 \"cabal repl cardano-ledegr-binary\""
be this
 \"cabal repl cardano-ledger-binary\""

libs/cardano-ledger-binary/src/Cardano/Ledger/Binary.hs
```diff
@@ -0,0 +1,30 @@
...
+import Codec.CBOR.Term (Term (..))
+import Control.Monad.Except (Except, MonadError (throwError))
+import Data.Text (Text)
+
+translateViaCBORAnnotator ::
+  (ToCBOR a, FromCBOR (Annotator b)) =>
+  Version ->
+  Text ->
+  a ->
+  Except DecoderError b
```
TimSheard: Add a comment about what this does?

libs/cardano-ledger-binary/src/Cardano/Ledger/Binary/Decoding.hs
```diff
@@ -0,0 +1,206 @@
...
+--   /Throws/: @'Read.DeserialiseFailure'@ if the given external
+--   representation is invalid or does not correspond to a value of the
+--   expected type.
+unsafeDeserialize :: FromCBOR a => Version -> BSL.ByteString -> a
+unsafeDeserialize version =
+  either (error . displayException) id . bimap fst fst . deserialiseDecoder version fromCBOR
+
+-- | Strict variant of 'deserialize'.
+unsafeDeserialize' :: FromCBOR a => Version -> BS.ByteString -> a
+unsafeDeserialize' version = unsafeDeserialize version . BSL.fromStrict
```
TimSheard: I  know you probably didn't write these comments, but it would be better to say

Deserialize a Haskell value from the external binary representation stored in a lazy bytestring

Deserialize a Haskell value from the external binary representation stored in a strict bytestring

Usually when we say a strict version we mean the function is strict in its argument(s).

libs/cardano-ledger-binary/src/Cardano/Ledger/Binary/Decoding/Coders.hs
```diff
@@ -0,0 +1,578 @@
...
+--
+-- In this example, let @Int@ and @C@ have 'ToCBOR' instances, and @dualB :: Dual B@.
+--
+-- @
+-- data C = C Text.Text
+-- instance ToCBOR C where toCBOR (C t) = toCBOR t
+-- instance FromCBOR C where fromCBOR = C <$> fromCBOR
+--
+-- data B = B Text.Text
+-- dualB = Dual (\ (B t) ->toCBOR t) (B <$> fromCBOR)
```
TimSheard: Does the new code still support Duals?  If not we need to change these comments.

libs/cardano-ledger-binary/src/Cardano/Ledger/Binary/Decoding/Decoder.hs
```diff
@@ -0,0 +1,1064 @@
...
+-- | Use monadic syntax to extract value level version of the decoder from its type
+--
+-- >>> decodeFullDecoder 3 "Version" getDecoderVersion ""
+-- Right 3
+getDecoderVersion :: Decoder s Version
+getDecoderVersion = Decoder pure
+
+-- | Conditionoly choose the decoder newer or older deceder, depending on the current
+-- version. Supplied version acts as a pivot.
+--
```
TimSheard: conditionally  (I think)

libs/cardano-ledger-binary/src/Cardano/Ledger/Binary/Decoding/Decoder.hs
```diff
@@ -0,0 +1,1064 @@
...
+--------------------------------------------------------------------------------
+
+-- | @'D.Decoder'@ for list.
+decodeList :: Decoder s a -> Decoder s [a]
+decodeList decodeValue =
+  ifDecoderVersionAtLeast
+    (natVersion @2)
+    (decodeCollection decodeListLenOrIndef decodeValue)
+    (decodeListWith decodeValue)
+
```
TimSheard: Is this a backwards compatibility thing?   Give an explanation why we use the magic "2"

libs/cardano-ledger-binary/src/Cardano/Ledger/Binary/Decoding/Decoder.hs
```diff
@@ -0,0 +1,1064 @@
...
+decodeMapByKey decodeKey decodeValueFor =
+  uncurry Exts.fromListN
+    <$> decodeCollectionWithLen decodeMapLenOrIndef decodeInlinedPair
+  where
+    decodeInlinedPair = do
+      !key <- decodeKey
+      !value <- decodeValueFor key
+      pure (key, value)
+
+-- | We want to make a uniform way of encoding and decoding `Map.Map`. The otiginal ToCBOR
```
TimSheard: Original  not otiginal (I think)

libs/cardano-ledger-binary/src/Cardano/Ledger/Binary/Encoding/ToCBOR.hs
```diff
@@ -0,0 +1,709 @@
...
+
+newtype PreEncoded = PreEncoded {unPreEncoded :: BS.ByteString}
+
+instance ToCBOR PreEncoded where
+  toCBOR = encodePreEncoded . unPreEncoded
+
+--------------------------------------------------------------------------------
+-- Size expressions
+--------------------------------------------------------------------------------
+
```
TimSheard, line 116: We should ask around if anyone uses the size stuff, and get rid of it if not.  Maybe we should just deprecate it at first.

libs/cardano-ledger-binary/src/Cardano/Ledger/Binary/Version.hs
```diff
@@ -0,0 +1,73 @@
...
+-- | Protocol version number that is used during encoding and decoding. All supported
+-- versions are in the range from `MinVersion` to `MaxVersion`.
+newtype Version = Version Word
+  deriving (Eq, Ord, Show, Enum)
+
+-- | Minimum supported version
+type MinVersion = 1
+
+-- | Maximum supported version
+type MaxVersion = 8
```
TimSheard, line 39: Why are we using type versions of Nat.  None of the these types seem to be used as indexes, so what does it buy us?

libs/cardano-ledger-binary/src/Cardano/Ledger/Binary/Version.hs
```diff
@@ -0,0 +1,73 @@
...
+  maxBound = Version (fromInteger (natVal (Proxy @MaxVersion)))
+
+-- | Same as `natVersionProxy`, construct a version from a type level `Nat`, except it can be
+-- supplied through @TypeApplications@.
+natVersion :: forall v. (KnownNat v, MinVersion <= v, v <= MaxVersion) => Version
+natVersion = natVersionProxy (Proxy @v)
+
+-- | Safely construct a `Version` from a type level `Nat`, which is supplied as a `Proxy`
+natVersionProxy :: (KnownNat v, MinVersion <= v, v <= MaxVersion) => Proxy v -> Version
+natVersionProxy = Version . fromInteger . natVal
```
TimSheard, line 52: So maybe this is where the type level Nat  is used.

libs/cardano-ledger-binary/test-lib/Test/Cardano/Ledger/Binary/RoundTrip.hs
```diff
@@ -0,0 +1,150 @@
...
+    where
+      bytes = BSL.toStrict rtfEncodedBytes
+      prettyTerm =
+        case decodeFullDecoder' rtfVersion "Term" decodeTerm bytes of
+          Left err -> "Could not decode as Term: " ++ show err
+          Right term -> showExpr term
+
+-- | A definition of a CBOR trip through binary representation of one type to
+-- another. In this module this is called an embed. When a source and target type is the
+-- exect same one then it would be a dual and is expected to round trip.
```
TimSheard: exact

libs/cardano-ledger-binary/test-lib/Test/Cardano/Ledger/Binary/Twiddle.hs
```diff
@@ -0,0 +1,190 @@
...
+{-# LANGUAGE UndecidableInstances #-}
+
+module Test.Cardano.Ledger.Binary.Twiddle
+  ( Twiddler (unTwiddler),
+    Twiddle (..),
+    encodingToTerm,
+    toTwiddler,
+    toTerm,
+  )
+where
```
TimSheard, line 20: This is a huge file. Somebody should explain what's going on here.

TimSheard (approved):
Aside from Twiddle, all this looks familiar, and in good shape.
Just a few comments need to be added, and a few mispelling typos to be fixed.

libs/cardano-ledger-binary/test-lib/Test/Cardano/Ledger/Binary/Twiddle.hs
Soupstraw: That's true, I should properly document this module. I'll create an issue for that.

Other than that, this module looks fine to me, as the Twiddle instances should not care about the deserializer version unless we make some low level changes such as requiring all terms to be in canonical form or such.

JaredCorduan (approved):
Thank you @lehins , this is extremely exciting.

I was given a video tour of the PR on friday, and today I looked over the next instances. Looks great to me.

Note that this PR resolves #3014

libs/cardano-ledger-binary/src/Cardano/Ledger/Binary/Decoding/Decoder.hs
lehins: Done

libs/cardano-ledger-binary/.ghcid
lehins: Yes. Thank you. Done

libs/cardano-ledger-binary/src/Cardano/Ledger/Binary/Version.hs
lehins: Yes, `natVersion` and `natVersionProxy` are **total functions** that allow us constructing `Version` using type level `Nat`s

libs/cardano-ledger-binary/src/Cardano/Ledger/Binary/Version.hs
lehins: This gives us a range of valid versions at the type level, which we can use later to construct `Version` without any call to `error`. In particular `natVersion` and `natVersionProxy` let us do that.

libs/cardano-ledger-binary/src/Cardano/Ledger/Binary/Decoding/Decoder.hs
lehins: Yep. Done

libs/cardano-ledger-binary/src/Cardano/Ledger/Binary/Decoding/Decoder.hs
lehins: Magic number `2` is the major protocol version for Shelley. This is used throughout the library therefore it doesn't make sense to describe what does the version mean at each use case. The reader should check the haddock for `Version` type and possibly functions `ifDecoderVersionAtLeast` and `natVersion` to see what it all means.

That being said I did document the difference in behavior of `decodeList` between versions.

libs/cardano-ledger-binary/src/Cardano/Ledger/Binary/Decoding/Coders.hs
lehins: No, it does not. I've adjusted the example to reflect lack of `Dual` type.

libs/cardano-ledger-binary/src/Cardano/Ledger/Binary/Encoding/ToCBOR.hs
lehins: This is a good idea. I'd rather create a separate task for this, since I do have a suspicion that it might be used in some obscure cases. 

Here is the ticket: https://github.com/input-output-hk/cardano-ledger/issues/3071

libs/cardano-ledger-binary/test-lib/Test/Cardano/Ledger/Binary/RoundTrip.hs
lehins: Done

libs/cardano-ledger-binary/test-lib/Test/Cardano/Ledger/Binary/Twiddle.hs
lehins: Here is a link to the ticket that tracks documenting the module: https://github.com/input-output-hk/cardano-ledger/issues/3067

libs/cardano-ledger-binary/src/Cardano/Ledger/Binary.hs
lehins: Not a function that I introduced, but I did add some haddock.

libs/cardano-ledger-binary/src/Cardano/Ledger/Binary/Decoding.hs
lehins: Yeah, it is a good point. I updated haddock for all deserialization functions.

lehins:
> Note that this PR resolves https://github.com/input-output-hk/cardano-ledger/issues/3014

Yes, it does, except only the first half of it. This PR only implements the library that is capable of versioned CBOR. A subsequent PR will be submitted that actually applies this new interface throughout ledger.

Issue #3035, opened 2022-09-13
benchmark TICKF


JaredCorduan:
We need realistic benchmarks of the `TICKF` rule in order to determine if #3034 is a real problem.

---

The consensus layer calls `TICKF` once per slot to determine if a block producing node has been elected leader. When the slot in question is in a new epoch (relative to the last ledger view that the consensus layer has stored), this results in a potentially expensive computation (see #3034).

`TICKF` is defined here (every ledger era has used the Shelley TICKF):

https://github.com/input-output-hk/cardano-ledger/blob/1176affe9b2a7a3b50c1ceda00cc92f237ea6360/eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Tick.hs#L262

---

In order to get realistic data for our benchmarks, we can use [db-analyser](https://github.com/input-output-hk/ouroboros-network/blob/master/ouroboros-consensus-cardano-tools/Documentation.md) to get the ledger state on mainnet just before some of the epoch boundaries. The benchmarks will deserialize the ledger state obtained from db-analyser, and project the `NewEpochState` for use in the `TICKF` rule.

Some slots we might consider (this issue was created in the middle of epoch 363):

|slot number| block number| last block of epoch X|
|---|---|---|
|71452796|7749373|362|
|71020795|7728107|361|
|70588777|7706865|360|
|70156668|7685639|359|
|69724795|7664616|358|

Note that the benchmarks will need to use a `Slot` past the epoch boundary in order to trigger the large calculations.

---

Tasks:

- [ ] Write a tool that takes a ledger state snapshot from db-analyser and benchmarks the `TICKF` rule.
- [ ] Create ledger state snapshots for the slots in the table above.
- [ ] Record the times obtained from running the new tool on the snapshots. (The report can be placed in this GitHub issue.)

lehins:
First task is practically done, we have benchmarks that take **mainnet** `NewEpochState` in cbor encoded form and run various benchmarks using that state:

https://github.com/input-output-hk/cardano-ledger/blob/1176affe9b2a7a3b50c1ceda00cc92f237ea6360/libs/ledger-state/bench/Performance.hs#L39-L45

All it needs is a change of `CurrentEra` to `Babbage`, since now it is set to `Alonzo`

dnadales:
I'll link this PR here in case it provides relevant information for y'all https://github.com/input-output-hk/ouroboros-network/pull/4014

TimSheard:
[...]

PR #3068, opened 2022-10-10
Ts new calc pool distr

TimSheard:
Add some benchmarks to speed up the tickf rule


eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState/IncrementalStake.hs
```diff
@@ -1,14 +1,25 @@
+{-# LANGUAGE BangPatterns #-}
 {-# LANGUAGE DataKinds #-}
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE RankNTypes #-}
 
+-- ===========================================================================
+-- There are three parts to IncrementalStaking.
+-- 1) The incremental part, where we keep track of each update to the UtXO
+--    adding Inputs and deleting consumed Outputs. Done in the Utxo rules.
+-- 2) Finalizing and aggregating by stake cedential to create a Snapshot.
```
ltouro:
```suggestion
-- 2) Finalizing and aggregating by stake credential to create a Snapshot.
```

TimSheard: done

JaredCorduan (approved):
This looks wonderful @TimSheard , thank you for adding in all those very helpful comments!

Do you mind adding notes to #3035 with the results of the benchmarks? it will be really great to have a bit of a record of the results (and the community might be interested to learn as well).

eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState/IncrementalStake.hs
```diff
@@ -1,14 +1,25 @@
+{-# LANGUAGE BangPatterns #-}
 {-# LANGUAGE DataKinds #-}
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE RankNTypes #-}
 
+-- ===========================================================================
+-- There are three parts to IncrementalStaking.
+-- 1) The incremental part, where we keep track of each update to the UtXO
```
lehins:
```suggestion
-- 1) The incremental part, where we keep track of each update to the UTxO
```

eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState/IncrementalStake.hs
```diff
@@ -301,8 +222,120 @@ resolveActiveIncrementalPtrs isActive ptrMap_ (IStake credStake ptrStake) =
...
+data Filtered era = Filtered
+  { -- Only the first component is strict on purpose
+
+    -- | These are registered, but in the ShelleyEra they are ignored because of backward compatibility
+    --  in other Eras, this field will be the Map.empty
+    registeredF :: !(Map (Credential 'Staking (EraCrypto era)) (Set (Reward (EraCrypto era)))),
+    shelleyIgnoredF :: Map (Credential 'Staking (EraCrypto era)) (Set (Reward (EraCrypto era))),
+    unregisteredF :: Set (Credential 'Staking (EraCrypto era)),
+    totalUnregisteredF :: Coin
+  }
```
lehins: I think it would be helpful to name the type `FilteredRewards` and make it a return type of the `applyRUpd'` (which would be good to rename too.

Also we should stick to the convention with prefixing, not suffixing, with `FilteredRewards` record names would become:

```suggestion
data FilteredRewards era = FilteredRewards
  { -- Only the first component is strict on purpose

    -- | These are registered, but in the ShelleyEra they are ignored because of backward compatibility
    --  in other Eras, this field will be the Map.empty
    frRegistered :: !(Map (Credential 'Staking (EraCrypto era)) (Set (Reward (EraCrypto era)))),
    frShelleyIgnored :: Map (Credential 'Staking (EraCrypto era)) (Set (Reward (EraCrypto era))),
    frUnregistered :: Set (Credential 'Staking (EraCrypto era)),
    frTotalUnregistered :: Coin
  }
```

eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState/IncrementalStake.hs
```diff
@@ -301,8 +222,120 @@ resolveActiveIncrementalPtrs isActive ptrMap_ (IStake credStake ptrStake) =
...
+  EpochState era ->
+  ( EpochState era,
+    Map (Credential 'Staking (EraCrypto era)) (Set (Reward (EraCrypto era))),
+    Map (Credential 'Staking (EraCrypto era)) (Set (Reward (EraCrypto era))),
+    Set (Credential 'Staking (EraCrypto era))
+  )
+applyRUpd'
+  ru
+  es@(EpochState as ss ls pr pp _nm) =
+    (epochStateAns, registered, registeredIgnored, unregistered)
```
lehins: I think it would make this function cleaner if we utilized the newly added type (sure it would also include previously unreturned `totalUnregistered`, but I don't see it as a problem). I also think that current name is just terrible, so maybe we can take this opportunity and improve it to something like `applyFilteredRUpd` or anything else that doesn't have a tick
```suggestion
  (EpochState era, FilteredRewards era))
applyFilteredRUpd
  ru
  es@(EpochState as ss ls pr pp _nm) = (epochStateAns, filteredRewards)
```

eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState/IncrementalStake.hs
```diff
@@ -301,8 +222,120 @@ resolveActiveIncrementalPtrs isActive ptrMap_ (IStake credStake ptrStake) =
...
+    where
+      utxoState_ = lsUTxOState ls
+      delegState = lsDPState ls
+      !epochStateAns = EpochState as' ss ls' pr pp nm'
+      dState = dpsDState delegState
+      Filtered registered registeredIgnored unregistered totalUnregistered = filterAllRewards (rs ru) es
```
lehins: If we return the whole type, then no need to bind all of values. Also `NamedFieldPuns` is a useful extension in such cases, since then we don't need to rely on the order of records in the type:
```suggestion
      filteredRewards@FilteredRewards {frRegistered, frTotalUnregistered} = filterAllRewards (rs ru) es
```

eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState/IncrementalStake.hs
```diff
@@ -301,8 +222,120 @@ resolveActiveIncrementalPtrs isActive ptrMap_ (IStake credStake ptrStake) =
...
+
+    -- | These are registered, but in the ShelleyEra they are ignored because of backward compatibility
+    --  in other Eras, this field will be the Map.empty
+    registeredF :: !(Map (Credential 'Staking (EraCrypto era)) (Set (Reward (EraCrypto era)))),
+    shelleyIgnoredF :: Map (Credential 'Staking (EraCrypto era)) (Set (Reward (EraCrypto era))),
+    unregisteredF :: Set (Credential 'Staking (EraCrypto era)),
+    totalUnregisteredF :: Coin
+  }
+
+-- | Return aggregated information from a reward mapping from the previous Epocc
```
lehins:
```suggestion
-- | Return aggregated information from a reward mapping from the previous Epoch
```

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/NewEpoch.hs
```diff
@@ -244,3 +248,22 @@ instance
   where
   wrapFailed = MirFailure
   wrapEvent = MirEvent
+
+-- ===========================================
+
+updateRewards ::
+  ( HasField "_protocolVersion" (PParams era) ProtVer
+  ) =>
```
lehins:
```suggestion
  (HasField "_protocolVersion" (PParams era) ProtVer) =>
```

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/NewEpoch.hs
```diff
@@ -244,3 +248,22 @@ instance
...
+  EpochState era ->
+  EpochNo ->
+  RewardUpdate (EraCrypto era) ->
+  Rule (ShelleyNEWEPOCH era) 'Transition (EpochState era)
+updateRewards es e ru'@(RewardUpdate dt dr rs_ df _) = do
+  let totRs = sumRewards (esPrevPp es) rs_
+  Val.isZero (dt <> (dr <> toDeltaCoin totRs <> df)) ?! CorruptRewardUpdate ru'
+  let (!es', regRU, eraIgnored, unregistered) = applyRUpd' ru' es
+  tellEvent $ RestrainedRewards e eraIgnored unregistered
```
lehins: By switching `applyRUpd'` to return `FilteredRewards` we could also add `totalUnregistered` to the `RestrainedRewards` event, maybe it is useful info for the event consumer?

libs/cardano-ledger-test/bench/Bench/Cardano/Ledger/StakeDistr.hs
```diff
@@ -0,0 +1,388 @@
...
+newepochFilePath = "/home/sheard/Two/Data/71452796-NewEpochState.cbor"
+
+-- ==============================
+-- Stored constants read from files. Globals and NewEpochState
+
+instance NFData Globals where
+  rnf (Globals {}) = ()
+
+instance NFData (Filtered era) where
+  rnf (Filtered a b c d) = seq (rnf a) (seq (rnf b) (seq (rnf c) (rnf d)))
```
lehins: This instance should be moved right next to the type. Same goes for the `Globals` type above. There is never a reason to define `NFData` Instances as orphans.

Another minor suggestion that maybe will make the implementation a bit easier to read
```suggestion
  rnf (Filtered a b c d) = a `deepseq` b `deepseq` c `deepseq` rnf d
```

libs/cardano-ledger-test/bench/Bench/Cardano/Ledger/StakeDistr.hs
```diff
@@ -0,0 +1,388 @@
...
+
+readGlobals :: IO Globals
+readGlobals = pure globals
+  where
+    pp = def
+    !globals = mkGlobals shelleyGenesis pp
```
lehins: There  is no need to `seq` globals, since `env` function will evaluate it to NF
```suggestion
readGlobals = pure $ mkGlobals shelleyGenesis pp
  where
    pp = def
```

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/NewEpoch.hs
```diff
@@ -209,20 +203,30 @@ calculatePoolDistr' :: forall c. (KeyHash 'StakePool c -> Bool) -> SnapShot c ->
...
-              includeHash d
-          ]
+      poolstake :: Map.Map (KeyHash 'StakePool c) Word64
+      poolstake = calculatePoolStake includeHash delegs stake
    in PoolDistr $
         Map.intersectionWith
-          IndividualPoolStake
-          sd
-          (toMap (VMap.map _poolVrf poolParams))
+          (\word64 poolparam -> IndividualPoolStake (fromIntegral word64 % nonZeroTotal) (_poolVrf poolparam))
```
lehins: Less polymorphic alternative to `fromIntegral` and more Haskell camelCase convention. Also if we make poolStake to contain `CompactCoin` instead of `Word64` we need to unwrap it
```suggestion
          (\(CompactCoin cc) poolParam -> IndividualPoolStake (toInteger cc % nonZeroTotal) (_poolVrf poolParam))
```

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/NewEpoch.hs
```diff
@@ -209,20 +203,30 @@ calculatePoolDistr' :: forall c. (KeyHash 'StakePool c -> Bool) -> SnapShot c ->
...
+          poolstake
+          (VMap.toMap poolParams)
+
+-- | Sum up the Coin (as CompactForm Coin = Word64) for each StakePool
+calculatePoolStake ::
+  (KeyHash 'StakePool c -> Bool) ->
+  VMap VB VB (Credential 'Staking c) (KeyHash 'StakePool c) ->
+  Stake c ->
+  Map.Map (KeyHash 'StakePool c) Word64
+calculatePoolStake includeHash delegs stake = VMap.foldlWithKey accum Map.empty delegs
```
lehins, line 222: I suggest we still return `CompactCoin` to disambiguate it from more plain `Word64`, this would also remove the need for `== Word64` comment in the haddock of this function. This ca easily and efficiently be achieved with `coerce`
```suggestion
  Map.Map (KeyHash 'StakePool c) CompactCoin
calculatePoolStake includeHash delegs stake = coerce $ VMap.foldlWithKey accum Map.empty delegs
```

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/NewEpoch.hs
```diff
@@ -209,20 +203,30 @@ calculatePoolDistr' :: forall c. (KeyHash 'StakePool c -> Bool) -> SnapShot c ->
...
-      sd =
-        Map.fromListWith (+) $
-          [ (d, c % nonZeroTotal)
-            | (hk, compactCoin) <- VMap.toAscList (unStake stake),
-              let Coin c = fromCompact compactCoin,
-              Just d <- [VMap.lookup hk delegs],
-              includeHash d
-          ]
+      poolstake :: Map.Map (KeyHash 'StakePool c) Word64
+      poolstake = calculatePoolStake includeHash delegs stake
```
lehins: Just a suggestion about convention:
```suggestion
      poolStake :: Map.Map (KeyHash 'StakePool c) Word64
      poolStake = calculatePoolStake includeHash delegs stake
```

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/NewEpoch.hs
```diff
@@ -209,20 +203,30 @@ calculatePoolDistr' :: forall c. (KeyHash 'StakePool c -> Bool) -> SnapShot c ->
...
+  VMap VB VB (Credential 'Staking c) (KeyHash 'StakePool c) ->
+  Stake c ->
+  Map.Map (KeyHash 'StakePool c) Word64
+calculatePoolStake includeHash delegs stake = VMap.foldlWithKey accum Map.empty delegs
+  where
+    accum ans cred keyhash =
+      if includeHash keyhash
+        then case VMap.lookup cred (unStake stake) of
+          Nothing -> ans
+          Just (CompactCoin c) -> Map.insertWith (+) keyhash c ans
```
lehins: camelCase it up :) 
```suggestion
    accum ans cred keyHash =
      if includeHash keyHash
        then case VMap.lookup cred (unStake stake) of
          Nothing -> ans
          Just (CompactCoin c) -> Map.insertWith (+) keyHash c ans
```

eras/shelley/test-suite/src/Test/Cardano/Ledger/Shelley/Rewards.hs
```diff
@@ -750,7 +751,7 @@ eventsMirrorRewards events nes = same eventRew compRew
             (completed, lastevent) = complete pulser
     total = getMostRecentTotalRewardEvent events
     aggevent = aggIncrementalRewardEvents events
-    (aggFilteredEvent, _, _, _) = filterAllRewards aggevent (nesEs nes)
+    Filtered aggFilteredEvent _ _ _ = filterAllRewards aggevent (nesEs nes)
```
lehins: It is safer to use record accessor function:
```suggestion
    aggFilteredEvent = frRegistered $ filterAllRewards aggevent (nesEs nes)
```

eras/shelley/test-suite/src/Test/Cardano/Ledger/Shelley/Serialisation/EraIndepGenerators.hs
```diff
@@ -631,8 +634,26 @@ instance Arbitrary PerformanceEstimate where
...
 deriving instance Arbitrary (CompactForm Coin)
 
+-- | In the system, Stake never contains more than the sum of all Ada (which is constant).
+-- This makes it safe to store individual Coins (in CompactForm) as Word64. But we must
+-- be careful that we never generate Stake where the sum of all the coins exceeds (maxBound :: Word64)
+-- There will never be a real Stake in the system with that many Ada, because total Ada is constant.
+-- So using a restricted Arbitrary Generator is OK.
 instance CC.Crypto c => Arbitrary (Stake c) where
-  arbitrary = Stake <$> arbitrary
+  arbitrary = Stake <$> (VMap.fromMap <$> theMap)
```
lehins, line 642:
```suggestion
  arbitrary = Stake . VMap.fromMap <$> theMap
```

libs/cardano-ledger-test/bench/Bench/Cardano/Ledger/StakeDistr.hs
```diff
@@ -0,0 +1,388 @@
...
+import Data.Maybe.Strict (StrictMaybe (..))
+import Data.Strings (lazyBytes)
+import Data.Text (pack)
+import qualified Data.UMap as UM
+import GHC.IO.Exception (IOException (..))
+import Test.Cardano.Ledger.Tickf (oldCalculatePoolDistr)
+
+-- ======================================================================
+
+type CurrentEra = AlonzoEra StandardCrypto
```
lehins, line 75: Maybe we should switch it to `Babbage`? This won't change the outcome of the benchmarks, but maybe it will look more accurate :)

libs/cardano-ledger-test/bench/Bench/Cardano/Ledger/StakeDistr.hs
```diff
@@ -0,0 +1,388 @@
...
+
+-- ==============================
+-- Data files
+
+-- | A filepath to a CBOR file that can be decoded to a NewEpochState.
+--   If the file does not  exist at this path the benchmark will run, but be meaningless.
+--   See the README.md file in ledger-state/libs/ledger-state/README.md for instructions for
+--   how one might obtain such a file. Get your own copy and adjust the path here.
+newepochFilePath :: String
+newepochFilePath = "/home/sheard/Two/Data/71452796-NewEpochState.cbor"
```
lehins: Well, that path is definitely not reproducable for all platforms :smile: 

I suggest you use the same approach of reading an evironment variable as in `ledger-state`: 
https://github.com/input-output-hk/cardano-ledger/blob/8817304c300dd60099054dd66d9a1ecbd33bfb3a/libs/ledger-state/bench/Performance.hs#L41-L43

libs/cardano-ledger-test/bench/Bench/Cardano/Ledger/StakeDistr.hs
```diff
@@ -0,0 +1,388 @@
...
+        fromCBOR
+        lazybytes of
+        Left err -> error (show err)
+        Right nes -> pure nes
+    Left e -> do
+      putStrLn $ "The path to the NewEpochState CBOR file\n   '" ++ newepochFilePath ++ "' is not readable."
+      putStrLn $ "Description of error\n   " ++ show (ioe_description e)
+      putStrLn $ "The error no.\n  " ++ show (ioe_errno e)
+      putStrLn $ "The benchmark results are meaningless."
+      pure bogusNewEpochState
```
lehins: Here is where you could use the env var (note that `System.Environment` import is needed and there is no longer need for usage `try`):

```suggestion
readNewEpochState = do
  let ledgerVarName = "BENCH_LEDGER_STATE_PATH"
  ledgerStateFilePath <- lookupEnv ledgerVarName
  case ledgerStateFilePath of
    Just newEpochFilePath -> do
      lazyBytes <- Lazy.readFile newepochFilePath
      let lbl = "NewEpochState from " <> pack newepochFilePath
      case decodeFullDecoder @(NewEpochState CurrentEra) lbl fromCBOR lazyBytes of
          Left err -> error (show err)
          Right nes -> pure nes
    Nothing -> bogusNewEpochState <$ do
      putStrLn $ unlines [ "No path to the NewEpochState CBOR was provided",
                                      "The benchmark results are meaningless."]
```

libs/cardano-ledger-test/bench/Bench/Cardano/Ledger/StakeDistr.hs
```diff
@@ -0,0 +1,388 @@
...
+    (_, fails) -> error (show fails)
+
+mirR :: Globals -> EpochState CurrentEra -> EpochState CurrentEra
+mirR globals es' =
+  case flip
+    runReader
+    globals
+    (applySTS @(ShelleyMIR CurrentEra) (TRC ((), es', ()))) of
+    Left fails -> error (show fails)
+    Right x -> x
```
lehins: I think it would be good to reduce duplication in this module by adding a helper function of sorts:

```haskell
-- | Run a reader transformer that will throw an error on a `Left` result.
runReaderPartial :: Show a => env -> Reader env (Either a b) -> b 
runReaderPartial env action =
  case runReader action env of
    Left fails -> error (show fails)
    Right x -> x
```

libs/cardano-ledger-test/bench/Bench/Cardano/Ledger/StakeDistr.hs
```diff
@@ -0,0 +1,388 @@
...
+
+instance Semigroup a => Semigroup (StrictMaybe a) where
+  SNothing <> x = x
+  x <> SNothing = x
+  (SJust x) <> (SJust y) = SJust (x <> y)
+
+instance Semigroup a => Monoid (StrictMaybe a) where
+  mempty = SNothing
```
lehins: Merging with these instances will create work for someone else down the road. I suggest we first merge this PR: https://github.com/input-output-hk/cardano-base/pull/314 and use instances from newer `cardano-strict-containers`

I can help with making a new release of cardano-strict-containers` and using it here, once we get the above PR merged

libs/cardano-ledger-test/bench/Bench/Cardano/Ledger/StakeDistr.hs
```diff
@@ -0,0 +1,388 @@
...
+tickfRuleBench :: Benchmark
+tickfRuleBench =
+  env
+    readGlobals
+    ( \globals ->
+        env
+          readNewEpochState
+          ( \nes ->
+              bgroup
+                "Tickf Benchmarks"
```
lehins: Ormolized formatting will be cleaner if we do this instead
```suggestion
  env readGlobals $ \globals ->
     env readNewEpochState $ \nes ->
       bgroup
          "Tickf Benchmarks"
```

libs/cardano-ledger-test/bench/Bench/Cardano/Ledger/StakeDistr.hs
```diff
@@ -0,0 +1,388 @@
...
+                    "Tick subparts"
+                    [ bench "adoptGenesisDelegs" $ whnf (adoptGenesisDelegsR (SlotNo 156953303)) nes,
+                      bench "newEpoch" $ whnf (newEpochR globals (nesEL nes + 1)) nes,
+                      bgroup
+                        "NewEpoch sub parts"
+                        [ bench "updateRewards" $ whnf (updateRewardsX globals) nes,
+                          bgroup
+                            "updateRewards subparts"
+                            [ bench "sumRewards" $ whnf (sumRewards (esPp (nesEs nes))) (rs (getRewardUpdate nes)),
+                              bench "applyRUpd' Incremental" $ whnf (applyRUpd' (getRewardUpdate nes)) (nesEs nes),
```
lehins: Do you want to benchmark `applyRUpd'` or `applyRUpd`? Because `FilteredRewards` are lazy which means you need to use `nf` here instead of `whnf` in order to get the full result benchmarked.

On the other hand we can just switch to `applyRUpd` if only `registered` needs to be evaluated

eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState/IncrementalStake.hs
```diff
@@ -301,8 +222,120 @@ resolveActiveIncrementalPtrs isActive ptrMap_ (IStake credStake ptrStake) =
...
+      nm' = nonMyopic ru
+
+-- | A data structure aggregating different information about the rewards mapping
+data Filtered era = Filtered
+  { -- Only the first component is strict on purpose
+
+    -- | These are registered, but in the ShelleyEra they are ignored because of backward compatibility
+    --  in other Eras, this field will be the Map.empty
+    registeredF :: !(Map (Credential 'Staking (EraCrypto era)) (Set (Reward (EraCrypto era)))),
+    shelleyIgnoredF :: Map (Credential 'Staking (EraCrypto era)) (Set (Reward (EraCrypto era))),
```
lehins: I'd like to suggest a comment here why theses fields are kept lazy. I can easily see someone down the road trying to "optimize" things by making these fields strict

libs/cardano-ledger-test/bench/Bench/Cardano/Ledger/StakeDistr.hs
```diff
@@ -0,0 +1,388 @@
...
+                                    ( \ ~(Filtered registered _ _ _) ->
+                                        bench "aggregateRewards" $ whnf (aggregateRewards (esPp (nesEs nes))) registered
+                                    ),
+                                  env
+                                    ( pure
+                                        ( aggregateRewards
+                                            (esPp (nesEs nes))
+                                            ( (\ ~(Filtered a _ _ _) -> a)
+                                                (filterAllRewards (rs (getRewardUpdate nes)) (nesEs nes))
+                                            )
```
lehins:
```suggestion
                                            ( frRegistered $
                                                filterAllRewards (rs (getRewardUpdate nes)) (nesEs nes)
                                            )
```

libs/cardano-ledger-test/bench/Bench/Cardano/Ledger/StakeDistr.hs
```diff
@@ -0,0 +1,388 @@
...
+                            [ bench "sumRewards" $ whnf (sumRewards (esPp (nesEs nes))) (rs (getRewardUpdate nes)),
+                              bench "applyRUpd' Incremental" $ whnf (applyRUpd' (getRewardUpdate nes)) (nesEs nes),
+                              bgroup
+                                "applyRUpd subparts"
+                                [ bench "filterAllRewards" $ whnf (filterAllRewards (rs (getRewardUpdate nes))) (nesEs nes),
+                                  env
+                                    (pure (filterAllRewards (rs (getRewardUpdate nes)) (nesEs nes)))
+                                    ( \ ~(Filtered registered _ _ _) ->
+                                        bench "aggregateRewards" $ whnf (aggregateRewards (esPp (nesEs nes))) registered
+                                    ),
```
lehins:
```suggestion
                                    (bench "aggregateRewards" . whnf (aggregateRewards (esPp (nesEs nes))) . frRegistered),
```

libs/cardano-ledger-test/bench/Bench/Cardano/Ledger/StakeDistr.hs
```diff
@@ -0,0 +1,388 @@
...
+instance NFData (RUArgs (AlonzoEra StandardCrypto)) where
+  rnf (RUArgs a b c) = seq (rnf a) $ seq (rnf b) (rnf c)
+
+-- ==================================================================
+
+-- | The inital ShelleyGenesis structure, encoded in json format as a Haskell String
+shelleyGenesis :: ShelleyGenesis CurrentEra
+shelleyGenesis =
+  fromJust $
+    Aeson.decode $
```
lehins: `fromJust` should be banned. It is impossible to find which occurrence of it in the codebase is at fault when it fails. There is always a better way: 
```suggestion
  either error id $
    Aeson.eitherDecode $
```

libs/cardano-ledger-test/bench/Bench/Cardano/Ledger/StakeDistr.hs
```diff
@@ -0,0 +1,388 @@
...
+                ]
+          )
+    )
+
+getSnap :: NewEpochState (AlonzoEra StandardCrypto) -> SnapShot StandardCrypto
+getSnap nes = (_pstakeSet . esSnapshots . nesEs) nes
+
+getRewardUpdate :: NewEpochState era -> RewardUpdate (EraCrypto era)
+getRewardUpdate nes =
+  case (nesRu nes) of
```
lehins: Redundant parenthesis
```suggestion
  case nesRu nes of
```

libs/cardano-ledger-test/bench/Bench/Cardano/Ledger/StakeDistr.hs
```diff
@@ -0,0 +1,388 @@
...
+
+decodeRUArgs :: Decode ('Closed 'Dense) (RUArgs (AlonzoEra StandardCrypto))
+decodeRUArgs = RecD RUArgs <! From <! From <! From
+
+instance ToCBOR (RUArgs (AlonzoEra StandardCrypto)) where toCBOR x = encode (encodeRUArgs x)
+
+instance FromCBOR (RUArgs (AlonzoEra StandardCrypto)) where fromCBOR = decode decodeRUArgs
+
+instance NFData (RUArgs (AlonzoEra StandardCrypto)) where
+  rnf (RUArgs a b c) = seq (rnf a) $ seq (rnf b) (rnf c)
```
lehins: I can't seem to find where these `RUArgs` are used. Is this dead code?

libs/cardano-ledger-test/bench/Bench/Cardano/Ledger/StakeDistr.hs
```diff
@@ -0,0 +1,388 @@
...
+
+getSnap :: NewEpochState (AlonzoEra StandardCrypto) -> SnapShot StandardCrypto
+getSnap nes = (_pstakeSet . esSnapshots . nesEs) nes
+
+getRewardUpdate :: NewEpochState era -> RewardUpdate (EraCrypto era)
+getRewardUpdate nes =
+  case (nesRu nes) of
+    SNothing -> error "No RewardUpdate in new epoch state"
+    SJust (Pulsing _ _) -> error "RewardUpdate in new epoch state is pulsing"
+    SJust (Complete ru) -> ru
```
lehins: Extract this into separate function and reuse in `updateRewardsX`:
```suggestion
    SJust c -> getComplete c
    
getComplete = \case ->
    Pulsing _ _ -> error "RewardUpdate in new epoch state is pulsing"
    Complete ru -> ru
```

libs/cardano-ledger-test/bench/Bench/Cardano/Ledger/StakeDistr.hs
```diff
@@ -0,0 +1,388 @@
...
+      epochstate = nesEs newepochstate
+   in case flip
+        runReader
+        globals
+        ( runRule
+            (TRC ((), newepochstate, epochNo))
+            ( case nesRu newepochstate of
+                SNothing -> pure epochstate
+                SJust (Complete ru) -> updateRewards epochstate epochNo ru
+                SJust (Pulsing _ _) -> error "Update is still pulsing"
```
lehins:
```suggestion
                SJust p -> updateRewards epochstate epochNo $ getComplete p
```

libs/cardano-ledger-test/bench/Bench/Cardano/Ledger/StakeDistr.hs
```diff
@@ -0,0 +1,388 @@
...
+-- Functions based on Rules we can Benchmark
+
+updateRewardsX :: Globals -> NewEpochState CurrentEra -> EpochState CurrentEra
+updateRewardsX globals newepochstate =
+  let epochNo = nesEL newepochstate
+      epochstate = nesEs newepochstate
+   in case flip
+        runReader
+        globals
+        ( runRule
```
lehins: This pattern is used twice. Also warrants a helper function

TimSheard:
So I added some notes to https://github.com/input-output-hk/cardano-ledger/issues/3035. Did you make some changes. I note that you did a force-push .  Do I need to do anything else? (like rebase on master?)

libs/cardano-ledger-test/bench/Bench/Cardano/Ledger/StakeDistr.hs
```diff
@@ -0,0 +1,388 @@
...
+  rnf (RUArgs a b c) = seq (rnf a) $ seq (rnf b) (rnf c)
+
+-- ==================================================================
+
+-- | The inital ShelleyGenesis structure, encoded in json format as a Haskell String
+shelleyGenesis :: ShelleyGenesis CurrentEra
+shelleyGenesis =
+  fromJust $
+    Aeson.decode $
+      lazyBytes
```
lehins: This is unnecessary. Just enable `OverloadedStrings` function and it will work just fine. If in fact that string actually contained unicode characters, then this is as easy as `Text.encodeUtf8 . Text.pack`, where both functions come from `text` package.
```suggestion
```

libs/cardano-ledger-test/cardano-ledger-test.cabal
```diff
@@ -166,10 +170,14 @@ benchmark bench
     QuickCheck,
     cardano-ledger-shelley,
     cardano-ledger-shelley-test,
+    cardano-slotting,
     microlens,
+    mtl,
     random,
     small-steps,
     small-steps-test,
+    strings,
```
lehins: This package is really outdated and is unneeded

lehins (changes requested):
Looks great, but few things need fixing though.

libs/cardano-ledger-test/bench/Bench/Cardano/Ledger/StakeDistr.hs
TimSheard: My choice of Alonzo was indicated because the slot that we got the NewEpochState from was in the Alonzo era.
Because the code we benchmark runs the same in every era, it probably doesn't matter.

libs/cardano-ledger-test/bench/Bench/Cardano/Ledger/StakeDistr.hs
TimSheard: Not sure what pattern we are taking about.

libs/cardano-ledger-test/bench/Bench/Cardano/Ledger/StakeDistr.hs
TimSheard: good idea

libs/cardano-ledger-test/bench/Bench/Cardano/Ledger/StakeDistr.hs
TimSheard: I am not sure the code will compile without these instances.

libs/cardano-ledger-test/bench/Bench/Cardano/Ledger/StakeDistr.hs
TimSheard: I am pretty sure some version of ormlou formatted the code that way.

libs/cardano-ledger-test/bench/Bench/Cardano/Ledger/StakeDistr.hs
TimSheard: Not sure I understand this. I will think about it.

libs/cardano-ledger-test/bench/Bench/Cardano/Ledger/StakeDistr.hs
TimSheard: it might be. I will check.

libs/cardano-ledger-test/bench/Bench/Cardano/Ledger/StakeDistr.hs
lehins: This one. It is also used in `tickfR2`
```haskell
case flip  runReader globals action of
 (es, []) -> es
 (_, fails) -> error (show fails)
```

lehins: I've learned it long ago. If I am doing a copy and paste of some chunk of code, especially in the same module, I will spend an extra moment to figure out how can I abstract away the common part in order to make it more readable and reduce future duplication

libs/cardano-ledger-test/bench/Bench/Cardano/Ledger/StakeDistr.hs
lehins: That's my point we need to import this instances from `cardano-base`, rather then redefine them again here. You submitted a PR that adds those instances to `cardano-base` and what I am saying is that we should use those instances here.

libs/cardano-ledger-test/bench/Bench/Cardano/Ledger/StakeDistr.hs
lehins: Just try and apply my suggestion and you'll see the difference.

libs/cardano-ledger-test/bench/Bench/Cardano/Ledger/StakeDistr.hs
TimSheard: Has that PR been merged?

lehins: Not yet, there are comments that needs to be addressed.

eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState/IncrementalStake.hs
TimSheard: done, it cleans things up nicely

eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState/IncrementalStake.hs
TimSheard: done

eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState/IncrementalStake.hs
TimSheard: done

eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState/IncrementalStake.hs
TimSheard: done

eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState/IncrementalStake.hs
TimSheard: done

eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState/IncrementalStake.hs
TimSheard: done

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/NewEpoch.hs
TimSheard: done

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/NewEpoch.hs
TimSheard: I have swittched, so the info is there. I won't add something to the event on speculation.

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/NewEpoch.hs
TimSheard: the toInteger is fine, but what we are summing is Word64, so embeddig it in a CompactCoin  is just confusing.

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/NewEpoch.hs
TimSheard: don't agree

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/NewEpoch.hs
TimSheard: done

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/NewEpoch.hs
TimSheard: done

libs/cardano-ledger-test/bench/Bench/Cardano/Ledger/StakeDistr.hs
TimSheard: done

libs/cardano-ledger-test/bench/Bench/Cardano/Ledger/StakeDistr.hs
TimSheard: You are write, some leftover thing from some attempt at something I forget.

libs/cardano-ledger-test/bench/Bench/Cardano/Ledger/StakeDistr.hs
TimSheard: done

libs/cardano-ledger-test/cardano-ledger-test.cabal
TimSheard: removed

libs/cardano-ledger-test/bench/Bench/Cardano/Ledger/StakeDistr.hs
TimSheard: OK I understand now. I tried it both ways, and it doesn't seem to make much of a difference.

```
bench "applyRUpdFiltered Incremental" $ whnf (applyRUpdFiltered (getRewardUpdate nes)) (nesEs nes),
benchmarking Tickf Benchmarks/Tick subparts/NewEpoch sub parts/updateRewards subparts/applyRUpdFiltered Incremental
time                 919.5 ms   (869.5 ms .. 955.5 ms)
                     1.000 R²   (0.999 R² .. 1.000 R²)
mean                 931.6 ms   (922.1 ms .. 939.4 ms)
std dev              9.524 ms   (3.750 ms .. 11.97 ms)
variance introduced by outliers: 19% (moderately inflated)

bench "applyRUpdFiltered Incremental" $ nf (applyRUpdFiltered (getRewardUpdate nes)) (nesEs nes),
benchmarking Tickf Benchmarks/Tick subparts/NewEpoch sub parts/updateRewards subparts/applyRUpdFiltered Incremental
time                 1.010 s    (938.1 ms .. 1.051 s)
                     0.999 R²   (0.999 R² .. 1.000 R²)
mean                 997.1 ms   (974.8 ms .. 1.007 s)
std dev              16.70 ms   (1.813 ms .. 21.16 ms)
variance introduced by outliers: 19% (moderately inflated)
```

TimSheard: I changed it to be 'nf'

libs/cardano-ledger-test/bench/Bench/Cardano/Ledger/StakeDistr.hs
TimSheard: So Ieft the instances here for now, so I could compile and run. I will address the comments after I finish these comments.

libs/cardano-ledger-test/bench/Bench/Cardano/Ledger/StakeDistr.hs
TimSheard: done

lehins (approved):

PR #3069, opened 2022-10-10
Move to get Haskell packages from CHaP

michaelpj:
This moves the `cardano-ledger` repository to use CHaP, murdering most of the `source-repository-package`s in the process.

I tried to explain all the process differences, please let me know if it's clear enough!

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -12,9 +12,49 @@ We use tags on the release branches to indicate patches, of the form `ledger/a.b
 We also use tags to indicate what version of the ledger was used in
 cardano-node releases, of the form `node/a.b.c` (possible with a `rc` or `rc1`, etc).
 
+### Releasing the ledger packages to CHaP
+
+When the ledger packages are released they should be released to [CHaP](https://github.com/input-output-hk/cardano-haskell-packages).
+See the CHaP README for [instructions](https://github.com/input-output-hk/cardano-haskell-packages#-from-github).
+
 ## Building
 
 See the [Readme](https://github.com/input-output-hk/cardano-ledger#building) for instructions on building.
+
+## Updating dependencies
+
+Our Haskell packages come from two package repositories:
+- Hackage
+- [CHaP](https://github.com/input-output-hk/cardano-haskell-packages) (which is essentially another Hackage)
+
+The "index state" of each repository is pinned to a particular time in `cabal.project`.
+This tells Cabal to treat the repository "as if" it was the specified time, ensuring reproducibility.
+If you want to use a package version from repository X which was added after the pinned index state time, you need to bump the index state for X.
+This is not a big deal, since all it does is change what packages `cabal` considers to be available when doing solving, but it will change what package versions cabal picks for the plan, and so will likely result in significant recompilation, and potentially some breakage.
+That typically just means that we need to fix the breakage (and add a lower-bound on the problematic package), or add an upper-bound on the problematic package.
+
+Note that `cabal` itself keeps track of what index states it knows about, so when you bump the pinned index state you may need call `cabal update` in order for `cabal` to be happy.
+
+The Nix code which builds our packages also cares about the index state.
+This is represented by inputs managed by `niv`:
+You can update these by running:
+- `niv update hackage.nix` for Hackage
+- `niv update cardano-haskell-packages` for CHaP
+
+If you fail to do this you may get an error like this from Nix:
+```
+error: Unknown index-state 2021-08-08T00:00:00Z, the latest index-state I know about is 2021-08-06T00:00:00Z. You may need to update to a newer hackage.nix.
+```
+
+### Use of `source-repository-package`s
+
+We *can* use Cabal's `source-repository-package` mechanism to pull in un-released package versions.
+However, we should try and avoid this.
+In particular, we should not release our packages to CHaP while we depend on a `source-repository-package`.
+
+If we are stuck in a situation where we need a long-running fork of a package, we should release it to CHaP instead (see the https://github.com/input-output-hk/cardano-haskell-packages[CHaP README] for more).
+
+If you do add a `source-repository-package`, you need to provide a `--sha256` comment in `cabal.project` so that Nix knows the hash of the content.
 
 ## Warnings
 
--- a/cabal.project
+++ b/cabal.project
@@ -1,4 +1,21 @@
+-- Custom repository for cardano haskell packages, see CONTRIBUTING for more
+repository cardano-haskell-packages
+  url: https://input-output-hk.github.io/cardano-haskell-packages
+  secure: True
+  root-keys:
+    3e0cce471cf09815f930210f7827266fd09045445d65923e6d0238a6cd15126f
+    443abb7fb497a134c343faf52f0b659bd7999bc06b7f63fa76dc99d631f9bea1
+    a86a1f6ce86c449c46666bda44268677abf29b5b2d2eb5ec7af903ec2f117a82
+    bcec67e8e99cabfa7764d75ad9b158d72bfacf70ca1d0ec8bc6b4406d1bf8413
+    c00aae8461a256275598500ea0e187588c35a5d5d7454fb57eac18d9edb86a56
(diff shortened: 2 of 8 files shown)


michaelpj:
Okay, so it's unhappy building `moo-1.2`, since your GHA CI is using cabal 3.8 which doesn't like the duplicate module name. That makes sense... I'm just surprised it's only happening now, since that incompatibility has been there for a while!

michaelpj:
I'm just going to go ahead and purge goblins and save everyone some pain.

JaredCorduan:
> I'm just going to go ahead and purge goblins and save everyone some pain.

Thank you!

lehins:
Also use this `moo`: 

```
source-repository-package
  type: git
  location: https://github.com/input-output-hk/moo
  tag: 8c487714fbfdea66188fcb85053e7e292e0cc348
  --sha256: sha256-HWxnkf6xEAFfY14QxQFU/RsQ09skzLSMKvrAB1EQstU=
```

I have no idea why it started complaining about it only recently.

https://github.com/input-output-hk/cardano-ledger/pull/3063/files#diff-b8ed06757045fac949c89f2139a862498ad2b6d1f82c61a31e7c91d6cf0eaa70R121-R125

michaelpj:
Thanks @lehins , I'll update this PR with that until I kill goblins.

cabal.project
```diff
@@ -44,43 +61,25 @@ benchmarks: true
...
-source-repository-package
-  type: git
-  location: https://github.com/input-output-hk/cardano-crypto
-  tag: f73079303f663e028288f9f4a9e08bcca39a923e
-  --sha256: 1n87i15x54s0cjkh3nsxs4r1x016cdw1fypwmr68936n3xxsjn6q
+constraints:
+  -- bizarre issue: in earlier versions they define their own 'GEq', in newer
+  -- ones they reuse the one from 'some', but there isn't e.g. a proper version
+  -- constraint from dependent-sum-template (which is the library we actually use).
+  , dependent-sum > 0.6.2.0
```
andreabedini, line 73: these constraints are here rather than in the cabal file? easier transition?

cabal.project
```diff
@@ -44,43 +61,25 @@ benchmarks: true
...
-    cardano-prelude-test
+  -- Plutus dependency
+  , algebraic-graphs >= 0.7
 
-source-repository-package
-  type: git
-  location: https://github.com/input-output-hk/goblins
-  tag: cde90a2b27f79187ca8310b6549331e59595e7ba
-  --sha256: 17c88rbva3iw82yg9srlxjv2ia5wjb9cyqw44hik565f5v9svnyg
+  , protolude >= 0.3.2
```
andreabedini: why are these constraints here rather than in the cabal file? easier transition?

andreabedini:
FWIW cabal 3.8.1.0 fails to compile moo but cabal 3.6.2.0 succeeds. So one could do

```
$ cabal-3.6.2.0 repl --build-depends moo
```

to populate the cabal store. :shrug:

cabal.project
michaelpj: They're for packages not in this repository...

michaelpj:
Some of the project file constraints can be tidied up, I'll do that in a follow up to avoid conflicts.

michaelpj:
Just wanted to register that this _slightly_ changes the packages that are depended on. In particular, you're now using the `cardano-base` that was published as `2.0.0.1`, rather than the precise sha you had before. Hopefully that's fine - in future if you need more changes we should just release them from `cardano-base`.

michaelpj:
Also @JaredCorduan please let me know if the information about releasing packages to CHaP makes sense to you - let's not merge this until everyone is happy with the new workflow!

michaelpj:
(rebased on goblins-removal, so hopefully this will be green now)

michaelpj:
Was missing the just-added s-r-p for `cborg`, restored.

JaredCorduan:
> Also @JaredCorduan please let me know if the information about releasing packages to CHaP makes sense to you - let's not merge this until everyone is happy with the new workflow!

thanks @michaelpj ! I haven't had time to look over it all yet, but I will try to make time soon. The overall goal is glorious! I've added the whole team no the reviewers, so everyone has an excuse to get familiar.

CONTRIBUTING.md
```diff
@@ -12,10 +12,50 @@ We use tags on the release branches to indicate patches, of the form `ledger/a.b
...
+If you want to use a package version from repository X which was added after the pinned index state time, you need to bump the index state for X.
+This is not a big deal, since all it does is change what packages `cabal` considers to be available when doing solving, but it will change what package versions cabal picks for the plan, and so will likely result in significant recompilation, and potentially some breakage.
+That typically just means that we need to fix the breakage (and add a lower-bound on the problematic package), or add an upper-bound on the problematic package.
+
+Note that `cabal` itself keeps track of what index states it knows about, so when you bump the pinned index state you may need call `cabal update` in order for `cabal` to be happy.
+
+The Nix code which builds our packages also cares about the index state.
+This is represented by inputs managed by `niv`:
+You can update these by running:
+- `niv upate hackage.nix` for Hackage
```
aniketd:
```suggestion
- `niv update hackage.nix` for Hackage
```

teodanciu:
@michaelpj - thank you also for the generous explanations about the index state  and niv. 

Something that is not clear to me:  (sorry if it's explained and I missed it, or maybe if it's too trivial  :grimacing: ) 

In this new version,  what is the way of knowing what commit of a certain dependency  is used?  For example: heapwords.  I expect we're looking at the CHaP commit specified  in `sources.json`, so in this case: 
https://github.com/input-output-hk/cardano-haskell-packages/commits/d4af732a53321a1d1760023a2fa5b3f656cc3bfd
Then the answer is  in /_sources/heapwords/0.1.0.1/meta.toml ?  Could there be several versions of this though? 

And  the same in reverse: if you needed to use a certain hash of a depenency, how would you go about that?

michaelpj:
> In this new version, what is the way of knowing what commit of a certain dependency is used?

So, the idea is that mostly you shouldn't need to know commit hashes: versions should be specific enough (like on Hackage). But several people have asked about this so we'll probably do something to make it easier regardless :) For now, the answer is that you can go to CHaP and look at the metadata file specifying the package version, exactly the one you found.

You do need to be able to work out what _version_ of the package you're using. This is now a general cabal problem, you have the same problem if you want to know what version of `containers` you're using. The simplest way I know of is just to run `cabal freeze` and take a look at the freeze file.

> And the same in reverse: if you needed to use a certain hash of a depenency, how would you go about that?

I think the question is: why do you need to do this? Generally we want to get to a place where we're mainly using released versions of software. But if you do need to you can always use `source-repository-package`... just try not to leave it like that, otherwise we end up in a situation where we can't release packages because they depend on unreleased random commits of their dependencies (aka, the historical status quo :sweat_smile: ).

-----

I'm gradually trying to answer questions in the CHaP readme. I'll make sure these answers get captured!

aniketd (commented):
Thanks for the explanations :slightly_smiling_face:

lehins (approved):

teodanciu (approved):
Looks good to me

JaredCorduan (approved):

CONTRIBUTING.md
```diff
@@ -12,10 +12,50 @@ We use tags on the release branches to indicate patches, of the form `ledger/a.b
...
+
+Note that `cabal` itself keeps track of what index states it knows about, so when you bump the pinned index state you may need call `cabal update` in order for `cabal` to be happy.
+
+The Nix code which builds our packages also cares about the index state.
+This is represented by inputs managed by `niv`:
+You can update these by running:
+- `niv update hackage.nix` for Hackage
+- `niv update cardano-haskell-packages` for CHaP
+
+iF you fail to do this you may get an error like this from Nix:
```
andreabedini:
```suggestion
If you fail to do this you may get an error like this from Nix:
```

Soupstraw (approved):

andreabedini (approved):

andreabedini:
🎉🎉🎉

PR #3070, opened 2022-10-10
Purge `goblins` from the ledger

michaelpj:
This reduces the amount of testing of Byron, in particular it removes invalid trace generation.

I've tried to remove only the things that strictly relied on `goblins`, so in fact _some_ of the invalid trace generation survives! Please take a careful look in case I've been too zealous.


lehins (approved):
LGTM

JaredCorduan (approved):
:no_entry_sign: :japanese_ogre:

lehins:
@michaelpj Looks like me upgrading ormolu created a conflict

PR #3075, opened 2022-10-12
Simplify CostModel parameter representation 

### LINKED ISSUE #3042: Simplify CostModel internals
The `CostModel` type contains a `Map Text Integer` named `cmMap`:

https://github.com/input-output-hk/cardano-ledger/blob/3f5123c3ae009ec1ac2779410da6604a236e67f2/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Scripts.hs#L231-L235

Change the type of `cm` to be `[Integer]`.

---

## Context for those curious

This map `cmMap` is _only_ used by the ledger to create a Plutus evaluation context (of type `EvaluationContext`), and it is only stored in the ledger so that we can serialize it (deserializing it allows us to construct the `EvaluationContext` again).

The newest version of Plutus now follows [this plan](https://github.com/input-output-hk/cardano-ledger/issues/2902#issuecomment-1183644628), and hence only needs `[Integer]` to be constructed (ie no longer a `Map Text Integer`).

Moreover, there is no longer any need to use the `Text` keys in the ledger. So `costModelParamsNames` and `costModelParamsNamesSet` can be deleted as a part of this change. (introduced here https://github.com/input-output-hk/cardano-ledger/pull/2992/files#diff-f481d9eb9f793235d9fc4a1965ec1e0a01bed313f90adb11dcd9aa099a4cd738R320-R326)

teodanciu:
In the new version of Plutus API, we are passing costmodel parameter values as a list, instead of a map indexed by parameter names. So there is no need to keep the parameter names in the model .

Closes https://github.com/input-output-hk/cardano-ledger/issues/3042

### DIFF
--- a/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Genesis.hs
+++ b/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Genesis.hs
@@ -212,7 +212,7 @@ instance FromJSONKey Language where
   fromJSONKey = Aeson.FromJSONKeyTextParser languageFromText
 
 validateCostModel :: MonadFail m => Language -> Map Text Integer -> m (Language, CostModel)
-validateCostModel lang cmps = case mkCostModel lang cmps of
+validateCostModel lang cmps = case mkCostModel lang (Map.elems cmps) of
   Left err -> fail $ show err
   Right cm -> pure (lang, cm)
 
--- a/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Genesis.hs
+++ b/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Genesis.hs
@@ -27,7 +27,6 @@ import Cardano.Ledger.Alonzo.Scripts
     ExUnits (..),
     ExUnits',
     Prices (..),
-    costModelParamsNames,
     getCostModelParams,
     mkCostModel,
   )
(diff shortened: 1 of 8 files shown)


JaredCorduan (approved):
this is perfect, thank you @teodanciu ! Let's update the Changelog, though.

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Genesis.hs
```diff
@@ -211,27 +216,31 @@ instance ToJSONKey Language where
...
-translateLegacyV1paramNames :: Map Text Integer -> Map Text Integer
-translateLegacyV1paramNames cmps =
-  Map.fromList $ zip (costModelParamsNames PlutusV1) (Map.elems cmps)
+cmParamValues :: Language -> Map Text Integer -> [Integer]
+cmParamValues lang cmMap =
+  case lang of
+    PlutusV1 -> filterPreservingOrder $ paramNames (enumerate @PV1.ParamName)
+    PlutusV2 -> filterPreservingOrder $ paramNames (enumerate @PV2.ParamName)
+  where
+    filterPreservingOrder = mapMaybe (`Map.lookup` cmMap)
```
teodanciu, line 230: Should add some explanations here for this acrobatics

PR #3079, opened 2022-10-13
Update haskell.nix sources

teodanciu:
which fixes the coverageReport derivation build failure


JaredCorduan (approved):
:tada: thank you @teodanciu !

PR #3089, opened 2022-10-18
Remove unused `Data.BiMap` module from cardano-data

### LINKED ISSUE #3086: Remove `BiMap` implementation
Looks like `BiMap` is no longer used anyhere in the codebase.  Instead of continuing maintaining that data structure we should rip it out from `cardano-data` package. This will require removal of `BiMap` rules in the `set-algebra` package.

If it will ever happen that we need it again, we can refer to this ticket and bring it back.

teodanciu:
Closes https://github.com/input-output-hk/cardano-ledger/issues/3086


JaredCorduan (approved):
Red is the color of :heart: s. I :heart: removing dead code! This looks good to me, once we get past the CI failures. thank you!

CHANGELOG.md
```diff
@@ -83,6 +83,7 @@ in the naming of release branches.
 - Remove model test framework #3019
 - The `Cardano.Ledger.Alonzo.Scripts` module no longer re-exports the
   `plutus-ledger-api`'s `assertWellFormedCostModelParams`. #3065
+- Removed unused `Data.BiMap` module from `cardano-data #3081
```
JaredCorduan:
```suggestion
- Removed unused `Data.BiMap` module from `cardano-data #3089
```

libs/set-algebra/src/Control/Iterate/BaseTypes.hs
```diff
@@ -248,46 +245,11 @@ instance Iter Sett where
...
-  isnull (MkBiMap f _g) = isnull f
-  lookup x (MkBiMap left _right) = Map.lookup x left
-  haskey k (MkBiMap left _right) = haskey k left
-
 -- ============== Basic Map =========================
 
 instance Basic Map.Map where
-  addkv (k, v) m comb = Map.insertWith (mapflip comb) k v m
+  -- Data.Map uses(\ new old -> ...) while our convention is (\ old new -> ...)
+  addkv (k, v) m comb = Map.insertWith (flip comb) k v m
```
teodanciu, line 252: @JaredCorduan  My only concern in this PR is this - so basically this `mapflip` defined in the BiMap was used also in the implementation of the map. Here is the implementation: https://github.com/input-output-hk/cardano-ledger/pull/3089/commits/e38e4048065c54a639593001b1ec66cc5ca6232b#diff-528417e48d8c20cf8e058ea8ec356378763095d96fc6841d76c65415772255c3L145

It seems to me that it's a simple `flip`, but would you mind double checking my logic?

JaredCorduan: `mapflip` looks like a less general version of `flip` that doesn't have anything to do with maps :laughing: . LGTM!

teodanciu: Thanks for confirming!

libs/cardano-data/src/Data/UMap.hs
```diff
@@ -11,9 +11,9 @@
 {-# LANGUAGE TypeFamilies #-}
 {-# LANGUAGE ViewPatterns #-}
 
--- | A 'UMap' (for Unified map) represents 2 'Data.Map's and one 'Data.BiMap' into a single structure.
+-- | A 'UMap' (for Unified map) represents 2 'Data.Map's into a single structure.
```
teodanciu: This comment is odd.. The reference to the BiMap is only in the comment, not in code.. the definition is just this:

```data UMap coin cred pool ptr = UnifiedMap !(Map cred (Trip coin ptr pool)) !(Map ptr cred)```

I thought maybe the comment went out of sync with the code, but looking at the history, it seems to have always had this definition. So i guess the comment is just inaccurate, unless it has some other meaning that I don't see

Comment (JaredCorduan) [reply libs/cardano-data/src/Data/UMap.hs]: let's ask @TimSheard

File: CHANGELOG.md [hunk 4]
```diff
@@ -83,6 +83,7 @@ in the naming of release branches.
 - Remove model test framework #3019
 - The `Cardano.Ledger.Alonzo.Scripts` module no longer re-exports the
   `plutus-ledger-api`'s `assertWellFormedCostModelParams`. #3065
+- Removed unused `Data.BiMap` module from `cardano-data #3089
```
```suggestion
- Removed unused `Data.BiMap` module from `cardano-data` #3089
```

lehins (approved):
Looks great!

libs/cardano-data/src/Data/UMap.hs
lehins: As I suggested during the meeting, we should change the comment to something along the lines:

```suggestion
-- | A 'UMap' (for Unified map) represents three Maps of the same size in one direction with @cred@ for 
-- keys and one more in the inverse direction with @ptr@ for keys and @cred@ for values.
```

CHANGELOG.md
JaredCorduan: uh oh, we've "ticked" @lehins off. :drum:

teodanciu: good catch, it would have annoyed me to no end to see if after i merged, thanks

PR #3095, opened 2022-10-20
Added a law and a test to the `Twiddler` class

Soupstraw:
This PR adds a test for the `Twiddle` law. A correct `Twiddle` implementation should perserve all data when twiddle-encoded and then decoded.

I added `Twiddle` law test to Alonzo and Babbage eras.

I also moved the `twiddleStrictMaybe` and `emptyOrNothing` utility functions into the `Twiddle` module.

### DIFF
--- a/libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/Properties.hs
+++ b/libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/Properties.hs
@@ -318,6 +318,17 @@ adaIsPreservedInEachEpoch proof genSize =
         trcInit = _traceInitState trc
         trcLast = lastState trc
 
+twiddlerInvariantHolds ::
+  forall era.
+  ( Reflect era
+  ) => 
+  Proof era ->
+  GenSize ->
+  TestTree
+twiddlerInvariantHolds proof genSize =
+  testProperty (show proof) $
+    _
+
 -- ==============================================================
 -- Infrastrucure for running individual tests, with easy replay.
 -- In ghci just type
--- a/libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/Properties.hs
+++ b/libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/Properties.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE AllowAmbiguousTypes #-}
 {-# LANGUAGE DataKinds #-}
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE FlexibleInstances #-}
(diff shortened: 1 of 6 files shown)


JaredCorduan (approved):
:heart:

lehins (changes requested):
That's great. Thank you. I've marked this PR as a Draft. I'll fix it up and merge it after we've made the full switch to versioned CBOR implemented in cardano-ledger-binary.

Reasons we don't want to merge it now:

* Adding an orphan instance is a breaking change that propagates downstream
* Adding modules to cardano-data and changes to Twiddler in cardano-data will have to be removed and will only create conflicts
* There is no reason to maintain now two identical implementations

Soupstraw:
Alright, sounds good

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/Properties.hs
```diff
@@ -318,6 +325,25 @@ adaIsPreservedInEachEpoch proof genSize =
...
+  TestTree
+twiddleInvariantHolds proof =
+  testProperty (show proof) $ twiddleInvariantProp @a
+
+twiddleInvariantHoldsEras :: TestTree
+twiddleInvariantHoldsEras =
+  testGroup
+    "Twiddle invariant holds for TxBody"
+    [ twiddleInvariantHolds @(AlonzoTxBody (AlonzoEra Mock)) (Alonzo Mock)
+    , twiddleInvariantHolds @(BabbageTxBody (BabbageEra Mock)) (Babbage Mock)
```
lehins: You don't need actual `Proof` here, so just use an era name and also avoid using `Mock`, we should always whenever possible use `StandardCrypto`:

```suggestion
twiddleInvariantHolds ::
  forall a. (Arbitrary a, Show a, Twiddle a) =>
  String ->
  TestTree
twiddleInvariantHolds name =
  testProperty name $ twiddleInvariantProp @a

twiddleInvariantHoldsEras :: TestTree
twiddleInvariantHoldsEras =
  testGroup
    "Twiddle invariant holds for TxBody"
    [ twiddleInvariantHolds @(TxBody Alonzo) "Alonzo"
    , twiddleInvariantHolds @(TxBody Babbage) "Babbage"
```

libs/cardano-ledger-binary/test-lib/Test/Cardano/Ledger/Binary/Twiddle.hs
```diff
@@ -201,17 +204,55 @@ instance Twiddle Term where
...
+  (a -> t b) ->
+  Gen (Maybe Term)
+emptyOrNothing v x f =
+  if null $ f x
+    then
+      oneof
+        [ Just <$> twiddle @(t Void) v mempty,
+          pure Nothing
+        ]
+    else Just <$> twiddle v (f x)
```
lehins: This looks like overfunctionalization, instead this would be cleaner:

```suggestion
emptyOrNothing ::
  forall t b.
  (Foldable t, Twiddle (t Void), Monoid (t Void), Twiddle (t b)) =>
  Version ->
  t b ->
  Gen (Maybe Term)
emptyOrNothing v x =
  if null x
    then
      oneof
        [ Just <$> twiddle @(t Void) v mempty,
          pure Nothing
        ]
    else Just <$> twiddle v x
```


At call sites you just write instead of:
```haskell
cert' <- emptyOrNothing v txBody btbCerts
```

You write
```haskell
cert' <- emptyOrNothing v $ btbCerts txBody
```

lehins (changes requested):
Couple of suggestions, otherwise looks great

lehins (approved):
:+1:

PR #3098, opened 2022-10-24
Build with newly-released 1.1 versions of `plutus` libs

michaelpj:
This removes the `s-r-p` and allows GHC 9.2 compatibility for everything except `plutus-preprocessor` (it'll take us a while to get the plugin updated, and `plutus-preprocessor` isn't on the critical path to the node).


michaelpj:
Hmmm, what's going on with the GHA...

michaelpj:
Hmph, I cannot reproduce locally. Maybe the runner's cabal state got messed up :/

michaelpj:
The error makes no sense to me:
- `nothunks` has had an instance for `Vector` forever
- `cardano-crypto-class` has had `NoThunks (VerKeyDSIGN d)` as a superclass constraint on `DSIGNAlgorithm` since 2020

:thinking:

michaelpj:
@JaredCorduan I think this is good to go! Pretty minimal in the end.

JaredCorduan (approved):
looks great, thank you!

PR #3099, opened 2022-10-24
GHC 9.2 support

michaelpj:
VERY WIP

Depends on https://github.com/input-output-hk/cardano-ledger/pull/3098.

This is mostly here to get Hydra building the shell in the background for easier iteration. Then we need to actually push the changes to this branch to get things compiling with 9.2.

----

This moves the ledger to build primarily with GHC 9.2. I say "primarily", because there annoyingly remains one package that doesn't build with 9.2, namely `plutus-preprocessor`. This is because we haven't fixed up `plutus-tx-plugin` to work with 9.2 yet. 

The compromise position I've taken is to move the default build environment to 9.2, but leave instructions for how to get an 8.10 shell if you need to use `plutus-preprocessor`. This requires us to not break 8.10 support for a while, but hopefully not too long. And it still lets us progress towards a 9.2 node.

The alternative would be to just not try to keep 8.10 support at all, and not try to keep `plutus-preprocessor` building. Then we would resurrect it once we fix `plutus-tx-plugin`. I think that's worth considering, since it's not used that frequently and it's not that much code. This approach would allow you to start using 9.2 features everywhere (and give you a simpler CI), but at the cost of losing `plutus-preprocessor` for a bit. Up to you.


erikd:
I have everything other than `plutus-preprocessor` fixed in the `erikd/ghc-9.2` branch.

michaelpj:
Great, then I'll merge that into here shortly.

michaelpj:
There *was* an existing PR for the right branch, oops :facepalm:

PR #3101, opened 2022-10-25
Fix costmodel creation in Alonzo Genesis

teodanciu:


eras/alonzo/test-suite/cardano-ledger-alonzo-test.cabal
```diff
@@ -18,6 +18,7 @@ data-files:
   cddl-files/real/crypto.cddl
   cddl-files/mock/extras.cddl
   golden/*.cbor
+  golden/mainnet-alonzo-genesis.json
```
JaredCorduan, line 21: :heart:

JaredCorduan (approved):
This looks great to me, thank you @teodanciu !

@Jimbo4350 - you should be aware that the CLI is going to need to use `legacyParseCostModels` for parsing the original alonozo genesis file.

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Genesis.hs
```diff
@@ -216,38 +211,30 @@ instance ToJSONKey Language where
...
+-- Plutus API does no longer require the map as a parameter to `mkEvaluationContext`, but the list of Integers representing the values of the map.
+-- The expectation on this list of Integers is that they are sorted in the order given by the `ParamName` enum,
+-- so even though we just have to pass the list to plutus, we still need to use the names of the parameters in order to sort the list.
+-- In new versions, we want to represent the costmodel parameters directly as a list, so we can avoid this reordering, hence the name of this function.
+legacyParseCostModels :: Object -> Parser CostModels
+legacyParseCostModels o =
+  do
+    plutusV1 <- o .:? "PlutusV1"
+    cms <- traverse (validateCostModel PlutusV1 . cmParamValues) plutusV1
+    pure . CostModels . Map.fromList $ maybeToList cms
```
JaredCorduan: @teodanciu - I talked with @Jimbo4350 and learned how node is using the Alonzo genesis for testing. They want the ability to load up a bunch of cost models (V2, etc) in the alonzo era. Let's add parsing for the other language versions, but have them all just parse a list of integers (instead of the map).

teodanciu: Ah, sorry, just saw your comment.  
I just added a commit that parses V1 and V2 in the ToJSON instance of CostModels, that is not used by the code (only by an example NewGenesis data type that I added in tests). 

Do you know if they need the parsing of AlonzoGenesis or that of CostModels (that is contained within the AlonzoGenesis)? 
1)  If they need only the CostModels deserialization, then this new commit that I added should be fine - they only need to call FromJSON and will get CostModels for both versions.
2) However, if they want to deserialize AlonzoGenesis and expect CostModels for both V1 and V2, then I can make a simple change and not make the differentiation between legacy and non-legacy: let the FromJSON instance parse both V1 and V2 and construct AlonzoGenesis with both. 

What do you think between 1) and 2)

Sorry for the long comment - i hope it makes some sense.

JaredCorduan: I think we need 2 for testing purposes.

teodanciu: Ok!

teodanciu:
Hello @Jimbo4350 - would you mind taking a look at this before we're merging it? Just to make sure it will  work with your plans for testing.

Jimbo4350 (approved):
LGTM! Thanks :+1:

PR #3102, opened 2022-10-25
Fix doc site

michaelpj:
This was not required before, but is required by cicero, which threw up some errors.


teodanciu (approved):
```
Looks good to me
================
```

PR #3103, opened 2022-10-26
Get ormolu from haskell.nix

michaelpj:
`haskell.nix` has nice support for this kind of thing, we can just use it. I ran `./scripts/ormolise.sh` from a `nix-shell` and it was a no-op, so I think it worked.


JaredCorduan (approved):
wonderful, thank you @michaelpj !

Issue #3061, opened 2022-10-05
Remove _ prefix from record accessor names under shelley/impl.


aniketd:
Post the move away from using lenses, some of the record accessor names remained prefixed with `_`.
- [x] Replace `_` prefixed names with better names
- [x] ~~Add `DEPRECATED` standalone functions for those that are being used downstream to avoid any breakage~~

This is related to #2927 and its corresponding PR #3045 with some code overlap.

List of data types that would need to be addressed with removal of `_` prefix in record accessors.

1. [x] `Cardano.Ledger.Shelley.Delegation.Certificates.Delegation`
1. [x] `Cardano.Ledger.Shelley.EpochBoundary.SnapShot`
1. [x] `Cardano.Ledger.Shelley.EpochBoundary.SnapShots`
1. [x] `Cardano.Ledger.Shelley.LedgerState.DPState.DState`
1. [x] `Cardano.Ledger.Shelley.LedgerState.DPState.PState`
1. [x] `Cardano.Ledger.Shelley.LedgerState.Types.AccountState`
1. [x] `Cardano.Ledger.Shelley.LedgerState.Types.UTxOState`
1. [x] `Cardano.Ledger.Shelley.PParams.ShelleyPParamsHKD`
1. [x] `Cardano.Ledger.Shelley.PoolParams.PoolMetadata`
1. [x] `Cardano.Ledger.Shelley.PoolParams.PoolParams`
1. [x] `Cardano.Ledger.Shelley.Tx.TxRaw`
1. [x] `Cardano.Ledger.Shelley.TxBody.TxBodyRaw`
1. [x] `Cardano.Ledger.Shelley.TxBody.(pattern ShelleyTxBody)`

aniketd:
Starting with just `UTxOState` (across downstream repositories too)

aniketd:
`DState` and `PState` next.

aniketd:
Closing this, as all the points here now appear to have been addressed.

PR #3108, opened 2022-10-31
Rename UTxOState fields without underscores

aniketd:
Partially addresses #3061

- [x] Rename fields across current code base
~~- [ ] Add DEPRECATED warning for all renamed fields~~ They won't be useful as setters.
~~- [ ] Open PRs against downstream packages to not use the deprecated accessors.~~ The CHANGELOG update could be enough.


JaredCorduan (approved):
looks wonderful to me, thank you @aniketd !

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Rules/Ledger.hs
```diff
@@ -158,7 +158,7 @@ instance
         ( \(TRC (LedgerEnv {ledgerPp}, _, _))
            (LedgerState utxoSt DPState {dpsDState, dpsPState}) ->
               obligation ledgerPp (rewards dpsDState) (_pParams dpsPState)
-                == _deposited utxoSt
+                == utxosDeposited utxoSt
```
teodanciu, line 161: Just curious, is it `utxos` as in the plural of `utxo` or from `utxoState`  ?

lehins (changes requested):
This is a breaking change and requires a bump in version. So, currently released version of `cardano-ledger-shelley` is `0.1.0.0`, so we need to bump it up to `0.2.0.0`. Otherwise it looks great. This is a fairly new process, which we should all adopt

Also, we need to add a changelog to all of our packages and start placing changes there, instead of top level CHNAGELOG file. @aniketd This could be your next task, which you could do in a follow up PR

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Rules/Ledger.hs
aniketd: :+1:  It stands for `UTxO{S}tate` as in `State`, and not for plurality. :slightly_smiling_face:

teodanciu:
Should we also [release the new version  to cHaP](https://github.com/input-output-hk/cardano-haskell-packages#how-to-add-a-new-package-version-to-chap)?

lehins:
No, we don't need to make a release for every PR. Only when a downstream library is ready to integrate and use the new version, we should make a release

lehins (approved):
Looks great. It just needs a conflict resolution and I suggest to squash all commits into one

aniketd:
Done!

teodanciu (approved):
Looks good to me

Soupstraw (approved):
LGTM

teodanciu:
@aniketd  - this is the most highly approved PR I've seen!  :smile:  :tada:

PR #3109, opened 2022-10-31
Move some modules to core Version 4

TimSheard:
Moved Shelley.Orphans, Shelley.PoolParams and Shelley.EpochBoundary to Cardano.Ledger core.


eras/shelley/impl/src/Cardano/Ledger/Shelley/Orphans.hs
```diff
@@ -0,0 +1,13 @@
...
+-- but makes a bunch of orphan instances, and it is these orphan instances we
+-- want to be exported. The dodgy-exports Flag does not seem to understand about instances.
+
+module Cardano.Ledger.Shelley.Orphans
+  {-# DEPRECATED "Use 'import Cardano.Ledger.Orphans' instead" #-}
+  ( module X,
+  )
+where
+
+import Cardano.Ledger.Orphans as X ()
```
lehins: GHC option is not needed, it is indeed warning you about unnecessary thing. Instances are reexported implicitly. So this is a correct approach:
```haskell
module Cardano.Ledger.Shelley.Orphans
  {-# DEPRECATED "Use 'import Cardano.Ledger.Orphans' instead" #-}
  ( )
where

import Cardano.Ledger.Orphans ()
```

lehins (changes requested):
That's great. History is preserved. I can suggest squashing last three commits together, but that is just a suggestion. GHC flag needs fixing though, I left it as a comment on a commit

eras/shelley/impl/src/Cardano/Ledger/Shelley/Orphans.hs
TimSheard: So if you do this, it gives another error, something like Cardano.Ledger.Orphans does not import anything, and is redundant

lehins: If you do without `as X` it will not give this error. So, just doing this:
```
import Cardano.Ledger.Orphans ()
```
should not give you any errors

TimSheard: So I fixed this, as suggested, since it worked.

lehins (approved):
Awesome

CHANGELOG.md
```diff
@@ -9,6 +9,14 @@ in the naming of release branches.
 
 ## [Unreleased]
 
+
+### Changed
+
+- Moved Cardano.Ledger.Shelley.Orphans to Cardano.Ledger.Orphans  in the core modules
+- Moved Cardano.Ledger.Shelley.PoolParms to Cardano.Ledger.PoolParams  in the core modules
+- Moved Cardano.Ledger.Shelley.EpochBoundary to Cardano.Ledger.EpochBoundary  in the core modules
+- Added deprecated modules that import the moved ones and export their contents
```
JaredCorduan: We should move these to the top of the `Changed` section below.

PR #3110, opened 2022-10-31
Move UTxO to Core

TimSheard:
Split the UTxO file in Shelley into two parts.
One that deals just with UTxO  Cardano.Ledger.UTxO and the othet which deals with what happens in the 
Shelley era, Cardano.Ledger.Shelley.UTxO


eras/shelley/test-suite/src/Test/Cardano/Ledger/Shelley/Generator/Core.hs
```diff
@@ -109,17 +109,17 @@ import qualified Cardano.Ledger.Shelley.Tx as Ledger
...
   ( BlockNo (..),
     Duration (..),
     SlotNo (..),
     epochInfoFirst,
     (*-),
   )
+import Cardano.Ledger.UTxO
+  ( UTxO,
+    pattern UTxO,
+  )
```
lehins: Let's fix this while we at it: 
```suggestion
import Cardano.Ledger.UTxO (UTxO(UTxO))
```

lehins (approved):
This is great. Just one minor recommendation

eras/shelley/test-suite/src/Test/Cardano/Ledger/Shelley/Generator/Core.hs
TimSheard: done

TimSheard:
After some magic, managed to restore the git history to both UTxO filles.

lehins (approved):
Very nice!!!

libs/cardano-ledger-core/src/Cardano/Ledger/UTxO.hs
```diff
@@ -0,0 +1,251 @@
...
+    DSignable c (CH.Hash (HASH c) EraIndependentTxBody)
+  ) =>
+  SafeHash c EraIndependentTxBody ->
+  KeyPair kr c ->
+  WitVKey 'Witness c
+makeWitnessVKey safe keys =
+  WitVKey (asWitness $ vKey keys) (coerce $ signedDSIGN @c (sKey keys) (extractHash safe))
+
+-- | Create witnesses for transaction
+makeWitnessesVKey ::
```
JaredCorduan, line 178: this is just a side comment while I'm thinking about it, definitely ignore for now in this PR. but I think lots of the functions in this module are not really that related to the `UTxO` and should be in their own module.

JaredCorduan (approved):
fantastic, thank you @TimSheard !

libs/cardano-ledger-core/src/Cardano/Ledger/UTxO.hs
TimSheard: For the next cleanup PR

lehins (changes requested):
@TimSheard Squashing all of the commits together erased all of the work you did in the attempt to preserve git history. Are you sure that is what you wanted?

lehins (approved):
:+1:

PR #3114, opened 2022-11-02
Bump all cabal versions in preparation for uploading to CHaP

erikd:


erikd:
Not ready for this yet in `node` so closing it.

lehins:
Indiscriminate bumping up of all versions also goes against PVP, so that is not the way we should do releases

PR #3126, opened 2022-11-04
Rename TxBody and TxBodyRaw fields and patterns without underscores

aniketd:
Partially addresses #3061 
- [x] Update CHANGELOG

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,7 +28,107 @@ in the naming of release branches.
 
 ### Changed
 
-- Renamed records fields in `Cardano.Ledger` to names without `_` (underscores) #3118
+- Renamed records fields in `Cardano.Ledger` to names without `_` (underscores) #3126
+  - `Alonzo.TxBody.TxBodyRaw` to `Alonzo.TxBody.AlonzoTxBodyRaw`
+    - `_inputs` to `atbrInputs`
+    - `_collateral` to `atbrCollateral`
+    - `_outputs` to `atbrOutputs`
+    - `_certs` to `atbrCerts`
+    - `_wdrls` to `atbrWdrls`
+    - `_txfee` to `atbrTxFee`
+    - `_vldt` to `atbrValidityInterval`
+    - `_update` to `atbrUpdate`
+    - `_reqSignerHashes` to `atbrReqSignerHashes`
+    - `_mint` to `atbrMint`
+    - `_scriptIntegrityHash` to `atbrScriptIntegrityHash`
+    - `_adHash` to `atbrAuxDataHash`
+    - `_txnetworkid` to `atbrTxNetworkId`
+  - `Alonzo.TxBody.AlonzoTxBody` pattern synonym
+    - `inputs` to `atbInputs`
+    - `collateral` to `atbCollateral`
+    - `outputs` to `atbOutputs`
+    - `txcerts` to `atbCerts`
+    - `txwdrls` to `atbWdrls`
+    - `txfee` to `atbTxFee`
+    - `txvldt` to `atbValidityInterval`
+    - `txUpdates` to `atbUpdate`
+    - `reqSignerHashes` to `atbReqSignerHashes`
+    - `mint` to `atbMint`
+    - `scriptIntegrityHash` to `atbScriptIntegrityHash`
+    - `adHash` to `atbAuxDataHash`
+    - `txnetworkid` to `atbTxNetworkId`
+  - `Babbage.TxBody.TxBodyRaw` to `Babbage.TxBody.BabbageTxBodyRaw`
+    - `_spendInputs` to `btbrSpendInputs`
+    - `_collateralInputs` to `btbrCollateralInputs`
+    - `_referenceInputs` to `btbrReferenceInputs`
+    - `_outputs` to `btbrOutputs`
+    - `_collateralReturn` to `btbrCollateralReturn`
+    - `_certs` to `btbrCerts`
+    - `_wdrls` to `btbrWdrls`
+    - `_txfee` to `btbrTxFee`
+    - `_vldt` to `btbrValidityInterval`
+    - `_update` to `btbrUpdate`
+    - `_reqSignerHashes` to `btbrReqSignerHashes`
+    - `_mint` to `btbrMint`
+    - `_scriptIntegrityHash` to `btbrScriptIntegrityHash`
+    - `_adHash` to `btbrAuxDataHash`
+    - `_txnetworkid` to `btbrTxNetworkId`
+  - `Babbage.TxBody.BabbageTxBody` pattern synonym
+    - `inputs` to `btbInputs`
+    - `collateral` to `btbCollateral`
+    - `referenceInputs` to `btbReferenceInputs`
+    - `outputs` to `btbOutputs`
+    - `collateralReturn` to `btbCollateralReturn`
+    - `totalCollateral` to `btbTotalCollateral`
+    - `txcerts` to `btbCerts`
+    - `txwdrls` to `btbWdrls`
+    - `txfee` to `btbTxFee`
+    - `txvldt` to `btbValidityInterval`
+    - `txUpdates` to `btbUpdate`
+    - `reqSignerHashes` to `btbReqSignerHashes`
+    - `mint` to `btbMint`
+    - `scriptIntegrityHash` to `btbScriptIntegrityHash`
+    - `adHash` to `btbAuxDataHash`
+    - `txnetworkid` to `btbrTxNetworkId`
+  - `ShelleyMA.TxBody.TxBodyRaw` to `ShelleyMA.TxBody.MATxBodyRaw`
+    - `inputs` to `matbrInputs`
+    - `outputs` to `matbrOutputs`
+    - `certs` to `matbrCerts`
+    - `wdrls` to `matbrWdrls`
+    - `txfee` to `matbrTxFee`
+    - `vldt` to `matbrValidityInterval`
+    - `update` to `matbrUpdate`
+    - `adHash` to `matbrAuxDataHash`
+    - `mint` to `matbrMint`
+  - `ShelleyMA.TxBody.MATxBody` pattern synonym
+    - `inputs` to `matbInputs`
+    - `outputs` to `matbOutputs`
+    - `certs` to `matbCerts`
+    - `wdrls` to `matbWdrls`
+    - `txfee` to `matbTxFee`
+    - `vldt` to `matbValidityInterval`
+    - `update` to `matbUpdate`
+    - `adHash` to `matbAuxDataHash`
+    - `mint` to `matbMint`
+  - `Shelley.TxBody.TxBodyRaw` to `Shelley.TxBody.ShelleyTxBodyRaw`
+    - `inputsX` to `stbrInputs`
+    - `outputsX` to `stbrOutputs`
+    - `certsX` to `stbrCerts`
+    - `wdrlsX` to `stbrWdrls`
+    - `txfeeX` to `stbrTxFee`
+    - `ttlX` to `stbrTTL`
+    - `txUpdateX` to `stbrUpdate`
+    - `mdHashX` to `stbrMDHash`
+  - `Shelley.TxBody.ShelleyTxBody` pattern synonym
+    - `_inputs` to `stbInputs`
+    - `_outputs` to `stbOutputs`
+    - `_certs` to `stbCerts`
+    - `_wdrls` to `stbWdrls`
+    - `_txfee` to `stbTxFee`
+    - `_ttl` to `stbTTL`
+    - `_txUpdate` to `stbUpdate`
+    - `_mdHash` to `stbMDHash`
+ - Renamed records fields in `Cardano.Ledger` to names without `_` (underscores) #3118
   - `Shelley.LedgerState.Types.AccountState`
     - `_treasury -> asTreasury`
     - `_reserves -> asReserves`
@@ -45,7 +145,7 @@ in the naming of release branches.
   - `PoolParams.PoolMetadata`
     - `_poolMDUrl -> pmUrl`
     - `_poolMDHash -> pmHash`
-- Renamed records fields in `Cardano.Ledger` to names without `_` (underscores) #?
+- Renamed records fields in `Cardano.Ledger` to names without `_` (underscores) #3120
   - `Shelley.Delegation.Certificates`
     - `_delegator` to `dDelegator`
     - `_delegatee` to `dDelegatee`
@@ -120,6 +220,16 @@ in the naming of release branches.
 
 ### Removed
 
+- Removed pattern synonym `Cardano.Ledger.ShelleyMA.TxBody.MATxBody.TxBody'` with fields #3126
+  - `adHash'`
+  - `certs'`
+  - `inputs'`
+  - `mint'`
+  - `outputs'`
+  - `txfee'`
+  - `update'`
+  - `vldt'`
+  - `wdrls'`
 - Deprecated the `validPlutusdata` function: #3006
 - Deprecated the misspelled `HasAlgorithm` type alias: #3007
 - Deprecated `CLI.evaluateMinLovelaceOutput` in favor of newly added
--- a/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxBody.hs
+++ b/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxBody.hs
@@ -26,19 +26,19 @@ module Cardano.Ledger.Alonzo.TxBody
     DataHash32,
     AlonzoTxBody
       ( AlonzoTxBody,
-        inputs,
-        collateral,
-        outputs,
-        txcerts,
-        txwdrls,
-        txfee,
-        txvldt,
-        txUpdates,
-        reqSignerHashes,
-        mint,
-        scriptIntegrityHash,
-        adHash,
-        txnetworkid
+        atbInputs,
+        atbCollateral,
+        atbOutputs,
+        atbCerts,
+        atbWdrls,
+        atbTxFee,
+        atbValidityInterval,
+        atbUpdate,
+        atbReqSignerHashes,
+        atbMint,
+        atbScriptIntegrityHash,
+        atbAuxDataHash,
+        atbTxNetworkId
       ),
     AlonzoEraTxBody (..),
     ShelleyEraTxBody (..),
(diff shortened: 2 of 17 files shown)


eras/shelley-ma/impl/src/Cardano/Ledger/Mary/Value.hs
```diff
@@ -367,14 +367,14 @@ decodeIntegerBounded64 = do
     TypeUInt64 -> pure ()
     TypeNInt -> pure ()
     TypeNInt64 -> pure ()
-    _ -> fail "expected major type 0 or 1 when decoding mint field"
+    _ -> fail "expected major type 0 or 1 when decoding smatbrMint field"
```
Soupstraw: I think it would be more readable when we left it to `mint` here

aniketd: Oops! Yes!

eras/shelley/impl/src/Cardano/Ledger/Shelley/TxBody.hs
```diff
@@ -237,29 +237,29 @@ txSparse (TxBodyRaw input output cert wdrl fee ttl update hash) =
...
-      _ttlX = SlotNo 0,
-      _certsX = StrictSeq.empty,
-      _wdrlsX = Wdrl Map.empty,
-      _txUpdateX = SNothing,
-      _mdHashX = SNothing
+  ShelleyTxBodyRaw
+    { stbrInputs = Set.empty,
+      stbrOutputs = StrictSeq.empty,
+      stbrTxFee = Coin 0,
+      stbrTtl = SlotNo 0,
```
Soupstraw: I think `stbrTTL` would be better, since it's an abbreviation (time-to-live)

eras/shelley/test-suite/src/Test/Cardano/Ledger/Shelley/Address/Bootstrap.hs
```diff
@@ -225,14 +225,14 @@ coinsToBob = Coin 1000
...
-      _txUpdate = SNothing,
-      _mdHash = SNothing
+    { sbtInputs = Set.fromList [TxIn genesisId minBound],
+      sbtOutputs = StrictSeq.fromList [ShelleyTxOut bobAddr coinsToBob, ShelleyTxOut aliceAddr change],
+      sbtCerts = StrictSeq.fromList mempty,
+      sbtWdrls = Wdrl Map.empty,
+      sbtTxfee = fee,
+      sbtTtl = SlotNo 10,
+      sbtTxUpdate = SNothing,
+      sbtMdHash = SNothing
```
Soupstraw: `sbtMDHash` would be better here

eras/shelley-ma/impl/src/Cardano/Ledger/ShelleyMA/Rules/Utxo.hs
```diff
@@ -180,7 +180,7 @@ utxoTransition = do
   -- runValidationTransMaybe fromShelleyFailure $ Shelley.validateInputSetEmptyUTxO txb
   runTest $ Shelley.validateInputSetEmptyUTxO txb
 
-  {- minfee pp tx ≤ txfee txb -}
+  {- minfee pp tx ≤ smatbrTxFee txb -}
```
lehins: We should revert the changes in all comments, since those are snippets from the spec
```suggestion
  {- minfee pp tx ≤ txfee txb -}
```

eras/shelley-ma/impl/src/Cardano/Ledger/ShelleyMA/Rules/Utxo.hs
```diff
@@ -201,8 +201,8 @@ utxoTransition = do
   ppup' <-
     trans @(EraRule "PPUP" era) $ TRC (PPUPEnv slot pp genDelegs, ppup, txup tx)
 
-  {- adaPolicy ∉ supp mint tx
-     above check not needed because mint field of type MultiAsset cannot contain ada -}
+  {- adaPolicy ∉ supp smatbrMint tx
+     above check not needed because smatbrMint field of type MultiAsset cannot contain ada -}
```
lehins: Same here

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxBody.hs
```diff
@@ -122,40 +122,40 @@ import Prelude hiding (lookup)
...
-    _outputs :: !(StrictSeq (Core.TxOut era)),
-    _certs :: !(StrictSeq (DCert (EraCrypto era))),
-    _wdrls :: !(Wdrl (EraCrypto era)),
-    _txfee :: !Coin,
-    _vldt :: !ValidityInterval,
-    _update :: !(StrictMaybe (Update era)),
-    _reqSignerHashes :: Set (KeyHash 'Witness (EraCrypto era)),
-    _mint :: !(MultiAsset (EraCrypto era)),
+data AlonzoTxBodyRaw era = AlonzoTxBodyRaw
+  { atbrInputs :: !(Set (TxIn (EraCrypto era))),
```
teodanciu, line 126: I find these names a bit difficult to process, because of how alien these abbreviations seem (`atbr`, `btbr`, `smat`) - only because they haven't been used, yet so the don't mean anything at the moment. 

On the other hand, I see there's not too many references to them, so maybe it's not so important. 
Just wondering, have you considered using the era name as a prefix? Would it be too long or maybe conflicting with something else? 
I mean 
```
alonzoInputs :: 
alonzoCollateral :: .. 
....
babbageSpendInputs  :: ... 

etc
```

aniketd: I completely agree with you on this @teodanciu. The abbreviations are rather unpronounciable. I don't think they are likely to conflict with anything else, though the names might get "too long" - this is a subjective presumption. I can change this to using full era names if you think it will be helpful.

teodanciu: Maybe let's see what others think about it before changing, maybe someone has a better idea altogether!

Soupstraw: I use a similar style as Aniket, so I'm used to these kinds of names

lehins: Same here, I was able to decipher `atbr` for `alonzo tx body raw` just fine. This is a common convention in Haskell, in order to avoid really long names.

The important part is the consistency. This gives us a predictable way of uniquely identifying the type the field works with. The naming convention of using scientists' names in alphabetical order helps us avoid those clashes as well if we use just the first latter (eg, `a` for `Alonzo`, `b` for `Babbage`)


Also with respect to all of the parameterized families, `Tx` , `TxBody`, etc. we should use lenses anyways, so these record field names are mostly an internal implementation detail.

eras/shelley/test-suite/src/Test/Cardano/Ledger/Shelley/Generator/ShelleyEraGen.hs
```diff
@@ -76,11 +76,11 @@ instance
   genEraTxBody _ge _utxo = genTxBody
   genEraAuxiliaryData = genMetadata
 
-  updateEraTxBody _utxo _pp _wits body' fee ins out =
+  updateEraTxBody _utxo _pp _trWits body' fee ins out =
```
lehins: Wrongful search and replacing function arguments. There are many occurrences of this, could you please revert them back
```suggestion
  updateEraTxBody _utxo _pp _wits body' fee ins out =
```

lehins (changes requested):
There are a few accidental renames that weren't suppose to happen,other than that it looks pretty good

eras/shelley/test-suite/src/Test/Cardano/Ledger/Shelley/Generator/ShelleyEraGen.hs
aniketd: Noted @lehins :memo: :pray: . Will get them reverted.

eras/shelley-ma/impl/src/Cardano/Ledger/ShelleyMA/Rules/Utxo.hs
aniketd: Noted! Fixed!

eras/shelley-ma/impl/src/Cardano/Ledger/ShelleyMA/Rules/Utxo.hs
aniketd: Noted! Fixed!

eras/shelley-ma/impl/src/Cardano/Ledger/ShelleyMA/TxBody.hs
```diff
@@ -287,19 +287,19 @@ pattern TxBody' ::
   StrictMaybe (AuxiliaryDataHash (EraCrypto era)) ->
   MultiAsset (EraCrypto era) ->
   MATxBody era
-pattern TxBody' {inputs', outputs', certs', wdrls', txfee', vldt', update', adHash', mint'} <-
+pattern TxBody' {smatbrInputs', smatbrOutputs', smatbrCerts', smatbrWdrls', smatbrTxFee', smatbrValidityInterval', smatbrTxUpdate', smatbrAuxDataHash', smatbrMint'} <-
```
lehins: These actually don't need the `r` at the end, since they don't operate on the `Raw` type, but on the . Also removing the ugly tick is a good idea:
```suggestion
pattern TxBody' {matbInputs, matbOutputs, matbCerts, matbWdrls, matbTxFee, matbValidityInterval, matrTxUpdate, matbAuxDataHash, matbMint} <-
```

eras/shelley-ma/impl/src/Cardano/Ledger/ShelleyMA/TxBody.hs
```diff
@@ -104,63 +104,63 @@ import NoThunks.Class (NoThunks (..))
...
-  { inputs :: !(Set (TxIn (EraCrypto era))),
-    outputs :: !(StrictSeq (ShelleyTxOut era)),
-    certs :: !(StrictSeq (DCert (EraCrypto era))),
-    wdrls :: !(Wdrl (EraCrypto era)),
-    txfee :: !Coin,
-    vldt :: !ValidityInterval, -- imported from Timelocks
-    update :: !(StrictMaybe (Update era)),
-    adHash :: !(StrictMaybe (AuxiliaryDataHash (EraCrypto era))),
-    mint :: !(MultiAsset (EraCrypto era))
+data ShelleyMATxBodyRaw era = ShelleyMATxBodyRaw
```
lehins: For consistency it should be just `MATxBodyRaw`  (in order to match `MATxBody`), which also means that prefix  will need to be switched from `smatbr` -> `matbr`
```suggestion
data MATxBodyRaw era = MATxBodyRaw
```

"Shelley" in the naming of the package is an artifact, so prefixing types with `Shelley` doesn't make much sense (especially since we are planning on splitting into two separate packages `Mary` and `Allegra`

lehins (changes requested):
This is looking very good. I noticed some more stuff that can be improved upon the second review. It would also be nice to make pattern synonym record names for `[Era]TxBody` constant for all eras as well. For example `AlonzoTxBody` is defined as:
```haskell
pattern AlonzoTxBody
  { inputs,
    collateral,
    outputs,
    txcerts,
    txwdrls,
    txfee,
    txvldt,
    txUpdates,
    reqSignerHashes,
    mint,
    scriptIntegrityHash,
    adHash,
    txnetworkid
  }
```

But it should be something like this:
```haskell
pattern AlonzoTxBody
  { atbInputs,
    atbCollateral,
    atbOutputs,
    atbCerts,
    atbWdrls,
    atbTxFee,
    atbValidityInterval,
    atbUpdate,
    atbReqSignerHashes,
    atbMint,
    atbScriptIntegrityHash,
    atbAuxDataHash,
    atbTxNetworkId
  }
```

We want to make sure that there are not a dozen different names for a similar concept (eg `txUpdates` vs `atbrTxUpdate` or `_txUpdate` prior to this PR - note that one is plural another one is not) 

I guess the goal is to have no ambiguity between era and the least amount of discrepancy between record fields for the raw type, the actual newtype wrapper around it as well as the lens names for the corresponding fields.

eras/shelley-ma/impl/src/Cardano/Ledger/ShelleyMA/TxBody.hs
aniketd: Thanks for the suggestion :+1: I also renamed `TxBody'` to `MATxBody_`

eras/shelley-ma/impl/src/Cardano/Ledger/ShelleyMA/TxBody.hs
aniketd: Totally Agree!

eras/shelley-ma/impl/src/Cardano/Ledger/ShelleyMA/TxBody.hs
```diff
@@ -21,20 +21,29 @@ module Cardano.Ledger.ShelleyMA.TxBody
...
+        matbWdrls,
+        matbAuxDataHash_,
+        matbCerts_,
+        matbInputs_,
+        matbMint_,
+        matbOutputs_,
+        matbTxFee_,
+        matbUpdate_,
+        matbValidityInterval_,
+        matbWdrls_
```
aniketd: Here the set of exports with trailing `_` have been added like the set of exports for identifiers without trailing `_`, owing to the two separate pattern synonyms existing for `MATxBody`

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxBody.hs
```diff
@@ -26,19 +26,19 @@ module Cardano.Ledger.Alonzo.TxBody
...
+        atbCerts,
+        atbWdrls,
+        atbTxFee,
+        atbValidityInterval,
+        atbUpdate,
+        atbReqSignerHashes,
+        atbMint,
+        atbScriptIntegrityHash,
+        atbAuxDataHash,
+        atbTxNetworkId
```
lehins, line 41: Adding these renames to the changelog would probably be a good idea.

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxBody.hs
```diff
@@ -300,79 +300,53 @@ pattern AlonzoTxBody ::
...
-            vldtX
-            updateX
-            reqSignerHashesX
-            mintX
-            scriptIntegrityHashX
-            adHashX
-            txnetworkidX
+    AlonzoTxBody inputs collateral outputs certs wdrls txfee validityInterval update reqSignerHashes mint scriptIntegrityHash auxDataHash txnetworkid =
+      mkAlonzoTxBody $
+        AlonzoTxBodyRaw inputs collateral outputs certs wdrls txfee validityInterval update reqSignerHashes mint scriptIntegrityHash auxDataHash txnetworkid
```
lehins: Sorry for asking you to do all this mundane stuff, but could you switch this to use record syntax, it would be much cleaner and follow the pattern above. Something along the lines:
```
AlonzoTxBody {atbInputs, ...} =
  mkAlonzoTxBodyRaw $
    AlonzoTxBodyRaw {
       atbrInputs = atbrInputs,
       ....
       }
````

File: eras/babbage/impl/src/Cardano/Ledger/Babbage/TxBody.hs [hunk 13]
```diff
@@ -37,22 +37,22 @@ module Cardano.Ledger.Babbage.TxBody
     babbageMinUTxOValue,
     BabbageTxBody
       ( BabbageTxBody,
-        inputs,
```
lehins, line 40: Same with these ones in the changelog

eras/babbage/impl/src/Cardano/Ledger/Babbage/TxBody.hs
```diff
@@ -533,85 +533,53 @@ pattern BabbageTxBody ::
...
-            vldtX
-            updateX
-            reqSignerHashesX
-            mintX
-            scriptIntegrityHashX
-            adHashX
-            txnetworkidX
+    BabbageTxBody inputs collateral referenceInputs outputs collateralReturn totalCollateral certs wdrls txfee vldt update reqSignerHashes mint scriptIntegrityHash adHash txnetworkid =
+      mkBabbageTxBodyFromRaw $
+        BabbageTxBodyRaw inputs collateral referenceInputs outputs collateralReturn totalCollateral certs wdrls txfee vldt update reqSignerHashes mint scriptIntegrityHash adHash txnetworkid
```
lehins: Same here with respect to record style syntax

eras/shelley-ma/impl/src/Cardano/Ledger/ShelleyMA/TxBody.hs
lehins: Sorry, I totally messed up. This is the only era that provides a special pattern synonym for deconstructing only.

We need to get rid of it and other accessors (like `inputs'`, `outputs'`, etc.) in other eras, but we  probably should do that in a separate PR. I created an issue for this : #3130 
Could you please revert this change of tick to underscore and the prefix as well `certs'` -> `matbCerts_` for this pattern `MATXBody'`/`MATXBody_`. Alternatively you can just get rid of this pattern synonym in this PR if you prefer.

lehins (changes requested):
Sorry, more review comments. I am truly enjoying seeing this cleanup!

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxBody.hs
aniketd: Sure! :smile: I did think about it before too, but thought that using it in one place will give enough safety. I will switch it out.

eras/babbage/impl/src/Cardano/Ledger/Babbage/TxBody.hs
aniketd: :+1:

eras/babbage/impl/src/Cardano/Ledger/Babbage/TxBody.hs
aniketd: Noted! All `[Era]TxBody` pattern synonyms should need the switch :+1:

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxBody.hs
aniketd: Oops! Missed it.

eras/shelley-ma/impl/src/Cardano/Ledger/ShelleyMA/TxBody.hs
aniketd: I got rid of it!

CHANGELOG.md
```diff
@@ -28,7 +28,107 @@ in the naming of release branches.
...
+    - `_wdrls` to `atbrWdrls`
+    - `_txfee` to `atbrTxFee`
+    - `_vldt` to `atbrValidityInterval`
+    - `_update` to `atbrUpdate`
+    - `_reqSignerHashes` to `atbrReqSignerHashes`
+    - `_mint` to `atbrMint`
+    - `_scriptIntegrityHash` to `atbrScriptIntegrityHash`
+    - `_adHash` to `atbrAuxDataHash`
+    - `_txnetworkid` to `atbrTxNetworkId`
+  - `Alonzo.TxBody.AlonzoTxBodyRaw` pattern synonym
```
```suggestion
  - `Alonzo.TxBody.AlonzoTxBody` pattern synonym
```

lehins (approved):
Looks awesome. Thank you!

aniketd:
Squashed and rebased!

aniketd:
Can I merge this?

teodanciu:
@aniketd It's approved,  good to merge I think! :rocket: 
We can always change later if some details could still be improved

JaredCorduan:
go for it @aniketd !

PR #3127, opened 2022-11-04
Redo how we manage deposits

TimSheard:
This PR explores one mechanism for tracking deposits for each Stake credential and each Pool.
Motivated by https://github.com/input-output-hk/cardano-ledger/issues/3113

Here is what happened
Added deposits (key deposits) to DState and (pool deposits) PState.
    Functions that computed deposits, obligation, and refunds (keyRefunds
    and totalDeposits, obigation etc) were changed to take DPState as input
    Added DPState to UtxoEnv replacing pool stake. Incorporated TreeDiff
    to show differences in two NewEpochState's. Gathered all the TreeDiff
    stuff into one module Cardano.Ledger.TreeDiff in cardano-binary. This
    includes all needed orphan instances from external types. Importing this module
    should supply everything needed, and carry the orphan instances.
    Added (.->>) as a version of (.->) that uses tree diff in small-steps-test
    Added property tests that test the new invariants that must hold
    1) utxosDepoits == sum (dsDeposited) + sum (psDeposited)
    2) dom rewards == dom dsDeposited


CHANGELOG.md
```diff
@@ -12,6 +12,13 @@ in the naming of release branches.
 
 ### Added
 
+- Add ToExpr (from tree-diff) instances for all types inside NewEpochState
+  - Add module Cardano.Ledger.TreeDiff in cardano-binary. Includes all needed operations, classes, and orphan instances	
+- Add deposits (key deposits) to DState{dsDeposits} and (pool deposits) PState{psDeposits}
+  - Added property tests that test the new invariants that must hold
+    - utxosDepoits == sum (dsDeposited) + sum (psDeposited)
```
teodanciu: A little typo here, missing an `s` in `utxosDeposits`

.gitignore
```diff
@@ -52,6 +52,13 @@ specs/**/.ghc.environment.x86_64-linux-*
 
 *.pdf
 .envrc
+## Ledger Shortcuts
+SHELLY
+ALLEGRA
+MARY
+ALONZO
+BABBAGE
+CORE
```
lehins: It is a common practice to put your personal things that the git should ignore into your own personal gitignore file. So add this to your `~/.gitconfig`:
```
[core]
        excludesfile = /home/tim/.gitignore_global
```
and then create a `/home/tim/.gitignore_global` file with the same contents that you tried adding here

libs/cardano-ledger-binary/src/Cardano/Ledger/TreeDiff.hs
```diff
@@ -0,0 +1,86 @@
+{-# LANGUAGE FlexibleInstances #-}
+{-# OPTIONS_GHC -Wno-orphans #-}
+
+-- | Collect all the basic TreeDiff stuff in one place.
+--   Including orphan instances from external types.
+--   So anyplace this module is imported to get access to the
+--   ToExpr class, one also gets the orphan instances.
+module Cardano.Ledger.TreeDiff
```
lehins, line 8: It think this module should go into `cardano-data` package instead. It doesn't really have anything to do with binary

libs/cardano-ledger-binary/src/Cardano/Ledger/TreeDiff.hs
```diff
@@ -0,0 +1,86 @@
...
+instance ToExpr a => ToExpr (StrictSeq a) where
+  toExpr x = App "StrictSeqFromList" [listToExpr (toList x)]
+
+instance ToExpr a => ToExpr (StrictMaybe a)
+
+instance (ToExpr a, ToExpr b) => ToExpr (VMap VB VB a b) where
+  toExpr x = toExpr (toMap x)
+
+instance (ToExpr a, ToExpr b, Prim b) => ToExpr (VMap VB VP a b) where
+  toExpr x = toExpr (toMap x)
```
lehins: These can be described in one instance and it should live in the `vector-map` package instead of as an orphan here.

```suggestion
instance (VG.Vector kv k, VG.Vector vv v, ToExpr k, ToExpr v) => ToExpr (VMap kv vv k v) where
  toExpr = toExpr . toMap
```

libs/cardano-ledger-binary/src/Cardano/Ledger/TreeDiff.hs
```diff
@@ -0,0 +1,86 @@
...
+import Data.TreeDiff.Expr (Expr (App, Lst, Rec))
+import Data.TreeDiff.QuickCheck (ediffEq)
+import Data.VMap (VB, VMap, VP, toMap)
+import Test.Hspec (Expectation, HasCallStack, expectationFailure)
+
+-- =====================================================
+-- Cardano functions that deal with TreeDiff and ToExpr
+
+trimExprViaShow :: Show a => Int -> a -> Expr
+trimExprViaShow n x = App (take n (drop 1 (show x)) ++ "..") []
```
lehins: What is being dropped in the `drop 1`?

libs/cardano-ledger-binary/src/Cardano/Ledger/TreeDiff.hs
```diff
@@ -0,0 +1,86 @@
...
+-- | Check that two values are equal and if they are not raise an exception with the
+-- `ToExpr` diff
+expectExprEqual :: (Eq a, ToExpr a) => a -> a -> Expectation
+expectExprEqual x y = expectExprEqualWithMessage "Expected two values to be equal:" x y
+
+expectExprEqualWithMessage :: (ToExpr a, Eq a, HasCallStack) => [Char] -> a -> a -> Expectation
+expectExprEqualWithMessage message expected actual =
+  unless (actual == expected) (expectationFailure msg)
+  where
+    msg = (if null message then "" else message ++ "\n") ++ diffExpr expected actual
```
lehins: These (`expectExprEqual` and `expectExprEqualWithMessage`) should go back into the test module `Test.Cardano.Ledger.Binary.TreeDiff`, because we do not want to depend on packages that we use for testing in main libraries

libs/cardano-ledger-binary/cardano-ledger-binary.cabal
```diff
@@ -78,6 +79,8 @@ library
                    , text
                    , time
                    , transformers >= 0.5
+                   , tree-diff
+                   , hspec
```
lehins: You see, if you ever have a need to add something like `QuickCheck` or `hspec`  or `tasty` as a dependency to the library it is an indicator that there is some functionality being misplaced and should move to the testing library instead
```suggestion
```

libs/cardano-ledger-core/src/Cardano/Ledger/BaseTypes.hs
```diff
@@ -697,3 +698,32 @@ word16FromInteger :: Integer -> Maybe Word16
...
+
+instance ToExpr Network
+
+instance ToExpr Port
+
+instance ToExpr Url
+
+instance ToExpr Nonce where
+  toExpr NeutralNonce = App "NeutralNonce" []
+  toExpr (Nonce x) = App "Nonce" [App (take 10 (drop 1 (show x)) ++ "..") []]
```
lehins:
```suggestion
  toExpr (Nonce x) = App "Nonce" [trimExprViaShow 10 x]
```

libs/cardano-ledger-binary/src/Cardano/Ledger/TreeDiff.hs
lehins: Also I'd suggest showing the prefix as well as the suffix. It is easy to introduce a bug that will only make the prefix match. 


Something along the lines: `take n (drop 1 (show x)) ++ ".." ++ reverse (take n (drop 1 (show (reverse x)))) ++ ".."`

.gitignore
JaredCorduan: how does this work? :thinking:

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Rules/Utxow.hs
```diff
@@ -76,15 +76,9 @@ import Cardano.Ledger.Shelley.Delegation.Certificates
...
-    propWits,
     witsFromTxWitnesses,
   )
-import Cardano.Ledger.Shelley.Rules
-  ( ShelleyUtxowEvent (UtxoEvent),
-    ShelleyUtxowPredFailure (..),
-    UtxoEnv (..),
-    validateNeededWitnesses,
-  )
+import Cardano.Ledger.Shelley.Rules (ShelleyUtxowEvent (UtxoEvent), ShelleyUtxowPredFailure (..), UtxoEnv (..), propWits, validateNeededWitnesses)
```
JaredCorduan: sorry for the bike shedding, but I prefer using the vertical space instead, it makes for better diff history. that's just my opinion, though, and if everyone else disagrees we can use the horizontal space.

lehins (changes requested):
I haven't looked through the full PR yet, here are my comments thus far

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Rules/Utxow.hs
lehins: For lines that are that long, I do agree.

libs/small-steps-test/src/Control/State/Transition/Trace.hs
```diff
@@ -413,7 +415,21 @@ mSt .- sig = do
     Right st' -> pure st'
 
 -- | Bind the state inside the first argument, and check whether it is equal to
--- the expected state, given in the second argument.
+-- the expected state, given in the second argument. If it isn't raise an Assertion that
+-- uses a tree-diff to describe the differences.
+(.->>) ::
```
JaredCorduan, line 421: just my preference, but think this `.->>` is superior and that we can just replace `.->` with it instead of adding a new operator.

eras/shelley/impl/src/Cardano/Ledger/Shelley/API/Wallet.hs
```diff
@@ -515,26 +521,22 @@ evaluateTransactionFee pp tx numKeyWits = getMinFeeTx pp tx'
...
+  DepositInfo era ->
   -- | The UTxO relevant to the transaction.
   UTxO era ->
-  -- | A predicate that a stake pool ID is new (i.e. unregistered).
-  -- Typically this will be:
-  --
-  -- @
-  --   (`Map.notMember` stakepools)
-  -- @
-  (KeyHash 'StakePool (EraCrypto era) -> Bool) ->
```
JaredCorduan, line 531: This was on optimization that we intentionally added for the wallet team, namely to not have to pass the whole map. I forget the details, but we need to make sure that removing this is okay. It looks like `keyCertsRefunds` needs the tracked deposits, so perhaps there is nothing better we can do. cc @HeinrichApfelmus

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Epoch.hs
```diff
@@ -193,11 +173,13 @@ epochTransition = do
       TRC (epochState', UpecState pp (utxosPpups utxoSt'), ())
   let utxoSt'' = utxoSt' {utxosPpups = ppupSt'}
 
-  let Coin oblgCurr = obligation pp (rewards dstate') (psStakePoolParams pstate'')
-      Coin oblgNew = obligation pp' (rewards dstate') (psStakePoolParams pstate'')
+  let -- Coin oblgCurr = obligationDPState (DPState dstate' pstate'')
+      -- Now that obligation does not depend on the PParams at all, the current
+      -- and the new are identical, so we don't have to adjust for the difference.
+      Coin oblgNew = obligationDPState (DPState dstate' pstate'')
```
JaredCorduan: is this left so that we can make use of it in a property test?

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Pool.hs
```diff
@@ -21,6 +21,7 @@ module Cardano.Ledger.Shelley.Rules.Pool
   )
 where
 
+-- import Debug.Trace(trace)
```
JaredCorduan: can remove :)

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/PoolReap.hs
```diff
@@ -113,24 +115,35 @@ instance
...
+      {- An alternate approach
+      (refunds,mRefunds) = Map.foldlWithKey accum (mempty,mempty) (eval (retired ◁ psStakePoolParams ps))
+        where accum ans@(refund,unclaim) poolHash poolparam  =
+                case Map.lookup poolHash retiringDeposits of
+                  Nothing -> ans
+                  Just coin -> if eval (cred  ∈ dom (rewards ds))
+                                  then (Map.insertWith (<+>) cred coin refund,unclaim)
+                                  else (refund, Map.insertWith (<+>) cred coin unclaim)
+                    where cred = getRwdCred (_poolRAcnt poolparam)
+      -}
```
JaredCorduan: is there a reason to keep this?

eras/shelley/test-suite/src/Test/Cardano/Ledger/Shelley/Rules/TestChain.hs
```diff
@@ -160,7 +154,45 @@ import Test.QuickCheck
     (.||.),
     (===),
   )
+import Test.Tasty (TestTree, defaultMain, testGroup)
+import qualified Test.Tasty.QuickCheck as TQC
 
+-- ==========================================
+
+test1 ::
```
JaredCorduan:
```suggestion
preservationOfAda ::
```

eras/shelley/test-suite/src/Test/Cardano/Ledger/Shelley/Rules/TestChain.hs
```diff
@@ -160,7 +154,45 @@ import Test.QuickCheck
...
+minimal =
+  testGroup
+    "Minimal Property Tests"
+    [ TQC.testProperty "total amount of Ada is preserved (Chain)" (test1 @era)
+    ]
+
+main :: IO ()
+main = defaultMain (minimal @(ShelleyEra TestCrypto))
+
+-- :main --quickcheck-replay=443873
```
JaredCorduan: a bug we've now squashed?

eras/shelley/test-suite/src/Test/Cardano/Ledger/Shelley/Rules/TestChain.hs
```diff
@@ -289,12 +321,15 @@ adaPreservationChain ::
   Property
 adaPreservationChain =
   forAllChainTrace @era longTraceLen $ \tr -> do
-    let ssts = sourceSignalTargets tr
+    let ssts :: [SourceSignalTarget (CHAIN era)]
+        -- Signal(CHAIN era) = Block (BHeader (EraCrypto era)) era
```
JaredCorduan, line 343: dead?

CHANGELOG.md
```diff
@@ -38,6 +45,8 @@ in the naming of release branches.
 
 ### Changed
 
+- Changed UtxoEnv by replacing pool info (Map (KeyHash 'StakePool c) (PoolParams c)) with DPState
+    to show differences in two NewEpochState's. G
```
JaredCorduan: we need to also mention the change in the API function `evaluateTransactionBalance`.

eras/allegra/impl/src/Cardano/Ledger/Allegra/UTxO.hs
```diff
@@ -19,6 +19,7 @@ import Cardano.Ledger.UTxO (EraUTxO (..))
 
 instance Crypto c => EraUTxO (AllegraEra c) where
   type ScriptsNeeded (AllegraEra c) = ShelleyScriptsNeeded (AllegraEra c)
+  type DepositInfo (AllegraEra c) = DPState c
```
JaredCorduan: I couldn't figure out why we needed to make `DepositInfo` a new type family. could you explain?

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Rules/Ledger.hs
```diff
@@ -101,14 +99,13 @@ ledgerTransition = do
             )
       else pure dpstate
 
-  let DPState dstate pstate = dpstate
+  let DPState dstate _pstate = dpstate
```
teodanciu, line 102: Interesting, _pstate doesn't seem used yet there is no unused warning, I wonder why :thinking:

eras/shelley/impl/src/Cardano/Ledger/Shelley/AdaPots.hs
```diff
@@ -72,6 +93,87 @@ totalAdaES cs =
...
+  {proOutputs :: !Coin, proFees :: Coin, proDeposits :: !Coin}
+
+instance Show Produced where
+  show (Produced (Coin out) (Coin f) (Coin d)) =
+    "Produced(Outputs " ++ show out ++ ", Fees " ++ show f ++ " Deposits " ++ show d ++ ") = " ++ show (out + f + d)
+
+-- =========================
+
+-- | Compute the Coin part of what is consumed by a TxBody, itemized as a 'Consume'
+consumedTxBody ::
```
teodanciu, line 122: Seems surprising to have these two  data types `Consumed` and `Produced` as well as the functions, `consumedTxBody` and `producedTxBody` (which are functions of a transaction)  here in `AdaPots`.   What is their relation with the rest of this module ?

eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState/DPState.hs
```diff
@@ -121,24 +149,26 @@ instance NoThunks (InstantaneousRewards c) => NoThunks (DState c)
...
       (Interns (Credential 'Staking c), Interns (KeyHash 'StakePool c))
   fromSharedPlusCBOR =
-    decodeRecordNamedT "DState" (const 4) $ do
+    decodeRecordNamedT "DState" (const 5) $ do
       unified <- fromSharedPlusCBOR
       fgs <- lift fromCBOR
       gs <- lift fromCBOR
       ir <- fromSharedPlusLensCBOR _1
-      pure $ DState unified fgs gs ir
+      ds <- fromSharedPlusLensCBOR (_1 . toMemptyLens _1 id)
```
teodanciu: Is this allowing this field to not exist and setting it to empty if that's the case? So that the deserialization is backwards compatible?

eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState/DPState.hs
```diff
@@ -235,19 +268,166 @@ instance Default (DState c) where
...
-delegations (DState unified _ _ _) = Delegations unified
+delegations (DState unified _ _ _ _) = Delegations unified
 
 -- | get the actual ptrs map, we don't need a view
 ptrsMap :: DState c -> Map Ptr (Credential 'Staking c)
-ptrsMap (DState (UnifiedMap _ ptrmap) _ _ _) = ptrmap
+ptrsMap (DState (UnifiedMap _ ptrmap) _ _ _ _) = ptrmap
+
+-- ==========================================================
+-- Functions that handle Deposits for stake credetials and key hashes.
```
teodanciu: Another small typo, missing `d` in `credentials`

eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState/DPState.hs
```diff
@@ -235,19 +268,166 @@ instance Default (DState c) where
...
+
+-- | One only pays a deposit on the initial key registration. If the key has been
+--   de-registered it should have been removed from the map. If it hasn't been
+--   de-registered, then it has no effect on the Deposits. In places where this function
+--   is called, there should be an explicit check that the credential is not in the map.
```
teodanciu: I don't understand exactly this: `In places where this function is called, there should be an explicit check that the credential is not in the map.`  For example, `Deleg.delegationTransition` calls this function,  so it's expected that it does a check before calling it?  
I can see this check: 
` eval (hk ∉ dom (rewards ds)) ?! StakeKeyAlreadyRegisteredDELEG hk` , so should it always check in the rewards map ?

eras/shelley/test-suite/src/Test/Cardano/Ledger/Shelley/Generator/Delegation.hs
```diff
@@ -443,8 +443,12 @@ genRetirePool pp poolKeys pState slot =
         (List.find (\x -> hk x == hk') poolKeys)
     EpochNo cepoch = epochFromSlotNo slot
     epochLow = cepoch + 1
-    EpochNo retirementBound = getField @"_eMax" pp
-    epochHigh = cepoch + retirementBound - 1
+    -- EpochNo retirementBound = getField @"_eMax" pp
+    -- epochHigh = cepoch + retirementBound - 1
```
teodanciu: Are the two lines above left by accident here?

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Rules/Ledger.hs
lehins: Because the underscore `_` prefix means: do not produce a warning :wink: 

That is why it is bad to use record fields with underscores prefixes in them, like we had until recently in many types.

eras/shelley/impl/src/Cardano/Ledger/Shelley/AdaPots.hs
```diff
@@ -72,6 +93,87 @@ totalAdaES cs =
...
+data Consumed = Consumed
+  {conInputs :: !Coin, conRefunds :: !Coin, conWithdrawals :: !Coin}
+
+instance Show Consumed where
+  show (Consumed (Coin i) (Coin r) (Coin w)) =
+    "Consumed(Inputs " ++ show i ++ ", Refunds " ++ show r ++ ", Withdrawals " ++ show w ++ ") = " ++ show (i + r + w)
+
+-- | Itemizing what is Produced by a transaction
+data Produced = Produced
+  {proOutputs :: !Coin, proFees :: Coin, proDeposits :: !Coin}
```
lehins: Was `proFees` left lazy on purpose. If so, we should have a comment on why this is necessary, otherwise someone will "fix" it eventually. On the other hand if it doesn't have to be lazy then it is better to make it strict

eras/shelley/impl/src/Cardano/Ledger/Shelley/AdaPots.hs
```diff
@@ -72,6 +93,87 @@ totalAdaES cs =
...
+    HasField "_keyDeposit" pp Coin,
+    HasField "_poolDeposit" pp Coin
+  ) =>
+  TxBody era ->
+  pp ->
+  DPState (EraCrypto era) ->
+  Produced
+producedTxBody txBody pp dpstate = Produced {proOutputs = out, proFees = f, proDeposits = d}
+  where
+    out = Val.coin (balance (txouts txBody))
```
lehins: This will be faster, since it sums the compact coin and it avoids mappending all the multi assets
```suggestion
    out = coinBalance (txouts txBody)
```

eras/shelley/test-suite/src/Test/Cardano/Ledger/Shelley/Rules/TestChain.hs
```diff
@@ -44,32 +46,24 @@ import Cardano.Ledger.Coin
...
-    deltaT,
-    iRReserves,
-    iRTreasury,
-    incrementalStakeDistr,
-    keyRefunds,
-    ptrsMap,
-    rewards,
-    rs,
+import Cardano.Ledger.Shelley.AdaPots (compareAdaPots)
+import Cardano.Ledger.Shelley.LedgerState (DPState (..), DState (..), EpochState (..), IncrementalStake (..), LedgerState (..), NewEpochState (..), PPUPState (..), PState (..), UTxOState (..), completeRupd, credMap, delegations, deltaF, deltaR, deltaT, iRReserves, iRTreasury, incrementalStakeDistr, keyTxRefunds, ptrsMap, rewards, rs, totalTxDeposits)
```
lehins: Switch to multiline. I can't even see the end of it in this github diff

eras/shelley/impl/src/Cardano/Ledger/Shelley/AdaPots.hs
```diff
@@ -72,6 +93,87 @@ totalAdaES cs =
...
+      oneline "rewardsAdaPot" rewardsAdaPot,
+      oneline "utxoAdaPot" utxoAdaPot,
+      oneline "keyDepositAdaPot" keyDepositAdaPot,
+      oneline "poolDepositAdaPot" poolDepositAdaPot,
+      oneline "depositsAdaPot" depositsAdaPot,
+      oneline "feesAdaPot" feesAdaPot
+    ]
+  where
+    n = 25
+    oneline name f = pad n name ++ pad n (show (f x)) ++ pad n (show (f y)) ++ pad n (show (f y <-> f x))
```
lehins: This should be moved to the test module, where it is being used: `Test.Cardano.Ledger.Shelley.Rules.TestChain`

CHANGELOG.md
```diff
@@ -12,6 +12,13 @@ in the naming of release branches.
 
 ### Added
 
+- Add ToExpr (from tree-diff) instances for all types inside NewEpochState
+  - Add module Cardano.Ledger.TreeDiff in cardano-binary. Includes all needed operations, classes, and orphan instances
```
```suggestion
- Add ToExpr (from tree-diff) instances for all types inside NewEpochState
  - Add module Cardano.Ledger.TreeDiff in cardano-data. Includes all needed operations, classes, and orphan instances	
```

eras/allegra/impl/src/Cardano/Ledger/Allegra/UTxO.hs
lehins: I know why :slightly_smiling_face: That is because `UTxO` module was recently moved into `cardano-ledger-core`, which does not know what `DPState` is. 

The consistent way to solve this would be to add ShelleyEraUTxO class into `Cardano.Ledger.Shelley.UTxO` and move the `getConsumedValue` function in that class:
```haskell
class EraUTxO era => ShelleyEraUTxO era where
  getConsumedValue :: HasField "_keyDeposit" pp Coin => pp -> DPState era -> UTxO era -> TxBody era -> Value era
```



The type family approach would make sense only if that type family `DepositInfo` had anyc chance of being set to a different type in different eras, but since it is always the same, type family approach isn't warranted

lehins: Or even better move the whole `EraUTxO` class into `cardano-ledger-shelley` into `Cardano.Ledger.Shelley.UTxO`, this will solve this problem in the easiest possible fashion.

JaredCorduan: ah, makes sense now, thank you!

.gitignore
TimSheard: These are symbolic Unix file-links so clicking on one takes you right to where they link.

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Deleg.hs
```diff
@@ -285,7 +289,8 @@ delegationTransition = do
           u1 = Set.singleton hk UM.⋪ Rewards u0
           u2 = Set.singleton hk UM.⋪ Delegations u1
           u3 = Ptrs u2 UM.⋫ Set.singleton hk
-      pure ds {dsUnified = u3}
+          (_doSomethingWithMeSomeday, newDState) = refundKeyDeposit hk (ds {dsUnified = u3})
```
teodanciu: So the first of the pair is the value that is being refunded, right? Just curious if there is a particular thing to do with it that you were thinking of

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Epoch.hs
```diff
@@ -193,11 +173,13 @@ epochTransition = do
       TRC (epochState', UpecState pp (utxosPpups utxoSt'), ())
   let utxoSt'' = utxoSt' {utxosPpups = ppupSt'}
 
-  let Coin oblgCurr = obligation pp (rewards dstate') (psStakePoolParams pstate'')
-      Coin oblgNew = obligation pp' (rewards dstate') (psStakePoolParams pstate'')
+  let -- Coin oblgCurr = obligationDPState (DPState dstate' pstate'')
```
teodanciu: I guess this particular comment is left by accident here ?

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Epoch.hs
```diff
@@ -193,11 +173,13 @@ epochTransition = do
...
+      -- Now that obligation does not depend on the PParams at all, the current
+      -- and the new are identical, so we don't have to adjust for the difference.
+      Coin oblgNew = obligationDPState (DPState dstate' pstate'')
       Coin reserves = asReserves acnt'
       utxoSt''' = utxoSt'' {utxosDeposited = Coin oblgNew}
-      acnt'' = acnt' {asReserves = Coin $ reserves + oblgCurr - oblgNew}
+      acnt'' = acnt' {asReserves = Coin $ reserves}
```
teodanciu: We could get rid of the `$` :  `Coin reserves`

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Newpp.hs
```diff
@@ -106,29 +99,21 @@ newPpTransition ::
...
       ppNew
       ) <-
     judgmentContext
 
   case ppNew of
     Just ppNew' -> do
-      let Coin oblgCurr = obligation pp (rewView (dsUnified dstate)) (psStakePoolParams pstate)
-          Coin oblgNew = obligation ppNew' (rewView (dsUnified dstate)) (psStakePoolParams pstate)
-          diff = oblgCurr - oblgNew
-          Coin availableReserves = availableAfterMIR ReservesMIR acnt (dsIRewards dstate)
```
teodanciu, line 120: How come we don't have to care about  MIR  here? Does this check follow as a consequence of the new implementation?

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Utxo.hs
```diff
@@ -327,15 +321,26 @@ instance
...
-  renderAssertionViolation AssertionViolation {avSTS, avMsg, avCtx, avState} =
-    "AssertionViolation ("
-      <> avSTS
-      <> "): "
-      <> avMsg
-      <> "\n"
-      <> show avCtx
-      <> "\n"
-      <> show avState
+  renderAssertionViolation
```
teodanciu, line 323: Wondering if these are part of the "API", so any chance someone could be parsing these failures? Should we mention them in the changelog?

eras/shelley/impl/src/Cardano/Ledger/Shelley/AdaPots.hs
teodanciu: Also, the logic seems to be that from `Shelley.UTxO.getConsumedCoin` \ `produced` - could we unify them or it's a problem of accessing one from the other

libs/cardano-ledger-core/src/Cardano/Ledger/Slot.hs
```diff
@@ -74,3 +74,6 @@ epochInfoSize ::
   EpochNo ->
   ShelleyBase EpochSize
 epochInfoSize ei = lift . EI.epochInfoSize ei
+
+-- ================================================
+-- TreeDiff ToExpr instances
```
teodanciu: But no instances defined?

libs/ledger-state/src/Cardano/Ledger/State/Query.hs
```diff
@@ -565,7 +565,8 @@ getDStateNoSharing dstateId = do
               iRTreasury = iRTreasury,
               deltaReserves = dStateIrDeltaReserves,
               deltaTreasury = dStateIrDeltaTreasury
-            }
+            },
+        dsDeposits = Map.empty -- FIXME, HELP ME FIX THIS
```
teodanciu, line 569: What's the story here? We have to somehow populate this before we can get it? 
Do we need to fix this before merging?

teodanciu:
Went through all the code, added some small remarks and asked (likely stupid) questions mostly for my education .
Had no idea this change would impact so much of the codebase and how many things would have to be adapted to the change, a lot of work!

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Rules/Ledger.hs
TimSheard: GHC doesn't complain about unused names that begin with "_"  like _pstate. This is quite usefull as it can be used to make self documenting code.

eras/shelley/impl/src/Cardano/Ledger/Shelley/AdaPots.hs
TimSheard: These are just another kind of Ada Pot, they are the Ada Pots induced by a transaction. We often get errors like produced is not equal to consumed, and the give us two coins, but without telling how those coins were computed. So this makes that explicit.

eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState/DPState.hs
TimSheard: This is a complicated way of telling how the deserialisation can share things, so that there is only one copy in memory.
It is complicated because DPState needs to share two different things (Credential 'Staking c) and (KeyHash 'PoolParam c)
this line tells how to separate the two things, so as to pass just  (Credential 'Staking c) sharing to the DState part of DPState.

eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState/DPState.hs
TimSheard: I changed the comment to read like this.

`-- | One only pays a deposit on the initial key registration. If the key has been

--   de-registered it should have been removed from the map. If it hasn't been

--   de-registered, then it has no effect on the Deposits. Paying a deposit on

--   a credential already registered should have no effect. (In fact it probably

--   should be an error) So to avoid this problem one should make an explicit

--   check that the credential is not registered in the places where this function is called.`

eras/shelley/test-suite/src/Test/Cardano/Ledger/Shelley/Generator/Delegation.hs
TimSheard: I changed this to look like this.

```
epochLow = cepoch + 1
    -- if epochHigh is more than a few epochs above epochLow, then
    -- because our traces are at most, maybe 6 or so traces long,
    -- we will never reap any pools. Choosing a delta between 1 and 10
    -- should give good mix of sometimes reaping, but mostly not.
    epochHigh = cepoch + 10
```

eras/shelley/impl/src/Cardano/Ledger/Shelley/AdaPots.hs
TimSheard: you are right.

eras/shelley/impl/src/Cardano/Ledger/Shelley/AdaPots.hs
TimSheard: good catch

eras/shelley/impl/src/Cardano/Ledger/Shelley/AdaPots.hs
TimSheard: This and several other things are used in the STS  methods 'renderAssertionViolation' , and these occur in the main modules.  These assertions fail quite often, when adding new features or changing old implementations, comparing the AdaPots is often the quickest way to help zero in on what went wrong. Even if they don't appear now, if a failure occurs, adding a temporary 'compareAdaPots' is often just the clue one needs to track things down.  This is a good investment for debugging future failures.  So in short I don't want to move them.

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Deleg.hs
TimSheard: You are right, that is the amount that is being refunded. It is strange that this is not used. But it is accounted for in Cardano.Ledger.Shelley.Rules.Epoch(epochTransition)
This surprised me at first, but after looking into it, it appears this value is not needed here. So I am going to change refundKeyDeposit, to not return a pair.

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Epoch.hs
TimSheard: So I have added an extensive comment here. The thing that makes this tricky, is that before this PR, the utxosDeposited field might change at the epoch boundary if the PParams changed. This is no longer the case, but it still needs to be adjusted for other reasons. So here is the comment I put there 
`  let -- At the epoch boundary refunds are made, so we need to change what
      -- the utxosDeposited field is. The other two places where deposits are
      -- kept (dsDeposits of DState and psDeposits of PState) are adjusted by
      -- the rules, So we can recompute the utxosDeposited field using adjustedDPState
      -- since we have the invariant that: obligationDPState dpstate == utxosDeposited utxostate
      Coin oblgNew = obligationDPState adjustedDPState`

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Epoch.hs
TimSheard: yes, good catch

eras/shelley/impl/src/Cardano/Ledger/Shelley/AdaPots.hs
lehins: Are you sure? I can only see that this function is used in testing. Could you provide a link where this function is being used outside the test framework?

CHANGELOG.md
TimSheard: good catch.

.gitignore
TimSheard: Alexi suggests a way to make this local.

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Rules/Utxow.hs
TimSheard: I thought ormolise did that automatically.  I changed it.

.gitignore
TimSheard: OK I am trying this.

libs/cardano-ledger-binary/src/Cardano/Ledger/TreeDiff.hs
TimSheard: so cardano-data depends on cardano-ledger-binary,  so the tests in cardano-ledger-binary can't depend on this file. There is certainly common stuff that should go in here.

lehins: Yeah, make sense. We can keep it here until we find a better home for it, if ever.

libs/cardano-ledger-core/src/Cardano/Ledger/BaseTypes.hs
TimSheard: done

eras/shelley/impl/src/Cardano/Ledger/Shelley/API/Wallet.hs
TimSheard: The problem is you can't compute this without knowing which Keys and Pools are already registered. We could make a version that took two maps as input, and then use them?
When we know what they need we can make some adjustments.

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Epoch.hs
TimSheard: I guess this was more of a comment explaining why certain things were deleted from the code. In the old system, changing the PParams could change what the obligation was.
Now it only depends on the DPState. So the code is much simpler. I have made this much cleaer in the comment.

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Pool.hs
TimSheard: yes, done

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/PoolReap.hs
TimSheard: Not really,  I thought we might discuss making this change. To me it is really much cleaner

eras/shelley/test-suite/src/Test/Cardano/Ledger/Shelley/Rules/TestChain.hs
TimSheard: done

eras/shelley/test-suite/src/Test/Cardano/Ledger/Shelley/Rules/TestChain.hs
TimSheard: Yes. sometimes it is nice to leave some infrastructure in place.
I moved it to a comment.

eras/shelley/test-suite/src/Test/Cardano/Ledger/Shelley/Rules/TestChain.hs
TimSheard: Just really a comment, so I changed it to look like this

```
    let ssts :: [SourceSignalTarget (CHAIN era)]
        -- In this test, the STS Signal has this definition
        -- Signal(CHAIN era) = Block (BHeader (EraCrypto era)) era
        ssts = sourceSignalTargets tr
```

libs/ledger-state/src/Cardano/Ledger/State/Query.hs
lehins: I think we can fix this in a follow up PR. This is needed for benchmarking the memory overhead of the ledger state, so need to block this PR on this problem.

What's needed for this is an extra table in the db schema and queries for storing and retrieving deposits from the sqlite db. 

@TimSheard Could you create a ticket about this? This would be a good task for @aniketd to pick up, since I know for sure he is familiar with persistent, so this would not be a hard task for him.

CHANGELOG.md

eras/allegra/impl/src/Cardano/Ledger/Allegra/UTxO.hs
TimSheard: This class is defined in the Core, and DPState is not inscope in the core modules, this is a depends on cycle issue. Alexi suggested a work around. I will see if I can make that work. If not, then we will have to keep this. Crossing my fingers.

eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState/DPState.hs
TimSheard: so I added the missing 'n', thanks for pointing it out.

eras/allegra/impl/src/Cardano/Ledger/Allegra/UTxO.hs
lehins: Another alternative would be to move `Caradno.Ledger.Shelley.LedgerStatet.DPState` module into `cardano-ledger-core`

eras/shelley/impl/src/Cardano/Ledger/Shelley/API/Wallet.hs
```diff
@@ -456,7 +456,13 @@ getRewardProvenance globals newepochstate =
...
 -- | The consumed calculation.
 -- Used for the default implentation of 'evaluateTransactionBalance'.
-evaluateConsumed :: EraUTxO era => PParams era -> UTxO era -> TxBody era -> Value era
+evaluateConsumed ::
+  (EraUTxO era, HasField "_keyDeposit" pp Coin) =>
+  pp ->
+  DepositInfo era ->
+  UTxO era ->
+  TxBody era ->
+  Value era
```
lehins: This function has been deprecated, so instead of changing it feel free to remove it

eras/shelley/test-suite/src/Test/Cardano/Ledger/Shelley/Rules/TestChain.hs
TimSheard: I thought ormolou did this. I guess not. Done.

eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState/DPState.hs
```diff
@@ -235,19 +268,166 @@ instance Default (DState c) where
...
+      Map.empty
 
 instance Default (PState c) where
   def =
-    PState Map.empty Map.empty Map.empty
+    PState Map.empty Map.empty Map.empty Map.empty
 
 rewards :: DState c -> ViewMap c (Credential 'Staking c) Coin
-rewards (DState unified _ _ _) = Rewards unified
+rewards (DState unified _ _ _ _) = Rewards unified
```
lehins: I think this is more future proof, so we don't have to change this line when we add fields to DState and we won't have problems if we change the order of fields
```suggestion
rewards = Rewards . dsUnified
```

eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState/DPState.hs
```diff
@@ -235,19 +268,166 @@ instance Default (DState c) where
...
 
 delegations ::
   DState c ->
   ViewMap c (Credential 'Staking c) (KeyHash 'StakePool c)
-delegations (DState unified _ _ _) = Delegations unified
+delegations (DState unified _ _ _ _) = Delegations unified
```
lehins: Same here:
```suggestion
delegations = Delegations . dsUnified
```

eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState/DPState.hs
```diff
@@ -235,19 +268,166 @@ instance Default (DState c) where
...
+  Credential 'Staking (EraCrypto era) ->
+  PParams era ->
+  DState (EraCrypto era) ->
+  DState (EraCrypto era)
+payKeyDeposit cred pp dstate = dstate {dsDeposits = newStake}
+  where
+    stake = dsDeposits dstate
+    newStake = case Map.lookup cred stake of
+      Nothing -> Map.insert cred (getField @"_keyDeposit" pp) stake
+      Just _ -> stake
```
lehins:
```suggestion
    newStake 
      | Map.member cred stake = Map.insert cred (getField @"_keyDeposit" pp) stake
      | otherwise = stake
```

eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState/DPState.hs
```diff
@@ -235,19 +268,166 @@ instance Default (DState c) where
...
+  PParams era ->
+  PState (EraCrypto era) ->
+  PState (EraCrypto era)
+payPoolDeposit keyhash pp pstate = pstate {psDeposits = newpool}
+  where
+    pool = psDeposits pstate
+    newpool = case Map.lookup keyhash pool of
+      Nothing -> Map.insert keyhash (getField @"_poolDeposit" pp) pool
+      Just _ -> pool -- Should it be overwritten with the current  (getField @"_poolDeposit" pp) ?
+      -- things are simpler if it is not.
```
lehins: Same here:
```suggestion
    newpool 
      | Map.member keyhash pool = Map.insert keyhash (getField @"_poolDeposit" pp) pool
       -- Should it be overwritten with the current (getField @"_poolDeposit" pp) ?
       -- things are simpler if it is not:
      | otherwise = pool
```

eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState/DPState.hs
```diff
@@ -235,19 +268,166 @@ instance Default (DState c) where
...
+  numKeys <×> getField @"_keyDeposit" pp
+    <+> snd (foldl' accum (regpools, Coin 0) certs)
+  where
+    numKeys = length $ filter isRegKey certs
+    regpools = psStakePoolParams (dpsPState dpstate)
+    accum (pools, ans) (DCertPool (RegPool poolparam)) =
+      if Map.member (ppId poolparam) pools -- We don't pay a deposit on a pool that is already registered
+        then (pools, ans)
+        else (Map.insert (ppId poolparam) poolparam pools, ans <+> getField @"_poolDeposit" pp)
+    accum ans _ = ans
```
lehins: Would this implementation be slightly simpler to read, granted I understood and reimplemented correctly? 

```suggestion
    <+> Set.size unregPoolIds <×> getField @"_poolDeposit" pp
  where
    numKeys = length $ filter isRegKey certs
    regPools = psStakePoolParams (dpsPState dpstate)
    unregPoolIds = 
      -- We don't pay a deposit on a pool that is already registered
      Set.filter (`Map.notMember` regPools) $
        Set.fromList [ppId poolParam | DCertPool (RegPool poolParam) <- certs]
```

This is just a suggestion, but if you decide not to take my suggestion make sure that the accumulator is strict, other wise it will leak memory, since tuple elements are lazy:
```haskell
    accum (!pools, !ans) (DCertPool (RegPool poolparam)) =
```

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Newpp.hs
TimSheard: Mir is adjusted in the Mir rule. It transfers money to and from treasury and reserves. Since these never change the deposits (and PParams no longer figure into what refunds are), all this code is no longer needed.

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Utxo.hs
TimSheard: No renderAssertionViolation returns a string. It is not meant to return something parseable.

eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState/IncrementalStake.hs
```diff
@@ -196,7 +196,7 @@ incrementalStakeDistr incstake ds ps =
     (VMap.fromMap poolParams)
   where
     UnifiedMap tripmap ptrmap = dsUnified ds
-    PState poolParams _ _ = ps
+    PState poolParams _ _ _ = ps
```
lehins: Future proofing
```suggestion
    poolParams = psStakePoolParams ps 
```

libs/cardano-ledger-core/src/Cardano/Ledger/Slot.hs
TimSheard: Instances have been moved to Cardano.Ledger.TreeDiff, where the ToExpr class is reexported, so any one who knows about TreeDiff sees these instances. So this should be deleted.

libs/ledger-state/src/Cardano/Ledger/State/Query.hs
TimSheard: This will depend on how we translate across the hard fork, so we will have to fill this in later. This whole module is for abstracting the ledgerstate from synching the chain. So this won't break anything for now.

libs/small-steps-test/src/Control/State/Transition/Trace.hs
TimSheard: Well (.->) is used in the Byron era, and adding the right ToExpr instances is too much work.  So for now we need both.

eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState/DPState.hs
```diff
@@ -235,19 +268,166 @@ instance Default (DState c) where
...
+    (coin, newpool) = case Map.lookup keyhash pool of
+      Just c -> (c, Map.delete keyhash pool)
+      Nothing -> (mempty, pool)
+
+-- | Calculate total possible refunds in the system. There is an invariant that
+--   this should be the same as the utxosDeposited field of the UTxOState. Note that
+--   this does not depend upon the current values of the Key and Pool deposits of the PParams.
+obligationDPState :: DPState era -> Coin
+obligationDPState (DPState DState {dsDeposits = keys} PState {psDeposits = stakePools}) =
+  foldl' (<>) (Coin 0) keys <> foldl' (<>) (Coin 0) stakePools
```
lehins:
```suggestion
  sumVal keys <> sumVal stakePools
```

eras/allegra/impl/src/Cardano/Ledger/Allegra/UTxO.hs
TimSheard: This would require moving Certificates to Core, but it looks like DCert may become a type family, so that might complicate things. Of course we might split DPState into two parts. That might not be a bad idea, since only a few functions on DPState need to know about DCert.

eras/shelley/impl/cardano-ledger-shelley.cabal
```diff
@@ -44,6 +44,7 @@ library
     Cardano.Ledger.Shelley.BlockChain
     Cardano.Ledger.Shelley.Core
     Cardano.Ledger.Shelley.CompactAddr
+    Cardano.Ledger.Shelley.Reports
```
lehins: We should move this module into the `other-modules:` section, since this module contains only internal helper functions.

Also to indicate better that this is used only in STS rules, we should move the module into the Rules folder, thus making its new name: `Cardano.Ledger.Shelley.Rules.Reports`

eras/shelley/impl/src/Cardano/Ledger/Shelley/RewardUpdate.hs
```diff
@@ -340,3 +341,9 @@ instance (CC.Crypto c) => FromCBOR (PulsingRewUpdate c) where
       decPS n = Invalid n
 
 instance NFData (PulsingRewUpdate c)
+
+-- ===============================================================
+
+-- | You really don't want to see what is inside this.
```
lehins, line 347: :rofl:

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Newpp.hs
```diff
@@ -106,29 +99,21 @@ newPpTransition ::
   TransitionRule (ShelleyNEWPP era)
 newPpTransition = do
   TRC
-    ( NewppEnv dstate pstate utxoSt acnt,
+    ( NewppEnv dstate pstate utxoSt _,
```
lehins: Instead of ignoring the value in the `NewppEnv` it makes more sense to just remove the `AccountState` from the `NewppEnv` type instead

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Pool.hs
```diff
@@ -215,11 +218,19 @@ poolDelegationTransition = do
...
+                { psStakePoolParams = eval (psStakePoolParams ps ∪ singleton hk poolParam)
+                }
         else do
           tellEvent $ ReregisterPool hk
+          -- hk is already registered, so we want to reregister it. That means adding it to the
+          -- Future pool params (if it is not there already), and overriding the range with the new 'poolParam',
+          -- if it is (using ⨃ ). We must also unretire it, if it has been schedule for retirement.
+          -- The deposit does not change. One pays the deposit just once. Only if it is fully retired
+          -- (i.e. it's deposit has been refunded, and it has been removed from the registered pools).
+          -- does it need to pay a new deposit (at the current current deposit amount). But of course,
```
lehins:
```suggestion
          -- does it need to pay a new deposit (at the current deposit amount). But of course,
```

eras/shelley/test-suite/test/Test/Cardano/Ledger/Shelley/Examples/Combinators.hs
```diff
@@ -175,30 +176,74 @@ newLab b cs =
...
+            (utxosDeposited utxoSt)
+              <+> (length stakes <×> _keyDeposit ppEx)
+              <+> (newcount <×> _poolDeposit ppEx),
+          utxosFees = (utxosFees utxoSt) <+> newFees
+        }
+    ls' = ls {lsUTxOState = utxoSt', lsDPState = dpstate'}
+    -- Count the number of new pools, because we don't take a deposit for existing pools
+    -- This strategy DOES NOT WORK if there are duplicate PoolParams in one call
+    newcount = foldl accum 0 pools
+    accum n x = if Map.member (ppId x) (psDeposits pstate) then (n :: Integer) else n + 1
```
lehins:
```suggestion
    newcount = length $ filter (\x -> ppId x `Map.notMember` psDeposits pstate) pools
```

eras/shelley/test-suite/test/Test/Cardano/Ledger/Shelley/Examples/Combinators.hs
```diff
@@ -175,30 +176,74 @@ newLab b cs =
...
-feesAndDeposits newFees depositChange cs = cs {chainNes = nes'}
+feesAndKeyRefund newFees key cs = cs {chainNes = nes'}
   where
     nes = chainNes cs
     es = nesEs nes
     ls = esLState es
+    DPState dstate pstate = lsDPState ls
+    refund = case Map.lookup key (dsDeposits dstate) of
+      Nothing -> Coin 0
+      Just coin -> coin
```
lehins:
```suggestion
    refund = Map.findWithDefault (Coin 0) key (dsDeposits dstate)
```

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/AggPropTests.hs
```diff
@@ -105,3 +122,61 @@ forAllChainTrace p@(Allegra _) n propf =
...
+-- ===========================================================
+
+-- | Check that the sum of dsDeposits and the psDepoits are equal to the utsosDeposits
+depositInvariant ::
+  SourceSignalTarget (MOCKCHAIN era) ->
+  Property
+depositInvariant SourceSignalTarget {source = mockChainSt} =
+  let LedgerState {lsUTxOState = utxost, lsDPState = DPState dstate pstate} = (esLState . nesEs . mcsNes) mockChainSt
+      allDeposits = utxosDeposited utxost
+      sumCoin m = Map.foldl' (<+>) (Coin 0) m
```
lehins, line 135: This is almost exact monomorphic definition of [`sumVal`](https://github.com/input-output-hk/cardano-ledger/blob/b613b13139c9be9ba19c57cd2ab73a9b64446dd4/libs/cardano-ledger-core/src/Cardano/Ledger/Val.hs#L95)
```suggestion
      sumCoin m = Map.foldl' (<+>) (Coin 0) m
```

CHANGELOG.md
lehins: Also there is some lonely `G` at the end of this entry

lehins (changes requested):
Here are some more comments and suggestions

I don't fully grasp the logic in this PR, since I had limited exposure to staking and delegations, however from what I do understand I don't see any real problems.

libs/cardano-ledger-binary/src/Cardano/Ledger/TreeDiff.hs
TimSheard: So the instances for vector-map are problematic,  cardano-ledger-binary depends on  vector-map, so the best we can do is leave the single orphan instance in TreeDiff file in cardano-binary or make vector-map depend on the external TreeDiff.  So we have done the second choice.

libs/cardano-ledger-binary/src/Cardano/Ledger/TreeDiff.hs
TimSheard: Ok done, makes it a bit harder to import what you want.

libs/cardano-ledger-binary/cardano-ledger-binary.cabal
TimSheard: done

CHANGELOG.md
TimSheard: we didn't move this, so this suggestion is moot.

eras/shelley/impl/src/Cardano/Ledger/Shelley/API/Wallet.hs
TimSheard: OK, but why this one and not other deprecated ones?

eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState/DPState.hs
TimSheard: done

eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState/DPState.hs
TimSheard: done

CHANGELOG.md
lehins: That's right, however there still a small fix is needed: `cardano-binary` -> `cardano-ledger-binary`

eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState/DPState.hs
TimSheard: I am pretty sure your suggestion has it backwards

 ```
newStake = case Map.lookup cred stake of
      Nothing -> Map.insert cred (getField @"_keyDeposit" pp) stake
      Just _ -> stake
```

Should be recast as

 ```
newState | Map.notMember cred stake = Map.insert cred (getField @"_keyDeposit" pp) stake
          | otherwise = state
```

Any way I have changed it.

lehins: Yeah, you are right! That's exactly what I meant to say :wink:

eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState/DPState.hs
TimSheard: done

lehins: Same here, it should have been `notMember`, but I am sure you caught it too

eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState/DPState.hs
TimSheard: So I made the accumulator strict. I think this way is clearer.

lehins: Somehow I knew that you gonna make that decision :smile: but I wanted to make the suggestion anyway

eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState/DPState.hs
TimSheard: some day soon, dsDeposits and psDeposits are going to hold CompactCoin, so unless we make CompactCoin an instance of Val (which I don't think we can because of the method 'invert') am going to leave this as it is.

eras/shelley/impl/src/Cardano/Ledger/Shelley/RewardUpdate.hs
TimSheard: I agree.

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Newpp.hs
TimSheard: done

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Pool.hs
TimSheard: done

eras/shelley/test-suite/test/Test/Cardano/Ledger/Shelley/Examples/Combinators.hs
TimSheard: done

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/AggPropTests.hs
TimSheard: Yes but then again, these things are going to end up CompactCoin, so the changes will be easier to make if this is explicit. Some thing like

Map.foldl' addCompact (CompactCoin 0) m

eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState/IncrementalStake.hs
TimSheard: done

libs/cardano-ledger-binary/src/Cardano/Ledger/TreeDiff.hs
TimSheard: What we are trying to do is hide the surrounding " that are introduced by 'show', we use this only for Nonces and Hashes etc. where we really only want to see some short prefix. But I agree since the comparison is done on the whole thing the shortening it has dangers.
It would be nice if there was a constructor of Expr, that was just a string and a number
 Short :: Int -> String -> Expr.
then we could get the best of both worlds on things built with exprViaShow, compare on the string, but only show the first 'n' things when it is displayed. So for now I will just make it a synonym for exprViaShow.

eras/shelley/impl/cardano-ledger-shelley.cabal
TimSheard: We maybe we will want to use these in tests someday as well?

lehins: When that day comes, we can reexport either the module or individual functions. But it is not the case now, so we should avoid polluting the public space with helper functions. Note that if you make them public in this PR then you should also make an entry about it in the changelog and any time in the future if one of us changes any of these functions then we'd also have to update the changelog and potentially update the package version.

So, we should always strike for exporting the absolute minimum amount of stuff that is required for proper usability of the package.

TimSheard: done

eras/shelley/test-suite/test/Test/Cardano/Ledger/Shelley/Examples/Combinators.hs
TimSheard: I like it the way it is. It suggests that we are accumulating individual pool deposits, which is exactly what we are doing.

lehins: Then you should switch `foldl` to `foldl'` to avoid memory leaks

lehins: Also, you aren't really accumulating deposits, you are just counting here ppIds that aren't in the deposits pstate

TimSheard: I changed it to foldl'

eras/shelley/impl/src/Cardano/Ledger/Shelley/AdaPots.hs
```diff
@@ -72,6 +92,87 @@ totalAdaES cs =
...
+    f = txBody ^. feeTxBodyL
+    d = totalTxDeposits pp dpstate txBody
+
+-- ==============================================
+-- Compare two AdaPots, item by item
+
+pad :: Int -> String -> String
+pad n x = if m < n then x ++ [const ' ' i | i <- [1 .. n - m]] else x
+  where
+    m = length x
```
aniketd: I feel that this would be easier to read. `pad n x = x ++ replicate (n - length x) ' '`

TimSheard: good suggestion

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/AggPropTests.hs
```diff
@@ -105,3 +122,61 @@ forAllChainTrace p@(Allegra _) n propf =
   property $ propf <$> genTrace p n (def {blocksizeMax = 4, slotDelta = (6, 12)}) (initStableFields p)
 forAllChainTrace p@(Shelley _) n propf =
   property $ propf <$> genTrace p n (def {blocksizeMax = 4, slotDelta = (6, 12)}) (initStableFields p)
+
+-- ===========================================================
+
+-- | Check that the sum of dsDeposits and the psDepoits are equal to the utsosDeposits
```
aniketd:
```suggestion
-- | Check that the sum of dsDeposits and the psDepoits are equal to the utxosDeposits
```

aniketd (approved):
:+1:

.gitignore
```diff
@@ -58,3 +58,6 @@ specs/**/.ghc.environment.x86_64-linux-*
 
 ## python for generating the doc site's transitive dependencies
 .venv
+
+## Symbolic links
+excludesfile = /home/.gitignore_global
```
lehins: This should not be here. Ping me if you can't figure it out
```suggestion
```

eras/shelley/impl/src/Cardano/Ledger/Shelley/AdaPots.hs
```diff
@@ -72,6 +92,85 @@ totalAdaES cs =
...
+      oneline "rewardsAdaPot" rewardsAdaPot,
+      oneline "utxoAdaPot" utxoAdaPot,
+      oneline "keyDepositAdaPot" keyDepositAdaPot,
+      oneline "poolDepositAdaPot" poolDepositAdaPot,
+      oneline "depositsAdaPot" depositsAdaPot,
+      oneline "feesAdaPot" feesAdaPot
+    ]
+  where
+    n = 25
+    oneline name f = pad n name ++ pad n (show (f x)) ++ pad n (show (f y)) ++ pad n (show (f y <-> f x))
```
lehins: I'll post here one more time. This should go into `Test.cardano.Ledger.Shelley.Rules.TestChain`, because that is the only place where it is being used

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/AggPropTests.hs
TimSheard: done

.gitignore
TimSheard: got it

eras/shelley/impl/src/Cardano/Ledger/Shelley/AdaPots.hs
TimSheard: NO,  the problem is that you never debug test failures. This is the most useful tool ever to debug test failures, since it itemizes where things have changed. I use it in debugging by putting trace calls which use this function. In order for me to make such trace calls, this function must be in scope. If it is in the tests, it is not in scope in the src files. 

Once the the failures have been debugged, the trace calls are removed, and the it APPEARS that this has only one use. But in fact I have used it many times. It needs to be in scope. It is such a little thing, but it is the exception that breaks the rule.

TimSheard: So now this has been moved to Cardano.Ledger.Shelley.Internal(compareAdaPots)

aniketd: Could we keep this part (slightly edited) in the haddock for the function :slightly_smiling_face: ?
> This is the most useful tool ever to debug test failures, since it itemizes where things have changed. Use it for debugging by putting trace calls composed with this function.

eras/allegra/impl/src/Cardano/Ledger/Allegra/UTxO.hs
TimSheard: So after rebasing on master that had been merged with https://github.com/input-output-hk/cardano-ledger/pull/3202, It has been very easy to remove DepositInfo.
DPState is now in scope in the core, so DepositInfo is not needed.

lehins (approved):
Awesome work!!!

PR #3128, opened 2022-11-07
Draft: Enable UndecidableSuperClasses where needed

milloni:
I am getting several errors related to UndecidableSuperClasses when building with GHC 9.4

My suspicion is that this is due to stricter class recursion checking rules in GHC 9.4 when type families are involved

I can get it to compile by enabling the UndecidableSuperClasses language extension

Would you accept this patch or alternative could you suggest a different fix?

Regards
Max

### DIFF
--- a/.github/workflows/haskell.yml
+++ b/.github/workflows/haskell.yml
@@ -17,7 +17,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        ghc: ["8.10.7", "9.2.4"]
+        ghc: ["8.10.7", "9.2.4", "9.4.2"]
         os: [ubuntu-latest]
 
     steps:
--- a/cabal.project
+++ b/cabal.project
@@ -98,3 +98,16 @@ source-repository-package
   --sha256: 18apsg2lqjv9cc29nbd3hzj2hqhksqjj0s4xp2rdv8cbd27racjh
   subdir:
     cborg
+
+source-repository-package
+  type: git
+  location: https://github.com/milloni/cardano-base
+  tag: a82587d621d6b64cf5001cf2845a958887aef9cc
+  --sha256: 1zx8li963wfc7kq5p2nxdj8lnw2nj59jli47bww7h2silbr60apf
+  subdir:
+    binary
+    binary/test
+    cardano-crypto-class
+    cardano-crypto-praos
+    cardano-crypto-tests
+    slotting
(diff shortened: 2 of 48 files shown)


lehins:
Do you mind adding ghc `9.4.3` to CI here: https://github.com/input-output-hk/cardano-ledger/blob/836452c1ff7b38b16ca5d2d3b0c58383ef59f512/.github/workflows/haskell.yml#L20

We might not promise GHC 9.4 support at this point, but it would not hurt keeping the code forward compatible proactively, especially if all that is required right now is the patch in this PR

milloni:
@lehins Sure, I will add ghc 9.4 to the CI. Not sure that's the only patch needed, we will see soon..

lehins:
Looks like there is more work that needs to be done in order to add GHC-9.4 support than just these extensions.

If you willing to do some of that work into getting this compatibility in that would be awesome.

The first step would be to switch the upper bound on `base < 4.17` to `base < 5` for all packages in the `cardano-base` repo and in this repo as well.

It is ok if you are not up to the task. Unfortunately we can't merge this PR in such case. The rule is pretty simple, if we can't test it we can't merge it.

Please, let me know what you want to do. You got my full support in this effort, if you decide to pursue it.

milloni:
@lehins No problem, I'm working on it :)

milloni:
@lehins I pushed a commit that uses a patched version of cardano-base, but with that I can't access the CI log anymore

If I click on "details" I get a message from Google Workspace:

```
Access blocked: Production can only be used within its organization
```

Could you please advise what to do

lehins:
@milloni CI is now running.

milloni:
@lehins Could you advise how to plug my fork of cardano-base into this build? Commit `942ef225ae90c237fa8bfe85f75e6519702038ca` is how I tried to do it but as you can see that didn't work

lehins:
In order for cabal to pickup your changes in cardano-base you need to add the `subdir` for all packages that you've modified in `cardano-base`. You also need the hash nix `<hash-value>`, which you can get by running in cardano-base directory `nix-prefetch-git $PWD --rev HEAD` or waiting for cabal to complain at you with the correct value when building

```cabal
source-repository-package
  type: git
  location: https://github.com/milloni/cardano-base
  tag: a82587d621d6b64cf5001cf2845a958887aef9cc
  --sha256: <hash value>
  subdir:
    base-deriving-via
    binary
    binary/test
    cardano-crypto-class
    cardano-crypto-praos
    cardano-strict-containers
    measures
    slotting
    heapword
```

JaredCorduan:
it looks like this PR has gone stale. I'll close it for now, but would be happy for it to be revived.

PR #3133, opened 2022-11-10
1.1.x + 9.2

angerman:
Let's see what CI thinks of this


angerman:
according to hydra [native.checks.tests.cardano-ledger-alonzo-test.cardano-ledger-alonzo-test.x86_64-darwin](https://hydra.iohk.io/build/21516623) fails. So something specific to x86_64-darwin.

angerman:
I can't reproduce the failure :-/

angerman:
Looks like hydra is now also happy with this. 🤯

angerman:
The only job that hasn't build _for hours_ now is the `build-version` job, which is _still_ pending. At this point, I'll just merge this.

angerman:
All green!

.github/workflows/haskell.yml
```diff
@@ -62,7 +62,7 @@ jobs:
       id: setup-haskell
       with:
         ghc-version: ${{ matrix.ghc }}
-        cabal-version: 3.4.0.0
+        cabal-version: 3.6.2.0
```
teodanciu, line 65: Wonder whether the [cabal version in shell.nix](https://github.com/input-output-hk/cardano-ledger/pull/3133/files#diff-c09cc8dc973415b1221308559b96942a1c4b872e690bf360fc22d9b42822d519L65) doesn't need updating as well

eras/alonzo/impl/cardano-ledger-alonzo.cabal
```diff
@@ -69,7 +69,6 @@ library
     cardano-crypto-class,
     cardano-ledger-core,
     cardano-ledger-shelley-ma,
-    cardano-prelude,
```
teodanciu, line 72: Strange that this wasn't included in the earlier commit which removes the dependency on cardano-prelude. Interesting that it compiled with the dependency even though it was unnecessary (unless it became unnecessary because of something in this particular commit :thinking: )

eras/byron/ledger/impl/test/Test/Options.hs
```diff
@@ -18,9 +20,13 @@ module Test.Options
...
 where
 
-import Cardano.Prelude hiding (Option)
 import Hedgehog (Gen, Group (..), Property, PropertyT, TestLimit, withTests)
 import Hedgehog.Internal.Property (GroupName (..), PropertyName (..))
+#if __GLASGOW_HASKELL__ >= 900
+import Cardano.Prelude
+#else
+import Cardano.Prelude hiding (Option)
+#endif
```
angerman, line 29: @teodanciu And here we re-introduce Cardano-prelude 🙈

eras/alonzo/impl/cardano-ledger-alonzo.cabal
angerman: See my other comment about re-introducing our dependency on Cardano-prelude. In the end it only cherry-picked some parts. For the ledger/1.1.x branch I hope that's ok.

.github/workflows/haskell.yml
angerman: I guess we can upgrade the shell.

eras/byron/ledger/impl/test/Test/Options.hs
Soupstraw: It was there before, I just had to add pre-processor directives to prevent a name clash with the new Haskell prelude.

angerman: Oh, yea it was there. I don't deny that. It's just that we didn't fully apply the remove Cardano-prelude patch. But for the release/1.1.x branch, I still don't think it matters. This is just our jumping board for 9.2 compatibility until we have a release from ledger @ master, to push into network on node.

PR #3135, opened 2022-11-10
Add PlutusDebug to TransactionScriptFailure

### LINKED ISSUE #3122: Add script input to the error type in `evaluateTransactionExecutionUnits`
When building a transaction using cardano api or cli, `Cardano.Ledger.Alonzo.Tools.evaluateTransactionExecutionUnits` is called to evaluate the plutus scripts in the transaction. To make it easier to debug script failures, the input to the Plutus script (Datum, Redeemer, ScriptContext) should be included in the error (`TransactionScriptFailure`), so that users can reproduce and debug the failure.

aniketd:
Closes #3122

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@ in the naming of release branches.
 
 ### Added
 
+- Add `PlutusDebug` to `TransactionScriptFailure.ValidationFailedV{1|2}`: #3135
+  - Also add a harmless dummy `Show` instance for `PlutusDebug` with a constant `"PlutusDebug Omitted"`
 - Create a `cardano-ledger-binary` package that does versioned serialization as a
   replacement for `cardano-binary` package: #3063, #3078
 - Start on the `cardano-ledger-api` package and implement
--- a/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Tools.hs
+++ b/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Tools.hs
@@ -24,6 +24,7 @@ import Cardano.Ledger.Alonzo.Tx (AlonzoEraTx, ScriptPurpose (..), rdptr)
 import Cardano.Ledger.Alonzo.TxBody (AlonzoEraTxOut (..))
 import Cardano.Ledger.Alonzo.TxInfo
   ( ExtendedUTxO (txscripts),
+    PlutusDebug (..),
     TranslationError,
     VersionedTxInfo (..),
     exBudgetToExUnits,
@@ -71,9 +72,9 @@ data TransactionScriptFailure c
   | -- | Missing datum.
     MissingDatum !(DataHash c)
   | -- | Plutus V1 evaluation error.
-    ValidationFailedV1 !PV1.EvaluationError ![Text]
+    ValidationFailedV1 !PV1.EvaluationError ![Text] PlutusDebug
   | -- | Plutus V2 evaluation error.
-    ValidationFailedV2 !PV2.EvaluationError ![Text]
+    ValidationFailedV2 !PV2.EvaluationError ![Text] PlutusDebug
   | -- | A redeemer points to a transaction input which is not
     --  present in the current UTxO.
     UnknownTxIn !(TxIn c)
@@ -196,7 +197,7 @@ evaluateTransactionExecutionUnitsWithLogs pp tx utxo ei sysS costModels = do
       RdmrPtr ->
       (Data era, ExUnits) ->
       Either (TransactionScriptFailure (EraCrypto era)) ([Text], ExUnits)
-    findAndCount pparams info pointer (rdmr, _) = do
+    findAndCount pparams info pointer (rdmr, exunits) = do
       (sp, mscript, sh) <-
         note (RedeemerPointsToUnknownScriptHash pointer) $
           Map.lookup pointer ptrToPlutusScript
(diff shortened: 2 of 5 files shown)


eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Tools.hs
```diff
@@ -22,16 +24,7 @@ import Cardano.Ledger.Alonzo.Scripts
...
-  ( ExtendedUTxO (txscripts),
-    TranslationError,
-    VersionedTxInfo (..),
-    exBudgetToExUnits,
-    transExUnits,
-    transProtocolVersion,
-    txInfo,
-    valContext,
-  )
+import Cardano.Ledger.Alonzo.TxInfo (ExtendedUTxO (txscripts), PlutusDebug (..), TranslationError, VersionedTxInfo (..), exBudgetToExUnits, transExUnits, transProtocolVersion, txInfo, valContext)
```
JaredCorduan: let's keep this vertical

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Tools.hs
```diff
@@ -87,6 +80,12 @@ data TransactionScriptFailure c
     NoCostModelInLedgerState !Language
   deriving (Show, Eq)
 
+-- | This 'Show' instance is intentionally an orphan,
+-- meant to be used only as part of 'TransactionScriptFailure'
+-- See the comment at the definition of 'PlutusDebug' about
+-- the cost of 'Show'
```
JaredCorduan: Hmm, I am not sure what to do here. The show here is not in a testing module, and could therefore be abused.

Maybe we could require end-users to use a standalone function that we provide? What do you think @lehins ?

JaredCorduan (commented):
@aniketd looks great, thank you! I'm not sure what to do about the `Show` instance, but besides that it looks great. Did we abandon the idea to use a GADT to keep better consistency with the plutus versions? (maybe that was to be a follow up PR?).

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Tools.hs
lehins: Do we need to have `Show` instance for `TransactionScriptFailure`?

If we only need it for tests, then moving as standalone into a test package is the best approach IMHO

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Tools.hs
aniketd: Done!

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Tools.hs
aniketd: The `Show` on `TransactionScriptFailure` is forcing it. Would it hurt if we removed it? (Will try and see)

aniketd:
@JaredCorduan I started with the GADT-based implementation, but it felt unrelated to the changes required for the issue. I will make a follow up PR with those changes :+1:

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Tools.hs
lehins: That being said, if we really need this Show instance here or downstream, then adding one for `PlutusDebug` would be acceptable, we just need to be really sure not to use it. One way or another it should not be an orphan instance in this module

lehins: > The Show on TransactionScriptFailure is forcing it.

I understand that, that's why I asked: "Do we need to have Show instance for TransactionScriptFailure?"

> Would it hurt if we removed it?

Try and remove it. See what breaks

JaredCorduan: We could also ask @Jimbo4350 and @ak3n if they need a `Show` instance for `TransactionScriptFailure` (they are the two folks that I am aware of who use the tool).

aniketd: I'm pushing a version with the `Show` instance for both: `PlutusDebug` and `TransactionScriptFailure`, moved to the `Test` module, where those are used, for review.

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Alonzo/Tools.hs
```diff
@@ -1,9 +1,14 @@
 {-# LANGUAGE DataKinds #-}
+{-# LANGUAGE DeriveAnyClass #-}
```
Jimbo4350: Are all of these extensions needed? Shouldn't `StandaloneDeriving` + `FlexibleInstances` be enough?

Jimbo4350 (approved):

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Alonzo/Tools.hs
aniketd: @Jimbo4350 you are right. Thanks! :raised_hands:

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ in the naming of release branches.
 
 ### Added
 
+- Add `PlutusDebug` to `TransactionScriptFailure.ValidationFailedV{1|2}`: #3135
```
lehins, line 15: It is also worth adding here that the `Show` instance for `TransactionScriptFailure` has been removed

lehins (approved):
LGTM

CHANGELOG.md
aniketd: Note: Since the latest discussion we have decided to keep a dummy `Show` instance for `PlutusDebug`.

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Tools.hs
aniketd: Update: Following the discussion today, I have pushed a dummy `Show` instance for `PlutusDebug`, keeping the `Show` instance for `TransactionScriptFailure` intact.

JaredCorduan (approved):
:raised_hands:

PR #3136, opened 2022-11-10
Some adjustment to the .ghcid files

aniketd:
- Remove `-fwrite-interface`.
  - This avoids `cabal build` from facing problems with incompatible interface files.
- Add `--reverse-errors`.
  - This shows errors in reverse order so that the developer can see the "head" of the error-list in the visible part of terminal scroll window, best combined with `--clear`.


JaredCorduan (approved):
:heart:

PR #3137, opened 2022-11-10
modernize tullia CI setup

dermetfan:
- only one task that builds everything for faster invocation
- build for all supported systems
  - report one status per system


teodanciu (approved):
Looks good to me and works well - I rebased other branches on top with successful builds :+1:

PR #3078, opened 2022-10-12
Add versioned CBOR instances for crypto primitives

lehins:
This PR moves all ToCBOR and FromCBOR instances from `cardano-crypto-class` and `cardano-crypto-praos` packages. Actual decoders are still defined in `cardano-crypto-class`

 Instances for `PraosVRF` in this PR depends on  https://github.com/input-output-hk/cardano-base/pull/316

PR #3138, opened 2022-11-10
Switch to versioned cbor

### LINKED ISSUE #3014: Versioned serialization schemes for types

lehins:
This PR fixes #3014 
It implements the switch from `cardano-binary` to `cardano-ledger-binary`, which was implemented in #3063 and #3078
There are also a few improvements to `cardano-ledger-binary`, see the commit message and changes to changelog for the full list of changes.

~Things left to do~:

* [x] Update changelog
* [x] Cleanup git commit history
* [x] Update PR message


eras/alonzo/test-suite/test/Test/Cardano/Ledger/Alonzo/Serialisation/Tripping.hs
```diff
@@ -8,160 +8,155 @@
...
 
-tripping :: (Eq src, Show src, ToCBOR src, FromCBOR src) => src -> Property
-tripping = trippingF roundTrip
+-- tripping :: (Eq src, Show src, ToCBOR src, FromCBOR src) => src -> Property
+-- tripping = trippingF roundTrip
 
-trippingWithTwiddling :: (Twiddle src, FromCBOR src) => (src -> src -> Property) -> src -> Property
-trippingWithTwiddling comp x = property $ trippingM roundTripWithTwiddling comp x
+-- trippingWithTwiddling :: (Twiddle src, FromCBOR src) => (src -> src -> Property) -> src -> Property
+-- trippingWithTwiddling comp x = property $ trippingM roundTripWithTwiddling comp x
```
JaredCorduan, line 82: dead code, or  turned off temporarily?

JaredCorduan (approved):
This is the dawn of a glorious new day! Thank you for this colossal effort, @lehins !

There was a call to review a lot of this PR earlier today, and I've just finished going over it myself. LGTM!

eras/conway/impl/src/Cardano/Ledger/Conway/Translation.hs
```diff
@@ -113,13 +110,21 @@ instance
...
-    pure $ Tx $ AlonzoTx bdy txwits validating aux
+    pure $ Tx $ AlonzoTx txBody txWits validating auxData
+
+-- bdy <- translateViaCBORAnn "txbody" $ Alonzo.body tx
+-- txwits <- translateViaCBORAnn "txwitness" $ Alonzo.wits tx
+-- aux <- case Alonzo.auxiliaryData tx of
+--   SNothing -> pure SNothing
+--   SJust axd -> SJust <$> translateViaCBORAnn "auxiliarydata" axd
+-- let validating = Alonzo.isValid tx
+-- pure $ Tx $ AlonzoTx bdy txwits validating aux
```
JaredCorduan: dead?

eras/shelley/impl/src/Cardano/Ledger/Shelley/Genesis.hs
```diff
@@ -196,20 +209,33 @@ instance Era era => FromJSON (ShelleyGenesis era) where
...
+        <*> obj
+          .: "maxKESEvolutions"
+        <*> obj
+          .: "slotLength"
+        <*> obj
+          .: "updateQuorum"
+        <*> obj
+          .: "maxLovelaceSupply"
+        <*> obj
+          .: "protocolParams"
```
JaredCorduan: did you still want to collapse this?

eras/alonzo/test-suite/test/Test/Cardano/Ledger/Alonzo/Serialisation/Tripping.hs
lehins: Yes, I intentionally commented it out. There will be a subsequent PR that fixes this part properly. I am partially done with its implementation too. These twiddler tests were skipped anyways, so I didn't see any harm in commenting them out in this PR.

eras/conway/impl/src/Cardano/Ledger/Conway/Translation.hs
lehins: Dead indeed. Removed

eras/shelley/impl/src/Cardano/Ledger/Shelley/Genesis.hs
lehins: Yes, this was a formatting accident. Fixed

eras/alonzo/test-suite/test/Test/Cardano/Ledger/Alonzo/Serialisation/Tripping.hs
```diff
@@ -8,160 +8,155 @@
...
-  where
-    f' a = pure (f a, toLazyByteString $ toCBOR a)
+-- trippingF ::
+--   (Eq src, Show src, Show target, ToCBOR src) =>
+--   (src -> Either target (BSL.ByteString, src)) ->
+--   src ->
+--   Property
+-- trippingF f x = runIdentity $ trippingM f' (===) x
+--   where
+--     f' a = pure (f a, toLazyByteString $ toCBOR a)
```
aniketd, line 36: The reason behind leaving these comments here is not clear because there is no mention. May be a special comment at the top will make it clear.

libs/cardano-ledger-binary/test-lib/Test/Cardano/Ledger/Binary/Random.hs
```diff
@@ -0,0 +1,51 @@
...
+    MkGen (\r _n -> runStateGen_ r (uniformShortByteString k))
+  {-# INLINE uniformShortByteString #-}
+
+-- | It is possible to use a hash of a binary representation of any type as a source of
+-- randomness, since hash value by its definiteion is uniformly distributed.
+mkDummyHash :: forall h a b. (HashAlgorithm h, ToCBOR a) => a -> Hash h b
+mkDummyHash = coerce . hashWithEncoder @h shelleyProtVer toCBOR
+
+-- | Use a hash of the binary representation of a type as a seed to construct `StdGen`,
+-- taht can be further used to generate random values.
```
aniketd:
```suggestion
-- that can be further used to generate random values.
```

aniketd (approved):
This is massive! :+1:

libs/cardano-ledger-test/src/Test/Cardano/Ledger/BaseTypes.hs
```diff
@@ -39,66 +40,61 @@ boundedRationalTests ::
...
-          localOption (mkTimeout 500000) $
-            testCase "Check divergence" $
-              expectLeft (boundedFromJSON "10e1234567893456")
-        ]
-          ++ [ testCase testName $ expectLeft $ boundedFromJSON invalidInput
-               | (testName, invalidInput) <- badJSONValues
-             ],
-      testProperty "CBOR roundtrip" $ \(br :: a) ->
-        either (error . show) (br ==) $ Binary.decodeFull (Binary.serialize br),
-      testProperty "CBOR BoundedRational roundtrip" $ \(br :: a) ->
```
teodanciu, line 97: This `CBOR BoundedRational roundtrip`  test removed on purpose?

lehins: Yes, `boundedRationalToCBOR`/`boundedRationalFromCBOR` functions have been removed

eras/shelley/impl/src/Cardano/Ledger/Shelley/TxWits.hs
```diff
@@ -215,12 +212,12 @@ pattern ShelleyTxWits {addrWits, scriptWits, bootWits} <-
             if null x then Nothing else Just (encodeWord ix <> enc x)
           l =
             catMaybes
-              [ encodeMapElement 0 encodeFoldable awits,
-                encodeMapElement 1 encodeFoldable scriptWitMap,
-                encodeMapElement 2 encodeFoldable bootstrapWits
+              [ encodeMapElement 0 toCBOR awits,
+                encodeMapElement 1 toCBOR (Map.elems scriptWitMap),
```
teodanciu, line 216: The previous  version `encodeFoldable scriptWitMap` was also encoding only the  values of the map?

teodanciu (approved):
Looks EPIC to me,  nothing pertinent that I can comment I'm afraid. 

I'm in awe at the tenacity that this work must have required!  :bowing_woman:

eras/shelley/impl/src/Cardano/Ledger/Shelley/TxWits.hs
lehins: That's right. Foldable instance for Map only concerns the elements, not the keys. So with this approach it is a bit more explicit that we discard the keys.

teodanciu: Makes sense, thanks for the explanation

PR #3140, opened 2022-11-11
[not to merge] TICKF calculatePoolDistr

aniketd:


libs/ledger-state/src/Cardano/Ledger/State/Query.hs
```diff
@@ -36,14 +36,16 @@ import qualified Data.VMap as VMap
...
 
 -- Populate database
 
 insertGetKey ::
   ( MonadIO m,
     PersistUniqueWrite backend,
     PersistRecordBackend record backend,
-    AtLeastOneUniqueKey record
+    AtLeastOneUniqueKey record,
+    SafeToInsert record
```
aniketd: @nfrisby Hi! :slightly_smiling_face: I was wondering if we could do without this contraint. If we could, we would not need to bump the `persistent < 2.14` constraint present in `cabal.project`.

nfrisby: I only put it there to fix an error I was seeing. Might be totally inconsequential now.

In fact, in did a little more dev today on top of a rebase that @amesgen did for me. Looks like he removed it already 👍

nfrisby:
I just forced pushed a commit that has two improvements:

- Esgen had rebased it onto the `node/1.35.3` tag in this repo.

- See commit message 0b8ca35854234517f8387135d30c7cbca090bf54 for more details; I _think_ that commit introduces a `TICKF` implementation that is sound and could not ever be any more minimal. Please do read the commit message, though.

nfrisby (changes requested):
I'm Requesting Changes just to further ensure this does not get merged as-is.

History: Aniket opened this PR on a branch that I had shared with the Ledger team (presumably so he could post some questions about it). I just pushed to the branch before realizing that Aniket open this PR using my branch instead of his branch (which is `aniket/`+my branch name). So: sorry for force-pushing to your PR (but my branch 😵).

libs/ledger-state/src/Cardano/Ledger/State/Query.hs
amesgen: > Looks like he removed it already

Exactly, as I got compilation errors with the new constraint :+1:

aniketd: Got it! Thanks!

aniketd:
> So: sorry for force-pushing to your PR (but my branch dizzy_face).

Your pushes are welcome @nfrisby :slightly_smiling_face: . I had also pushed another branch, just to avoid making any changes to yours, here: #3141, that is being rebased on top of `master`. I will get your recent changes also merged into that one soon.

nfrisby:
@amesgen has confirmed that this latest commit does eliminate stuttering he was observing in the related benchmark he's been developing. (We don't have any overall speedup measurements yet; we'll share once we do).

So: **as long as we confirm my extensive simplified here doesn't introduce any unsound logic**, I think this establishes a best-case limit for the `TICKF` performance we could achieve without explicitly incrementalizing the epoch boundary computations.

cc @JaredCorduan

JaredCorduan:
> So: **as long as we confirm my extensive simplified here doesn't introduce any unsound logic**, I think this establishes a best-case limit for the `TICKF` performance we could achieve without explicitly incrementalizing the epoch boundary computations.

I think our protections are:

* human code reviews
* the property tests that use the traces (with the hope that any change to the pool distribution would be caught)
* syncing mainnet

I will take responsibility for these, thank you again for the fantastic idea!

amesgen:
> syncing mainnet

FTR: I successfully synced to mainnet via a modified 1.35.3 (only changed the ledger pin to this branch).

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Tick.hs
```diff
@@ -163,6 +168,64 @@ validatingTickTransition nes slot = do
...
+    let pp' =
+            maybe pp id
+          $ votedValue
+              (proposals $ _ppups $ lsUTxOState ls)
+              pp
+              (fromIntegral coreNodeQuorum)
+
+    pure $! nes {
+        nesPd = pd'
+      , nesEs = adoptGenesisDelegs
```
JaredCorduan, line 224: just for posterity, let me mention that the genesis delegation mapping also needs to be updating in the `e =/= e + 1` case. The re-delegation of the genesis keys is staged for adoption one security window from the time the delegation certificate is submitted (ie not necessarily on an epoch boundary).

hooray for property tests! :tada:

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Tick.hs
```diff
@@ -163,6 +168,64 @@ validatingTickTransition nes slot = do
...
+    -- We can skip 'MIR', 'POOLREAP', 'UPEC' (which calls 'NEWPP'); we don't
+    -- need to do the checks: if the checks would fail, then the node will fail
+    -- in the 'TICK' rule if it ever tries to validate blocks for which the
+    -- return value here was used to validate their headers.
+
+    coreNodeQuorum <- liftSTS $ asks quorum
+
+    let pp' =
+            maybe pp id
+          $ votedValue
```
JaredCorduan, line 217: just for posterity, let me mention that just because the core nodes reach quorum on protocol parameter update doesn't mean it will be accepted. At the epoch boundary, `UPEC` does some final sanity checks. (This turns out to be problematic, see [ADR-3](https://github.com/input-output-hk/cardano-ledger/pull/3190/files#diff-fb71f1e0f9d8938e0f10af3f59332fa73b3c539fbb4fc4065c623b23327a12b3R56) and will be changing in the next ledger era.

hooray for property tests! :tada:

JaredCorduan:
replaced by #3209

Issue #3034, opened 2022-09-12
Use of calculatePoolDistr on the epoch boundary


JaredCorduan:
Large computations on the epoch boundary are problematic. In the past they have lead to delayed block production in new epochs. For this reason, we have made the reward calculation and the (per stake credential) stake distribution calculation incremental. At least one remaining epoch boundary calculation remains, however, that we should consider handling in a better way, namely the per-stake-pool stake distribution calculation, performed by `calculatePoolDistr` in the `NEWEPOCH` transition.

Note that the consensus layer calls the `TICK` rule when processing block headers. This rule is usually pretty trivial, but when a block header corresponds to the first block of a new epoch, this rule includes extra work for the epoch boundary. When nodes are at the end of an epoch, and are trying to determine if they can make the first block of the next epoch, the `TICK` rule can be called multiple times. This can exacerbate situations like https://github.com/input-output-hk/cardano-node/issues/4421.

Even though `calculatePoolDistr` is called on the epoch boundary, the data that it uses for the calculation has been stable for a long time (`7k/f`-many slots).

We should consider calculating the per-stake-pool stake distribution well in advance of the epoch boundary, so that it is guaranteed to only every be computed once.

AndrewWestberg:
Please prioritize this fix high if you can. This is a lot of lost blocks across many pools that happen in this epoch transition dead-zone.

JaredCorduan:
yes, it's a high priority. we are  building out some bench-marking tools for this, so that we don't end up on a wild goose chase, and so that we can be sure we've solved it.

JaredCorduan:
We have a great stop-gap measure for this issue, thanks to @nfrisby . The idea is in this branch: https://github.com/input-output-hk/cardano-ledger/tree/nfrisby/share-tickf-pooldistr-calculation Someone from the ledger team will see the idea through to completion (Nick may have done all the work already, we just need to be careful).

PR #3141, opened 2022-11-11
TICKF calculatePoolDistr

### LINKED ISSUE #3132: earlier thunk for the per-stake-pool stake distribution
This is a stop-gap measure for the problem described in #3034.

Create the thunk for the `calculatePoolDistr` when the corresponding snapshot is created (when crossing the previous epoch boundary), not in the `NEWEPOCH` rule. This way, when the consensus layer forecasts across the boundary multiple times, the expensive computation only happens once per epoch boundary per node per fork.

aniketd:
- [ ] Get approval about exhaustiveness of changes required
- [x] Revert `pattern` implementation and move `calc` from within `SnapShotRaw` to within `SnapShots` as it was earlier.
- [ ] Update changelog
- [ ] Squash all history and write an elaborate commit message with links to issues

References
- https://github.com/input-output-hk/cardano-node/issues/4421
- #3034 
- #3132 

Closes #3132


aniketd:
@nfrisby I have just merged your latest changes on to this one, which remains rebased on top of `master`. Would really like to have your eyes go over this.

cabal.project
```diff
@@ -62,6 +62,72 @@ benchmarks: true
 -- The only sensible test display option
 test-show-details: streaming
 
+-- <<<<<<< HEAD
```
lehins: Conflict

eras/shelley-ma/impl/src/Cardano/Ledger/Mary/Translation.hs
```diff
@@ -153,7 +153,7 @@ instance Crypto c => TranslateEra (MaryEra c) ShelleyTxOut where
 
 instance Crypto c => TranslateEra (MaryEra c) UTxO where
   translateEra ctxt utxo =
-    return $ UTxO (translateEra' ctxt <$> unUTxO utxo)
+    return $ UTxO (translateEra' ctxt `Map.map` unUTxO utxo)
```
lehins: `Map` has `Functor` instance, why this change?

cabal.project
aniketd: Yes :slightly_smiling_face: , I kept it just in case they were needed from the original PR. Going to remove them in case they aren't needed. Tests seem to be failing. Will look closer what's happening.

eras/shelley-ma/impl/src/Cardano/Ledger/Mary/Translation.hs
aniketd: This change was part of the original PR. Will check the implementation and then revert this change if it is unnecessary :+1:

aniketd: `fmap` is indeed defined in terms of `map`, for both `v0.6.6` as well as `0.6.5.1` of `containers`. Reverting this change. :+1:

amesgen: @nfrisby might clarify, but I think he did this to prevent extra thunks as the `Functor Map` instance is lazy in the values, while `Data.Map.Strict.map` is strict.

lehins: @amesgen That is very good point. We probably want to add a comment about this, so no-one switches it back to `fmap` in the future.

aniketd: @amesgen @nfrisby sorry didn't catch that while I was looking. Thanks for the clarification!

lehins: Here is a separate PR: https://github.com/input-output-hk/cardano-ledger/pull/3143

nfrisby: Yes, that diff should only be in PR 3143, not this one. It was merely a "leftover" in my commits.

aniketd (commented):
@nfrisby this commit moves the `PoolDistr` from the record `SnapShots` to inside each individual `SnapShot` as suggested by @TimSheard. I'd be glad to receive both your reviews on this change.

aniketd:
@JaredCorduan I have updated this PR to reflect our latest discussion.
1. Move `PoolDistr` back to `SnapShots` from `SnapShot`
2. Remove `SnapShotRaw` and corresponding `pattern synonym`

This is ready for your review!

JaredCorduan:
replaced by #3209

PR #3143, opened 2022-11-15
Fix the thunks due to lazy map

yogeshsajanikar:
- Use Data.Map.Strict.map instead of lazy <$> to fix the thunk issue


nfrisby (commented):
Looks good to me 👏 (but I'm not Approving, since I'm not an owner here)

lehins (approved):
Looks great. Thank you for investigating this

PR #3162, opened 2022-11-21
Clean hardfork switches

### LINKED ISSUE #3104: Clean up protocol version switching
The [HardForks](https://github.com/input-output-hk/cardano-ledger/blob/8817304c300dd60099054dd66d9a1ecbd33bfb3a/eras/shelley/impl/src/Cardano/Ledger/Shelley/HardForks.hs) and [SoftForks](https://github.com/input-output-hk/cardano-ledger/blob/8817304c300dd60099054dd66d9a1ecbd33bfb3a/eras/shelley/impl/src/Cardano/Ledger/Shelley/SoftForks.hs) modules contain all the occurrences in the code where we switch on the protocol version. In at least one occasion, namely [translateTimeForPlutusScripts](https://github.com/input-output-hk/cardano-ledger/blob/8817304c300dd60099054dd66d9a1ecbd33bfb3a/eras/shelley/impl/src/Cardano/Ledger/Shelley/HardForks.hs#L61), the problematic behavior (ie the logic prior to the version where we switch) never arose on mainnet. Therefore some of these switching functions can be removed, and the code can be cleaned up.

Go through all of the functions in HardForks and SoftForks and see which ones are semantically meaningful on mainnet (ie see if mainnet syncs without it, up until the version in question). Note that none of this logic touches the Byron era, so to speed up the testing it is recommended to start with a snapshot of the ledger just before the Shelley hardfork.

Leave comments on the remaining functions that we must maintain forever, noting that they can never be removed (we might, in the future, have to add more of these with the hope of being able to remove them).

teodanciu:
Closes: https://github.com/input-output-hk/cardano-ledger/issues/3104

Answers the question: which HardFork switches can we remove from the code without affecting the state

How I checked:

1. Checked out cardano-node (tag 1.35.3) and synced up with mainnet until slot 76138419 (the tip at the time)
2. Checked out db-analyser (part of ouroboros-network), at branch release/cardano-node-1.35.x 
2. Took snapshots of the ledger state, using db-analyser, corresponding to the first slots of each era
3. For each tested switch:
	- set the switch to True in cardano-ledger in a branch off tag node/1.35.3
	- run db-analyzer pointing to the aforementioned cardano-ledger branch:
		- analyzing from a slot that is not affected by the switch (earlier than the protocol version from the condition)
		- taking a snapshot of the ledger state for slots corresponding to the snapshots taken at step 2
	- this db-analyzer had two possible outcomes: an error or a snapshot. The error is an indication that the switch cannot be removed. In case there was no error, the sha256 sum of the produced snapshot was compared with that of the corresponding snapshot produced at step 2 (using the code with the HardFork switch intact)

### 1. allowMIRTransfer pp = pvMajor (getField @"_protocolVersion" pp) > 4  :negative_squared_cross_mark: 

Ran: 
```
cabal run ouroboros-consensus-cardano-tools:db-analyser2 -- --db /home/teodora/cardano-mainnet/db/db-mainnet --only-immutable-db --analyse-from 23068800 --store-ledger 39916975  cardano --config /home/teodora/cardano-mainnet/config.json
```
because 23068800 is the first slot in Mary (version 4) and 39916975 is the first slot in the next era

This resulted in an error:  `ExtValidationErrorLedger (HardForkLedgerErrorFromEra S (S (S (Z (WrapLedgerErr {unwrapLedgerErr = BBodyError (BlockTransitionError [LedgersFailure (LedgerFailure (DelegsFailure (WithdrawalsNotInRewardsDELEGS (fromList [(RewardAcnt {getRwdNetwork = Mainnet, getRwdCred = KeyHashObj (KeyHash "7f5287ac749c44f8b517165a35ea60a0a165904b25da29749d2366e4")},Coin 220383)]))))])})))))`

So I concluded it can't be removed. 

--------------------------------------------------------

### 2. aggregatedRewards pp = pvMajor (getField @"_protocolVersion" pp) > 2 :negative_squared_cross_mark: 

set to True, trying to take a snapshot from 4492800 (first Shelley slot(v2)) to 16588800 (first Allegra slot (v3)) resulted in: 
` ExtValidationErrorLedger (HardForkLedgerErrorFromEra S (Z (WrapLedgerErr {unwrapLedgerErr = BBodyError (BlockTransitionError [LedgersFailure (LedgerFailure (DelegsFailure (WithdrawalsNotInRewardsDELEGS (fromList [(RewardAcnt {getRwdNetwork = Mainnet, getRwdCred = KeyHashObj (KeyHash "24f1b3791037e55363815fc4b788b886cd1d82a2f2d39e17196cd1a4")},Coin 431236474)]))))])})))`

--------------------------------------------------------

### 3. validatePoolRewardAccountNetID pp = pvMajor (getField @"_protocolVersion" pp) > 4  :negative_squared_cross_mark: 
set to True, trying to take a snapshot from 23068800 (first Mary slot(v2)) to 39916975 (first Alonzo slot (v3)) resulted in: `ExtValidationErrorLedger (HardForkLedgerErrorFromEra S (S (S (Z (WrapLedgerErr {unwrapLedgerErr = BBodyError (BlockTransitionError [LedgersFailure (LedgerFailure (DelegsFailure (DelplFailure (PoolFailure (WrongNetworkPOOL Mainnet Testnet (KeyHash "ae8ee8f5676a3c3aadefd0ee7adb57ebbd39dd7126c8517e718dc4c1"))))))])})))))`

--------------------------------------------------------

### 4. forgoRewardPrefilter pp = pvMajor (getField @"_protocolVersion" pp) > 6 :negative_squared_cross_mark: 
  
set to True, trying to take a snapshot from 43372972 (v6) to 76138419 resulted in:  `ExtValidationErrorLedger (HardForkLedgerErrorFromEra S (S (S (S (Z (WrapLedgerErr {unwrapLedgerErr = BBodyError (BlockTransitionError [ShelleyInAlonzoPredFail (LedgersFailure (LedgerFailure (DelegsFailure (DelplFailure (DelegFailure (StakeKeyNonZeroAccountBalanceDELEG (Just (Coin 658031))))))))])}))))))`

--------------------------------------------------------
### 5. translateTimeForPlutusScripts pp = pvMajor (getField @"_protocolVersion" pp) > 5  :heavy_check_mark: 
set to True, a snapshot of 76138419 (latest slot that I had synced with mainnet) was generated with the same sha256 of the snapshot taken with the switch enabled (9aafb9e44500f84d2a59dfd8019934dd42a947edf976c35d90fa6bb4136604dd  ) 

--------------------------------------------------------
### 6. allowScriptStakeCredsToEarnRewards pp = pvMajor (getField @"_protocolVersion" pp) > 4 :heavy_check_mark: 
set to True, a snapshot of 76138419 with the original sha256 is generated

--------------------------------------------------------
### 7. allowOutsideForecastTTL pp =let mv = pvMajor (getField @"_protocolVersion" pp) in mv == 5 || mv == 6 :heavy_check_mark: 
set to True, a snapshot  of 76138419 with the original sha256 is generated

--------------------------------------------------------
### 8. missingScriptsSymmetricDifference  :grey_question: 
I forgot why we can't switch this off even if the state is the same

### DIFF
--- a/eras/shelley/impl/src/Cardano/Ledger/Shelley/Rewards.hs
+++ b/eras/shelley/impl/src/Cardano/Ledger/Shelley/Rewards.hs
@@ -220,14 +220,12 @@ instance CC.Crypto c => FromCBOR (PoolRewardInfo c) where
       )
 
 notPoolOwner ::
-  HasField "_protocolVersion" pp ProtVer =>
-  pp ->
   PoolParams c ->
   Credential 'Staking c ->
   Bool
-notPoolOwner pp pps = \case
+notPoolOwner pps = \case
   KeyHashObj hk -> hk `Set.notMember` ppOwners pps
-  ScriptHashObj _ -> HardForks.allowScriptStakeCredsToEarnRewards pp
+  ScriptHashObj _ -> True
 
 -- | The stake pool member reward calculation
 rewardOnePoolMember ::
@@ -257,7 +255,7 @@ rewardOnePoolMember
   rewardInfo
   hk
   (Coin c) =
-    if prefilter && notPoolOwner pp (poolPs rewardInfo) hk && r /= Coin 0
+    if prefilter && notPoolOwner (poolPs rewardInfo) hk && r /= Coin 0
       then Just r
       else Nothing
     where
--- a/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/PlutusScriptApi.hs
+++ b/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/PlutusScriptApi.hs
@@ -165,7 +165,7 @@ collectTwoPhaseScriptInputs ei sysS pp tx utxo =
             (Right [])
   where
     scriptsAvailable = txscripts utxo tx
-    txinfo lang = txInfo pp lang ei sysS utxo tx
+    txinfo lang = txInfo lang ei sysS utxo tx
     AlonzoScriptsNeeded scriptsNeeded' = getScriptsNeeded utxo (tx ^. bodyTxL)
     neededAndConfirmedToBePlutus =
       mapMaybe (knownToNotBe1Phase scriptsAvailable) scriptsNeeded'
--- a/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Rules/Utxo.hs
+++ b/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Rules/Utxo.hs
@@ -76,7 +76,6 @@ import Cardano.Ledger.Rules.ValidationMode
     runTest,
     runTestOnSignal,
   )
-import Cardano.Ledger.Shelley.HardForks (allowOutsideForecastTTL)
 import qualified Cardano.Ledger.Shelley.LedgerState as Shelley
 import Cardano.Ledger.Shelley.Rules (ShelleyUtxoPredFailure, UtxoEnv)
 import qualified Cardano.Ledger.Shelley.Rules as Shelley
@@ -341,22 +340,18 @@ validateOutsideForecast ::
     AlonzoEraTxWits era,
     EraTx era
   ) =>
-  PParams era ->
   EpochInfo (Either a) ->
   -- | Current slot number
   SlotNo ->
   SystemStart ->
   Tx era ->
   Test (AlonzoUtxoPredFailure era)
-validateOutsideForecast pp ei slotNo sysSt tx =
+validateOutsideForecast ei slotNo sysSt tx =
   {-   (_,i_f) := txvldt tx   -}
   case tx ^. bodyTxL . vldtTxBodyL of
     ValidityInterval _ (SJust ifj)
       | not (nullRedeemers (tx ^. witsTxL . rdmrsTxWitsL)) ->
-          let ei' =
-                if allowOutsideForecastTTL pp
-                  then unsafeLinearExtendEpochInfo slotNo ei
-                  else ei
+          let ei' = unsafeLinearExtendEpochInfo slotNo ei
            in -- ◇ ∉ { txrdmrs tx, i_f } ⇒
               failureUnless (isRight (epochInfoSlotToUTCTime ei' sysSt ifj)) $ OutsideForecast ifj
     _ -> pure ()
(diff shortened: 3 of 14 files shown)


eras/shelley/impl/src/Cardano/Ledger/Shelley/HardForks.hs.orig
```diff
@@ -0,0 +1,73 @@
+{-# LANGUAGE DataKinds #-}
```
JaredCorduan: I suspect you didn't mean to commit this file. :)

JaredCorduan (approved):
looks great, thank you @teodanciu , wonderful clean up! (my only comment is about the `orig` file).

eras/shelley/impl/src/Cardano/Ledger/Shelley/HardForks.hs.orig
teodanciu: Eek

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rewards.hs
```diff
@@ -220,14 +220,12 @@ instance CC.Crypto c => FromCBOR (PoolRewardInfo c) where
...
-  HasField "_protocolVersion" pp ProtVer =>
-  pp ->
   PoolParams c ->
   Credential 'Staking c ->
   Bool
-notPoolOwner pp pps = \case
+notPoolOwner pps = \case
   KeyHashObj hk -> hk `Set.notMember` ppOwners pps
-  ScriptHashObj _ -> HardForks.allowScriptStakeCredsToEarnRewards pp
+  _ -> True
```
lehins: It is often good to exhaustively list all patterns. Whenever the sum type is modified we get warnings about non-exhaustive pattern matches in all such places. This is not terribly important, since I don't think we plan on changing `Credential` type, but just in case it is a good practice in general
```suggestion
  ScriptHashObj _ -> True
```

lehins (approved):
This is a great cleanup!!! Thank you. 
Just one minor comment

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rewards.hs
teodanciu: I was actually between the two, but didn't think this. Good to know, thank you

PR #3163, opened 2022-11-21
Update ruby version in github actions, in order to fix the build 

teodanciu:


lehins (approved):
Thank you for taking care of it!

lehins:
I merged this PR without waiting for cicero, since this PR only affects GithubActions config, which has passed.

We need these changes to get faster feedback in other PRs, since GA is still much faster

JaredCorduan:
thank you @teodanciu !!

PR #3172, opened 2022-11-29
Remove `EncodeMint`/`DecodeMint` classes

lehins:
These two type classes are remnants of the time when mint field was a `Value` instead of `MultiAsset`. No longer needed. With classes gone also are gone the bogus instances for `Coin`

### DIFF
--- a/eras/shelley-ma/impl/src/Cardano/Ledger/Mary/Value.hs
+++ b/eras/shelley-ma/impl/src/Cardano/Ledger/Mary/Value.hs
@@ -57,12 +57,7 @@ import Cardano.Ledger.Coin (Coin (..), CompactForm (..), integerToWord64)
 import Cardano.Ledger.Compactible (Compactible (..))
 import qualified Cardano.Ledger.Crypto as CC
 import Cardano.Ledger.Shelley.Scripts (ScriptHash (..))
-import Cardano.Ledger.Val
-  ( DecodeMint (..),
-    DecodeNonNegative (..),
-    EncodeMint (..),
-    Val (..),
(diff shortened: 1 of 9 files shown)


eras/shelley-ma/impl/src/Cardano/Ledger/Mary/Value.hs
```diff
@@ -152,11 +147,11 @@ instance Group (MultiAsset c) where
...
 
-instance CC.Crypto c => DecodeMint (MultiAsset c) where
-  decodeMint = decodeMultiAssetMaps decodeIntegerBounded64
+instance CC.Crypto c => FromCBOR (MultiAsset c) where
+  fromCBOR = decodeMultiAssetMaps decodeIntegerBounded64
 
-instance CC.Crypto c => EncodeMint (MultiAsset c) where
-  encodeMint = encodeMultiAssetMaps
+instance CC.Crypto c => ToCBOR (MultiAsset c) where
+  toCBOR = encodeMultiAssetMaps
```
teodanciu, line 154: Is this `encodeMultiAssetMaps` function even necessary? It seems to be doing only `toCBOR m`.  But maybe for symmetry with `decodeMultiAssetMaps`

lehins: Yeah, I think it is good to keep it for consistency with `decodeMultiAssetMaps`. Both of these are internal functions, so there is no worry of anyone using it outside this module.

teodanciu (approved):
Looks good to me

PR #3174, opened 2022-11-30
Don't deserialize Conway outputs with pointer addresses

### LINKED ISSUE #3161: prevent ptr address from being created in conway
Starting in the conway ledger era, do not allow the creation of transaction outputs that contain pointer addresses. Implementation options include: 1) adding a new predicate failure, perhaps the `UTXO` rule, 2) use the version CBOR to prevent pointer address from being deserialized starting in major protocol version 9.

teodanciu:
Closes https://github.com/input-output-hk/cardano-ledger/issues/3161

### DIFF
--- a/eras/babbage/impl/src/Cardano/Ledger/Babbage/TxOut.hs
+++ b/eras/babbage/impl/src/Cardano/Ledger/Babbage/TxOut.hs
@@ -40,7 +40,7 @@ module Cardano.Ledger.Babbage.TxOut
 where
 
 import Cardano.Crypto.Hash
-import Cardano.Ledger.Address (Addr (..))
+import Cardano.Ledger.Address (Addr (..), isPtrAddr)
 import Cardano.Ledger.Alonzo.Data
   ( BinaryData,
     Data,
@@ -62,11 +62,7 @@ import qualified Cardano.Ledger.Alonzo.TxBody as Alonzo
 import Cardano.Ledger.Babbage.Era (BabbageEra)
 import Cardano.Ledger.Babbage.PParams (_coinsPerUTxOByte)
 import Cardano.Ledger.Babbage.Scripts ()
-import Cardano.Ledger.BaseTypes
-  ( StrictMaybe (..),
-    maybeToStrictMaybe,
-    strictMaybeToMaybe,
-  )
+import Cardano.Ledger.BaseTypes (StrictMaybe (..), maybeToStrictMaybe, natVersion, strictMaybeToMaybe)
 import Cardano.Ledger.Binary
   ( Annotator (..),
     Decoder,
@@ -87,6 +83,7 @@ import Cardano.Ledger.Binary
     getDecoderVersion,
     interns,
     peekTokenType,
+    whenDecoderVersionAtLeast,
   )
 import Cardano.Ledger.Binary.Coders
 import Cardano.Ledger.Coin (Coin (..))
@@ -107,7 +104,7 @@ import Cardano.Ledger.Val
     Val (..),
   )
 import Control.DeepSeq (NFData (rnf), rwhnf)
-import Control.Monad ((<$!>))
+import Control.Monad (when, (<$!>))
 import qualified Data.ByteString.Lazy as LBS
 import Data.Maybe (fromMaybe)
 import qualified Data.Text as T
@@ -435,14 +432,21 @@ instance
 -- toCBOR (TxOutCompactRefScript addr cv d rs) = encodeTxOut addr cv d (SJust rs)
 
 instance
-  ( Era era,
-    Val (Value era),
-    FromCBOR (Annotator (Script era)),
-    DecodeNonNegative (Value era)
+  ( EraTxOut era,
+    Core.TxOut era ~ BabbageTxOut era,
+    FromCBOR (Annotator (Script era))
   ) =>
   FromCBOR (BabbageTxOut era)
   where
-  fromCBOR = fromCborTxOutWithAddr fromCborBackwardsBothAddr
+  fromCBOR = do
+    txOut <- fromCborTxOutWithAddr fromCborBackwardsBothAddr
+    -- Prevent pointer addresses from being deserialized, starting with version 9.
+    -- This implementation is according to ADR #5
+    whenDecoderVersionAtLeast (natVersion @9) $ do
+      when
+        (isPtrAddr (txOut ^. addrTxOutL))
+        (fail "Pointer addresses not supported starting with version 9")
+    pure txOut
 
 instance
   ( Era era,
--- a/eras/conway/impl/src/Cardano/Ledger/Conway/Era.hs
+++ b/eras/conway/impl/src/Cardano/Ledger/Conway/Era.hs
@@ -25,12 +25,12 @@ import Cardano.Ledger.Shelley.Rules
 
 -- =====================================================
 
--- | The Babbage era
+-- | The Conway era
 data ConwayEra c
 
 instance CC.Crypto c => Era (ConwayEra c) where
   type EraCrypto (ConwayEra c) = c
-  type ProtVerLow (ConwayEra c) = 8
+  type ProtVerLow (ConwayEra c) = 9
 
 type instance Value (ConwayEra c) = MaryValue c
 
--- a/eras/conway/test-suite/src/Test/Cardano/Ledger/Conway/Serialisation/Roundtrip.hs
+++ b/eras/conway/test-suite/src/Test/Cardano/Ledger/Conway/Serialisation/Roundtrip.hs
@@ -1,27 +1,73 @@
 {-# LANGUAGE AllowAmbiguousTypes #-}
+{-# LANGUAGE DataKinds #-}
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE RankNTypes #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE TypeApplications #-}
+{-# LANGUAGE TypeFamilies #-}
 
 module Test.Cardano.Ledger.Conway.Serialisation.Roundtrip (allprops) where
 
+import Cardano.Ledger.Babbage (BabbageTxOut)
+import Cardano.Ledger.BaseTypes (natVersion)
 import Cardano.Ledger.Conway.Genesis (ConwayGenesis (..))
-import Cardano.Ledger.Core (Era (..), eraProtVerHigh)
+import Cardano.Ledger.Core (Era (..), EraTxOut, Script, TxOut, Value, addrTxOutL, eraProtVerHigh)
 import Data.Data (Proxy (..), typeRep)
+import Data.List (isInfixOf)
+import Lens.Micro ((&), (.~))
+import Test.Cardano.Ledger.Alonzo.AlonzoEraGen ()
 import Test.Cardano.Ledger.Binary.RoundTrip
 import Test.Cardano.Ledger.Conway.Serialisation.Generators ()
+import Test.Cardano.Ledger.Shelley.ConcreteCryptoTypes (Mock)
+import Test.Cardano.Ledger.Shelley.Serialisation.EraIndepGenerators (genNonPtrAddress, genPtrAddress)
+import Test.Cardano.Ledger.ShelleyMA.Serialisation.Generators ()
+import Test.QuickCheck (Arbitrary, Property, counterexample, forAll)
 import Test.Tasty (TestTree, testGroup)
 import Test.Tasty.QuickCheck (testProperty)
 
 allprops ::
   forall e.
-  ( Era e
+  ( EraTxOut e,
+    TxOut e ~ BabbageTxOut e,
+    Mock (EraCrypto e),
+    Arbitrary (Value e),
+    Arbitrary (Script e)
   ) =>
   TestTree
 allprops =
   testGroup
     (show $ typeRep (Proxy @e))
     [ testProperty "ConwayGenesis" $
-        roundTripCborExpectation @(ConwayGenesis (EraCrypto e)) (eraProtVerHigh @e)
+        roundTripCborExpectation @(ConwayGenesis (EraCrypto e)) (eraProtVerHigh @e),
+      testProperty "conway/TxOut with pointer addresses" (ptrAddrTxOutNotDeserialized @e),
+      testProperty "conway/TxOut without pointer addresses" (nonPtrAddrTxOutDeserialized @e)
     ]
+
+ptrAddrTxOutNotDeserialized ::
+  forall e.
+  EraTxOut e =>
+  TxOut e ->
+  Property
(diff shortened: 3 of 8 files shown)


eras/babbage/impl/src/Cardano/Ledger/Babbage/TxOut.hs
```diff
@@ -435,14 +432,28 @@ instance
...
+  ( EraTxOut era,
+    Core.TxOut era ~ BabbageTxOut era,
+    FromCBOR (Annotator (Script era))
   ) =>
   FromCBOR (BabbageTxOut era)
   where
-  fromCBOR = fromCborTxOutWithAddr fromCborBackwardsBothAddr
+  fromCBOR = do
+    txOut <- fromCborTxOutWithAddr fromCborBackwardsBothAddr
+    whenDecoderVersionAtLeast (natVersion @9) $ do
```
JaredCorduan, line 445: thinking out loud here, maybe we could leave a comment which points to the ADR I just made? :thinking: 

https://github.com/input-output-hk/cardano-ledger/pull/3192

JaredCorduan (approved):
The change looks great, thank you! We should update the changelog, and also maybe leave a bit more of a trail about why this change was needed (below I suggested a comment linking to the ADR that I just put up, which in turn links to the CIP...).

eras/babbage/impl/src/Cardano/Ledger/Babbage/TxOut.hs
```diff
@@ -435,14 +432,28 @@ instance
...
+    pure txOut
+
+checkNotPtrAddr ::
+  forall era s.
+  (EraTxOut era) =>
+  Core.TxOut era ->
+  Decoder s ()
+checkNotPtrAddr txOut = do
+  let addr = txOut ^. addrTxOutL
+  if (isPtr addr)
```
lehins, line 454: Redundant parenthesis
```suggestion
  if isPtr addr
```

eras/babbage/impl/src/Cardano/Ledger/Babbage/TxOut.hs
```diff
@@ -435,14 +432,28 @@ instance
...
+    Core.TxOut era ~ BabbageTxOut era,
+    FromCBOR (Annotator (Script era))
   ) =>
   FromCBOR (BabbageTxOut era)
   where
-  fromCBOR = fromCborTxOutWithAddr fromCborBackwardsBothAddr
+  fromCBOR = do
+    txOut <- fromCborTxOutWithAddr fromCborBackwardsBothAddr
+    whenDecoderVersionAtLeast (natVersion @9) $ do
+      checkNotPtrAddr txOut
```
lehins: There is no need to create helper function for everything. It is better to write it like so:
```suggestion
      when (isPtr (txOut ^. addrTxOutL)) $
        fail "Pointer addresses are not supported starting with protocol version 9"
```

eras/babbage/impl/src/Cardano/Ledger/Babbage/TxOut.hs
```diff
@@ -435,14 +432,28 @@ instance
...
+checkNotPtrAddr ::
+  forall era s.
+  (EraTxOut era) =>
+  Core.TxOut era ->
+  Decoder s ()
+checkNotPtrAddr txOut = do
+  let addr = txOut ^. addrTxOutL
+  if (isPtr addr)
+    then fail "Pointer addresses not supported starting with version 9"
+    else pure ()
```
lehins: Not needed
```suggestion
```

eras/conway/impl/src/Cardano/Ledger/Conway/Era.hs
```diff
@@ -25,12 +25,12 @@ import Cardano.Ledger.Shelley.Rules
...
 -- =====================================================
 
--- | The Babbage era
+-- | The Conway era
 data ConwayEra c
 
 instance CC.Crypto c => Era (ConwayEra c) where
   type EraCrypto (ConwayEra c) = c
-  type ProtVerLow (ConwayEra c) = 8
+  type ProtVerLow (ConwayEra c) = 9
```
lehins, line 33: Could you also update the upper version in babbage: https://github.com/input-output-hk/cardano-ledger/blob/6b41318acd067a38f3f12faa2397a9f81921ded2/eras/babbage/impl/src/Cardano/Ledger/Babbage/Era.hs#L36-L37

```diff
+ type ProtVerHigh (BabbageEra c) = 8
```

eras/conway/test-suite/src/Test/Cardano/Ledger/Conway/Serialisation/Roundtrip.hs
```diff
@@ -23,5 +32,25 @@ allprops =
...
     ]
+
+ptrAddrTxOutNotDeserialized :: Property
+ptrAddrTxOutNotDeserialized =
+  forAll (genPtrAddress @(EraCrypto Conway)) $ \ptrAddr txOut ->
+    do
+      let ptrAddrTxOut = txOut & addrTxOutL .~ ptrAddr
+      let messageMatch s = "Pointer addresses not supported starting with version 9" `isInfixOf` s
+      property $
+        case (roundTrip (natVersion @9) (cborTrip @(BabbageTxOut Conway)) ptrAddrTxOut) of
```
lehins: Redundant parenthesis
```suggestion
        case roundTrip (natVersion @9) (cborTrip @(BabbageTxOut era)) ptrAddrTxOut of
```

eras/conway/test-suite/src/Test/Cardano/Ledger/Conway/Serialisation/Roundtrip.hs
```diff
@@ -23,5 +32,25 @@ allprops =
...
     [ testProperty "ConwayGenesis" $
-        roundTripCborExpectation @(ConwayGenesis (EraCrypto e)) (eraProtVerHigh @e)
+        roundTripCborExpectation @(ConwayGenesis (EraCrypto e)) (eraProtVerHigh @e),
+      testProperty "conway/TxOut with pointer addresses" ptrAddrTxOutNotDeserialized,
+      testProperty "conway/TxOut without pointer addresses" nonPtrAddrTxOutDeserialized
     ]
+
+ptrAddrTxOutNotDeserialized :: Property
+ptrAddrTxOutNotDeserialized =
+  forAll (genPtrAddress @(EraCrypto Conway)) $ \ptrAddr txOut ->
```
lehins: In order to make this test forward compatible we need to expose the `era` parameter and not restrict it to `Conway` only.  After all this test should pass for all eras that use `BabbageTxOut` .

```suggestion
ptrAddrTxOutNotDeserialized :: EraTxOut era => BabbageTxOut era -> Property
ptrAddrTxOutNotDeserialized txOut =
  forAll (genPtrAddress @era) $ \ptrAddr ->
```

eras/conway/test-suite/src/Test/Cardano/Ledger/Conway/Serialisation/Roundtrip.hs
```diff
@@ -23,5 +32,25 @@ allprops =
...
+      let ptrAddrTxOut = txOut & addrTxOutL .~ ptrAddr
+      let messageMatch s = "Pointer addresses not supported starting with version 9" `isInfixOf` s
+      property $
+        case (roundTrip (natVersion @9) (cborTrip @(BabbageTxOut Conway)) ptrAddrTxOut) of
+          Left (RoundTripFailure _ _ _ _ _ _ (Just e)) -> messageMatch (show e)
+          _ -> False
+
+nonPtrAddrTxOutDeserialized :: Property
+nonPtrAddrTxOutDeserialized =
+  forAll (genNonPtrAddress @(EraCrypto Conway)) $ \nonPtrAddr txOut ->
```
lehins: Same comments as for the above property test applies here.

eras/conway/test-suite/src/Test/Cardano/Ledger/Conway/Serialisation/Roundtrip.hs
```diff
@@ -23,5 +32,25 @@ allprops =
...
+ptrAddrTxOutNotDeserialized :: Property
+ptrAddrTxOutNotDeserialized =
+  forAll (genPtrAddress @(EraCrypto Conway)) $ \ptrAddr txOut ->
+    do
+      let ptrAddrTxOut = txOut & addrTxOutL .~ ptrAddr
+      let messageMatch s = "Pointer addresses not supported starting with version 9" `isInfixOf` s
+      property $
+        case (roundTrip (natVersion @9) (cborTrip @(BabbageTxOut Conway)) ptrAddrTxOut) of
+          Left (RoundTripFailure _ _ _ _ _ _ (Just e)) -> messageMatch (show e)
+          _ -> False
```
lehins:
```suggestion
          _ -> counterexample "This was not suppose to parse" False
```

eras/conway/test-suite/src/Test/Cardano/Ledger/Conway/Serialisation/Roundtrip.hs
```diff
@@ -23,5 +32,25 @@ allprops =
...
+      testProperty "conway/TxOut with pointer addresses" ptrAddrTxOutNotDeserialized,
+      testProperty "conway/TxOut without pointer addresses" nonPtrAddrTxOutDeserialized
     ]
+
+ptrAddrTxOutNotDeserialized :: Property
+ptrAddrTxOutNotDeserialized =
+  forAll (genPtrAddress @(EraCrypto Conway)) $ \ptrAddr txOut ->
+    do
+      let ptrAddrTxOut = txOut & addrTxOutL .~ ptrAddr
+      let messageMatch s = "Pointer addresses not supported starting with version 9" `isInfixOf` s
```
lehins: If we don't add these explanation we'll not know why exactly our test is failing
```suggestion
      let messageMatch s = counterexample ("Unexpected decoder failure message: " ++ s) $
                                              "Pointer addresses not supported starting with version 9" `isInfixOf` s
```

eras/conway/test-suite/src/Test/Cardano/Ledger/Conway/Serialisation/Roundtrip.hs
```diff
@@ -23,5 +32,25 @@ allprops =
...
+      property $
```
lehins: No longer needed with my other comments
```suggestion
```

eras/conway/test-suite/src/Test/Cardano/Ledger/Conway/Serialisation/Roundtrip.hs
```diff
@@ -23,5 +32,25 @@ allprops =
...
+        case (roundTrip (natVersion @9) (cborTrip @(BabbageTxOut Conway)) ptrAddrTxOut) of
+          Left (RoundTripFailure _ _ _ _ _ _ (Just e)) -> messageMatch (show e)
+          _ -> False
+
+nonPtrAddrTxOutDeserialized :: Property
+nonPtrAddrTxOutDeserialized =
+  forAll (genNonPtrAddress @(EraCrypto Conway)) $ \nonPtrAddr txOut ->
+    do
+      let nonPtrAddrTxOut = txOut & addrTxOutL .~ nonPtrAddr
+      property $ isRight $ (roundTrip (natVersion @9) (cborTrip @(BabbageTxOut Conway)) nonPtrAddrTxOut)
```
lehins: Redundant parenthesis, also isRight will not tell us anything why this test fails, if it does:
```suggestion
      counterexample ("Unexpected decoder failure: " ++ show err) $
      case roundTrip (natVersion @9) (cborTrip @(BabbageTxOut era)) nonPtrAddrTxOut of
        Left err ->  False
        Right  -> True
```

libs/cardano-ledger-core/src/Cardano/Ledger/Address.hs
```diff
@@ -401,6 +379,10 @@ isBootstrapRedeemer :: BootstrapAddress c -> Bool
 isBootstrapRedeemer (BootstrapAddress (Byron.Address _ _ Byron.ATRedeem)) = True
 isBootstrapRedeemer _ = False
 
+isPtr :: Addr era -> Bool
+isPtr (Addr _ _ (StakeRefPtr _)) = True
+isPtr _ = False
```
lehins: Consider a little more descriptive name eg. `isStakePtrAddr`?

lehins (approved):
This is great! Few recommendations, otherwise I see no blockers.

eras/babbage/impl/src/Cardano/Ledger/Babbage/TxOut.hs
teodanciu: Yep, I think the scala-tendencies took over!  Your version is way easier to read, thanks

eras/babbage/impl/src/Cardano/Ledger/Babbage/TxOut.hs
teodanciu: Yep, nice idea!  Will update

libs/cardano-ledger-core/src/Cardano/Ledger/Address.hs
JaredCorduan: elsewhere, these are described as "pointer addresses", so I would probably go with `isPtrAddr`. there isn't a notion of "payment pointer address", for example.

eras/conway/test-suite/src/Test/Cardano/Ledger/Conway/Serialisation/Roundtrip.hs
teodanciu: Hey @lehins , I'm only able to pass the era as a parameter with these signature: 
```
ptrAddrTxOutNotDeserialized ::
  forall e.
  ( EraTxOut e,
    TxOut e ~ BabbageTxOut e,
    Mock (EraCrypto e),
    Arbitrary (Value e),
    Arbitrary (Script e),
    Show (Script e)
  ) =>
  BabbageTxOut e ->
  Property
```

which then I have to repeat in the top-level function .
Is there a better way? 

Also: these functions get invoked ultimately from  conway `Main`, like this: 
```
Roundtrip.allprops @Conway
```

So passing the era as a parameter will come in handy if in the future eras we will explicitly call this function passing the new era as a parameter  - that's what you meant, correct?

lehins: That's right, that's exactly what I meant. Also you can simplify this a bit:
```haskell
ptrAddrTxOutNotDeserialized ::
  forall e.
  ( EraTxOut e,
    Mock (EraCrypto e),
    Arbitrary (Value e),
    Arbitrary (Script e),
    Show (Script e)
  ) =>
  TxOut e ->
  Property
```

JaredCorduan:
@teodanciu we should mention this in the changelog

Issue #3171, opened 2022-11-28
Split cardano-shelley-ma into two Allegra and Mary


lehins:
Current implementation of ShelleyMA is unlike all other eras and is quite awkward and sometimes gets into way of implementing new features.

The plan is to:

* [x] - (done in #3175) Split `cardano-ledger-shelley-ma` into `cardano-ledger-allegra` and `cardano-ledger-mary`
  * Keep `cardano-ledger-shelley-ma` for a while with some deprecations instructions how to transition to new packages. Breakage should be minimal
  * Get rid of `ShelleyMA ma era` type and `MAClass` class as no longer useful
* [ ] - Move `cardano-ledger-shelley-ma-test` functionality into `testlib` sublibraries in the `cardano-ledger-allegra` and `cardano-ledger-mary`

lehins:
This is now a subtask of #3623

PR #3175, opened 2022-11-30
Split shelley-ma into allegra and mary

lehins:
This PR splits the package `cardano-ledger-shelley-ma` into two packages that implement proper eras:

* `cardano-ledger-allegra`
* `cardano-ledger-mary`. 

It is important to point out that in order to keep this PR as small as possible the test package `cardano-ledger-shelley-ma-test` is still a thing. That test package will be restructured into separate libraries in the  corresponding `eras/allegra/` and `eras/mary/` directories in a later PR.  `shelley-ma` era also contains the spec, which I am not quite sure yet where it should eventually live, since it does combine the two eras together as one document.

There is no longer a `ShelleyMAEra ma c`, there are proper types that replace it:
* `AllegraEra c`
*  and `MaryEra c`.

Conceptually the only thing that is different between the two eras is the support of multi assets by the latter era, therefore the only type that had seen a somewhat impactful change was the `TxBody`. 

There is no longer ` MATxBody`, but two:
* `AllegraTxBody`
*  and `MaryTxBody`. 

Care was take to reduce duplication so, both of those types are backed by the same raw type underneath. This also means that corresponding type class was split as well: 

There is no longer `ShelleyMAEraTxBody`, but instead these two:
* `AllegraEraTxBody` 
* and `MaryEraTxBody`.


CHANGELOG.md
```diff
@@ -205,6 +197,7 @@ in the naming of release branches.
 
 ### Removed
 
+- `MAClass` is gone: #3175
```
JaredCorduan, line 200: :wave:

eras/shelley-ma/impl/cardano-ledger-shelley-ma.cabal
```diff
@@ -2,10 +2,9 @@ cabal-version: 3.0
 
 name:                cardano-ledger-shelley-ma
 version:             0.1.0.0
-synopsis:            Shelley ledger with multiasset and time lock support.
+synopsis:            Deprecated package that used to implement Allegra and Mary eras.
```
teodanciu, line 5: Just curious if there is a particular reason to leave the deprecated ones around, or just common sense?
Any idea of when to remove them?

eras/shelley-ma/test-suite/test/Test/Cardano/Ledger/Mary/Examples/Cast.hs
```diff
@@ -22,73 +22,73 @@ module Test.Cardano.Ledger.Mary.Examples.Cast
...
 import Cardano.Ledger.Keys
   ( KeyPair (..),
     KeyRole (..),
   )
-import Test.Cardano.Ledger.EraBuffet (TestCrypto)
 import Test.Cardano.Ledger.Shelley.Utils (RawSeed (..), mkAddr, mkKeyPair)
 
 -- | Alice's payment key pair
-alicePay :: KeyPair 'Payment TestCrypto
+alicePay :: KeyPair 'Payment StandardCrypto
```
teodanciu, line 33: Is this  `StandardCrypto` to `TestCrypto` change related to the split, or just an general improvement?

eras/shelley-ma/impl/cardano-ledger-shelley-ma.cabal
lehins: This is to help downstream packages migrate to new packages. We on=always try to give users some time to migrate and then we can get rid of deprecated stuff

eras/shelley-ma/test-suite/test/Test/Cardano/Ledger/Mary/Examples/Cast.hs
lehins: This is general improvement. Whenever possible we should use `StandardCrypto`. THere are very few places were `TestCrypto` actually makes sense, in particular it is relevant when dealing with actual Blocks

eras/allegra/impl/src/Cardano/Ledger/Allegra/TxBody.hs
```diff
@@ -0,0 +1,356 @@
...
+          emptyAllegraTxBodyRaw
+          bodyFields
+          [(0, "atbrInputs"), (1, "atbrOutputs"), (2, "atbrTxFee")]
+      )
+
+instance AllegraEraTxBody era => FromCBOR (Annotator (AllegraTxBodyRaw () era)) where
+  fromCBOR = pure <$> fromCBOR
+
+-- Sparse encodings of AllegraTxBodyRaw, the key values are fixed by backward compatibility
+-- concerns as we want the Shelley era TxBody to deserialise as a Shelley-ma TxBody.
```
teodanciu: to deserialize as an Allegra TxBody ?

lehins: yep. Fixed

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/Fields.hs
```diff
@@ -237,98 +236,29 @@ initVI = ValidityInterval SNothing SNothing
...
-    initWdrl
-    (Coin 0)
-    initVI
-    SNothing
-    Set.empty
-    initMultiAsset
-    SNothing
-    SNothing
-    SNothing
+initialTxBody (Shelley _) = mkBasicTxBody
```
teodanciu, line 240: Awesome mass deduplications in this file!

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/Consensus.hs
```diff
@@ -45,7 +47,7 @@ import Cardano.Ledger.Conway.Genesis (ConwayGenesis (..))
 import qualified Cardano.Ledger.Conway.PParams as ConwayPP
 import Cardano.Ledger.Conway.Translation ()
 import Cardano.Ledger.Core as Core
-import Cardano.Ledger.Crypto as CC
+import Cardano.Ledger.Crypto
```
teodanciu, line 49: This `CC`  alias for `Cardano.Ledger.Crypto`  seems to be used a lot throughout  the codebase, but perhaps it's necessary in some places as opposed to here

eras/shelley-ma/test-suite/src/Test/Cardano/Ledger/ShelleyMA/TxBody.hs
```diff
@@ -5,75 +5,53 @@
...
-    (Wdrl Map.empty)
-    (Coin 6)
-    (ValidityInterval (SJust (SlotNo 3)) (SJust (SlotNo 42)))
-    SNothing
-    SNothing
-    testmint
+  mkBasicTxBody
+    & feeTxBodyL .~ Coin 6
+    & vldtTxBodyL .~ ValidityInterval (SJust (SlotNo 3)) (SJust (SlotNo 42))
+    & mintTxBodyL .~ testMint
```
teodanciu, line 47: So much easier to understand what's being set

eras/shelley-ma/impl/src/Cardano/Ledger/ShelleyMA/Core.hs
```diff
@@ -1,22 +1,12 @@
...
 
-class ShelleyEraTxBody era => ShelleyMAEraTxBody era where
-  vldtTxBodyL :: Lens' (TxBody era) ValidityInterval
-
-  mintTxBodyL :: Lens' (TxBody era) (MultiAsset (EraCrypto era))
-
-  mintValueTxBodyF :: SimpleGetter (TxBody era) (Value era)
-
-  mintedTxBodyF :: SimpleGetter (TxBody era) (Set (ScriptHash (EraCrypto era)))
+type ShelleyMAEraTxBody era = (AllegraEraTxBody era, MaryEraTxBody era)
```
teodanciu, line 12: Won't this introduce confusion, since this name meant something else before the refactoring?  Will it be useful defined like this?

eras/shelley-ma/impl/src/Cardano/Ledger/ShelleyMA/TxWits.hs
```diff
@@ -1,39 +1,5 @@
...
-
-import Cardano.Ledger.Core (EraTxWits (..))
-import Cardano.Ledger.Crypto (StandardCrypto)
-import Cardano.Ledger.Shelley.TxWits
-  ( ShelleyTxWits,
-    addrShelleyTxWitsL,
-    bootAddrShelleyTxWitsL,
-    scriptShelleyTxWitsL,
+module Cardano.Ledger.ShelleyMA.TxWits
+  {-# DEPRECATED "Should not have been exposed in the first place" #-}
```
teodanciu, line 2: It shouldn't have been exposed?  We're still exposing Mary.TxWits in the new implementation:  https://github.com/input-output-hk/cardano-ledger/pull/3175/files#diff-3c54618278f3efba674af5e7fc84e6fe414da600e08b9ddc3603868ab5603ed5R11

eras/shelley-ma/impl/src/Cardano/Ledger/ShelleyMA/Era.hs
```diff
@@ -1,162 +0,0 @@
...
-    NoThunks (MAValue ma c),
-    KnownNat (MAProtVer ma),
-    MinVersion <= MAProtVer ma,
-    MAProtVer ma <= MaxVersion
-  ) =>
-  MAClass (ma :: MaryOrAllegra) c
-  where
-  type MAValue (ma :: MaryOrAllegra) c :: Type
-  getScriptHash :: proxy ma -> MultiAsset c -> Set.Set (ScriptHash c)
-  promoteMultiAsset :: proxy ma -> MultiAsset c -> Value (ShelleyMAEra ma c)
```
teodanciu, line 84: Great that this hack is gone now :tada:

teodanciu (approved):
This is awesome!  Looks good to me, just some minor questions and remarks.

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/Consensus.hs
lehins: We used to have a clash, these two had the same name `Cardano.Ledger.Crypto.Crypto` and `Cardano.Lerdge.Core.EraCrypto`. This resulted in all those qualified as CC. I suggested a rename ` Crypto` -> `EraCrypto` #2933, so this is no longer necessary.

eras/shelley-ma/impl/src/Cardano/Ledger/ShelleyMA/Core.hs
lehins: This is only defined so people get a compilation warning rather than compilation error. Besides that, this constraint synonym actually means the same thing as before. Only thing is you can't create instances for such constraint

eras/shelley-ma/impl/src/Cardano/Ledger/ShelleyMA/TxWits.hs
lehins: Actually you are almost right, we are exposing it, but in Allegra, not in Mary. I'll fix that. See:

https://github.com/input-output-hk/cardano-ledger/blob/01f523a72c132259466f32c69a4a88d4b555f96d/eras/mary/impl/cardano-ledger-mary.cabal#L47-L53

lehins: Fixed

lehins: FYI. It shouldn't have been exposed, because it does not export anything. The only thing that it exports is type class instance for EraTxWits, which is unusable unless you have the concrete era type (eg. `MaryEra`) in scope, which itself is exported together with the instance. So `TxWits` modules in Allegra and Mary are suppose to be hidden

teodanciu: Makes sense! Thanks for the explanation

eras/allegra/impl/src/Cardano/Ledger/Allegra/Era.hs
```diff
@@ -0,0 +1,97 @@
...
+
+type instance EraRule "SNAP" (AllegraEra c) = ShelleySNAP (AllegraEra c)
+
+type instance EraRule "TICK" (AllegraEra c) = API.ShelleyTICK (AllegraEra c)
+
+type instance EraRule "TICKF" (AllegraEra c) = ShelleyTICKF (AllegraEra c)
+
+type instance EraRule "UPEC" (AllegraEra c) = ShelleyUPEC (AllegraEra c)
+
+-- These rules are defined anew in the Allegra era(s)
```
JaredCorduan:
```suggestion
-- These rules are defined anew in the Allegra era
```

eras/allegra/impl/src/Cardano/Ledger/Allegra/Rules/Utxo.hs
```diff
@@ -36,11 +38,9 @@ import Cardano.Ledger.Binary
     invalidKey,
     serialize,
   )
-import Cardano.Ledger.Coin
+import Cardano.Ledger.Coin (Coin)
```
JaredCorduan, line 41: :pray:

eras/allegra/impl/src/Cardano/Ledger/Allegra/Scripts.hs
```diff
@@ -0,0 +1,312 @@
...
+import Data.ByteString.Lazy (fromStrict)
+import Data.ByteString.Short (fromShort)
+import Data.Sequence.Strict (StrictSeq (Empty, (:<|)))
+import Data.Set (Set, member)
+import GHC.Generics (Generic)
+import NoThunks.Class (NoThunks (..))
+
+-- ================================================================
+-- A pair of optional SlotNo.
+
```
JaredCorduan:
```suggestion
```

eras/allegra/impl/src/Cardano/Ledger/Allegra/Scripts.hs
```diff
@@ -0,0 +1,312 @@
...
+   in case tl of
+        Signature s -> rewrap $ Signature s
+        AllOf l -> rewrap . AllOf $ translateTimelock <$> l
+        AnyOf l -> rewrap . AnyOf $ translateTimelock <$> l
+        MOfN n l -> rewrap $ MOfN n (translateTimelock <$> l)
+        TimeStart x -> rewrap $ TimeStart x
+        TimeExpire x -> rewrap $ TimeExpire x
+
+-- These coding choices are chosen so that a MultiSig script
+-- can be deserialised as a Timelock script
```
JaredCorduan, line 141: is this supposed to be attached to the `ToCBOR` instance? I know you didn't write these... but there seems to be a lot of stray comments that are not as lovely at attached haddocks...

eras/allegra/impl/src/Cardano/Ledger/Allegra/Translation.hs
```diff
@@ -38,14 +38,6 @@ import qualified Data.Map.Strict as Map
 -- being total. Do not change it!
 --------------------------------------------------------------------------------
 
--- | Return the subset of UTxO corresponding to Byron-era AVVM addresses, which
--- are to be removed on the Shelley/Allegra boundary. This set will be passed
--- _back_ to the translation functions as the UTxO, allowing these addresses to
--- be removed. This is needed because we cannot do a full scan on the UTxO at
--- this point, since it has been persisted to disk.
-shelleyToAllegraAVVMsToDelete :: NewEpochState (ShelleyEra c) -> UTxO (ShelleyEra c)
-shelleyToAllegraAVVMsToDelete = stashedAVVMAddresses
```
JaredCorduan: huh, I thought this was still needed. :thinking:

eras/allegra/impl/src/Cardano/Ledger/Allegra/Scripts.hs
lehins: Honestly all these comments are totally useless, since they apply to all of our types. Adding them as haddock to each individual instance is pointless too, it would be better to have a proper documentation on serialization of core types and why and how we handle backwards compatibility. 

We should also document how annotated/memoized serialization is being used with those types, but that is an implementation detail that people working on ledger should be aware of.

eras/allegra/impl/src/Cardano/Ledger/Allegra/Translation.hs
lehins: You are totally right, I thought no-one uses it because I couldn't find any place where it is being used, but it turns out it still didn't land in consensus master and is part of this UTxO-HD long running PR: input-output-hk/ouroboros-network#3863

I'll bring it back.

eras/shelley-ma/test-suite/test/Test/Cardano/Ledger/Mary/Examples/MultiAssets.hs
```diff
@@ -45,52 +46,40 @@ import qualified Data.Map.Strict as Map
...
 
-type MaryTest = MaryEra TestCrypto
-
 ------------------------------
 -- Set Up the Initial State --
 ------------------------------
 
 aliceInitCoin :: Coin
-aliceInitCoin = Coin $ 10 * 1000 * 1000 * 1000 * 1000 * 1000
+aliceInitCoin = Coin $ 10_000_000_000_000_000
```
JaredCorduan, line 60: :raised_hands:

JaredCorduan (approved):
Looks good, thank you @lehins !

I can't recall why `shelleyToAllegraAVVMsToDelete` is still around if it is not used...

Is the plan to keep `eras/shelley-ma/test-suite` for testing? or punt for now and move it later? (I'd love to get my tab-complete on "shel" back :laughing: )

lehins:
> Is the plan to keep eras/shelley-ma/test-suite for testing? 

Yes, I will split `cardano-ledger-shelley-ma-test` in a subsequent PR. I've updated the #3171 ticket to reflect it

> (I'd love to get my tab-complete on "shel" back laughing )

Unfortunately we'll have to keep `cardano-ledger-shelley-ma` for a little bit until downstream reflect on these changes. But afterwards we'll definitely be able to get rid of `era/shelley-ma` folder. One thing I am not certain about is where to put the `era/shelley-ma/formal-spec` folder, since it crams both Allegra and Mary in there

JaredCorduan:
> Unfortunately we'll have to keep `cardano-ledger-shelley-ma` for a little bit until downstream reflect on these changes. But afterwards we'll definitely be able to get rid of `era/shelley-ma` folder. One thing I am not certain about is where to put the `era/shelley-ma/formal-spec` folder, since it crams both Allegra and Mary in there

IMHO, it is the Mary spec. There is no Allegra spec, you just have to know what parts of the Mary spec to leave out... Maybe we can rename it to the Mary spec and have an appendix about the "Allegra spec".

cc @polinavino  @WhatisRT

eras/allegra/impl/src/Cardano/Ledger/Allegra/Translation.hs
lehins: Done

PR #3182, opened 2022-12-02
Bootstrapped the public facing API of cardano-ledger-api

koslambrou:
**Summary**

* Created the `Cardano.Ledger.Api` module which exports anything the end-user will need in order to build and inspect transactions.

* Completing some missing exports in lower modules

* Bootstrapped very simple documentation on how to build and inspect transactions

**Notes**

After trying out `cardano-ledger-api`, this is my simple attempt to bootstrap some documentation on how to use the API from an external user's perspective.

I made some opinionated changes, so that I can get something started.

### DIFF
--- a/libs/cardano-ledger-api/src/Cardano/Ledger/Api.hs
+++ b/libs/cardano-ledger-api/src/Cardano/Ledger/Api.hs
@@ -1,0 +1,26 @@
+-- | This module provides a library interface for working with types that will allow a user to
+-- interact with Cardano ledger.
+--
+-- It is intended to be the complete API covering everything but without exposing constructors that
+-- reveal any lower level types.
+--
+-- In the interest of simplicity it glosses over some details of the system.
+-- Most tools should be able to work just using this interface, however you can go deeper and
+-- experiment with internal modules if necessary, such as "Cardano.Ledger.Core",
+-- "Cardano.Ledger.Shelley", "Cardano.Ledger.Babbage", etc.
+module Cardano.Ledger.Api
+  ( -- | Definition of Cardano Ledger era types.
+    module Cardano.Ledger.Api.Era,
+    -- | Building and inspecting transactions.
+    module Cardano.Ledger.Api.Tx,
+    -- | Protocol parameters.
+    module Cardano.Ledger.Api.PParams,
+    -- | Scripts
+    module Cardano.Ledger.Api.Scripts,
+  )
+where
+
+import Cardano.Ledger.Api.Era
+import Cardano.Ledger.Api.PParams
+import Cardano.Ledger.Api.Scripts
+import Cardano.Ledger.Api.Tx
--- a/libs/cardano-ledger-api/src/Cardano/Ledger/Api/Tx/Out.hs
+++ b/libs/cardano-ledger-api/src/Cardano/Ledger/Api/Tx/Out.hs
@@ -4,17 +4,34 @@
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE TypeApplications #-}
 
+-- \* Building and inspecting transaction outputs
+--
+-- Here's an example on how to build a very basic Babbage era transaction output with a random
+-- address and value, and without any datum or reference script.
+--
+-- >>> :set -XTypeApplications
+-- >>> import Test.QuickCheck
+-- >>> import Test.Cardano.Ledger.Babbage.Serialisation.Generators ()
+-- >>> import Lens.Micro
+-- prop> \addr val ->
+--     let
+--         -- Defining a Babbage era transaction output with some random address and value.
+--         txOut = mkBasicTxOut @Babbage addr val
+--      in
+--         -- We verify that the transaction output contains our random address and value.
+--         txOut ^. addrTxOutL == addr && txOut ^. valueTxOutL == val
 module Cardano.Ledger.Api.Tx.Out
-  ( EraTxOut (..),
+  ( module Cardano.Ledger.Api.Scripts.Data,
+    EraTxOut (..),
 
-    -- * Shelley Era
+    -- * Shelley, Allegra and Mary Era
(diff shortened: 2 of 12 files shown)


libs/cardano-ledger-api/src/Cardano/Ledger/Api.hs
```diff
@@ -0,0 +1,100 @@
...
+-- interact with Cardano.
+--
+-- It is intended to be the complete API covering everything but without exposing constructors that
+-- reveal any lower level types.
+--
+-- In the interest of simplicity it glosses over some details of the system.
+-- Most simple tools should be able to work just using this interface, however you can go deeper and
+-- expose the types from the underlying libraries using "Cardano.Ledger.Core",
+-- "Cardano.Ledger.Shelley", "Cardano.Ledger.Babbage", etc.
+module Cardano.Ledger.Api
```
lehins, line 11: It is a good idea to have the top level `Cardano.Ledger.Api` module that exports everything that is needed for working with ledger. However, it should re-export other module form the `cardano-ledger-api`, rather than export individual types with haddock. This is approximately how it suppose to look like:

```haskell
module Cardano.Ledger.Api (
  module Cardano.Ledger.Api.Era,
  module Cardano.Ledger.Api.Tx,
) where

import Cardano.Ledger.Api.Era
import Cardano.Ledger.Api.Tx
```

```haskell
module Cardano.Ledger.Api.Tx (
  module Cardano.Ledger.Api.Tx.Body,
  module Cardano.Ledger.Api.Tx.AuxData,
  module Cardano.Ledger.Api.Tx.Wits,
  ... all of Tx related stuff
) where

import Cardano.Ledger.Api.Tx.Body
-- these below aren't implemented yet, but should be here
import Cardano.Ledger.Api.Tx.AuxData
import Cardano.Ledger.Api.Tx.Wits
```

```haskell
module Cardano.Ledger.Api.Tx.Body (
  module Cardano.Ledger.Api.Tx.Out,
  module Cardano.Ledger.Api.Scripts,
  ... all of TxBody related stuff
) where

import Cardano.Ledger.Api.Tx.Out
import Cardano.Ledger.Api.Scripts
```

```haskell
module Cardano.Ledger.Api.Tx.Out (
  module Cardano.Ledger.Api.Tx.Out,
  module Cardano.Ledger.Api.Scripts.Data,
  ... all of TxOut related stuff
) where

import Cardano.Ledger.Api.Tx.Out
import Cardano.Ledger.Api.Scripts.Data
```

If it is not done this way, then we will either end up duplicating documentation in all places or one of modules will miss the relevant documentation. But if we do module re-exporting then all of haddocks and examples stay inside their respectful modules. Another problem that it solves: it is way too easy to export a new function or a type from one module, but not the top level. So, this hierarchy is pretty important.


I can finish up the full module hierarchy, but since you added the `Cardano.Ledger.Api`, could you re-export existing modules in the suggested way and move the example into the `Cardano.Ledger.Api.Tx` module.

lehins (changes requested):
Thank you for adding this. One important comment, otherwise looks good.

libs/cardano-ledger-api/src/Cardano/Ledger/Api.hs
koslambrou: Totally makes sense. Thanks for reviewing this.

Will make the requested changes

JaredCorduan (approved):
Thank you @koslambrou , this looks wonderful!

libs/cardano-ledger-api/src/Cardano/Ledger/Api.hs
```diff
@@ -0,0 +1,26 @@
+-- | This module provides a library interface for working with types that will allow a user to
+-- interact with Cardano.
```
lehins: Perhaps?
```suggestion
-- interact with Cardano ledger.
```

libs/cardano-ledger-api/src/Cardano/Ledger/Api.hs
```diff
@@ -0,0 +1,26 @@
...
+--
+-- It is intended to be the complete API covering everything but without exposing constructors that
+-- reveal any lower level types.
+--
+-- In the interest of simplicity it glosses over some details of the system.
+-- Most simple tools should be able to work just using this interface, however you can go deeper and
```
lehins: This interface should be as powerful as using the underlying internals, but in a simpler way.
```suggestion
-- Most tools should be able to work just using this interface, however you can go deeper and
```

libs/cardano-ledger-api/src/Cardano/Ledger/Api.hs
```diff
@@ -0,0 +1,26 @@
...
+-- expose the types from the underlying libraries using "Cardano.Ledger.Core",
```
lehins:
```suggestion
-- and experiment with internal modules if necessary, such as  "Cardano.Ledger.Core",
```

libs/cardano-ledger-api/src/Cardano/Ledger/Api/Tx/Out.hs
```diff
@@ -4,17 +4,35 @@
...
 {-# LANGUAGE TypeApplications #-}
 
+-- \* Building and inspecting transaction outputs
+--
+-- Here's an example on how to build a very basic Babbage era transaction output with a random
+-- address and value, and without any datum or reference script.
+--
+-- >>> :set -XTypeApplications
+-- >>> import Test.QuickCheck
+-- >>> import qualified Data.Sequence.Strict as StrictSeq
```
lehins: Copy paste error :wink: 
```suggestion
```

lehins (approved):
This looks great. Just a few minor comments. 

Thank you for your effort!

koslambrou:
Thanks for reviewing @lehins :) 

More to come.

PR #3184, opened 2022-12-03
Create core test sublib

lehins:
This PR cleans up, organizes Arbitrary instances for all of the `cardano-ledger-core` types in the newly created `cardano-ledger-core:testlib` . Also it removes some of the duplicate hedgehog generators, that are now fully replaced by `QuickCheck` instances.

It is important to note that this is just the beginning of the test suite cleanup, therefore more PRs like this one are expected to be coming in the future.


libs/cardano-ledger-binary/src/Cardano/Ledger/Binary/Decoding/FromCBOR.hs
```diff
@@ -54,7 +54,7 @@ import Codec.Serialise as Serialise (Serialise (decode))
 import qualified Data.ByteString as BS
 import qualified Data.ByteString.Lazy as BSL
 import qualified Data.ByteString.Short as SBS
-#if __GLASGOW_HASKELL__ < 900
+#if !MIN_VERSION_bytestring(0,11,1)
 import qualified Data.ByteString.Short.Internal as SBS
 #endif
```
TimSheard: How about a comment here, that tells why we have this  conditional import

eras/shelley/test-suite/src/Test/Cardano/Ledger/Shelley/Serialisation/EraIndepGenerators.hs
```diff
@@ -194,29 +140,10 @@ instance Mock c => Arbitrary (BHeader c) where
...
-    WitVKey
-      <$> arbitrary
-      <*> arbitrary
+instance Crypto c => Arbitrary (TP.HashHeader c) where
+  arbitrary = TP.HashHeader <$> arbitrary
 
-instance CC.Crypto c => Arbitrary (Wdrl c) where
+instance Crypto c => Arbitrary (Wdrl c) where
   arbitrary = genericArbitraryU
   shrink = genericShrink
```
TimSheard, line 149: I always wonder why Wdrl defined in  Cardano.Ledger.Shelley.Core
newtype Wdrl c = Wdrl {unWdrl :: Map (RewardAcnt c) Coin}
  deriving (Show, Eq, Generic)
  deriving newtype (NoThunks, NFData)

Isn't in the Core?  It probably should be introduced in  Cardano.Ledger.Address where RewardAcnt  is defined.

eras/shelley/test-suite/src/Test/Cardano/Ledger/Shelley/Serialisation/EraIndepGenerators.hs
```diff
@@ -15,59 +16,25 @@
 
 module Test.Cardano.Ledger.Shelley.Serialisation.EraIndepGenerators
   ( genCoherentBlock,
```
TimSheard, line 18: Maybe we should call this  Test.Cardano.Ledger.Shelley.Serialisation.Generators
since none of the type here are Era independent, now.  That is how the other Era  generators are consistently named:

module Test.Cardano.Ledger.ShelleyMA.Serialisation.Generators
module Test.Cardano.Ledger.Alonzo.Serialisation.Generators
module Test.Cardano.Ledger.Babbage.Serialisation.Generators

TimSheard (approved):
Some small comments, I think the Wdrl and renaming EraIndependentGenerators are worth doing

libs/cardano-ledger-binary/src/Cardano/Ledger/Binary/Decoding/FromCBOR.hs
lehins: I'll do one better. I'll be more explicit about the import, which will depict the exact reason why this CPP is needed:
```haskell
#if MIN_VERSION_bytestring(0,11,1)
import Data.ByteString.Short (ShortByteString(SBS))
#else
import Data.ByteString.Short.Internal (ShortByteString(SBS))
#endif
```

eras/shelley/test-suite/src/Test/Cardano/Ledger/Shelley/Serialisation/EraIndepGenerators.hs
lehins: Yeah, this is a really good point. By moving `Wdrl` to `Cardano.Ledger.Address` we would also be able to move `wdrlsTxBodyL` from `ShelleyEraTxBody` to `EraTxBody`.

I suggest this is done in a separate PR though. Cretaetd a ticket https://github.com/input-output-hk/cardano-ledger/issues/3194

eras/shelley/test-suite/src/Test/Cardano/Ledger/Shelley/Serialisation/EraIndepGenerators.hs
```diff
@@ -15,59 +16,25 @@
...
-    genHash,
-    genShelleyAddress,
-    genByronAddress,
```
teodanciu, line 20: What is the best  way then to generate a different, more specific Gen , as I did in my [PR](https://github.com/input-output-hk/cardano-ledger/pull/3174/commits/e06d141c708df0222270972e3c8998a9406bd614#diff-d2955d72f315240794b22235ed710e10b4896b47c5d39af211f37a81a270e2f4R348)?  
Perhaps I can keep this specialized function alongside the arbitrary instance, or what is more idiomatic to do?

JaredCorduan (approved):
LGTM!

eras/shelley/test-suite/src/Test/Cardano/Ledger/Shelley/Serialisation/EraIndepGenerators.hs
lehins: First of all this module should be called `Arbitrary`, because `Generators` implies that you export functions that do generation, rather than instances.

Second of all there should be no `Serialization` in there either, since these instances are used many places, not only for testing serilization.

That will be done as part of my follow up test reorganization and cleanup

eras/shelley/test-suite/src/Test/Cardano/Ledger/Shelley/Serialisation/EraIndepGenerators.hs
lehins: Yes, in that case you generate a value with a specialized function and use that instead. Hopefully with short documentation describing why this is necessary.

Ideally Arbitrary instance suppose to explore all possible states a type can be in.

eras/shelley/test-suite/src/Test/Cardano/Ledger/Shelley/Serialisation/EraIndepGenerators.hs
```diff
@@ -194,29 +140,10 @@ instance Mock c => Arbitrary (BHeader c) where
...
-
-instance CC.Crypto c => Arbitrary (TP.HashHeader c) where
-  arbitrary = TP.HashHeader <$> genHash
-
-instance (Typeable kr, CC.Crypto c) => Arbitrary (WitVKey kr c) where
-  arbitrary =
-    WitVKey
-      <$> arbitrary
-      <*> arbitrary
+instance Crypto c => Arbitrary (TP.HashHeader c) where
```
teodanciu, line 143: How come some of these instances stayed here?  What's the criteria?   they are to  do only with Shelley maybe?

.github/workflows/bench.yml
```diff
@@ -20,7 +20,7 @@ jobs:
           nix_path: nixpkgs=channel:nixos-unstable
           extra_nix_config: |
             trusted-public-keys = hydra.iohk.io:f/Ea+s+dFdN+3Y/G+FDgSq+a5NEWhJGzdjvKNGv0/EQ= cache.nixos.org-1:6NCHdD59X431o0gWypbMrAURkbJ16ZPMQFGspcDShjY=
-            substituters = https://hydra.iohk.io https://cache.nixos.org/
+            substituters = https://cache.iog.io https://cache.nixos.org/
```
teodanciu, line 23: This change could be in a different commit, but it doesn't matter too much, maybe not worth the hassle

hie.yaml
```diff
@@ -117,6 +117,9 @@ cradle:
     - path: "libs/cardano-ledger-core/src"
       component: "lib:cardano-ledger-core"
 
+    - path: "libs/cardano-ledger-core/test-lib"
```
teodanciu: I think the folder was renamed to `testlib` as well, no?

libs/cardano-ledger-binary/test/Test/Cardano/Ledger/Binary/Vintage/Serialization.hs
```diff
@@ -87,7 +88,7 @@ genTestStruct =
     <*> (V.fromList <$> Gen.list (Range.constant 0 10) Gen.bool)
     <*> (BS.Lazy.fromStrict <$> Gen.bytes (Range.linear 0 20))
     <*> (BS.Short.toShort <$> Gen.bytes (Range.linear 0 20))
-    <*> genUTCTime
+    <*> arbitrary
```
teodanciu, line 91: Awesome noise reduction!

.github/workflows/bench.yml
lehins: Yeah, I don't wanna deal with it :smiling_face_with_tear:

eras/shelley/test-suite/src/Test/Cardano/Ledger/Shelley/Serialisation/EraIndepGenerators.hs
lehins: Criteria is that instances for types that are defined in `cardano-ledger-core` should go into `cardano-ledger-core:testlib`. So the `HashHeader`, would need to be migrated into `cardano-protocol-tpraos`, cause that is where it is defined. That should be different PR though. If you feel like it, you could submit such PR, basing on this PR as an exmaple.

hie.yaml
lehins: Good catch!!! Thank you!

lehins: Fixed

.github/workflows/bench.yml
lehins: Since I had to change the commit anyways, I moved this change into a separate commit. :+1:

teodanciu (approved):
Looks good to me

Issue #3015, opened 2022-08-31
Add doctests to our CI


JaredCorduan:
https://hackage.haskell.org/package/doctest

Note that we disabled doc tests in https://github.com/input-output-hk/cardano-ledger/pull/2793, they were depending on oleg.fi being available.

lehins:
Here is an example of the correct way to run doctests: https://github.com/haskell/random/blob/877522162c959d757b638d57bfaff2324063fd3a/scripts/doctest.sh#L5

It was wrong to try and add them as a separate section in a cabal file, since that is not a supported way of running them.

JaredCorduan:
❤

PR #3188, opened 2022-12-03
Added support for doctest in cardano-ledger-api

koslambrou:
* Added a test-suite for running all doctests in Haddock comments

* Fixed existing "doctest" in cardano-ledger-api's doc which wasn't conforming to the doctest syntax.

Linked to https://github.com/input-output-hk/cardano-ledger/issues/3015, but this PR only enables doctests in `cardano-ledger-api`

### DIFF
--- a/libs/cardano-ledger-api/src/Cardano/Ledger/Api/Tx.hs
+++ b/libs/cardano-ledger-api/src/Cardano/Ledger/Api/Tx.hs
@@ -1,28 +1,32 @@
--- \* Building and inspecting transactions
---
-
 -- | Transaction building and inspecting relies heavily on lenses (`microlens`). Therefore, some
 -- familiarity with those is necessary. However, you can probably go a long way by simply
 -- looking at the examples and try to go from there.
 --
--- Here's an example on how to build a very simple Babbage era unbalanced transaction using the
--- provided interface.
+-- Let's start by defining the GHC extensions and imports.
 --
--- >>> :set -XTypeApplications
+-- >>> :set -XScopedTypeVariables
 -- >>> import Test.QuickCheck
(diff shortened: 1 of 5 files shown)


lehins:
@koslambrou If you look at the #3015 ticket the goal is to setup doctest with `doctest` package, not `cabal-doctest`, which has been deprecated on hackage by the maintainer. Moreover `cabal-doctest` requires a custom setup, which is another thing we are trying to avoid

koslambrou:
No problem, I'll close this PR.

lehins:
I have tried the `doctest` package instead of `cabal-doctest` and it seems to be pretty broken for ghc-9

I think this approach is the next best option, at least until the time when the `doctest` gets fixed.

lehins (approved):
Looks great. I am not fond of custom cabal setup that comes with cabal-doctest, but this is the best that we can do, as far as I am aware.

libs/cardano-ledger-api/src/Cardano/Ledger/Api/Tx.hs
```diff
@@ -1,28 +1,31 @@
...
--- >>> import Test.Cardano.Ledger.Babbage.Serialisation.Generators ()
+-- >>> import Cardano.Ledger.Api.Era (Babbage)
 -- >>> import Lens.Micro
--- prop> \txOut ->
+-- >>> import Test.Cardano.Ledger.Babbage.Serialisation.Generators ()
+--
+-- Here's an example on how to build a Babbage era unbalanced transaction containing a single
+-- transaction output using the provided interface.
+--
+-- >>> :{
```
lehins, line 17: Example below needs this:
```suggestion
-- >>> :set -XScopedTypeVariables
-- >>> :{
```

PR #3191, opened 2022-12-05
ADR - untranslatable data in script context

JaredCorduan:


teodanciu (approved):
I'm not familiar with this issue, but the description and explanation look clear to me

PR #3192, opened 2022-12-05
ADR - remove ptr addresses

JaredCorduan:


docs/adr/2022-12-05_005-remove-ptr-addresses.md
```diff
@@ -0,0 +1,37 @@
...
+[Ouroboros Leios](https://iohk.io/en/research/library/papers/ouroboros-leios-design-goals-and-concepts/).
+
+## Decision
+
+We will remove support for pointer addresses.
+
+## Consequences
+
+Pointer addresses will no longer be available.
+The existing ~11 pointer addresses will be converted to enterprise addresses.
```
teodanciu, line 35: I forgot if this document is  meant to be a definitive decision?  And also I forgot if we were totally decided on this part, I thought maybe there were some problems with it? 
But anyway, i seem to don't remember well, so please ignore me if that's the case!

JaredCorduan: yes, this is suppose to reflect a decision that we all agreed to.

I think we are all in agreement that this is the plan, but I should make a note of the fact that we are doing it in two stages!
1. prevent the creation of pointer addresses
2. later (after a hard fork) remove pointer addresses everywhere

JaredCorduan: I updated the prose, let me know if it makes sense now, thanks!

teodanciu (approved):
Looks good to me

PR #3197, opened 2022-12-06
Add cabal-doctest setup for all packages

aniketd:
We exclude, for now, due to incompatibilty with internal libraries:
1. `cardano-ledger-binary`
1. `cardano-ledger-core`
1. `plutus-preprocessor`


lehins:
Wee have more internal libraries coming, so this approach is not gonna work. We'll need to figure out something else later

PR #3201, opened 2022-12-08
Add setMinFeeTx function to ledger-api

### LINKED ISSUE #3181: Result of evaluateMinLovelaceOutput and  evaluateMinFee is not valid after transaction update.
### Issue
Related :  https://github.com/input-output-hk/cardano-node/issues/2631

Since Babbage, era the encoded size of `Coin` data depends on the value it has. 

so, if we use `evaluateMinFee` function with `0 Lovelace` fee and then update the transaction with the calculated fee, the size of transaction will increase, and thus the fee is no longer sufficient.

### Solution 1
 Straight forward solution is to  change the `minfee`  function defined in  cardano-ledger-alonzo :
  [cardano-ledger/Tx.hs](https://github.com/input-output-hk/cardano-ledger/blob/c7c63dabdb215ebdaed8b63274965966f2bf408f/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Tx.hs#L333)

- Calculate fee for the transaction
- Calculate new size of transaction as `newSize = oldSize - sizeOf oldFee  + sizeOf minFee `
- return Fee based on new size.

### Solution 2
Since we do not need to do the double computation when we are just validating a transaction, 

We should have function pairs.  
- `minfee`   for validation. (same as current min fee)
- `minfeeForUpdate`, calculates fee as suggested in solution 1.

mesudip:
Should fix https://github.com/input-output-hk/cardano-ledger/issues/3181.

### DIFF
--- a/libs/cardano-ledger-api/src/Cardano/Ledger/Api/Tx.hs
+++ b/libs/cardano-ledger-api/src/Cardano/Ledger/Api/Tx.hs
@@ -54,7 +54,6 @@ import Lens.Micro ((&), (.~), (^.))
 
 -- | Calculate and update the fee in the transaction until it has the smallest possible value according to
 -- the settings in the protocol parameters.
-
 setMinFeeTx :: EraTx era => PParams era -> Tx era -> Tx era
 setMinFeeTx pp tx =
   let curMinFee = getMinFeeTx pp tx
--- a/libs/cardano-ledger-api/src/Cardano/Ledger/Api/Tx.hs
+++ b/libs/cardano-ledger-api/src/Cardano/Ledger/Api/Tx.hs
@@ -52,7 +52,8 @@ import Cardano.Ledger.Core (EraPParams (PParams), EraTx (..))
 import Cardano.Ledger.Shelley.Tx (ShelleyTx)
 import Lens.Micro ((&), (.~), (^.))
 
--- Calcluate minfee and create new transaction until the minfee is same.
+-- | Calculate and update the fee in the transaction until it has the smallest possible value according to
+-- the settings in the protocol parameters.
 
 setMinFeeTx :: EraTx era => PParams era -> Tx era -> Tx era
 setMinFeeTx pp tx =
--- a/libs/cardano-ledger-api/src/Cardano/Ledger/Api/Tx.hs
+++ b/libs/cardano-ledger-api/src/Cardano/Ledger/Api/Tx.hs
@@ -40,6 +40,7 @@ module Cardano.Ledger.Api.Tx
     -- * Alonzo and Babbage Era
     AlonzoTx,
     AlonzoEraTx (..),
+    setMinFeeTx,
   )
 where
 
@@ -47,5 +48,17 @@ import Cardano.Ledger.Alonzo.Tx (AlonzoEraTx (..), AlonzoTx)
 import Cardano.Ledger.Api.Tx.AuxData
 import Cardano.Ledger.Api.Tx.Body
 import Cardano.Ledger.Api.Tx.Wits
-import Cardano.Ledger.Core (EraTx (..))
+import Cardano.Ledger.Core (EraPParams (PParams), EraTx (..))
 import Cardano.Ledger.Shelley.Tx (ShelleyTx)
+import Lens.Micro ((&), (.~), (^.))
+
+-- Calcluate minfee and create new transaction until the minfee is same.
+
+setMinFeeTx :: EraTx era => PParams era -> Tx era -> Tx era
+setMinFeeTx pp tx =
+  let curMinFee = getMinFeeTx pp tx
+      curFee = tx ^. bodyTxL ^. feeTxBodyL
+      modifiedTx = tx & bodyTxL . feeTxBodyL .~ curMinFee
+   in if curFee == curMinFee
+        then tx
+        else setMinFeeTx pp modifiedTx
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ in the naming of release branches.
 
 ### Added
 
+- Add `setMinFeeTx` to  `Cardano.Ledger.Api.Tx`: #3201
 - Add ToExpr (from tree-diff) instances for all types inside NewEpochState
   - Add module Cardano.Ledger.TreeDiff in cardano-binary. Includes all needed operations, classes, and orphan instances
 - Add deposits (key deposits) to DState{dsDeposits} and (pool deposits) PState{psDeposits}


lehins (changes requested):
I am sorry, but that is not the way it should be implemented. 

First of all it should not be `getMinFeeTxFixed`, but instead `setMinFeeTxOut`, since setting the fee changes the size.

Secondly, none of the functions or classes require any changes in the era packages. It should be just one function in the `cardano-ledger-api` package. 

You can draw inspiration from a very similar in concept function that computes the minimum coin on the TxOut:

https://github.com/input-output-hk/cardano-ledger/blob/b613b13139c9be9ba19c57cd2ab73a9b64446dd4/libs/cardano-ledger-api/src/Cardano/Ledger/Api/Tx/Out.hs#L61-L69

If you will not have any luck figuring it out, let me know and I can take of this issue myself.

FYI, we use ormolu for formattin. You can run `scripts/ormolise.sh` script to ensure formatting is in expected form

mesudip:
@lehins 
> First of all it should not be `getMinFeeTxFixed`, but instead setMinFeeTxOut, since setting the fee changes the size.

well by naming `getMinFeeTxFixed` I was trying to infer that the  returned value  of `minFee` is calculated in such a way that after updating the transaction with it, the `getMinFeeTx pp newTx` will return the same fee value.

I agree that the most straightforward solution is to add `setMinFeeTx` to `Cardano.Ledger.Api.Tx`, with something like this.
```haskell
setMinFeeTx :: EraTx era => PParams era -> Tx era -> Tx era
setMinFeeTx pp  = go
  where
    go tx =
      let curMinFee = getMinFeeTx pp tx
          curFee =  tx ^. bodyTxL ^. feeTxBodyL
          modifiedTx = tx &  bodyTxL . feeTxBodyL .~ curMinFee
       in if curFee == curMinFee
            then tx
            else go  modifiedTx
```

The reason  I didn't  go with this approach is because `getMinFeeTx`  is called multiple times when implementing it like this. `getMinFeeTx` at least in alonzo era requires `sizeTxF` and `sizeTxF` is basically `LBS.length $ serializeEncoding $ toCBOR $ tx`.

This means we are re-serializing the entire tx multiple times (at least 3 times if we start with 0 txFee). 

in, Current PR, I solve this issue of re-serializing transaction multiple times by serializing `tx` once and instead serializing `fee` multiple times. 

**Note** that the change is transaction size is only due to change of serialized size of `fee` value. For this `getMinFeeTxFixed` had to access the internal details of feeCalculation and it ended up becoming era dependent. 


If You think that re-serializing the transaction is not an issue, I can update the PR and implement simpler `setMinFeeTx`.

mesudip:
### For Reference : CBOR serialization size of coin, and value of txMinFee

Fee  and   minFeeTx  for a dummy transaction with mainnet PParams.
```
initial Fee,  CBOR length of `Coin` (Bytes), minFeeTx
0 Lovelace      -> CoinSize: 1,  minFeeTx: 165061
2 Lovelace      -> CoinSize: 1,  minFeeTx: 165061
4 Lovelace      -> CoinSize: 1,  minFeeTx: 165061
8 Lovelace      -> CoinSize: 1,  minFeeTx: 165061
16 Lovelace     -> CoinSize: 1,  minFeeTx: 165061
32 Lovelace     -> CoinSize: 2,  minFeeTx: 165105
64 Lovelace     -> CoinSize: 2,  minFeeTx: 165105
128 Lovelace    -> CoinSize: 2,  minFeeTx: 165105
256 Lovelace    -> CoinSize: 3,  minFeeTx: 165149
512 Lovelace    -> CoinSize: 3,  minFeeTx: 165149
1024 Lovelace   -> CoinSize: 3,  minFeeTx: 165149
2048 Lovelace   -> CoinSize: 3,  minFeeTx: 165149
4096 Lovelace   -> CoinSize: 3,  minFeeTx: 165149
8192 Lovelace   -> CoinSize: 3,  minFeeTx: 165149
16384 Lovelace  -> CoinSize: 3,  minFeeTx: 165149
32768 Lovelace  -> CoinSize: 3,  minFeeTx: 165149
65536 Lovelace  -> CoinSize: 5,  minFeeTx: 165237
131072 Lovelace         -> CoinSize: 5,  minFeeTx: 165237
262144 Lovelace         -> CoinSize: 5,  minFeeTx: 165237
524288 Lovelace         -> CoinSize: 5,  minFeeTx: 165237
1048576 Lovelace        -> CoinSize: 5,  minFeeTx: 165237
2097152 Lovelace        -> CoinSize: 5,  minFeeTx: 165237
4194304 Lovelace        -> CoinSize: 5,  minFeeTx: 165237
8388608 Lovelace        -> CoinSize: 5,  minFeeTx: 165237
16777216 Lovelace       -> CoinSize: 5,  minFeeTx: 165237
33554432 Lovelace       -> CoinSize: 5,  minFeeTx: 165237
67108864 Lovelace       -> CoinSize: 5,  minFeeTx: 165237
134217728 Lovelace      -> CoinSize: 5,  minFeeTx: 165237
268435456 Lovelace      -> CoinSize: 5,  minFeeTx: 165237
536870912 Lovelace      -> CoinSize: 5,  minFeeTx: 165237
1073741824 Lovelace     -> CoinSize: 5,  minFeeTx: 165237
2147483648 Lovelace     -> CoinSize: 5,  minFeeTx: 165237
4294967296 Lovelace     -> CoinSize: 9,  minFeeTx: 165413
8589934592 Lovelace     -> CoinSize: 9,  minFeeTx: 165413
17179869184 Lovelace    -> CoinSize: 9,  minFeeTx: 165413
34359738368 Lovelace    -> CoinSize: 9,  minFeeTx: 165413
68719476736 Lovelace    -> CoinSize: 9,  minFeeTx: 165413
137438953472 Lovelace   -> CoinSize: 9,  minFeeTx: 165413
274877906944 Lovelace   -> CoinSize: 9,  minFeeTx: 165413
549755813888 Lovelace   -> CoinSize: 9,  minFeeTx: 165413
1099511627776 Lovelace  -> CoinSize: 9,  minFeeTx: 165413
2199023255552 Lovelace  -> CoinSize: 9,  minFeeTx: 165413
4398046511104 Lovelace  -> CoinSize: 9,  minFeeTx: 165413
8796093022208 Lovelace  -> CoinSize: 9,  minFeeTx: 165413
17592186044416 Lovelace         -> CoinSize: 9,  minFeeTx: 165413
35184372088832 Lovelace         -> CoinSize: 9,  minFeeTx: 165413
70368744177664 Lovelace         -> CoinSize: 9,  minFeeTx: 165413
140737488355328 Lovelace        -> CoinSize: 9,  minFeeTx: 165413
281474976710656 Lovelace        -> CoinSize: 9,  minFeeTx: 165413
562949953421312 Lovelace        -> CoinSize: 9,  minFeeTx: 165413
1125899906842624 Lovelace       -> CoinSize: 9,  minFeeTx: 165413
2251799813685248 Lovelace       -> CoinSize: 9,  minFeeTx: 165413
4503599627370496 Lovelace       -> CoinSize: 9,  minFeeTx: 165413
9007199254740992 Lovelace       -> CoinSize: 9,  minFeeTx: 165413
18014398509481984 Lovelace      -> CoinSize: 9,  minFeeTx: 165413
36028797018963968 Lovelace      -> CoinSize: 9,  minFeeTx: 165413
72057594037927936 Lovelace      -> CoinSize: 9,  minFeeTx: 165413
144115188075855872 Lovelace     -> CoinSize: 9,  minFeeTx: 165413
288230376151711744 Lovelace     -> CoinSize: 9,  minFeeTx: 165413
576460752303423488 Lovelace     -> CoinSize: 9,  minFeeTx: 165413
1152921504606846976 Lovelace    -> CoinSize: 9,  minFeeTx: 165413
2305843009213693952 Lovelace    -> CoinSize: 9,  minFeeTx: 165413
4611686018427387904 Lovelace    -> CoinSize: 9,  minFeeTx: 165413
9223372036854775808 Lovelace    -> CoinSize: 9,  minFeeTx: 165413
```

lehins:
> If You think that re-serializing the transaction is not an issue

I am pretty sure reserializing the transaction is not an issue. Especially since the fixed point is reached quite quickly. MOreover, it  is the most correct and the most general approach.

mesudip:
I've updated PR according to the discussion.

libs/cardano-ledger-api/src/Cardano/Ledger/Api/Tx.hs
```diff
@@ -40,12 +40,25 @@ module Cardano.Ledger.Api.Tx
...
 import Cardano.Ledger.Api.Tx.AuxData
 import Cardano.Ledger.Api.Tx.Body
 import Cardano.Ledger.Api.Tx.Wits
-import Cardano.Ledger.Core (EraTx (..))
+import Cardano.Ledger.Core (EraPParams (PParams), EraTx (..))
 import Cardano.Ledger.Shelley.Tx (ShelleyTx)
+import Lens.Micro ((&), (.~), (^.))
+
+-- Calcluate minfee and create new transaction until the minfee is same.
+
```
lehins: It would be better to convert a comment into an actual haddock
```suggestion
-- | Calculate and update the fee in the transaction until it has the smallest possible value according to
-- the settings in the protocol parameters.
```

lehins (approved):
This looks great!

If you could add a few property tests to the test suite that would be even more awesome.
You could follow  by the example of tests in https://github.com/input-output-hk/cardano-ledger/blob/master/libs/cardano-ledger-api/test/Test/Cardano/Ledger/Api/Tx/Out.hs
And in fact you could probably use your previous implementation to validate the logic in tests.
Let me know if that is something you'd be willing to do and we'll hold off on merging this PR until we have some tests in place.

JaredCorduan (approved):
fantastic, thank you @mesudip !

PR #3205, opened 2022-12-10
Cleanup. Remove some deprecated type synonyms.

lehins:
# Description

This is just a cleanup PR. It takes care of a few things:

* Remove deprecated type synonyms (node has already accounted for those): `ValidatedTx`, `TxOut`, `TxBody`, `TxSeq`, `Script`, `Value`, `PParamDelta`.
* `as Core`  qualified import is not needed in most places due to removal of above deprcations
* Stop importing `Crypto` qualified as `CC`. It no longer conflicts with recently renamed `EraCrypto`
* Rename `Cardano.Ledger.Shelley.Metadata` module to `Cardano.Ledger.Shelley.TxAuxData`

 There aren't really any semantic changes here. Just the cleanup

# Checklist

- [x] Self-reviewed the diff


eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxBody.hs
```diff
@@ -137,34 +132,30 @@ import Prelude hiding (lookup)
...
     atbrWdrls :: !(Wdrl (EraCrypto era)),
     atbrTxFee :: !Coin,
     atbrValidityInterval :: !ValidityInterval,
     atbrUpdate :: !(StrictMaybe (Update era)),
     atbrReqSignerHashes :: Set (KeyHash 'Witness (EraCrypto era)),
     atbrMint :: !(MultiAsset (EraCrypto era)),
-    -- The spec makes it clear that the mint field is a
-    -- Cardano.Ledger.Mary.Value.MaryValue, not a Cardano.Ledger.Core.Value.
-    -- Operations on the TxBody in the AlonzoEra depend upon this.
-    -- We now store only the MultiAsset part of a Mary.Value.
```
Soupstraw, line 151: why is this comment removed?

eras/shelley/impl/src/Cardano/Ledger/Shelley/Tx.hs
```diff
@@ -405,33 +400,34 @@ extractKeyHashWitnessSet = foldr accum Set.empty
...
-  Core.Tx era ->
+  PParams era ->
+  Tx era ->
   Coin
 shelleyMinFeeTx pp tx =
   Coin $
     fromIntegral (getField @"_minfeeA" pp)
-      * tx ^. sizeTxF
+      * tx
+        ^. sizeTxF
```
Soupstraw, line 413: I prefer the way it was written before (or did ormolu do this?)

eras/shelley/impl/src/Cardano/Ledger/Shelley/UTxO.hs
```diff
@@ -57,7 +57,7 @@ import Cardano.Ledger.Shelley.TxBody
   )
 import Cardano.Ledger.TxIn (TxIn (..))
 import Cardano.Ledger.UTxO
-import qualified Cardano.Ledger.UTxO as Core
+import qualified Cardano.Ledger.UTxO as UTxO
```
Soupstraw: Nice catch

eras/shelley/impl/src/Cardano/Ledger/Shelley/UTxO.hs
```diff
@@ -172,7 +172,8 @@ produced ::
 produced pp isNewPool txBody =
   balance (txouts txBody)
     <+> Val.inject
-      ( txBody ^. feeTxBodyL
+      ( txBody
+          ^. feeTxBodyL
```
Soupstraw: I prefer the old style here as well

Soupstraw (approved):
Looks good, just a few nitpicks

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxBody.hs
lehins: That is because the spec needs to be fixed. The spec does not have a separate type `MultiAsset` like we do in Haskell, that does not contain ADA. We also removed a rule that checks that mint can't contain ADA, precisely because it can't happen by construction. So this comment is wrong, that is why I removed it.

Here I created a ticket that tracks fixining the spec: https://github.com/input-output-hk/cardano-ledger/issues/3207

eras/shelley/impl/src/Cardano/Ledger/Shelley/Tx.hs
lehins: I prefer it too. It is the stupid ormolu. It is their recent addition of `--no-cabal` flag that is messing with this. So much for "we don't have any configuration that affects the formatting" mantra. 

I am gonna give it a try and see if I can supply `--no-cabal` flag through hls in my emacs, but I suspect we will see more discrepancies like this between formatting on CI and local

eras/shelley/impl/src/Cardano/Ledger/Shelley/UTxO.hs
lehins: Me too :cry:

Soupstraw:
sorry for the force pushes, I accidentally pushed some changes onto your branch and then undid them :s

lehins:
> sorry for the force pushes, I accidentally pushed some changes onto your branch and then undid them :s

No worries.

JaredCorduan (approved):
fantastic! :raised_hands:

PR #3208, opened 2022-12-12
Replace NominalDiffTime with a newtype wrapper.

### LINKED ISSUE #3185: CBOR serialization of NominalDiffTime is broken
Current `Decoder`/`Encoding` and consequently `FromCBOR`/`ToCBOR` instances for `NominalDiffTime` assume/round the value down to 1/100000 of precision that the type is capable of representing. It seems that the only reason for this strange behavior is the  `ShelleyGenesis.sgSlotLength` field.

This needs fixing. I do not know for sure the reason for the comment in this instance:
```
-- | For backwards compatibility we round pico precision to micro
instance FromCBOR NominalDiffTime where
  fromCBOR = fromRational . (% 1e6) <$> fromCBOR
```

but from the looks of it the only place `NominalDiffTime` is being used in is the `ShelleyGenesis` type, which we have full control over because the value is always set to 1s (thus encoded as 1000000).

So, this ticket requires some investigation and if it turns out that in fact the only place that the `NominalDiffTime` is being used is in the `ShelleyGenesis`, then it needs to be switched to some newtype that wraps around the `NominalDiffTime` with the rounded instance, while the original type `NominalDiffType` should have its instances fixed (or even better totally removed). There is already a `newtype` like this in roundtrip testing, which could be adopted for this purpose.

aniketd:
# Description

Remove CBOR instances of NominalDiffTime and only keep those for the newtype wrapper. This is because we don't use full precision during serialisation.

Closes #3185 

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -58,6 +58,7 @@ in the naming of release branches.
 
 ### Changed
 
+- Replace `NominalDiffTime` with a `newtype` wrapper `NominalDiffTimeMicro`. Remove use of `NominalDiffTime`, as we don't use its full precision. #3208
 - Switched `PlutusDebug` to use a `GADT` with `singletons` for better type-safety. #3167
   - Made `Plutus` imports uniform.
 - Renamed module `Cardano.Ledger.Shelley.Metadata` -> `Cardano.Ledger.Shelley.TxAuxData` #3205
--- a/eras/shelley/impl/src/Cardano/Ledger/Shelley/Genesis.hs
+++ b/eras/shelley/impl/src/Cardano/Ledger/Shelley/Genesis.hs
@@ -28,7 +28,6 @@ module Cardano.Ledger.Shelley.Genesis
     validateGenesis,
     describeValidationErr,
     mkShelleyGlobals,
-    microsecondsToNominalDiffTimeMicro,
     nominalDiffTimeMicroToMicroseconds,
     nominalDiffTimeMicroToSeconds,
     toNominalDiffTimeMicro,
@@ -167,7 +166,7 @@ fromNominalDiffTimeMicro =
 
 toNominalDiffTimeMicroWithRounding :: NominalDiffTime -> NominalDiffTimeMicro
 toNominalDiffTimeMicroWithRounding =
-  microsecondsToNominalDiffTimeMicro . picoToMicro . nominalDiffTimeToSeconds
+  secondsToNominalDiffTimeMicro . picoToMicro . nominalDiffTimeToSeconds
 
 toNominalDiffTimeMicro :: NominalDiffTime -> Maybe NominalDiffTimeMicro
 toNominalDiffTimeMicro ndt
@@ -176,11 +175,8 @@ toNominalDiffTimeMicro ndt
   where
     ndtm = toNominalDiffTimeMicroWithRounding ndt
 
-microsecondsToNominalDiffTimeMicro :: Micro -> NominalDiffTimeMicro
-microsecondsToNominalDiffTimeMicro = NominalDiffTimeMicro
-
-secondsToNominalDiffTimeMicro :: Pico -> NominalDiffTimeMicro
-secondsToNominalDiffTimeMicro = NominalDiffTimeMicro . picoToMicro
+secondsToNominalDiffTimeMicro :: Micro -> NominalDiffTimeMicro
+secondsToNominalDiffTimeMicro = NominalDiffTimeMicro
 
 nominalDiffTimeMicroToMicroseconds :: NominalDiffTimeMicro -> Micro
 nominalDiffTimeMicroToMicroseconds (NominalDiffTimeMicro microseconds) = microseconds
--- a/eras/shelley/impl/src/Cardano/Ledger/Shelley/Genesis.hs
+++ b/eras/shelley/impl/src/Cardano/Ledger/Shelley/Genesis.hs
@@ -5,18 +5,22 @@
 {-# LANGUAGE DerivingStrategies #-}
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE FlexibleInstances #-}
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
 {-# LANGUAGE NamedFieldPuns #-}
+{-# LANGUAGE NumericUnderscores #-}
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE StandaloneDeriving #-}
 {-# LANGUAGE TypeApplications #-}
 {-# LANGUAGE TypeFamilies #-}
 {-# LANGUAGE UndecidableInstances #-}
+{-# OPTIONS_GHC -Wno-orphans #-}
 
 module Cardano.Ledger.Shelley.Genesis
   ( ShelleyGenesisStaking (..),
     ShelleyGenesis (..),
     ValidationErr (..),
+    NominalDiffTimeMicro (..),
     emptyGenesisStaking,
     sgActiveSlotCoeff,
     genesisUTxO,
@@ -24,6 +28,13 @@ module Cardano.Ledger.Shelley.Genesis
     validateGenesis,
     describeValidationErr,
     mkShelleyGlobals,
+    microsecondsToNominalDiffTimeMicro,
+    nominalDiffTimeMicroToMicroseconds,
+    nominalDiffTimeMicroToSeconds,
+    toNominalDiffTimeMicro,
+    toNominalDiffTimeMicroWithRounding,
+    fromNominalDiffTimeMicro,
+    secondsToNominalDiffTimeMicro,
   )
 where
 
@@ -47,8 +58,10 @@ import Cardano.Ledger.Binary
     FromCBOR (..),
     ToCBOR (..),
     cborError,
+    decodeInteger,
     decodeRational,
     decodeRecordNamed,
+    encodeInteger,
     encodeListLen,
     enforceDecoderVersion,
     enforceEncodingVersion,
@@ -70,14 +83,21 @@ import Cardano.Slotting.EpochInfo (EpochInfo)
 import Cardano.Slotting.Time (SystemStart (SystemStart))
 import Data.Aeson (FromJSON (..), ToJSON (..), (.!=), (.:), (.:?), (.=))
 import qualified Data.Aeson as Aeson
+import Data.Fixed (E6, Fixed (..), HasResolution (resolution), Micro, Pico)
 import qualified Data.ListMap as LM
 import Data.Map.Strict (Map)
 import qualified Data.Map.Strict as Map
 import Data.Maybe (catMaybes)
 import Data.Proxy (Proxy (..))
+import Data.Ratio (numerator, (%))
 import Data.Text (Text)
 import qualified Data.Text as Text
-import Data.Time (NominalDiffTime, UTCTime (..))
+import Data.Time.Clock
+  ( NominalDiffTime,
+    UTCTime (..),
+    nominalDiffTimeToSeconds,
+    secondsToNominalDiffTime,
+  )
 import Data.Unit.Strict (forceElemsToWHNF)
 import Data.Word (Word32, Word64)
 import GHC.Generics (Generic)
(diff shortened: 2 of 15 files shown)


eras/shelley/impl/src/Cardano/Ledger/Shelley/Genesis.hs
```diff
@@ -129,6 +143,37 @@ emptyGenesisStaking =
       sgsStake = mempty
     }
 
+-- | We do not handle the full precision of NominalDiffTime.
+newtype NominalDiffTimeRounded = NominalDiffTimeRounded NominalDiffTime
```
lehins: On a second thought. Let's switch this `newtype` around the [`Micro`](https://hackage.haskell.org/package/base-4.14.1.0/docs/Data-Fixed.html#t:Micro) type.
```suggestion
-- Unlike `NominalDiffTime` that supports `Pico` precision, this type only supports `Micro` precision
newtype NominalDiffTimeMicro = NominalDiffTimeMicro Micro
```

Also implementing and exporting these three functions:
```
toNominalDiffTimeMicro :: NominalDiffTime -> Maybe NominalDiffTimeMicro

toNominalDiffTimeMicroWithRounding :: NominalDiffTime -> NominalDiffTimeMicro

fromNominalDiffTimeMicro :: NominalDiffTimeMicro -> NominalDiffTime
```

This way the type will be correct b y construction, instead of it rounding at the time of serialization

lehins (changes requested):
This looks great, but I just realized  we can do even better. Let me know if you  have questions about this

eras/shelley/impl/src/Cardano/Ledger/Shelley/Genesis.hs
aniketd: Awesome! :raised_hands:

JaredCorduan (approved):
looks great to me, thank you @aniketd !

eras/shelley/impl/src/Cardano/Ledger/Shelley/Genesis.hs
aniketd: I have updated the PR with these additions! :raised_hands:

lehins: It is still a newtype wrapper around the `NominalDiffTime`. The point is that `NominalDiffTime` is a newtype wrapper around [`Pico`](https://downloads.haskell.org/ghc/9.2.5/docs/html/libraries/base-4.16.4.0/Data-Fixed.html#t:Pico) and we want a type that is a wrapper around [`Micro`](https://downloads.haskell.org/ghc/9.2.5/docs/html/libraries/base-4.16.4.0/Data-Fixed.html#t:Micro)

CHANGELOG.md
```diff
@@ -45,6 +45,7 @@ in the naming of release branches.
 
 ### Changed
 
+- Replace `NominalDiffTime` with a `newtype` wrapper `NominalDiffTimeRounded`. Remove use of `NominalDiffTime`, as we don't use its full precision. #3208
```
```suggestion
- Replace `NominalDiffTime` with a `newtype` wrapper `NominalDiffTimeMicro`. Remove use of `NominalDiffTime`, as we don't use its full precision. #3208
```

eras/shelley/impl/src/Cardano/Ledger/Shelley/Genesis.hs
```diff
@@ -129,6 +146,52 @@ emptyGenesisStaking =
...
+
+-- | For backwards compatibility we round pico precision to micro
+decodeNominalDiffTimeMicro :: Decoder s NominalDiffTimeMicro
+decodeNominalDiffTimeMicro = NominalDiffTimeMicro <$> fromRational . (% 1_000_000) <$> decodeInteger
+
+nominalDiffTimeMicroToSeconds :: NominalDiffTimeMicro -> Pico
+nominalDiffTimeMicroToSeconds (NominalDiffTimeMicro ndtm) = nominalDiffTimeToSeconds ndtm
+
+secondsToNominalDiffTimeMicro :: Pico -> NominalDiffTimeMicro
+secondsToNominalDiffTimeMicro (MkFixed s) =
```
lehins: With my suggestion this function would become:
```suggestion
secondsToNominalDiffTimeMicro :: Micro -> NominalDiffTimeMicro
secondsToNominalDiffTimeMicro  = NominalDiffTimeMicro
```

eras/shelley/impl/src/Cardano/Ledger/Shelley/Genesis.hs
```diff
@@ -129,6 +146,52 @@ emptyGenesisStaking =
...
+  | otherwise = Nothing
+  where
+    ndtm = toNominalDiffTimeMicroWithRounding ndt
+
+toNominalDiffTimeMicroWithRounding :: NominalDiffTime -> NominalDiffTimeMicro
+toNominalDiffTimeMicroWithRounding ndt =
+  secondsToNominalDiffTimeMicro $ nominalDiffTimeToSeconds ndt
+
+fromNominalDiffTimeMicro :: NominalDiffTimeMicro -> NominalDiffTime
+fromNominalDiffTimeMicro (NominalDiffTimeMicro ndt) = ndt
```
lehins: This would become
```suggestion
fromNominalDiffTimeMicro (NominalDiffTimeMicro (MkFixed ndt)) = NominalDiffTimeMicro (MkFixed (ndtm * 1_000_000))
```

aniketd: I get it now! Thanks! :raised_hands:

eras/shelley/impl/src/Cardano/Ledger/Shelley/Genesis.hs
aniketd: I don't know how I missed that detail. :face_exhaling:. I have pushed the latest changes now :raised_hands:

eras/shelley/impl/src/Cardano/Ledger/Shelley/Genesis.hs
```diff
@@ -129,6 +149,71 @@ emptyGenesisStaking =
...
+-- | Unlike @'NominalDiffTime'@ that supports @'Pico'@ precision, this type
+-- only supports @'Micro'@ precision.
+newtype NominalDiffTimeMicro = NominalDiffTimeMicro Micro
+  deriving (Show, Eq, Generic)
+  deriving anyclass (NoThunks)
+  deriving newtype (Ord, Num, Fractional, Real, ToJSON, FromJSON)
+
+deriving instance Generic Micro
+
+deriving anyclass instance NoThunks Micro
```
lehins: It is better to move this instance into `Cardano.Ledger.Orphans` and leave  a `TODO` that we should upstream instance `HasResolution p => NoThunks (Fixed p)` into `nothunks` package
```suggestion
deriving anyclass instance NoThunks Micro
```

eras/shelley/impl/src/Cardano/Ledger/Shelley/Genesis.hs
```diff
@@ -129,6 +149,71 @@ emptyGenesisStaking =
...
+
+instance ToCBOR NominalDiffTimeMicro where
+  toCBOR = encodeNominalDiffTimeMicro
+
+encodeNominalDiffTimeMicro :: NominalDiffTimeMicro -> Encoding
+encodeNominalDiffTimeMicro (NominalDiffTimeMicro micro) =
+  encodeInteger . toMicroseconds $ micro
+  where
+    toMicroseconds t =
+      numerator (toRational t * toRational (resolution $ Proxy @E6))
```
lehins: This one is much easier ;) 
```suggestion
encodeNominalDiffTimeMicro (NominalDiffTimeMicro (MkFixed micro)) = encodeInteger micro
```

eras/shelley/impl/src/Cardano/Ledger/Shelley/Genesis.hs
```diff
@@ -129,6 +149,71 @@ emptyGenesisStaking =
...
+
+instance FromCBOR NominalDiffTimeMicro where
+  fromCBOR = decodeNominalDiffTimeMicro
+
+decodeNominalDiffTimeMicro :: Decoder s NominalDiffTimeMicro
+decodeNominalDiffTimeMicro =
+  NominalDiffTimeMicro . fromRational . (% 1_000_000) <$> decodeInteger
```
lehins: I think this one is wrong, it should be:
```suggestion
  NominalDiffTimeMicro . MKFixed <$> decodeInteger
```

eras/shelley/impl/src/Cardano/Ledger/Shelley/Genesis.hs
```diff
@@ -129,6 +149,71 @@ emptyGenesisStaking =
...
+  microsecondsToNominalDiffTimeMicro . picoToMicro . nominalDiffTimeToSeconds
+
+toNominalDiffTimeMicro :: NominalDiffTime -> Maybe NominalDiffTimeMicro
+toNominalDiffTimeMicro ndt
+  | roundedPicoseconds == picoseconds = Just ndtm
+  | otherwise = Nothing
+  where
+    ndtm@(NominalDiffTimeMicro microseconds) = toNominalDiffTimeMicroWithRounding ndt
+    picoseconds = nominalDiffTimeToSeconds ndt
+    roundedPicoseconds = microToPico microseconds
```
lehins: This can also be simpler:
```suggestion
  | fromNominalDiffTimeMicro ndtm == ndt = Just ndtm
  | otherwise = Nothing
  where
    ndtm = toNominalDiffTimeMicroWithRounding ndt
```

eras/shelley/impl/src/Cardano/Ledger/Shelley/Genesis.hs
```diff
@@ -129,6 +149,71 @@ emptyGenesisStaking =
...
+  deriving (Show, Eq, Generic)
+  deriving anyclass (NoThunks)
+  deriving newtype (Ord, Num, Fractional, Real, ToJSON, FromJSON)
+
+deriving instance Generic Micro
+
+deriving anyclass instance NoThunks Micro
+
+instance ToCBOR NominalDiffTimeMicro where
+  toCBOR = encodeNominalDiffTimeMicro
```
lehins: In fact the `To/FromCBOR` instances can be derived using GND. All you need is to add these standalone instances to:

`Cardano.Ledger.Binary.Encoding.ToCBOR` and `Cardano.Ledger.Binary.Decoding.FromCBOR`:

```haskell
deriving newtype instance ToCBOR (MkFixed p)
deriving newtype instance FromCBOR (MkFixed p)
```

eras/shelley/test-suite/src/Test/Cardano/Ledger/Shelley/Serialisation/EraIndepGenerators.hs
```diff
@@ -155,6 +155,10 @@ instance (Era era, Mock (EraCrypto era)) => Arbitrary (Update era) where
   arbitrary = genericArbitraryU
   shrink = genericShrink
 
+instance Arbitrary NominalDiffTimeMicro where
+  arbitrary = secondsToNominalDiffTimeMicro <$> arbitrary
+  shrink = fmap secondsToNominalDiffTimeMicro . shrink . nominalDiffTimeMicroToSeconds
```
lehins: This is too much work, instead
```suggestion
deriving newtype instance Arbitrary NominalDiffTimeMicro
```

lehins (changes requested):
Looks good. Deserializer is a bit borked, but there is a really easy fix for it

eras/shelley/impl/src/Cardano/Ledger/Shelley/Genesis.hs
aniketd: Awesome! Thanks! :bow:

eras/shelley/impl/src/Cardano/Ledger/Shelley/Genesis.hs
aniketd: I found instances already present for `Nano`, `Pico`, so I added similar ones directly for `Micro` instead of deriving. Is that okay?

eras/shelley/impl/src/Cardano/Ledger/Shelley/Genesis.hs
aniketd: Now that we have the instance for `Micro` the codec was easy to convert to `deriving newtype (ToCBOR, FromCBOR)`

eras/shelley/impl/src/Cardano/Ledger/Shelley/Genesis.hs
aniketd: Same here as for the `encode`. This could be changed to `deriving newtype (FromCBOR)` as we added the instance for `Micro` to `Cardano.Ledger.Binary.Decoding.FromCBOR`. :+1:

eras/shelley/impl/src/Cardano/Ledger/Shelley/Genesis.hs
aniketd: Awesome! Thanks! :bow:

eras/shelley/impl/src/Cardano/Ledger/Shelley/Genesis.hs
aniketd: I have replaced all instances (`Nano`, `Pico`, `Micro`) with `deriving newtype instance Typeable p => {To|From}CBOR (Fixed p)`

eras/shelley/impl/src/Cardano/Ledger/Shelley/Genesis.hs
```diff
@@ -129,6 +146,48 @@ emptyGenesisStaking =
...
+  | fromNominalDiffTimeMicro ndtm == ndt = Just ndtm
+  | otherwise = Nothing
+  where
+    ndtm = toNominalDiffTimeMicroWithRounding ndt
+
+microsecondsToNominalDiffTimeMicro :: Micro -> NominalDiffTimeMicro
+microsecondsToNominalDiffTimeMicro = NominalDiffTimeMicro
+
+secondsToNominalDiffTimeMicro :: Pico -> NominalDiffTimeMicro
+secondsToNominalDiffTimeMicro = NominalDiffTimeMicro . picoToMicro
```
lehins: Having both of these functions is confusing and unnecessary. Because:
```haskell
> secondsToNominalDiffTimeMicro 15 == microsecondsToNominalDiffTimeMicro 15
True
```

I suggest keep only one of them and rename it:
```haskell
microToNominalDiffTimeMicro :: Micro -> NominalDiffTimeMicro
microToNominalDiffTimeMicro = NominalDiffTimeMicro
```
Alternative name for the above function could also be `secondsToNominalDiffTimeMicro`, which would stay inline with the previous version. Point is that when you call these functions above you don't supply microseconds or picoseconds, but rather seconds with microsecond or picosecond precision respectively. So, feel free to choose whichever name you like better, the important part is not to keep two functions with morally conflicting names.

libs/cardano-ledger-core/src/Cardano/Ledger/Orphans.hs
```diff
@@ -60,6 +66,12 @@ instance NoThunks WC.XSignature where
 instance SignableRepresentation (Hash.Hash a b) where
   getSignableRepresentation = Hash.hashToBytes
 
+-- | TODO: We should upstream instance
+-- HasResolution p => NoThunks (Fixed p) into the nothunks package.
+deriving anyclass instance NoThunks Micro
+
+deriving instance Generic Micro
```
lehins: I am worried that it can lead to duplicate instance somewhere down stream. I don't think we need this either

libs/cardano-ledger-core/src/Cardano/Ledger/Orphans.hs
```diff
@@ -60,6 +66,12 @@ instance NoThunks WC.XSignature where
 instance SignableRepresentation (Hash.Hash a b) where
   getSignableRepresentation = Hash.hashToBytes
 
+-- | TODO: We should upstream instance
+-- HasResolution p => NoThunks (Fixed p) into the nothunks package.
+deriving anyclass instance NoThunks Micro
```
lehins: `Fixed` is a just a newtype wrapper around Integer, so this way is better
```suggestion
deriving newtype instance NoThunks (Fixed p)
```

libs/cardano-ledger-core/src/Cardano/Ledger/Orphans.hs
```diff
@@ -1,7 +1,11 @@
+{-# LANGUAGE DeriveAnyClass #-}
+{-# LANGUAGE DeriveGeneric #-}
 {-# LANGUAGE DerivingStrategies #-}
+{-# LANGUAGE FlexibleInstances #-}
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE StandaloneDeriving #-}
+{-# LANGUAGE TypeSynonymInstances #-}
```
lehins: These are no longer needed with my comments below
```suggestion
{-# LANGUAGE DerivingStrategies #-}
{-# LANGUAGE FlexibleInstances #-}
{-# LANGUAGE GeneralizedNewtypeDeriving #-}
{-# LANGUAGE ScopedTypeVariables #-}
{-# LANGUAGE StandaloneDeriving #-}
```

lehins (approved):
Looks great. Just a few final comments.

eras/shelley/impl/src/Cardano/Ledger/Shelley/Genesis.hs
aniketd: > when you call these functions above you don't supply microseconds or picoseconds, but rather seconds

Thanks for that! :100: :bow: I completely failed at understanding this :heart:

libs/cardano-ledger-core/src/Cardano/Ledger/Orphans.hs
aniketd: This is necessary due to the `NoThunks` derivation. GHC complains without it. Is there a better place to put it may be? Or some other way?
```haskell
deriving newtype instance NoThunks (Fixed p)
deriving stock instance Generic (Fixed p)
```
...seems to work out.

aniketd: Sorry, I'm wrong here, by keeping the `DeriveGeneric`. Removing it and just keeping `newtype` with `GeneralizedNewtypeDeriving` did the trick. Thanks a ton! :bow: :heart_eyes: :spiral_notepad: :man_student: :exploding_head:

PR #3209, opened 2022-12-12
optimize TICKF

JaredCorduan:
# Description

This PR optimizes the `TICKF` transition, particularly for the problematically slow computations on the epoch boundary. See the ADR in this PR for more context. See also #3034.

This PR consists of four commits: one for each of the two optimizations mentioned in the ADR, one for the ADR, and one to fix a broken link in the PR template.

I did not update the changelog since all the changes are invisible outside of this code base.

replaces #3140 and #3141

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/docs/adr/2022-12-12_007-optimize-ledger-view.md
+++ b/docs/adr/2022-12-12_007-optimize-ledger-view.md
@@ -1,0 +1,63 @@
+---
+slug: 7
+title: |
+  7. Optimize TICKF
+authors: [Jared Corduan]
+tags: [Accepted]
+---
+
+## Status
+
+Accepted
+
+## Context
+
+Large computations on the epoch boundary are problematic.
+In the past they have led to delayed block production in new epochs.
+We have removed most of the problems surrounding large computations in the ledger on the epoch
+boundary (such as making the reward calculation and the stake distribution calculation incremental),
+but benchmarks from the consensus team still indicate performance problems in the ledger.
+In particular, the `TICKF` transition, which is used for forecasting the ledger state in order
+to check block headers and to check for leadership, still take too long.
+The node needs to check for leadership once per slot (so once per second on mainnet),
+so if `TICKF` takes a second to compute then the node is guaranteed to start skipping leadership checks.
+The performance problems with `TICKF` can exacerbate situations like
+[this](https://github.com/input-output-hk/cardano-node/issues/4421).
+
+Upon examination, the `TICKF` rule is still performing at least one large computation which can
+be optimized, namely computing the per-pool stake distribution from the per-stake-credential
+stake distribution using the delegation relation.
+The computation is currently being done on the epoch boundary when it is first needed,
+but could be performed on the previous epoch boundary.
+In particular, the leadership check appears to be computing it repeatedly on every slot
+crossing the epoch boundary until the first block in the new epoch is made.
+
+The per-pool stake distribution, however, only appears to account for about half of the
+duration of the `TICKF` rule (from our benchmarks).
+
+## Decision
+
+@nfrisby suggested two very helpful optimizations
+(but we take full responsibility for the implementation).
+
+The first suggestion was to create a thunk for the pool distribution on the first epoch
+boundary when the computation is able to be performed.
+This way the node will only have to pay the computational cost once, instead of at every
+slot in a new epoch until the first block is produced.
+In particular, we can add a lazy field to the `SnapShots` record to memoize the computation.
+
+The second suggestion was to strip down `TICKF` to only what is required to compute the 'LedgerView'.
+
+Preliminary benchmarks from the consensus team show that these two optimizations remove the
+problematic behavior on the epoch boundary.
+
+## Consequences
+
+The most dramatic consequence of these two decisions is the rapid leader checks on the epoch
+boundary.
+
+On the negative side, we have now diverged further from the formal specification
+(though it is semantically the same).
+Furthermore, the `TICKF` implementation is less obviously correct, as it has to duplicate logic.
+To combat the problem that `TICKF` could diverge from `TICK` with respect to the
+ledger view, we added a property test.
--- a/eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/NewEpoch.hs
+++ b/eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/NewEpoch.hs
@@ -17,10 +17,9 @@ module Cardano.Ledger.Shelley.Rules.NewEpoch
     ShelleyNewEpochPredFailure (..),
     ShelleyNewEpochEvent (..),
     PredicateFailure,
+    updateRewards,
     calculatePoolDistr,
     calculatePoolDistr',
-    updateRewards,
-    calculatePoolStake,
   )
 where
 
@@ -30,12 +29,12 @@ import Cardano.Ledger.BaseTypes
     ShelleyBase,
     StrictMaybe (SJust, SNothing),
   )
-import Cardano.Ledger.Coin (Coin (Coin), CompactForm (CompactCoin), toDeltaCoin)
+import Cardano.Ledger.Coin (toDeltaCoin)
 import Cardano.Ledger.Core
 import Cardano.Ledger.Credential (Credential)
 import Cardano.Ledger.EpochBoundary
-import Cardano.Ledger.Keys (KeyHash, KeyRole (StakePool, Staking))
-import Cardano.Ledger.PoolDistr (IndividualPoolStake (..), PoolDistr (..))
+import Cardano.Ledger.Keys (KeyRole (Staking))
+import Cardano.Ledger.PoolDistr (PoolDistr (..))
 import Cardano.Ledger.Shelley.AdaPots (AdaPots, totalAdaPotsES)
 import Cardano.Ledger.Shelley.Era (ShelleyNEWEPOCH)
 import Cardano.Ledger.Shelley.LedgerState
@@ -43,18 +42,14 @@ import Cardano.Ledger.Shelley.Rewards (sumRewards)
 import Cardano.Ledger.Shelley.Rules.Epoch
 import Cardano.Ledger.Shelley.Rules.Mir (ShelleyMIR, ShelleyMirEvent, ShelleyMirPredFailure)
 import Cardano.Ledger.Shelley.Rules.Rupd (RupdEvent (..))
-import Cardano.Ledger.Shelley.TxBody (PoolParams (ppVrf))
 import Cardano.Ledger.Slot (EpochNo (EpochNo))
 import qualified Cardano.Ledger.Val as Val
 import Control.State.Transition
 import Data.Default.Class (Default, def)
 import qualified Data.Map.Strict as Map
-import Data.Ratio ((%))
 import Data.Set (Set)
-import Data.VMap as VMap
 import GHC.Generics (Generic)
 import GHC.Records (HasField)
-import GHC.Word (Word64)
 import NoThunks.Class (NoThunks (..))
 
 data ShelleyNewEpochPredFailure era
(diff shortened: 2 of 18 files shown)


libs/cardano-ledger-core/src/Cardano/Ledger/EpochBoundary.hs
```diff
@@ -181,15 +186,19 @@ instance CC.Crypto c => FromSharedCBOR (SnapShot c) where
     pure SnapShot {ssStake, ssDelegations, ssPoolParams}
 
 -- | Snapshots of the stake distribution.
+--
+-- Note that ssStakeMark and ssStakeMarkPoolDistr are lazy on
+-- purpose since we only want to force the thunk after one stability window
+-- when we know that they are stable (so that we do not compute them if we do not have to).
```
lehins, line 192: Maybe adding a url to the ADR in this PR:
```suggestion
-- when we know that they are stable (so that we do not compute them if we do not have to).
-- See more info in the [Optimize TICKF ADR](https://github.com/input-output-hk/cardano-ledger/blob/master/docs/adr/2022-12-12_007-optimize-ledger-view.md)
```

libs/cardano-ledger-core/src/Cardano/Ledger/EpochBoundary.hs
```diff
@@ -200,6 +209,7 @@ instance
   toCBOR (SnapShots {ssStakeMark, ssStakeSet, ssStakeGo, ssFee}) =
     encodeListLen 4
       <> toCBOR ssStakeMark
+      -- We intentionalyl do not serialize the redundant ssStakeMarkPoolDistr
```
lehins:
```suggestion
      -- We intentionaly do not serialize the redundant ssStakeMarkPoolDistr
```

docs/adr/2022-12-12_007-optimize-ledger-view.md
```diff
@@ -0,0 +1,63 @@
...
+
+## Status
+
+Accepted
+
+## Context
+
+Large computations on the epoch boundary are problematic.
+In the past they have led to delayed block production in new epochs.
+We have removed most of the problems surrounding large computations in the ledger on the epooch
```
aniketd:
```suggestion
We have removed most of the problems surrounding large computations in the ledger on the epoch
```

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Tick.hs
```diff
@@ -167,6 +175,63 @@ validatingTickTransition nes slot = do
...
+  if unEpochNo epoch /= unEpochNo (nesEL nes) + 1
+    then pure $ nes {nesEs = adoptGenesisDelegs es slot}
+    else do
+      -- We can skip 'SNAP'; we already have the equivalent pd'.
+
+      -- We can skip 'MIR' and 'POOLREAP';
+      -- we don't need to do the checks:
+      -- if the checks would fail, then the node will fail in the 'TICK' rule
+      -- if it ever then node tries to validate blocks for which the
+      -- return value here was used to validate their headers.
```
aniketd, line 220: I found it hard to read these two lines. It feels like some words may be missing.

eras/shelley/test-suite/test/Test/Cardano/Ledger/Shelley/RulesTests.hs
```diff
@@ -41,8 +56,15 @@ import Test.Cardano.Ledger.Shelley.MultiSigExamples
     applyTxWithScript,
     bobOnly,
   )
+import Test.Cardano.Ledger.Shelley.Serialisation.EraIndepGenerators ()
+-- EraIndepGenerators is neded for Arbitrary NewEpochState
+import Test.Cardano.Ledger.Shelley.Serialisation.Generators ()
+-- Generators is neded for Arbitrary ShelleyPParams
```
aniketd:
```suggestion
-- Generators is needed for Arbitrary ShelleyPParams
```

aniketd (changes requested):
I understand this a tad-bit better after reviewing this PR. Thank you! :raised_hands:

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Tick.hs
JaredCorduan: yea, it's not great. how does this sound:

```
We can skip 'MIR' and 'POOLREAP'.
We do not need to do the checks in the 'TICK' rule since they are not actually
validating user input, but are instead a protection for the internal consistency
of the ledger state.
```

eras/shelley/test-suite/test/Test/Cardano/Ledger/Shelley/RulesTests.hs
```diff
@@ -520,3 +542,50 @@ testRwdAliceSignsAlone''' =
...
+      SJust (Complete rewardUpdate) ->
+        SJust . Complete $ rewardUpdate {rs = removeEmptyRewards (rs rewardUpdate)}
+
+setDepositsToObligation :: NewEpochState Shelley -> NewEpochState Shelley
+setDepositsToObligation (NewEpochState el bprev bcur (EpochState as ss ls ppp pp nm) ru pd stash) =
+  NewEpochState el bprev bcur (EpochState as ss ls' ppp pp nm) ru pd stash
+  where
+    LedgerState (UTxOState utxo _deposited fees ppups stakeDistr) dpstate = ls
+    deposited = obligationDPState dpstate
+    ls' = LedgerState (UTxOState utxo deposited fees ppups stakeDistr) dpstate
```
lehins: FYI, I am not suggesting to change this implementation, I just wanted to make a comment that is related to our discussion about pattern matches vs record syntax. Precisely in this kind of patterns it would be much clearer for the future reader and the current reader of the code to understand what is going on, because it is obvious what is being modified and which fields are needed for this update. All because we remove all the unnecessary noise. More over it  would make it future proof (i.e. account for field reordering, addition or removal):
```suggestion
setDepositsToObligation nes = nes { nesEs = es { esLState = ls { lsUTxOState =  utxoState }} }
  where
    es = nesEs nes
    ls = esLState es
    utxoState = (lsUTxOState ls) { utxosDeposited = obligationDPState $ lsDPState ls }
```

Naturally, with lenses this would be even more clear. Maybe in the future will define some lenses for all our types.

eras/shelley/test-suite/test/Test/Cardano/Ledger/Shelley/RulesTests.hs
```diff
@@ -520,3 +542,50 @@ testRwdAliceSignsAlone''' =
...
+
+-- | This property test checks the correctness of the TICKF transation.
+-- TICKF is used by the consensus layer to get a ledger view in a computationally
+-- cheaper way than using the TICK rule.
+-- Therefore TICKF and TICK need to compute the same ledger view.
+propTickfPerservesLedgerView :: NewEpochState Shelley -> Property
+propTickfPerservesLedgerView nes =
+  let (EpochNo e) = nesEL nes
+      slot = slotFromEpoch (EpochNo $ e + 1)
+      nes' = (setDepositsToObligation . filterEmptyRewards) nes
```
lehins: Two characters and one function less :slightly_smiling_face: 
```suggestion
      nes' = setDepositsToObligation (filterEmptyRewards nes)
```

eras/shelley/test-suite/test/Test/Cardano/Ledger/Shelley/RulesTests.hs
```diff
@@ -520,3 +542,50 @@ testRwdAliceSignsAlone''' =
...
+      tickNes = runShelleyBase $ applySTSTest @(ShelleyTICK Shelley) (TRC ((), nes', slot))
+      tickFNes = runShelleyBase $ applySTSTest @(ShelleyTICKF Shelley) (TRC ((), nes', slot))
+   in case (tickNes, tickFNes) of
+        (_, Left _) -> discard
+        (Left _, _) -> discard
+        (Right tickNes', Right tickFNes') ->
+          currentLedgerView tickNes' === currentLedgerView tickFNes'
```
lehins: Pattern matches with tuples are usually hard to read, because what you mentally have to do is go through all potential cases, which can usually be replaced with monadic or even applicative syntax. Consider this alternative:
```suggestion
   in either discard id $ do
          tickNes' <- tickNes
          tickNes' <- tickFNes
          pure $ currentLedgerView tickNes' === currentLedgerView tickFNes'
```
Or this:
```suggestion
   in either discard id $ 
         (===) 
         <$> (currentLedgerView <$> tickNes)
         <*> (currentLedgerView <$> tickFNes)
```

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Tick.hs
```diff
@@ -167,6 +175,63 @@ validatingTickTransition nes slot = do
...
+
+  epoch <- liftSTS $ do
+    ei <- asks epochInfoPure
+    epochInfoEpoch ei slot
+
+  let es = nesEs nes
+      ss = esSnapshots es
+
+  -- the relevant 'NEWEPOCH' logic
+  let pd' = ssStakeMarkPoolDistr ss
```
lehins, line 208: I know that `pd'` is lazy, but for clarity that it is not used only in the `else` clause, maybe move it there as well

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/NewEpoch.hs
```diff
@@ -175,8 +170,24 @@ newEpochTransition = do
       es''' <- trans @(EraRule "EPOCH" era) $ TRC ((), es'', e)
       let adaPots = totalAdaPotsES es'''
       tellEvent $ TotalAdaPotsEvent adaPots
-      let ss = esSnapshots es'''
-          pd' = calculatePoolDistr (ssStakeSet ss)
+      let pd' = ssStakeMarkPoolDistr (esSnapshots es)
+      -- The spec sets pd' with:
+      -- pd' = ssStakeSet = calculatePoolDistr (ssStakeSet $ esSnapshots es'''),
```
lehins:
```suggestion
      -- pd' = calculatePoolDistr (ssStakeSet $ esSnapshots es'''),
```

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/NewEpoch.hs
```diff
@@ -175,8 +170,24 @@ newEpochTransition = do
...
+      --
+      -- This is equivalent to:
+      -- pd' = ssStakeMarkPoolDistr (esSnapshots es)
+      --
+      -- since:
+      --
+      -- \* SNAP rotates mark to set, so
+      -- \* the set snapshot in es''' is mark in es
+      -- \* ssStakeMarkPoolDistr is computed by calling calculatePoolDistr
+      --    on the mark snapshot at the previous epoch boundary.
```
lehins: In order to disambiguate from actual verbs in a sentence "mark" and "set"
```suggestion
      -- \* SNAP rotates `ssStakeMark` to `ssStakeSet`, so
      -- \* the `ssStakeSet` snapshot in es''' is `ssStakeMark` in es
      -- \* `ssStakeMarkPoolDistr` is computed by calling `calculatePoolDistr`
      --    on the `ssStakeMark` snapshot at the previous epoch boundary.
```

lehins (approved):
This makes sense. Just a few minor suggestions

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Tick.hs
aniketd: Yes, that sounds better! :+1:

eras/shelley/test-suite/test/Test/Cardano/Ledger/Shelley/RulesTests.hs
JaredCorduan: that's much nicer! I think I prefer the do notation.

eras/shelley/test-suite/test/Test/Cardano/Ledger/Shelley/RulesTests.hs
```diff
@@ -520,3 +542,48 @@ testRwdAliceSignsAlone''' =
...
+propTickfPerservesLedgerView :: NewEpochState Shelley -> Property
+propTickfPerservesLedgerView nes =
+  let (EpochNo e) = nesEL nes
+      slot = slotFromEpoch (EpochNo $ e + 1)
+      nes' = setDepositsToObligation (filterEmptyRewards nes)
+      tickNes = runShelleyBase $ applySTSTest @(ShelleyTICK Shelley) (TRC ((), nes', slot))
+      tickFNes = runShelleyBase $ applySTSTest @(ShelleyTICKF Shelley) (TRC ((), nes', slot))
+   in either discard id $ do
+          tickNes' <- tickNes
+          tickNes' <- tickFNes
```
JaredCorduan: @lehins I'll need to undo the suggestion, the return types unfortunately do not match. the trouble is that the predicate failures are different, so the `Left`s are not the same type. I'll revert it to what is was before, unless you have another clever way to make it more readable.

eras/shelley/test-suite/test/Test/Cardano/Ledger/Shelley/RulesTests.hs
lehins: > the return types unfortunately do not match

Oh yeah, that makes sense, the `Left` type doesn't match

> unless you have another clever way to make it more readable

Of course I do :slightly_smiling_face: :
```haskell
   in fromMaybe discard $ do
          Right tickNes' <- pure tickNes
          Right tickNes' <- pure tickFNes
          pure $ currentLedgerView tickNes' === currentLedgerView tickFNes'
```

JaredCorduan: :metal:

aniketd (approved):
Awesome stuff! :+1:

PR #3210, opened 2022-12-13
Move `KeyPair` to test lib

### LINKED ISSUE #3186: KeyPair should be moved to tests
The `KeyPair` type in `Cardano.Ledger.Keys` contains both the private and the public keys, which are never used in the ledger. It doesn't seem like anyone downstream uses that type. The only place were it is useful is in tests, therefore it should be moved to the new `cardano-ledger-core:teslib`  sublibrary introduced in #3184

teodanciu:
# Description

Closes https://github.com/input-output-hk/cardano-ledger/issues/3186

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/eras/shelley/impl/src/Cardano/Ledger/Shelley/API/Types.hs
+++ b/eras/shelley/impl/src/Cardano/Ledger/Shelley/API/Types.hs
@@ -50,7 +50,7 @@ import Cardano.Ledger.Keys as X
     Hash,
     KESignable,
     KeyHash (..),
-    KeyPair (..),
+    KeyPair (..), -- deprecated
     KeyRole (..),
     SignKeyDSIGN,
     SignKeyKES,
@@ -85,7 +85,7 @@ import Cardano.Ledger.Shelley.LedgerState as X
     EpochState (..),
     IncrementalStake (..),
     InstantaneousRewards (..),
-    KeyPairs,
+    KeyPairs, -- deprecated
     LedgerState (..),
     NewEpochState (..),
     PPUPState (..),
--- a/eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState/Types.hs
+++ b/eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState/Types.hs
@@ -16,6 +16,7 @@
 {-# LANGUAGE TypeFamilies #-}
 {-# LANGUAGE TypeOperators #-}
 {-# LANGUAGE UndecidableInstances #-}
+{-# OPTIONS_GHC -Wno-deprecations #-}
 
 module Cardano.Ledger.Shelley.LedgerState.Types where
 
@@ -44,7 +45,7 @@ import Cardano.Ledger.EpochBoundary
   )
 import Cardano.Ledger.Keys
   ( KeyHash (..),
-    KeyPair,
+    KeyPair, -- deprecated
     KeyRole (..),
   )
 import Cardano.Ledger.PoolDistr (PoolDistr (..))
@@ -73,8 +74,9 @@ import NoThunks.Class (NoThunks (..))
 
 -- ==================================
 
--- | Representation of a list of pairs of key pairs, e.g., pay and stake keys
 type KeyPairs c = [(KeyPair 'Payment c, KeyPair 'Staking c)]
+
+{-# DEPRECATED KeyPairs "Use `Test.Cardano.Ledger.Core.KeyPair (KeyPairs)` instead" #-}
 
 type RewardAccounts c =
   Map (Credential 'Staking c) Coin
--- a/libs/cardano-ledger-core/src/Cardano/Ledger/Keys.hs
+++ b/libs/cardano-ledger-core/src/Cardano/Ledger/Keys.hs
@@ -25,7 +25,7 @@ module Cardano.Ledger.Keys
     -- * DSIGN
     DSignable,
     VKey (..),
-    KeyPair (..),
+    KeyPair (..), -- deprecated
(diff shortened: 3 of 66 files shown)


eras/shelley/impl/src/Cardano/Ledger/Shelley/API/Types.hs
```diff
@@ -50,7 +50,7 @@ import Cardano.Ledger.Keys as X
     Hash,
     KESignable,
     KeyHash (..),
-    KeyPair (..),
+    -- KeyPair (..),
```
JaredCorduan: I think the right thing to do here is to make a new type, also called `KeyPair`, and deprecate it with a helpful message telling people where to find the new one.

libs/cardano-ledger-core/src/Cardano/Ledger/Keys.hs
```diff
@@ -187,23 +187,23 @@ instance
...
+--     NFData (DSIGN.VerKeyDSIGN (DSIGN c)),
+--     NFData (DSIGN.SignKeyDSIGN (DSIGN c))
+--   ) =>
+--   NFData (KeyPair kd c)
 
-instance Crypto c => NoThunks (KeyPair kd c)
+-- instance Crypto c => NoThunks (KeyPair kd c)
 
-instance HasKeyRole KeyPair
+-- instance HasKeyRole KeyPair
```
JaredCorduan: dead code ?

libs/cardano-ledger-pretty/src/Cardano/Ledger/Pretty.hs
```diff
@@ -84,12 +84,12 @@ import Cardano.Ledger.Keys
     GenDelegPair (..),
     GenDelegs (..),
     KeyHash (..),
-    KeyPair (..),
     KeyRole (Staking),
     VKey (..),
     VerKeyKES,
     hashKey,
   )
+-- import Test.Cardano.Ledger.Core.KeyPair (KeyPair (..))
```
JaredCorduan, line 93:
```suggestion
```

JaredCorduan (commented):
I'm very happy to be moving this to the tests, thank you! I had just one comment about deprecating `KeyPair` from the API.

eras/shelley/impl/src/Cardano/Ledger/Shelley/API/Types.hs
teodanciu: Makes sense!

libs/cardano-ledger-core/src/Cardano/Ledger/Keys.hs
teodanciu: eeek

libs/cardano-ledger-core/src/Cardano/Ledger/Keys.hs
```diff
@@ -188,23 +188,8 @@ instance
...
-    NFData (DSIGN.VerKeyDSIGN (DSIGN c)),
-    NFData (DSIGN.SignKeyDSIGN (DSIGN c))
-  ) =>
-  NFData (KeyPair kd c)
-
-instance Crypto c => NoThunks (KeyPair kd c)
-
-instance HasKeyRole KeyPair
+newtype KeyPair c = KeyPair ()
+{-# DEPRECATED KeyPair "Use `Test.Cardano.Ledger.Core.KeyPair (KeyPair)` instead" #-}
```
teodanciu, line 195: @JaredCorduan  - something like this maybe?  I mean leaving it here where the original type was but deprecated, or defining it somewhere else ?

JaredCorduan: yea, exactly, except that I think we want the type to remain the same (ie keep the `KeyRole` and the two fields). The idea is that we want downstream code to still compile, just emit a warning (I don't think that the warning will fire (or be as obvious) if it doesn't type check, but we could test that out).

teodanciu: Yeah fair enough, at first I left it as you say but then I wanted to get rid of the imports. But you're right, if we're gonna deprecate it, it's stupid to change the type and leave the name.

eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState/Types.hs
```diff
@@ -16,6 +16,7 @@
 {-# LANGUAGE TypeFamilies #-}
 {-# LANGUAGE TypeOperators #-}
 {-# LANGUAGE UndecidableInstances #-}
+{-# OPTIONS_GHC -Wno-deprecations #-}
```
teodanciu, line 19: I checked that trying to use deprecated types from this module (eg: KeyPairs), WILL result in deprecation error, so this no-deprecation applies within this file rather than affecting the usages.

JaredCorduan: awesome, good to know, thank you!

libs/cardano-ledger-core/src/Cardano/Ledger/Keys.hs
JaredCorduan: I am still learning too!

JaredCorduan (approved):
looks great, thank you!

lehins (approved):
:+1: 

Awesome. I don't think anyone outside our tests uses the `KeyPair` type, but just in case deprecating is not a bad idea.

PR #3212, opened 2022-12-13
Adjust deposits event

kderme:
# Description

The tx hash is used to associate the event with the tx since the events within a block can appear out of order. The adjustment from deposits to deposits change is useful for db-sync.

# Checklist

- [x] Self-reviewed the diff


lehins (approved):
:+1:

JaredCorduan (approved):
looks good. I think db-sync is the only consumer of the ledger events at the moment, but I think in the future we'll probably want to mention changes to them in the changelog.

kderme:
> looks good. I think db-sync is the only consumer of the ledger events at the moment, but I think in the future we'll probably want to mention changes to them in the changelog.

A changelog entry is now added

PR #3214, opened 2022-12-14
Hand written FromCBOR instance for Coin, disallowing negative Coins

TimSheard:
We really shouldn't deserialize negative coins. So staring with Version 9, we will raise an error

Decoder error: Deserialisation failure while decoding Coin.
CBOR failed with error: DeserialiseFailure 0 "expected word"

As we use (CompactForm Coin) for more things, this makes more sense.


lehins (approved):
:+1:

PR #3217, opened 2022-12-16
Fold the keyDeposit into the UMap

TimSheard:
We change the Rewards View of the UMap so that it map to a Reward-Deposit-Pair, so the single view stores both
the reward amount (a (CompactForm Coin)) for a staking credential, and a deposit amount (a (CompactForm Coin)) 
for the same credential.

Because we have UMap version of deposits, we no longer need the dsDeposits field for DState. It was removed.

The purpose of this PR is to save space, as there are millions of staking credentials, Since we are using compact coins, this should take about 1 extra Word64 per credential for deposits, and 2 less Word64 per credential for rewards (which used to be Coin). So this actually uses less memory.

A few things in the tools  Cardano.Ledger.State.Query  will need to fixed in a follow on PR


CHANGELOG.md
```diff
@@ -59,6 +59,7 @@ in the naming of release branches.
 
 ### Changed
 
+- Changed the representaion of  key deposits. Deleted the field dsDeposits. Moved it to the UMap. Added some new functions to UMapCompact module to accomodate changes.
```
JaredCorduan, line 65: I'm almost always a huge fan of adding to the changelog! I think in this case, though, the details are so internal that no one besides the ledger team would notice (outside of the reduced memory consumption and the structure of the ledger state when dumped from the node). I'd say leave this line out, but happy to hear if others disagree!

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Delegs.hs
```diff
@@ -215,15 +214,16 @@ delegsTransition = do
               (Map.mapKeys (mkRwdAcnt network) (UM.rewView (dsUnified ds)))
           )
 
-      let wdrls_' :: RewardAccounts (EraCrypto era)
-          wdrls_' =
+      let wdrls2 :: Map.Map (Credential 'Staking (EraCrypto era)) UM.RDPair
```
JaredCorduan: maybe `drainedRewardAccounts` would be a better name?

libs/cardano-ledger-core/src/Cardano/Ledger/DPState.hs
```diff
@@ -136,26 +132,24 @@ instance NoThunks (InstantaneousRewards c) => NoThunks (DState c)
 instance NFData (InstantaneousRewards c) => NFData (DState c)
 
 instance (CC.Crypto c, ToCBOR (InstantaneousRewards c)) => ToCBOR (DState c) where
-  toCBOR (DState unified fgs gs ir ds) =
-    encodeListLen 5
+  toCBOR (DState unified fgs gs ir) =
+    encodeListLen 4
```
JaredCorduan, line 136: :pray:

libs/ledger-state/src/Cardano/Ledger/State/Query.hs
```diff
@@ -530,7 +531,8 @@ getDStateNoSharing dstateId = do
       rws <- selectList [RewardDstateId ==. dstateId] []
       forM rws $ \(Entity _ Reward {..}) -> do
         Credential credential <- getJust rewardCredentialId
-        pure (Keys.coerceKeyRole credential, rewardCoin)
+        pure (Keys.coerceKeyRole credential, UM.RDPair (UM.compactCoinOrError rewardCoin) (UM.CompactCoin 0))
+  -- FIXME the deposit is not accounted for ^
```
JaredCorduan, line 535: I realize that you did not introduce this, but I'm realizing now that I do not understand the original FIXME. is this tool not able to properly load the deposits into the database?

JaredCorduan (approved):
Looks great to me, thank you Tim! I made a few comments, but only very minor stuff.

CHANGELOG.md
TimSheard: Although this is internal, it seems a shame not to note it, since it affects so many things.

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Delegs.hs
TimSheard: Good idea

libs/cardano-ledger-core/src/Cardano/Ledger/DPState.hs
TimSheard: Yeah .... , I didn't catch this one at first.

libs/ledger-state/src/Cardano/Ledger/State/Query.hs
TimSheard: Yes, anything that changes the data structures breaks this tool. Alexei promised to fix this. I will remind him. I added this comment.

 -- FIXME the deposit is not accounted for ^
  -- The PR ts-keydeposit-intoUMap breaks this tool since it changes the DPState data type.
  -- https://github.com/input-output-hk/cardano-ledger/pull/3217
  -- All the FIXME-s in this file will have to be addressed in a follow on PR

And Alexey agreed to fix it later.

eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState/IncrementalStake.hs
```diff
@@ -226,22 +226,25 @@ resolveActiveIncrementalPtrs isActive ptrMap_ (IStake credStake ptrStake) =
...
 aggregateActiveStake :: Ord k => Map k (Trip c) -> Map k Coin -> Map k Coin
 aggregateActiveStake =
   Map.mergeWithKey
     -- How to merge the ranges of the two maps where they have a common key. Below
     -- 'coin1' and 'coin2' have the same key, '_k', and the stake is active if the delegation is SJust
-    (\_k trip coin2 -> (<> coin2) <$> (fromCompact <$> UM.tripRewardActiveDelegation trip))
+    (\_k trip coin2 -> extractAndAdd coin2 <$> UM.tripRewardActiveDelegation trip)
     -- what to do when a key appears just in 'tripmap', we only add the coin if the key is active
-    (Map.mapMaybe (\trip -> fromCompact <$> UM.tripRewardActiveDelegation trip))
+    (Map.mapMaybe (\trip -> (fromCompact . UM.rdReward) <$> UM.tripRewardActiveDelegation trip))
```
lehins: Redundant parenthesis
```suggestion
    (Map.mapMaybe (\trip -> fromCompact . UM.rdReward <$> UM.tripRewardActiveDelegation trip))
```

eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState/IncrementalStake.hs
```diff
@@ -226,22 +226,25 @@ resolveActiveIncrementalPtrs isActive ptrMap_ (IStake credStake ptrStake) =
...
     -- what to do when a key is only in 'incremental', keep everything, because at
     -- the call site of aggregateActiveStake, the arg 'incremental' is filtered by
     -- 'resolveActiveIncrementalPtrs' which guarantees that only active stake is included.
     id
 
+extractAndAdd :: Coin -> UM.RDPair -> Coin
+extractAndAdd coin (UM.RDPair rew _dep) = coin <> fromCompact rew
+
```
lehins: This is only used in one place, might as well turn into a local function:
```suggestion
  where
    extractAndAdd :: Coin -> UM.RDPair -> Coin
    extractAndAdd coin (UM.RDPair rew _dep) = coin <> fromCompact rew
```

eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState/RefundsAndDeposits.hs
```diff
@@ -81,17 +88,18 @@ keyCertsRefunds ::
...
+keyCertsRefunds pp dpstate certs = snd (foldl' accum (initialKeys, Coin 0) certs)
   where
-    regkeys = dsDeposits (dpsDState dpstate)
+    initialKeys = (Rewards . dsUnified . dpsDState) dpstate
+    keyDeposit = compactCoinOrError (getField @"_keyDeposit" pp)
     accum (keys, ans) (DCertDeleg (RegKey k)) =
       -- Deposit is added locally to the growing 'keys'
-      (Map.insert k (getField @"_keyDeposit" pp) keys, ans)
+      (Rewards $ UM.insert k (RDPair mempty keyDeposit) keys, ans)
     accum (keys, ans) (DCertDeleg (DeRegKey k)) =
```
lehins: Unrelated to your changes, just an observation. It would be good to make the tuple acc strict: `(!keys, !ans)` in both positions (line 95 and line 98).

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Deleg.hs
```diff
@@ -277,21 +276,22 @@ delegationTransition = do
       -- (hk ∉ dom (rewards ds))
       UM.notMember hk (rewards ds) ?! StakeKeyAlreadyRegisteredDELEG hk
       let u1 = dsUnified ds
-          u2 = Rewards u1 UM.∪ (hk, mempty)
+          deposit = (compactCoinOrError (getField @"_keyDeposit" pp))
```
lehins: Redundant parenthesis
```suggestion
          deposit = compactCoinOrError (getField @"_keyDeposit" pp)
```

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Deleg.hs
```diff
@@ -277,21 +276,22 @@ delegationTransition = do
...
+          u2 = Rewards u1 UM.∪ (hk, RDPair (UM.CompactCoin 0) deposit)
           u3 = Ptrs u2 UM.∪ (ptr, hk)
-      pure (payKeyDeposit hk pp (ds {dsUnified = u3}))
+      pure $ (ds {dsUnified = u3}) -- payKeyDeposit hk pp (ds {dsUnified = u3})
```
lehins: Redundant `$` and some dead code:
```suggestion
      pure (ds {dsUnified = u3})
```

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Deleg.hs
```diff
@@ -277,21 +276,22 @@ delegationTransition = do
...
-      let rewardCoin = UM.lookup hk (rewards ds)
+      let rewardCoin = rdReward <$> UM.lookup hk (rewards ds)
       rewardCoin == Just mempty ?! StakeKeyNonZeroAccountBalanceDELEG (fromCompact <$> rewardCoin)
 
       let u0 = dsUnified ds
           u1 = Set.singleton hk UM.⋪ Rewards u0
           u2 = Set.singleton hk UM.⋪ Delegations u1
           u3 = Ptrs u2 UM.⋫ Set.singleton hk
-          u4 = refundKeyDeposit hk (ds {dsUnified = u3})
+          u4 = (ds {dsUnified = u3}) -- refundKeyDeposit hk (ds {dsUnified = u3})
```
lehins: Same here:
```suggestion
          u4 = ds {dsUnified = u3}
```

libs/cardano-ledger-core/src/Cardano/Ledger/UMapCompact.hs
```diff
@@ -239,7 +274,7 @@ umInvariant stake ptr (UMap tripmap ptrmap) = forwards && backwards
 data View c k v where
   Rewards ::
```
lehins: I think it would make sense to rename this to something like`RewardsDeposits`, Otherwise it is a bit misleading cause there are places were it is used with deposits only while the reward part of the RDPair is mempty.

libs/cardano-ledger-core/src/Cardano/Ledger/UMapCompact.hs
```diff
@@ -98,15 +107,40 @@ import GHC.Stack (HasCallStack)
...
+-- and the deposit for a given (Credential 'Staking c)
+data RDPair = RDPair
+  { rdReward :: {-# UNPACK #-} !(CompactForm Coin)
+  , rdDeposit :: {-# UNPACK #-} !(CompactForm Coin)
+  }
+  deriving (Show, Eq, Ord, Generic, NoThunks, NFData)
+
+rdPair :: Coin -> Coin -> RDPair
+rdPair rew dep = RDPair (compactCoinOrError rew) (compactCoinOrError dep)
+
```
lehins: This function is dangerous and is not used anywhere. There is a similar one redefined in tests, but this one should be removed
```suggestion
```

eras/shelley/test-suite/src/Test/Cardano/Ledger/Shelley/Generator/Trace/Chain.hs
```diff
@@ -239,12 +242,13 @@ registerGenesisStaking
       -- about updating the 'ssDelegations' field.
       --
       -- See STS DELEG for details
+      zeroPair = UM.RDPair (UM.CompactCoin 0) keyDeposit
```
lehins: Why is it called `zeroPair`? `keyDeposit` isn't guaranteed to be zero, right?
Maybe something like `zeroRewardKeyDeposit`?

eras/shelley/test-suite/test/Test/Cardano/Ledger/Shelley/Examples/Combinators.hs
```diff
@@ -290,7 +299,7 @@ newStakeCred cred ptr cs = cs {chainNes = nes'}
       ds
         { dsUnified =
             let um0 = dsUnified ds
-                um1 = (UM.insert cred (UM.CompactCoin 0) (Rewards um0))
+                um1 = (UM.insert cred (UM.RDPair (UM.CompactCoin 0) (UM.CompactCoin 0)) (Rewards um0))
```
lehins: Redundant parenthesis:
```suggestion
                um1 = UM.insert cred (UM.RDPair (UM.CompactCoin 0) (UM.CompactCoin 0)) (Rewards um0)
```

libs/cardano-ledger-binary/testlib/Test/Cardano/Ledger/Binary/TreeDiff.hs
```diff
@@ -100,6 +100,8 @@ expectExprEqual x y = expectExprEqualWithMessage "Expected two values to be equa
 
 expectExprEqualWithMessage :: (ToExpr a, Eq a, HasCallStack) => [Char] -> a -> a -> Expectation
 expectExprEqualWithMessage message expected actual =
-  unless (actual == expected) (expectationFailure msg)
+  unless
+    (actual == expected)
+    (putStrLn msg >> expectationFailure "expectExprEqual fails (expected red, actual green)")
```
lehins: Could you please revert this change:

```suggestion
  unless (actual == expected) (expectationFailure msg)
```

and add and use this function instead with tasty (You'll need to also add `import Test.Tasty.HUnit`)

```haskell
assertExprEqualWithMessage :: (ToExpr a, Eq a, HasCallStack) => [Char] -> a -> a -> Assertion
assertExprEqualWithMessage message expected actual =
  unless (actual == expected) (assertFailure msg)
  where
    msg = (if null message then "" else message ++ "\n") ++ diffExpr expected actual
```

The issue is that `assertFailure` and `expectationFailure` throw different type of exceptions and cannot be mixed, unfortunately. This `assertExprEqualWithMessage` will eventually go away when we fully switch to `hspec`

libs/cardano-ledger-core/src/Cardano/Ledger/UMapCompact.hs
```diff
@@ -536,13 +590,18 @@ view ⨃ mp = unView $ Map.foldlWithKey' accum view mp
...
+  View c k RDPair ->
   Map k (CompactForm Coin) ->
   UMap c
-(Rewards (UMap tm pm)) ∪+ mp = UMap (unionHelp tm mp) pm
+(Rewards (UMap tripmap ptrmap)) ∪+ aggRewMap = UMap (unionHelp tripmap aggRewMap) ptrmap
 
--- The other Views are not reachable, since they do not have (CompactForm Coin) as 3rd parameter
+addPairAndCoin :: StrictMaybe RDPair -> CompactForm Coin -> StrictMaybe RDPair
+addPairAndCoin SNothing _ = SNothing
+addPairAndCoin (SJust (RDPair rew deposit)) delta = SJust (RDPair (addCompact rew delta) deposit)
```
lehins: Maybe call it `addReward` or something else that mentions that addition happens only to the reward, not the deposit

lehins (approved):
This looks great! Few suggestions, some more important than others, but nothing that is blocking this PR.

libs/cardano-ledger-binary/testlib/Test/Cardano/Ledger/Binary/TreeDiff.hs
TimSheard: done

libs/cardano-ledger-core/src/Cardano/Ledger/UMapCompact.hs
TimSheard: I called it  addCoinToJustRewardsPartOfRDPair

eras/shelley/test-suite/src/Test/Cardano/Ledger/Shelley/Generator/Trace/Chain.hs
TimSheard: I renamed it  pairWithDepositsButNoRewards

eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState/RefundsAndDeposits.hs
TimSheard: Good catch, done

libs/cardano-ledger-core/src/Cardano/Ledger/UMapCompact.hs
TimSheard: done

PR #3218, opened 2022-12-16
Cleanup address deserialization

lehins:
# Description

Currently there are 4 separate implementations for deserializing an `Addr`. I am not gonna go into history why this is the case, but this PR moves old implementations into tests and uses them for ensuring that deserialization doesn't break accidentally over time and also as a baseline for checking performance. Summary:

* There is now one single deserialization implementation for both `CompactAddr` and `Addr`
* Constructor for `CompactAddr` is no longer exposed, making it impossible to create invalid and dangerous binary representations for addresses
* Combines `Cardano.Ledger.CompactAddress` into `Cardano.Ledger.Address`, while preserving git history
* There is no longer a need to supply a version for rountrip tests, they will iterate through all versions of serialization, since that property must hold for all versions
* Switch to `fromCBOR` in `BabbageTxBody` and `BabbagetxOut` for address. Now possible thanks to versioned deserialization

# Checklist

- [x] Self-reviewed the diff


libs/cardano-ledger-binary/cardano-ledger-binary.cabal
```diff
@@ -158,5 +158,4 @@ test-suite tests
                    , vector
                    , vector-map
 
-  ghc-options:      -threaded
-                    -rtsopts
+  ghc-options:      -threaded -rtsopts -with-rtsopts=-N
```
teodanciu, line 161: `with-rtsopts=N` is for parallelism?  (Sorry, can't find the specs the opts passed to -with-rtsopts` ) .
Any reason for which  it's set for this particular package? ( I see a few more which have it too).  Is it to do with the serialization?

lehins: Yeah, That's a very good question. What `-N` actually means is use all cores that are available on the machine. It is also possible to specify a concrete number too, eg: `-N4`, which will tell RTS to use  4 capabilities (i.e. OS threads), and if there are at least 4 cores than it will be able to utilize them.

 I don't remember why I added multi-core execution here, might have been working on multiple projects at the same time and mixed em up. That being said, it is possible to parallelize tests and we should probably experiment with it, hspec has this ability out of the box with [`parallel`](https://hackage.haskell.org/package/hspec-2.10.8/docs/Test-Hspec.html#v:parallel). So, I'll leave the `-N` here, since it doesn't hurt and can be useful in the future.

libs/cardano-ledger-core/testlib/Test/Cardano/Ledger/Common.hs
```diff
@@ -0,0 +1,30 @@
...
+      configTimes = True
+    }
+
+ledgerTestMainWith :: Config -> Spec -> IO ()
+ledgerTestMainWith conf spec = do
+  hSetBuffering stdout LineBuffering
+  hSetEncoding stdout utf8
+  hspecWith conf spec
+
+ledgerTestMain :: Spec -> IO ()
```
teodanciu, line 35: I forgot what you said about this, I'm sorry!   Who will it be re-used by, so that it's called Common? 
So far it seems only used from the Main of cardano-ledger-core test suite, but it's generic enough to be used from where else for example?

teodanciu: Oh, judging by [this](https://github.com/input-output-hk/cardano-ledger/pull/3218/commits/0929d563f72d19fa3dfa552110736dcced6935a8#diff-3766dcbff475881d6bfc018ec82386fcb978ac2c3cfe0c46bc144c6a9f581c5aR8) import , this is  meant like a little prelude, grouping useful imports and stuff I guess?

lehins: > Oh, judging by [this](https://github.com/input-output-hk/cardano-ledger/pull/3218/commits/0929d563f72d19fa3dfa552110736dcced6935a8#diff-3766dcbff475881d6bfc018ec82386fcb978ac2c3cfe0c46bc144c6a9f581c5aR8) import , this is meant like a little prelude, grouping useful imports and stuff I guess?

That's exactly right.

We can avoid adding these in every test module:
```haskell
import Test.Hspec
import Test.QuickCheck
...
```

libs/cardano-ledger-core/test/Test/Cardano/Ledger/AddressSpec.hs
```diff
@@ -145,3 +143,17 @@ propDeserializeRewardAcntErrors v acnt = do
...
+    prop "compactAddr/decompactAddr round trip" $
+      propCompactAddrRoundTrip @StandardCrypto
+    prop "Compact address binary representation" $
+      propCompactSerializationAgree @StandardCrypto
+    prop "Ensure Addr failures on incorrect binary data" $
+      propDecompactErrors @StandardCrypto
+    prop "Ensure RewardAcnt failures on incorrect binary data" $
+      propDeserializeRewardAcntErrors @StandardCrypto
+    prop "Decompacting an address is still valid" $
+      propValidateNewDecompact @StandardCrypto
```
teodanciu: Three of these were part of the `minimalPropertyTests`  in addition to the `propertyTests`, so I guess the restructuring in of tests in `cardano-ledger-core` has an impact on what tests the "nightly builds" are running. This is just an observation, because I don't even know if  the nightly builds are running at all, or if it matters. Just noticing the difference.

libs/cardano-ledger-binary/test/Test/Cardano/Ledger/Binary/RoundTripSpec.hs
```diff
@@ -70,201 +70,196 @@ import Test.Hspec
...
-            roundTripSpec
-              @( Hash Blake2b_224 (),
-                 Hash Blake2b_256 (),
-                 Hash SHA256 (),
-                 Hash SHA3_256 (),
-                 Hash Keccak256 (),
-                 Hash ShortHash ()
-               )
-              version
-              cborTrip
```
teodanciu, line 267: Bah, Github got  surprisingly confused with this diff

libs/cardano-ledger-binary/testlib/Test/Cardano/Ledger/Binary/RoundTrip.hs
```diff
@@ -49,15 +49,14 @@ import Test.QuickCheck hiding (label)
 
 -- =====================================================================
 
--- | Tests the roundtrip property using QuickCheck generators
+-- | Tests the roundtrip property using QuickCheck generators fro all possible versions
```
teodanciu: Small typo:  `fro`

libs/cardano-ledger-binary/testlib/Test/Cardano/Ledger/Binary/RoundTrip.hs
```diff
@@ -83,30 +82,29 @@ roundTripFailureExpectation version x =
       expectationFailure $
         "Should not have deserialized: " ++ showExpr (CBORBytes (serialize' version x))
 
--- | Verify that round triping through the binary form holds.
+-- | Verify that round triping through the binary form holds fro all possible versions
```
teodanciu: Same small typo `fro`

libs/cardano-ledger-binary/testlib/Test/Cardano/Ledger/Binary/RoundTrip.hs
```diff
@@ -83,30 +82,29 @@ roundTripFailureExpectation version x =
...
   t ->
   Expectation
-roundTripExpectation version trip t =
-  case roundTrip version trip t of
-    Left err -> expectationFailure $ "Failed to deserialize encoded: " ++ show err
-    Right tDecoded -> tDecoded `shouldBe` t
+roundTripExpectation trip t =
+  forM_ [minBound .. maxBound] $ \version ->
+    case roundTrip version trip t of
+      Left err -> expectationFailure $ "Failed to deserialize encoded: " ++ show err
```
teodanciu, line 102: Does `err` include the version number for which it failed to deserialize?

lehins: Yep, that was the part I showed during the meeting. The roundTrip produces failure result that contains the version, along with other relevant stuff

libs/cardano-ledger-core/src/Cardano/Ledger/Address.hs
```diff
@@ -364,14 +366,10 @@ getByron =
...
 bootstrapAddressAttrsSize :: BootstrapAddress c -> Int
 bootstrapAddressAttrsSize (BootstrapAddress addr) =
-  -- I'm sorry this code is formatted so weridly below.
-  -- It is to apease the capricious god Ormolu. A sacrifice is demanded!
-  maybe
-    0
-    (BS.length . Byron.getHDAddressPayload)
-    (Byron.aaVKDerivationPath (Byron.attrData attrs))
-    + Byron.unknownAttributesLength attrs
+  maybe 0 payloadLen derivationPath + Byron.unknownAttributesLength attrs
```
teodanciu, line 302: hehe nice solution to the formatting problem :laughing:

libs/cardano-ledger-binary/testlib/Test/Cardano/Ledger/Binary/RoundTrip.hs
lehins: Fixed. Thank you.

libs/cardano-ledger-core/test/Test/Cardano/Ledger/AddressSpec.hs
lehins: Moving those tests to hspec will run them regardless of where the test suite is executed, which is what we want. 

I am not sure what was the reasoning behind running some tests on nightly only, but I think we need to move away from that. All tests should run at all times. One reason I can think of is runtime. If some test takes long time, then that test should be fixed. Rather than making it run once a day only.

In any case, this was a good observation, but this minimal vs normal vs nightly distinction is not important for this PR.

libs/cardano-ledger-binary/test/Test/Cardano/Ledger/Binary/RoundTripSpec.hs
lehins: Yeah, I noticed that too :man_shrugging:

libs/cardano-ledger-binary/testlib/Test/Cardano/Ledger/Binary/RoundTrip.hs
lehins: Fixed too.

libs/cardano-ledger-core/src/Cardano/Ledger/Address.hs
lehins: Yeah, I got tired of reading the comment about ormolu gods

libs/cardano-ledger-core/test/Test/Cardano/Ledger/AddressSpec.hs
```diff
@@ -56,6 +57,13 @@ spec = do
     prop "RewardAcnt (fromCborRewardAcnt)" $
       roundTripExpectation @(RewardAcnt StandardCrypto) (mkTrip toCBOR fromCborRewardAcnt)
 
+propSameAsOldDecompactAddr :: forall c. Crypto c => CompactAddr c -> Property
+propSameAsOldDecompactAddr cAddr =
+  (addr === decompactAddrOld @c cAddr)
+    .&&. (addr == decompactAddrOldLazy @c cAddr)
+  where
+    addr = decompactAddr @c cAddr
+
```
teodanciu, line 66: Curious, if we wrote this as an Expectation instead: 
```
  do
    addr `shouldBe` decompactAddrOld @c cAddr
    addr `shouldBe` decompactAddrOldLazy @c cAddr
  where
    addr = decompactAddr @c cAddr
```

Then would we get finer grained information about whether the first or the second  line failed?  That's what I remember from the meeting, at least :thinking:

libs/cardano-ledger-core/test/Test/Cardano/Ledger/AddressSpec.hs
teodanciu: Fair enough

eras/babbage/impl/src/Cardano/Ledger/Babbage/TxBody.hs
```diff
@@ -732,34 +724,19 @@ instance
...
-      bodyFields 18 =
-        field (\x tx -> tx {btbrReferenceInputs = x}) From
-      bodyFields 1 =
-        field
-          (\x tx -> tx {btbrOutputs = x})
-          (D (decodeStrictSeq (decodeSized (fromCborTxOutWithAddr fromCborBothAddr))))
-      bodyFields 16 =
-        ofield
-          (\x tx -> tx {btbrCollateralReturn = x})
-          (D (decodeSized (fromCborTxOutWithAddr fromCborBothAddr)))
```
teodanciu, line 748: `fromCborTxOutWithAddr` doesn't seem used outside `TxOut` anymore, I wonder if it still makes sense to expose it, since it's just the implementation of `FromCBOR`

teodanciu (approved):
Looks good to me.
First two commits could be probably squashed into one, to make  each remaining commit  work correctly, and same with next two commits, but don't think it matters too much, with all the moving around.

libs/cardano-ledger-core/test/Test/Cardano/Ledger/AddressSpec.hs
lehins: You are absolutely right, I'll change the implementation to your suggestion

eras/babbage/impl/src/Cardano/Ledger/Babbage/TxBody.hs
lehins: Very good point. There is no reason for anyone using it, they should use `fromCBOR` instead. I'll hide it

libs/cardano-ledger-core/test/Test/Cardano/Ledger/AddressSpec.hs
lehins: Done

eras/babbage/impl/src/Cardano/Ledger/Babbage/TxBody.hs
lehins: Done

lehins:
> First two commits could be probably squashed into one, to make each remaining commit work correctly, and same with next two commits

I was unable to squash last two commits, since we then loose git history, however squashing the subsequent two worked out fine.

PR #3222, opened 2022-12-27
Switch from `Shelley`/`BabbageTxOut` to a more general `TxOut`

lehins:
# Description

By switching to more general type family `TxOut` from a concrete one, we can reduce number of constraints needed in many places, thus simplifying the code a bit. There are no semantic changes in this PR, just cleanup.

# Checklist

- [x] Self-reviewed the diff


JaredCorduan (approved):
fantastic clean up, thank you!

PR #3224, opened 2022-12-27
Switch ShelleyGenesis parameterization on era in favor of crypto

lehins:
# Description

This PR:

* Switch `ShelleyGenesis era` -> `ShelleyGenesis c`, which means that `PParams` in the type are restricted to `PParams (ShelleyEra c)`, instead of `ShelleyParams era`. This is necessary, because we do in fact want the pparams from the Shelley era, instead of any other era that uses the same type.
* Removes `TranslateEra` instance for `Shelleygenesis`, since it makes no sense to upgrade a type that is not parameterized by an era. It was wrong to add that instance to begin with.
* Improves `CanStartFromGenesis` class, by abstracting the PParam abstract logic with `fromShelleyPParams`, which is the only part that is different  between eras.
* Fixes `ConwayGenesis`, by adding `AlonzoGenesis`, which is needed for the ability to upgrade PParams all the way from ShelleyEra.
* Removed `Cardano.Ledger.Babbage.Genesis` module, since now it is completely empty, because duplicated logic for upgrading PParams was removed.

# Checklist

- [x] Self-reviewed the diff


JaredCorduan (approved):
this looks really nice! let's just check with @bartfrenk before we merge it, to make sure we are not working against their plans to separate the genesis data from the can start data.

eras/alonzo/impl/src/Cardano/Ledger/Alonzo.hs
```diff
@@ -89,7 +89,7 @@ instance (Crypto c, DSignable c (Hash c EraIndependentTxBody)) => API.ApplyBlock
 instance Crypto c => API.CanStartFromGenesis (AlonzoEra c) where
   type AdditionalGenesisConfig (AlonzoEra c) = AlonzoGenesis
 
-  initialState = API.initialStateFromGenesis extendPPWithGenesis
+  fromShelleyPParams ag = translatePParams ag . translateEra' () . translateEra' ()
```
Soupstraw, line 92: Maybe we should rename `translateEra'` to `translateEraV` for `Void` to make it more clear what the function is for?

eras/shelley/impl/src/Cardano/Ledger/Shelley/API/Genesis.hs
```diff
@@ -35,40 +34,49 @@ import Data.Kind (Type)
 import qualified Data.Map.Strict as Map
 
 -- | Indicates that this era may be bootstrapped from 'ShelleyGenesis'.
-class CanStartFromGenesis era where
+class
+  ( EraTxOut era
+  , Default (State (EraRule "PPUP" era))
```
Soupstraw, line 39: Conway won't have a `PPUP` rule, so does that mean we can't start it from genesis?

lehins: Nah, this `Default` constraint will just change to whatever the `Default` of type of this field will be:

https://github.com/input-output-hk/cardano-ledger/blob/9bb89710943cf14cca968551a935f1c7c058cf1c/eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState/Types.hs#L257

Soupstraw: It looks like the `Default` instance is only necessary for the default implementation of `initialState` so maybe it's a better idea to create a `default` signature for that function and add the constraints there instead of adding them to the class signature.

eras/alonzo/impl/src/Cardano/Ledger/Alonzo.hs
lehins: I would not opposite getting rid of another tick. I am not sure about the `V` suffix, since `Void` is gone by that point. Maybe `translateEraTotal`, or something else that is descriptive. Let's do it in a separate PR though. Wanna create a short ticket about it?

eras/shelley/impl/src/Cardano/Ledger/Shelley/API/Genesis.hs
lehins: > maybe it's a better idea to create a default signature for that function

Then you'll have to create a new implementation for that function for no reason for all subsequent eras starting with Conway. More over, this is not the only place where this problem will appear, you'll have to solve in all places where constraint for `State (EraRule "PPUP" era)` appears. Here are just a few examples:

https://github.com/input-output-hk/cardano-ledger/blob/9bb89710943cf14cca968551a935f1c7c058cf1c/eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState/Types.hs#L120-L122

https://github.com/input-output-hk/cardano-ledger/blob/9bb89710943cf14cca968551a935f1c7c058cf1c/eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState/Types.hs#L128-L130

https://github.com/input-output-hk/cardano-ledger/blob/9bb89710943cf14cca968551a935f1c7c058cf1c/eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState/Types.hs#L148-L150

lehins: In fact this constraint comes from here:

https://github.com/input-output-hk/cardano-ledger/blob/9bb89710943cf14cca968551a935f1c7c058cf1c/eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState/Types.hs#L508-L509

So once it is changed in the `Default` instance for `UTxOState`, it will change here as well

Soupstraw: Alright, I'll fix it in my Conway rules PR, since I'm adding a new type family to get rid of these `State (EraRule "PPUP" era)` things.

Soupstraw (approved):
Looks good

PR #3225, opened 2022-12-28
Fourmolize!

lehins:
# Description

Leading commas at last!
Diff friendly imports and exports at last!

# Checklist

- [x] Self-reviewed the diff


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -10,5 +10,5 @@ to the issue.
 - [ ] Commits have useful messages
 - [ ] New tests are added if needed and existing tests are updated
 - [ ] Any changes are noted in the [changelog](https://github.com/input-output-hk/cardano-ledger/blob/master/CHANGELOG.md)
-- [ ] Code is formatted with ormolu (which can be run with `scripts/ormolise.sh`
+- [ ] Code is formatted with [`fourmolu`](https://github.com/fourmolu/fourmolu) (which can be run with `scripts/fourmolize.sh`
```
JaredCorduan, line 13: :metal:

fourmolu.yaml
```diff
@@ -0,0 +1,14 @@
...
+comma-style: leading
+import-export-style: diff-friendly
+indent-wheres: true
+record-brace-space: true
+newlines-between-decls: 1
+haddock-style: single-line
+haddock-style-module:
+let-style: auto
+in-style: right-align
+unicode: never
```
JaredCorduan, line 12: if I've understood correctly from the docs, this means to "**never** complain about unicode".

fourmolu.yaml
```diff
@@ -0,0 +1,14 @@
+indentation: 2
+function-arrows: trailing
+comma-style: leading
```
JaredCorduan, line 3: :cartwheeling:

shell.nix
```diff
@@ -33,8 +33,8 @@ let
     tools = {
       cabal = "3.8.1.0";
       ghcid = "0.8.7";
-      ormolu = "0.5.0.1";
-      haskell-language-server="latest";
+      fourmolu = "0.10.1.0";
```
JaredCorduan, line 36: so was this already in the nix store, or is this related to you changing the nix sources?

JaredCorduan (approved):
hallelujah!

shell.nix
lehins: I updated nix sources in order to get the latest version of fourmolu, cause `0.10.1.0` is a fairly recent release

fourmolu.yaml
lehins: Nah, as far as I understand it, this means to not do the autoconversion to unicode symbols: `->` -> `→`, `::` -> `∷`, etc. 
https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/unicode_syntax.html

In other words when this setting is set to `auto`, then it will detect if `UnicodeSyntax` extension is enabled and will automatically convert to unicode. `never` means do nothing.

PR #3228, opened 2022-12-28
Simplify libsecp256k1 installation

newhoggy:
# Description

Simplify `libsecp256k1` installation.  This allows the installation to be maintain from the one repo: https://github.com/input-output-hk/setup-secp256k1

# Checklist

- [ ] Self-reviewed the diff


lehins (approved):
I am a bit on a fence here. This PR could be made much simpler, because we don't run Github Actions on anything other than `ubuntu-latest`, since ledger codebase is pure Haskell that is very much OS agnostic. However, it doesn't mean we couldn't start running GA on both MacOS and Windows in the future. So, maybe a bit of extra complexity is justified for potential future compatibility.

lehins (dismissed):
I just looked at this PR: https://github.com/input-output-hk/cardano-base/pull/356/files and it looks like it is still possible to use the `haskell/actions` instead of defining manual cabal installer. If that is indeed the case, I'd rather switch it back to haskell/actions

newhoggy:
It used to be the case that `haskell/actions` was the better option but that was when `ghcup` wasn't installed by default on all Github Actions VMs.

Now that `ghcup` is installed by default, I think `input-output-hk/setup-haskell` which is just a thin wrapper around `ghcup` is the better option.

There are some know problems with `haskell/actions`:

1. `haskell/actions` runs `cabal update` during the install, which is bad because that means `haskell/actions` can fail to install if there is something wrong with the `cabal.project` or a network flake that happens with the `cabal update`.  This makes it looks like a problem with installing GHC or cabal, but it is not.
2. `haskell/actions` has an explicit list of cabal and GHC versions it can install and it can be behind what `ghcup` supports.

lehins:
@newhoggy If you say `input-output-hk/setup-haskell` is better then `actions/haskell`  that's great. I am definitely in favor if it works better. I just don't like all the complexity in this PR that uses neither. For example this PR is so much cleaner: https://github.com/input-output-hk/cardano-base/pull/356/files, albeit not fully functional yet :win

newhoggy:
> I just don't like all the complexity in this PR that uses neither.

Oops, I thought I did that already.  Ok, updated.

lehins (approved):
Very nice. Thank you!

PR #3231, opened 2023-01-03
update the LICENCE and NOTICE for 2023

JaredCorduan:
The appendix should have never been in the LICENCE, and the NOTICE needs to be updated to the current year.

Happy new year :fireworks: 

# Description

to the issue.

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -175,28 +175,3 @@
       of your accepting any such warranty or additional liability.
 
    END OF TERMS AND CONDITIONS
-
-   APPENDIX: How to apply the Apache License to your work.
-
-      To apply the Apache License to your work, attach the following
-      boilerplate notice, with the fields enclosed by brackets "[]"
-      replaced with your own identifying information. (Don't include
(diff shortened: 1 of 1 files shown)


NOTICE
```diff
@@ -1,4 +1,4 @@
-Copyright 2019 IOHK
+Copyright 2023 IOHK
```
lehins: Shouldn't it be 2019-2023?

lehins (approved):
Looks good. Just one minor question.

NOTICE
JaredCorduan: thanks for prodding on this, I'm not going to merge this until I get an answer from our legal team.

JaredCorduan: @lehins on second thought (in case this takes a while to get an answer for), I would like to remove the obviously wrong appendix from the `LICENSE` and leave the `NOTICE` untouched. (and follow up with another PR if/when I get an answer.) yea?

lehins: Yeah, that sounds great.

PR #3241, opened 2023-01-09
Disallow decoding a 0-value MultiAsset

### LINKED ISSUE #3240: Zero-Quantity Assets in Block Data
There is an inconsistency in what is allowed in the block data and what is stored in the ledger state for zero quantity assets in multi asset bundles. A quantity of zero in multi asset bundles is allowed in the block data but they are removed when the Cardano node deserializes the data.

This isn't a security issue at the ledger level, but it can cause confusion for end users and developers when other services deserialize block data themselves. While it can, and arguably should, be fixed at the application level, removing this inconsistency at the chain level would make it easier to build on top of the chain and reduces the chances that a developer displays information inconsistent with the ledger state that could easily be exploited. 

Here is an example change to the current CDDL that could be used to remove the zero quantity assets from the block data.
> negInt64 = -9223372036854775808 .. -1
 posInt64 = 1 .. 9223372036854775807
 nonZeroInt64 = negInt64 / posInt64 ; this is the same as the current int64 definition but without zero
 natNum = 1 .. 4294967295 ; non-zero 32bit uint
 mint = multiasset\<nonZeroInt64\>
 value = coin / [coin,multiasset\<natNum\>]

aniketd:
# Description

Closes #3240

# Checklist

- [x] Self-reviewed the diff


eras/conway/test-suite/cddl-files/conway.cddl
```diff
@@ -429,8 +429,15 @@ multiasset<a> = { * policy_id => { * asset_name => a } }
 policy_id = scripthash
 asset_name = bytes .size (0..32)
 
+negInt64 = -9223372036854775808 .. -1
+posInt64 = 1 .. 9223372036854775808
+nonZeroInt64 = negInt64 / posInt64 ; this is the same as the current int64 definition but without zero
+natNum = 1 .. 4294967295 ; non-zero 32bit uint
+
 value = coin / [coin,multiasset<uint>]
```
aniketd: Change to `value = coin / [coin, multiasset<posInt64>]`

eras/conway/test-suite/cddl-files/conway.cddl
```diff
@@ -429,8 +429,15 @@ multiasset<a> = { * policy_id => { * asset_name => a } }
...
-mint = multiasset<int64>
+mint = multiasset<nonZeroInt64>
+
+old_mint = multiasset<int64>
```
aniketd: Remove this.

eras/mary/impl/src/Cardano/Ledger/Mary/Value.hs
```diff
@@ -262,58 +265,85 @@ roundupBytesToWords b = quot (b + wordLength - 1) wordLength
...
+decodeValue =
+  ifDecoderVersionAtLeast
+    (natVersion @9)
+    (decode' conwayCoin)
+    (decode' preConwayCoin)
+  where
+    conwayCoin :: Decoder s (MaryValue c)
+    conwayCoin = inject . word64ToCoin <$> decodeWord64
+    preConwayCoin :: Decoder s (MaryValue c)
+    preConwayCoin = inject . Coin <$> decodeInteger
```
aniketd: `inject . word64ToCoin <$> decodeWord64` here too. But for Conway guard against zero too.

eras/mary/impl/src/Cardano/Ledger/Mary/Value.hs
```diff
@@ -262,58 +265,85 @@ roundupBytesToWords b = quot (b + wordLength - 1) wordLength
...
+-- | MultiAsset can be of two types: Value and Mint.
+-- Value is always a non-negative `uint` (as per the CDDL), and should allow zero values.
+decodeMultiAssetValue ::
+  Crypto c =>
+  Decoder s (MultiAsset c)
+decodeMultiAssetValue =
+  MultiAsset
+    <$> ifDecoderVersionAtLeast
+      (natVersion @9)
+      (decode' decodePositiveInteger)
```
aniketd: Make it `decodePositiveInteger64`.

libs/cardano-ledger-core/src/Cardano/Ledger/Coin.hs
```diff
@@ -65,7 +65,7 @@ instance FromCBOR Coin where
   fromCBOR =
     ifDecoderVersionAtLeast
       (natVersion @9)
-      (Coin . fromIntegral <$> decodeWord64)
+      (word64ToCoin <$> decodeWord64) -- this was always the case, since CDDL was always `coin = uint`. Fixed in Conway.
```
aniketd: This change is not necessary at all. Revert.

eras/shelley-ma/test-suite/test/Test/Cardano/Ledger/Mary/Value.hs
```diff
@@ -354,6 +361,85 @@ compactRoundTrip = forAll gen $ \v ->
...
+          roundTripVersionExpectation version (Coin (0))
+    , testCase "fail decode 0 as Coin for conway" $
+        roundTripFailureExpectation (natVersion @9) (Coin (0))
+    , testCase "decode -1 as Coin for pre-conway" $
+        forM_ [natVersion @2 .. natVersion @8] $ \version ->
+          roundTripVersionExpectation version (Coin (-1))
+    , testCase "fail decode -1 as Coin for conway" $
+        roundTripFailureExpectation (natVersion @9) (Coin (-1))
+    , testCase "decode 1 as Coin for all eras" $
+        roundTripCborExpectation (Coin (1))
```
aniketd: All these test currently pass and I think they are correct.

eras/shelley-ma/test-suite/test/Test/Cardano/Ledger/Mary/Value.hs
```diff
@@ -354,6 +361,85 @@ compactRoundTrip = forAll gen $ \v ->
...
+    "MultiAsset CBOR roundtrips"
+    [ testCase "decode 0 as MultiAsset for pre-conway" $
+        forM_ [natVersion @2 .. natVersion @8] $ \version ->
+          roundTripVersionExpectation version (MultiAsset (Map.singleton (PolicyID (mkScriptHash @C_Crypto 1)) (Map.singleton (AssetName "asset") (0))))
+    , testCase "fail decode 0 as MultiAsset for conway" $
+        roundTripFailureExpectation (natVersion @9) (MultiAsset (Map.singleton (PolicyID (mkScriptHash @C_Crypto 1)) (Map.singleton (AssetName "asset") (0))))
+    , testCase "decode -1 as MultiAsset for all eras" $
+        roundTripCborExpectation (MultiAsset (Map.singleton (PolicyID (mkScriptHash @C_Crypto 1)) (Map.singleton (AssetName "asset") (-1))))
+    , testCase "decode 1 as MultiAsset for all eras" $
+        roundTripCborExpectation (MultiAsset (Map.singleton (PolicyID (mkScriptHash @C_Crypto 1)) (Map.singleton (AssetName "asset") (1))))
```
aniketd: All these tests currently pass and I think they are correct.

eras/shelley-ma/test-suite/test/Test/Cardano/Ledger/Mary/Value.hs
```diff
@@ -354,6 +361,85 @@ compactRoundTrip = forAll gen $ \v ->
...
+  testGroup
+    "Coin CBOR roundtrips"
+    [ testCase "decode 0 as Coin for pre-conway" $
+        forM_ [natVersion @2 .. natVersion @8] $ \version ->
+          roundTripVersionExpectation version (Coin (0))
+    , testCase "fail decode 0 as Coin for conway" $
+        roundTripFailureExpectation (natVersion @9) (Coin (0))
+    , testCase "decode -1 as Coin for pre-conway" $
+        forM_ [natVersion @2 .. natVersion @8] $ \version ->
+          roundTripVersionExpectation version (Coin (-1))
```
lehins: I always recommend switching to property instead of a test case whenever there is a chance for many potential inputs:
```suggestion
    , testProperty "decode -1 as Coin for pre-conway" $ $ \(Negative c) ->
          roundTripRanngeExpectation minBound (natVersion @8) (Coin c)
```

Try to go through all of the test and see what can be generated instead of hard coded

eras/shelley-ma/test-suite/test/Test/Cardano/Ledger/Mary/Value.hs
```diff
@@ -354,6 +361,85 @@ compactRoundTrip = forAll gen $ \v ->
...
+    [ testCase "decode 0 as MaryValue for pre-conway" $
+        forM_ [natVersion @2 .. natVersion @8] $ \version ->
+          roundTripVersionExpectation version (MaryValue 1 (MultiAsset (Map.singleton (PolicyID (mkScriptHash @C_Crypto 1)) (Map.singleton (AssetName "asset") (0)))))
+    , testCase "fail decode 0 as MaryValue for conway" $
+        roundTripFailureExpectation (natVersion @9) (MaryValue 1 (MultiAsset (Map.singleton (PolicyID (mkScriptHash @C_Crypto 1)) (Map.singleton (AssetName "asset") (0)))))
+    , testCase "fail decode -1 as MaryValue for all eras" $
+        forM_ [natVersion @2 .. natVersion @9] $ \version ->
+          roundTripFailureExpectation version (MaryValue 1 (MultiAsset (Map.singleton (PolicyID (mkScriptHash @C_Crypto 1)) (Map.singleton (AssetName "asset") (-1)))))
+    , testCase "decode 1 as MaryValue for all eras" $
+        roundTripCborExpectation (MaryValue 1 (MultiAsset (Map.singleton (PolicyID (mkScriptHash @C_Crypto 1)) (Map.singleton (AssetName "asset") (1)))))
```
aniketd: All these tests currently pass and I think they are correct.

eras/shelley-ma/test-suite/test/Test/Cardano/Ledger/Mary/Value.hs
```diff
@@ -354,6 +361,85 @@ compactRoundTrip = forAll gen $ \v ->
 compactTest :: TestTree
 compactTest = testProperty "fromCompact . toCompact == id" compactRoundTrip
 
+coinCBORTests :: TestTree
+coinCBORTests =
+  testGroup
+    "Coin CBOR roundtrips"
+    [ testCase "decode 0 as Coin for pre-conway" $
+        forM_ [natVersion @2 .. natVersion @8] $ \version ->
```
lehins: There is no nee to supply versions here, `roundTripVersionExpectation` already goes through all versions
```suggestion
        roundTripVersionExpectation minBound (natVersion @8) (Coin 0)
```

eras/shelley-ma/test-suite/test/Test/Cardano/Ledger/Mary/Value.hs
```diff
@@ -354,6 +361,85 @@ compactRoundTrip = forAll gen $ \v ->
...
+          roundTripVersionExpectation version (Coin (0))
```
lehins: Redundant parenthesis
```suggestion
          roundTripVersionExpectation version (Coin 0)
```

eras/shelley-ma/test-suite/test/Test/Cardano/Ledger/Mary/Value.hs
```diff
@@ -354,6 +361,85 @@ compactRoundTrip = forAll gen $ \v ->
...
+      -- , testCase "fail decode 0 as CompactForm MaryValue for conway" $
+      --     roundTripFailureExpectation (natVersion @9) (toCompact (MaryValue 1 (MultiAsset (Map.singleton (PolicyID (mkScriptHash @C_Crypto 1)) (Map.singleton (AssetName "asset") (0)) ))))
+      testCase "decode -1 as CompactForm MaryValue for all eras" $
+        roundTripCborExpectation (toCompact (MaryValue 1 (MultiAsset (Map.singleton (PolicyID (mkScriptHash @C_Crypto 1)) (Map.singleton (AssetName "asset") (-1))))))
+    , -- , testCase "fail decode -1 as CompactForm MaryValue for all eras" $
+      --     forM_ [natVersion @2 .. natVersion @9] $ \version ->
+      --       roundTripFailureExpectation version (toCompact (MaryValue 1 (MultiAsset (Map.singleton (PolicyID (mkScriptHash @C_Crypto 1)) (Map.singleton (AssetName "asset") (-1))))))
+      testCase "decode 1 as CompactForm MaryValue for all eras" $
+        roundTripCborExpectation (toCompact (MaryValue 1 (MultiAsset (Map.singleton (PolicyID (mkScriptHash @C_Crypto 1)) (Map.singleton (AssetName "asset") (1))))))
```
aniketd: All these tests currently pass, but I think they are incorrect. They should pass/fail identically to the ones for `MaryValue`.

libs/cardano-ledger-binary/testlib/Test/Cardano/Ledger/Binary/RoundTrip.hs
```diff
@@ -84,6 +85,13 @@ roundTripFailureExpectation version x =
       expectationFailure $
         "Should not have deserialized: " ++ showExpr (CBORBytes (serialize' version x))
 
+-- Tests that a decoder error does not happen for a specific version
+roundTripVersionExpectation :: forall a. (Eq a, Show a, ToCBOR a, FromCBOR a) => Version -> a -> Expectation
+roundTripVersionExpectation version x =
+  case roundTrip version (cborTrip @a) x of
+    Left err -> expectationFailure $ "Failed to deserialize encoded: " ++ show err
+    Right tDecoded -> tDecoded `shouldBe` x
+
```
lehins: I recommend you do this instead:

```
roundTripExpectation = roundTripRangeExpectation minBound maxBound

-- | Just like `roundTripExpectation`, but allows to supply an inclusive version range,
-- instead of testing all versions
roundTripRangeExpectation ::
  (Show t, Eq t, Typeable t, HasCallStack) =>
  -- | From version
  Version ->
  -- | To version
  Version ->
  Trip t t ->
  t ->
  Expectation
roundTripRangeExpectation fromVersion toVersion trip t =
  forM_ [fromVersion .. toVersion] $ \version ->
     case roundTrip version trip t of
       Left err -> expectationFailure $ "Failed to deserialize encoded: " ++ show err
       Right tDecoded -> tDecoded `shouldBe` t
```

This way whenever you want to test roundtrip with custom version range you can

eras/shelley-ma/test-suite/test/Test/Cardano/Ledger/Mary/Value.hs
lehins: Generate using `Positive` instead of `1`

lehins (changes requested):
Here are my fly by comments on the testing

libs/cardano-ledger-binary/testlib/Test/Cardano/Ledger/Binary/RoundTrip.hs
aniketd: Awesome!

CHANGELOG.md
```diff
@@ -60,7 +60,7 @@ in the naming of release branches.
 
 ### Changed
 
-
+- Disallow decoding a 0-value `MultiAsset`. #3241
```
JaredCorduan:
```suggestion
- Disallow decoding a 0-value `MultiAsset` in the de-serialization. #3241
```

libs/cardano-ledger-core/src/Cardano/Ledger/Coin.hs
```diff
@@ -135,3 +137,10 @@ instance ToExpr DeltaCoin
 
 instance ToExpr (CompactForm Coin) where
   toExpr x = toExpr (fromCompact x)
+
+decodePositiveInteger64 :: Decoder s Integer
+decodePositiveInteger64 = do
+  n <- fromIntegral <$> decodeWord64
+  if n == 0
+    then fail "Expected a positive integer. Got 0 (zero)."
+    else pure n
```
lehins: Since you put it in this module, might as well call it `decodePositiveCoin` and make return `Coin`

```suggestion
decodePositiveCoin :: Decoder s Coin
decodePositiveCoin = do
  n <- decodeWord64
  if n == 0
    then fail "Expected a positive Coin. Got 0 (zero)."
    else pure $ Coin (toInteger n)
```

eras/mary/impl/src/Cardano/Ledger/Mary/Value.hs
```diff
@@ -262,59 +266,79 @@ roundupBytesToWords b = quot (b + wordLength - 1) wordLength
...
+    preConwayCoin = inject . word64ToCoin <$> decodeWord64
+    decode' :: Decoder s (MaryValue c) -> Decoder s (MaryValue c)
+    decode' decodeCoin = do
+      tt <- peekTokenType
+      case tt of
+        TypeUInt -> decodeCoin
+        TypeUInt64 -> decodeCoin
+        TypeListLen -> decodeValuePair decodeInteger
+        TypeListLen64 -> decodeValuePair decodeInteger
+        TypeListLenIndef -> decodeValuePair decodeInteger
```
lehins:
```suggestion
    conwayCoin :: Decoder s Coin
    conwayCoin = decodePositiveCoin
    preConwayCoin :: Decoder s Coin
    preConwayCoin = word64ToCoin <$> decodeWord64
    decode' :: Decoder s Coin -> Decoder s (MaryValue c)
    decode' decodeCoin = do
      tt <- peekTokenType
      case tt of
        TypeUInt -> inject <$> decodeCoin
        TypeUInt64 -> inject <$> decodeCoin
        TypeListLen -> decodeValuePair (unCoin <$> decodeCoin)
        TypeListLen64 -> decodeValuePair (unCoin <$> decodeCoin)
        TypeListLenIndef -> decodeValuePair (unCoin <$> decodeCoin)
```

eras/mary/impl/src/Cardano/Ledger/Mary/Value.hs
```diff
@@ -262,59 +266,79 @@ roundupBytesToWords b = quot (b + wordLength - 1) wordLength
...
-  MultiAsset . prune <$> decodeMap fromCBOR (decodeMap fromCBOR decodeAmount)
+decodeMultiAssetValue =
+  MultiAsset
+    <$> ifDecoderVersionAtLeast
+      (natVersion @9)
+      (decode' decodePositiveInteger64)
+      (decode' decodeNonNegativeInteger)
+  where
+    decode' decoder =
+      decodeMap fromCBOR (decodeMap fromCBOR decoder)
```
lehins: It is actually better to bring back the previous implementation, because it can be reused in both places (MaryValue and MultiAsset):
```suggestion
decodeMultiAssetMaps decodeAmount =
  MultiAsset . prune <$> decodeMap fromCBOR (decodeMap fromCBOR decodeAmount)
```

eras/mary/impl/src/Cardano/Ledger/Mary/Value.hs
```diff
@@ -262,59 +266,79 @@ roundupBytesToWords b = quot (b + wordLength - 1) wordLength
...
+  Decoder s (MultiAsset c)
+decodeMultiAssetMint =
+  MultiAsset
+    <$> ifDecoderVersionAtLeast
+      (natVersion @9)
+      (decode' True)
+      (decode' False)
+  where
+    decode' nonZero =
+      decodeMap fromCBOR (decodeMap fromCBOR (decodeIntegerBounded64 nonZero))
```
lehins:
```suggestion
 ifDecoderVersionAtLeast
      (natVersion @9)
      (decodeMultiAssetMaps (decodeIntegerBounded64 True))
      (decodeMultiAssetMaps (decodeIntegerBounded64 False))
```

eras/mary/impl/src/Cardano/Ledger/Mary/Value.hs
```diff
@@ -262,59 +266,79 @@ roundupBytesToWords b = quot (b + wordLength - 1) wordLength
...
   (forall t. Decoder t Integer) ->
   Decoder s (MaryValue c)
-decodeValuePair decodeAmount =
+decodeValuePair decoder =
   decode $
     RecD MaryValue
-      <! D decodeAmount
-      <! D (decodeMultiAssetMaps decodeAmount)
+      <! D decoder
+      <! D decodeMultiAssetValue
```
lehins: This `decodeValuePair` function does not need to change either

eras/shelley-ma/test-suite/src/Test/Cardano/Ledger/ShelleyMA/Serialisation/Roundtrip.hs
```diff
@@ -60,4 +88,129 @@ allEraRoundtripTests =
...
     , eraRoundTripProps @Mary
+    , valueCBORTests
+    ]
+
+coinCBORTests :: TestTree
+coinCBORTests =
+  testGroup
+    "Coin CBOR roundtrips"
+    [ testCase "decode 0 as Coin for pre-conway" $
+        roundTripCborRangeExpectation minBound (natVersion @8) (Coin (0))
```
lehins: What's the point of wrapping literals with parenthesis?

```suggestion
        roundTripCborRangeExpectation minBound (natVersion @8) (Coin 0)
```

eras/shelley-ma/test-suite/src/Test/Cardano/Ledger/ShelleyMA/Serialisation/Roundtrip.hs
```diff
@@ -60,4 +88,129 @@ allEraRoundtripTests =
...
+    , testCase "fail decode 0 as Coin for conway" $
+        roundTripFailureExpectation (natVersion @9) (Coin (0))
```
lehins: Future proof
```suggestion
        roundTripRangeFailureExpectation (natVersion @9) maxBound (Coin 0)
```

Will require addition of `roundTripRangeFailureExpectation`

lehins (changes requested):
Bunch of suggestions

eras/mary/impl/src/Cardano/Ledger/Mary/Value.hs
```diff
@@ -262,59 +266,79 @@ roundupBytesToWords b = quot (b + wordLength - 1) wordLength
...
 
 encodeMultiAssetMaps ::
   Crypto c =>
   MultiAsset c ->
   Encoding
 encodeMultiAssetMaps (MultiAsset m) = toCBOR m
 
-decodeMultiAssetMaps ::
+-- | MultiAsset can be of two types: Value and Mint.
```
lehins: There aren't any Haskell type `Mint`, so this comment can be a bit confusing, maybe instead
```suggestion
-- | `MultiAsset` can be used in two different circumstances: during minting and in `MaryValue` while sending.
-- In latter case MultiAsset amounts cannot be negative, while in minting negative indicates burning.
```

eras/shelley-ma/test-suite/src/Test/Cardano/Ledger/ShelleyMA/Serialisation/Roundtrip.hs
```diff
@@ -1,4 +1,8 @@
 {-# LANGUAGE AllowAmbiguousTypes #-}
+{-# LANGUAGE StandaloneDeriving #-}
+{-# LANGUAGE DeriveAnyClass #-}
+{-# LANGUAGE ViewPatterns #-}
+{-# LANGUAGE OverloadedStrings #-}
```
lehins: What are these for? There are no instances defined here?
```suggestion
```

eras/shelley-ma/test-suite/src/Test/Cardano/Ledger/ShelleyMA/Serialisation/Roundtrip.hs
```diff
@@ -60,4 +88,129 @@ allEraRoundtripTests =
...
+
+coinCBORTests :: TestTree
+coinCBORTests =
+  testGroup
+    "Coin CBOR roundtrips"
+    [ testCase "decode 0 as Coin for pre-conway" $
+        roundTripCborRangeExpectation minBound (natVersion @8) (Coin (0))
+    , testCase "fail decode 0 as Coin for conway" $
+        roundTripFailureExpectation (natVersion @9) (Coin (0))
+    , testProperty "decode (Negative Int) as Coin for pre-conway" $ \(getNegative -> i) ->
```
lehins: Why rely on extension here?
```suggestion
    , testProperty "decode (Negative Int) as Coin for pre-conway" $ \(Negative i) ->
```

eras/shelley-ma/test-suite/src/Test/Cardano/Ledger/ShelleyMA/Serialisation/Roundtrip.hs
```diff
@@ -60,4 +88,129 @@ allEraRoundtripTests =
...
+    , testCase "fail decode 0 as MaryValue for conway" $
+        roundTripFailureExpectation (natVersion @9) 
+          (MaryValue 1 (MultiAsset (Map.singleton (PolicyID (mkScriptHash @C_Crypto 1)) (Map.singleton (AssetName "asset") 0))))
+    , testProperty "fail decode (Negative Int) as MaryValue for all eras" $ \(getNegative -> i) ->
+        forM_ [minBound .. maxBound] $ \version ->
+          roundTripFailureExpectation version 
+            (MaryValue 1 (MultiAsset (Map.singleton (PolicyID (mkScriptHash @C_Crypto 1)) (Map.singleton (AssetName "asset") i))))
+    , testProperty "decode (Positive Int) as MaryValue for all eras" $ \(getPositive -> i) ->
+        roundTripCborExpectation 
+          (MaryValue 1 (MultiAsset (Map.singleton (PolicyID (mkScriptHash @C_Crypto 1)) (Map.singleton (AssetName "asset") i))))
```
lehins: I think it would be better to create a generator for MultiAsset, that could be used in the definition of the Arbitrary instance for MaryValue as well

```suggestion
    , testProperty "decode (Positive Int) as MaryValue for all eras" $ \ (Positive c) ->
      forAll $ genPositiveMultiAsset $ \ma ->  roundTripCborExpectation (MaryValue c ma)
```
Here is an idea for a potential generator  that does the right thing:

```haskell
genMap genKey genVal = ...

genMultiAsset genAmount = MultiAsset <$> genMap arbitrary (genMap arbitrary genAmount)

genMultiAssetMint = genMultiAsset (getPositive <$> arbitrary)
```
```

File: eras/conway/test-suite/cddl-files/conway.cddl [hunk 26]
```diff
@@ -429,8 +429,14 @@ multiasset<a> = { * policy_id => { * asset_name => a } }
 policy_id = scripthash
 asset_name = bytes .size (0..32)
 
-value = coin / [coin,multiasset<uint>]
-mint = multiasset<int64>
+negInt64 = -9223372036854775808 .. -1
+posInt64 = 1 .. 9223372036854775808
```
yHSJ: I made a typo in my recommended changes to the CDDL in https://github.com/input-output-hk/cardano-ledger/issues/3240. This would cause an overflow, as the maximum value of a 64-bit integer is `9223372036854775807` 

The corrected definition is:
> posInt64 = 1 .. 9223372036854775807

eras/conway/test-suite/cddl-files/conway.cddl
```diff
@@ -429,8 +429,14 @@ multiasset<a> = { * policy_id => { * asset_name => a } }
...
+nonZeroInt64 = negInt64 / posInt64 ; this is the same as the current int64 definition but without zero
+natNum = 1 .. 4294967295 ; non-zero 32bit uint
+
+value = coin / [coin,multiasset<posInt64>]
```
yHSJ: In my original recommended changes to the CDDL in https://github.com/input-output-hk/cardano-ledger/issues/3240, I had a typo in my markdown that hid the type of the `multiasset` in the `value` definition. As a result, these changes are much more drastic than they were intended to be. 

This is the corrected definition:
> value = coin / [coin,multiasset\< natNum\>]

eras/conway/test-suite/cddl-files/conway.cddl
JaredCorduan: which maybe would have lead us to continue to allow zero :laughing: 

thank you for the heads up!

lehins: Nah it would have just wrapped around into a `-9223372036854775808`. So instead of minting it would be trying to burn :smile:

eras/mary/impl/src/Cardano/Ledger/Mary/Value.hs
```diff
@@ -334,16 +358,12 @@ instance
...
-  where
-  decodeNonNegative = decodeNonNegativeValue
-
 -- Note: we do not use `decodeInt64` from the cborg library here because the
 -- implementation contains "-- TODO FIXME: overflow"
-decodeIntegerBounded64 :: Decoder s Integer
-decodeIntegerBounded64 = do
+-- Also, the caller may control whether to allow mint values of 0 (zero),
+-- as a necessary distinction between pre-Conway and other eras
+decodeIntegerBounded64 :: Bool -> Decoder s Integer
```
JaredCorduan: @lehins I'm curious about your opinion here. Do you think that in order to minimize "boolean blindness" here, it would be helpful to introduce a type `data AllowZeros = AllowZero | DoNotAllowZero`?

eras/shelley-ma/test-suite/test/Test/Cardano/Ledger/Mary/Value.hs
```diff
@@ -368,4 +453,5 @@ valTests =
     , monoValueTests
     , valueGroupTests
     , compactTest
+    -- , valueCBORTests
```
JaredCorduan: are these all commented out since they are failing?

eras/mary/impl/src/Cardano/Ledger/Mary/Value.hs
lehins: Nah, it is an internal function that is used in two places within the same module, there is little benefit in doing this. If this function was exported and used in other places that would definitely be the right way to go.

JaredCorduan: for myself, I always worry I'll mess this kind of thing up. or that it'll spread later. but for now, indeed it's more quarantined and correct.

JaredCorduan (approved):
looks great to me (other than the commented out tests)

eras/conway/test-suite/cddl-files/conway.cddl
aniketd: Not `posInt64` but `natNum`.

eras/shelley-ma/test-suite/test/Test/Cardano/Ledger/Mary/Value.hs
aniketd: These just cause `heap exhausted` errors now.

eras/shelley-ma/test-suite/test/Test/Cardano/Ledger/Mary/Value.hs
aniketd: These are moved out to another module. So I have removed these lines from this file in the latest update.

eras/mary/impl/src/Cardano/Ledger/Mary/Value.hs
```diff
@@ -292,28 +306,20 @@ encodeMultiAssetMaps ::
...
-  Decoder s Integer ->
-  Decoder s (MultiAsset c)
+-- | `MultiAsset` can be used in two different circumstances:
+-- during minting and in `MaryValue` while sending.
+-- In the latter case, MultiAsset amounts cannot be negative,
+-- while in minting negative indicates burning, and should not be zero.
+decodeMultiAssetMaps :: Crypto c => (forall t. Decoder t Integer) -> Decoder s (MultiAsset c)
 decodeMultiAssetMaps decodeAmount =
-  MultiAsset . prune <$> decodeMap fromCBOR (decodeMap fromCBOR decodeAmount)
+  MultiAsset <$> decodeMap fromCBOR (decodeMap fromCBOR decodeAmount)
```
aniketd: @lehins if `prune` is retained here, the tests throw `No Decoder error` `No Dropper error`. This is most likely due to cases where there are `0` or `[]`.

eras/shelley-ma/test-suite/src/Test/Cardano/Ledger/ShelleyMA/Serialisation/Roundtrip.hs
```diff
@@ -60,4 +97,118 @@ allEraRoundtripTests =
...
+    ]
+
+valueCBORTests :: TestTree
+valueCBORTests =
+  testGroup
+    "Value CBOR roundtrips"
+    [ coinCBORTests
+    , multiassetCBORTests
+    , maryvalueCBORTests
+    -- , compactvalueCBORTests
```
aniketd: @lehins These tests are commented out because no matter how `CompactValue` is generated: via `toCompact (MultiAsset c)` or `genCompactValue`, they continue to result in `Heap exhausted`.

libs/cardano-ledger-core/src/Cardano/Ledger/Coin.hs
```diff
@@ -65,7 +67,7 @@ instance FromCBOR Coin where
   fromCBOR =
     ifDecoderVersionAtLeast
       (natVersion @9)
-      (Coin . fromIntegral <$> decodeWord64)
+      decodePositiveCoin
```
lehins: This is wrong, please revert. Coins can actually be zero, because they are used in other places, not only in outputs 
```suggestion
      (Coin . fromIntegral <$> decodeWord64)
```

eras/mary/impl/src/Cardano/Ledger/Mary/Value.hs
lehins: Yeah, that's right, we don't want to prune zero values, we want to fail on those. However that brings up another point. We were before pruning empty maps too, so we need to preserve this behavior, but also I think we need to fail whenever empty map is supplied starting with Conway.

@JaredCorduan What's your opinion here. Do we continue to prune empty maps or fail when one is supplied? I lean towards the latter

eras/shelley-ma/test-suite/src/Test/Cardano/Ledger/ShelleyMA/Serialisation/Roundtrip.hs
lehins: Does each and every one of the four individually causes heap exhaustion?

eras/shelley-ma/test-suite/src/Test/Cardano/Ledger/ShelleyMA/Serialisation/Roundtrip.hs
```diff
@@ -60,4 +89,118 @@ allEraRoundtripTests =
...
+            roundTripFailureExpectation version cv
+    ]
+
+valueCBORTests :: TestTree
+valueCBORTests =
+  testGroup
+    "Value CBOR roundtrips"
+    [ coinCBORTests
+    , multiassetCBORTests
+    , maryvalueCBORTests
```
lehins: Camel case
```suggestion
    , multiAssetCBORTests
    , maryValueCBORTests
```

eras/shelley-ma/test-suite/src/Test/Cardano/Ledger/ShelleyMA/Serialisation/Roundtrip.hs
```diff
@@ -60,4 +89,118 @@ allEraRoundtripTests =
...
+  -- elements [CompactValueAdaOnly (CompactCoin i), CompactValueMultiAsset (CompactCoin i) (1 :: Word32) "a"]
+  elements [CompactValueAdaOnly (CompactCoin i), CompactValueMultiAsset (CompactCoin i) w32 sbs]
+
+compactvalueCBORTests :: TestTree
+compactvalueCBORTests =
+  testGroup
+    "CompactValue CBOR roundtrips"
+    [ testProperty "decode 0 as (CompactValue c = CompactCoin) for pre-conway" $
+        forAll (genCompactValue (pure 0)) $ \cv ->
+          roundTripCborRangeExpectation minBound (natVersion @8) cv
```
lehins: Eta reduce:
```suggestion
        forAll (genCompactValue (pure 0)) $ roundTripCborRangeExpectation minBound (natVersion @8) 
```
Same with tests below

eras/shelley-ma/test-suite/src/Test/Cardano/Ledger/ShelleyMA/Serialisation/Roundtrip.hs
```diff
@@ -60,4 +89,118 @@ allEraRoundtripTests =
...
+          \ma -> roundTripFailureExpectation (natVersion @9) ma
+    ]
+
+maryvalueCBORTests :: TestTree
+maryvalueCBORTests =
+  testGroup
+    "MaryValue CBOR roundtrips"
+    [ testProperty "decode (Positive Int) as MaryValue for all eras" $
+        forAll genMaryValuePositive $
+          \mv -> roundTripCborExpectation mv
```
lehins: Same here. There is no need for lambdas
```suggestion
        forAll genMaryValuePositive $ roundTripCborExpectation
```

lehins (changes requested):
Yeah, it is looking very nice. We need to get all the tests working of course

eras/babbage/impl/src/Cardano/Ledger/Babbage/TxOut.hs
```diff
@@ -422,23 +419,23 @@ instance
...
   , FromCBOR (Annotator (Script era))
-  , DecodeNonNegative (Value era)
+  , FromCBOR (Value era)
   ) =>
   FromSharedCBOR (BabbageTxOut era)
   where
   type Share (BabbageTxOut era) = Interns (Credential 'Staking (EraCrypto era))
   fromSharedCBOR credsInterns =
-    internTxOut <$!> decodeBabbageTxOut
+    internTxOut <$!> fromCBOR
```
lehins: Why this change?
```suggestion
    internTxOut <$!> decodeBabbageTxOut
```

eras/mary/impl/src/Cardano/Ledger/Mary/Value.hs
JaredCorduan: I don't have a strong opinion about this. If we do want to disallow empty maps, let's update the CDDL spec to use `+` instead of `*`.  so I think we'd want `multiasset<a> = { + policy_id => { + asset_name => a } }` (ie no empty map of policies, and no empty map of assets within a given policy.

eras/shelley-ma/test-suite/src/Test/Cardano/Ledger/ShelleyMA/Serialisation/Roundtrip.hs
aniketd: I have updated the code with commented out examples and `NOTE`s (inline) accompanying each case separately, in the hope that it can add more information for debugging.

libs/cardano-ledger-core/src/Cardano/Ledger/Coin.hs
aniketd: Got you! Thanks!

eras/babbage/impl/src/Cardano/Ledger/Babbage/TxOut.hs
aniketd: Oh no! This is from the `rebase`. Sorry! I hadn't reviewed all the changes yet. I'm doing so now and keeping this ready for another review.

eras/shelley-ma/test-suite/src/Test/Cardano/Ledger/ShelleyMA/Serialisation/Roundtrip.hs
```diff
@@ -2,31 +2,66 @@
 {-# LANGUAGE DataKinds #-}
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE MonoLocalBinds #-}
+{-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE TypeApplications #-}
 
 module Test.Cardano.Ledger.ShelleyMA.Serialisation.Roundtrip where
 
+-- import Cardano.Ledger.Compactible (Compactible(toCompact))
```
aniketd: There is some code commented out in this file in order to provide sufficient context for inspecting test failures.

aniketd:
This is now again ready for review :slightly_smiling_face:

eras/shelley-ma/test-suite/src/Test/Cardano/Ledger/ShelleyMA/Serialisation/Roundtrip.hs
aniketd: Since `CompactValue` is never directly codec'ed (or (de)serialised) directly, but only via the "non-compact" form, these tests have been omitted. @lehins please correct me if I'm wrong.

eras/mary/impl/src/Cardano/Ledger/Mary/Value.hs
lehins: If we are not pruning,  we need to guard against them
```suggestion
    ma <- decodeMap fromCBOR (decodeMap fromCBOR decodeAmount)
    ifDecoderVersionAtLeast
    (natVersion @9)
    (MultiAsset ma <$ mapM_ ma (\m -> when (Map.empty m) $ fail "Empty Assets are not allowed")
    (pure $ MultiAsset $ prune ma)
```

lehins (changes requested):
Decoding of multi asset allowed empty maps. Besides fixing it, we also need a test that ensures that checks it.

eras/mary/impl/src/Cardano/Ledger/Mary/Value.hs
lehins: Note, that unlike the spec we do allow multiasset to be empty, because otherwise we'd have to make `mint` field and mulit asset part of the MaryValue be `Maybe MultiAsset`, instead we treating empty as `Nothing`

JaredCorduan: yep, great point @lehins. fortunately, I do not have any evidence that this will cause any confusion for folks following the chain making guesses about the semantics.

lehins (approved):
Beautiful

eras/alonzo/test-suite/test/Test/Cardano/Ledger/Alonzo/Serialisation/Tripping.hs
```diff
@@ -59,11 +61,11 @@ tests =
...
           roundTripAnnTwiddledProperty @(TxBody Alonzo) (zipMemoRawType (===))
     , testProperty "alonzo/CostModels" $
-        roundTripCborExpectation @CostModels
+        roundTripCborRangeExpectation @CostModels (natVersion @2) maxBound
     , testProperty "alonzo/PParams" $
-        roundTripCborExpectation @(PParams Alonzo)
+        roundTripCborRangeExpectation @(PParams Alonzo) (natVersion @2) maxBound
     , testProperty "alonzo/PParamsUpdate" $
-        roundTripCborExpectation @(PParamsUpdate Alonzo)
+        roundTripCborRangeExpectation @(PParamsUpdate Alonzo) (natVersion @2) maxBound
```
lehins: We only need to test for Alonzo
```suggestion
        roundTripCborRangeExpectation @(PParamsUpdate Alonzo) (eraProtVerLow @Alonzo) (eraProtVerHigh @Alonzo)

```

eras/alonzo/test-suite/test/Test/Cardano/Ledger/Alonzo/Serialisation/Tripping.hs
```diff
@@ -59,11 +61,11 @@ tests =
         testProperty "alonzo/TxBody twiddled" $
           roundTripAnnTwiddledProperty @(TxBody Alonzo) (zipMemoRawType (===))
     , testProperty "alonzo/CostModels" $
-        roundTripCborExpectation @CostModels
+        roundTripCborRangeExpectation @CostModels (natVersion @2) maxBound
     , testProperty "alonzo/PParams" $
-        roundTripCborExpectation @(PParams Alonzo)
+        roundTripCborRangeExpectation @(PParams Alonzo) (natVersion @2) maxBound
```
lehins:
```suggestion
        roundTripCborRangeExpectation @(PParams Alonzo) (eraProtVerLow @Alonzo) (eraProtVerHigh @Alonzo)

```

eras/babbage/test-suite/test/Test/Cardano/Ledger/Babbage/Serialisation/Tripping.hs
```diff
@@ -45,5 +47,5 @@ tests =
     , testProperty "babbage/BabbageUtxoPredFailure" $
         roundTripCborExpectation @(BabbageUtxoPredFailure Babbage)
     , testProperty "babbage/Block" $
-        roundTripAnnExpectation @(Block (BHeader StandardCrypto) Babbage)
+        roundTripAnnRangeExpectation @(Block (BHeader StandardCrypto) Babbage) (eraProtVerLow @Babbage) (eraProtVerHigh @Babbage)
```
lehins: Line is too long
```suggestion
        roundTripAnnRangeExpectation @(Block (BHeader StandardCrypto) Babbage) (eraProtVerLow @Babbage) (eraProtVerHigh @Babbage)
```

eras/babbage/test-suite/test/Test/Cardano/Ledger/Babbage/Serialisation/Tripping.hs
```diff
@@ -31,11 +33,11 @@ tests =
...
         roundTripAnnExpectation @(TxBody Babbage)
     , testProperty "babbage/CostModel" $
-        roundTripCborExpectation @CostModels
+        roundTripCborRangeExpectation @CostModels (natVersion @2) maxBound
     , testProperty "babbage/PParams" $
-        roundTripCborExpectation @(PParams Babbage)
+        roundTripCborRangeExpectation @(PParams Babbage) (natVersion @2) maxBound
     , testProperty "babbage/PParamsUpdate" $
-        roundTripCborExpectation @(PParamsUpdate Babbage)
+        roundTripCborRangeExpectation @(PParamsUpdate Babbage) (natVersion @2) maxBound
```
lehins:
```suggestion
        roundTripCborRangeExpectation @(PParamsUpdate Babbage) 
          (eraProtVerLow @Babbage) (eraProtVerHigh @Babbage)
```

eras/babbage/test-suite/test/Test/Cardano/Ledger/Babbage/Serialisation/Tripping.hs
```diff
@@ -31,11 +33,11 @@ tests =
     , testProperty "babbage/TxBody" $
         roundTripAnnExpectation @(TxBody Babbage)
     , testProperty "babbage/CostModel" $
-        roundTripCborExpectation @CostModels
+        roundTripCborRangeExpectation @CostModels (natVersion @2) maxBound
     , testProperty "babbage/PParams" $
-        roundTripCborExpectation @(PParams Babbage)
+        roundTripCborRangeExpectation @(PParams Babbage) (natVersion @2) maxBound
```
lehins:
```suggestion
        roundTripCborRangeExpectation @(PParams Babbage) 
          (eraProtVerLow @Babbage) (eraProtVerHigh @Babbage)
```

eras/babbage/impl/src/Cardano/Ledger/Babbage/TxOut.hs
```diff
@@ -526,7 +523,7 @@ decodeTxOut decAddr = do
     bodyFields 1 =
       field
         (\x txo -> txo {decodingTxOutVal = x})
-        (D decodeNonNegative)
+        (D fromCBOR)
```
lehins:
```suggestion
        From
```

eras/alonzo/test-suite/test/Test/Cardano/Ledger/Alonzo/Serialisation/Tripping.hs
```diff
@@ -77,7 +79,7 @@ tests =
     , testProperty "alonzo/Tx" $
         roundTripAnnExpectation @(Tx Alonzo)
     , testProperty "alonzo/Block" $
-        roundTripAnnExpectation @(Block (BHeader StandardCrypto) Alonzo)
+        roundTripAnnRangeExpectation @(Block (BHeader StandardCrypto) Alonzo) (eraProtVerLow @Alonzo) (eraProtVerHigh @Alonzo)
```
lehins: Line is too long
```suggestion
        roundTripAnnRangeExpectation @(Block (BHeader StandardCrypto) Alonzo) 
          (eraProtVerLow @Alonzo) (eraProtVerHigh @Alonzo)
```

lehins (changes requested):
This is all good, however there is still a problem with OOM. We need to get to the bottom of it. It is too dangerous to merge a PR that has flaky tests like that. For all we know it could be the decoder that is doing something bad, although it is pretty unlikely.

eras/shelley/test-suite/src/Test/Cardano/Ledger/Shelley/Serialisation/EraIndepGenerators.hs
```diff
@@ -334,7 +334,10 @@ genBlock ::
   , h ~ BHeader (EraCrypto era)
   ) =>
   Gen (Block h era)
-genBlock = Block <$> arbitrary <*> (toTxSeq @era <$> arbitrary)
+genBlock =
+  Block
+    <$> (scale (`div` 15) arbitrary)
+    <*> (toTxSeq @era <$> scale (`div` 15) arbitrary)
```
lehins: This is wrong. You don't want to scale down *ALL* of the fields in the Block. So you need to revert this change.
```suggestion
genBlock = Block <$> arbitrary <*> (toTxSeq @era <$> arbitrary)
```

eras/shelley-ma/test-suite/test/Test/Cardano/Ledger/ShelleyMA/Serialisation/Golden/Encoding.hs
```diff
@@ -330,22 +330,6 @@ goldenEncodingTestsMary =
...
-                . TkInteger 0
-                . TkMapLen 1
-            )
-            <> S policyID1
-            <> T
-              ( TkMapLen 1
-                  . TkBytes (SBS.fromShort assetName1)
-                  . TkInteger (-19)
-              )
-        )
```
lehins: Why was this test removed? I can't see anything wrong with it.

eras/shelley-ma/test-suite/test/Test/Cardano/Ledger/ShelleyMA/Serialisation/Golden/Encoding.hs
```diff
@@ -427,7 +411,7 @@ goldenEncodingTestsMary =
 
 assetName32Bytes :: Assertion
 assetName32Bytes =
-  roundTripFailureExpectation (eraProtVerHigh @Mary) $
+  roundTripFailureCborRangeExpectation (eraProtVerHigh @Mary) (eraProtVerHigh @Mary) $
```
lehins: This should work for all eras, starting with `Mary`
```suggestion
  roundTripFailureCborRangeExpectation (eraProtVerHigh @Mary) maxBound $
```

eras/mary/impl/cardano-ledger-mary.cabal
```diff
@@ -63,6 +63,26 @@ library
         cardano-strict-containers,
         text
 
+library testlib
+    exposed-modules: Test.Cardano.Ledger.Mary.ValueSpec
```
lehins, line 67: This is wrong! We should not be using the same module in different cabal targets. If you want to reuse generators that where defined in `Test.Cardano.Ledger.Mary.ValueSpec` then those `Arbitrary` instances should be moved into a new module `Test.Cardano.Ledger.Mary.Arbitrary` and then they can be reused in different test suites. In other words, that's exactly what you are doing in #3285 

That being said, if you are planning to fix this in #3285 then we can merge this as a temperately solution

eras/mary/impl/src/Cardano/Ledger/Mary/Value.hs
```diff
@@ -350,8 +358,11 @@ decodeIntegerBounded64 = do
     TypeNInt64 -> pure ()
     _ -> fail "expected major type 0 or 1 when decoding mint field"
   x <- decodeInteger
-  if x >= minval && x <= maxval
-    then pure x
+  if minval <= x && x <= maxval
+    then
+      if x == 0 && nonZero
+        then fail "mint field cannot be 0 (zero)"
+        else pure x
```
lehins: Possibly inline the if `ifDecoderVersionAtLeast` here instead of `decodeMultiAssetMint`

eras/babbage/test-suite/src/Test/Cardano/Ledger/Babbage/Serialisation/Generators.hs
```diff
@@ -68,16 +68,16 @@ instance
       <$> arbitrary
       <*> arbitrary
       <*> arbitrary
-      <*> arbitrary
-      <*> arbitrary
+      <*> scale (`div` 15) arbitrary
+      <*> scale (`div` 15) arbitrary
```
lehins: You already adjusted the `TxOut`, so no need to scale it again. This effectively would resize the value to an empty map all the time in the TxOut
```suggestion
      <*> arbitrary
      <*> arbitrary
```

eras/alonzo/test-suite/src/Test/Cardano/Ledger/Alonzo/Serialisation/Generators.hs
```diff
@@ -187,14 +187,14 @@ instance
     AlonzoTxBody
       <$> arbitrary
       <*> arbitrary
-      <*> arbitrary
+      <*> scale (`div` 15) arbitrary
```
lehins: You already adjusted the size of Value in `TxOut`
```suggestion
      <*> arbitrary
```

lehins (changes requested):
Recent changes made generator from generating a lot to generating no multiassets for TxOuts at all. I do recommend trying out in ghci and see the affects of any changes to any generator you are doing. It is very easy to mess it up and cause poor generators to mask bugs in implementation

eras/shelley-ma/test-suite/test/Test/Cardano/Ledger/ShelleyMA/Serialisation/Golden/Encoding.hs
aniketd: This has now been replaced with two tests: one asserting that `MultiAsset` decodes correctly with negative values, and another asserting that `MaryValue` decoding fails with negative values.

eras/mary/impl/cardano-ledger-mary.cabal
aniketd: Yes, I will fix this in the very next PR #3285 for sure :+1:

eras/shelley-ma/test-suite/src/Test/Cardano/Ledger/ShelleyMA/Serialisation/Generators.hs
```diff
@@ -166,40 +166,20 @@ instance
   arbitrary =
     MaryTxBody
       <$> arbitrary
-      <*> arbitrary
+      <*> scale (`div` 15) arbitrary
```
lehins: One last place. Size for value in TxOut has already been adjusted
```suggestion
      <*> arbitrary
```

lehins (approved):
One last change is needed otherwise this is great.

Awesome work @aniketd!

JaredCorduan:
yep, thanks again for all this work @aniketd ! it turned out to be much stickier than I would have guessed!

:raised_hands:

Issue #3037, opened 2022-09-15
Make the CI abort actions on older commits when a new commit gets pushed


Soupstraw:
Right now the CI seems to continue running actions on older commits when I push newer commits to a PR branch. This results in a lot of spam in my notifications and also probably clogs up the CI job action queue. 

It might be a good idea to halt all previous actions once a new commit gets added, so that the actions will only be run on the latest commit in a PR.

lehins:
Fixed in #3244

PR #3244, opened 2023-01-10
Speedup CI

lehins:
# Description

* Improved runtime of "Download Mainnet Mirror" from 5~8 Min to 1.5 Min. Renamed it from a less descriptive "Git Clone" name
* Fixed caching. Shaves off 30-50 minutes upon a cache hit.
* Stop triggering CI on pushes, which caused duplicate runs. We want to run upon any PR, regardless of the target branch. Fixed.

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/.github/workflows/haskell.yml
+++ b/.github/workflows/haskell.yml
@@ -92,12 +92,16 @@ jobs:
     - name: Build
       run: cabal build all
 
-    - name: Git clone
-      run: git clone https://github.com/input-output-hk/cardano-mainnet-mirror
+    - name: Download Mainnet Mirror
+      run: |
+        REF=a31ac75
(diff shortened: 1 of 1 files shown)


Soupstraw:
Does this fix #3037?

lehins:
> Does this fix https://github.com/input-output-hk/cardano-ledger/issues/3037?

No this is unrelated. I have no idea how to depend one run on another. So, killing one run depending when another one for the same trigger is present, seems like an impossible task for me with GithubActions. However, Github Actions surprised me in positive ways many times before and I might be totally wrong on this.

JaredCorduan (approved):
wonderful!

PR #3245, opened 2023-01-11
Rename and export specs

### LINKED ISSUE #2810: Version the formal specs
It would be very helpful to have a way to store versioned (perhaps with git tags) copies of the formal specs. A permanent link, somewhere in GitHub, would be ideal.

It would be nice to have a CI job that packages up the PDFs into tar.gz upon a creation of `cardano-ledger-spec-x.y.z` tag, and uploads it into the github releases section.

teodanciu:
# Description
* Renames generated specs in a more consistent way.
* Removes a [bak file](https://github.com/input-output-hk/cardano-ledger/blob/master/eras/shelley/design-spec/delegation_design_spec.tex.bak) that I assume was added by accident 
* Adds a github action that:
  * gets triggered when a tag is pushed that has name: cardano-ledger-specs-[0-9]+-[0-9]+-[0-9]+   (for example: cardano-ledger-specs-1.0.0)
  * creates a release for this tag, and attaches to the release the corresponding specs, as generated by `nix-build -A specs` 
* Update the links the specs from the dangling hydra ones to the latest github release ones

For an example of how the releases with the attached docs look, you can see my fork here: 
 releases:  https://github.com/bienpulenta/cardano-ledger/releases and
 actions that created these releases: https://github.com/bienpulenta/cardano-ledger/actions 
 
(I used this repo as a test, in order to not create releases on our main repo)
Closes https://github.com/input-output-hk/cardano-ledger/issues/2810

# Checklist

- [x] Self-reviewed the diff


.github/workflows/push-specs.yml
```diff
@@ -0,0 +1,35 @@
...
+          mkdir -p deref
+          for d in output/* ; do
+            dr="$(basename $d)";
+            cp -L $d/*.pdf deref;
+          done
+      - name: Create specs release
+        uses: ncipollo/release-action@v1
+        with:
+          artifacts: "deref/*"
+
```
lehins, line 35: @teodanciu If I am not mistaken I think we decided to push releases to this repo: https://github.com/input-output-hk/formal-ledger-specifications to a `gh-pages` branch. This way we can setup github pages and people will be able to access the pdfs with a nice url

JaredCorduan: actually, @teodanciu and I went round in circles yesterday trying to find the best solution. The trans-repo solution involved a few things that we didn't love, the worst of which was having to tie the security to a single individual's github account, and having that token be updated once a year. so we came back to the release idea, and we discovered that we could create a stable link ("latest") to the pdfs in a release, which can be triggered by pushing a tag.

lehins: Oh, so it will upload pdfs to the release page? That's even better. That's what I liked and advocated for all along :smile:

JaredCorduan: Indeed it will, that's exactly what this PRs does. I was about to say "wasn't it you that suggested this idea in the first place" :laughing: .

JaredCorduan (approved):
this is wonderful, thank you @teodanciu ! great catch on that `bak` file.

PR #3248, opened 2023-01-12
Implement a more useful `ensureMinCoinTxOut`

### LINKED ISSUE #3247: Add `ensureMinCoinTxOut`
I think it is also possible that someone would like to just ensure that there TxOut satisfies the minimum output condition, rather than set it to the minimum possible value. It would be good to add a function that is very similar to `setMinCoinTxOut`, but is a bit more lenient.

lehins:
# Description

Fixes #3247 

Also switches `cardano-ledger-api` test suite to `hspec`

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/libs/cardano-ledger-api/src/Cardano/Ledger/Api/Tx/Out.hs
+++ b/libs/cardano-ledger-api/src/Cardano/Ledger/Api/Tx/Out.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE BangPatterns #-}
 {-# LANGUAGE DataKinds #-}
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE RankNTypes #-}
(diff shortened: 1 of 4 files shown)


libs/cardano-ledger-api/src/Cardano/Ledger/Api/Tx/Out.hs
```diff
@@ -43,45 +48,86 @@ module Cardano.Ledger.Api.Tx.Out (
...
+  TxOut era ->
+  TxOut era
+setMinCoinTxOutInternal f pp = go
   where
-    go txOut =
+    go !txOut =
       let curMinCoin = getMinCoinTxOut pp txOut
           curCoin = txOut ^. coinTxOutL
-       in if curCoin == curMinCoin
+       in if curCoin `f` curMinCoin
```
JaredCorduan, line 74: sure you can't be tempted by some fun and obscure unicode symbol? :laughing:

JaredCorduan (approved):
looks great! I love the great docs!

libs/cardano-ledger-api/src/Cardano/Ledger/Api/Tx/Out.hs
lehins: I don't think it is possible to have infix operators as function argument names .

JaredCorduan: of well, `f` is a great name too :grin:

PR #3253, opened 2023-01-17
Forgo resolving pointer addresses

teodanciu:
# Description

* The first 3 commits are a reorganization of `TestChain` in cardano-ledger-shelley-test.
   TestChain is large module that defines and aggregates chain properties pertaining to different rules, as well as defining 
   functions for working with the trace generation.
   This PR splits the module into several cohesive ones, leaving TestChain to contain utilities used by the others. 
   It also renames and simplifies some functions for testing incremental state (in the newly extracted module IncrementalStake), 
   in preparation for making the incremental stake computation dependent on PParams.
* The next 3 commits are the actual point of this PR: introducing a Hardfork switch that turns off the pointer addresses starting with version 9, for both:  Utxo (`IncrementalStake.updateStakeDistribution`) and Snap (`IncrementalStake.incrementalStakeDistr`
* The remaining commits deal with making the tests pass with the new switch. For this, a new constant `maxMajorPV` is controlling the protocol version generated in the trace, which in our case we set to 9, corresponding to the Hardfork switch. 
Without this, the tests fail with an Assertion Violation error  in Ledger Rule:
```   
AssertionViolation (ShelleyLEDGER (ShelleyEra C_Crypto))
   Deposit pot must equal obligation (ShelleyLedger)
``` 

# Checklist

- [x] Self-reviewed the diff


JaredCorduan (commented):
This is looking great @teodanciu ! I know this is still a WIP... You've addressed the point where dangling pointers are addressed immediately before we take a stake snapshot. (and it all looks exactly as I'd have expected).  We still need to address the spot in the incremental computation, namely here:

https://github.com/input-output-hk/cardano-ledger/blob/bb0b5a2c4218a7704c45e2066a2de31fe2683b8f/eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState/IncrementalStake.hs#L124

https://github.com/input-output-hk/cardano-ledger/blob/bb0b5a2c4218a7704c45e2066a2de31fe2683b8f/eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState/IncrementalStake.hs#L92-L93

This makes me now realize that "pointer stake" which is already accounted for in the stake distribution may be a bit tricky to deal with. We may have to address this by deleting it in the hard fork combinator translation from babbage to conway. :thinking: 
This function `incrementalAggregateUtxoCoinByCredential` is used both to add and to remove stake associated by pointer addresses, when observing txout and txins respectively:

SebastienGllmt:
Twice in one week?!

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Epoch.hs
```diff
@@ -75,7 +75,7 @@ data ConwayEpochEvent era
 instance
   ( EraTxOut era
   , Embed (EraRule "SNAP" era) (ConwayEPOCH era)
-  , Environment (EraRule "SNAP" era) ~ LedgerState era
```
JaredCorduan, line 76: good catch!

eras/shelley/impl/src/Cardano/Ledger/Shelley/HardForks.hs
```diff
@@ -65,3 +66,8 @@ translateUpperBoundForPlutusScripts ::
   ProtVer ->
   Bool
 translateUpperBoundForPlutusScripts pv = pvMajor pv > natVersion @8
+
+forgoPointerAddressResolution ::
+  ProtVer ->
+  Bool
+forgoPointerAddressResolution pv = pvMajor pv > natVersion @9
```
JaredCorduan:
```suggestion
forgoPointerAddressResolution pv = pvMajor pv > natVersion @8
```

JaredCorduan: We can stop counting pointer stake starting with the Conway era, ie version 9

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Snap.hs
```diff
@@ -80,16 +83,14 @@ instance EraTxOut era => STS (ShelleySNAP era) where
...
+  TRC (snapEnv, s, _) <- judgmentContext
 
-  let LedgerState
-        (UTxOState _utxo _ fees _ incStake)
-        (DPState dstate pstate)
-        _ = lstate
-      -- stakeSnap = stakeDistr @era utxo dstate pstate  -- HISTORICAL NOTE
-      istakeSnap = incrementalStakeDistr @(EraCrypto era) incStake dstate pstate
+  let SnapEnv (LedgerState (UTxOState _utxo _ fees _ incStake) (DPState dstate pstate) _) pp = snapEnv
+      -- stakeSnap = stakeDistr @era utxo dstate pstate  -EraCrypto- HISTORICAL NOTE
```
JaredCorduan: :thinking: how can we make this clearer? (I realize you did not introduce it...). maybe:

> This is equivalent to what is written in the spec, namely `stakeSnap = stakeDistr @era utxo dstate pstate`

?

eras/shelley/test-suite/cardano-ledger-shelley-test.cabal
```diff
@@ -163,6 +168,7 @@ test-suite cardano-ledger-shelley-test
       -- The 4 megabytes stack bound and 250 megabytes heap bound were
       -- determined ad-hoc.
       "-with-rtsopts=-K4m -M250m"
+      -fprint-potential-instances
```
JaredCorduan: TIL!

eras/shelley/test-suite/src/Test/Cardano/Ledger/Shelley/Generator/Update.hs
```diff
@@ -100,7 +100,7 @@ genPParams ::
   (EraPParams era, ProtVerAtMost era 4, ProtVerAtMost era 6) =>
   Constants ->
   Gen (PParams era)
-genPParams c@Constants {maxMinFeeA, maxMinFeeB, minMajorPV} = do
+genPParams c@Constants {maxMinFeeA, maxMinFeeB, minMajorPV, maxMajorPV} = do
```
JaredCorduan, line 103: I'm surprised we can get away with updating the max major protocol here without a ton of headache...

JaredCorduan (approved):
This looks great, thank you @teodanciu ! All the cleanup and organization is much appreciated!!

eras/shelley/test-suite/cardano-ledger-shelley-test.cabal
teodanciu: Oh no, i left this in by mistake!! Well it helped me debug, but we probably should leave it in, so good catch on you!

eras/shelley/impl/src/Cardano/Ledger/Shelley/HardForks.hs
teodanciu: Oops :facepalm:   Thanks!

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Snap.hs
teodanciu: Yes good idea, a few times I deleted this before reading the `HISTORICAL NOTE` end,  much better to write a human explanation like you proposed

eras/shelley/impl/src/Cardano/Ledger/Shelley/HardForks.hs
lehins: IMHO using `>=` would make this a bit more obvious. Not suggesting changing it though, unless we change it in the whole module, cause otherwise  inconsistency would make it even more confusing.

eras/shelley/test-suite/src/Test/Cardano/Ledger/Shelley/Generator/Update.hs
teodanciu: It's the new one `maxMajorPV` in Constants that I introduced, not the existing one in `Globals`. I haven't managed to fully understand the impact of changing that one, so I decided it's better to control how we generate the data than to tune it from that flag..

eras/shelley/impl/src/Cardano/Ledger/Shelley/HardForks.hs
teodanciu: I agree, the number (8 or 9) is a stronger signal that stays in your mind than the operator is, so you associate the idea with the number. Happy to change it in another commit

eras/shelley/test-suite/bench/Main.hs
```diff
@@ -223,9 +224,9 @@ epochAt x =
       ("UTxO=" ++ show x ++ ",  address=" ++ show n)
       [ bench "stakeDistr" (nf action2m arg)
       , bench "incrementalStakeDistr" (nf action2im arg)
-      , env (pure (updateStakeDistribution mempty mempty utxo)) $ \incStake ->
+      , env (pure (updateStakeDistribution def mempty mempty utxo)) $ \incStake ->
           bench "incrementalStakeDistr (no update)" $
-            nf (incrementalStakeDistr incStake dstate) pstate
+            nf (\d -> incrementalStakeDistr (def :: PParams Shelley) incStake dstate d) pstate
```
lehins: Eta expansion isn't really necessary, also instead of importing `def` and adding `data-default-core`, it can be more explicit to use `emptyPParams`, which is already in scope
```suggestion
            nf (incrementalStakeDistr (emptyPParams @Shelley) incStake dstate) pstate
```

eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState/IncrementalStake.hs
```diff
@@ -186,12 +193,14 @@ smartUTxOState utxo c1 c2 st =
...
 --   This function has a non-incremental analog, 'stakeDistr', mosty used in tests, which does use the UTxO.
 incrementalStakeDistr ::
-  forall c.
+  forall c era.
+  EraPParams era =>
+  PParams era ->
   IncrementalStake c ->
   DState c ->
   PState c ->
   SnapShot c
```
lehins: `c` depends on `era`, so now it has to be:
```haskell
  forall era.
  EraPParams era =>
  PParams era ->
  IncrementalStake (EraCrypto era) ->
  DState (EraCrypto era) ->
  PState (EraCrypto era) ->
  SnapShot (EraCrypto era)
```

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Snap.hs
```diff
@@ -80,16 +83,14 @@ instance EraTxOut era => STS (ShelleySNAP era) where
...
 
-  let LedgerState
-        (UTxOState _utxo _ fees _ incStake)
-        (DPState dstate pstate)
-        _ = lstate
-      -- stakeSnap = stakeDistr @era utxo dstate pstate  -- HISTORICAL NOTE
-      istakeSnap = incrementalStakeDistr @(EraCrypto era) incStake dstate pstate
+  let SnapEnv (LedgerState (UTxOState _utxo _ fees _ incStake) (DPState dstate pstate) _) pp = snapEnv
+      -- per the spec: stakeSnap = stakeDistr @era utxo dstate pstate
+      istakeSnap = incrementalStakeDistr @(EraCrypto era) pp incStake dstate pstate
```
lehins: Type application is no longer needed, because `pp` will resolve the crypto as well.
```suggestion
      istakeSnap = incrementalStakeDistr pp incStake dstate pstate
```

This comment applies to all places where `incrementalStakeDistr` is used with `@(EraCrypto era)`

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Epoch.hs
```diff
@@ -180,7 +179,7 @@ epochTransition = do
   let utxoSt = lsUTxOState ls
   let DPState dstate pstate = lsDPState ls
   ss' <-
-    trans @(EraRule "SNAP" era) $ TRC (ls, ss, ())
+    trans @(EraRule "SNAP" era) $ TRC ((SnapEnv ls pp), ss, ())
```
lehins: Redundant parenthesis
```suggestion
    trans @(EraRule "SNAP" era) $ TRC (SnapEnv ls pp, ss, ())
```

eras/shelley/test-suite/bench/Main.hs
```diff
@@ -237,12 +238,14 @@ action2m ::
...
 
 action2im ::
+  forall era.
   EraTxOut era =>
   (DState (EraCrypto era), PState (EraCrypto era), UTxO era) ->
   EB.SnapShot (EraCrypto era)
 action2im (dstate, pstate, utxo) =
-  let incStake = updateStakeDistribution mempty mempty utxo
-   in incrementalStakeDistr incStake dstate pstate
+  let pp = def :: PParams era
```
lehins: This also is needed to remove `Data.Default` import
```suggestion
  let pp = emptyPParams @era
```

eras/shelley/test-suite/src/Test/Cardano/Ledger/Shelley/Rules/Pool.hs
```diff
@@ -0,0 +1,208 @@
+{-# LANGUAGE AllowAmbiguousTypes #-}
```
lehins, line 1: We should try our best to preserve git history whenever possible when we move large chunks of code around. Here it could have been done easily first with
```
$ git mv eras/shelley/test-suite/src/Test/Cardano/Ledger/Shelley/Rules/TestPool.hs eras/shelley/test-suite/src/Test/Cardano/Ledger/Shelley/Rules/Pool.hs
```

as a separate commit , followed by a commit that modifies the new file.


Ability to resolve conflicts easily when rebasing is the most important reason why we should try and preserve history. Of course ability to easily looking up history of a piece of code with git-blame  is quite valuable as well.

eras/shelley/test-suite/src/Test/Cardano/Ledger/Shelley/Rules/PoolReap.hs
```diff
@@ -0,0 +1,116 @@
+{-# LANGUAGE AllowAmbiguousTypes #-}
```
lehins, line 1: Same goes for this module and the removed `TestPoolReap`. It should not be too hard to fix that, let me know if you need help with getting that history back

eras/shelley/test-suite/src/Test/Cardano/Ledger/Shelley/Rules/AdaPreservation.hs
```diff
@@ -0,0 +1,656 @@
+{-# LANGUAGE AllowAmbiguousTypes #-}
+{-# LANGUAGE ConstraintKinds #-}
+{-# LANGUAGE DataKinds #-}
+{-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE NamedFieldPuns #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeApplications #-}
+{-# LANGUAGE TypeFamilies #-}
+
+module Test.Cardano.Ledger.Shelley.Rules.AdaPreservation (
```
lehins, line 10: Same with this module. Over half a thousand lines of code have moved from another module. Any modifications to the original code that are inflight right now in other PRs will have a hell of a time rebasing on this work. So it is easier to fix it in this PR by retaining the git history. Also it would make this PR a lot easier to review, cause right now all new modules look like new code, despite most of it is old with some modifications.

libs/ledger-state/src/Cardano/Ledger/State/Query.hs
```diff
@@ -11,6 +11,7 @@ module Cardano.Ledger.State.Query where
 
 import Cardano.Ledger.Alonzo.TxBody as Alonzo
 import Cardano.Ledger.Binary
+import Cardano.Ledger.Core (PParams)
```
lehins:
```suggestion
import Cardano.Ledger.Core (emptyPParams)
```

libs/ledger-state/src/Cardano/Ledger/State/Query.hs
```diff
@@ -31,6 +32,7 @@ import Control.Monad
 import Control.Monad.Trans.Reader
 import Data.Conduit.Internal (zipSources)
 import Data.Conduit.List (sourceList)
+import Data.Default.Class (def)
```
lehins:
```suggestion
```

libs/ledger-state/src/Cardano/Ledger/State/Query.hs
```diff
@@ -516,7 +518,7 @@ getLedgerState utxo LedgerState {..} dstate = do
   pure
     Shelley.LedgerState
       { Shelley.lsUTxOState =
-          Shelley.smartUTxOState utxo utxoStateDeposited utxoStateFees utxoStatePpups -- Maintain invariant
+          Shelley.smartUTxOState (def :: PParams CurrentEra) utxo utxoStateDeposited utxoStateFees utxoStatePpups -- Maintain invariant
```
lehins:
```suggestion
          Shelley.smartUTxOState (emptyPParams @CurrentEra) utxo utxoStateDeposited utxoStateFees utxoStatePpups -- Maintain invariant
```

libs/ledger-state/ledger-state.cabal
```diff
@@ -39,6 +39,7 @@ library
                     , cardano-ledger-shelley
                     , conduit
                     , containers
+                    , data-default-class
```
lehins:
```suggestion
```

eras/shelley/test-suite/bench/Main.hs
```diff
@@ -57,6 +57,7 @@ import Criterion.Main (
   whnf,
   whnfIO,
  )
+import Data.Default.Class (def)
```
lehins:
```suggestion
```

eras/shelley/test-suite/bench/Main.hs
```diff
@@ -223,9 +224,9 @@ epochAt x =
       ("UTxO=" ++ show x ++ ",  address=" ++ show n)
       [ bench "stakeDistr" (nf action2m arg)
       , bench "incrementalStakeDistr" (nf action2im arg)
-      , env (pure (updateStakeDistribution mempty mempty utxo)) $ \incStake ->
+      , env (pure (updateStakeDistribution def mempty mempty utxo)) $ \incStake ->
```
lehins:
```suggestion
      , env (pure (updateStakeDistribution emptyPParams mempty mempty utxo)) $ \incStake ->
```

eras/shelley/test-suite/cardano-ledger-shelley-test.cabal
```diff
@@ -224,6 +228,7 @@ benchmark mainbench
                 cardano-strict-containers,
                 containers,
                 criterion,
+                data-default-class,
```
lehins: With all other comments this one shouldn't be necessary
```suggestion
```

eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState/IncrementalStake.hs
```diff
@@ -200,8 +209,13 @@ incrementalStakeDistr incstake ds ps =
...
     delegs_ = UM.viewToVMap (delegations ds)
-    -- A credential is active, only if it is being delegated
-    step1 = resolveActiveIncrementalPtrs (`VMap.member` delegs_) ptrmap incstake
+    -- A credential is active, only if it is being delegated: (dom activeDelegs ◁ credStake)
+    activeCreds = Map.filterWithKey (\k _ -> VMap.member k delegs_) credStake
+    ignorePtrs = HardForks.forgoPointerAddressResolution (pp ^. ppProtocolVersionL)
+    step1 =
+      if ignorePtrs
+        then activeCreds
+        else resolveActiveIncrementalPtrs (`VMap.member` delegs_) ptrmap (IStake activeCreds ptrStake)
```
lehins: The `step1A` was extracted from `resolveActiveIncrementalPtrs`, so the function looses its usefulness in my opinion and makes it just a little more confusing. I think it'll be cleaner if we inline it:
```suggestion
        else Map.foldlWithKey' addResolvedPointer activeCreds ptrStake
    addResolvedPointer ans ptr coin =
      case Map.lookup ptr ptrsMap of -- Map of ptrs to Credentials
        Just cred
          | VMap.member cred delegs_ -> Map.insertWith (<>) cred coin ans
        _ -> ans
```

Could you also rename `ptrmap` to `ptrsMap` and `tripmap` to `triplesMap`?

eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState/IncrementalStake.hs
```diff
@@ -200,8 +209,13 @@ incrementalStakeDistr incstake ds ps =
     UMap tripmap ptrmap = dsUnified ds
     PState {psStakePoolParams = poolParams} = ps
     delegs_ = UM.viewToVMap (delegations ds)
-    -- A credential is active, only if it is being delegated
-    step1 = resolveActiveIncrementalPtrs (`VMap.member` delegs_) ptrmap incstake
+    -- A credential is active, only if it is being delegated: (dom activeDelegs ◁ credStake)
+    activeCreds = Map.filterWithKey (\k _ -> VMap.member k delegs_) credStake
+    ignorePtrs = HardForks.forgoPointerAddressResolution (pp ^. ppProtocolVersionL)
+    step1 =
```
lehins, line 214:
```suggestion
    -- For pre Conway: (dom activeDelegs ◁ credStake) ∪ (dom activeDelegs ◁ ptrStake)
    -- afterwards we forgo ptr resolution: (dom activeDelegs ◁ credStake)
    step1 =
```

lehins (changes requested):
Great work!

There a few minor things that can be improved. In particular retention git history and type signature for `incrementalStakeDistr`. Please, let me know if you need help with the former, it could be tricky to make git remember things.

eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState/IncrementalStake.hs
teodanciu: Love this version, less noise! 
I suppose we could look up directly in `activeCreds` instead, since we need that map anyway for the "forgo" case, so
`Just cred | Map.member cred activeCreds ->  Map.insertWith (<>) cred coin ans`

lehins: > I suppose we could look up directly in activeCreds instead

I am not sure we can do that. I don't think this is equivalent: `Map.member cred activeCreds` === `VMap.member cred delegs_` for `cred` in the PtrsMap. Because it could be the case that cred is in`deleg_`, but not in `activeCreds`, because it is not part of `credStake`, but instead is part of `ptrStake`

teodanciu: It took me a while, but indeed!

eras/shelley/test-suite/src/Test/Cardano/Ledger/Shelley/Rules/AdaPreservation.hs
teodanciu: Thanks for the suggestion, I separated the renames in a different commit (after using `git mv` as suggested) for the three files that I have renamed  (Pool, PoolReap and Deleg). 

However, not sure about this technique for the files that I created by extracting content from the big original file, `TestChain` (including this file AdaPreservation). If i do `git mv` it will rename the whole file, no?  
Is there a way to git mv just a part of a file? Or how can  i keep the history for parts of a file?

eras/shelley/test-suite/src/Test/Cardano/Ledger/Shelley/Rules/Pool.hs
teodanciu: Done by making all renames in a [separate commit](https://github.com/input-output-hk/cardano-ledger/pull/3253/commits/6f323603f2bdc81dbcf2ca2db0b48a485767c66a).

eras/shelley/test-suite/src/Test/Cardano/Ledger/Shelley/Rules/AdaPreservation.hs
lehins: Very good question. This can be done with two separate commits:

```shell
$ cp TestChain.hs TestChainBak.hs
$ git add TestChainBak.hs
$ git mv AdaPreservation.hs
$ git commit TestChainBak.hs AdaPreservation.hs -m "Make a copy of two files with the same history"
$ git mv TestChainBak.hs TestChain.hs
$ git commit TestChain.hs -m "Moved back into the original"
```

At this point you have two files with the desired names and the same git history. Subsequent commit can include modification to both of the files: 
* `TestChain.hs` has stuff removed that now will live in `AdaPreservation.hs`
* `AdaPreservation.hs` has the stuff removed that stayed in `TestChain.hs`

This is were I stole it from: https://stackoverflow.com/questions/3887736/keep-git-history-when-splitting-a-file#answer-33992444

I know it is a bit elaborate way of going about it., but it is worth it.

teodanciu: That's crazy. I love it

teodanciu: Done, but at the cost of many extra commits (2  extra for each extraction, so 8 extra commits). 
Now github history of the 4 new files (`IncrementalStake.hs`, `AdaPreservation.hs`, `CollisionFreeness.hs`, `Deposits.hs`)  shows the rename which then can be followed, and blame shows like in the initial file

lehins: Hell yeah!!! Thank you Teodora for putting this extra effort. Extra commits is a small price to pay for keeping all that git history.

lehins (approved):
Looks great!

JaredCorduan:
:raised_hands:

PR #3260, opened 2023-01-17
Update release tag pattern 

teodanciu:
so that releases get sorted in a deterministic way.
# Description

Without this change, if we realease:
`cardano-ledger-spec-1.0.0`
`cardano-ledger-spec-2.0.0`
....
`cardano-ledger-spec-10.0.0` 

then the latter won't be considered the latest,  because the sorting is done alphabetically. 

If we use `cardano-ledger-spec-YYYY-MM-DD` tag as proposed in this PR and tag based on the current date, then the provided alphabetical sorting should give us the desired sorting by latest. 

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/.github/workflows/push-specs.yml
+++ b/.github/workflows/push-specs.yml
@@ -3,7 +3,7 @@ name: Push Specs
 on:
   push:
     tags:
-      - 'cardano-ledger-specs-[0-9]+.[0-9]+.[0-9]+'
+      - 'cardano-ledger-spec-[0-9][0-9][0-9][0-9]-[0-9][0-9]-[0-9][0-9]'
 
 jobs:
   push-docs:
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -91,6 +91,22 @@ cabal configure <package-name> --ghc-options="-Wwarn"
 cabal build <package-name>
 ```
 
+## Publishing specifications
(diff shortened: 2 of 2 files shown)


.github/workflows/push-specs.yml
```diff
@@ -3,7 +3,7 @@ name: Push Specs
 on:
   push:
     tags:
-      - 'cardano-ledger-specs-[0-9]+.[0-9]+.[0-9]+'
+      - 'cardano-ledger-spec-[0-9][0-9][0-9][0-9]-[0-9][0-9]-[0-9][0-9]'
```
teodanciu, line 6: [0-9]{4} syntax doesn't seem supported

CONTRIBUTING.md
```diff
@@ -91,6 +91,22 @@ cabal configure <package-name> --ghc-options="-Wwarn"
 cabal build <package-name>
 ```
 
+## Publishing specifications
+Pdf specs are stored as attachments to [github releases](https://github.com/input-output-hk/cardano-ledger/releases)
```
JaredCorduan:
```suggestion
PDF specs are stored as attachments to [github releases](https://github.com/input-output-hk/cardano-ledger/releases)
```

JaredCorduan (approved):
fantastic!

PR #3263, opened 2023-01-18
Fix Redeemer codec

### LINKED ISSUE #3237: Gracefully fix `Redeemer` (de)serialization
Gracefully fix `Redeemers` encoder/decoder length value from `4` to `3`. In other words:

* For Conway deserializer should be able to accept both values `3` and `4`
* For Conway serializer should use the correct value `3`
* Create a ticket that tracks disabling incorrect value `4` in deserializer in the era after `Conway` 

Encoder:

https://github.com/input-output-hk/cardano-ledger/blob/71c31ace17467ee02f065ad9fc5f640e5f1e4091/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxWits.hs#L139-L142

Decoder:

https://github.com/input-output-hk/cardano-ledger/blob/71c31ace17467ee02f065ad9fc5f640e5f1e4091/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxWits.hs#L489-L495

aniketd:
# Description

Closes #3237 

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxWits.hs
+++ b/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxWits.hs
@@ -23,6 +23,7 @@ module Cardano.Ledger.Alonzo.TxWits (
     Redeemers,
     Redeemers'
   ),
+  RedeemersRaw (..),
   unRedeemers,
   nullRedeemers,
   TxDats (TxDats, TxDats'),
@@ -58,17 +59,24 @@ import Cardano.Ledger.Alonzo.Language (Language (..))
 import Cardano.Ledger.Alonzo.Scripts (AlonzoScript (..), ExUnits (..), Tag)
 import Cardano.Ledger.Alonzo.Scripts.Data (Data, hashData)
 import Cardano.Ledger.Binary (
-  Annotator,
+  Annotator (..),
   Decoder,
+  DecoderError (DecoderErrorSizeMismatch),
   FromCBOR (..),
   FromCBORGroup (..),
   ToCBOR (..),
   ToCBORGroup (..),
+  cborError,
+  decodeBreakOr,
   decodeList,
+  decodeListLenOrIndef,
   encodeFoldableEncoder,
   encodeListLen,
+  natVersion,
  )
 import Cardano.Ledger.Binary.Coders
+import Cardano.Ledger.Binary.Decoding (DecoderError (DecoderErrorCustom), ifDecoderVersionAtLeast)
+import Cardano.Ledger.Binary.Encoding (ifEncodingVersionAtLeast)
 import Cardano.Ledger.Core
 import Cardano.Ledger.Crypto (Crypto (DSIGN, HASH), StandardCrypto)
 import Cardano.Ledger.Keys (KeyRole (Witness))
@@ -84,12 +92,14 @@ import Cardano.Ledger.MemoBytes (
 import Cardano.Ledger.SafeHash (SafeToHash (..))
 import Cardano.Ledger.Shelley.TxBody (WitVKey)
 import Control.DeepSeq (NFData)
+import Control.Monad (unless, when)
 import Data.Map.Strict (Map)
 import qualified Data.Map.Strict as Map
 import Data.Maybe (mapMaybe)
 import Data.Proxy (Proxy (..))
 import Data.Set (Set)
 import qualified Data.Set as Set
+import qualified Data.Text as Text
 import Data.Typeable (Typeable)
 import Data.Word (Word64)
 import GHC.Generics (Generic)
@@ -133,10 +143,15 @@ newtype RedeemersRaw era = RedeemersRaw (Map RdmrPtr (Data era, ExUnits))
 deriving instance HashAlgorithm (HASH (EraCrypto era)) => Show (RedeemersRaw era)
 
 instance Typeable era => ToCBOR (RedeemersRaw era) where
-  toCBOR (RedeemersRaw rs) = encodeFoldableEncoder keyValueEncoder $ Map.toAscList rs
+  toCBOR (RedeemersRaw rs) =
+    ifEncodingVersionAtLeast
+      (natVersion @9)
+      (encode' 3)
+      (encode' 4)
     where
-      keyValueEncoder (ptr, (dats, exs)) =
-        encodeListLen 4
+      encode' n = encodeFoldableEncoder (keyValueEncoder n) $ Map.toAscList rs
+      keyValueEncoder numFields (ptr, (dats, exs)) =
+        encodeListLen numFields
           <> toCBORGroup ptr
           <> toCBOR dats
           <> toCBOR exs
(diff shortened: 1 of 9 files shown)


libs/cardano-ledger-binary/src/Cardano/Ledger/Binary/Decoding/Decoder.hs
```diff
@@ -529,7 +529,7 @@ decodeRecordNamedT name getRecordSize decoder = do
   lenOrIndef <- lift decodeListLenOrIndef
   x <- decoder
   lift $ case lenOrIndef of
-    Just n -> matchSize (Text.pack "Record " <> name) n (getRecordSize x)
+    Just n -> matchSize (Text.pack "Record " <> name) (getRecordSize x) n
```
aniketd, line 532: Argument positions have been swapped to correct the earlier mistake.

aniketd:
@lehins This is ready for your initial quick review of the implementation. I still need to add some tests and necessary instances.

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxWits.hs
```diff
@@ -480,20 +497,41 @@ instance
...
+      decodeRedeemerRecord :: Decoder s (Annotator (RdmrPtr, (Data era, ExUnits)))
+      decodeRedeemerRecord = do
+        rdmrPtr <- fromCBORGroup
+        dataEra <- fromCBOR
+        exUnits <- fromCBOR
+        let f :: RdmrPtr -> (Data era) -> ExUnits -> (RdmrPtr, (Data era, ExUnits))
+            f x y z = (x, (y, z))
+        pure $ f rdmrPtr <$> dataEra <*> pure exUnits
+      decodePreConway :: Decoder s (Annotator (RdmrPtr, (Data era, ExUnits)))
+      decodePreConway = decodeRecordNamed "redeemer" (const 4) $ decodeRedeemerRecord
```
lehins: Redundant
```suggestion
      decodePreConway = decodeRecordNamed "redeemer" (const 4) decodeRedeemerRecord
```

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxWits.hs
```diff
@@ -480,20 +497,41 @@ instance
...
-            pure $ f rdmrPtr <$> dat <*> pure ex
-        pure $ Map.fromList <$> entries
+      dec :: Decoder s (Annotator (RdmrPtr, (Data era, ExUnits))) -> Decoder s (Annotator (Map RdmrPtr (Data era, ExUnits)))
+      dec decoder = fmap (fmap Map.fromList) (fmap sequence . decodeList $ decoder)
+      decodeRedeemerRecord :: Decoder s (Annotator (RdmrPtr, (Data era, ExUnits)))
+      decodeRedeemerRecord = do
+        rdmrPtr <- fromCBORGroup
+        dataEra <- fromCBOR
+        exUnits <- fromCBOR
+        let f :: RdmrPtr -> (Data era) -> ExUnits -> (RdmrPtr, (Data era, ExUnits))
```
lehins:
```suggestion
        let f :: RdmrPtr -> Data era -> ExUnits -> (RdmrPtr, (Data era, ExUnits))
```

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxWits.hs
```diff
@@ -480,20 +497,41 @@ instance
...
+        record <- lift decodeRedeemerRecord
+        lift $ case lenOrIndef of
+          Just n ->
+            when (n `notElem` [3, 4]) $
+              cborError $
+                DecoderErrorSizeMismatch (Text.pack "Record redeemers should be sized 3 OR 4 for Conway") n n
+          Nothing -> do
+            isBreak <- decodeBreakOr
+            unless isBreak $ cborError $ DecoderErrorCustom "redeemers" "Excess terms in array"
+        pure record
```
lehins: Identity transformer wrapper is not need. Also we don't need to report 4 as an option in the error message. 3 is the correct value, 4 is merely a compatibility check.
```suggestion
      decodeConway = do
        lenOrIndef <- decodeListLenOrIndef
        record <- decodeRedeemerRecord
        case lenOrIndef of
          Just n ->
            when (n `notElem` [3, 4]) $
              cborError $
                DecoderErrorSizeMismatch (Text.pack "Record redeemers should be sized 3 for Conway") 3 n
          Nothing -> do
            isBreak <- decodeBreakOr
            unless isBreak $ cborError $ DecoderErrorCustom "redeemers" "Excess terms in array"
        pure record
```

libs/cardano-ledger-binary/src/Cardano/Ledger/Binary/Decoding/Decoder.hs
lehins: Good catch

lehins (changes requested):
Yeah, this is looking great. Some tests and it'll be ready to go

libs/cardano-ledger-binary/testlib/Test/Cardano/Ledger/Binary/RoundTrip.hs
```diff
@@ -113,12 +118,60 @@ roundTripAnnExpectation ::
...
+
+roundTripAnnExpectation' ::
+  (Show t, Eq t, ToCBOR t, FromCBOR (Annotator t), HasCallStack) =>
+  Version ->
+  t ->
+  Expectation
+roundTripAnnExpectation' version t =
+  case roundTripAnn version t of
+    Left err -> expectationFailure $ "Failed to deserialize encoded: " ++ show err
+    Right tDecoded -> tDecoded `shouldBe` t
```
lehins: Reduce duplication.
```suggestion
roundTripAnnExpectation' version = roundTripAnnRangeExpectation version version
```

Also I strongly advise against using ticks in names, because they carry no meaning. Maybe naming it `roundTripAnnExactExpectation`? Or any other name, but not the tick.

That being said, I think it should be rather rare that only a particular version should be tested, I think it always will and should be a range, so maybe this function isn't that useful in the end. Feel free how you wanna go about it.

eras/conway/test-suite/src/Test/Cardano/Ledger/Conway/Serialisation/Roundtrip.hs
```diff
@@ -1,26 +1,34 @@
...
   TestTree
 allprops =
   testGroup
-    (show $ typeRep (Proxy @e))
-    [ testProperty "ConwayGenesis" $ roundTripCborExpectation @(ConwayGenesis (EraCrypto e))
+    ("Conway CBOR roundtrip " <> show (typeRep (Proxy @e)))
+    [ testProperty "ConwayGenesis" $
+        roundTripCborExpectation @(ConwayGenesis (EraCrypto e))
+    , testProperty "Redeemers" $
+        roundTripAnnExpectation' @(Redeemers e) (natVersion @9)
```
lehins: I don't see a reason to test it only for the version `9`, it should work for future protocol versions as well
```suggestion
        roundTripAnnExpectation @(Redeemers e) (natVersion @9) maxBound
```

libs/cardano-ledger-binary/testlib/Test/Cardano/Ledger/Binary/RoundTrip.hs
```diff
@@ -113,12 +118,60 @@ roundTripAnnExpectation ::
...
+roundTripAnnFailureExpectation' ::
+  (ToCBOR t, FromCBOR (Annotator t), HasCallStack) =>
+  Version ->
+  t ->
+  Expectation
+roundTripAnnFailureExpectation' version t =
+  case roundTripAnn version t of
+    Left _ -> pure ()
+    Right _ ->
+      expectationFailure $ "Should not have deserialized: " ++ showExpr (CBORBytes (serialize' version t))
```
lehins: Same comment here as for `roundTripAnnExpectation' `

libs/cardano-ledger-binary/testlib/Test/Cardano/Ledger/Binary/RoundTrip.hs
```diff
@@ -113,12 +118,60 @@ roundTripAnnExpectation ::
...
+  Version ->
+  Version ->
+  t ->
+  Expectation
+roundTripAnnFailureRangeExpectation fromVersion toVersion t =
+  forM_ [fromVersion .. toVersion] $ \version ->
+    case roundTripAnn version t of
+      Left _ -> pure ()
+      Right _ ->
+        expectationFailure $ "Should not have deserialized: " ++ showExpr (CBORBytes (serialize' version t))
```
lehins: We need to show the version in the message that causes the failure
```suggestion
        expectationFailure $ "Should not have deserialized: <version: " ++ show version ++"> " ++ showExpr (CBORBytes (serialize' version t))
```

eras/conway/test-suite/src/Test/Cardano/Ledger/Conway/Serialisation/Roundtrip.hs
```diff
@@ -1,26 +1,34 @@
...
+  (Era e) =>
   TestTree
 allprops =
   testGroup
-    (show $ typeRep (Proxy @e))
-    [ testProperty "ConwayGenesis" $ roundTripCborExpectation @(ConwayGenesis (EraCrypto e))
+    ("Conway CBOR roundtrip " <> show (typeRep (Proxy @e)))
+    [ testProperty "ConwayGenesis" $
+        roundTripCborExpectation @(ConwayGenesis (EraCrypto e))
+    , testProperty "Redeemers" $
```
lehins, line 34: The test we also need here is the embed test. Where we generate and serialize Babbage Redeemers , but deserialize as Conway. There are helper functions for testing this as well.  Checkout `embedTripExpectation`

lehins (approved):
Looks great. There are just a few improvements on testing that can be done, other than that it can be merged no problem.

libs/cardano-ledger-binary/testlib/Test/Cardano/Ledger/Binary/RoundTrip.hs
aniketd: Got it! Yes I agree that ticks are meaningless. My initial attempt was to directly use `roundTrip... <version> <version>` in the tests, but then reverted it back. I will put it that way reducing one level of indirection, if that is acceptable in the tests directly.

eras/conway/test-suite/src/Test/Cardano/Ledger/Conway/Serialisation/Roundtrip.hs
aniketd: Got it! I will use `roundTripAnnRangeExpectation` and add `maxBound` to it. However, as discussed on Slack, I was unsure if this test should have lower protocol versions as well, like: `(natVersion @2) maxBound`, as it fails for lower versions with the following error:
```
Encoder version: 2
Decoder version: 2
[...]
...expected size of 4, but found 3
```
...which is indicative of encoding version being 9 rather than 2?

eras/conway/test-suite/src/Test/Cardano/Ledger/Conway/Serialisation/Roundtrip.hs
aniketd: :exploding_head: Awesome! Thanks!

libs/cardano-ledger-binary/testlib/Test/Cardano/Ledger/Binary/RoundTrip.hs
```diff
@@ -82,7 +85,12 @@ roundTripFailureExpectation version x =
     Left _ -> pure ()
     Right _ ->
       expectationFailure $
-        "Should not have deserialized: " ++ showExpr (CBORBytes (serialize' version x))
+        unlines
```
lehins, line 88: `unlines`? Did you mean to do `mconcat`?
Otherwise it'll be:
```
Should not have deserialized: <version: 
9
>
...
```

aniketd:
It was discovered that `FromCBORGroup` for `RdmrPtr` uses a `listLen` of `2`, and therefore `4` happens to be the correct number of elements in the codec.

PR #3264, opened 2023-01-18
Move Annotator from cardano-binary

lehins:
# Description

When #3063 was implemented `Annotator` was reused from `cardano-binary` in order to minimize the diff. This is a follow up PR that moves the `Annotator`+`Annotated` types and logic completely into `cardano-ledger-binary`. There are no semantic changes, so changelog does not need any new entries

# Checklist

- [x] Self-reviewed the diff


JaredCorduan (approved):
the new haddocks are fantastic!!

lehins:
> the new haddocks are fantastic!!

I didn't write them, I only slightly adjusted them :slightly_smiling_face:

JaredCorduan:
> I didn't write them, I only slightly adjusted them slightly_smiling_face

oh, they are from cardano-binary. still lovely :grin:

PR #3265, opened 2023-01-19
Fix index-state syntax in cabal.project

andreabedini:
The index-state stanza in cabal.project is "last wins", so 
```
index-state: 2022-12-24T00:00:00Z
index-state: cardano-haskell-packages 2022-11-02T15:34:17Z
```
would set the index state of hackage to HEAD.

The correct syntax to set the index-state of both repositories is
```
index-state:
  , hackage.haskell.org 2022-12-24T00:00:00Z
  , cardano-haskell-packages 2022-11-02T15:34:17Z
```
haskell.nix has trouble parsing the index-state syntax with multiple repositories, so we repeat the index-state for Hackage. Cabal will ignore the first index-state stanza anyway.

# Description

to the issue.

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -12,10 +12,14 @@ repository cardano-haskell-packages
 
 -- See CONTRIBUTING for some Nix commands you will need to run if you
 -- update either of these.
--- Bump this if you need newer packages from Hackage
+-- NOTE: repeating the index-state for Hackage is a temporary workaround
+-- for hackage.nix limited parsing ability.
+-- Bump both the following dates if you need newer packages from Hackage
 index-state: 2022-12-24T00:00:00Z
+index-state:
+  , hackage.haskell.org 2022-12-24T00:00:00Z
 -- Bump this if you need newer packages from CHaP
-index-state: cardano-haskell-packages 2022-11-02T15:34:17Z
+  , cardano-haskell-packages 2022-11-02T15:34:17Z
 
 packages:
   eras/allegra/impl


lehins:
> Cabal will ignore the first index-state stanza anyway.

Are you sure about this? I definitely did not observe this behavior: "would set the index state of hackage to HEAD."

I can say for sure because this repo would have problems building for a month if hackage was at HEAD. It was deliberately set to:
```
index-state: 2022-12-24T00:00:00Z
```
because starting with 26th of December our tooling was broken due to incorrect bounds on Hackage starting with this index:
```
index-state: 2022-12-26T00:00:00Z
```

It was only recently fixed, but my point still stands, if hackage was set to HEAD I would expect it to be broken.
Could you also please link to some issues on cabal or haskell.nix repos, so we know exactly why we have to do this duplicate work, and more importantly when we can stop doing.

cabal.project
```diff
@@ -12,10 +12,14 @@ repository cardano-haskell-packages
 
 -- See CONTRIBUTING for some Nix commands you will need to run if you
 -- update either of these.
--- Bump this if you need newer packages from Hackage
+-- NOTE: repeating the index-state for Hackage is a temporary workaround
```
lehins, line 15: Yeah, this "temporary" work around will turn out to be permanent unless we check in some way that this problem has been fixed

andreabedini: I might have been seduced by optimism here :)

andreabedini:
> Are you sure about this? I definitely did not observe this behavior: "would set the index state of hackage to HEAD."

100% sure, the cabal issue reference is https://github.com/haskell/cabal/issues/8568. Also the prompt to fix the all the repos I can find is because it broke cardano-node today ([link](https://input-output-rnd.slack.com/archives/C0441DTDH8R/p1674036694944989?thread_ts=1673273768.464599&cid=C0441DTDH8R)).

Note that if you use nix, you are always safe because haskell.nix can parse the `index-state: 2022-12-24T00:00:00Z` line and give you an appropriate hackage tarball. So what cabal thinks is irrelevant in that case.

JaredCorduan (approved):

PR #3269, opened 2023-01-23
Use listLen for RmdrPtr codec

aniketd:
# Description

This is a followup from #3263, taking only the general enhancements made in that PR.

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxWits.hs
+++ b/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxWits.hs
@@ -136,7 +136,7 @@ instance Typeable era => ToCBOR (RedeemersRaw era) where
   toCBOR (RedeemersRaw rs) = encodeFoldableEncoder keyValueEncoder $ Map.toAscList rs
     where
       keyValueEncoder (ptr, (dats, exs)) =
-        encodeListLen 4
+        encodeListLen (listLen ptr + 2)
           <> toCBORGroup ptr
           <> toCBOR dats
           <> toCBOR exs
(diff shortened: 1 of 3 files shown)


JaredCorduan (approved):
LGTM! (I tried to think if there was something more clever we could do besides using that dummy value, but didn't come up with anything. maybe @lehins has an idea?) thanks!

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxWits.hs
```diff
@@ -494,7 +494,15 @@ instance
...
-          . decodeRecordNamed "redeemer" (const 4)
+          . decodeRecordNamed
+            "redeemer"
+            ( const
+                ( 2
+                    +
+                    {- This is a dummy value, but it enforces getting the correct listLen for the FromCBORGroup of RdmrPtr -}
+                    fromIntegral (listLen (RdmrPtr Spend 1))
+                )
+            )
```
lehins: It is wrong to construct a dummy value, because you do have access to the actual value:
```suggestion
            (\ (rdmPtr, _) -> fromIntegral (listLen rdmPtr) + 2)
```

lehins (approved):
Looks good. Needs one change though.

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxWits.hs
aniketd: I tried this earlier, but the type there happens to be `Int -> Annotator (RdmrPtr, (Data era, ExUnits)) -> Int`. So, even `(\(fmap fst . runAnnotator -> rdmrPtr) -> fromIntegral (listLen rdmrPtr) + 2)` does not work, as the `FullByteString -> a` remains a `Functor` that expects its argument to be supplied during decoding, I presume?

lehins: That is a fairly easy problem to solve. All you gotta do is push the `decodeRecordName` inside where `rdmrPtr` hasn't been pushed inside an `Annotator`. Here is the full working solution for you:


```haskell

instance Era era => FromCBOR (Annotator (RedeemersRaw era)) where
  fromCBOR = do
    entries <- fmap sequence . decodeList $ decodeAnnElement
    pure $ RedeemersRaw . Map.fromList <$> entries
    where
      decodeAnnElement :: forall s. Decoder s (Annotator (RdmrPtr, (Data era, ExUnits)))
      decodeAnnElement = do
        (rdmrPtr, dat, ex) <- decodeElement
        let f x y z = (x, (y, z))
        pure $ f rdmrPtr <$> dat <*> pure ex
      decodeElement :: forall s. Decoder s (RdmrPtr, Annotator (Data era), ExUnits)
      decodeElement = do
        decodeRecordNamed
          "Redeemer"
          (\(rdmrPtr, _, _) -> fromIntegral (listLen rdmrPtr) + 2)
          $ (,,) <$> fromCBORGroup <*> fromCBOR <*> fromCBOR
```
Just copy an paste it and it will work

aniketd: > All you gotta do is push the decodeRecordName inside where rdmrPtr hasn't been pushed inside an Annotator.

Sorry I didn't think of that!  :bowing_man:

PR #3270, opened 2023-01-23
Containers internal tests

lehins:
# Description

Some property tests that check custom functions that use internals from containers package. Nothing bad was discovered

Also adds asserts on two usages of dangerous `mergeWithKey` functions.

Fixes show instance for AsetName

# Checklist

- [x] Self-reviewed the diff


JaredCorduan (approved):
LGTM, it's always great to have more tests!

PR #3271, opened 2023-01-24
replaced balanceR and balancedL with link

TimSheard:
# Description
Fixes the loss of balance when using cannonical maps when given offsetting tokens (ie (policy asset -3) and (policy asset +3)
# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/libs/cardano-data/src/Data/CanonicalMaps.hs
+++ b/libs/cardano-data/src/Data/CanonicalMaps.hs
@@ -12,8 +12,8 @@ where
 
 import Data.Map.Internal (
   Map (..),
-  balanceL,
-  balanceR,
+  -- balanceL,
+  -- balanceR,
   link,
   link2,
  )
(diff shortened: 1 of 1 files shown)


libs/cardano-data/src/Data/CanonicalMaps.hs
```diff
@@ -36,14 +36,14 @@ instance CanonicalZero Integer where
...
-instance (Ord k, CanonicalZero v) => CanonicalZero (Map k v) where
+instance (Show k, Show v, Ord k, CanonicalZero v) => CanonicalZero (Map k v) where
   zeroC = Map.empty
   joinC = canonicalMapUnion joinC
 
 -- Note that the class CanonicalZero and the function canonicalMapUnion are mutually recursive.
 
 canonicalMapUnion ::
-  (Ord k, CanonicalZero a) =>
+  (Show k, Show a, Ord k, CanonicalZero a) =>
```
treeowl, line 46: What's with the `Show` constraints?

lehins: Leftover from debug tracing, that would be my guess.

PR #3274, opened 2023-01-30
version the new cost model deserializers

### LINKED ISSUE #2902: Problematic cost model serialization

The are two problems:

JaredCorduan:
# Description

Enforce that the cost model deserializers prior to version 9 use a list of the expected length. This is required for full backwards compatibility (otherwise the current deserializers in ledger are more flexible than those currently used by the node).

Note that the `conway` CDDL file has been changed to reflect this change (and thus the CDDL tests as well).

~This is the last changed needed to resolve #2902.~ In order to resolve #2902, we still need to support deserializing cost models for future version of Plutus (we do now support adding new builtins).

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Scripts.hs
+++ b/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Scripts.hs
@@ -69,6 +69,7 @@ import Cardano.Ledger.Binary (
   encodeFoldableAsDefLenList,
   encodeMap,
   getVersion64,
+  ifDecoderVersionAtLeast,
  )
 import Cardano.Ledger.Binary.Coders (
   Decode (Ann, D, From, Invalid, RecD, SumD, Summands),
@@ -80,6 +81,7 @@ import Cardano.Ledger.Binary.Coders (
   (<!),
   (<*!),
  )
+import Cardano.Ledger.Binary.Version (natVersion)
 import Cardano.Ledger.Coin (Coin (..))
 import Cardano.Ledger.Core
 import Cardano.Ledger.Crypto (Crypto)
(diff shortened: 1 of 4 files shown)


eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Scripts.hs
```diff
@@ -559,7 +561,43 @@ getCostModelParams :: CostModel -> [Integer]
...
+-- an array of integers of a specific length (depending on the version of Plutus).
+-- Starting in version 9, we allow the decoders to accept lists longer than what they
+-- require, so that new fields can be added in the future.
+-- For this reason, we must hard code the length expectation into the deserializers
+-- prior to version 9.
+--
+-- Note that the number of elements in the V1 and V2 cost models
+-- may change in the future, they are only fixed prior to version 9.
+--
+-- See https://github.com/input-output-hk/cardano-ledger/issues/2902
```
teodanciu, line 580: Nice explanation! Would it be worth to reference the adr as well? https://github.com/input-output-hk/cardano-ledger/blob/master/docs/adr/2022-12-05_006-cost-model-serialization.md

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Scripts.hs
```diff
@@ -559,7 +561,43 @@ getCostModelParams :: CostModel -> [Integer]
...
+legacyCostModelLength PlutusV1 = 166
+legacyCostModelLength PlutusV2 = 175
+
+-- | See the note for 'legacyCostModelLength'.
+legacyDecodeCostModel :: Language -> Decoder s CostModel
+legacyDecodeCostModel lang = do
+  values <- fromCBOR
+  let numValues = length values
+      expectedNumValues = legacyCostModelLength lang
+  when (length values /= expectedNumValues) $
```
teodanciu: A nitpick: could use `numValues` instead of `length values`

JaredCorduan: :facepalm: that's what I meant to do! thank you for picking that nit before it became a lice infestation!!

eras/alonzo/test-suite/test/Test/Cardano/Ledger/Alonzo/Serialisation/CDDL.hs
```diff
@@ -33,6 +34,7 @@ tests n = withResource combinedCDDL (const (pure ())) $ \cddl ->
     , cddlTest @(PParamsUpdate Alonzo) (eraProtVerHigh @Alonzo) n "protocol_param_update"
     , cddlAnnotatorTest @(Redeemers Alonzo) (eraProtVerHigh @Alonzo) n "[* redeemer]"
     , cddlAnnotatorTest @(Tx Alonzo) (eraProtVerHigh @Alonzo) n "transaction"
+    , cddlTest @CostModels (eraProtVerHigh @Alonzo) n "costmdls"
```
teodanciu, line 37: Just for my own clarification, this wouldn't have failed before this change no?  Was it missing by accident?

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Scripts.hs
JaredCorduan: thanks! good idea, let me remind myself how the right way is to do that for ADRs...

teodanciu: I remember we talked about this, but helpfully, I don't remember anything about the output of the discussion! 

I see one reference here: https://github.com/input-output-hk/cardano-ledger/blob/master/libs/cardano-ledger-core/src/Cardano/Ledger/EpochBoundary.hs#L193  

```
-- See more info in the [Optimize TICKF ADR](https://github.com/input-output-hk/cardano-ledger/blob/master/docs/adr/2022-12-12_007-optimize-ledger-view.md)
```

Can't remember at all if this is what we decided  ( though It's not a terrible choice either way)

JaredCorduan: oh!! I had it backwards. I thought you meant "update the ADR to state that we need to care about backwards compatibility". Which isn't what you said at all :laughing: , you meant to leave a comment. yes, great idea!

eras/alonzo/test-suite/test/Test/Cardano/Ledger/Alonzo/Serialisation/CDDL.hs
JaredCorduan: that's correct, it would not fail before or after this change. and in fact, cost models can end up in `PParamsUpdate`, so it is not strictly necessary. but I thought it would be nice to add them for easier future-proofing and debugging.

teodanciu: Makes sense, thanks for the explanation!

teodanciu (approved):
Looks good to me

lehins (approved):
I am still a little fuzzy on how #2902 will work, but as far as this PR is concerned it looks good

PR #3277, opened 2023-02-01
enforce unique map keys in CBOR decoding, ver. 9

### LINKED ISSUE #2965: duplicate keys in CBOR map decoding
The helper function used by the ledger to deserialize CBOR maps does not reject duplicate keys as it should. Unfortunately, a hard fork is required to fix this issue.

Since we have to fix this at a hard fork, and maintain the existing serialization up until the moment of the hard fork, this is a tricky issue to fix. Especially since it involves any map used inside a transactions, and since our `To/FromCBOR` instances have no notion of versioning (we usually change serialization by introducing new types).

Our current best plan to fix this issues came from @lehins. The next time we introduce a new transaction type inside a new ledger era, the `FromCBOR` instance for the transaction can use custom deserializers for everything (ie not using the `FromCBOR` instances that are flawed).

JaredCorduan:
# Description

Starting in major version 9, we raise a CBOR decoding error if a map contains duplicate keys.

resolves #2965

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/libs/cardano-ledger-binary/src/Cardano/Ledger/Binary/Decoding/Decoder.hs
+++ b/libs/cardano-ledger-binary/src/Cardano/Ledger/Binary/Decoding/Decoder.hs
@@ -674,7 +674,11 @@ decodeMapByKey ::
 decodeMapByKey decodeKey decodeValue =
   ifDecoderVersionAtLeast
     (natVersion @2)
-    (decodeIsListByKey decodeKey decodeValue)
+    ( ifDecoderVersionAtLeast
+        (natVersion @9)
+        (decodeMapEnforceNoDuplicates decodeKey decodeValue)
+        (decodeIsListByKey decodeKey decodeValue)
+    )
     (decodeMapSkel Map.fromDistinctDescList decodeKey decodeValue)
 
 -- | Decode `VMap`. Unlike `decodeMap` it does not behavee differently for
@@ -829,6 +833,32 @@ decodeMapNoDuplicates decodeKey decodeValue =
     decodeInlinedPair = do
       !key <- decodeKey
       !value <- decodeValue
+      pure (key, value)
+
+-- | Just like `decodeMap`, but fails on duplicate keys
+decodeMapEnforceNoDuplicates ::
+  forall s k v.
+  Ord k =>
+  Decoder s k ->
+  (k -> Decoder s v) ->
+  Decoder s (Map.Map k v)
+decodeMapEnforceNoDuplicates decodeKey decodeValueFor = do
+  decodeMapLenOrIndef >>= \case
+    Just len -> loop (\x -> pure (x - 1, x <= 0)) len Map.empty
+    Nothing -> loop (\x -> (,) x <$> decodeBreakOr) () Map.empty
+  where
+    loop :: (a -> Decoder s (a, Bool)) -> a -> Map.Map k v -> Decoder s (Map.Map k v)
+    loop condition prevStep !acc = do
+      (nextStep, shouldStop) <- condition prevStep
+      if shouldStop
+        then pure acc
+        else do
+          (k, v) <- decodeInlinedPair
+          when (k `Map.member` acc) $ fail "Duplicate key detected in the Map"
+          loop condition nextStep (Map.insert k v acc)
+    decodeInlinedPair = do
+      !key <- decodeKey
+      !value <- decodeValueFor key
       pure (key, value)
 
 decodeMapContents :: Decoder s a -> Decoder s [a]
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -342,6 +342,7 @@ in the naming of release branches.
 - Fixed typo in makeHashWithExplicitProxys phantom type (indexl to index). #3072
 - Fixed the incorrect conversion of the validity interval's upper bound in `transVITime` (fixes #3043). #3200
 - Enforce that the CostModel deserializers expect a specific length prior to version 9.
+- Starting in version 9, duplicate keys in CBOR maps are not longer allowed
 
 ## Release branch 1.1.x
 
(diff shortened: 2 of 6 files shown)


libs/cardano-ledger-binary/src/Cardano/Ledger/Binary/Decoding/Decoder.hs
```diff
@@ -831,6 +835,20 @@ decodeMapNoDuplicates decodeKey decodeValue =
...
       pure (key, value)
 
+-- | Just like `decodeMap`, but fails on duplicate keys
+decodeMapEnforceNoDuplicates ::
+  Ord k => Decoder s k -> (k -> Decoder s v) -> Decoder s (Map.Map k v)
+decodeMapEnforceNoDuplicates decodeKey decodeValueFor = do
+  (numValues, xs) <- decodeCollectionWithLen decodeMapLenOrIndef decodeInlinedPair
+  unless (numValues == (length . Set.toList . fromList . (map fst)) xs) $
+    fail "Map has duplicates"
+  pure $ Map.fromList xs
```
lehins: This is way too expensive:
* construct a list
* construct another list with keys
* construct a set from a list
* construct a map from a list

There is a much better way to do this.

Construct a Map and each time a key value pair is decoded check weather it is already in a Map or not.
*Disclaimer* - In PR programming, expect some problems, but the idea should be clear.

```suggestion
  decodeMapLenOrIndef >>= \case
    Just len -> loop (\(x -> pure (x - 1, x  <= 0)) len Map.empty
    Nothing -> loop (\x -> (,) x <$> decodeBreakOr) () Map.empty
  where
    loop condition prevStep !acc = do
      (nextStep, shouldStop) <- condition prevStep
      if shouldStop
        then pure acc
        else do
          (k, v) <- decodeInlinedPair
          when (k `Map.member` acc) $ fail "Duplicate key detected in the Map"
          loop condition nextStep (Map.insert k v acc)
```

libs/cardano-ledger-binary/test/Test/Cardano/Ledger/Binary/Vintage/Failure.hs
```diff
@@ -42,6 +45,16 @@ genNegativeInteger :: Gen Integer
...
 
+-- | Generate an association list with at least one duplicate key
+genDuplicateAssocList :: Gen [(Int, Int)]
+genDuplicateAssocList = do
+  xs <- Gen.list (Range.constant 1 20) genPair -- note nonempty
+  (a, _) <- Gen.element xs -- pick a key to duplicate
+  c <- Gen.int Range.constantBounded -- pick a value for duplicate key
+  Gen.shuffle ((a, c) : xs)
+  where
+    genPair = (,) <$> (Gen.int Range.constantBounded) <*> (Gen.int Range.constantBounded)
```
lehins:
```suggestion
    genPair = (,) <$> Gen.int Range.constantBounded <*> Gen.int Range.constantBounded
```

libs/cardano-ledger-binary/test/Test/Cardano/Ledger/Binary/Vintage/Failure.hs
```diff
@@ -80,6 +93,12 @@ prop_shouldFailNegativeNatural = property $ do
   n <- forAll genNegativeInteger
   assertIsLeft (decode byronProtVer (toCBOR n) :: Either DecoderError Natural)
 
+prop_shouldFailMapWithDupKeys :: Property
+prop_shouldFailMapWithDupKeys = property $ do
+  xs <- forAll genDuplicateAssocList
+  let encodedMap = encodeMapLen (fromIntegral $ length xs) <> mconcat (toCBOR <$> xs)
```
lehins: I am not sure what's the name of function in Hedgehog, but something an equivalent to `elements` from QuickCheck can be used to switch between indefinite and exact length encoding:
```suggestion
  encodedMap <- 
    elements [ encodeMapLen (fromIntegral $ length xs) <> foldMap toCBOR xs
             , encodeMapLenIndef <> foldMap toCBOR xs <> encodeBreak ]
```

lehins (changes requested):
There is a more efficient way to implement stricter Map decoder, which I suggested in a comment.

Also, maybe you wanna try adding a quickcheck property test to hspec test suite? I think it would be a good exercise to get first hand experience with `hspec`.  This is just a suggestion, so feel free to ignore it.

Hedgehog definitely works, so the test that is already included in this PR will  work just fine, but since we mostly use QuckCheck, I thought it would be better if we stick to one tool for consistency. However, if you prefer Hedgehog there is nothing wrong with using both of the random property testing libraries.

libs/cardano-ledger-binary/src/Cardano/Ledger/Binary/Decoding/Decoder.hs
JaredCorduan: awesome, thank you!

JaredCorduan:
> Also, maybe you wanna try adding a quickcheck property test to hspec test suite? I think it would be a good exercise to get first hand experience with `hspec`. This is just a suggestion, so feel free to ignore it.

I only used Hedgehog since that was what was already being used in the `Failure` module. In general, I'd like to move completely over to QC.. I'll definitely poke around with using QuickCheck in hspec tomorrow, probably in a new test module, thanks for the motivation.

JaredCorduan:
@lehins I redid the tests in QuickCheck. let me know if this is what you had in mind.

lehins (approved):
Beautiful!!! :slightly_smiling_face:

jmhrpr:
What was the behaviour when decoding a map with duplicate keys? Like, what value for the duplicate key was used?

JaredCorduan:
> What was the behaviour when decoding a map with duplicate keys? Like, what value for the duplicate key was used?

it was overwriting (so right biased).

PR #3282, opened 2023-02-03
Integrate cardano binary

lehins:
# Description

We use versioned CBOR throughout ledger that is implemented in cardano-ledger-binary, however many downstream components do not care about the versioning, so we need to have ability for ledger to work with both `cardano-ledger-binary` and `cardano-binary`. This PR makes it possible.

* Define To/FromCBOR for all Byron types, which guarantees that they will not to change
* Define To/FromCBOR for majority of the types that are parameterized by `era` 
* MemoBytes allows to derive only the non-versioned `ToCBOR`, because the type is already pre-encoded and versioning happens on the raw type upon construction. Every `EncCBOR` for memoized type indicates with a comment that it is already pre-encoded.
* Created a nicer way to create golden tests and switched NewEpochState test to use it, which is now located in the new test suite inside `cardano-ledger-shelley`
* Restrict `MaxVersion` constraint in such a way to guarantee upgradeability

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -19,7 +19,7 @@ index-state: 2022-12-24T00:00:00Z
 index-state:
   , hackage.haskell.org 2022-12-24T00:00:00Z
 -- Bump this if you need newer packages from CHaP
-  , cardano-haskell-packages 2022-11-02T15:34:17Z
+  , cardano-haskell-packages 2023-02-21T23:27:00Z
 
 packages:
   eras/allegra/impl
@@ -77,7 +77,8 @@ test-show-details: streaming
 allow-newer:
   monoidal-containers:aeson,
   size-based:template-haskell,
-  Unique:hashable
+  Unique:hashable,
+  plutus-core:cardano-crypto-class
 
 constraints:
   -- bizarre issue: in earlier versions they define their own 'GEq', in newer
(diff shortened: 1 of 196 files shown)


cabal.project
```diff
@@ -110,3 +110,21 @@ source-repository-package
   --sha256: 18apsg2lqjv9cc29nbd3hzj2hqhksqjj0s4xp2rdv8cbd27racjh
   subdir:
     cborg
+
+source-repository-package
```
lehins: This will be removed, once releases of packages from cardano-base are made to CHaPs after this PR is approved

eras/shelley/test-suite/test/Test/Cardano/Ledger/Shelley/UnitTests.hs
```diff
@@ -710,8 +710,8 @@ testProducedOverMaxWord64 =
       st =
         runShelleyBase $
           applySTSTest @(ShelleyLEDGER C) (TRC (ledgerEnv, ledgerState, tx))
-   in -- We test that the serialization of the predicate failure does not return bottom
-      serialize' shelleyProtVer st @?= serialize' shelleyProtVer st
+   in -- We test that the predicate failure does not return bottom
+      pure $! rnf st
```
lehins, line 714: @JaredCorduan this is the proper way to check that the type does not return bottom :wink:

JaredCorduan: much better, thanks!

JaredCorduan (approved):
yet another colossal undertaking! looks great to me!

PR #3283, opened 2023-02-03
introduce flexible cost model decoding

### LINKED ISSUE #2902: Problematic cost model serialization

JaredCorduan:
# Description

Starting in version 9, `CostModels` can now be deserialized from any map of `Word8` values to lists of integers. Only valid cost models are actually converted to evaluation contexts that can be used. Errors and unrecognized language versions are stored in the `CostModels` type so that:

* they can accept cost models that they do not yet understand
* upon deserializing after a software update, new cost models are available from the prior serialization.

resolves #2902

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Scripts.hs
+++ b/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Scripts.hs
@@ -33,7 +33,10 @@ module Cardano.Ledger.Alonzo.Scripts (
 
   -- * Cost Model
   CostModel,
+  CostModelError (..),
+  emptyCostModels,
   mkCostModel,
+  mkCostModelsLenient,
   encodeCostModel,
   getCostModelLanguage,
   getCostModelParams,
@@ -41,8 +44,7 @@ module Cardano.Ledger.Alonzo.Scripts (
   ExUnits (ExUnits, exUnitsMem, exUnitsSteps, ..),
   ExUnits',
   Prices (..),
-  decodeCostModelMap,
-  decodeCostModel,
+  decodeCostModelFailHard,
   CostModels (..),
   PV1.CostModelApplyError (..),
 )
@@ -67,7 +69,6 @@ import Cardano.Ledger.Binary (
   cborError,
   decodeMapByKey,
   encodeFoldableAsDefLenList,
-  encodeMap,
   getVersion64,
   ifDecoderVersionAtLeast,
  )
@@ -85,6 +86,7 @@ import Cardano.Ledger.Binary.Version (natVersion)
 import Cardano.Ledger.Coin (Coin (..))
 import Cardano.Ledger.Core
 import Cardano.Ledger.Crypto (Crypto)
+import Cardano.Ledger.Language (mkLanguageEnum)
 import Cardano.Ledger.SafeHash (SafeToHash (..))
 import Cardano.Ledger.Shelley.Scripts (nativeMultiSigTag)
 import Cardano.Ledger.TreeDiff (Expr (App), ToExpr (..), defaultExprViaShow)
@@ -120,7 +122,7 @@ import Data.Typeable (Typeable)
 import Data.Vector as Vector (toList)
 import Data.Word (Word64, Word8)
 import GHC.Generics (Generic)
-import NoThunks.Class (InspectHeapNamed (..), NoThunks)
+import NoThunks.Class (InspectHeapNamed (..), NoThunks (..), allNoThunks)
 import Numeric.Natural (Natural)
 import PlutusCore.Evaluation.Machine.CostModelInterface (CostModelApplyWarn)
 import qualified PlutusLedgerApi.V1 as PV1 (
@@ -328,7 +330,7 @@ instance FromJSON CostModels where
   parseJSON = withObject "CostModels" $ \o -> do
     v1CostModels <- legacyParseCostModels o
     v2CostModels <- parseCostModels o
-    pure $ CostModels $ Map.fromList (v1CostModels ++ v2CostModels)
+    pure $ CostModels (Map.fromList (v1CostModels ++ v2CostModels)) mempty mempty
 
 -- | The costmodel parameters in Alonzo Genesis are represented as a map.  Plutus API does
 -- no longer require the map as a parameter to `mkEvaluationContext`, but the list of
@@ -560,12 +562,19 @@ getCostModelLanguage (CostModel lang _ _) = lang
 getCostModelParams :: CostModel -> [Integer]
 getCostModelParams (CostModel _ cm _) = cm
 
-decodeCostModelMap :: Decoder s (Map Language CostModel)
-decodeCostModelMap =
+decodeCostModelsCollectingErrors :: Decoder s CostModels
+decodeCostModelsCollectingErrors = mkCostModelsLenient <$> fromCBOR
+
+decodeCostModelsFailingOnError :: Decoder s CostModels
+decodeCostModelsFailingOnError =
+  CostModels <$> decodeMapByKey fromCBOR legacyDecodeCostModel <*> pure mempty <*> pure mempty
+
+decodeCostModels :: Decoder s CostModels
+decodeCostModels =
   ifDecoderVersionAtLeast
     (natVersion @9)
-    (decodeMapByKey fromCBOR decodeCostModel)
-    (decodeMapByKey fromCBOR legacyDecodeCostModel)
+    decodeCostModelsCollectingErrors
+    decodeCostModelsFailingOnError
 
 -- | Prior to version 9, each 'CostModel' was expected to be serialized as
 -- an array of integers of a specific length (depending on the version of Plutus).
@@ -600,8 +609,8 @@ legacyDecodeCostModel lang = do
     Left e -> fail $ show e
     Right cm -> pure cm
 
-decodeCostModel :: Language -> Decoder s CostModel
-decodeCostModel lang = do
+decodeCostModelFailHard :: Language -> Decoder s CostModel
+decodeCostModelFailHard lang = do
   checked <- mkCostModel lang <$> fromCBOR
   case checked of
     Left e -> fail $ show e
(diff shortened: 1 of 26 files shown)


eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Scripts.hs
```diff
@@ -610,21 +619,85 @@ decodeCostModel lang = do
...
   deriving stock (Eq, Show, Ord, Generic)
-  deriving newtype (NFData, NoThunks)
+
+emptyCostModels :: CostModels
+emptyCostModels = CostModels mempty mempty mempty
+
+-- | This function attempts to add a new cost model to a given 'CostModels'.
+-- If it is a valid cost model for a known version of Plutus, it is added to
+-- 'validCostModels'. If it is an invalid cost model for a known version of Plutus,
+-- the error is storted in 'costModelErrors' and the cost model is stored in
```
teodanciu: `stored` instead of `storted`? 
However: can we make `storted` a thing? to mean stored in a sorted way :grin:

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Scripts.hs
```diff
@@ -33,16 +33,18 @@ module Cardano.Ledger.Alonzo.Scripts (
 
   -- * Cost Model
   CostModel,
+  CostModelError (..),
+  emptyCostModels,
   mkCostModel,
+  mkCostModels,
```
teodanciu: A pity we have to expose both, especially since `mkCostModel` is meant to be the "smart constructor" (I think it's written in some doc). Now we have an even smarter one, a genius constructor :grin: 
But we need them in tests, which ultimately is a good thing!

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Scripts.hs
```diff
@@ -610,21 +619,85 @@ decodeCostModel lang = do
...
+-- | For a known version of Plutus, attempting to construct a cost model with
+-- too few parameters (depending on the version) will result in an error.
+-- 'CostModelApplyError' exists to collect these errors in the 'CostModels' type.
+-- The 'CostModels' type itself needs to be flexible enough to accept any map
+-- of 'Word8' to '[Integer]', so that cost models can be placed in the protocol parameters
+-- ahead of changes to the Plutus evaluation context. In this way, serializing a cost model,
+-- updating software, and deserializing can result in errors going away.
+--
+-- Additionally, 'CostModels' needs to be able to store cost models for future version
+-- of Plutus, which we cannot yet even validate. These are stored in 'invalidCostModels'.
```
teodanciu, line 655: Great and useful explanation! Would it be better positioned on top of `data CostModels` just below this line, rather than it is now on the `instance NFData` ? 
Not sure how the haddock looks if it's here

teodanciu (approved):
Looks good to me.

The `Scripts` module is becoming a big unwieldy and losing  cohesion, so maybe  if we need to add stuff to `CostModel` we can make it a separate module.
Just some other small remarks :+1:

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Scripts.hs
JaredCorduan: yea, can you think of anything better to do?

maybe I just need more documentation to make it clear when you might use each one? `mkCostModel` is when you are _serious_ :neutral_face: about needing a given cost model. and `mkCostModels` is much more laid back :relaxed:, happy to take the good things and recording the bad things for future contemplation.

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Scripts.hs
JaredCorduan: :rofl: :facepalm: :rofl: 

good catch, thank you!

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Scripts.hs
lehins: Maybe call it something like `mkLenientCostModels` or `mkCostModelsLenient`

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Scripts.hs
```diff
@@ -560,12 +562,19 @@ getCostModelLanguage (CostModel lang _ _) = lang
...
 getCostModelParams (CostModel _ cm _) = cm
 
-decodeCostModelMap :: Decoder s (Map Language CostModel)
-decodeCostModelMap =
+decodeCostModelsCollectingErrors :: Decoder s CostModels
+decodeCostModelsCollectingErrors = mkCostModels <$> fromCBOR
+
+decodeCostModelsFailingOnError :: Decoder s CostModels
+decodeCostModelsFailingOnError =
+  CostModels <$> (decodeMapByKey fromCBOR legacyDecodeCostModel) <*> (pure mempty) <*> (pure mempty)
```
```suggestion
  CostModels <$> decodeMapByKey fromCBOR legacyDecodeCostModel <*> pure mempty <*> pure mempty
```

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Scripts.hs
```diff
@@ -610,21 +619,85 @@ decodeCostModel lang = do
...
+-- updating software, and deserializing can result in errors going away.
+--
+-- Additionally, 'CostModels' needs to be able to store cost models for future version
+-- of Plutus, which we cannot yet even validate. These are stored in 'invalidCostModels'.
+instance NFData CostModelError
+
+data CostModels = CostModels
+  { validCostModels :: Map Language CostModel
+  , costModelErrors :: Map Language CostModelError
+  , invalidCostModels :: Map Word8 [Integer]
```
lehins: Any reason not to make 'em strict? Also in order to follow convention and keep things consistent maybe renaming them would be a good idea? Maybe calling the "invalid" ones "unknown" instead would be better? you already call them as such in tests, which is where I got this suggestion from.
```suggestion
  { costModelsValid :: !(Map Language CostModel)
  , costModelsErrors :: !(Map Language CostModelError)
  , costModelsUnknown :: !(Map Word8 [Integer])
```

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Scripts.hs
```diff
@@ -610,21 +619,85 @@ decodeCostModel lang = do
 getEvaluationContext :: CostModel -> PV1.EvaluationContext
 getEvaluationContext (CostModel _ _ ec) = ec
 
-newtype CostModels = CostModels {unCostModels :: Map Language CostModel}
+-- | See 'CostModels' for an explanation of how 'CostModelError' is used.
+--
+-- It would be preferable to store the 'PV1.CostModelApplyError' type itself,
+-- but there is currently no 'Eq' instance for it, which we require.
+data CostModelError = CostModelError String
+  deriving stock (Eq, Show, Ord, Generic)
```
lehins: Why not create an `Eq` instance that uses the `Show` instance now and add the `Eq` instance upstream later? 

This way we would give us the correct interface right now, but it would be slightly inefficient for the unusual case, until the issue is fixed upstream in Plutus.

So something like:

```suggestion
newtype CostModelError = CostModelError PV1.CostModelApplyError
  deriving stock (Show, Generic)

instance Eq CostModelError where
  CostModelError x1 == CostModelError x2 = show x1 == show x2
```

It looks a bit hacky, but semantically it is exactly the same as the current implementation. At least it is more obvious that it is hacky :smile: 

Related note. Plutus should export `CostModelApplyError` from `PlutusLedgerApi`, since it is shared between versions.

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Scripts.hs
```diff
@@ -610,21 +619,85 @@ decodeCostModel lang = do
...
+emptyCostModels = CostModels mempty mempty mempty
+
+-- | This function attempts to add a new cost model to a given 'CostModels'.
+-- If it is a valid cost model for a known version of Plutus, it is added to
+-- 'validCostModels'. If it is an invalid cost model for a known version of Plutus,
+-- the error is storted in 'costModelErrors' and the cost model is stored in
+-- 'invalidCostModels'. Lastly, if the Plutus version is unknown,
+-- the cost model is also stored in 'invalidCostModels'.
+addRawCostModel :: Word8 -> [Integer] -> CostModels -> CostModels
+addRawCostModel lang params (CostModels validCMs errs invalidCMs) =
```
lehins: Lets call them something different. There are too many params in this repo. Also instead of using ticks, how about adding `8` or `W8` to the lang to indicate that it is not `Language`, but an 8bit numeric argument?
```suggestion
addRawCostModel langW8 cmIds (CostModels validCMs errs invalidCMs) =
```

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Scripts.hs
```diff
@@ -610,21 +619,85 @@ decodeCostModel lang = do
...
+  if minLang <= lang && lang <= maxLang
+    then case mkCostModel lang' params of
+      Right cm -> CostModels (Map.insert lang' cm validCMs) errs invalidCMs
+      Left e -> CostModels validCMs (addError lang' e errs) invalidCMs'
+    else CostModels validCMs errs invalidCMs'
+  where
+    minLang = fromIntegral . fromEnum $ (minBound :: Language)
+    maxLang = fromIntegral . fromEnum $ (maxBound :: Language)
+    lang' = toEnum . fromIntegral $ lang
```
lehins: Lets add a function to the `Cardano.Ledger.Language`:

```haskell
-- | Make a language form its `Enum` index.
mkLanguageEnum :: Int -> Maybe Language
mkLanguageEnum iLang
  | minLang <= iLang && iLang <= maxLang = Just $ toEnum lang
  | otherwise = Nothing
  where
    minLang = fromEnum (minBound :: Language)
    maxLang = fromEnum (maxBound :: Language)
```

This way the partial `toEnum` is contained near the definition of the `Language` and we get a reusable function that might come in handy elsewhere.

```suggestion
  case mkLanguageEnum (fromIntegral lang8) of
    Just lang ->
      then case mkCostModel lang params of
        Right cm -> CostModels (Map.insert lang cm validCMs) errs invalidCMs
        Left e -> CostModels validCMs (addError lang e errs) invalidCMs'
    Nothing -> CostModels validCMs errs invalidCMs'
```

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Scripts.hs
```diff
@@ -610,21 +619,85 @@ decodeCostModel lang = do
...
+
+-- | Turn a 'CostModels' into a mapping of potential language versions and
+-- cost model values, with no distinction between valid and invalid cost models.
+-- This is used for serialization, so that judgements about validity can be made
+-- upon deserialization.
+flattenCostModel :: CostModels -> Map Word8 [Integer]
+flattenCostModel (CostModels validCMs _ invalidCMs) = validCMs' `Map.union` invalidCMs
+  where
+    validCMs' = Map.fromList $ map f (Map.toList validCMs)
+    f (lang, cm) = (fromIntegral . fromEnum $ lang, cmMap cm)
```
lehins: We can avoid creating intermediate lists with a fold:
```suggestion
flattenCostModel (CostModels validCMs _ invalidCMs) =
  Map.foldrWithKey (\lang cm -> Map.insert (fromIntegral $ fromEnum lang) (cmMap cm)) invalidCMs validCMs
```

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Scripts.hs
```diff
@@ -610,21 +619,85 @@ decodeCostModel lang = do
...
+
+instance NFData CostModels
 
 instance FromCBOR CostModels where
-  fromCBOR = CostModels <$> decodeCostModelMap
+  fromCBOR = decodeCostModels
 
 instance ToCBOR CostModels where
-  toCBOR = encodeMap toCBOR encodeCostModel . unCostModels
+  toCBOR cm = encodeMap toCBOR toCBOR (flattenCostModel cm)
```
lehins: That's just
```suggestion
  toCBOR = toCBOR . flattenCostModel
```

eras/alonzo/test-suite/src/Test/Cardano/Ledger/Alonzo/Serialisation/Generators.hs
```diff
@@ -233,19 +236,66 @@ instance Arbitrary Language where
...
+listAtLeast x = do
+  y <- getNonNegative <$> arbitrary
+  replicateM (x + y) arbitrary
+
+genCostModelValues :: Language -> Gen (Word8, [Integer])
+genCostModelValues lang =
+  oneof
+    [ (lang',) <$> listAtLeast (costModelParamsCount lang) -- Valid Cost Model for known language
+    , (lang',) <$> take tooFew <$> arbitrary -- Invalid Cost Model for known language
+    ]
```
lehins:
```suggestion
  (lang',) <$> oneof
    [ listAtLeast (costModelParamsCount lang) -- Valid Cost Model for known language
    , take tooFew <$> arbitrary -- Invalid Cost Model for known language
    ]
```

eras/conway/test-suite/cddl-files/conway.cddl
```diff
@@ -366,10 +366,17 @@ ex_unit_prices =
 language = 0 ; Plutus v1
          / 1 ; Plutus v2
 
-costmdls =
-  { ? 0 : [ 166* int ] ; Plutus v1, only 166 integers are used, but more are accepted (and ignored)
-  , ? 1 : [ 175* int ] ; Plutus v2, only 175 integers are used, but more are accepted (and ignored)
-  }
+potential_languages = 0 .. 255
+
+; The format for costmdls is flexible enough to allow adding Plutus builtins and languge
```
```suggestion
; The format for costmdls is flexible enough to allow adding Plutus built-ins and language
```

libs/cardano-ledger-binary/testlib/Test/Cardano/Ledger/Binary/RoundTrip.hs
```diff
@@ -104,10 +106,18 @@ roundTripExpectation ::
...
-      Right tDecoded -> tDecoded `shouldBe` t
+  forM_ [natVersion @2 .. maxBound] (roundTripExpectation' trip t)
+
+roundTripExpectation' ::
+  (Show t, Eq t, Typeable t, HasCallStack) =>
+  Trip t t ->
+  t ->
+  Version ->
+  Expectation
+roundTripExpectation' trip t version =
```
lehins: Please, no ticks :pray: 

```suggestion
-- | Test roundtripping with exact version
roundTripExactExpectation ::
  (Show t, Eq t, Typeable t, HasCallStack) =>
  Trip t t ->
  t ->
  Version ->
  Expectation
roundTripExactExpectation trip t version =
```

libs/cardano-ledger-binary/testlib/Test/Cardano/Ledger/Binary/RoundTrip.hs
```diff
@@ -116,6 +126,14 @@ roundTripCborExpectation ::
   Expectation
 roundTripCborExpectation = roundTripExpectation (cborTrip @t)
 
+roundTripCborExpectation' ::
+  forall t.
+  (Show t, Eq t, ToCBOR t, FromCBOR t, HasCallStack) =>
+  Version ->
+  t ->
+  Expectation
+roundTripCborExpectation' v t = roundTripExpectation' (cborTrip @t) t v
```
lehins: Here too :pray: 
```suggestion
roundTripCborExactExpectation ::
  forall t.
  (Show t, Eq t, ToCBOR t, FromCBOR t, HasCallStack) =>
  Version ->
  t ->
  Expectation
roundTripCborExactExpectation v t = roundTripExactExpectation (cborTrip @t) t v
```

lehins (approved):
This is really awesome!!!

It is only after going through this PR that I was able to convince myself that this is a really good approach to deal with forward compatibility of CostModels. I could not think of anything that could mess it up. Very nice!

I have a few suggestions, but nothing that would block the PR.

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Scripts.hs
JaredCorduan: great idea, I like that much better!

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Scripts.hs
JaredCorduan: ah yea, I went back and forth a bit on the name. In the end I decided "unknown" was better, good catch!

and yea, strict is better, I've gotten out of the habit it seems...

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Scripts.hs
JaredCorduan: great idea!

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Scripts.hs
JaredCorduan: fantastic :)

eras/alonzo/test-suite/src/Test/Cardano/Ledger/Alonzo/Serialisation/Generators.hs
JaredCorduan: much nicer!

libs/cardano-ledger-binary/testlib/Test/Cardano/Ledger/Binary/RoundTrip.hs
JaredCorduan: yea, I guess it's a bad habit. `roundTripExactExpectation` is much clearer :)

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Scripts.hs
JaredCorduan: @lehins I tried it out, looks like we will also need `Ord`, `NoThunks`, `NFData`, and `ToExpr`.  or really a `Generic` and the last three are trivial. probably still worth it, yea?

lehins: All of them are doable. `Ord` and `ToExpr` can be done with `Show`. `NFData`, `NoThunks` can be either defined manually or with standalone derived `Generic` instance. Important part is that the constructors for the type are available, so we can do all that we need to do in order to get it right, including temporarily introducing any orphan instances for the type if need be, although only if with intention to upstream them into plutus

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Scripts.hs
JaredCorduan: sounds good to me!

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Scripts.hs
JaredCorduan: ok, I used `allNoThunks []`. hopefully that is not too much of a hack.

PR #3290, opened 2023-02-09
clarifying the script integrity hash

JaredCorduan:
# Description

Clarifying the script integrity hash in the CDDL description:
* remove the outer square brackets so as to not imply that the bytes are inside of a CBOR array
* explicitly describe in prose that we are concatenating
* use the notation from cryptography of double pipe for concatenation
* make it clear that if there are no datums, that "empty string" here means the empty bytestring, ie that it is omitted.

# Checklist

- [x] Self-reviewed the diff


AndrewWestberg (approved):

teodanciu (approved):
Looks good to me

JaredCorduan:
The CI failure was the inability to generate a sufficiently long trace for a `small-step-semantics` test. I am unworried about this and will re-run.

PR #3291, opened 2023-02-10
Implemented `RATIFY` and `ENACT`

Soupstraw:
# Description

This PR brings the Conway era up to date with the current specifications. 

The `ENACTMENT` rule was renamed to `ENACT` to better reflect the spec.

The `NEWEPOCH` rule now calls `RATIFY`, which in turn calls `ENACT`. `RATIFY` is a new rule that is run at the epoch boundary and it loops over all active governance actions to decide whether it can be enacted. `ENACT` currently supports all governance actions except treasury payouts.

`GovernanceActionInfo` was removed since it's not in the spec and is very similar to `GovernanceActionState`. The `Vote` datatype was removed and that data is now kept directly in `GovernanceProcedure`. Votes and governance actions are now stored in `TxBody` as `GovernanceProcedure`s.

Governance now uses credentials instead of key hashes for voting. This is different from the current spec, but makes sense if we want to use staking credentials for voting. The alternative would be to only allow staking credentials that use key hashes to participate in voting, filtering out any staking credentials that use Plutus scripts.

Three new governance actions were added:
 * `NoConfidence`
 * `NewCommittee`
 * `NewConstitution`

# Checklist

- [x] Self-reviewed the diff


eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Ratify.hs
```diff
@@ -0,0 +1,142 @@
...
+  type Environment (ConwayRATIFY era) = RatifyEnv era
+  type PredicateFailure (ConwayRATIFY era) = Void
+  type Signal (ConwayRATIFY era) = RatifySignal era
+  type State (ConwayRATIFY era) = RatifyState era
+  type BaseM (ConwayRATIFY era) = ShelleyBase
+
+  initialRules = []
+  transitionRules = [ratifyTransition]
+
+--- Constants
```
JaredCorduan, line 76: I guess we are just hard-coding these temporarily, until we add them to the thresholds?

eras/conway/test-suite/cddl-files/conway.cddl
```diff
@@ -51,51 +51,94 @@ major_protocol_version = 1..next_major_protocol_version
...
-  , ? 20 : [* vote]                   ; New; votes
+  , ? 16 : transaction_output       ; collateral return
+  , ? 17 : coin                     ; total collateral
+  , ? 18 : set<transaction_input>   ; reference inputs
+  , ? 19 : [* governance_procedure] ; New; governance procedures
   }
 
-governance_action_info =
-  [ deposit_amount    : coin
+governance_action_state =
```
JaredCorduan, line 73: I might be forgetting something, why do we have the `governance_action_state` in the wire spec? it's not anywhere in a transaction, I don't think users will care about it...

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Enact.hs
```diff
@@ -0,0 +1,55 @@
...
+  type BaseM (ConwayENACT era) = ShelleyBase
+
+  initialRules = []
+  transitionRules = [enactmentTransition]
+
+enactmentTransition :: EraPParams era => TransitionRule (ConwayENACT era)
+enactmentTransition = do
+  TRC ((), st, act) <- judgmentContext
+  case act of
+    ParameterChange ppup -> pure $ st {ensPParams = newPP}
```
JaredCorduan, line 48: I guess the new enactment logic that checks a hash is not yet in yet?

JaredCorduan (approved):
Looks good to me! I didn't do a close comparison with the spec, I figure we will do that later when things settle down. thanks!

CHANGELOG.md
```diff
@@ -102,8 +103,10 @@ Changes recorded below are for all of the package versions in the list:
 - Added new governance related data types to Conway: #3216
   - `GovernanceActionState`
   - `GovernanceProcedure`
+  - `GovernanceMetadata`
   - `TallyEnv`
-
+  - `RatifyEnv`
+  - `RatifySignal`
```
lehins, line 109: All of this changes now need to go into the `era/conway/impl/CHANGELOG.md` file, since we switched to PVP for versioning

eras/conway/impl/src/Cardano/Ledger/Conway/Governance.hs
```diff
@@ -254,46 +187,162 @@ instance EraPParams era => NoThunks (GovernanceActionState era)
...
+
+data ConwayGovernance era = ConwayGovernance
+  { cgTally :: !(ConwayTallyState era)
+  , cgRatify :: !(RatifyState era)
+  , cgVoterRoles :: !(Map (Credential 'Voting (EraCrypto era)) VoterRole)
+  }
+  deriving (Generic, Eq, Show)
+
+cgTallyL :: Lens' (ConwayGovernance era) (ConwayTallyState era)
+cgTallyL = lens cgTally (\x y -> x {cgTally = y})
```
lehins, line 324: Why only one lens? Might as well define for the other two fields: `cgRatify` and `cgVoterRoles`.

eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState/Types.hs
```diff
@@ -408,6 +420,9 @@ data LedgerState era = LedgerState
   }
   deriving (Generic)
 
+lsUTxOStateL :: Lens' (LedgerState era) (UTxOState era)
+lsUTxOStateL = lens lsUTxOState (\x y -> x {lsUTxOState = y})
```
lehins, line 444: Same here. Defining only one lens selectively just for one field feels incomplete. If the goal is to define lenses for all of the types in this module, might as well make an extra step and at least define for all fields for the type

libs/cardano-ledger-pretty/src/Cardano/Ledger/Pretty/Conway.hs
```diff
@@ -203,12 +176,53 @@ instance PrettyA GovernanceActionIx where
 instance PrettyA VoterRole where
   prettyA = ppString . show
 
+instance PrettyA (GovernanceMetadata era) where
+  prettyA = viaShow
+
 instance PrettyA (PParamsUpdate era) => PrettyA (GovernanceActionState era) where
-  prettyA GovernanceActionState {..} =
+  prettyA (GovernanceActionState vs d ra act en) =
```
lehins: It is good that you want to protect against forgetting a field when one is added, but you don't need to loose readability of `RecordWildCards`. In fact it is better to retain RWC, because renaming fields should also be reflected in pretty and compiler will notify when that happens, so you can do this instead:
```suggestion
  prettyA gas@(GovernanceActionState _ _ _ _ _) =
    let GovernanceActionState {..} = gas -- guard against addition or removal of fields
    in
```

eras/shelley/impl/src/Cardano/Ledger/Shelley/Governance.hs
```diff
@@ -45,6 +45,7 @@ class
   type GovernanceState era = (r :: Type) | r -> era
 
   emptyGovernanceState :: GovernanceState era
+  emptyGovernanceState = def
```
lehins, line 48: Nice. The line no 57 can be removed
```diff
-emptyGovernanceState = ShelleyPPUPState emptyPPPUpdates emptyPPPUpdates
```

lehins (changes requested):
Looks great. One thing that definitely needs adjusting is the CHANGELOG, since the list that is been modified in this PR is for  the already released version. 

We now need to do proper PVP versioning for all our packages, otherwise we will end up in a dependency hell. So few things need adjustment:

* a minor version bump for `cardano-ledger-shelley`
* changelog entries for the newly added lenses in `cardano-ledger-shelley`
* a major version bump is needed for `cardano-ledger-conway` and a lower bound update on the `cardano-ledger-shelley`
* `cardano-ledger-pretty` needs a version bump and an upper bound change on `cardano-ledger-conway`

CHANGELOG.md
lehins: Could you please remove the changes in this file?

This changelog is for the previous release process and we should not change it

eras/conway/test-suite/src/Test/Cardano/Ledger/Conway/Serialisation/Generators.hs
```diff
@@ -120,8 +102,6 @@ instance
       <*> arbitrary
       <*> arbitrary
       <*> arbitrary
-      -- FIXME: For now this is turned off, rountrip tests in consensus are failing
```
lehins, line 123: This comment  is still relevant, could you please revert it?

libs/cardano-ledger-pretty/cardano-ledger-pretty.cabal
```diff
@@ -39,7 +39,7 @@ library
         cardano-ledger-alonzo >=1.0,
         cardano-ledger-babbage >=1.0,
         cardano-ledger-byron,
-        cardano-ledger-conway >=1.0,
+        cardano-ledger-conway >=1.0.1,
```
lehins:
```suggestion
        cardano-ledger-conway >=1.1,
```

libs/cardano-ledger-pretty/src/Cardano/Ledger/Pretty.hs
```diff
@@ -1,11 +1,14 @@
+{-# LANGUAGE AllowAmbiguousTypes #-}
```
lehins: I don't think this extension is needed
```suggestion
```

libs/cardano-ledger-pretty/src/Cardano/Ledger/Pretty.hs
```diff
@@ -1,11 +1,14 @@
...
 {-# LANGUAGE ConstraintKinds #-}
 {-# LANGUAGE DataKinds #-}
+{-# LANGUAGE DefaultSignatures #-}
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE FlexibleInstances #-}
 {-# LANGUAGE GADTs #-}
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE TypeApplications #-}
+{-# LANGUAGE TypeOperators #-}
```
lehins: This one too
```suggestion
```

libs/cardano-ledger-pretty/src/Cardano/Ledger/Pretty.hs
```diff
@@ -1,11 +1,14 @@
+{-# LANGUAGE AllowAmbiguousTypes #-}
 {-# LANGUAGE ConstraintKinds #-}
 {-# LANGUAGE DataKinds #-}
+{-# LANGUAGE DefaultSignatures #-}
```
lehins: So is this one
```suggestion
```

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/Same.hs
```diff
@@ -47,6 +47,7 @@ import Cardano.Ledger.Shelley.LedgerState (
   UTxOState (..),
  )
 import Cardano.Ledger.Shelley.PParams (ProposedPPUpdates (..))
+import Cardano.Ledger.Shelley.Translation ()
```
lehins: This is a redundant import
```suggestion
```

lehins (approved):
This is great. Just a few more minor fixes and this is good to go

JaredCorduan:
:tada:

PR #3294, opened 2023-02-13
Recompute stake distribution in Conway translation

teodanciu:
# Description

Since we [stop resolving pointer addresses](https://github.com/input-output-hk/cardano-ledger/pull/3253) starting with Conway, we also have to adjust the stake within the UTxOState. 

# Checklist

- [ ] Code is formatted with [`fourmolu`](https://github.com/fourmolu/fourmolu) (which can be run with `scripts/fourmolize.sh`)
- [x] Self-reviewed the diff

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -381,6 +381,7 @@ BabbageEraTxBody --> AlonzoEraTxBody --> ....
                 `--> BabbageEraTxOut -> AlonzoEraTxOut -->
 ```
 ### Changed
+- Updated `Translation.hs` in Conway to recompute the stake distribution
 - Introduced a new switch in `HardForks` that turns off pointer address resolution in `IncrementalStake` starting with version 9
 - Some types have been moved:
   - The `WitVKey` type has been moved into its own module in core.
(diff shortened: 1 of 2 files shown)


JaredCorduan:
@teodanciu this is totally my fault (for having you start on this before I was totally sure about all of the details in consensus), but I no long think that we can get away with something this simple. The trouble is that if we do it this way, we will pay the cost of this computation for _every leadership check_ crossing into the conway era until the point that the first block is accepted. Prior to the incremental stake distribution computation that we now enjoy, this was what every epoch boundary looked like, so it might appear that this is not a terrible option. however, the ledger state is much bigger today than it was before the incremental calculation, and the old way could be pretty detrimental.

So we need to rethink this. One thing we could do is reuse the trick in [ADR-7](https://github.com/input-output-hk/cardano-ledger/blob/6825d3b42fa59d26ad6a8a746b066939d3084db4/docs/adr/2022-12-12_007-optimize-ledger-view.md), namely create a thunk for the full per-stake-credential distribution as a new field in the `SnapShots`. (This does leave a sort of ugly wart, sadly, though perhaps we could clean it up after-the-fact, which would then only effect replay).

I'm open to thoughts and suggestions!

teodanciu:
Closing for now to remove PR Noise, will reopen with the better solution

PR #3295, opened 2023-02-13
Fixed that the preservation of Ada proof was missing

WhatisRT:
# Description

The latex file was there, it was just not included. I thought I'd fixed that years ago, but it seems that I've never committed the fix.


teodanciu (approved):
Built the doc out of curiosity, looks  good to me.
Maybe we should [make a new doc release](https://github.com/input-output-hk/cardano-ledger/blob/master/CONTRIBUTING.md#publishing-specifications) ?

JaredCorduan:
> Built the doc out of curiosity, looks good to me.
> Maybe we should [make a new doc release](https://github.com/input-output-hk/cardano-ledger/blob/master/CONTRIBUTING.md#publishing-specifications) ?

great idea!

PR #3296, opened 2023-02-14
Nightly builds  and tests reorganization

### LINKED ISSUE #3246: Recurring GithubActions instead of Nightly BuildKite job
GithubActions apparently have nowadays [recurring jobs](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#schedule). Managing separate configuration and access control for BuildKite does not make sense.

Setup the daily action to run somewhere around 6am UTC. I feel like it is late enough for America, but early enough for Europe.

Note that there is a separate flag ` --scenario=Nightly` for daily builds

teodanciu:
# Description

The main goal of this PR is to have nightly builds running as a github action. 
Choosing the nightly vs the default set of tests is happening based on the environment variable NIGHTLY. This is being set in the github action whenever the build is triggered by the scheduler (as opposed to by pushing ). 
Additionally, I simplified and reorganized some test code,  in order to make it easier to understand which tests are being run from the entry point, removing aggregations and indirections  that seemed arbitrary. 

Fixes #3246

# Checklist

- [x] Self-reviewed the diff


JaredCorduan (approved):
looks great! thank you for making things nicer and more organized!

.github/workflows/haskell.yml
```diff
@@ -5,6 +5,9 @@ on:
     branches: [ "master" ]
   pull_request:
     branches: [ "**" ]
+  schedule:
+    # Every night at 02:00 UTC
+    - cron: '0 2 * * *'
```
lehins: We figured 6am UTC is better for everyone on the team, since no-one is around during that time. See https://github.com/input-output-hk/cardano-ledger/issues/3246
```suggestion
    # Every night at 06:00 UTC
    - cron: '0 6 * * *'
```

lehins (approved):
This is perfect! Thank you for cleaning it up!

.github/workflows/haskell.yml
teodanciu: Ok., but that would make them matinal builds instead of nightly :laughing:

lehins: It would for London :stuck_out_tongue_winking_eye:

JaredCorduan: let's call them happy hour tests. it's five o'clock somewhere! :beers:

lehins: :rofl: That's awesome

PR #3288, opened 2023-02-08
Shelley arbitrary

lehins:
# Description

This PR starts a new `testlib` for shelley and moves all the arbitrary instances into the testlib. Also it fixes two minor problems with constraints 

# Checklist

- [x] Commit sequence broadly makes sense
- [x] Commits have useful messages
- [x] New tests are added if needed and existing tests are updated
- [x] Any changes are noted in the [changelog](https://github.com/input-output-hk/cardano-ledger/blob/master/CHANGELOG.md)
- [x] Code is formatted with [`fourmolu`](https://github.com/fourmolu/fourmolu) (which can be run with `scripts/fourmolize.sh`
- [x] Self-reviewed the diff

PR #3297, opened 2023-02-14
Collection of small serialization related fixes and improvements

lehins:
# Description

All fixes and improvements are separated into isolated commits. For some it might be easier to review one commit at a time.

Fixes:
* Fix `ToCBOR`/`FromCBOR` insatance for `PParams Babbage`. Same fix as it was done for `PParams Alonzo` in #3288 in the ae323208d6d1252e26bfe741d55a80b9297314e8 commit
* Fix `Show` instance for `PParams`/`PParamsUpdate`
* Make golden tests for Mary to use an example `MultiAsset` for the mint field of a `TxBody`, it was empty prior to this PR and it was `Coin 10` up to the latest release :open_mouth: 
* Make `ByronEra` type span both protocol versions `0` and `1`. Learned it from a comment in ouroboros-consensus repo.

Improvements:
* Addition of `encodeWithOrigin` and `decodeWithOrigin`.
* Change `mkVersion` to accept any `Integral`
* Add `getVersion` function that returns an `Integral`. This is a more general version of `getVersion64`.
* Improve error reporting during `CBORBytes` display. Prior to this PR any malformed CBOR would result in an exception, now it will try its best to show the malformed input. very helpful during test failure debugging

# Checklist

- [x] Self-reviewed the diff


eras/babbage/impl/src/Cardano/Ledger/Babbage/PParams.hs
```diff
@@ -239,7 +237,7 @@ instance Era era => ToCBOR (BabbagePParams Identity era) where
           !> To bppA0
           !> To bppRho
           !> To bppTau
-          !> E toCBORGroup bppProtocolVersion
+          !> To bppProtocolVersion
```
JaredCorduan, line 240: to make sure I am following:

we want `[protocol_version]` on the wire, to be consistent with how the updates are sent on the wire. and `E toCBORGroup` inlines the two numbers. yea?

eras/shelley-ma/test-suite/src/Test/Cardano/Ledger/Mary/Examples/Consensus.hs
```diff
@@ -22,17 +24,16 @@ ledgerExamplesMary =
...
-  Int ->
-  MaryValue c
-exampleMultiAssetValue x =
-  MaryValue 100 $ MultiAsset (Map.singleton policyId $ Map.singleton couttsCoin 1000)
+exampleMultiAssetValue :: Crypto c => Int -> MaryValue c
+exampleMultiAssetValue x = MaryValue 100 $ exampleMultiAsset x
+
+exampleMultiAsset :: forall c. Crypto c => Int -> MultiAsset c
+exampleMultiAsset x =
+  MultiAsset (Map.singleton policyId $ Map.singleton couttsCoin 1000)
```
JaredCorduan, line 36: :relaxed:

libs/cardano-ledger-binary/src/Cardano/Ledger/Binary/Version.hs
```diff
@@ -45,7 +45,7 @@ newtype Version = Version Word64
   deriving (Eq, Ord, Show, Enum, NFData, NoThunks)
 
 -- | Minimum supported version
-type MinVersion = 1
+type MinVersion = 0
```
JaredCorduan, line 48: ~we need to support Byron for something, I guess?~ we always wanted byron. but byron had two majors.

JaredCorduan (approved):
LGTM!

eras/babbage/impl/src/Cardano/Ledger/Babbage/PParams.hs
lehins: Correct

lehins: It inlines them, but the total number of elements in the list is not updated, thus causing it to create an invalid CBOR object, because the number of fields no longer matches the declared size. Luckily this only affects the locally stored ledger state

JaredCorduan: great, thank you for the sanity check!

PR #3298, opened 2023-02-14
Switch BHeader memoization to strict ByteString, but keep the field lazy

lehins:
# Description

It is wrong to have a field used for memoizing the binary representation of the type to be strict, even if the lazy bytestring is used for memoization.  It is wrong, because evaluating lazy ByteString to WHNF will force the first chunk of the ByteString, thus  causing the first 1KiB of a Block to be immediately serialized upon construction of the BHeader

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/libs/cardano-protocol-tpraos/src/Cardano/Protocol/TPraos/BHeader.hs
+++ b/libs/cardano-protocol-tpraos/src/Cardano/Protocol/TPraos/BHeader.hs
@@ -74,11 +74,10 @@ import Cardano.Ledger.Binary (
   peekTokenType,
   runByteBuilder,
   serialize',
-  serializeEncoding,
   szCases,
   withWordSize,
  )
-import qualified Cardano.Ledger.Crypto as CC
+import Cardano.Ledger.Crypto (Crypto)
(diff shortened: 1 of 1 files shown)


libs/cardano-protocol-tpraos/src/Cardano/Protocol/TPraos/BHeader.hs
```diff
@@ -291,37 +291,32 @@ pattern BHeader bHeaderBody' bHeaderSig' <-
...
-  CC.Crypto c =>
-  FromCBOR (Annotator (BHeader c))
-  where
+instance Crypto c => FromCBOR (Annotator (BHeader c)) where
   fromCBOR = annotatorSlice $
     decodeRecordNamed "Header" (const 2) $ do
       bhb <- fromCBOR
       sig <- decodeSignedKES
-      pure $ pure $ BHeader' bhb sig
+      pure $ pure $ BHeader' bhb sig . BSL.toStrict
```
JaredCorduan, line 315: so pure it floats! :soap:

JaredCorduan (approved):
LGTM!

PR #3299, opened 2023-02-14
Update `iohk-nix` in order to get the latest `libsodium`

lehins:
# Description

This brings in the new libsodium version that will be required for the Batch VRF functionality. This change alone makes ledger compatible with current cardano-base master and of course backwards compatible with older versions of cardano-base.

I am submitting it as a separate PR because it might cause some breakage locally for developers, which will require purging of `~/.cabal/store` and `dist-newstyle`

# Checklist

- [ ] Code is formatted with [`fourmolu`](https://github.com/fourmolu/fourmolu) (which can be run with `scripts/fourmolize.sh`)
- [ ] Cabal files are formatted (which can be done with `scripts/cabal-format.sh`)
- [x] Self-reviewed the diff

### DIFF
--- a/.github/workflows/haskell.yml
+++ b/.github/workflows/haskell.yml
@@ -12,7 +12,7 @@ jobs:
 
     env:
       # Modify this value to "invalidate" the cabal cache.
-      CABAL_CACHE_VERSION: "2022-12-30"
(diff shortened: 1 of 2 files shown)


JaredCorduan (approved):
:+1:

PR #3303, opened 2023-02-15
Rename `To/FromCBOR` -> `Enc/DecCBOR`

lehins:
# Description

* Rename `ToCBOR` -> `EncCBOR` and `FromCBOR` -> `DecCBOR`
* Rename `FromSharedCBOR` -> `DecShareCBOR` and all the related functions:
    
    * `fromSharedCBOR` -> `decShareCBOR`
    * `fromSharedPlusCBOR` -> `decSharePlusCBOR`
    * `fromSharedLensCBOR` -> `decShareLensCBOR`
    * `fromSharedPlusLensCBOR` -> `decSharePlusLensCBOR`
    * `fromNotSharedCBOR` -> `decNoShareCBOR`
    * `fromShareCBORfunctor` -> `decShareMonadCBOR`

This PR disambiguates between versioned CBOR implemented `cardano-ledger-binary` and regular plain CBOR interface implemented by `cardano-binary`. This is necessary for two major reasons:

1. Both APIs will be used within ledger
2. By renaming versioned `To/FromCBOR` to `Enc/DecCBOR` instead of the plain from `cardano-binary` we reduce breakage downstream, since it is mostly ledger codebase that is concerned with versioning by the protocol version.

This PR was fully generated by running `sed` and doing `git mv` on two files. In other words there are no changes that were done manually. This fact should simplify the review process. That also means that the changelog entries will follow in a subsequent PR

# Checklist

- [x] Self-reviewed the diff


JaredCorduan (approved):
LGTM

PR #3304, opened 2023-02-20
Add gen-hie to CI

aniketd:
# Description

Add `gen-hie` to the CI workflow.

# Checklist

- [x] Self-reviewed the diff


.github/workflows/haskell.yml
```diff
@@ -80,6 +80,9 @@ jobs:
     - name: Cabal update
       run: cabal update
 
+    - name: Install implicit-hie
+      run: cabal install implicit-hie
+
```
lehins: This is no longer needed, right?
```suggestion
```

.github/workflows/haskell.yml
```diff
@@ -152,3 +155,25 @@ jobs:
...
+    runs-on: ubuntu-latest
+
+    defaults:
+      run:
+        shell: bash
+
+    strategy:
+      fail-fast: false
+
+    steps:
```
lehins, line 173: Adding caching could speed things up:
```suggestion
    steps:
     - uses: actions/cache@v3 
       name: Cache cabal store 
       with: 
         path: | 
           ${{ steps.setup-haskell.outputs.cabal-store }} 
           dist-newstyle

```

See example here: https://github.com/input-output-hk/cardano-ledger/blob/61bcb797babc992cf079f0333d870720e83f2d10/.github/workflows/haskell.yml#L89-L95

lehins (changes requested):
Bring back all the `base` bounds and we can merge this PR. :+1:

.github/workflows/haskell.yml
aniketd: Awesome! Super thanks for this suggestion!

.github/workflows/haskell.yml
aniketd: Yep!

eras/allegra/impl/cardano-ledger-allegra.cabal
```diff
@@ -46,7 +46,7 @@ library
         -Wunused-packages
 
     build-depends:
-        base >=4.12 && <4.17,
+        base >=4.12 && <5,
```
lehins: We do not support ghc-9.4 yet. When I said we need to revert back all removed constraints on base I meant revert them literally
```suggestion
        base >=4.12 && <4.17,
```

scripts/gen-hie.sh
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+~/.cabal/bin/gen-hie > hie.yaml
```
lehins: We can't require `gen-hie` to be installed by `cabal`, so we need to require `gen-hie` be on the PATH. 

That means we need it to be available in our nix-shell as well, which I think is as easy as adding it to tooling: https://github.com/input-output-hk/cardano-ledger/blob/b9aa1ad1728c0ceeca62657ec94d6d099896c052/shell.nix#L33-L38

lehins (changes requested):
We only officially support ghc-8.10 and ghc-9.2 Which likely means ghc-9.0  also works fine.

So, if you want to update the lower bound on `base` from `>=4.12` to `>=4.14` that would be fine, but the upper bonds has to stay `<4.17` until we can properly verify that it works with those newer compiler versions. From what I know they don't, that's why it is good to be explicit about it.

scripts/gen-hie.sh
aniketd: Additing it to the `tools` didn't work (with `key not found` error), but adding it to the `nativeBuildInputs` worked :+1:

eras/byron/ledger/executable-spec/byron-spec-ledger.cabal
```diff
@@ -73,16 +73,16 @@ test-suite byron-spec-ledger-test
         "-with-rtsopts=-K4m -M150m"
 
     build-depends:
-        base,
-        bimap,
+        base >=4.14 && <4.17,
```
lehins: I removed those constraints on purpose, because there is no point in duplicating the same constraints for the library and the test-suite at the same time

lehins (approved):
Looks like you brought back some changes during conflict resolution that were deliberately removed. If you could go through and revert them , that would be great.

PR #3305, opened 2023-02-20
update tullia/cicero CI

dermetfan:
# Description

- support PRs from forks
- run with exec driver instead of podman
  - more reliable
  - much better caching
  - faster
- update tullia
  - update nix version (needed for tullia's std flake input) by updating nixpkgs so we can keep eval memory usage down

# Checklist

- ~~New tests are added if needed and existing tests are updated~~
- ~~Any changes are noted in the [changelog](https://github.com/input-output-hk/cardano-ledger/blob/master/CHANGELOG.md)~~
- ~~Code is formatted with [`fourmolu`](https://github.com/fourmolu/fourmolu) (which can be run with `scripts/fourmolize.sh`)~~
- ~~Cabal files are formatted (which can be done with `scripts/cabal-format.sh`)~~
- [x] Self-reviewed the diff

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -268,16 +268,16 @@
     },
     "nixpkgs": {
       "locked": {
-        "lastModified": 1669546925,
-        "narHash": "sha256-Gvtk9agz88tBgqmCdHl5U7gYttTkiuEd8/Rq1Im0pTg=",
+        "lastModified": 1676817468,
+        "narHash": "sha256-ovuJ1jQOC2/EEibufBkXmSN/O9mLx80Wh7aDmHmHAhA=",
         "owner": "NixOS",
         "repo": "nixpkgs",
-        "rev": "fecf05d4861f3985e8dee73f08bc82668ef75125",
+        "rev": "0cf4274b5d06325bd16dbf879a30981bc283e58a",
         "type": "github"
       },
       "original": {
         "owner": "NixOS",
-        "ref": "nixos-22.05",
+        "ref": "nixos-22.11",
         "repo": "nixpkgs",
         "type": "github"
       }
--- a/flake.lock
+++ b/flake.lock
@@ -115,21 +115,6 @@
         "type": "github"
       }
     },
-    "flake-utils_3": {
-      "locked": {
-        "lastModified": 1653893745,
-        "narHash": "sha256-0jntwV3Z8//YwuOjzhV2sgJJPt+HY6KhU7VZUL0fKZQ=",
-        "owner": "numtide",
-        "repo": "flake-utils",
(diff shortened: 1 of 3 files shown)


flake.lock
```diff
@@ -268,16 +268,16 @@
...
     "nixpkgs": {
       "locked": {
-        "lastModified": 1669546925,
-        "narHash": "sha256-Gvtk9agz88tBgqmCdHl5U7gYttTkiuEd8/Rq1Im0pTg=",
+        "lastModified": 1676817468,
+        "narHash": "sha256-ovuJ1jQOC2/EEibufBkXmSN/O9mLx80Wh7aDmHmHAhA=",
         "owner": "NixOS",
         "repo": "nixpkgs",
-        "rev": "fecf05d4861f3985e8dee73f08bc82668ef75125",
+        "rev": "0cf4274b5d06325bd16dbf879a30981bc283e58a",
```
teodanciu, line 275: How does it work,  are we using different packages when running in cicero rather than with regular nix?  So this change will impact only cicero runs?

dermetfan: No, we could do that but we don't. This line says we want to use the same nixpkgs for cicero:
https://github.com/input-output-hk/cardano-ledger/blob/d86964db9112031f1266363385babb0ff99216b3/flake.nix#L16

If we removed that line it would fall back to the nixpkgs version that happens to be defined in tullia's flake. That would be fine but unless there is a need to use different nixpkgs versions for different things, it's generally a good idea to only use one nixpkgs and share that with everything in a flake. Every nixpkgs instance needs about 300 MB of memory to evaluate (IIRC). So by sharing one we can speed up evaluation. Also wouldn't it be a good idea to update from 22.05 to 22.11 for our package build as well?

teodanciu: I still don't understand, sorry! 
So: we ARE using the same nixpkgs for cicero because of the  ` inputs.nixpkgs.follows = "nixpkgs"; ` . 
But we are not using flakes for anything else, at least not as a developer/CI. 

So then what is the effect of updating nixpkgs here,  if the package build is using the nixpkgs from sources.json (by not using flakes), and  we have this setting: ` inputs.nixpkgs.follows = "nixpkgs"; `  ?
(Sorry if I'm missing something very obvious)

dermetfan: Oh, it was me who missed something very obvious! :smile: I didn't look properly and just assumed a normal flake-based approach in this repo but we're using niv. So yes, this change will only impact cicero runs.

teodanciu: Thanks for the explanation :)

teodanciu (approved):
Looks good to me

dermetfan:
bors merge

dermetfan:
bors merge

PR #3307, opened 2023-02-21
Fix Address deserialization compatibility

lehins:
# Description

Starting with Babbage era it is not possible to decode a TxOut with a bad address, such as one that had trailing junk data or a Ptr address with values that are too large or trailed with junk. That being said, prior to Babbage there was a few Ptrs that violated those restrictions and thus need to be supported. This PR fixes this backwards compatibility of retaining buggy behavior for older eras:

* Implement lenient Ptr decoding
* Improve BabbageTxOut serialization

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/libs/cardano-ledger-core/src/Cardano/Ledger/Address.hs
+++ b/libs/cardano-ledger-core/src/Cardano/Ledger/Address.hs
@@ -26,15 +26,13 @@ module Cardano.Ledger.Address (
   RewardAcnt (..),
   serialiseRewardAcnt,
   deserialiseRewardAcnt,
-  -- internals exported for testing
   bootstrapKeyHash,
+  -- internals exported for testing
   putAddr,
   putCredential,
   putPtr,
   putRewardAcnt,
   putVariableLengthWord64,
-  -- TODO: these should live somewhere else
-  word64ToWord7s,
   Word7 (..),
   toWord7,
 
@@ -111,6 +109,7 @@ import qualified Data.ByteString.Base16 as B16
 import qualified Data.ByteString.Lazy as BSL
 import Data.ByteString.Short as SBS (ShortByteString, fromShort, index, length, toShort)
 import qualified Data.ByteString.Unsafe as BS (unsafeDrop, unsafeIndex, unsafeTake)
+import Data.Function (fix)
 import Data.Map.Strict (Map)
 import Data.Maybe (fromMaybe)
 import Data.Proxy (Proxy (..))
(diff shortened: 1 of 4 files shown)


libs/cardano-ledger-core/src/Cardano/Ledger/Address.hs
```diff
@@ -791,13 +808,24 @@ decodeVariableLengthWord32 name buf = do
         res <- decode7BitVarLength name buf (\_ -> failDecoding name "too many bytes.") acc
         -- Only while decoding the last 7bits we check if there was too many
         -- bits supplied at the beginning.
-        unless (bufUnsafeIndex buf off0 .&. 0b01110000 == 0) $
+        unless (bufUnsafeIndex buf off0 .&. 0b11110000 == 0b10000000) $
```
lehins, line 811: Change to this check is not needed, this is mostly to clarify that the most significant bit is always set. In other words it is impossible for it to not be set at this particular point.

libs/cardano-ledger-core/src/Cardano/Ledger/Address.hs
```diff
@@ -791,13 +808,24 @@ decodeVariableLengthWord32 name buf = do
...
           failDecoding name "More than 32bits was supplied"
         pure res
       {-# INLINE d7last #-}
   d7 (d7 (d7 (d7 d7last))) 0
 {-# INLINE decodeVariableLengthWord32 #-}
 
+-- | This decoder is here only with the purpose of preserving old buggy behavior. Should
+-- not be used for anything else.
```
JaredCorduan, line 819: want to give it a scarier name?

libs/cardano-ledger-core/src/Cardano/Ledger/Address.hs
```diff
@@ -291,12 +290,10 @@ isBootstrapRedeemer (BootstrapAddress (Byron.Address _ _ Byron.ATRedeem)) = True
...
-putPtr (Ptr slot (TxIx txIx) (CertIx certIx)) = do
-  putSlot slot
-  putVariableLengthWord64 ((fromIntegral @_ @Word64) txIx)
-  putVariableLengthWord64 ((fromIntegral @_ @Word64) certIx)
-  where
-    putSlot (SlotNo n) = putVariableLengthWord64 n
+putPtr (Ptr (SlotNo slot) (TxIx txIx) (CertIx certIx)) = do
+  putVariableLengthWord64 slot
+  putVariableLengthWord64 txIx
+  putVariableLengthWord64 certIx
```
JaredCorduan, line 296: lovely :)

JaredCorduan (approved):
LGTM! so nice that this is all the past now :relieved:

libs/cardano-ledger-core/src/Cardano/Ledger/Address.hs
lehins: Nah, its alright. It is an internal function that is restricted to this module only. So the danger of anyone using it is almost non-existent.

PR #3316, opened 2023-02-28
Improve nightly builds

teodanciu:
# Description

A few improvements to nightly builds:
* Rename the Github action run to make it clear it's a nightly build - hopefully this will make it easier to distinguish it in the emails if it fails
*  Add new substituters (so that we can use caching in github actions running nix )
* Run `nix-build` and `nix-shell` as part of nightly builds, so we are aware at latest the next day if they've been broken 

# Checklist

- [x] Self-reviewed the diff


JaredCorduan (approved):
thank you!!

Issue #3281, opened 2023-02-03
Create `testlib`s for all eras except Shelley and move Arbitrary instances over to those


aniketd:

lehins:
`testlib` for Shelley was handled in #3288 Posting here just for the record.

PR #3318, opened 2023-03-01
Move Shelley EraIndepGenerators to TPraos testlib

aniketd:
# Description

This is a part of #3281 

# Checklist

- [ ] Self-reviewed the diff


eras/shelley/test-suite/src/Test/Cardano/Ledger/Shelley/Rules/ClassifyTraces.hs
```diff
@@ -103,6 +102,9 @@ relevantCasesAreCovered ::
   , ChainProperty era
   , QC.HasTrace (CHAIN era) (GenEnv era)
   , ProtVerAtMost era 8
+  , Show (Environment (CHAIN era))
+  , State (CHAIN era) ~ ChainState era
+  , Signal (CHAIN era) ~ Block (BHeader (EraCrypto era)) era
```
aniketd, line 107: @lehins I wonder if these are necessary here because `ShelleyEraGen` equivalent is not imported. I'm following up on this too.

eras/shelley/test-suite/bench/BenchValidation.hs
```diff
@@ -82,10 +79,11 @@ validateInput ::
   , GetLedgerView era
   , MinLEDGER_STS era
   , EraGovernance era
+  , Environment (CHAIN era) ~ ()
```
aniketd, line 82: @lehins I haven't looked at why these are necessary yet. I'm following up on this.

PR #3320, opened 2023-03-06
Don't run unnecessary gh actions

teodanciu:
# Description
Reduces the number of github actions run by cancelling workflows when a new one for the same PR is starting
 
# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/.github/workflows/haskell.yml
+++ b/.github/workflows/haskell.yml
@@ -14,6 +14,10 @@ on:
   schedule:
     # "Nightly" builds: Every day at 06:00 UTC
     - cron: '0 6 * * *'
+# Cancel running actions when a new action on the same PR is started
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
 
 jobs:
   build:
--- a/.github/workflows/haskell.yml
+++ b/.github/workflows/haskell.yml
@@ -14,6 +14,9 @@ on:
   schedule:
     # "Nightly" builds: Every day at 06:00 UTC
     - cron: '0 6 * * *'
+  # for running the workflow manually - useful for branches without PRs, for which jobs don't get ran automatically
+  workflow_dispatch:
+
 # Cancel running actions when a new action on the same PR is started
 concurrency:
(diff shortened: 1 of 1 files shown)


.github/workflows/haskell.yml
```diff
@@ -14,10 +14,18 @@ on:
   schedule:
     # "Nightly" builds: Every day at 06:00 UTC
     - cron: '0 6 * * *'
+  # for running the workflow manually - useful for draft PRs, for which jobs don't get ran automatically
```
lehins: What is the point of a Draft PR if you don't run CI on it? In my workflow the only reason why I open a draft PR is precisely so the CI is executed. If you don't need the CI then there is no reason to even open a PR.

teodanciu: Fair enough, I didn't realize people used it for this. I usually open it to get early feedback or maybe for psychological comfort :laughing:  but I take your point. 
Happy to remove this commit,. Anyway, it's not gonna make a huge difference in terms of  number of runs - probably the biggest "offender" is stale branches running after a push.

JaredCorduan: I've wondered myself what exactly are draft PRs for. I concluded that they must be for getting feedback via the PR interface (and that CI is mostly redundant with running tests locally). but seems that was not what everyone expects.

lehins: Ability to run manually is a good feature as well (for example if it is too early for a PR even a draft one). 
I didn't realize there was an option to do that.  I am just not sure that disabling CI on Draft PRs is a good idea.

> I usually open it to get early feedback

That is exactly what draft PRs are also good for. Like this one for example :grin:

lehins: > and that CI is mostly redundant with running tests locally)

I almost never run the full test suite locally. I have better uses for my computer than running tests :smile:

lehins:
```suggestion
  # for running the workflow manually - useful for branches without PRs, for which jobs don't get ran automatically
```

lehins (approved):
Minor fix to a comment, otherwise looks great.

PR #3322, opened 2023-03-09
More work on constraint soundness property

UlfNorell:
and some bugfixes.

### DIFF
--- a/libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/Combinators.hs
+++ b/libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/Combinators.hs
@@ -75,7 +75,8 @@ fixSet mess numtrys size genA s = help numtrys s
       EQ -> pure set
       GT -> help (trys - 1) (iterate Set.deleteMin set !! (Set.size set - size))
       LT -> do
-        new <- Set.fromList <$> vectorOf (size - Set.size set) genA
+        let need = size - Set.size set
+        new <- Set.fromList . take need . filter (`Set.notMember` set) <$> vectorOf size genA
         help (trys - 1) (Set.union new set)
 
 {-
--- a/libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/Combinators.hs
+++ b/libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/Combinators.hs
@@ -25,7 +25,7 @@ errorMess extra mess = error (unlines ("\nGen-time error" : (reverse (extra : me
 
 -- | suchThat version that tracks Gen-time errors
 suchThatErr :: [String] -> Gen a -> (a -> Bool) -> Gen a
-suchThatErr msgs gen p = sized $ \ n -> try (10 :: Int) n
+suchThatErr msgs gen p = sized $ \n -> try (10 :: Int) n
   where
     try 0 _ = errorMess "SuchThat times out" msgs
     try k sz = do
@@ -80,7 +80,7 @@ fixSet mess numtrys size genA s = help numtrys s
       LT -> do
         let need = size - Set.size set
         new <- Set.fromList . take need . filter (`Set.notMember` set) <$> vectorOf size genA
-        sized $ \ n -> resize (n + 5) $ help (trys - 1) (Set.union new set)
+        sized $ \n -> resize (n + 5) $ help (trys - 1) (Set.union new set)
 
 {-
 x <- genA
--- a/libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/Combinators.hs
+++ b/libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/Combinators.hs
@@ -25,11 +25,14 @@ errorMess extra mess = error (unlines ("\nGen-time error" : (reverse (extra : me
 
 -- | suchThat version that tracks Gen-time errors
 suchThatErr :: [String] -> Gen a -> (a -> Bool) -> Gen a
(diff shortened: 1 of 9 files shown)


libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/Combinators.hs
```diff
@@ -25,11 +25,14 @@ errorMess extra mess = error (unlines ("\nGen-time error" : (reverse (extra : me
...
-    Just y -> pure y
-    Nothing -> errorMess "SuchThat times out" msgs
+suchThatErr msgs gen p = sized $ \n -> try (10 :: Int) n
+  where
+    try 0 _ = errorMess "SuchThat times out" msgs
+    try k sz = do
+      x <- resize sz $ suchThatMaybe gen p
+      case x of
+        Just y -> pure y
+        Nothing -> try (k - 1) (sz + 5)
```
TimSheard, line 36: I never really understood how resize works. Is there any reason to why you choose '5' as the amount to increase size by? Thanks for reworking this.

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/Rewrite.hs
```diff
@@ -131,19 +131,32 @@ removeSameVar :: [Pred era] -> [Pred era] -> [Pred era]
 removeSameVar [] ans = reverse ans
 removeSameVar ((Var v :=: Var u) : more) ans | Name v == Name u = removeSameVar more ans
 removeSameVar ((Var v :⊆: Var u) : more) ans | Name v == Name u = removeSameVar more ans
+removeSameVar (Disjoint (Var v@(V _ rep _)) (Var u) : more) ans | Name v == Name u = removeSameVar more ((Lit rep mempty :=: Var v) : ans)
```
TimSheard, line 134: Good catch. I forgot about this one.

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/Rewrite.hs
```diff
@@ -131,19 +131,32 @@ removeSameVar :: [Pred era] -> [Pred era] -> [Pred era]
...
 removeSameVar (m : more) ans = removeSameVar more (m : ans)
 
 removeEqual :: [Pred era] -> [Pred era] -> [Pred era]
 removeEqual [] ans = reverse ans
 removeEqual ((Var v :=: Var u) : more) ans | Name v == Name u = removeEqual more ans
-removeEqual ((Var v :=: expr) : more) ans = removeEqual (map sub more) ((Var v :=: expr) : (map sub ans))
+removeEqual ((Var v :=: expr@Lit {}) : more) ans = removeEqual (map sub more) ((Var v :=: expr) : map sub ans)
+  where
+    sub = substPred [SubItem v expr]
+removeEqual ((expr@Lit {} :=: Var v) : more) ans = removeEqual (map sub more) ((expr :=: Var v) : map sub ans)
```
TimSheard, line 143: Another obvious case that I missed.

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/Solver.hs
```diff
@@ -265,6 +268,16 @@ solveMap v1@(V _ r@(MapR dom rng) _) predicate = explain msg $ case predicate of
...
+  (expr `Subset` Rng (Var v2))
+    | Name v1 == Name v2 -> do
+        With _ <- hasOrd rng rng
+        With n <- simplifySet rng expr
+        mapSpec SzAny RelAny (RngRel (relSuperset rng n))
+  (Rng expr `Subset` (Var v2))
+    | Name v1 == Name v2 -> do
+        With _ <- hasOrd rng rng
+        With n <- simplifySet rng expr
+        mapSpec SzAny RelAny (RngRel (relSuperset rng n))
```
TimSheard, line 280: The problem with Dom and Rng, is that they really increase the number of cases, and those cases are really easy to miss.

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/Spec.hs
```diff
@@ -675,38 +688,12 @@ mergeRngSpec (RngRel RelAny) x = x
...
-                ++ show siz
-                ++ ") are not compatible."
-            ]
-mergeRngSpec b@(RngProj _ _ _) a@(RngElem _ _) = mergeRngSpec a b
+mergeRngSpec a@(RngElem _ xs) b
+  | runRngSpec xs b = a
+  | otherwise = RngNever ["The RngSpec's are inconsistent.\n  " ++ show a ++ "\n  " ++ show b]
+mergeRngSpec a b@(RngElem _ xs)
+  | runRngSpec xs a = b
+  | otherwise = RngNever ["The RngSpec's are inconsistent.\n  " ++ show a ++ "\n  " ++ show b]
```
TimSheard, line 696: In retrospect, this is so obvious. Brilliant!

TimSheard (approved):
Thanks for tracking down all those bugs.

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/Combinators.hs
UlfNorell: No particularly strong reasons. The size varies between 0 and 99 during a `quickCheck` run, so bumping by 5 for each retry seemed reasonable. You want to avoid the size getting too big, because it's used for things like lengths of lists and other datastructures.

PR #3323, opened 2023-03-13
Move `ToJSON` orphans from `cardano-node`

lehins:
# Description

## JSON instances

This PR moves `ToJSON` orphan instances for ledger types from [`cardano-node`](https://github.com/input-output-hk/cardano-node/blob/17edf34f0877507dea552e63ee85bd00a2a648e6/cardano-api/src/Cardano/Api/Orphans.hs) with a few key renames:

* For `PState`:
  * Renamed `"pParams pState"` -> `"pParams"`
  * Renamed `"fPParams pState"` -> `"fPParams"`
  * Renamed `"retiring pState"` -> `"retiring"`
  * Added `"deposits"`
* For `InstantaneousRewards`:
  * Addition of `deltaReserves` and `deltaTreasury` fields
* For `DState`:
  * `"unifiedRewards"` removed
  * `"unified"` added, which contains an object with both rewards and deposits

In this PR an unusual technique is utilized for defining JSON instances. For example:

```haskell
instance ToJSON ExUnits where
  toJSON exUnits@(ExUnits _ _) =
    let ExUnits {exUnitsMem, exUnitsSteps} = exUnits
     in object
          [ "exUnitsMem" .= toJSON exUnitsMem
          , "exUnitsSteps" .= toJSON exUnitsSteps
          ]
```

It relies on both of the features: the pattern matching and either `RecordWildCards` or `NamedFieldPuns` extension. Both of the extensions are equivalent for this technique, i.e. neither one of them is better then the other. At first glance this technique might seem redundant, but it provides two important safety guarantees:

* Pattern matching with `_` ensures that compilation will fail whenever a new field is added.
* `RecordWildCards`/`NamedFieldPuns` ensures that the compilation will fail whenever a field is renamed, which most likely need to be reflected in the ToJSON instance as well. Also those extensions make it easy to see the correlation between the field names and the the json keys.

## Superconstraints

Adds a few super constraints to reduce total number of constraints used.

* For `Val`:  `EncCBOR (CompactForm t)`, `DecCBOR (CompactForm t)`, `NoThunks t`, `EncCBOR t`, `DecCBOR t`, `ToJSON t`, `NFData t` and `Show t`
* `ToJSON` for classes: `EraTxOut`, `EraPParams`

## Field renames

This PR also renames two new fields in the Conway era:
* `ensCC` -> `ensCommittee`
* `rsES` -> `rsEnactState`

## Fixup PR template

Also fixes up the PR template by marking formatting checkboxes as checked by default

## Tests

All tests will be added in a separate PR and are tracked by #3324

# Checklist

- [ ] Self-reviewed the diff


eras/shelley/impl/src/Cardano/Ledger/Shelley/PParams.hs
```diff
@@ -1,3 +1,4 @@
+{-# LANGUAGE AllowAmbiguousTypes #-}
```
JaredCorduan: out of curiosity, what required this?

libs/cardano-ledger-binary/testlib/Test/Cardano/Ledger/Binary/Arbitrary.hs
```diff
@@ -50,6 +55,79 @@ import Test.Crypto.VRF ()
...
 import Test.QuickCheck.Instances ()
 
+firstUnreservedTag :: Word64
+firstUnreservedTag = 6
+
+-- | Simple values that are either unassigned or don't have a specialized type already
+simple :: [Word8]
+simple = [0 .. 19] ++ [23] ++ [32 ..]
+
+instance Arbitrary Term where
```
JaredCorduan, line 65: is this related to any of the other changes. :thinking: (curios how CBOR relates...)

eras/shelley/impl/src/Cardano/Ledger/Shelley/PParams.hs
lehins: Good catch. I initially used it, but this is no longer required. I've switched to passing `Proxy` instead.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -6,11 +6,10 @@ to the issue.
...
+- [ ] Commit sequence broadly makes sense and commits have useful messages
 - [ ] New tests are added if needed and existing tests are updated
-- [ ] Any changes are noted in the [changelog](https://github.com/input-output-hk/cardano-ledger/blob/master/CHANGELOG.md)
-- [ ] Code is formatted with [`fourmolu`](https://github.com/fourmolu/fourmolu) (use `scripts/fourmolize.sh`)
-- [ ] Cabal files are formatted (use `scripts/cabal-format.sh`)
-- [ ] `hie.yaml` has been updated (use `scripts/gen-hie.sh`)
+- [ ] Any changes are noted in the `CHANGELOG.md` for affected package
+- [x] Code is formatted with [`fourmolu`](https://github.com/fourmolu/fourmolu) (use `scripts/fourmolize.sh`)
+- [x] Cabal files are formatted (use `scripts/cabal-format.sh`)
+- [x] [`hie.yaml`](https://github.com/input-output-hk/cardano-ledger/blob/master/hie.yaml) has been updated (use `scripts/gen-hie.sh`)
```
Soupstraw, line 14: Are these supposed to be ticked?

lehins: Yeah, because they are enforced by CI. So, there is no reason to manually check it. All three are left mostly as a reminder and as instructions on how to do this formatting for new contributors.

JaredCorduan: @lehins likes these ticked by default, since you can't get past CI without them (so not really a human decision). and they are present just so that new folks can learn about them.

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Scripts.hs
```diff
@@ -199,6 +201,9 @@ instance Crypto c => EraScript (AlonzoEra c) where
       PlutusScript PlutusV1 _ -> "\x01"
       PlutusScript PlutusV2 _ -> "\x02"
 
+instance Era era => ToJSON (AlonzoScript era) where
+  toJSON = String . serializeAsHexText
```
lehins, line 205: @JaredCorduan This is where the serialization changes made to `cardano-ledger-binary`  come into play.

JaredCorduan: thanks! makes sense now!

JaredCorduan (approved):
LGTM!

PR #3327, opened 2023-03-13
OCert cbor and minor improvements

lehins:
# Description

Looks like OCert can't be easily CBOR versioned in the cardano-api, which is OK, since it is a pretty simple type and is unlikely to change. So this PR adds ToCBOR/FromCBOR instances and bases EncCBOR/DecCBOR on the unversioned implementation

Also other minor improvements

# Checklist

- [x] Self-reviewed the diff


JaredCorduan (approved):
:+1:

PR #3331, opened 2023-03-14
enforce unique elements in CBOR sets, ver. 9

### LINKED ISSUE #3340: duplicate keys in CBOR set decoding #2965
We currently allow CBOR arrays with duplicate elements to be serialized as sets (by dropping the older duplicated elements). We also currently do not allow for using the CBOR set tag 258. In the spirit of [proposed CIP-80](https://github.com/cardano-foundation/CIPs/pull/372), we consider the first issues (duplicates) a serious flaw that we should fix at the next hard fork, and we consider the second issue a non-serious flaw that we should migrate away from (so start permitting the tag at the next hard fork but do not enforce it).

JaredCorduan:
# Description

Starting in major version 9, we now raise a decoding error if a CBOR set contains duplicate keys. Additionally, we allow (but do not require) the set tag.

resolves #3340

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/libs/cardano-ledger-binary/CHANGELOG.md
+++ b/libs/cardano-ledger-binary/CHANGELOG.md
@@ -13,6 +13,8 @@
 * Fix deserializer for `Rational` and allow optionally tag `30` starting with protocol version `2`
 * Fix serializer for `Ratio` and encode tag `30` starting with protocol version `2`
 * Add new encoder `encodeRatioNoTag` for `Ratio`
+* Changed: Starting in version 9, duplicate keys in CBOR sets are not longer allowed.
+  Additionally, the CBOR set tag 258 is permitted but not enforced.
 
 ### `testlib`
 
--- a/libs/cardano-ledger-binary/cardano-ledger-binary.cabal
+++ b/libs/cardano-ledger-binary/cardano-ledger-binary.cabal
@@ -133,6 +133,7 @@ test-suite tests
     other-modules:
         Test.Cardano.Ledger.Binary.Failure
         Test.Cardano.Ledger.Binary.PlainSpec
+        Test.Cardano.Ledger.Binary.Success
         Test.Cardano.Ledger.Binary.RoundTripSpec
         Test.Cardano.Ledger.Binary.Vintage.Coders
         Test.Cardano.Ledger.Binary.Vintage.Drop
--- a/libs/cardano-ledger-binary/test/Test/Cardano/Ledger/Binary/Failure.hs
+++ b/libs/cardano-ledger-binary/test/Test/Cardano/Ledger/Binary/Failure.hs
@@ -1,11 +1,14 @@
 {-# LANGUAGE DataKinds #-}
+{-# LANGUAGE RankNTypes #-}
 {-# LANGUAGE TypeApplications #-}
 
 module Test.Cardano.Ledger.Binary.Failure (spec) where
 
 import Cardano.Ledger.Binary
-import Data.Either (isLeft)
 import Data.Map (Map)
+import Data.Proxy (Proxy (Proxy))
+import Data.Set (Set)
+import Test.Cardano.Ledger.Binary.RoundTrip (Trip (..), embedTripRangeFailureExpectation)
 import Test.Hspec
 import Test.Hspec.QuickCheck
 import Test.QuickCheck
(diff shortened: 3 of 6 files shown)


libs/cardano-ledger-binary/CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Version history for `cardano-ledger-binary`
 
+## 1.1.0.0
+
+Changed:
```
lehins: I don't think it makes sense to have these Changed/Added/Removed sections in the change log:

* we end up with related changes spread across multiple sections
* Also this redundant pattern becomes very common:
```
* Added:
  * Added Foo
  * Added Bar
* Removed:
  * Removed Baz
```

In my opinion as long as the change is listed it is more then enough information. Also, adding "Changed" to the "Changelog" seems somewhat redundant :smile: 

So, I suggest abandoning these sections and just stick with bullet points of changes.

libs/cardano-ledger-binary/test/Test/Cardano/Ledger/Binary/Failure.hs
```diff
@@ -28,6 +30,26 @@ genDuplicateAssocListEncoding = do
...
+genDuplicateListEncoding = do
+  xs <- genDuplicateList
+  let definite = encodeListLen (fromIntegral $ Prelude.length xs) <> foldMap encCBOR xs
+      indefinite = encodeListLenIndef <> foldMap encCBOR xs <> encodeBreak
+  oneof
+    [ pure $ definite
+    , pure $ encodeTag 258 <> definite
+    , pure $ indefinite
+    , pure $ encodeTag 258 <> indefinite
+    ]
```
lehins:
```suggestion
  elements
    [ definite
    , encodeTag 258 <> definite
    , indefinite
    , encodeTag 258 <> indefinite
    ]
```

libs/cardano-ledger-binary/test/Test/Cardano/Ledger/Binary.hs
```diff
@@ -0,0 +1,8 @@
+module Test.Cardano.Ledger.Binary (decode) where
+
+import Cardano.Ledger.Binary
+
+decode :: DecCBOR a => Version -> Encoding -> Either DecoderError a
+decode version enc =
+  let encoded = serialize version enc
+   in decodeFull version encoded
```
lehins: There is a much better alternative to this function in #3332 

I'll point it out in the places where it is used

libs/cardano-ledger-binary/test/Test/Cardano/Ledger/Binary/Failure.hs
```diff
@@ -36,12 +58,16 @@ prop_shouldFailMapWithDupKeys =
...
-decode version enc =
-  let encoded = serialize version enc
-   in decodeFull version encoded
+-- | Starting in version 9, do not accept duplicates in CBOR sets
+prop_shouldFailSetWithDupKeys :: Property
+prop_shouldFailSetWithDupKeys =
+  forAllBlind genDuplicateListEncoding $
+    ( \setEncoder ->
+        (property . isLeft) (decode (natVersion @9) setEncoder :: Either DecoderError (Set Int))
+    )
```
lehins: `embedTripRangeFailureExpectation` function from #3332 will produce better error reporting and will ensure this property holds true for future protocol versions
Also parenthesis are redundant.

```suggestion
     \setEncoding ->
      let trip = Trip id (decCBOR @(Set Int))
      in property $ embedTripRangeFailureExpectation trip (natVersion @9) maxBound setEncoding
```

lehins (changes requested):
Looking great

libs/cardano-ledger-binary/cardano-ledger-binary.cabal
```diff
@@ -131,8 +131,10 @@ test-suite tests
     main-is:          Main.hs
     hs-source-dirs:   test
     other-modules:
+        Test.Cardano.Ledger.Binary
```
lehins: This module is not needed

libs/cardano-ledger-binary/CHANGELOG.md
JaredCorduan: that's fine. I liked following a standard (KeepAChangelog) for the same reason I like fourmalo: less for folks to bikeshed about :laughing: . but as long as we are consistent and clear, I'm happy :)

libs/cardano-ledger-binary/cardano-ledger-binary.cabal
JaredCorduan: you mean now that we are dropping `decode`?

libs/cardano-ledger-binary/test/Test/Cardano/Ledger/Binary/Success.hs
```diff
@@ -0,0 +1,43 @@
...
+    , pure $ encodeTag 258 <> indefinite
+    ]
+
+-- | Starting in version 9, check set decoding with and without tag 258
+prop_shouldFailSetWithDupKeys :: Property
+prop_shouldFailSetWithDupKeys =
+  forAllBlind genUniqueListEncoding $
+    ( \setEncoder ->
+        (property . isRight) (decode (natVersion @9) setEncoder :: Either DecoderError (Set Int))
+    )
```
lehins: `isRight` isn't that great for debugging when failure occurs :smile: 
```suggestion
     \setEncoder ->
       let trip = Trip id (decCBOR @(Set Int))
       in property $ embedRoundTrip trip (natVersion @9) maxBound setEncoder
```

libs/cardano-ledger-binary/cardano-ledger-binary.cabal
lehins: yep

libs/cardano-ledger-binary/test/Test/Cardano/Ledger/Binary/Success.hs
```diff
@@ -0,0 +1,43 @@
...
+genUniqueListEncoding = do
+  xs <- genUniqueList
+  let definite = encodeListLen (fromIntegral $ Prelude.length xs) <> foldMap encCBOR xs
+      indefinite = encodeListLenIndef <> foldMap encCBOR xs <> encodeBreak
+  oneof
+    [ pure $ definite
+    , pure $ encodeTag 258 <> definite
+    , pure $ indefinite
+    , pure $ encodeTag 258 <> indefinite
+    ]
```
lehins:
```suggestion
  elements
    [ definite
    , encodeTag 258 <> definite
    , indefinite
    , encodeTag 258 <> indefinite
    ]
```

libs/cardano-ledger-binary/test/Test/Cardano/Ledger/Binary/Success.hs
```diff
@@ -0,0 +1,43 @@
...
+import Data.List (nub)
+import Data.Set (Set)
+import Test.Cardano.Ledger.Binary (decode)
+import Test.Hspec
+import Test.Hspec.QuickCheck
+import Test.QuickCheck
+
+-- | Generate a list with unique elements
+genUniqueList :: Gen [Int]
+genUniqueList = nub <$> arbitrary
```
lehins: This will be faster. Also default `arbitrary` instance for `Int` generates a lot of duplicates. So it is better to use the full range of Int
```suggestion
genUniqueList = Set.toList . Set.fromList <$> listOf chooseAny
```

libs/cardano-ledger-binary/test/Test/Cardano/Ledger/Binary/Success.hs
```diff
@@ -0,0 +1,43 @@
...
+import Data.Either (isRight)
+import Data.List (nub)
+import Data.Set (Set)
+import Test.Cardano.Ledger.Binary (decode)
+import Test.Hspec
+import Test.Hspec.QuickCheck
+import Test.QuickCheck
+
+-- | Generate a list with unique elements
+genUniqueList :: Gen [Int]
```
lehins: ~More general is better~

I see it is used for specific purpose only

libs/cardano-ledger-binary/test/Test/Cardano/Ledger/Binary/Success.hs
JaredCorduan: > `isRight` isn't that great for debugging when failure occurs smile

you missed the opportunity for a good pun:

> `isRight` isn't that great for debugging when failure occurs, amiright?

lehins: How about this one? :smile: 

`isRight` isn't right for debugging

libs/cardano-ledger-binary/test/Test/Cardano/Ledger/Binary/Success.hs
```diff
@@ -0,0 +1,48 @@
...
+  (,) xsSet
+    <$> elements
+      [ definite
+      , encodeTag 258 <> definite
+      , indefinite
+      , encodeTag 258 <> indefinite
+      ]
+
+-- | Starting in version 9, check set decoding with and without tag 258
+prop_shouldFailSetWithDupKeys :: Property
```
lehins: This doesn't sound like a right name for the test :smile: 
```suggestion
prop_setWithNoDuplicatesAndTag :: Property
```

lehins (approved):
Beautiful! :+1:

libs/cardano-ledger-binary/test/Test/Cardano/Ledger/Binary/Success.hs
JaredCorduan: :facepalm: :spaghetti:

PR #3333, opened 2023-03-15
Move Shelley.EraIndepGenerators to TPraos

aniketd:
# Description

This may be considered the second installment for #3281 

# Checklist

- [x] Self-reviewed the diff


eras/shelley/test-suite/src/Test/Cardano/Ledger/Shelley/Serialisation/EraIndepGenerators.hs
```diff
@@ -1,162 +1,3 @@
...
--- This generator uses 'mkBlock' provide more coherent blocks.
-genCoherentBlock ::
-  forall era h.
-  ( Mock (EraCrypto era)
-  , EraSegWits era
-  , Arbitrary (Tx era)
-  , h ~ BHeader (EraCrypto era)
-  ) =>
-  Gen (Block h era)
-genCoherentBlock = do
```
lehins: As the comment suggests, this function is used in `ouroboros-network`, therefore it is necessary to bring it back

aniketd: Sorry I missed this out. Thanks for the catch!

eras/alonzo/test-suite/cardano-ledger-alonzo-test.cabal
```diff
@@ -54,7 +54,7 @@ library
         cardano-ledger-shelley-test ^>=1.1,
         cardano-ledger-mary ^>=1.1,
         cardano-ledger-shelley-ma-test ^>=1.1,
-        cardano-protocol-tpraos ^>=1.0,
+        cardano-protocol-tpraos ^>=1.1,
```
lehins: You see, this bump is not needed, because there is no direct dependency on the `testlib`, which is what changed, right?  ``cardano-protocol-tpraos` library have not changed.
```suggestion
        cardano-protocol-tpraos ^>=1.0,
```

eras/babbage/test-suite/cardano-ledger-babbage-test.cabal
```diff
@@ -79,7 +79,7 @@ test-suite cardano-ledger-babbage-test
         cardano-ledger-shelley,
         cardano-ledger-shelley-ma-test,
         cardano-ledger-shelley-test,
-        cardano-protocol-tpraos >=1.0,
+        cardano-protocol-tpraos >=1.1,
```
lehins: Same here
```suggestion
        cardano-protocol-tpraos >=1.0,
```

libs/cardano-protocol-tpraos/cardano-protocol-tpraos.cabal
```diff
@@ -1,6 +1,6 @@
 cabal-version: 3.0
 name:          cardano-protocol-tpraos
-version:       1.0.1.0
+version:       1.1.0.0
```
lehins: Addition of a `testlib` is not a breaking change. Also the version has already been bumped, since the last released version `1.0.0.0`. So nothing to do here
```suggestion
version:       1.0.1.0
```

eras/shelley/test-suite/cardano-ledger-shelley-test.cabal
```diff
@@ -87,13 +87,12 @@ library
         cardano-ledger-core:{cardano-ledger-core, testlib} >=1.0,
         cardano-ledger-pretty,
         cardano-ledger-shelley:{cardano-ledger-shelley, testlib} ^>=1.1,
-        cardano-protocol-tpraos >=1.0,
+        cardano-protocol-tpraos:{cardano-protocol-tpraos, testlib} >=1.1,
```
lehins: Here a lower bound indeed has to be updated, bit not to `1.1` (see other comment)
```suggestion
        cardano-protocol-tpraos:{cardano-protocol-tpraos, testlib} >=1.0.1,
```

eras/shelley-ma/test-suite/cardano-ledger-shelley-ma-test.cabal
```diff
@@ -100,7 +100,7 @@ test-suite cardano-ledger-shelley-ma-test
         cardano-ledger-shelley-ma-test >=1.1,
         cardano-ledger-allegra >=1.1,
         cardano-ledger-mary:{cardano-ledger-mary, testlib},
-        cardano-protocol-tpraos >=1.0,
+        cardano-protocol-tpraos >=1.1,
```
lehins: Ditto
```suggestion
        cardano-protocol-tpraos >=1.0,
```

libs/cardano-protocol-tpraos/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Version history for `cardano-protocol-tpraos`
 
+## 1.1.0.0
```
lehins: If you run `git tag --list`:
```
$ git tag --list | grep praos
cardano-protocol-tpraos-1.0.0.0
```
you will see that the latest version that was released to CHaPs was `1.0.0.0`, which means you are updating the wrong section of the changelog. It should be the section below that needs to be updated.

That being said, if it was really a breaking chnage, then the section below would have to be renamed from `1.0.1.0` -> `1.1.0.0`. But because this change is non-breaking, this is not the case

eras/shelley/test-suite/src/Test/Cardano/Ledger/Shelley/Serialisation/EraIndepGenerators.hs
```diff
@@ -120,13 +40,15 @@ genBlock = Block <$> arbitrary <*> (toTxSeq @era <$> arbitrary)
 --
 -- This generator uses 'mkBlock' provide more coherent blocks.
 genCoherentBlock ::
```
lehins, line 42: I think it would be good to move this generator into `testlib` as well. Just make sure it is re-exported from this module

lehins (changes requested):
Issues with versioning, otherwise looks great.

eras/shelley/test-suite/src/Test/Cardano/Ledger/Shelley/Serialisation/EraIndepGenerators.hs
aniketd: Thanks Alexey! Assuming that you mean `TPraos.Arbitrary`, some of the other modules may also need moving to `tpraos`. Not all of the previous closed PR, but just a few, like the `Constants`, besides the `mkBlock` and `mkOCert` functions and the functions they depend on. Is that okay?

lehins: You know, on a second thought, don't worry about it. I'll do it myself later when I get to the test suite cleanup and organization

lehins (approved):
LGTM :+1:

eras/shelley/test-suite/src/Test/Cardano/Ledger/Shelley/Serialisation/EraIndepGenerators.hs
aniketd: :+1: Thanks! That sounds good. That way the entire extended task will be done in one sweep, and it might be better to plan together too. :raised_hands:

PR #3342, opened 2023-03-17
Improve evaluate transaction balance

lehins:
# Description

to the issue.

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/eras/mary/impl/src/Cardano/Ledger/Mary/UTxO.hs
+++ b/eras/mary/impl/src/Cardano/Ledger/Mary/UTxO.hs
@@ -6,13 +6,14 @@
 
 module Cardano.Ledger.Mary.UTxO (getConsumedMaryValue) where
 
-import Cardano.Ledger.Core
+import Cardano.Ledger.Coin (Coin)
+import Cardano.Ledger.Credential (StakeCredential)
 import Cardano.Ledger.Crypto
+import Cardano.Ledger.Mary.Core
 import Cardano.Ledger.Mary.Era (MaryEra)
-import Cardano.Ledger.Mary.TxBody (MaryEraTxBody (..))
+import Cardano.Ledger.Mary.TxBody ()
 import Cardano.Ledger.Mary.Value (MaryValue)
-import Cardano.Ledger.Shelley.LedgerState (DPState, keyTxRefunds)
-import Cardano.Ledger.Shelley.TxBody (Withdrawals (..))
+import Cardano.Ledger.Shelley.LedgerState (keyCertsRefunds)
 import Cardano.Ledger.Shelley.UTxO (
   ShelleyScriptsNeeded (..),
   getShelleyScriptsNeeded,
@@ -49,17 +50,17 @@ instance Crypto c => EraUTxO (MaryEra c) where
 getConsumedMaryValue ::
   (MaryEraTxBody era, Value era ~ MaryValue (EraCrypto era)) =>
   PParams era ->
-  DPState (EraCrypto era) ->
+  (StakeCredential (EraCrypto era) -> Maybe Coin) ->
   UTxO era ->
   TxBody era ->
   MaryValue (EraCrypto era)
-getConsumedMaryValue pp dpstate (UTxO u) txBody = consumedValue <> txBody ^. mintValueTxBodyF
+getConsumedMaryValue pp lookupRefund (UTxO u) txBody = consumedValue <> txBody ^. mintValueTxBodyF
   where
     {- balance (txins tx ◁ u) + wbalance (txwdrls tx) + keyRefunds pp tx -}
     consumedValue =
       balance (UTxO (Map.restrictKeys u (txBody ^. inputsTxBodyL)))
         <> inject (refunds <> withdrawals)
-    refunds = keyTxRefunds pp dpstate txBody
+    refunds = keyCertsRefunds pp lookupRefund (txBody ^. certsTxBodyG)
     withdrawals = fold . unWithdrawals $ txBody ^. withdrawalsTxBodyL
 
 -- | Computes the set of script hashes required to unlock the transaction inputs and the
--- a/eras/shelley/impl/src/Cardano/Ledger/Shelley/AdaPots.hs
+++ b/eras/shelley/impl/src/Cardano/Ledger/Shelley/AdaPots.hs
@@ -76,8 +76,9 @@ totalAdaPotsES (EpochState (AccountState treasury_ reserves_) _ ls _ _ _) =
     DPState dstate _ = lsDPState ls
     rewards_ = fromCompact $ sumRewardsView (rewards dstate)
     coins = coinBalance u
-    keyDeposits_ = (fromCompact . sumDepositView . RewardDeposits . dsUnified . dpsDState . lsDPState) ls
-    poolDeposits_ = fold ((psDeposits . dpsPState . lsDPState) ls)
+    keyDeposits_ =
+      fromCompact . sumDepositView . RewardDeposits . dsUnified . dpsDState $ lsDPState ls
+    poolDeposits_ = fold (psDeposits . dpsPState $ lsDPState ls)
 
 -- | Calculate the total ada in the epoch state
 totalAdaES :: EraTxOut era => EpochState era -> Coin
@@ -94,10 +95,10 @@ totalAdaES cs =
       , reservesAdaPot
       , rewardsAdaPot
       , utxoAdaPot
-      , -- keyDepositAdaPot,  -- We don't count these two, as their
-      -- poolDepositAdaPot, -- sum is always depositsAdaPot
-      depositsAdaPot
+      , depositsAdaPot
       , feesAdaPot
+      -- , keyDepositAdaPot  -- We don't count these two, as their
+      -- , poolDepositAdaPot -- sum is always depositsAdaPot
       } = totalAdaPotsES cs
 
 -- =============================================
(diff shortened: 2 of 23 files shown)


eras/shelley/impl/src/Cardano/Ledger/Shelley/AdaPots.hs
lehins, line 1: No semantic changes in this module, just a bit of code clarity improvement

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Epoch.hs
```diff
@@ -209,10 +208,10 @@ epochTransition = do
...
     -- the rules, So we can recompute the utxosDeposited field using adjustedDPState
     -- since we have the invariant that: obligationDPState dpstate == utxosDeposited utxostate
-    Coin oblgNew = obligationDPState adjustedDPstate
-    Coin reserves = asReserves acnt'
-    utxoSt''' = utxoSt'' {utxosDeposited = Coin oblgNew}
-    acnt'' = acnt' {asReserves = Coin reserves}
+    oblgNew = obligationDPState adjustedDPstate
+    reserves = asReserves acnt'
+    utxoSt''' = utxoSt'' {utxosDeposited = oblgNew}
+    acnt'' = acnt' {asReserves = reserves}
```
JaredCorduan, line 214: nice catch :)

libs/cardano-ledger-api/test/Test/Cardano/Ledger/Api/Tx/Body.hs
```diff
@@ -0,0 +1,142 @@
...
+import Test.Cardano.Ledger.Babbage.Arbitrary ()
+import Test.Cardano.Ledger.Common
+
+totalTxDeposits ::
+  ShelleyEraTxBody era =>
+  PParams era ->
+  DPState (EraCrypto era) ->
+  TxBody era ->
+  Coin
+totalTxDeposits pp dpstate txb =
```
JaredCorduan, line 36: maybe worth leaving a note that this includes both stake credential and stake pool registration deposits? (and DRep in the future...) I can see myself in the future asking "is this _really_ all the deposits"?

JaredCorduan (approved):
looks great to me!

libs/cardano-ledger-api/test/Test/Cardano/Ledger/Api/Tx/Body.hs
lehins: That's a great point. I'll leave a comment about this. 

This makes me realize that we will have to make `getProducedValue` part of the `EraUTxO` type class, because we will have DRep deposits in Conway.

lehins: Added a bit of haddock with this info

eras/mary/impl/src/Cardano/Ledger/Mary/UTxO.hs
```diff
@@ -49,17 +50,17 @@ instance Crypto c => EraUTxO (MaryEra c) where
...
   MaryValue (EraCrypto era)
-getConsumedMaryValue pp dpstate (UTxO u) txBody = consumedValue <> txBody ^. mintValueTxBodyF
+getConsumedMaryValue pp lookupRefund (UTxO u) txBody = consumedValue <> txBody ^. mintValueTxBodyF
   where
     {- balance (txins tx ◁ u) + wbalance (txwdrls tx) + keyRefunds pp tx -}
     consumedValue =
       balance (UTxO (Map.restrictKeys u (txBody ^. inputsTxBodyL)))
         <> inject (refunds <> withdrawals)
-    refunds = keyTxRefunds pp dpstate txBody
+    refunds = keyCertsRefunds pp lookupRefund (txBody ^. certsTxBodyG)
```
TimSheard, line 63: Just a question. Is lookupRefund, a partially applied function? So getConsumedMaryValue  does need the DPState?

eras/shelley/impl/src/Cardano/Ledger/Shelley/UTxO.hs
```diff
@@ -121,33 +127,52 @@ getShelleyScriptsNeeded u txBody =
...
 
+consumed ::
+  EraUTxO era =>
+  PParams era ->
+  DPState (EraCrypto era) ->
+  UTxO era ->
+  TxBody era ->
+  Value era
+consumed pp dpstate = getConsumedValue pp (lookupDepositDState (dpsDState dpstate))
+
```
TimSheard, line 138: Here is the partial application.

libs/cardano-ledger-api/test/Test/Cardano/Ledger/Api/Tx/Body.hs
TimSheard: Another question to think about is where will the DRep info reside in the state? Will we need a RState for Reps, and the DPState becomes the DPRState?

libs/cardano-ledger-api/test/Test/Cardano/Ledger/Api/Tx/Body.hs
```diff
@@ -0,0 +1,142 @@
...
+      balance (txouts txBody)
+        <+> inject (txBody ^. feeTxBodyL <+> totalTxDeposits pp dpstate txBody)
+    consumed =
+      (txBody ^. mintValueTxBodyF)
+        <> balance (UTxO (Map.restrictKeys u (txBody ^. inputsTxBodyL)))
+        <> inject (refunds <> withdrawals)
+    refunds = keyTxRefunds pp dpstate txBody
+    withdrawals = fold . unWithdrawals $ txBody ^. withdrawalsTxBodyL
+
+-- | Randomly lookup pool params and staking credentials to add them as unregestration and
```
TimSheard: unRegistration? or unReRegistration

libs/cardano-ledger-core/src/Cardano/Ledger/DPState.hs
```diff
@@ -183,6 +186,20 @@ toDStatePair DState {..} =
...
+        RDPair _ deposit <- UM.lookup k currentRewardDeposits
+        Just $! fromCompact deposit
+
+lookupRewardDState :: DState c -> (StakeCredential c -> Maybe Coin)
+lookupRewardDState dstate =
+  let currentRewardDeposits = RewardDeposits $ dsUnified dstate
+   in \k -> do
+        RDPair reward _ <- UM.lookup k currentRewardDeposits
+        Just $! fromCompact reward
+
```
TimSheard: Predefined, reusable, partially applied functions.  Might want to mention these are supplied purposely to construct the function arguments used by the new functions. Might want to list them in the comments as well.

TimSheard (approved):
Add a few comments about what's changed that makes passing the functions around instead of the DState.  Looks good to me.

eras/mary/impl/src/Cardano/Ledger/Mary/UTxO.hs
lehins: > Is lookupRefund, a partially applied function? So
Not necessarily. 

> So getConsumedMaryValue does need the DPState?

No. That is the whole point of converting this argument from just the DPState to a lookup function. There are couple of other use cases, besides what we used it for in the rules (where we have DPState readily available):
1. In cardano-cli, where a user can lookup only the refunds relevant to the transaction. So this would be a partially applied `lookup` function to some other Map instead of DPState, because cardano-cli does not have access to the full DPState.
* During creation of transactions for benchmarking. In that case there will be no registration/deregistration or delegation/undelegation certificates, so a simple `const Nothing` will suffice.

eras/shelley/impl/src/Cardano/Ledger/Shelley/UTxO.hs
lehins: Yes, the `consumed` function is used in UTXO rule only and is not part of the public API

libs/cardano-ledger-api/test/Test/Cardano/Ledger/Api/Tx/Body.hs
lehins: That is certainly a good question. Not sure renaming `DP` to `DPR` is worthwhile, considering that `DRep` is also delegation, just of a different kind. We'll solve this problem when we'll get to implementing it

libs/cardano-ledger-api/test/Test/Cardano/Ledger/Api/Tx/Body.hs
lehins: unRegistration :slightly_smiling_face: Problems with spelling. Fixed

libs/cardano-ledger-core/src/Cardano/Ledger/DPState.hs
lehins: * `lookupRewardDState` is currently not used anywhere and was added for completeness.

* `lookupDepositDState` is mentioned in the haddock of `evalBalanceTxBody`.

> Predefined, reusable, partially applied functions. 

FYI, the type signature was crafted to suggest that those are the use cases, but in the end those are just lookup functions and can be used anywhere. Note that these type signatures are equivalent:

* `lookupDepositDState :: DState c -> StakeCredential c -> Maybe Coin`
* `lookupRewardDState :: DState c -> StakeCredential c -> Maybe Coin`

> Might want to mention these are supplied purposely to construct the function arguments used by the new functions.

So, they don't have to be partially applied and can be used just as regular lookup function, therefore adding such comment would be misleading.

I will however add haddock for those two functions, cause that is always a good idea.

eras/mary/impl/src/Cardano/Ledger/Mary/UTxO.hs
TimSheard: OK, got it

eras/shelley/impl/src/Cardano/Ledger/Shelley/AdaPots.hs
TimSheard: got it

PR #3343, opened 2023-03-20
Ts fix cannonical

JaredCorduan:
# Description

This fixes the bug discovered in https://github.com/input-output-hk/cardano-node/issues/4826. This issue merits a full retrospective, which we will provide. For now, however, we are applying the patch to the main ledger branch, adding property tests, and providing a terse explanation in these PR notes.

The `Data.Map.Internal` functions `balanceR` and `balanceL` expect the size difference between their two arguments to be 0 or 1, which allows them to re-balance the underlying tree in constant time. The canonical map can, however, _shrink_ by one when it places the multi-assets in canonical form, resulting in the tree becoming unbalanced. Unlike `balanceR` and `balanceL`, `link` makes no assumptions about the relative balance, at the cost of spending log time re-balancing. The performance cost will not be noticeable.

The credit goes to @TimSheard for finding the problem so fast. Also many thanks to @lehins for his help in quickly narrowing things down.

replaces #3271

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/eras/shelley-ma/test-suite/test/Test/Cardano/Ledger/Mary/Value.hs
+++ b/eras/shelley-ma/test-suite/test/Test/Cardano/Ledger/Mary/Value.hs
@@ -9,10 +9,12 @@
 
 module Test.Cardano.Ledger.Mary.Value (valTests) where
 
+import Cardano.Crypto.Hash.Class (castHash, hashFromStringAsHex)
 import Cardano.Ledger.Coin (Coin (..))
 import Cardano.Ledger.Compactible (fromCompact, toCompact)
 import Cardano.Ledger.Crypto (StandardCrypto)
 import qualified Cardano.Ledger.Crypto as CC (Crypto)
+import Cardano.Ledger.Hashes (ScriptHash (..))
 import Cardano.Ledger.Mary.Value (
   AssetName (..),
   MaryValue (..),
@@ -22,6 +24,7 @@ import Cardano.Ledger.Mary.Value (
   lookupMultiAsset,
  )
 import Cardano.Ledger.Val (Val (..), invert)
+import Control.DeepSeq (rnf)
 import Control.Monad (replicateM)
 import Data.ByteString.Short (ShortByteString)
 import Data.CanonicalMaps (
@@ -29,8 +32,10 @@ import Data.CanonicalMaps (
   canonicalInsert,
   canonicalMapUnion,
  )
+import qualified Data.Group as G
 import Data.Map.Strict (empty, singleton)
 import qualified Data.Map.Strict as Map
+import Data.Maybe (fromMaybe)
 import Test.Cardano.Ledger.Mary.Arbitrary ()
 import Test.Cardano.Ledger.Shelley.Serialisation.EraIndepGenerators ()
 import Test.Cardano.Ledger.Shelley.Serialisation.Generators ()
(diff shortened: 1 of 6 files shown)


eras/shelley-ma/test-suite/test/Test/Cardano/Ledger/Mary/Value.hs
```diff
@@ -333,6 +340,86 @@ compactRoundTrip = forAll gen $ \v ->
...
+          , makeMultiAsset $ makeScriptHash "00"
+          ]
+        shuffle2 =
+          [ makeMultiAsset $ makeScriptHash "04"
+          , makeMultiAsset $ makeScriptHash "08"
+          ]
+        MultiAsset am = mconcat shuffle1
+        mm = mconcat [m | MultiAsset m <- shuffle1]
+        b = trace (unlines ["am ", show (valid am), showTree am, "mm:", show (valid mm), showTree mm]) $ mconcat shuffle2
+        boom = (MultiAsset mm) <> G.invert b
```
lehins: Redundant parenthesis
```suggestion
        boom = MultiAsset mm <> G.invert b
```

eras/shelley-ma/test-suite/test/Test/Cardano/Ledger/Mary/Value.hs
```diff
@@ -333,6 +340,86 @@ compactRoundTrip = forAll gen $ \v ->
...
+          , makeMultiAsset $ makeScriptHash "02"
+          , makeMultiAsset $ makeScriptHash "00"
+          ]
+        shuffle2 =
+          [ makeMultiAsset $ makeScriptHash "04"
+          , makeMultiAsset $ makeScriptHash "08"
+          ]
+        MultiAsset am = mconcat shuffle1
+        mm = mconcat [m | MultiAsset m <- shuffle1]
+        b = trace (unlines ["am ", show (valid am), showTree am, "mm:", show (valid mm), showTree mm]) $ mconcat shuffle2
```
lehins: We should not leave any traces that will pollute the tests output. If this is something that needs to be printed upon a test failure then we can do that, but it cannot be done with tracing that is resorted purely for debugging. I don't really think it is worth it though, since we know the problem is fixed. So I suggest we remove this:
```suggestion
        b = mconcat shuffle2
```

eras/shelley-ma/test-suite/test/Test/Cardano/Ledger/Mary/Value.hs
```diff
@@ -333,6 +340,86 @@ compactRoundTrip = forAll gen $ \v ->
...
+        boom = (MultiAsset mm) <> G.invert b
+     in boom == boom
```
lehins: We only need to check that there is no call to `error`, right? Therefore all we need to do is evaluate it to NF.
```suggestion
     in rnf boom
```

eras/shelley-ma/test-suite/test/Test/Cardano/Ledger/Mary/Value.hs
```diff
@@ -333,6 +340,86 @@ compactRoundTrip = forAll gen $ \v ->
...
+s12 = makeMultiAsset $ makeScriptHash "12"
+
+exampleMultiAssets :: [MultiAsset StandardCrypto]
+exampleMultiAssets = [s0, s1, s2, s3, s4, s5, s6, s7, s8, s9, s10, s11, s12]
+
+-- | Test that the subtraction of Multi-assets (and the underlying 'CanonicalMaps')
+-- is a total function.
+-- This was used to diagnose https://github.com/input-output-hk/cardano-node/issues/4826
+subtractionIsTotal :: TestTree
+subtractionIsTotal = testProperty "marymary" $
```
lehins: Interesting choice of test name :smile:
"Oh, Mary Mary"

eras/shelley-ma/test-suite/test/Test/Cardano/Ledger/Mary/Value.hs
```diff
@@ -333,6 +340,86 @@ compactRoundTrip = forAll gen $ \v ->
...
+          , makeMultiAsset $ makeScriptHash "05"
+          , makeMultiAsset $ makeScriptHash "04"
+          , makeMultiAsset $ makeScriptHash "07"
+          , makeMultiAsset $ makeScriptHash "02"
+          , makeMultiAsset $ makeScriptHash "00"
+          ]
+        shuffle2 =
+          [ makeMultiAsset $ makeScriptHash "04"
+          , makeMultiAsset $ makeScriptHash "08"
+          ]
```
lehins: Reduce duplication a bit. An even more concise way would be to just specify lists in terms of `[s10, s09, ...]`, but just in case if you prefer string values this will work too:
```suggestion
    let shuffle1 =
          [ makeMultiAsset $ makeScriptHash suffix
           | suffix <- ["10", 09", "11", "08", "01", "06", "03", "05", "04", "07", "02", "00"]]
          ]
        shuffle2 =
          [ makeMultiAsset $ makeScriptHash suffix
           | suffix <- ["04", "08"]
          ]
```

libs/cardano-data/cardano-data.cabal
```diff
@@ -1,6 +1,6 @@
 cabal-version:      3.0
 name:               cardano-data
-version:            1.0.0.0
+version:            1.1.0.0
```
lehins: This is not a breaking change, thus only a minor version bump is required:
```suggestion
version:            1.0.1.0
```

We also want to add a lower bound in mary, just to be on the safe side: `cardano-data >= 1.0.1`

eras/shelley-ma/test-suite/test/Test/Cardano/Ledger/Mary/Value.hs
JaredCorduan: sorry, I thought I fixed this, I'll `seq` it...

eras/shelley-ma/test-suite/test/Test/Cardano/Ledger/Mary/Value.hs
```diff
@@ -333,6 +340,86 @@ compactRoundTrip = forAll gen $ \v ->
 compactTest :: TestTree
 compactTest = testProperty "fromCompact . toCompact == id" compactRoundTrip
 
+-- | Create a script hash of length 28 with 27 leading zeros followed by one hex-encoded byte
+-- supplied by the caller.
+makeScriptHash :: String -> ScriptHash StandardCrypto
+makeScriptHash str = ScriptHash $ castHash (fromJust $ hashFromStringAsHex (pad <> str))
```
lehins: Usage of `fromJust` is banned :boom:  
```suggestion
makeScriptHash str =
  ScriptHash $ castHash (fromMaybe (error "Impossible") $ hashFromStringAsHex (pad <> str))
```

libs/cardano-data/CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Version history for `cardano-data`
 
+## 1.1.0.0
```
lehins:
```suggestion
## 1.0.1.0
```

libs/cardano-data/CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
...
+
+* Fix - A bug was fixed in the `canonicalInsert` function.
+  The bug manifested by creating an unbalanced tree in the `Data.Map` internals of the
+  'CanonicalMap', which can result in a crash.
```
lehins, line 7: Maybe adding a reference to the issue that describes the crash in the changelog would be a good thing to do?
https://github.com/input-output-hk/cardano-node/issues/4826

lehins (approved):
Looks great! Really happy to finally see this fix land in master :slightly_smiling_face:

eras/shelley-ma/test-suite/test/Test/Cardano/Ledger/Mary/Value.hs
lehins: you meed `deepseq` it, right? :see_no_evil:

lehins: :boom: == :boom:

JaredCorduan: that's the age old question of mankind!

eras/shelley-ma/test-suite/test/Test/Cardano/Ledger/Mary/Value.hs
JaredCorduan: yea, I swear I fixed all that along with the boom == boom

eras/shelley-ma/test-suite/test/Test/Cardano/Ledger/Mary/Value.hs
JaredCorduan: quite contrary... yet another thing I swear I got already...

PR #3347, opened 2023-03-22
Fix MultiAsset generation

### LINKED ISSUE #3312: Test failure on CI
A property test failure on CI. Looks like a bad asset name generator in the trace generator.
```
  Test.Cardano.Chain.Block.Model
ts_prop_generatedChainsAreValidated:            FAIL (91.57s)
       *** Failed! (after 97 tests):
       Failed to deserialize encoded:
       Encoder Version: Version 5
       Decoder Version: Version 5
       Decoder error: Deserialisation failure while decoding Annotator (AlonzoTx (AlonzoEra StandardCrypto)).
       CBOR failed with error: DeserialiseFailure 566114 "An error occured while decoding asset name exceeds 32 bytes:.\nError: 
....
Use --quickcheck-replay=431035 to reproduce.
Use -p '$0=="Alonzo tests.Alonzo CBOR round-trip.alonzo/Tx"' to rerun this test only.
```

aniketd:
# Description

Closes #3312 

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/eras/mary/impl/src/Cardano/Ledger/Mary/Value.hs
+++ b/eras/mary/impl/src/Cardano/Ledger/Mary/Value.hs
@@ -334,7 +334,7 @@ encodeMultiAssetMaps (MultiAsset m) = encCBOR m
 decodeMultiAssetMaps :: Crypto c => (forall t. Decoder t Integer) -> Decoder s (MultiAsset c)
 decodeMultiAssetMaps decodeAmount = do
   ma <- decodeMap decCBOR (decodeMap decCBOR decodeAmount)
-  -- compact form inequality: 
+  -- compact form inequality:
   --   8n (Word64) + 2n (Word16) + 2n (Word16) + 28p (policy ids) + sum of lengths of unique asset names <= 65535
   -- where: n = number of assets, p = number of unique policy ids
   let numUniqPolicies = length $ keysSet ma
@@ -342,9 +342,8 @@ decodeMultiAssetMaps decodeAmount = do
       sumLengthsUniqAssetNames = sum . fmap SBS.length . nub . foldr' (<>) [] $ fmap assetName . keys <$> ma
       highestOffset = 12 * numAssets + 28 * numUniqPolicies + sumLengthsUniqAssetNames
   if highestOffset >= 65535
-    then 
-      fail "MultiAsset too big to compact, has too many assets. Ideal number is < 910."
-    else 
+    then fail "MultiAsset too big to compact, has too many assets. Ideal number is < 910."
+    else
       ifDecoderVersionAtLeast
         (natVersion @9)
         (MultiAsset ma <$ forM_ ma (\m -> when (Map.null m) $ fail "Empty Assets are not allowed"))
--- a/eras/mary/impl/src/Cardano/Ledger/Mary/Value.hs
+++ b/eras/mary/impl/src/Cardano/Ledger/Mary/Value.hs
@@ -81,13 +81,13 @@ import Data.CanonicalMaps (
 import Data.Foldable (foldMap')
 import Data.Group (Abelian, Group (..))
 import Data.Int (Int64)
-import Data.List (sortOn)
+import Data.List (nub, sortOn)
 import Data.Map (Map)
 import Data.Map.Internal (
   link,
   link2,
  )
-import Data.Map.Strict (assocs)
+import Data.Map.Strict (assocs, foldr', keys, keysSet)
 import qualified Data.Map.Strict as Map
 import Data.Maybe (fromJust)
 import qualified Data.Primitive.ByteArray as BA
@@ -334,10 +334,21 @@ encodeMultiAssetMaps (MultiAsset m) = encCBOR m
 decodeMultiAssetMaps :: Crypto c => (forall t. Decoder t Integer) -> Decoder s (MultiAsset c)
 decodeMultiAssetMaps decodeAmount = do
   ma <- decodeMap decCBOR (decodeMap decCBOR decodeAmount)
-  ifDecoderVersionAtLeast
-    (natVersion @9)
-    (MultiAsset ma <$ forM_ ma (\m -> when (Map.null m) $ fail "Empty Assets are not allowed"))
-    (pure $ MultiAsset $ prune ma)
+  -- compact form inequality: 
+  --   8n (Word64) + 2n (Word16) + 2n (Word16) + 28p (policy ids) + sum of lengths of unique asset names <= 65535
+  -- where: n = number of assets, p = number of unique policy ids
+  let numUniqPolicies = length $ keysSet ma
+      numAssets = sum $ length <$> ma
+      sumLengthsUniqAssetNames = sum . fmap SBS.length . nub . foldr' (<>) [] $ fmap assetName . keys <$> ma
+      highestOffset = 12 * numAssets + 28 * numUniqPolicies + sumLengthsUniqAssetNames
+  if highestOffset >= 65535
+    then 
+      fail "MultiAsset too big to compact, has too many assets. Ideal number is < 910."
+    else 
+      ifDecoderVersionAtLeast
+        (natVersion @9)
+        (MultiAsset ma <$ forM_ ma (\m -> when (Map.null m) $ fail "Empty Assets are not allowed"))
+        (pure $ MultiAsset $ prune ma)
 
 decodeMultiAssetMint :: Crypto c => Decoder s (MultiAsset c)
 decodeMultiAssetMint = decodeMultiAssetMaps decodeIntegerBounded64
--- a/eras/mary/impl/src/Cardano/Ledger/Mary/Value.hs
+++ b/eras/mary/impl/src/Cardano/Ledger/Mary/Value.hs
@@ -81,13 +81,13 @@ import Data.CanonicalMaps (
 import Data.Foldable (foldMap')
 import Data.Group (Abelian, Group (..))
 import Data.Int (Int64)
-import Data.List (nub, sortOn)
+import Data.List (sortOn)
 import Data.Map (Map)
 import Data.Map.Internal (
   link,
   link2,
  )
-import Data.Map.Strict (assocs, foldr', keys, keysSet)
+import Data.Map.Strict (assocs)
 import qualified Data.Map.Strict as Map
 import Data.Maybe (fromJust)
 import qualified Data.Primitive.ByteArray as BA
@@ -336,13 +336,13 @@ decodeMultiAssetMaps decodeAmount = do
   ma <- decodeMap decCBOR (decodeMap decCBOR decodeAmount)
   -- compact form inequality:
   --   8n (Word64) + 2n (Word16) + 2n (Word16) + 28p (policy ids) + sum of lengths of unique asset names <= 65535
-  -- where: n = number of assets, p = number of unique policy ids
-  let numUniqPolicies = length $ keysSet ma
-      numAssets = sum $ length <$> ma
-      sumLengthsUniqAssetNames = sum . fmap SBS.length . nub . foldr' (<>) [] $ fmap assetName . keys <$> ma
-      highestOffset = 12 * numAssets + 28 * numUniqPolicies + sumLengthsUniqAssetNames
-  if highestOffset >= 65535
-    then fail "MultiAsset too big to compact, has too many assets. Ideal number is < 910."
+  -- maximum size for the asset name is 32 bytes, so:
+  -- 8n + 2n + 2n + 28p + 32n <= 65535
+  -- where: n = total number of assets, p = number of unique policy ids
+  let numPolicies = Map.size ma
+      numAssetNames = sum $ length <$> ma
+  if 44 * numAssetNames + 28 * numPolicies > 65535
+    then fail "MultiAsset too big to compact"
     else
       ifDecoderVersionAtLeast
         (natVersion @9)
--- a/eras/mary/impl/src/Cardano/Ledger/Mary/Value.hs
+++ b/eras/mary/impl/src/Cardano/Ledger/Mary/Value.hs
@@ -64,6 +64,7 @@ import Cardano.Ledger.Shelley.Scripts (ScriptHash (..))
 import Cardano.Ledger.TreeDiff (Expr (App), ToExpr (..))
 import Cardano.Ledger.Val (Val (..))
 import Control.DeepSeq (NFData (..), deepseq, rwhnf)
+import Control.Exception (assert)
 import Control.Monad (forM_, when)
 import Control.Monad.ST (runST)
 import Data.Aeson (FromJSON, FromJSONKey, ToJSON (..), object, (.=))
@@ -339,7 +340,7 @@ decodeMultiAssetMaps decodeAmount = do
   -- maximum size for the asset name is 32 bytes, so:
   -- 8n + 2n + 2n + 28p + 32n <= 65535
   -- where: n = total number of assets, p = number of unique policy ids
-  let numPolicies = Map.size ma
+  let numPolicies = length ma
       numAssetNames = sum $ length <$> ma
   if 44 * numAssetNames + 28 * numPolicies > 65535
     then fail "MultiAsset too big to compact"
@@ -571,8 +572,8 @@ to ::
 to (MaryValue ada (MultiAsset m))
   | Map.null m =
       CompactValueAdaOnly . CompactCoin <$> integerToWord64 ada
-to v = do
-  c <- integerToWord64 ada
+to v@(MaryValue _ (MultiAsset ma)) = do
+  c <- assert (44 * sum (length <$> ma) + 28 * length ma < 65535) (integerToWord64 ada)
   -- Here we convert the (pid, assetName, quantity) triples into
   -- (Int, (Word16,Word16,Word64))
   -- These represent the index, pid offset, asset name offset, and quantity.
@@ -707,7 +708,8 @@ representationSize xs = abcRegionSize + pidBlockSize + anameBlockSize
 from :: forall c. (Crypto c) => CompactValue c -> MaryValue c
 from (CompactValueAdaOnly (CompactCoin c)) = MaryValue (fromIntegral c) (MultiAsset Map.empty)
 from (CompactValueMultiAsset (CompactCoin c) numAssets rep) =
-  valueFromList (fromIntegral c) triples
+  let mv@(MaryValue _ (MultiAsset maMap)) = valueFromList (fromIntegral c) triples
+   in assert (44 * sum (length <$> maMap) + 28 * length maMap < 65535) mv
   where
     n = fromIntegral numAssets
 
--- a/eras/mary/impl/src/Cardano/Ledger/Mary/Value.hs
+++ b/eras/mary/impl/src/Cardano/Ledger/Mary/Value.hs
@@ -30,6 +30,7 @@ module Cardano.Ledger.Mary.Value (
   valueFromList,
   ToExpr (..),
   CompactValue (..),
+  isMultiAssetSmallEnough,
 )
 where
 
@@ -335,20 +336,13 @@ encodeMultiAssetMaps (MultiAsset m) = encCBOR m
 decodeMultiAssetMaps :: Crypto c => (forall t. Decoder t Integer) -> Decoder s (MultiAsset c)
 decodeMultiAssetMaps decodeAmount = do
   ma <- decodeMap decCBOR (decodeMap decCBOR decodeAmount)
-  -- compact form inequality:
-  --   8n (Word64) + 2n (Word16) + 2n (Word16) + 28p (policy ids) + sum of lengths of unique asset names <= 65535
-  -- maximum size for the asset name is 32 bytes, so:
-  -- 8n + 2n + 2n + 28p + 32n <= 65535
-  -- where: n = total number of assets, p = number of unique policy ids
-  let numPolicies = length ma
-      numAssetNames = sum $ length <$> ma
-  if 44 * numAssetNames + 28 * numPolicies > 65535
-    then fail "MultiAsset too big to compact"
-    else
+  if isMultiAssetSmallEnough (MultiAsset ma)
+    then
       ifDecoderVersionAtLeast
         (natVersion @9)
         (MultiAsset ma <$ forM_ ma (\m -> when (Map.null m) $ fail "Empty Assets are not allowed"))
         (pure $ MultiAsset $ prune ma)
+    else fail "MultiAsset too big to compact"
 
 decodeMultiAssetMint :: Crypto c => Decoder s (MultiAsset c)
 decodeMultiAssetMint = decodeMultiAssetMaps decodeIntegerBounded64
@@ -572,8 +566,8 @@ to ::
 to (MaryValue ada (MultiAsset m))
   | Map.null m =
       CompactValueAdaOnly . CompactCoin <$> integerToWord64 ada
-to v@(MaryValue _ (MultiAsset ma)) = do
-  c <- assert (44 * sum (length <$> ma) + 28 * length ma < 65535) (integerToWord64 ada)
+to v@(MaryValue _ ma) = do
+  c <- assert (isMultiAssetSmallEnough ma) (integerToWord64 ada)
   -- Here we convert the (pid, assetName, quantity) triples into
   -- (Int, (Word16,Word16,Word64))
   -- These represent the index, pid offset, asset name offset, and quantity.
@@ -687,6 +681,18 @@ to v@(MaryValue _ (MultiAsset ma)) = do
       q' <- integerToWord64 q
       pure (pidOffset pid, assetNameOffset aname, q')
 
+-- | Unlike `representationSize`, this function cheaply checks whether
+-- any offset within a MultiAsset compact representation is likely to overflow Word16.
+--
+-- compact form inequality:
+--   8n (Word64) + 2n (Word16) + 2n (Word16) + 28p (policy ids) + sum of lengths of unique asset names <= 65535
+-- maximum size for the asset name is 32 bytes, so:
+-- 8n + 2n + 2n + 28p + 32n <= 65535
+-- where: n = total number of assets, p = number of unique policy ids
+isMultiAssetSmallEnough :: MultiAsset c -> Bool
+isMultiAssetSmallEnough (MultiAsset ma) =
+  44 * sum (length <$> ma) + 28 * length ma <= 65535
+
 representationSize ::
   forall c.
   Crypto c =>
@@ -708,8 +714,8 @@ representationSize xs = abcRegionSize + pidBlockSize + anameBlockSize
 from :: forall c. (Crypto c) => CompactValue c -> MaryValue c
 from (CompactValueAdaOnly (CompactCoin c)) = MaryValue (fromIntegral c) (MultiAsset Map.empty)
 from (CompactValueMultiAsset (CompactCoin c) numAssets rep) =
-  let mv@(MaryValue _ (MultiAsset maMap)) = valueFromList (fromIntegral c) triples
-   in assert (44 * sum (length <$> maMap) + 28 * length maMap < 65535) mv
+  let mv@(MaryValue _ ma) = valueFromList (fromIntegral c) triples
+   in assert (isMultiAssetSmallEnough ma) mv
   where
     n = fromIntegral numAssets
 
--- a/eras/mary/impl/src/Cardano/Ledger/Mary/Value.hs
+++ b/eras/mary/impl/src/Cardano/Ledger/Mary/Value.hs
@@ -92,6 +92,7 @@ import Data.Map.Internal (
 import Data.Map.Strict (assocs)
 import qualified Data.Map.Strict as Map
 import Data.Maybe (fromJust)
+import qualified Data.Monoid as M (Sum (Sum, getSum))
 import qualified Data.Primitive.ByteArray as BA
 import Data.Proxy (Proxy (..))
 import qualified Data.Semigroup as Semigroup (Sum (..))
@@ -691,7 +692,7 @@ to v@(MaryValue _ ma) = do
 -- where: n = total number of assets, p = number of unique policy ids
 isMultiAssetSmallEnough :: MultiAsset c -> Bool
 isMultiAssetSmallEnough (MultiAsset ma) =
-  44 * sum (length <$> ma) + 28 * length ma <= 65535
+  44 * M.getSum (foldMap' (M.Sum . length) ma) + 28 * length ma <= 65535
 
 representationSize ::
   forall c.
--- a/eras/mary/impl/test/Test/Cardano/Ledger/Mary/ValueSpec.hs
+++ b/eras/mary/impl/test/Test/Cardano/Ledger/Mary/ValueSpec.hs
@@ -71,10 +71,12 @@ spec = do
     prop "Canonical generator" $
       \(ma :: MaryValue StandardCrypto) ->
         fromCompact @(MaryValue StandardCrypto) (fromJust (toCompact ma)) === ma
-    prop "Failing generator" $
-      forAll (genMaryValue (genMultiAssetToFail 2000)) $
-        \(ma :: MaryValue StandardCrypto) ->
-          fromCompact @(MaryValue StandardCrypto) (fromJust (toCompact ma)) =/= ma
+    it "Failing generator" $
+      expectFailure $
+        property $
+          forAll (genMaryValue genMultiAssetToFail) $
+            \(ma :: MaryValue StandardCrypto) ->
+              fromCompact @(MaryValue StandardCrypto) (fromJust (toCompact ma)) === ma
 
 instance IsString AssetName where
   fromString = AssetName . either error SBS.toShort . BS16.decode . BS8.pack
--- a/eras/mary/impl/test/Test/Cardano/Ledger/Mary/ValueSpec.hs
+++ b/eras/mary/impl/test/Test/Cardano/Ledger/Mary/ValueSpec.hs
@@ -67,16 +67,21 @@ spec = do
       prop "Empty MaryValue fails for Conway" $
         forAll (genMaryValue (genEmptyMultiAsset @StandardCrypto)) $
           roundTripCborRangeFailureExpectation (natVersion @9) maxBound
+      it "Too many assets should fail" $
+        expectFailure $
+          property $
+            forAll (genMaryValue (genMultiAssetToFail @StandardCrypto)) $
+              roundTripCborFailureExpectation
   describe "MaryValue compacting" $ do
     prop "Canonical generator" $
       \(ma :: MaryValue StandardCrypto) ->
-        fromCompact @(MaryValue StandardCrypto) (fromJust (toCompact ma)) === ma
+        fromCompact (fromJust (toCompact ma)) === ma
     it "Failing generator" $
       expectFailure $
         property $
-          forAll (genMaryValue genMultiAssetToFail) $
-            \(ma :: MaryValue StandardCrypto) ->
-              fromCompact @(MaryValue StandardCrypto) (fromJust (toCompact ma)) === ma
+          forAll (genMaryValue (genMultiAssetToFail @StandardCrypto)) $
+            \ma ->
+              fromCompact (fromJust (toCompact ma)) === ma
 
 instance IsString AssetName where
   fromString = AssetName . either error SBS.toShort . BS16.decode . BS8.pack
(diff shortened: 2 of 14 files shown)


JaredCorduan (approved):
looks great to me, thank you @aniketd !

eras/mary/impl/testlib/Test/Cardano/Ledger/Mary/Arbitrary.hs
```diff
@@ -26,17 +26,25 @@ import Data.String (IsString (fromString))
...
     AssetName
-      <$> oneof
-        [ do
-            len <- choose (1, 32)
-            genShortByteString len
-        , -- We need duplicates for quality tests
-          elements digitByteStrings
+      <$> frequency
+        [ (1, elements digitByteStrings)
+        , (99, genShortByteString =<< choose (1, 32))
```
lehins: Generating duplicates is important, so it should have a lot more than 1% chance. That's why it was 50/50 before. It could probably be as low as 30/70, but definitely not 1/99.

eras/mary/impl/test/Test/Cardano/Ledger/Mary/ValueSpec.hs
```diff
@@ -65,6 +67,14 @@ spec = do
       prop "Empty MaryValue fails for Conway" $
         forAll (genMaryValue (genEmptyMultiAsset @StandardCrypto)) $
           roundTripCborRangeFailureExpectation (natVersion @9) maxBound
+  describe "MaryValue compacting" $ do
+    prop "Canonical generator" $
+      \(ma :: MaryValue StandardCrypto) ->
+        fromCompact @(MaryValue StandardCrypto) (fromJust (toCompact ma)) === ma
+    prop "Failing generator" $
+      forAll (genMaryValue (genMultiAssetToFail 2000)) $
```
lehins: Such test is dangerous, because it reads out of bounds memory. In other words there is a chance that it could be killed with a segfault, which is not something we want to see in tests.

However, that generator `genMultiAssetToFail`  will be useful for testing deserialization failure. Also instead of `2000` we should be using the smallest possible value that triggers the failure or even better generate it randomly too inside `genMultiAssetToFail`, eg.: 
```
genMultiAssetToFail genAmount = do
  Positive n <- arbitrary
  let triplesSize = n + 910
  ...
```

eras/mary/impl/testlib/Test/Cardano/Ledger/Mary/Arbitrary.hs
```diff
@@ -89,9 +93,52 @@ instance Crypto c => Arbitrary (PolicyID c) where
...
+-- NOTE: There are some conditions due to which exceeding this number may not
+-- result in a guaranteed failure to compact without overflow, because, during compacting
+--   1. The asset names and policy ids are deduplicated
+--   2. Not all generated asset names are 32-bytes long
+-- But, exceeding this number does make the probability of causing overflow > 0.
 genMultiAsset :: Crypto c => Gen Integer -> Gen (MultiAsset c)
 genMultiAsset genAmount =
-  MultiAsset <$> genNonEmptyMap arbitrary (genNonEmptyMap arbitrary genAmount)
+  multiAssetFromListBounded
+    <$> resize 910 (listOf $ genMultiAssetTriple $ fromIntegral <$> genAmount)
```
lehins: This is not gonna work, because PolicyIDs are generated uniformly. This means all of them would only have a single AssetName. Also generating asset names on the order of hundreds as a common case is going to destroy the runtime and is guaranteed to be killed with OOM on some serialization roundtrippping, like Block for example.

That being said, I think it would be useful to generate MultiAssets in two different ways:

So, I think, it will be better this way:
```suggestion
  ma <- oneof
    [ genNonEmptyMap arbitrary (genNonEmptyMap arbitrary genAmount)
    , multiAssetFromListBounded  <$> (listOf $ genMultiAssetTriple $ fromIntegral <$> genAmount)
    ]
  if sum (length <$> ma) > 910
    then scale (`div` 2) $ genMultiAsset genAmount
    else MultiAsset <$> ma
```

eras/mary/impl/testlib/Test/Cardano/Ledger/Mary/Arbitrary.hs
```diff
@@ -89,9 +93,52 @@ instance Crypto c => Arbitrary (PolicyID c) where
...
+-- For negative tests
+genMultiAssetToFail :: Crypto c => Int -> Gen (MultiAsset c)
+genMultiAssetToFail triplesSize =
+  multiAssetFromListBounded @Int64
+    <$> vectorOf
+      triplesSize
+      ( (,,)
+          <$> arbitrary
+          <*> (AssetName <$> genShortByteString 32)
+          <*> (fromIntegral <$> choose (1 :: Int, maxBound))
```
lehins:
```suggestion
          <*> (fromIntegral <$> chooseInt (1, maxBound))
```

lehins (changes requested):
Great job on investigating this problem and figuring out the value of 910. Taking a quick peek at the compaction implementation it does seem to be the correct value indeed.

There are some problems that I commented on in the PR itself, but the overall direction is totally correct.

We are still missing fixes to the deserializer, but I am sure you knew this already anyways.

eras/mary/impl/test/Test/Cardano/Ledger/Mary/ValueSpec.hs
aniketd: Awesome! Thanks for this suggestion, I will try this out.

eras/mary/impl/testlib/Test/Cardano/Ledger/Mary/Arbitrary.hs
aniketd: Noted! :+1:

eras/mary/impl/testlib/Test/Cardano/Ledger/Mary/Arbitrary.hs
aniketd: > This is not gonna work, because PolicyIDs are generated uniformly. This means all of them would only have a single AssetName.

:facepalm: I didn't think of this! :exploding_head: Thanks!

> Also generating asset names on the order of hundreds as a common case is going to destroy the runtime and is guaranteed to be killed with OOM on some serialization roundtrippping, like Block for example.

:+1: :100: :memo:

libs/cardano-ledger-binary/testlib/Test/Cardano/Ledger/Binary/RoundTrip.hs
```diff
@@ -156,37 +159,43 @@ roundTripRangeFailureExpectation ::
...
 
 embedTripFailureExpectation ::
   (Typeable b, HasCallStack) =>
   Trip a b ->
   a ->
   Expectation
-embedTripFailureExpectation trip = embedTripRangeFailureExpectation trip minBound maxBound
+embedTripFailureExpectation trip = embedTripRangeFailureExpectation Maximal trip minBound maxBound
+
+data FailureVerbosity = Maximal | Minimal
```
lehins: We don't want to hide the output upon a failure, even if it is very large. All of the output is written to a log file that can be inspected when a test fails. So, please remove this type and its usage.

eras/mary/impl/testlib/Test/Cardano/Ledger/Mary/Arbitrary.hs
```diff
@@ -89,9 +96,67 @@ instance Crypto c => Arbitrary (PolicyID c) where
...
+    <$> resize
+      triplesSize
+      ( vectorOf
+          triplesSize
+          ( (,,)
+              <$> arbitrary
+              <*> (AssetName <$> genShortByteString 32)
+              <*> (fromIntegral <$> chooseInt (1, maxBound))
+          )
+      )
```
lehins: It is essentially a linear inequality with two unknowns. We generate one and solve for the other one. If my math is right it should be something along the lines:
```suggestion
  (numPolicyId, numAssetName) <- oneof
     [ do
        Positive numAsssetNames <- arbitary
        let minNumPolicyId = (65535 - 44 * numAsssetNames) `div` 28
        numPolicyIds <- (minNumPolicyId +) . getPositive <$> arbitrary
        pure (numPolicyIds, numAsssetNames)
    , do
        Positive numPolicyIds <- arbitary
        let minNumAssetNames = (65535 - 28 * numPolicyIds) `div` 44
        numAssetNames <- (minNumAssetNames +) . getPositive <$> arbitrary
        pure (numPolicyIds, numAsssetNames)
     ]
  -- Here we generate separately a list of asset names and a list of policyids and
  -- randomly shuffle them into a MultiAsset
  ...

```

eras/mary/impl/test/Test/Cardano/Ledger/Mary/ValueSpec.hs
```diff
@@ -47,24 +55,48 @@ spec = do
...
           roundTripCborRangeFailureExpectation (natVersion @9) maxBound
+      it "Too many assets should fail" $
+        expectFailure $
+          property $
+            forAll
+              (genMaryValue (genMultiAssetToFail @StandardCrypto))
+              ( embedTripRangeFailureExpectation @(MaryValue StandardCrypto) @(MaryValue StandardCrypto)
+                  Minimal
+                  cborTrip
+                  minBound
```
lehins: The starting version could be switched to the era when MaryValue was introduced. We don't need to test eras prior to Mary
```suggestion
                  (eraProtVerLow @Mary)
```

eras/mary/impl/test/Test/Cardano/Ledger/Mary/ValueSpec.hs
```diff
@@ -47,24 +55,48 @@ spec = do
...
+                  maxBound
+              )
+
+  describe "MaryValue compacting" $ do
+    prop "Canonical generator" $
+      \(ma :: MaryValue StandardCrypto) ->
+        fromCompact (fromJust (toCompact ma)) == ma
```
lehins: This will produce more informative failure messages
```suggestion
        fromCompact (fromJust (toCompact ma)) `shouldBe` ma
```

eras/mary/impl/test/Test/Cardano/Ledger/Mary/ValueSpec.hs
```diff
@@ -47,24 +55,48 @@ spec = do
...
+    it "Failing generator" $
+      expectFailure $
+        property $
+          forAll (genMaryValue (genMultiAssetToFail @StandardCrypto)) $
+            \ma ->
+              fromCompact (fromJust (toCompact ma)) == ma
```
lehins: I don't get it. How does this test pass?

eras/mary/impl/src/Cardano/Ledger/Mary/Value.hs
```diff
@@ -334,10 +334,20 @@ encodeMultiAssetMaps (MultiAsset m) = encCBOR m
...
-    (MultiAsset ma <$ forM_ ma (\m -> when (Map.null m) $ fail "Empty Assets are not allowed"))
-    (pure $ MultiAsset $ prune ma)
+  -- compact form inequality:
+  --   8n (Word64) + 2n (Word16) + 2n (Word16) + 28p (policy ids) + sum of lengths of unique asset names <= 65535
+  -- where: n = number of assets, p = number of unique policy ids
+  let numUniqPolicies = length $ keysSet ma
+      numAssets = sum $ length <$> ma
+      sumLengthsUniqAssetNames = sum . fmap SBS.length . nub . foldr' (<>) [] $ fmap assetName . keys <$> ma
+      highestOffset = 12 * numAssets + 28 * numUniqPolicies + sumLengthsUniqAssetNames
+  if highestOffset >= 65535
```
lehins: All we need is a safe upper bound on the number of assets names and policy ids, calculation does not have to be exact. In other words, we want to avoid expensive computations (like `nub`), while not disallowing realistic values.

```suggestion
  let numPolicies = Map.size ma
      numAssetNames = sum $ length <$> ma
  if 44 * numAssetNames + 28 * numPolicies > 65535
```

Of course, this would mean that we are over restricting some of the potentially valid values. So, what's the smallest over restricted value. The worst case that can happen is when we have maximum number of policies, all of which have the same 1 byte long asset name, in which case we get:
```
13 * 1 + 28 * numPolicies <= 65535
numPolicies <= 2340
```
This would mean that serialized version of the `MaryValue` would be around 75KiB, which is around x15 higher than what is allowed in the rules. So, I think we are good with not checking for duplicates and counting the actual length of each individual asset.

 By the way I just stumbled upon a function that can compute this value exactly: `representationSize` :slightly_smiling_face:

eras/mary/impl/src/Cardano/Ledger/Mary/Value.hs
```diff
@@ -334,10 +334,20 @@ encodeMultiAssetMaps (MultiAsset m) = encCBOR m
 decodeMultiAssetMaps :: Crypto c => (forall t. Decoder t Integer) -> Decoder s (MultiAsset c)
 decodeMultiAssetMaps decodeAmount = do
   ma <- decodeMap decCBOR (decodeMap decCBOR decodeAmount)
-  ifDecoderVersionAtLeast
-    (natVersion @9)
-    (MultiAsset ma <$ forM_ ma (\m -> when (Map.null m) $ fail "Empty Assets are not allowed"))
-    (pure $ MultiAsset $ prune ma)
+  -- compact form inequality:
+  --   8n (Word64) + 2n (Word16) + 2n (Word16) + 28p (policy ids) + sum of lengths of unique asset names <= 65535
+  -- where: n = number of assets, p = number of unique policy ids
```
lehins:
```suggestion
  --   8n (Word64) + 2n (Word16) + 2n (Word16) + 28p (policy ids) + sum of lengths of unique asset names <= 65535
  -- maximum size for the asset name is 32bytes, so:
  -- 8n + 2n + 2n + 28p + 32n <= 65535
  -- 44n + 28p <= 65535
  -- where: n = number of assets, p = number of unique policy ids
```

eras/mary/impl/testlib/Test/Cardano/Ledger/Mary/Arbitrary.hs
```diff
@@ -89,9 +96,67 @@ instance Crypto c => Arbitrary (PolicyID c) where
...
+--   2. All offsets (including those for AssetName and PolicyID) are relative to the whole
+--      of the representation (binary blob) rather than the start of their respective regions.
+--   3. If the offsets exceed their maxBounds, they will overflow.
+--   4. So, we need to ensure that at least the last of the offsets (AssetName offsets) do
+--      not exceed 65535.
+--   5. With `n` as the total number of assets, the inequality to be satisfied is thus:
+--           8n -- Word64 asset quantities
+--        +  2n -- Word16 policy id offsets
+--        +  2n -- Word16 asset name offsets
+--        + 28n -- 28-byte policy ids
```
lehins: Now I am realizing that this calculation of 910 is wrong, because in this computation you are assuming number of policyIDs an AssetNames should be the same. This should be:
```suggestion
--        + 28p -- 28-byte policy ids
```
where `p` is the number of policyIds

lehins (changes requested):
More work is needed in order to get this right.

eras/mary/impl/testlib/Test/Cardano/Ledger/Mary/Arbitrary.hs
aniketd: Absolutely, thanks Alexey! This carried over from the thinking of `triples`. And I think it is easy assuming that each policy id has only one asset. There are some properties in reality that I had thought about:
1. If policies cannot have empty assets, then
    1. number of policies is always <= number of assets
    2. in the worst case number of policies == number of assets
    3. and, this number can be atmost 910.
1. If policies can have empty assets
    1. number of policies can be > number of assets
    2. the equality stands inapplicable
    3. but we can still consider constraining number of policies as well as assets to be < 910, by using the generator for triples, but because, as you pointed out earlier, it is uniform, we can't use only that.

I also wondered whether there was a better way to approach the implementation of generators for this problem than recursively scaling down, like you suggested earlier, but I could not think of any approach yet. Generating policy ids and assets separately then begets the question of how to randomly distribute those assets across the policies. One could do a first pass by perhaps iteratively generating assets by `chooseInt (1, 5)` for each policy, and keeping the count to not exceed a number, but I'm not sure that will be too restrictive and hide good corner cases.

lehins: > 1. If policies cannot have empty assets, then

That is correct, it is enforced by the deserializer and canonical monoid instance

> * number of policies is always <= number of assets

Depends from which point of view you actually look at it. From the point of view of the MultiAsset representation and the deserializer, then yes. However, if you look from the point of view of the compactor/decompactor, in the worst case the number of assets could be 1, while number of policies could be up to 2340. This happens when each PolicyID contains the same single AssetName. This clearly breaks this assumption.

> * in the worst case number of policies == number of assets

That's not the worst case. The worst case is when there is only one policyId and maximum number of unique 32byte AssetNames. With such setup  there can be at most 1488 AssetNames (i.e. entries in the triplet)


> 2. If policies can have empty assets

This can't happen, you've already fixed the deserializer to disallow empty MultiAssets in a previous PR

eras/mary/impl/test/Test/Cardano/Ledger/Mary/ValueSpec.hs
aniketd: I was looking for a function that would make the test pass when the property failed for _any_ input. I found `expectFailure`.

eras/mary/impl/src/Cardano/Ledger/Mary/Value.hs
aniketd: That's really useful! Thanks for pointing out!

eras/mary/impl/test/Test/Cardano/Ledger/Mary/ValueSpec.hs
lehins: I see, I didn't notice the `expectFailure`, So, you expect it to fail. However, as I said earlier, creating tests that rely on buffer overflows to fail is a bad idea! Remember that `fromCompact` will read memory that it is not suppose to and if it so happens that the memory doesn't belong to the process, the process (test suite in this case) will crash with a segfault. We don't want our tests to crash with segfaults.

That being said, there is a way to make such a test. What you need is an `assert` inside the `from` function and a test that expects an `assert` to be raised on bad input

eras/mary/impl/testlib/Test/Cardano/Ledger/Mary/Arbitrary.hs
aniketd: This ends up generating not enough large MultiAssets, unfortunately.

eras/mary/impl/src/Cardano/Ledger/Mary/Value.hs
aniketd: Would this be a good place to directly call `representationSize` at?

eras/mary/impl/testlib/Test/Cardano/Ledger/Mary/Arbitrary.hs
```diff
@@ -89,9 +98,88 @@ instance Crypto c => Arbitrary (PolicyID c) where
...
+--   2. Not all generated asset names are 32-bytes long
+-- But, exceeding this number does make the probability of causing overflow > 0.
+genMultiAsset :: forall c. Crypto c => Gen Integer -> Gen (MultiAsset c)
+genMultiAsset genAmount = do
+  MultiAsset ma <-
+    oneof
+      [ MultiAsset <$> genNonEmptyMap arbitrary (genNonEmptyMap arbitrary genAmount)
+      , multiAssetFromListBounded <$> listOf1 (genMultiAssetTriple $ fromIntegral <$> genAmount)
+      ]
+  if sum (length <$> ma) > 910
```
lehins: This needs to be fixed to match the deserializer

eras/mary/impl/testlib/Test/Cardano/Ledger/Mary/Arbitrary.hs
```diff
@@ -89,9 +98,88 @@ instance Crypto c => Arbitrary (PolicyID c) where
...
+    then scale (`div` 2) $ genMultiAsset genAmount
+    else pure $ MultiAsset ma
+
+-- | For tests, because `insertMultiAsset` called through `genMultiAsset` filters out zero values
+genMultiAssetZero :: Crypto c => Gen (MultiAsset c)
+genMultiAssetZero = MultiAsset <$> genNonEmptyMap arbitrary (genNonEmptyMap arbitrary (pure 0))
+
+-- For negative tests, we need a definite generator that causes overflow
```
lehins: We don't want to cause overflow. We want negative tests that check that deserializer fails as expected

eras/mary/impl/testlib/Test/Cardano/Ledger/Mary/Arbitrary.hs
```diff
@@ -89,9 +98,88 @@ instance Crypto c => Arbitrary (PolicyID c) where
...
+      , do
+          Positive numPolicyIds <- arbitrary
+          let minNumAssetNames = (65535 - 28 * numPolicyIds) `div` 44
+          numAssetNames <- (minNumAssetNames +) . getPositive <$> arbitrary
+          pure (numPolicyIds, numAssetNames)
+      ]
+  -- Here we generate separately a list of asset names and a list of policy ids and
+  -- randomly shuffle them into a MultiAsset
+
+  if numPolicyId > numAssetName -- Since we don't want to generate policies with empty assets
```
lehins: Here is a simpler strategy that you can use:

Generate a list with `Int`s of length `numPolicyId` with positive numbers that sum up to `numAssetName`. Then you can use later to generate the nested structure. eg.
```
numPloicyId = 2
numAssetName = 5
```
Then one of lists that could be generated is:
```
lengths = [4, 1]
```

This would mean you can generate a multiasset with 2 policyIds where first one has 4 asset names and the second one has 1. This can be achived with two simple recursive functions.

Also. Do not use `resize` like this, this is not what it was designed for.

eras/mary/impl/testlib/Test/Cardano/Ledger/Mary/Arbitrary.hs
```diff
@@ -89,9 +98,88 @@ instance Crypto c => Arbitrary (PolicyID c) where
...
+    then genMultiAssetToFail
+    else do
+      ps <- resize numPolicyId $ vectorOf numPolicyId (arbitrary :: Gen (PolicyID c))
+      as <- resize numAssetName $ vectorOf numAssetName (arbitrary :: Gen (AssetName, Int))
+      let
+        initialTriples = zipWith (\p (a, v) -> (p, a, v)) ps as -- All policies should have at least one asset
```
lehins: `zipWith` will drop generated elements. In other words `length initialTriples == min (length ps) (length as)

So this whole thing is wrong

eras/mary/impl/src/Cardano/Ledger/Mary/Value.hs
lehins: No, because `representationSize` does what we don't want to do: nubs the asset names and counts their length

aniketd: Got it! I was half-thinking the same, but wanted to run this by you. :+1:

eras/mary/impl/testlib/Test/Cardano/Ledger/Mary/Arbitrary.hs
```diff
@@ -89,9 +98,88 @@ instance Crypto c => Arbitrary (PolicyID c) where
...
+    oneof
+      [ do
+          Positive numAsssetNames <- arbitrary
+          let minNumPolicyId = (65535 - 44 * numAsssetNames) `div` 28
+          numPolicyIds <- (minNumPolicyId +) . getPositive <$> arbitrary
+          pure (numPolicyIds, numAsssetNames)
+      , do
+          Positive numPolicyIds <- arbitrary
+          let minNumAssetNames = (65535 - 28 * numPolicyIds) `div` 44
+          numAssetNames <- (minNumAssetNames +) . getPositive <$> arbitrary
```
aniketd: Also add `numPolicyId` to this to ensure that the number is greater.

eras/mary/impl/testlib/Test/Cardano/Ledger/Mary/Arbitrary.hs
```diff
@@ -89,9 +98,88 @@ instance Crypto c => Arbitrary (PolicyID c) where
...
+
+-- For negative tests, we need a definite generator that causes overflow
+genMultiAssetToFail :: forall c. Crypto c => Gen (MultiAsset c)
+genMultiAssetToFail = do
+  (numPolicyId, numAssetName) <-
+    oneof
+      [ do
+          Positive numAsssetNames <- arbitrary
+          let minNumPolicyId = (65535 - 44 * numAsssetNames) `div` 28
+          numPolicyIds <- (minNumPolicyId +) . getPositive <$> arbitrary
```
aniketd: Add `max numAssetNames .` here to ensure number is always larger or equal.

eras/mary/impl/testlib/Test/Cardano/Ledger/Mary/Arbitrary.hs
```diff
@@ -89,9 +98,88 @@ instance Crypto c => Arbitrary (PolicyID c) where
...
+      remainingTriples <-
+        traverse
+          ( \(a, v) -> do
+              policy <- elements ps
+              pure (policy, a, v)
+          )
+          remainingAs
+      pure $
+        let final = multiAssetFromListBounded $ initialTriples <> remainingTriples
+         in traceShow ("FINAL" :: String, numPolicyId, numAssetName, representationSize $ flattenMultiAsset final) final -- TODO
```
lehins:
```suggestion
      let ma =
        let final = multiAssetFromListBounded $ initialTriples <> remainingTriples
         in traceShow ("FINAL" :: String, numPolicyId, numAssetName, representationSize $ flattenMultiAsset final) final -- TODO
      if length ma == numPolicyIds && sum (length <$> ma) == numAsssetNames then pure ma else genMultiAssetToFail
```

eras/mary/impl/test/Test/Cardano/Ledger/Mary/ValueSpec.hs
```diff
@@ -47,24 +56,47 @@ spec = do
...
       prop "Empty MaryValue fails for Conway" $
         forAll (genMaryValue (genEmptyMultiAsset @StandardCrypto)) $
           roundTripCborRangeFailureExpectation (natVersion @9) maxBound
+      it "Too many assets should fail" $
+        expectFailure $
+          property $
+            forAll
+              (genMaryValue (genMultiAssetToFail @StandardCrypto))
+              ( embedTripRangeFailureExpectation @(MaryValue StandardCrypto) @(MaryValue StandardCrypto)
+                  cborTrip
```
lehins:
```suggestion
              ( roundTripCborRangeFailureExpectation @(MaryValue StandardCrypto) @(MaryValue StandardCrypto)
```

eras/mary/impl/testlib/Test/Cardano/Ledger/Mary/Arbitrary.hs
aniketd: > Noting for posterity: This part was re-reviewed and it was decided that with minor adjustments, overall it works well.

lehins: Yes. It was my mistake. This whole thing is totally right :slightly_smiling_face:

eras/mary/impl/testlib/Test/Cardano/Ledger/Mary/Arbitrary.hs
aniketd: This has been updated! :raised_hands:

eras/mary/impl/test/Test/Cardano/Ledger/Mary/ValueSpec.hs
aniketd: Totally agree! :100: Thanks for explaining this to me :heart: I have updated the test.

eras/mary/impl/testlib/Test/Cardano/Ledger/Mary/Arbitrary.hs
aniketd: Thanks! This is now updated.

eras/mary/impl/testlib/Test/Cardano/Ledger/Mary/Arbitrary.hs
aniketd: > with positive numbers that sum up to numAssetName

I could not get to doing this today. However, I have updated the generator as per our previous discussion, and it results in sizes slightly above 65535 in most cases but also almost double that in some. Let me know what you think. I will be back to pick this up, if the current update does not   make the cut.

eras/mary/impl/src/Cardano/Ledger/Mary/Value.hs
```diff
@@ -561,8 +572,8 @@ to ::
 to (MaryValue ada (MultiAsset m))
   | Map.null m =
       CompactValueAdaOnly . CompactCoin <$> integerToWord64 ada
-to v = do
-  c <- integerToWord64 ada
+to v@(MaryValue _ (MultiAsset ma)) = do
+  c <- assert (44 * sum (length <$> ma) + 28 * length ma < 65535) (integerToWord64 ada)
```
lehins: Looks like we could use a helper function that can be used in all three places in this module where this condition is checked. Note that I changed `<` to `<=`. This actually matters.
```haskell
isMultiAssetNotTooBig (MultiAsset ma)) =
  44 * sum (length <$> ma) + 28 * length ma <= 65535
```

eras/mary/impl/testlib/Test/Cardano/Ledger/Mary/Arbitrary.hs
```diff
@@ -89,9 +102,90 @@ instance Crypto c => Arbitrary (PolicyID c) where
...
+      , do
+          numPolicyIds <- chooseInt (1, 2340) -- p > 2340 will produce negative `minNumAssetNames`
+          let minNumAssetNames = (65535 - 28 * numPolicyIds) `div` 44
+          numAssetNames <- (minNumAssetNames +) . getPositive <$> arbitrary
+          pure (min numPolicyIds numAssetNames, max numPolicyIds numAssetNames)
+      ]
+
+  -- Here we generate separately a list of asset names and a list of policy ids and
+  -- randomly shuffle them into a MultiAsset, ensuring that each policy has at least one asset.
+  ps <- vectorOf numP (arbitrary :: Gen (PolicyID c))
```
lehins: This will ensure there will be no duplicates generated. `Arbitrary` instance for `PolicyID` intentionally generates values that can clash
```suggestion
  ps <- vectorOf numP (PolicyID . ScriptHash <$> arbitrary)
```

eras/mary/impl/testlib/Test/Cardano/Ledger/Mary/Arbitrary.hs
```diff
@@ -89,9 +102,90 @@ instance Crypto c => Arbitrary (PolicyID c) where
...
+      ( \(a, v) -> do
+          policy <- elements ps -- For every remaining asset, randomly assign a policy
+          pure (policy, a, v)
+      )
+      remainingAs
+  let MultiAsset ma = multiAssetFromListBounded $ initialTriples <> remainingTriples
+  if length ma == numP && sum (length <$> ma) == numA
+    then -- Ensure that the large numbers aren't reduced due to duplicates
+    -- In practice, this occurs more often than anticipated, especially,
+    -- with duplicate policy ids.
```
lehins: This comment should no longer be valid with my other suggestion
```suggestion
    then
    -- Ensure that the large numbers aren't reduced due to duplicates.
    -- This impossible in practice, since PRNG produces uniform values
```

eras/mary/impl/testlib/Test/Cardano/Ledger/Mary/Arbitrary.hs
```diff
@@ -89,9 +102,90 @@ instance Crypto c => Arbitrary (PolicyID c) where
...
+          numAssetNames <- chooseInt (1, 1489) -- n > 1489 will produce negative `minNumPolicyIds`
+          let minNumPolicyIds = (65535 - 44 * numAssetNames) `div` 28
+          numPolicyIds <- (minNumPolicyIds +) . getPositive <$> arbitrary
+          pure (min numPolicyIds numAssetNames, max numPolicyIds numAssetNames)
+      , do
+          numPolicyIds <- chooseInt (1, 2340) -- p > 2340 will produce negative `minNumAssetNames`
+          let minNumAssetNames = (65535 - 28 * numPolicyIds) `div` 44
+          numAssetNames <- (minNumAssetNames +) . getPositive <$> arbitrary
+          pure (min numPolicyIds numAssetNames, max numPolicyIds numAssetNames)
+      ]
```
lehins: I think this is going to be the best strategy. I realized that there is no need to have two separate approaches for generating `numP` and `numA`, both of them rely on the same math, so one of them is redundant.
Also, I realized that when `minNumPolicyIds` is negative, it means that the number of asset names is big enough by itself to overflow the compaction, which means it sufficient to have just one policy that contains all of those asset names.
```suggestion
    do
          -- when numAssetNames > 1489 it is enough to have at least 1 PolicyId
          numAssetNames <- chooseInt (1, 1500)
          -- Make sure there is at least 1 PolicyId:
          let minNumPolicyIds = max 0 ((65535 - 44 * numAssetNames) `div` 28)
          numPolicyIds <- (minNumPolicyIds +) . getPositive <$> arbitrary
          -- Ensure we have at least as many asset names as there are PolicyIds
          pure (numPolicyIds, max numPolicyIds numAssetNames)
```

lehins (changes requested):
Couple of minor changes are still needed, but other than that this looks awesome!

eras/mary/impl/testlib/Test/Cardano/Ledger/Mary/Arbitrary.hs
aniketd: Got it! :raised_hands:

eras/mary/impl/src/Cardano/Ledger/Mary/Value.hs
```diff
@@ -676,6 +681,18 @@ to v = do
...
+-- any offset within a MultiAsset compact representation is likely to overflow Word16.
+--
+-- compact form inequality:
+--   8n (Word64) + 2n (Word16) + 2n (Word16) + 28p (policy ids) + sum of lengths of unique asset names <= 65535
+-- maximum size for the asset name is 32 bytes, so:
+-- 8n + 2n + 2n + 28p + 32n <= 65535
+-- where: n = total number of assets, p = number of unique policy ids
+isMultiAssetSmallEnough :: MultiAsset c -> Bool
+isMultiAssetSmallEnough (MultiAsset ma) =
+  44 * sum (length <$> ma) + 28 * length ma <= 65535
```
lehins: Slight performance improvement:
```suggestion
  44 * getSum (foldMap' (Sum . length) ma) + 28 * length ma <= 65535
```

eras/mary/impl/testlib/Test/Cardano/Ledger/Mary/Arbitrary.hs
```diff
@@ -89,9 +104,89 @@ instance Crypto c => Arbitrary (PolicyID c) where
...
+    pure (numPolicyIds, max numPolicyIds numAssetNames)
+
+  -- Here we generate separately a list of asset names and a list of policy ids and
+  -- randomly shuffle them into a MultiAsset, ensuring that each policy has at least one asset.
+  ps <- vectorOf numP genPolicyIDToFail
+  as <-
+    vectorOf numA $
+      (,)
+        <$> genAssetNameToFail
+        <*> bool (getNonZero @Int <$> arbitrary) (getPositive @Int <$> arbitrary) isForMaryValue
```
lehins: I usually suggest avoiding `bool`, because it is rarely batter than `if then else` and only confuses people. Even after many years of using Haskell every time with `bool` I have to go and lookup the order of arguments in order to convince my self it is correct.
```suggestion
        <*> if isForMaryValue then getNonZero @Int <$> arbitrary else getPositive @Int <$> arbitrary
```

lehins (approved):
Two final comments. This is awesome stuff! Thank you Aniket!

eras/mary/impl/src/Cardano/Ledger/Mary/Value.hs
aniketd: Awesome! Thanks! :raised_hands:

eras/mary/impl/testlib/Test/Cardano/Ledger/Mary/Arbitrary.hs
aniketd: Got it! :+1:

PR #3349, opened 2023-03-23
Shrinking with constraints

UlfNorell:
# Description

Add shrinking capability to the "generators subject to constraints".

# Checklist

- [ ] ~~Any changes are noted in the `CHANGELOG.md` for affected package~~
- [x] Self-reviewed the diff

### DIFF
--- a/libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/Ast.hs
+++ b/libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/Ast.hs
@@ -21,7 +21,7 @@ import qualified Data.Universe as Univ (Any (..))
 import Data.Word (Word64)
 import Lens.Micro
 import Test.Cardano.Ledger.Constrained.Classes (Adds (..), Count (..), Sizeable (..), Sums (..))
-import Test.Cardano.Ledger.Constrained.Env (Access (..), AnyF (..), Env, Field (..), Name (..), V (..), findVar, storeVar)
+import Test.Cardano.Ledger.Constrained.Env (Access (..), AnyF (..), Env (..), Field (..), Name (..), Payload (..), V (..), findVar, storeVar)
 import Test.Cardano.Ledger.Constrained.Monad (Typed (..), failT)
 import Test.Cardano.Ledger.Constrained.Size (OrdCond (..), Size (..), runOrdCond, runSize)
 import Test.Cardano.Ledger.Constrained.TypeRep (Rep (..), synopsis, testEql, (:~:) (Refl))
@@ -236,11 +236,19 @@ varsOfSum ans (Project _ x) = varsOfTerm ans x
 -- =====================================================
 -- Subtitution of (V era t) inside of (Spec era t)
 
-substToEnv :: [SubItem era] -> Env era -> Typed (Env era)
+type Subst era = [SubItem era]
+
+substToEnv :: Subst era -> Env era -> Typed (Env era)
 substToEnv [] ans = pure ans
 substToEnv ((SubItem v (Lit _ t)) : more) ans =
   substToEnv more (storeVar v t ans)
 substToEnv ((SubItem _ e) : _) _ = failT ["Not Literal expr in substToEnv: " ++ show e]
+
+envToSubst :: Env era -> Subst era
+envToSubst (Env env) =
+  [ SubItem (V x rep acc) (Lit rep t)
+  | (x, Payload rep t acc) <- Map.toList env
+  ]
 
 data SubItem era where
   SubItem :: V era t -> Term era t -> SubItem era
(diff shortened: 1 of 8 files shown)


libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/Ast.hs
```diff
@@ -236,12 +236,20 @@ varsOfSum ans (Project _ x) = varsOfTerm ans x
 -- =====================================================
 -- Subtitution of (V era t) inside of (Spec era t)
 
-substToEnv :: [SubItem era] -> Env era -> Typed (Env era)
+type Subst era = [SubItem era]
```
TimSheard, line 239: This is definitely a good idea. Thanks

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/Monad.hs
```diff
@@ -21,6 +21,9 @@ module Test.Cardano.Ledger.Constrained.Monad (
 newtype Typed x = Typed {runTyped :: Either [String] x}
   deriving (Functor, Applicative, Monad)
 
+instance MonadFail Typed where
+  fail err = failT [err]
+
```
TimSheard, line 26: Another good idea.

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/Shrink.hs
```diff
@@ -0,0 +1,66 @@
...
+--    * pick a variable
+--    * shrink its value to something that still satisfies the defining constraints for the variable
+--    * adjust the values for variables depending on the picked variables to fix constraint
+--      violations resulting from the changed value
+shrinkEnv :: Era era => DependGraph era -> Env era -> [Env era]
+shrinkEnv (DependGraph vs) env =
+  [ env'
+  | (before, (x, cs), after) <- splits vs
+  , env' <- shrinkOneVar env (map fst before) x cs after
+  ]
```
TimSheard: Say some thing about how the ordering allows one to know which variables depend upon the variable one picked.

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/Shrink.hs
```diff
@@ -0,0 +1,66 @@
...
+
+splits :: [a] -> [([a], a, [a])]
+splits [] = []
+splits (x : xs) = ([], x, xs) : [(x : ys, y, zs) | (ys, y, zs) <- splits xs]
```
TimSheard: A comment about how this is to be used. I assume it is applied to the dependency graph?

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/TypeRep.hs
```diff
@@ -451,6 +453,50 @@ genpup (PPUPStateR (Alonzo _)) = arbitrary
...
+shrinkRep RationalR t = shrink t
+shrinkRep Word64R t = shrink t
+shrinkRep IntR t = shrink t
+shrinkRep NaturalR t = shrink t
+shrinkRep FloatR t = shrink t
+shrinkRep TxInR t = shrink t
+shrinkRep (ValueR _) _ = []
+shrinkRep (TxOutR _) _ = []
+shrinkRep (UTxOR _) _ = []
+shrinkRep (PParamsR _) _ = []
```
TimSheard: Why are somethings using (shrink t) and others ([]).  How do you know? experience? or something else?

TimSheard (approved):
An excellent addition. Thanks for doing this.

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/TypeRep.hs
UlfNorell: We'd like them all to be `shrink t`. It's `[]` for the ones that don't have an `Arbitrary` instance. I'll add a todo.

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/Shrink.hs
UlfNorell: It's only used in the function above. I made it local.

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/Shrink.hs
UlfNorell: 👍 Fixed

TimSheard:
This PR has been incorporated into https://github.com/input-output-hk/cardano-ledger/pull/3374

TimSheard:
This is incorporated into https://github.com/input-output-hk/cardano-ledger/pull/3374

PR #3350, opened 2023-03-23
Fix some issues pointed out in the Babbage audit

WhatisRT:
In particular, I've removed the rounding from checking the balancing of the collateral since if n is a natural number and x is a rational (or real) number, n >= x is equivalent to n >= ceil x. Without the rounding it's way more obvious that you can just multiply the equation by 100 without changing the semantics.


JaredCorduan (approved):
looks great!

JaredCorduan:
@WhatisRT I need to also add something to the shelley spec. So after I do that, we can release another round of specs.

PR #3360, opened 2023-03-27
Translate bad `Ptr`s to valid `Ptr`s during translation to Conway.

lehins:
Bad pointers and ugly addresses are not allowed in transactions starting with Babbage era. However, Ptr's that did land on chain in prior eras are still present in the LedgerState. Thanks to deprecation of `Ptr`s we can simplify the Ptr decoding logic, but only after the bad ones have been translated.

# Description

to the issue.

# Checklist

- [x] Self-reviewed the diff


JaredCorduan (approved):
thanks for fixing this and cleaning it up too!

Issue #3337, opened 2023-03-16
Improve CanStartFromGenesis as a concept


lehins:
* Change `AdditionalGenesisConfig` to `CumulativeTranslationContext`
* Make it possible to extract `TranslationContext` from `CumulativeTranslationContext`
* Decouple `ConwayGenesis` from `CanStartFromGenesis`

JaredCorduan:
I believe this is related: #1633

PR #3361, opened 2023-03-28
Drop alonzo genesis from conway genesis

lehins:
# Description

This is the first step to taking care of #3337 

`AlonzoGenesis` has no business being in `ConwayGenesis`, so it is being removed. This breaks `CanStartFromGenesis`, but that will be fixed in a subsequent PR

# Checklist

- [x] Self-reviewed the diff


JaredCorduan (approved):
awesome!

PR #3363, opened 2023-03-29
Move tpraos utils

lehins:
# Description

Recent PR #3333 changed semantics of Arbitrary instances, which ended up breaking a test in consensus repo. In order to fix it we need to expose previous generator functionality. This PR takes care of that and some more:

* Improve `cardano-protocol-tpraos:testlib`
  * Create `VRFKeyPair` and `KESKeyPair`. And start using them everywhere instead of tuples
  * Move `FakeVRF` from `cardano-ledger-shelley-test` to `VRF.Fake` module.
  * Add `mkBHeader`
  * Move from `cardano-ledegr-shelley-test`: `mkOCert`, `mkBHBody`, `mkBlock`
  * Move `AllIssuerKeys` from `cardano-ledger-shelley-test`. Rename its fields:
    *  `cold` - > `aikCold`
    *  `hot` - > `aikHot`
    *  `vrf` - > `aikVrf`
    *  `hk` - > `aikColdKeyHash`
  * Bring back `genBlock`
  * Move `genCoherentBlock` from `cardano-ledegr-shelley-test` and change it to accept `AllIssuerKeys` as an argument.
* Move `bBodySize` into `Cardano.Ledger.Core`
* Remove `Test.Cardano.Ledger.Generic.Consensus` and `Test.Cardano.Ledger.Examples.Consensus` because:
  * These two modules duplicated functionality from the individual era test suites
  * Consensus examples are only used for golden tests in consensus, which means the above two modules where unused
  * `cardano-ledger-test` is not released to CHaPs, so it is not an appropriate place for such definitions.

# Checklist

- [ ] ~New tests are added if needed and existing tests are updated~
- [x] Self-reviewed the diff


libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/Consensus.hs
```diff
@@ -1,1054 +0,0 @@
...
-{-# LANGUAGE ScopedTypeVariables #-}
-{-# LANGUAGE TypeApplications #-}
-{-# LANGUAGE TypeFamilies #-}
-
--- | Generic Consensus examples
---   This module is meant to replace a bunch of Consensus examples
---   distributed across many Era. By defining the Consensus examples
---   for every era once, by writing code that is parameterized by a (Proof era),
---   it should be easier to maintain. It is meant to be byte for byte exact.
-module Test.Cardano.Ledger.Generic.Consensus where
```
JaredCorduan, line 16: why did you remove these? I thought the plan was to switch the consensus example to use these.

lehins: > I thought the plan was to switch the consensus example to use these.

Why would you wanna do that?

libs/cardano-protocol-tpraos/testlib/Test/Cardano/Protocol/TPraos/Create.hs
```diff
@@ -0,0 +1,294 @@
...
+  ) =>
+  -- | Hash of previous block
+  HashHeader (EraCrypto era) ->
+  -- | All keys in the stake pool
+  AllIssuerKeys (EraCrypto era) r ->
+  -- | Transactions to record
+  [Tx era] ->
+  -- | Current slot
+  SlotNo ->
+  -- | Block number\/chain length\/chain "difficulty"
```
JaredCorduan, line 271: difficulty?

JaredCorduan (approved):
This is some nice cleanup!

libs/cardano-protocol-tpraos/testlib/Test/Cardano/Protocol/TPraos/Create.hs
lehins: I have no idea what it means. I didn't write it :smile:

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/Consensus.hs
JaredCorduan: the generic ones can be reused in future eras.

libs/cardano-protocol-tpraos/testlib/Test/Cardano/Protocol/TPraos/Create.hs
JaredCorduan: :laughing: , gotcha. I think "Block number" is i pretty good explanation myself :)

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/Consensus.hs
lehins: If people stop coding with a copy+paste strategy then current consesus examples can also be reused in future eras. Otherwise there is nothing wrong with them. We have lenses for all our types which allow for much better and safer reusability of code, instead of the generic approach using `Proof`.

Another down side of using generic approach for consensus examples is we would have to release cardano-ledger-test to CHaPs and make this whole Proof thing into a public API. We don't want to do that.

PR #3364, opened 2023-03-30
WIP  Change STS from LEDGER down to use ILC Diff

TimSheard:
Change the STS instanced from LEDGER down to use ILC Diffs for the STS State type family.
So far we have added 
  1) The ILC class
   2) ILC instances for Coin, UMap, DPState, DState, PState, InstantaneousRewards
   3) Added a bunch of property tests

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/libs/cardano-data/src/Data/Incremental.hs
+++ b/libs/cardano-data/src/Data/Incremental.hs
@@ -1,0 +1,251 @@
+{-# LANGUAGE BangPatterns #-}
+{-# LANGUAGE ConstraintKinds #-}
+{-# LANGUAGE DeriveAnyClass #-}
+{-# LANGUAGE DeriveGeneric #-}
+{-# LANGUAGE DerivingStrategies #-}
+{-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE FlexibleInstances #-}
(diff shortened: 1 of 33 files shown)


TimSheard:
Here is a picture of what we are trying to do
Untitled presentation

libs/cardano-data/src/Data/Incremental.hs
```diff
@@ -0,0 +1,203 @@
...
+-- 2) an entry is changed or created, so there is a new range value
+--
+-- To do this we introduce two datatypes MonoidRngD and BinaryRngD. They
+-- will become part of the definition for the Diff(Map dom rng). It also
+-- turns out thet Both of them are Semigroups (but not Monoids as neither
+-- has a notion of No-Change. This is deliberate, but might be reconsidered
+-- at some point)
+
+-- | The range is deleted, overwritten, or combined using a Monoid
+data MonoidRngD v = Del | Write !v | Comb !v
```
Soupstraw: Could we just use a function `Maybe v -> Maybe v` instead?

Soupstraw: Never mind, that would make it hard to show the diffs..

PR #3365, opened 2023-03-30
Plutus V3

JaredCorduan:
# Plutus V3

This PR provides support for Plutus `V3`. Most of the types and functions that use `Language` are from `cardano-ledger-alonzo`, but note that both the Alonzo and the Babbage ledger rules forbid `V3`.

Note that we still anticipate changes to `V3` in Plutus, such as a new type for the script context (as given by CIP-85, "sums of products"), a new script purpose (for DReps), and new fields in the script context (supporting CIP-1694, so things like votes).

## Tests

The Babbage TxInfo tests were moved from `test-suite/test` to `test-suite/src` and made polymorphic using the supporting lenses. Conway TxInfo tests were added so that: 1) `V1` and `V2` use the same tests as Babbage, 2) the semantics of `V3` in conway are identical (for now!) to `V2`.

I did not add tests for executing any `V3` scripts. Once we have more of the testing infrastructure for conway in place we need to add some.

# Checklist

- [x] Self-reviewed the diff


eras/alonzo/impl/src/Cardano/Ledger/Alonzo/PParams.hs
```diff
@@ -687,6 +688,10 @@ getLanguageView pp lang =
       LangDepView
         (serialize' version lang)
         costModelEncoding
+    PlutusV3 ->
+      LangDepView
+        (serialize' version lang)
+        costModelEncoding
   where
```
lehins: Slight reduction of duplication:
```suggestion
      latestLangDepView
    PlutusV3 -> latestLangDepView
  where
    -- LangDepView for PlutusV1 differs from the rest
    latestLangDepView = LangDepView (serialize' version lang) costModelEncoding
```

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxInfo.hs
```diff
@@ -795,6 +817,18 @@ explainPlutusFailure _proxy pv lang scriptbytestring e ds cm eu =
...
+                  Nothing ->
+                    [ "The data is: " ++ show dat
+                    , "The redeemer is: " ++ show redeemer
+                    , "The third data argument, does not translate to a V2 script context\n" ++ show info
+                    ]
+                  Just info2 ->
+                    [ "The data is: " ++ show dat
+                    , "The redeemer is: " ++ show redeemer
+                    , "The script context is:\n" ++ show (pretty (info2 :: PV3.ScriptContext))
+                    ]
```
lehins: It would be nice to create a helper function that does this  and reuse for all plutus versions.

Copy/paste programming is evil!

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxInfo.hs
```diff
@@ -817,6 +851,16 @@ explainPlutusFailure _proxy pv lang scriptbytestring e ds cm eu =
                     [ "The redeemer is: " ++ show redeemer
                     , "The script context is:\n" ++ show (pretty (info2 :: PV2.ScriptContext))
                     ]
+              PlutusV3 ->
+                case PV3.fromData info of
+                  Nothing ->
+                    [ "The redeemer is: " ++ show redeemer
+                    , "The second data argument, does not translate to a V2 script context\n" ++ show info
```
lehins: Same here. Copy paste programming is the worst of evil. 
Here is proof of it:  Should be `V3` instead of `V2` 

It is best to abstract this away with a helper function

eras/alonzo/test-suite/test/Test/Cardano/Ledger/Alonzo/TxInfo.hs
```diff
@@ -134,12 +141,9 @@ tests =
...
     , testGroup
-        "Plutus V2"
-        [ testCase "translation error for V2 in Alonzo" $
-            expectTranslationError
-              PlutusV2
-              (txEx shelleyInput shelleyOutput)
-              (LanguageNotSupported PlutusV2)
+        "Plutus VX, X > 1"
+        [ testCase "translation error for V2 in Alonzo" (expectLanguageNotSupported PlutusV2)
+        , testCase "translation error for V3 in Alonzo" (expectLanguageNotSupported PlutusV3)
```
lehins, line 146: `expectLanguageNotSupported` 
Nice! That's what I am talking about. :rocket: 
Avoiding duplication is a necessity for coherent and maintainable code

eras/babbage/impl/src/Cardano/Ledger/Babbage/TxInfo.hs
```diff
@@ -176,61 +176,95 @@ babbageTxInfo ::
...
+      , PV1.txInfoId = PV1.TxId (Alonzo.transSafeHash (hashAnnotated txBody))
+      }
   where
     txBody = tx ^. bodyTxL
     witnesses = tx ^. witsTxL
     outs = txBody ^. outputsTxBodyL
     fee = txBody ^. feeTxBodyL
     multiAsset = txBody ^. mintTxBodyL
-    interval = txBody ^. vldtTxBodyL
+    datpairs = Map.toList (unTxDats $ txdats' witnesses)
```
lehins:
```suggestion
    datpairs = Map.toList (unTxDats $ witnesses ^. datsTxWitsL)
```

eras/babbage/impl/src/Cardano/Ledger/Babbage/TxInfo.hs
```diff
@@ -176,61 +176,95 @@ babbageTxInfo ::
...
+      , PV2.txInfoId = PV2.TxId (Alonzo.transSafeHash (hashAnnotated txBody))
+      }
+  where
+    txBody = tx ^. bodyTxL
+    witnesses = tx ^. witsTxL
+    outs = txBody ^. outputsTxBodyL
+    fee = txBody ^. feeTxBodyL
+    multiAsset = txBody ^. mintTxBodyL
     datpairs = Map.toList (unTxDats $ txdats' witnesses)
     rdmrs = Map.toList (unRedeemers $ txrdmrs' witnesses)
```
lehins: Might as well switch these to lenses as well
```suggestion
    datpairs = Map.toList (unTxDats $ witnesses ^. datsTxWitsL)
    rdmrs = Map.toList (unRedeemers $ witnesses ^. rdmrsTxWitsL)
```

eras/conway/impl/src/Cardano/Ledger/Conway/TxInfo.hs
```diff
@@ -0,0 +1,107 @@
...
+      [minBound ..]
+      (foldr (:) [] outs)
+  rdmrs' <- mapM (B.transRedeemerPtr txBody) rdmrs
+  pure . TxInfoPV3 $
+    PV3.TxInfo -- TODO Add relevant CIP-1694 data to PV3.TxInfo
+      { PV3.txInfoInputs = inputs
+      , PV3.txInfoOutputs = outputs
+      , PV3.txInfoReferenceInputs = refInputs
+      , PV3.txInfoFee = Alonzo.transValue (inject @(MaryValue (EraCrypto era)) fee)
+      , PV3.txInfoMint = Alonzo.transMintValue multiAsset
```
lehins: I think it might be a good idea to fix this ugliness with PlutusV3 where zero ada is passed as one of the minted values
```suggestion
      , PV3.txInfoMint = Alonzo.transMultiAsset multiAsset
```

eras/conway/impl/src/Cardano/Ledger/Conway/TxInfo.hs
```diff
@@ -0,0 +1,107 @@
...
+      , PV3.txInfoId = PV3.TxId (Alonzo.transSafeHash (hashAnnotated txBody))
+      }
+  where
+    txBody = tx ^. bodyTxL
+    witnesses = tx ^. witsTxL
+    outs = txBody ^. outputsTxBodyL
+    fee = txBody ^. feeTxBodyL
+    multiAsset = txBody ^. mintTxBodyL
+    datpairs = Map.toList (unTxDats $ txdats' witnesses)
+    rdmrs = Map.toList (unRedeemers $ txrdmrs' witnesses)
```
lehins:
```suggestion
    datpairs = Map.toList (unTxDats $ witnesses ^. datsTxWitsL)
    rdmrs = Map.toList (unRedeemers $ witnesses ^. rdmrsTxWitsL)
```

eras/conway/impl/src/Cardano/Ledger/Conway/TxInfo.hs
```diff
@@ -0,0 +1,107 @@
...
+    PlutusV3 -> conwayTxInfoV3 timeRange tx utxo
+  where
+    interval = tx ^. bodyTxL ^. vldtTxBodyL
+
+conwayTxInfoV3 ::
+  forall era.
+  ( EraTx era
+  , BabbageEraTxBody era
+  , Value era ~ MaryValue (EraCrypto era)
+  , TxWits era ~ AlonzoTxWits era
```
lehins: Suggestion to use lenses should allow you to remove this constraint
```suggestion
```

eras/conway/impl/src/Cardano/Ledger/Conway/TxInfo.hs
```diff
@@ -0,0 +1,107 @@
...
+import Data.Text (Text)
+import Lens.Micro
+import qualified PlutusLedgerApi.V3 as PV3
+
+conwayTxInfo ::
+  forall era.
+  ( EraTx era
+  , BabbageEraTxBody era
+  , Value era ~ MaryValue (EraCrypto era)
+  , TxWits era ~ AlonzoTxWits era
```
lehins: And here
```suggestion
```

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxAuxData.hs
```diff
@@ -210,6 +220,7 @@ instance Era era => DecCBOR (Annotator (AlonzoTxAuxDataRaw era)) where
           (D (sequence <$> decodeStrictSeq decCBOR))
       auxDataField 2 = fieldA (addPlutusScripts PlutusV1) From
       auxDataField 3 = fieldA (addPlutusScripts PlutusV2) From
+      auxDataField 4 = fieldA (addPlutusScripts PlutusV3) From
```
lehins: We need to fail here for eras before Conway. One way of do this that I can think of is something like this:
```suggestion
      auxDataField 4 = 
        fieldA (addPlutusScripts PlutusV3) $ 
        D (unlessDecoderVersionAtLeast (natVersion @9) (fail "PlutusV3 is not supported yet") >> decCBOR)
```

Creating a specialized reusable decoder like this would be a good investment, since we'll have to do it for every new plutus version and we could use it in the actual Script decoder as well

This will need a new reusable function in `Cardano.Ledger.Binary.Decoding.Decoder` next to `whenDecoderVersionAtLeast`:
```haskell
-- | Optionally run a decoder depending on the current version and the supplied one.
unlessDecoderVersionAtLeast ::
  Version ->
  -- | Run this decoder whenever current decoder version is smaller to the supplied
  -- `Version`
  Decoder s a ->
  Decoder s ()
unlessDecoderVersionAtLeast atLeast decoder = do
  cur <- getDecoderVersion
  unless (cur >= atLeast) (void decoder)
```

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxWits.hs
```diff
@@ -561,6 +570,10 @@ instance
         fieldA
           addScripts
           (fmap (PlutusScript PlutusV2) <$> From)
+      txWitnessField 7 =
+        fieldA
+          addScripts
+          (fmap (PlutusScript PlutusV3) <$> From)
```
lehins: If I remember correctly witnesses with V3 in Babbage will not go through due to lack of CostModel, right?

Maybe as an extra precausion we should disable it at serialization level prior to Conway as well?

eras/babbage/impl/src/Cardano/Ledger/Babbage/TxInfo.hs
```diff
@@ -176,61 +176,95 @@ babbageTxInfo ::
...
     multiAsset = txBody ^. mintTxBodyL
-    interval = txBody ^. vldtTxBodyL
+    datpairs = Map.toList (unTxDats $ txdats' witnesses)
 
+babbageTxInfoV2 ::
+  forall era.
+  ( EraTx era
+  , BabbageEraTxBody era
+  , Value era ~ MaryValue (EraCrypto era)
+  , TxWits era ~ AlonzoTxWits era
```
lehins: With lenses this is gone
```suggestion
```

eras/babbage/impl/src/Cardano/Ledger/Babbage/TxInfo.hs
```diff
@@ -176,61 +176,95 @@ babbageTxInfo ::
...
+    _ -> Left $ LanguageNotSupported lang
+  where
+    interval = tx ^. bodyTxL ^. vldtTxBodyL
+
+babbageTxInfoV1 ::
+  forall era.
+  ( EraTx era
+  , BabbageEraTxBody era
+  , Value era ~ MaryValue (EraCrypto era)
+  , TxWits era ~ AlonzoTxWits era
```
lehins: Here too
```suggestion
```

eras/babbage/test-suite/src/Test/Cardano/Ledger/Babbage/TxInfo.hs
```diff
@@ -0,0 +1,350 @@
...
+  TxIn (EraCrypto era) ->
+  Maybe (TxIn (EraCrypto era)) ->
+  TxOut era ->
+  TxBody era
+txb i mRefInp o =
+  mkBasicTxBody
+    & inputsTxBodyL .~ (Set.singleton i)
+    & B.referenceInputsTxBodyL .~ (maybe mempty Set.singleton mRefInp)
+    & outputsTxBodyL .~ (StrictSeq.singleton o)
+    & feeTxBodyL .~ (Coin 2)
```
lehins: Redundant parnethesis
```suggestion
    & inputsTxBodyL .~ Set.singleton i
    & B.referenceInputsTxBodyL .~ maybe mempty Set.singleton mRefInp
    & outputsTxBodyL .~ StrictSeq.singleton o
    & feeTxBodyL .~ Coin 2
```

eras/babbage/test-suite/src/Test/Cardano/Ledger/Babbage/TxInfo.hs
lehins, line 1: It is better to move files with `git mv`, this way git history travels with the file. Let me know if you need help fixing it?

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Scripts.hs
```diff
@@ -835,6 +847,7 @@ instance Era era => DecCBOR (Annotator (AlonzoScript era)) where
       decodeScript 0 = Ann (SumD TimelockScript) <*! From
       decodeScript 1 = Ann (SumD $ PlutusScript PlutusV1) <*! Ann From
       decodeScript 2 = Ann (SumD $ PlutusScript PlutusV2) <*! Ann From
+      decodeScript 3 = Ann (SumD $ PlutusScript PlutusV3) <*! Ann From
```
lehins: We need to forbid deserialization of PlutusV3 before `natVersion @9`  here too. This would take care of the inline script in the TxOut too.

lehins (changes requested):
This looks great. We should probably guard against PlutusV3 being deserialized in Babbage too, by fixing it in this instance: `DecCBOR Language`

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxWits.hs
JaredCorduan: correct. I think it is fine to serialize them, but I agree that we should block deserialization of V3.

eras/babbage/test-suite/src/Test/Cardano/Ledger/Babbage/TxInfo.hs
JaredCorduan: absolutely,  thanks for catching that. I should be fine.

eras/conway/impl/src/Cardano/Ledger/Conway/TxInfo.hs
JaredCorduan: We do need to be extremely careful with this, and make sure that we have good way to communicate this to script authors who may have grown used to this assumption...

JaredCorduan: there is an issues for this exact topic in the Plutus repo: https://github.com/input-output-hk/plutus/issues/5039

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxWits.hs
JaredCorduan: sorry, just realized that this _is_ deserialization. so yea, let's block it here too just for good measure (even though yes, no cost model also prevents this situation).

eras/babbage/test-suite/src/Test/Cardano/Ledger/Babbage/TxInfo.hs
JaredCorduan: done

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxInfo.hs
JaredCorduan: indeed that was a great suggestion, thanks!

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Scripts.hs
```diff
@@ -348,16 +354,20 @@ legacyParseCostModels o = do
...
-  pure $ maybeToList maybeCostModels
+parseCostModels :: Key -> Object -> Parser [(Language, CostModel)]
+parseCostModels key o = do
+  plutusV2 <- o .:? key
+  let valueDescription = show key <> " values"
+  maybeCostModelsV2 <-
+    traverse
+      (withArray valueDescription (parseCostModelsValues PlutusV2))
+      plutusV2
+  pure $ maybeToList maybeCostModelsV2
```
lehins: You almost got it :smiley: 
```suggestion
parseCostModels :: Language -> Object -> Parser [(Language, CostModel)]
parseCostModels lang o = do
  costModelsArray <- o .:? GHC.Exts.fromString (show lang)
  let valueDescription = show lang <> " values"
  maybeCostModels <-
    traverse
      (withArray valueDescription (parseCostModelsValues lang))
      costModelsArray
  pure $ maybeToList maybeCostModels
```

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxInfo.hs
```diff
@@ -767,65 +789,44 @@ explainPlutusFailure _proxy pv lang scriptbytestring e ds cm eu =
       firstLine = "\nThe " ++ show lang ++ " script (" ++ name ++ ") fails."
       pvLine = "The protocol version is: " ++ show pv
       plutusError = "The plutus error is: " ++ show e
+
+      getCtxAsString :: Language -> PV1.Data -> Maybe String
+      getCtxAsString PlutusV1 d = (show . pretty) <$> (PV1.fromData d :: Maybe PV1.ScriptContext)
+      getCtxAsString PlutusV2 d = (show . pretty) <$> (PV2.fromData d :: Maybe PV2.ScriptContext)
+      getCtxAsString PlutusV3 d = (show . pretty) <$> (PV3.fromData d :: Maybe PV3.ScriptContext)
```
lehins: Redundant
```suggestion
      getCtxAsString PlutusV1 d = show . pretty <$> (PV1.fromData d :: Maybe PV1.ScriptContext)
      getCtxAsString PlutusV2 d = show . pretty <$> (PV2.fromData d :: Maybe PV2.ScriptContext)
      getCtxAsString PlutusV3 d = show . pretty <$> (PV3.fromData d :: Maybe PV3.ScriptContext)
```

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Scripts.hs
```diff
@@ -348,16 +354,20 @@ legacyParseCostModels o = do
...
   where
-    parseCostModelsV2 :: Array -> Parser ((Language, CostModel))
-    parseCostModelsV2 array = do
+    parseCostModelsValues :: Language -> Array -> Parser ((Language, CostModel))
```
```suggestion
    parseCostModelsValues :: Language -> Array -> Parser (Language, CostModel)
```

eras/conway/impl/CHANGELOG.md
```diff
@@ -37,6 +37,9 @@
 * Add `EncCBOR`/`DecCBOR` and `ToCBOR`/`FromCBOR` for `ConwayTallyPredFailure`
 * Add `ToCBOR`/`FromCBOR` for `ConwayGovernance`
 * Remove `cgAlonzoGenesis` from `ConwayGenesis`.
+* Change the Conway txInfo to allow Plutus V3
+  NOTE - unlike V1 and V2, the the ledger will no longer place the "zero ada" value
```
```suggestion
  NOTE - unlike V1 and V2, the ledger will no longer place the "zero ada" value
```

libs/cardano-ledger-core/CHANGELOG.md
```diff
@@ -26,6 +26,7 @@
 * Remove custom `Fail` type, in favor of
   [`FailT`](https://hackage.haskell.org/package/FailT) package
 * Move `bBodySize` into `Cardano.Ledger.Core`
+* Add support for Plutus V3
```
```suggestion
* Add support for `PlutusV3`
```

lehins (approved):
Found one more bug. Despite that it looks ready to go.

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Scripts.hs
JaredCorduan: :facepalm: thanks!

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxInfo.hs
```diff
@@ -767,65 +789,44 @@ explainPlutusFailure _proxy pv lang scriptbytestring e ds cm eu =
...
+      getCtxAsString PlutusV2 d = show . pretty <$> (PV2.fromData d :: Maybe PV2.ScriptContext)
+      getCtxAsString PlutusV3 d = show . pretty <$> (PV3.fromData d :: Maybe PV3.ScriptContext)
+
+      ctxMessage info =
+        case getCtxAsString lang info of
+          Nothing ->
+            concat
+              [ "The third data argument does not translate to a "
+              , show lang
+              , "script context\n"
```
Soupstraw:
```suggestion
              , " script context\n"
```

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxWits.hs
```diff
@@ -451,7 +452,7 @@ instance (Era era, Script era ~ AlonzoScript era) => EncCBOR (AlonzoTxWitsRaw er
   encCBOR (AlonzoTxWitsRaw vkeys boots scripts dats rdmrs) =
     encode $
       Keyed
-        (\a b c d e f g -> AlonzoTxWitsRaw a b (c <> d <> e) f g)
+        (\a b c d e f g h -> AlonzoTxWitsRaw a b (c <> d <> e <> f) g h)
```
Soupstraw, line 454: I have no idea what the variables here mean

Soupstraw (approved):

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxInfo.hs
JaredCorduan: great catch, thank you!

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxWits.hs
JaredCorduan: When I myself use single letter variable names, my goal is to communicate "you do not need to understand what these things _are_, only that we are doing some kind of very local reshuffling". And by having a very clear ordering (in this case, the alphabet), I'm hoping to make the reshuffling visually easy to see. In other words, I was trying to communicate that "you do not need to know what the variables here mean".

Maybe no one else thinks this way, and full variable names would be less confusing to the majority of people.

The intended interpretation was supposed to be:

"smash the third, fourth, fifth, and sixth elements together, but otherwise keep the ordering the same".

eras/alonzo/impl/testlib/Test/Cardano/Ledger/Alonzo/Arbitrary.hs
```diff
@@ -205,14 +208,22 @@ instance
...
+genAlonzoScript langs = do
+  lang <- elements langs -- The language is not present in the Script serialization
+  frequency
+    [ (1, pure (alwaysSucceeds lang 1))
+    , (1, pure (alwaysFails lang 1))
+    , (10, TimelockScript <$> arbitrary)
+    ]
+
+instance forall c. Crypto c => Arbitrary (AlonzoScript (AlonzoEra c)) where
+  arbitrary = genAlonzoScript [PlutusV1]
```
JaredCorduan, line 226: heads up @lehins , I did something that I now regret. After I found a way to fix the serialization test, I used `--amend` and forced pushed, and now the relatively small changes I made are lost in the biff diff. apologies.

---

I started trying to use the `Twiddler`, but found that I was still going to have to use newtype wrappers and separate generators. So I tried this instead: have the `Arbitrary (AlonzoScript era)` instance specialized for each era, and passed the languages they support. ie, here you see `Arbitrary (AlonzoScript (AlonzoEra c))`, etc, for each era.

Also, clearly these are not great generators for serialization tests, as we suspected.

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/Same.hs
```diff
@@ -229,7 +229,7 @@ sameWithDependency (SomeM labx actx x1 x2 : more) =
     ansx -> extendLabel (labx ++ " ") ansx ++ sameWithDependency more
 
 instance
-  Reflect era =>
+  (Reflect era, Eq (TranslationContext era), Show (TranslationContext era)) =>
```
JaredCorduan: I do not understand why I now need to add these two new constraints. :(

JaredCorduan: ok, weird, CI is complaining that these are redundant. I'm not sure why when I build locally I get an an error about needing this constraint...

Soupstraw: I had a similar issue on my branch..

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxAuxData.hs
```diff
@@ -206,8 +217,11 @@ instance Era era => DecCBOR (Annotator (AlonzoTxAuxDataRaw era)) where
           (D (sequence <$> decodeStrictSeq decCBOR))
       auxDataField 2 = fieldA (addPlutusScripts PlutusV1) From
       auxDataField 3 = fieldA (addPlutusScripts PlutusV2) From
+      auxDataField 4 = fieldA (addPlutusScripts PlutusV3) decodeAtV9
       auxDataField n = field (\_ t -> t) (Invalid n)
 
+      decodeAtV9 = guardUntilAtLeast "PlutusV3 is not supported yet" (natVersion @9)
```
aniketd: May be `decodeAtV9` can be made a top-level function to reuse at the earlier site too? Just a suggestion, but there might be a compelling reason not to make it top-level if it isn't expected to be used more than in these two places, though.

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxInfo.hs
```diff
@@ -684,6 +689,13 @@ debugPlutus version db =
                       (getEvaluationContext costModel)
                       (transExUnits exUnits)
                       scripts
+                  evalV3 =
```
aniketd, line 694: Perhaps there is a missing addition of
```haskell
, plutusDebugLangDecoder (Proxy @'PlutusV3)
```
as part of the list argument to `F.asum` in `plutusDebugDecoder`, a few lines above this in the same function. I am not certain, but it looked like it. I may be wrong.

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxWits.hs
```diff
@@ -561,8 +571,14 @@ instance
         fieldA
           addScripts
           (fmap (PlutusScript PlutusV2) <$> From)
+      txWitnessField 7 =
+        fieldA
+          addScripts
+          (fmap (PlutusScript PlutusV3) <$> decodeAtV9)
       txWitnessField n = field (\_ t -> t) (Invalid n)
 
+      decodeAtV9 = guardUntilAtLeast "PlutusV3 is not supported yet" (natVersion @9)
```
aniketd: Another `decodeAtV9` :slightly_smiling_face:

eras/conway/test-suite/cddl-files/conway.cddl
```diff
@@ -219,6 +204,8 @@ script_data_hash = $hash32
 ;   - the language ID tag is encoded as expected.
 ;     Concretely, this means that the language version for V2 is encoded as
 ;     01 in hex.
+; For PlutusV3 (language id 2), the language view is the following:
+;   - the value of costmdls map at key 2 is encoded as an definite length list.
```
aniketd:
```suggestion
;   - the value of costmdls map at key 2 is encoded as a definite length list.
```

aniketd (approved):
LGTM! :raised_hands:

eras/alonzo/impl/CHANGELOG.md
```diff
@@ -14,6 +14,8 @@
 * Made it possible for `FromJSON` to decode `CostModels` both as the new approach:
   1. as a list of cost models values,
   2. and the old approach of mapping from the parameter name to the cost model value
+* Added support for Plutus V3 in the types and functions that use `Language`.
+  (Note that the Alonzo ledger era rules do not allow V3 scripts, however.).
```
lehins: This has to migrate to a new version, since `1.1` has already been released

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Arbitrary.hs
```diff
@@ -33,6 +38,9 @@ instance Crypto c => Arbitrary (ConwayDCert c) where
       , ConwayDCertConstitutional <$> arbitrary
       ]
 
+instance forall c. Crypto c => Arbitrary (AlonzoScript (ConwayEra c)) where
```
lehins: This `forall` is redundant
```suggestion
instance Crypto c => Arbitrary (AlonzoScript (ConwayEra c)) where
```

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Arbitrary.hs
```diff
@@ -1,23 +1,28 @@
 {-# LANGUAGE DataKinds #-}
 {-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE FlexibleInstances #-}
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE RankNTypes #-}
```
lehins: So is this extension
```suggestion
```

eras/alonzo/impl/testlib/Test/Cardano/Ledger/Alonzo/Arbitrary.hs
```diff
@@ -205,14 +208,22 @@ instance
...
+  Gen (AlonzoScript era)
+genAlonzoScript langs = do
+  lang <- elements langs -- The language is not present in the Script serialization
+  frequency
+    [ (1, pure (alwaysSucceeds lang 1))
+    , (1, pure (alwaysFails lang 1))
+    , (10, TimelockScript <$> arbitrary)
+    ]
+
+instance forall c. Crypto c => Arbitrary (AlonzoScript (AlonzoEra c)) where
```
lehins: Is there a reason you are adding these `forall`s?
```suggestion
instance Crypto c => Arbitrary (AlonzoScript (AlonzoEra c)) where
```

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/Classes.hs
```diff
@@ -40,6 +40,7 @@ import Test.Cardano.Ledger.Alonzo.Arbitrary ()
 import Test.Cardano.Ledger.Babbage.Arbitrary ()
 import Test.Cardano.Ledger.Constrained.Combinators (errorMess)
 import Test.Cardano.Ledger.Constrained.Size (AddsSpec (..), Size (..), genFromIntRange, genFromNonNegIntRange)
+import Test.Cardano.Ledger.Conway.Arbitrary ()
 import Test.Cardano.Ledger.Core.Arbitrary ()
```
lehins: The one below is now redundant
```suggestion
import Test.Cardano.Ledger.Conway.Arbitrary ()
```

libs/cardano-ledger-api/test/Test/Cardano/Ledger/Api/Tx/Out.hs
```diff
@@ -19,6 +19,7 @@ import Data.Functor.Identity
 import Lens.Micro
 import Test.Cardano.Ledger.Babbage.Arbitrary ()
 import Test.Cardano.Ledger.Common
+import Test.Cardano.Ledger.Conway.Arbitrary ()
```
lehins: All instances from `Test.Cardano.Ledger.Babbage.Arbitrary` are implicitly re-exported from `Test.Cardano.Ledger.Conway.Arbitrary`
```suggestion
import Test.Cardano.Ledger.Common
import Test.Cardano.Ledger.Conway.Arbitrary ()
```

eras/conway/impl/CHANGELOG.md
```diff
@@ -38,6 +38,9 @@
 * Add `ToCBOR`/`FromCBOR` for `ConwayGovernance`
 * Remove `cgAlonzoGenesis` from `ConwayGenesis`.
 * Set `ConwayGenesis` as `TranslationContext`
+* Change the Conway txInfo to allow Plutus V3
+  NOTE - unlike V1 and V2, the ledger will no longer place the "zero ada" value
+  in the script context for the transaction minting field.
```
lehins: This needs to move now to a new section, 1.1.0.0 is now released

eras/babbage/impl/CHANGELOG.md
```diff
@@ -6,6 +6,8 @@
 * Add `ToJSON` instance for `BabbagePParams Identity` and `BabbagePParams StrictMaybe`
 * Removed validation function `validateOutputTooBigUTxO`, in favor of the same function
   from `cardano-ledger-alonzo`.
+* Added support for Plutus V3 in the types and functions that use `Language`.
+  (Note that the Alonzo ledger era rules do not allow V3 scripts, however.).
```
lehins: Same here, new minor version bump needs to start with this change

lehins (commented):
We've now released version of packages that you've started with, so some of the versions and changelogs need to be updated

zliu41 (changes requested):
Please use version 1.5.0.0 for plutus core.
A small change to `Cardano.Ledger.Alonzo.TxInfo` is needed since `SatInt` no longer has `Integral` instance. See https://github.com/input-output-hk/cardano-ledger/pull/3382/files

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxWits.hs
JaredCorduan: I'll close this, but happy to change my style if the majority doesn't like it :)

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxAuxData.hs
JaredCorduan: I actually did try to do this, but got stuck on some type errors. I'd love help.

The type signature needed by `Cardano.Ledger.Alonzo.Scripts` is `Decode ('Closed 'Dense) ShortByteString`. The type needed here is a _list_ of short bytestrings, namely `Decode ('Closed d0) [Data.ByteString.Short.Internal.ShortByteString]` (for any `d0`, it seems).

So, letting `decodeAtV9 :: Decode ('Closed 'Dense) ShortByteString`, I tried decoding the `AlonzoTxWitsRaw` with:

```
      txWitnessField 7 =
        fieldA
          addScripts
          (fmap (PlutusScript PlutusV3) <$> (decodeList decodeAtV9))
```

but I get this error:
```
    • Couldn't match type: Decoder s0
                     with: Decode ('Closed d0)
```

So then I tried making `decodeAtV9` more parametric, using `Decode ('Closed a) ShortByteString` instead, but then the use site in `Scripts` fails with:

```
    • Couldn't match type ‘a’ with ‘'Dense’
      Expected: Decode ('Closed a) ShortByteString
        Actual: Decode ('Closed 'Dense) ShortByteString
```

So I'm not sure how to make this parametric in a way that works with the Coders library.

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxInfo.hs
JaredCorduan: yes indeed, thank you!! I wish I knew of a way to get the type checker to catch this for us... :thinking:

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxWits.hs
JaredCorduan: yep, I had the same trouble here as I described above in `TxAuxData`.

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxAuxData.hs
lehins: You don't need a reusable function for Coders, just for regular versioned decoding:

```haskell
      auxDataField 4 = fieldA (addPlutusScripts PlutusV3) (D (guardPlutus (natVersion @9) PlutusV3 >> decCBOR)
      auxDataField n = field (\_ t -> t) (Invalid n)
```

Somewhere in `Cardano.Ledger.Language`:

```haskell
guardPlutus v lang = 
  unlessDecoderVersionAtLeast v (fail (show lang <> " is not supported until " <> show v <> " major protocol version"))
```
Or even better:
```haskell
guardPlutus lang = 
  let v = case lang of
          PlutusV1 -> natVersion @5
          PlutusV2 -> natVersion @7
          PlutusV3 -> natVersion @9
  in unlessDecoderVersionAtLeast v $ 
        fail (show lang <> " is not supported until " <> show v <> " major protocol version")
```

michaelpj:
Is there a test for using PlutusV3? I think that should fail, since there's a key change we need to make on the Plutus side. We have a guard for the minimum protocol version in which a language is enabled, and at the moment it's set to `maxBound` for PlutusV3 because we didn't have confirmation about what protocol version it was going to be enabled in. So we need to change that before things will work, I would expect?

lehins:
> Is there a test for using PlutusV3?

There are not many tests for Conway yet and PlutusV3 will only work in Conway

> We have a guard for the minimum protocol version in which a language is enabled

Good. We have one too :slightly_smiling_face:

michaelpj:
Okay cool, just letting you know that you'll probably hit that as soon as you do try to write a test. If we're definitely sure that it's going to be version 9 then we can set it to that on our side and it'll work in the next release.

lehins:
@michaelpj Yes, it'll be version 9. Before major protocol version 9 Plutus script V3 is not deserializable.

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxAuxData.hs
JaredCorduan: that sounds great, thank you @lehins !

eras/alonzo/impl/CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Version history for `cardano-ledger-alonzo`
 
+## 1.2.0.0 (unreleased)
```
lehins: Let's not add the `(unreleased)`. I'll explain why tomorrow during the meeting, but if you are anxious to know the reasoning is described in #3379 
```suggestion
## 1.2.0.0
```

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxWits.hs
```diff
@@ -560,9 +569,15 @@ instance
...
           addScripts
-          (fmap (PlutusScript PlutusV2) <$> From)
+          (decodePlutus PlutusV2)
+      txWitnessField 7 =
+        fieldA
+          addScripts
+          (decodePlutus PlutusV3)
       txWitnessField n = field (\_ t -> t) (Invalid n)
 
+      decodePlutus lang = (fmap (PlutusScript lang) <$> D (guardPlutus lang >> decCBOR))
```
lehins: Redundant
```suggestion
      decodePlutus lang = fmap (PlutusScript lang) <$> D (guardPlutus lang >> decCBOR)
```

eras/babbage/impl/CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Version history for `cardano-ledger-babbage`
 
+## 1.2.0.0 (unreleased)
```
lehins: Same here
```suggestion
## 1.2.0.0
```

eras/babbage/impl/testlib/Test/Cardano/Ledger/Babbage/Arbitrary.hs
```diff
@@ -18,13 +18,18 @@ import Cardano.Ledger.Babbage.Tx
...
+import Cardano.Ledger.Language (Language (..))
 import Control.State.Transition (STS (PredicateFailure))
 import Data.Functor.Identity (Identity)
-import Test.Cardano.Ledger.Alonzo.Arbitrary ()
+import Test.Cardano.Ledger.Alonzo.Arbitrary (genAlonzoScript)
 import Test.QuickCheck
 
 deriving instance Arbitrary CoinPerByte
 
+instance forall c. Crypto c => Arbitrary (AlonzoScript (BabbageEra c)) where
```
lehins: Redundant `forall`
```suggestion
instance Crypto c => Arbitrary (AlonzoScript (BabbageEra c)) where
```

eras/conway/impl/CHANGELOG.md
```diff
@@ -1,9 +1,12 @@
 # Version history for `cardano-ledger-conway`
 
-## 1.2.0.0
+## 1.2.0.0 (unreleased)
```
lehins: ditto
```suggestion
## 1.2.0.0
```

eras/conway/test-suite/cddl-files/conway.cddl
```diff
@@ -432,6 +410,7 @@ potential_languages = 0 .. 255
 ;
 ; { ? 0 : [ 166* int ] ; Plutus v1, only 166 integers are used, but more are accepted (and ignored)
 ; , ? 1 : [ 175* int ] ; Plutus v2, only 175 integers are used, but more are accepted (and ignored)
+; , ? 2 : [ 175* int ] ; Plutus v3, only 175 integers are used, but more are accepted (and ignored)
```
```suggestion
; , ? 2 : [ 179* int ] ; Plutus v3, only 179 integers are used, but more are accepted (and ignored)
```

libs/cardano-ledger-binary/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Version history for `cardano-ledger-binary`
 
+## 1.2.0.0 (unreleased)
```
lehins: This is a minor version bump, because it is not a breaking change. Also requires a version change in the cabal file.

```suggestion
## 1.1.1.0
```


The other important part of every version bump is going and updating every affected dependency in this repo. In case of a minor bump, like this one only the packages that use the new functionality need adjustment. For example because `cardano-ledger-core` now uses `unlessDecoderVersionAtLeast` it will require the minor version update `cardano-ledger-binary >=1.1.1 && <1.2`

libs/cardano-ledger-api/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Version history for `cardano-ledger-api`
 
+## 1.2.0.0 (unreleased)
```
lehins:
```suggestion
## 1.2.0.0
```

libs/plutus-preprocessor/plutus-preprocessor.cabal
```diff
@@ -30,9 +30,9 @@ executable plutus-preprocessor
         cardano-ledger-alonzo >=1.1,
         flat,
         hashable >=1.4.1.0,
-        plutus-tx ^>=1.1,
-        plutus-tx-plugin ^>=1.1.1,
-        plutus-ledger-api ^>=1.1,
+        plutus-tx ^>=1.5,
+        plutus-tx-plugin ^>=1.5,
+        plutus-ledger-api ^>=1.5,
```
lehins: We can remove this, since this package is never released to CHaP. But feel free to leave it as is if you like.
```suggestion
        plutus-tx,
        plutus-tx-plugin,
        plutus-ledger-api,
```

libs/cardano-ledger-core/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Version history for `cardano-ledger-core`
 
+## 1.2.0.0 (unreleased)
```
```suggestion
## 1.2.0.0
```

eras/babbage/impl/CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
...
+
+* Added support for Plutus V3 in the types and functions that use `Language`.
+  (Note that the Alonzo ledger era rules do not allow V3 scripts, however.).
```
lehins:
```suggestion
  (Note that the Alonzo and Babbage ledger era rules do not allow V3 scripts, however.).
```

lehins (changes requested):
With respect to Haskell code it looks great. Versioning stuff still needs a little fixing. 

Every time the version is bumped, it also, must be updated in the cabal file. In case when the bump is major, it will require the upper bound in the dependencies to be updated too and potentially the lower bound as well.

eras/conway/test-suite/cddl-files/conway.cddl
JaredCorduan: good catch, thank you!!

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxInfo.hs
aniketd: Me too! :+1:

libs/cardano-ledger-binary/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Version history for `cardano-ledger-binary`
 
+## 1.1.1.0
```
lehins, line 3: It doesn't look like you updated the version in the `cardano-ledegr-binary.cabal` and the bounds for it in `cardano-ledger-core.cabal`? That is more important than even the changelog entries.

eras/babbage/test-suite/cardano-ledger-babbage-test.cabal
```diff
@@ -23,6 +23,7 @@ library
     exposed-modules:
         Test.Cardano.Ledger.Babbage.Examples.Consensus
         Test.Cardano.Ledger.Babbage.Serialisation.Generators
+        Test.Cardano.Ledger.Babbage.TxInfo
```
lehins, line 26: Addition of a new module requires a minor version bump: `1.1.1.0`

eras/conway/test-suite/cardano-ledger-conway-test.cabal
```diff
@@ -60,6 +60,7 @@ test-suite cardano-ledger-conway-test
     hs-source-dirs:   test
     other-modules:
         Test.Cardano.Ledger.Conway.Serialisation.CDDL
+        Test.Cardano.Ledger.Conway.TxInfo
```
lehins, line 63: Same here addition of a new module requires a minor bump: `1.1.1.0`

eras/conway/test-suite/cardano-ledger-conway-test.cabal
```diff
@@ -75,6 +76,7 @@ test-suite cardano-ledger-conway-test
         cardano-ledger-allegra,
         cardano-ledger-alonzo,
         cardano-ledger-babbage,
+        cardano-ledger-babbage-test,
```
lehins: This package now requires functionality in `cardano-ledger-babbage`, so bounds need to be reflected accordingly
```suggestion
        cardano-ledger-babbage-test >=1.1.1 && <1.2,
```

Note, that if this dependency was in the library section too, then it would be sufficient to update in the library section only, but since it is not present there bounds need to be set here.

eras/babbage/impl/CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Version history for `cardano-ledger-babbage`
 
+## 1.2.0.0
```
lehins: Version in cabal need to be updated too, as well as the bounds in cardano-ledger-conway

libs/cardano-ledger-api/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Version history for `cardano-ledger-api`
 
+## 1.2.0.0
```
lehins, line 3: Version in cabal need to be updated too

eras/babbage/impl/CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
...
+
+* Added support for Plutus V3 in the types and functions that use `Language`.
+  (Note that the Alonzo and Babbage ledger era rules do not allow V3 scripts, however.).
```
lehins, line 11:
```suggestion
  (Note that the Alonzo and Babbage ledger era rules do not allow V3 scripts, however.).
  Addition of `babbageTxInfoV1` and `babbageTxInfoV2`
```

libs/cardano-ledger-pretty/src/Cardano/Ledger/Pretty/Alonzo.hs
```diff
@@ -69,6 +69,7 @@ instance
 ppLanguage :: Language -> PDoc
 ppLanguage PlutusV1 = ppString "PlutusV1"
 ppLanguage PlutusV2 = ppString "PlutusV2"
+ppLanguage PlutusV3 = ppString "PlutusV3"
```
lehins, line 72: This now requires a `PlutusV3` constructor, which is only available in `cardano-ledger-core-1.2.0.0`, so bounds need to be updated in `cardano-ledger-pretty`

lehins (changes requested):
Versions in cabal files haven't been updated yet

eras/babbage/impl/CHANGELOG.md
JaredCorduan: good catch, thank you!

eras/babbage/test-suite/cardano-ledger-babbage-test.cabal
JaredCorduan: ah, so the plan is to **_not_** do change log version for test packages, but still do cabal versioning?

lehins: Yes versioning is more important than changelogs, without it releasing to CHaPs will not work.

If you'd like, we could do changelog for test-suites as well, but considering we are almost exclusive users of those test suites and the fact that we'll deprecate them eventually, I figured it would be an unnecessary overhead. However, we still do release them to CHaPs, cause consensus uses some parts of it, so versioning is necessary.

That is another major reason why I want to deprecate them and move the tests into the era packages themselves, to remove this overhead of separate versioning and releasing

JaredCorduan: nope, just trying to ensure we are all agreed to a common plan. I appreciate all your help with this!

eras/conway/impl/CHANGELOG.md
```diff
@@ -4,13 +4,17 @@
 
 * Added `ConwayDelegCert` and `Delegatee` #3372
 * Removed `toShelleyDCert` and `fromShelleyDCertMaybe` #3372
+<<<<<<< HEAD
```
lehins: Conflict resolution bug :slightly_smiling_face: 
```suggestion
```

libs/cardano-ledger-pretty/CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
  * Added `PrettyA` instances for:
    * `Delegatee era` #3372
    * `ConwayDelegCert c` #3372
+ * Plutus V3 is now pretty
```
lehins, line 8: Very pretty :smile:

lehins (approved):
Awesome stuff!

eras/conway/impl/CHANGELOG.md
JaredCorduan: :facepalm: thanks

JaredCorduan:
I don't know if I'm going mad or what, this says:

```
1 out of 201 tests failed (1305.54s)
Test suite cardano-ledger-shelley-ma-test: FAIL
```

But I don't see a single failed test.

PR #3366, opened 2023-04-03
Minor api improvements

lehins:
# Description

* Rename TxId field from _unTxId to unTxId
* Some TxAuxData realeted cleanups:
  * Remove redundant pattern synonym `AllegraTxAuxData'` and `AlonzoTxAuxData'`
  * Hide internal `AllegraTxAuxDataRaw` constructor with `atadrMetadata` and `atadrTimelock`
    record fields.

# Checklist

- [ ] ~New tests are added if needed and existing tests are updated~
- [x] Self-reviewed the diff

### DIFF
--- a/libs/cardano-ledger-core/src/Cardano/Ledger/TxIn.hs
+++ b/libs/cardano-ledger-core/src/Cardano/Ledger/TxIn.hs
@@ -17,6 +17,7 @@
 
 module Cardano.Ledger.TxIn (
   TxId (..),
+  _unTxId,
   TxIn (TxIn),
   mkTxInPartial,
   TxIx,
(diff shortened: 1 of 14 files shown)


libs/cardano-ledger-core/src/Cardano/Ledger/TxIn.hs
```diff
@@ -52,7 +52,7 @@ import NoThunks.Class (NoThunks (..))
 -- ====================================================================================
 
 -- | A unique ID of a transaction, which is computable from the transaction.
-newtype TxId c = TxId {_unTxId :: SafeHash c EraIndependentTxBody}
```
JaredCorduan, line 55: given that this one might be used by sever downstream libraries, would it makes sense to make a standalone `_unTxId` function with the same semantics, and immediately deprecate it?

JaredCorduan (approved):
LGTM!

libs/cardano-ledger-core/src/Cardano/Ledger/TxIn.hs
lehins: We can't really create a function with exact semantics as a record field has. In other words a function cannot be used to for updates or construction, eg:
```haskell
let txId = TxId { _unTxId = "...."}
```

However, adding at least a deprecated function `_unTxId` will not hurt. Done.

JaredCorduan: of course, that makes sense, thanks! what you added is exactly what I had in mind.

PR #3367, opened 2023-04-04
Adjust `CostModel`'s `FromJSON` instance

lehins:
# Description

It turns out cardano-node relies on JSON for CostModel for PlutusV2 to also contain the parameter names, not just a list of values. This PR implements this backwards compatibility in a forward compatible way.

* Addition of `costModelToMap`, `costModelFromMap` and `costModelParamNames`
* Made it possible for `FromJSON` to decode `CostModels` both as the new approach:
  1. as a list of cost models values,
  2. and the old approach of mapping from the parameter name to the cost model value

# Checklist

- [x] Self-reviewed the diff


JaredCorduan (approved):
looks great!

PR #3369, opened 2023-04-05
Added `VState`, Renamed `DPState` to `CertState`

Soupstraw:
# Description

This PR does the following:
 * Rename `DPState` to `CertState`, and any functions that reference `DPState` are renamed to refer to `CertState`.
 * ~~Replace `DELPL` rule with `CERT` starting from Conway era.~~
 * Add the `VState` type and add it to `CertState`
 * Parametrize `PState` and `DState` by `era`

Although this is a large PR, most of the changes are due to the `CertState` renaming which was done with `sed`. There are also a lot of changes due to `CertState` now being parametrized by era. To help with reviewing, here is a list with some of the files that have important changes:
* `libs/cardano-ledger-core/src/Cardano/Ledger/CertState.hs`
This file defines the new `VState` type and adds it to the `CertState`. This is also the file where the `DPState` is renamed to `CertState` and is parametrized by era. Also, `PState` and `DState` are now parametrized by era.

The files that I did not mention should be pretty straightforward changes - renaming functions/types, updating constraints, updating tests etc.

# Checklist

- [x] Self-reviewed the diff


JaredCorduan (approved):
Looks great to me!

I suspect that the "crypto parametrization to era parametrization" change will cause a bit of downstream work, so that we should not merge this until @lehins is ready for the final ledger release.

libs/cardano-ledger-core/src/Cardano/Ledger/CertState.hs
```diff
@@ -174,94 +176,126 @@ instance (Crypto c, DecShareCBOR (InstantaneousRewards c)) => DecShareCBOR (DSta
...
   , "retiring" .= psRetiring
   , "deposits" .= psDeposits
   ]
 
+data VState era = VState
+  { vsDReps :: !(Set (Credential 'Voting (EraCrypto era)))
+  , vsCCHotKeys ::
+      !( Map
+          (KeyHash 'Voting (EraCrypto era))
+          (KeyHash 'Voting (EraCrypto era))
```
lehins, line 261: Not suggest we do this in this PR. We need to think about either creating newtype wrappers around these KeyHashes, or creating another KeyRole. We need to disambiguate these two somehow, otherwise it is way too easy to mess them up.

libs/cardano-ledger-core/src/Cardano/Ledger/CertState.hs
```diff
@@ -174,94 +176,126 @@ instance (Crypto c, DecShareCBOR (InstantaneousRewards c)) => DecShareCBOR (DSta
...
+
 -- | The state associated with the DELPL rule, which combines the DELEG rule
 -- and the POOL rule.
-data DPState c = DPState
-  { dpsDState :: !(DState c)
-  , dpsPState :: !(PState c)
+data CertState era = CertState
+  { certDState :: !(DState era)
+  , certPState :: !(PState era)
+  , certVState :: !(VState era)
```
lehins: Considering that you are already changing this type can you swap the order of `certPState` and `certDState` as well? It will be less confusing because that is the order they are being serialized. Also it is better for sharing to put `VState` first (or at least before `DState`), this is because the credentials of DReps then can be shared inside the `DState`. I'll submit a separate PR that fixes the sharing, cause I don't think I'll be able to figure it out directly in github without the help of a type checker.
```suggestion
  { certVState :: !(VState era)
  , certPState :: !(PState era)
  , certDState :: !(DState era)
```

libs/cardano-ledger-core/src/Cardano/Ledger/CertState.hs
```diff
@@ -277,36 +311,38 @@ instance Crypto c => DecShareCBOR (InstantaneousRewards c) where
...
-toDPStatePairs :: (KeyValue a, Crypto c) => DPState c -> [a]
-toDPStatePairs DPState {..} =
-  [ "dstate" .= dpsDState
-  , "pstate" .= dpsPState
+instance Era era => EncCBOR (CertState era) where
+  encCBOR CertState {certPState, certDState, certVState} =
+    encodeListLen 3
+      <> encCBOR certPState -- We get better sharing when encoding pstate before dstate
+      <> encCBOR certDState
+      <> encCBOR certVState
```
lehins: As I mentioned in the previous comment, this will be better for sharing
```suggestion
      <> encCBOR certVState
      <> encCBOR certPState
      <> encCBOR certDState
```

libs/cardano-ledger-core/src/Cardano/Ledger/CertState.hs
```diff
@@ -277,36 +311,38 @@ instance Crypto c => DecShareCBOR (InstantaneousRewards c) where
...
+
+instance Era era => DecShareCBOR (CertState era) where
+  type Share (CertState era) = (Interns (Credential 'Staking (EraCrypto era)), Interns (KeyHash 'StakePool (EraCrypto era)))
+  decSharePlusCBOR = decodeRecordNamedT "CertState" (const 3) $ do
+    certPState <- decSharePlusLensCBOR _2
+    certDState <- decSharePlusCBOR
+    certVState <- lift decCBOR
```
lehins: I'll fix it properly later, but let's at least fix the ordering.
```suggestion
    certVState <- lift decCBOR -- TODO: add sharing of DRep credentials
    certPState <- decSharePlusLensCBOR _2
    certDState <- decSharePlusCBOR
```

libs/cardano-ledger-core/testlib/Test/Cardano/Ledger/Core/Arbitrary.hs
```diff
@@ -468,21 +468,25 @@ instance Crypto c => Arbitrary (UMap c) where
...
   shrink = genericShrink
 
-instance Crypto c => Arbitrary (PState c) where
+instance Era era => Arbitrary (PState era) where
   arbitrary = PState <$> arbitrary <*> arbitrary <*> arbitrary <*> arbitrary
   shrink = genericShrink
 
+instance Era era => Arbitrary (VState era) where
+  arbitrary = VState <$> arbitrary <*> arbitrary
+  shrink = genericShrink
```
lehins: Generic shrink doesn't work right on Maps. We should probably remove it fro `PState` and `DState` as well
```suggestion
```

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/Same.hs
```diff
@@ -114,26 +116,32 @@ eqVia pcf x y = if x == y then Nothing else Just (notEq (pcf x) (pcf y))
 class Same era t where
   same :: Proof era -> t -> t -> [(String, Maybe PDoc)]
 
-instance (EraCrypto era ~ c) => Same era (DPState c) where
-  same proof (DPState d1 p1) (DPState d2 p2) =
-    extendLabel "DState " (same proof d1 d2) ++ extendLabel "PState " (same proof p1 p2)
+instance Same era (CertState era) where
+  same proof (CertState d1 p1 v1) (CertState d2 p2 v2) =
+    extendLabel "DState " (same proof d1 d2) ++ extendLabel "PState " (same proof p1 p2) ++ extendLabel "VState " (same proof v1 v2)
```
lehins: Line is getting a little too long
```suggestion
    extendLabel "DState " (same proof d1 d2) ++ 
    extendLabel "PState " (same proof p1 p2) ++ 
    extendLabel "VState " (same proof v1 v2)
```

eras/allegra/impl/CHANGELOG.md
```diff
@@ -1,5 +1,14 @@
 # Version history for `cardano-ledger-allegra`
 
+## 1.2.0.0
+
+* Replace `DPState c` with `CertState era`
+* Parametrize `DState` and `PState` by era
```
lehins: This is not a relevant change for Allegra
```suggestion
```

eras/allegra/impl/CHANGELOG.md
```diff
@@ -1,5 +1,14 @@
 # Version history for `cardano-ledger-allegra`
 
+## 1.2.0.0
```
lehins: As far as `cardano-ledger-allegra` package is concerned this is not a breaking change.

Also need to be updated in cabal file
```suggestion
## 1.1.1.0
```

eras/alonzo/impl/CHANGELOG.md
```diff
@@ -1,5 +1,14 @@
 # Version history for `cardano-ledger-alonzo`
 
+## 1.2.0.0
+
+* Replace `DPState c` with `CertState era`
+* Parametrize `DState` and `PState` by era
```
lehins: Same here and cabal file
```suggestion
## 1.1.1.0

* Replace `DPState c` with `CertState era`
```

eras/babbage/impl/CHANGELOG.md
```diff
@@ -1,5 +1,14 @@
 # Version history for `cardano-ledger-babbage`
 
+## 1.2.0.0
+
+* Replace `DPState c` with `CertState era`
+* Parametrize `DState` and `PState` by era
```
lehins: Same here and cabal file
```suggestion
## 1.1.1.0

```

eras/conway/impl/CHANGELOG.md
```diff
@@ -4,6 +4,14 @@
 
 * Added `ConwayDelegCert` and `Delegatee` #3372
 * Removed `toShelleyDCert` and `fromShelleyDCertMaybe` #3372
+* Replace `DELPL` rule with `CERT`
+* Replace `DPState c` with `CertState era`
+* Parametrize `DState` and `PState` by era
```
lehins: Not a change relevant for this package
```suggestion
* Replace `DPState c` with `CertState era`
```

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Delegs.hs
```diff
@@ -0,0 +1,89 @@
...
+import Data.Map.Strict (Map)
+import Data.Sequence (Seq)
+import GHC.Generics (Generic)
+
+data ConwayDelegsPredFailure era
+  = DelegateeNotRegisteredDELEG
+      !(KeyHash 'StakePool (EraCrypto era)) -- target pool which is not registered
+  | WithdrawalsNotInRewardsDELEGS
+      !(Map (RewardAcnt (EraCrypto era)) Coin) -- withdrawals that are missing or do not withdrawal the entire amount
+  | CertFailure (PredicateFailure (EraRule "CERT" era)) -- Subtransition Failures
```
lehins: It would be good to turn those into haddock, rather tyhan comments
```suggestion
  = DelegateeNotRegisteredDELEG !(KeyHash 'StakePool (EraCrypto era))
   -- ^ Target pool which is not registered
  | WithdrawalsNotInRewardsDELEGS !(Map (RewardAcnt (EraCrypto era)) Coin)
   -- ^ Withdrawals that are missing or do not withdrawal the entire amount
  | CertFailure (PredicateFailure (EraRule "CERT" era))
   -- ^ Sub-transition Failures
```

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Delegs.hs
```diff
@@ -0,0 +1,89 @@
...
+module Cardano.Ledger.Conway.Rules.Delegs (
+  ConwayDELEGS,
+  ConwayDelegsPredFailure (..),
+  ConwayDelegsEvent (..),
+) where
+
+import Cardano.Ledger.BaseTypes (ShelleyBase)
+import Cardano.Ledger.Conway.Core (Era (..), EraRule, EraTx)
+import Cardano.Ledger.Conway.Era (ConwayCERT, ConwayDELEGS)
+import Cardano.Ledger.Shelley.API (CertState, Coin, DCert, DState, DelegsEnv, DelplEnv, KeyHash, KeyRole (..), PState, RewardAcnt)
```
lehins: Maybe turn into a multiline?

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Delegs.hs
```diff
@@ -0,0 +1,89 @@
...
+  ) =>
+  STS (ConwayDELEGS era)
+  where
+  type State (ConwayDELEGS era) = CertState era
+  type Signal (ConwayDELEGS era) = Seq (DCert (EraCrypto era))
+  type Environment (ConwayDELEGS era) = DelegsEnv era
+  type BaseM (ConwayDELEGS era) = ShelleyBase
+  type
+    PredicateFailure (ConwayDELEGS era) =
+      ConwayDelegsPredFailure era
```
lehins: On the other hand this isn't too long, and switching it a single line will make it more consistent with the rest of type family instances above and below
```suggestion
  type PredicateFailure (ConwayDELEGS era) = ConwayDelegsPredFailure era
```

libs/cardano-ledger-core/CHANGELOG.md
```diff
@@ -1,5 +1,18 @@
 # Version history for `cardano-ledger-core`
 
+## 1.2.0.0
```
lehins: This is a correct bump. However, all versions also need to be bumped in the cabal file, not only in the changelog.

Also, because this change is now required by all of the packages that have been modified in this PR, they all need to get a lower bound update: `cardano-ledger-core ^>= 1.2`

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Translation.hs
```diff
@@ -101,12 +104,49 @@ instance Crypto c => TranslateEra (AlonzoEra c) EpochState where
...
 
+instance Crypto c => TranslateEra (AlonzoEra c) DState where
+  translateEra _ ls =
+    pure
+      DState
+        { dsUnified = dsUnified ls
+        , dsIRewards = dsIRewards ls
+        , dsGenDelegs = dsGenDelegs ls
+        , dsFutureGenDelegs = dsFutureGenDelegs ls
+        }
```
lehins: I think this is one one of those places where RecordWildCards really shine:
I'd suggest this change on the introduced translations for `PState` and `DState` in other packages as well.

However, there is nothing wrong with current implementation, so it is up to you if you wanna change it or not.
```suggestion
  translateEra _ DState {..} = pure DState {..}
```

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Ledger.hs
```diff
@@ -244,7 +241,7 @@ ledgerTransition = do
             )
       else pure dpstate
 
-  let DPState dstate _pstate = dpstate
+  let CertState dstate _pstate _vstate = dpstate
```
lehins: This will also need a rename of the binding as well
```suggestion
  let dstate = certDState certState
```

eras/mary/impl/CHANGELOG.md
```diff
@@ -1,5 +1,14 @@
 # Version history for `cardano-ledger-mary`
 
+## 1.2.0.0
+
+* Replace `DPState c` with `CertState era`
+* Parametrize `DState` and `PState` by era
```
lehins: Ditto
```suggestion
## 1.1.1.0

```

eras/shelley/impl/CHANGELOG.md
```diff
@@ -1,8 +1,14 @@
 # Version history for `cardano-ledger-shelley`
 
-## 1.1.1.0
+## 1.2.0.0
 
 * Disable `TICKF` rule optimization: [#3375](https://github.com/input-output-hk/cardano-ledger/pull/3375)
+* Replace `DPState c` with `CertState era`
+* Parametrize `DState` and `PState` by era
```
lehins, line 7: In `cardano-ledegr-shelley` this change is somewhat relevant, but only because it affected types of arguments for some of the functions, which is a user facing feature.

eras/shelley/impl/CHANGELOG.md
```diff
@@ -1,8 +1,14 @@
 # Version history for `cardano-ledger-shelley`
 
-## 1.1.1.0
+## 1.2.0.0
```
lehins, line 3: This is indeed a breaking change, which means all of the downstream packages (like `cardano-ledger-shelley-test`) that have been affected by this change now need a lower bound update. Naturally this version also need to be updated in the cabal file

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Delegs.hs
```diff
@@ -193,26 +211,31 @@ instance
...
+  forall era someCERT someDELEGS certFailure.
   ( EraTx era
-  , ShelleyEraTxBody era
-  , Embed (EraRule "DELPL" era) (ShelleyDELEGS era)
-  , Environment (EraRule "DELPL" era) ~ DelplEnv era
-  , State (EraRule "DELPL" era) ~ DPState (EraCrypto era)
-  , Signal (EraRule "DELPL" era) ~ DCert (EraCrypto era)
+  , Environment someCERT ~ DelplEnv era
+  , State someCERT ~ CertState era
+  , Signal someCERT ~ DCert (EraCrypto era)
```
lehins: This whole approach of reusing `delegsTransition` in CONWAY isn't gonna work, at least not the way you are doing it. 

Note that the certificate `DCert` will be different in Conway, so that is already a problem. 

I suggest reverting changes to DELEGS in this PR and handling it as a separate PR in a different manner.

There is a better way of solving this problem. The classic way: extracting parts of the `delegsTransition` into separate functions and reusing them directly in a newly defined CERT rule. This whole complication with `ShelleyDelegsFail`  is definitely an unnecessary complication.

eras/shelley/impl/CHANGELOG.md
```diff
@@ -1,8 +1,14 @@
...
 
 * Disable `TICKF` rule optimization: [#3375](https://github.com/input-output-hk/cardano-ledger/pull/3375)
+* Replace `DPState c` with `CertState era`
+* Parametrize `DState` and `PState` by era
+* Rename `obligationDPState` to `obligationCertState`
+* Rename `keyCertsRefundsDPState` to `keyCertsRefundsCertState`
+* Rename `totalCertsDepositsDPState` to `totalCertsDepositsCertState`
+* Add `ShelleyDelegsFail` typeclass
```
lehins: Not to forget removing it from the changelog too
```suggestion
```

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Epoch.hs
```diff
@@ -187,12 +187,12 @@ epochTransition = do
     trans @(EraRule "POOLREAP" era) $
       TRC (pp, PoolreapState utxoSt acnt dstate pstate', e)
 
-  let adjustedDPstate = DPState dstate' pstate''
+  let adjustedDPstate = CertState dstate' pstate'' vstate
```
lehins:
```suggestion
  let adjustedCertState = CertState dstate' pstate'' vstate
```

libs/cardano-ledger-core/src/Cardano/Ledger/CertState.hs
```diff
@@ -174,94 +176,126 @@ instance (Crypto c, DecShareCBOR (InstantaneousRewards c)) => DecShareCBOR (DSta
...
-toPStatePair :: (KeyValue a, Crypto c) => PState c -> [a]
+toPStatePair :: (KeyValue a, Era era) => PState era -> [a]
 toPStatePair PState {..} =
   [ "stakePoolParams" .= psStakePoolParams
   , "futureStakePoolParams" .= psFutureStakePoolParams
   , "retiring" .= psRetiring
   , "deposits" .= psDeposits
   ]
 
+data VState era = VState
```
lehins, line 256: So, we plan to add DRep Deposits into this type later, right?

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/PoolReap.hs
```diff
@@ -93,7 +93,7 @@ instance (Default (ShelleyPoolreapState era), EraPParams era) => STS (ShelleyPOO
     [ PostCondition
         "Deposit pot must equal obligation (PoolReap)"
         ( \(TRC (_, _, _)) st ->
-            obligationDPState (DPState (prDState st) (prPState st)) == utxosDeposited (prUTxOSt st)
+            obligationCertState (CertState (prDState st) (prPState st) def) == utxosDeposited (prUTxOSt st)
```
lehins: I suspect this `def`  will be a problem when we add DRep deposits. I think we might have to add the full `CertState` to the `ShelleyPoolreapState`. 

We don't have to solve it now in this PR though. However, might be worth noting as a TODO.

eras/shelley/test-suite/src/Test/Cardano/Ledger/Shelley/Generator/Update.hs
```diff
@@ -327,7 +327,7 @@ genUpdate
       nodes <- take 5 <$> QC.shuffle coreNodes
 
       let e = epochFromSlotNo s
-          (GenDelegs genDelegs) = (dsGenDelegs . dpsDState) delegPoolSt
+          (GenDelegs genDelegs) = (dsGenDelegs . certDState) delegPoolSt
```
lehins: I know you didn't write this, but this is bugging me anyways :slightly_smiling_face: 
```suggestion
          GenDelegs genDelegs = dsGenDelegs (certDState delegPoolSt)
```

eras/shelley/test-suite/src/Test/Cardano/Ledger/Shelley/Rules/Deposits.hs
```diff
@@ -102,7 +101,7 @@ rewardDepositDomainInvariant ::
   SourceSignalTarget (CHAIN era) ->
   Property
 rewardDepositDomainInvariant SourceSignalTarget {source = chainSt} =
-  let LedgerState {lsDPState = DPState dstate _} = (esLState . nesEs . chainNes) chainSt
+  let LedgerState {lsCertState = CertState dstate _ _} = (esLState . nesEs . chainNes) chainSt
```
lehins: This will help with fields reordering
```suggestion
  let LedgerState {lsCertState = CertState {certDState = dstate}} = (esLState . nesEs . chainNes) chainSt
```

eras/shelley/test-suite/src/Test/Cardano/Ledger/Shelley/Rules/IncrementalStake.hs
```diff
@@ -140,8 +140,8 @@ incrStakeComp SourceSignalTarget {source = chainSt, signal = block} =
         where
           utxoBal = coinBalance u'
           incrStakeBal = fold (credMap sd') <> fold (ptrMap sd')
-          ptrs = ptrsMap . dpsDState $ dp
-          ptrs' = ptrsMap . dpsDState $ dp'
+          ptrs = ptrsMap . certDState $ dp
+          ptrs' = ptrsMap . certDState $ dp'
```
lehins: I just don't get it why people write it that way
```suggestion
          ptrs = ptrsMap $ certDState dp
          ptrs' = ptrsMap $ certDState dp'
```

eras/shelley/test-suite/src/Test/Cardano/Ledger/Shelley/Rules/TestChain.hs
```diff
@@ -199,7 +198,7 @@ poolTraceFromBlock chainSt block =
       let (LedgerEnv s _ pp _) = ledgerEnv
        in PoolEnv s pp
     poolSt0 =
-      let LedgerState _ (DPState _ poolSt0_) = ledgerSt0
+      let LedgerState _ (CertState _ poolSt0_ _) = ledgerSt0
        in poolSt0_
```
lehins: This is way more compact and forward compatbile
```suggestion
       credPState (lsCredState ledgerSt0)
```

eras/shelley/test-suite/src/Test/Cardano/Ledger/Shelley/Rules/TestChain.hs
```diff
@@ -230,7 +229,7 @@ delegTraceFromBlock chainSt block =
           ptr = Ptr s txIx dummyCertIx
        in DelegEnv s ptr reserves pp
     delegSt0 =
-      let LedgerState _ (DPState delegSt0_ _) = ledgerSt0
+      let LedgerState _ (CertState delegSt0_ _ _) = ledgerSt0
        in delegSt0_
```
lehins: Same here
```suggestion
      credDState (lsCredState ledgerSt0)
```

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Delegs.hs
```diff
@@ -0,0 +1,89 @@
...
+  type Event (ConwayDELEGS era) = ConwayDelegsEvent era
+
+  transitionRules = [delegsTransition @_ @(EraRule "CERT" era) @(ConwayDELEGS era) @(ShelleyDelplPredFailure era)]
```
lehins: We gonna have to redefine this rule from scratch. Reusing it like that will not work
```suggestion
  transitionRules = []
```
Let's do it in a separate PR though.

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Delegs.hs
```diff
@@ -0,0 +1,89 @@
...
+
+deriving instance Show (PredicateFailure (EraRule "CERT" era)) => Show (ConwayDelegsPredFailure era)
+
+instance
+  certFailure ~ PredicateFailure (EraRule "CERT" era) =>
+  ShelleyDelegsFail certFailure (ConwayDelegsPredFailure era) era
+  where
+  delegateeNotRegisteredDELEG = DelegateeNotRegisteredDELEG
+  withdrawalsNotInRewardsDELEGS = WithdrawalsNotInRewardsDELEGS
+  certFailure = CertFailure
```
lehins: Not applicable
```suggestion
```

lehins (changes requested):
This is great work. Couple of general comments:

* We need to be more diligent with versions and bounds. It is much easier to fix the bounds in the same PR that introduces breaking changes, rather than afterwards
* Reusing DELEG rule from shelley can be better. We can discuss this one later

eras/allegra/impl/CHANGELOG.md
Soupstraw: My thinking was that these changes cause the constraints of some of the functions in this package to change.

eras/shelley/test-suite/src/Test/Cardano/Ledger/Shelley/Rules/IncrementalStake.hs
Soupstraw: I usually don't write it like this either, but this one somehow slipped in

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Translation.hs
Soupstraw: That's pretty clever. I updated the other `translateEra`s as well.

libs/cardano-ledger-core/src/Cardano/Ledger/CertState.hs
Soupstraw: Right now the plan is to keep track of all the deposits in UTxOState.

eras/allegra/impl/CHANGELOG.md
lehins: Which functions? I didn't see any changes to constraints

libs/cardano-ledger-core/src/Cardano/Ledger/CertState.hs
lehins: That is strange. Deposits for PoolStake operators and Delegators is kept in CertState. What is the reasoning for DReps' deposits to be in a different place?

libs/cardano-ledger-core/src/Cardano/Ledger/CertState.hs
Soupstraw: ~That sounds like an easy fix, I'll do it in this PR.~ Never mind, turns out to be more complicated than it looks.

eras/allegra/impl/CHANGELOG.md
Soupstraw: Oh, ok that's right

eras/conway/impl/CHANGELOG.md
Soupstraw: But this one does have some changes in constraints related to the `CertState` renaming

lehins: Fair enough. Then only one of those entries is needed.

eras/conway/impl/CHANGELOG.md
```diff
@@ -4,6 +4,13 @@
 
 * Added `ConwayDelegCert` and `Delegatee` #3372
 * Removed `toShelleyDCert` and `fromShelleyDCertMaybe` #3372
+* Replace `DELPL` rule with `CERT`
```
lehins, line 7: Technically not in this PR, but feel free to leave it in since you are doing this in #3384 anyways

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Delegs.hs
```diff
@@ -108,16 +111,27 @@ data ShelleyDelegsPredFailure era
...
+  withdrawalsNotInRewardsDELEGS :: Map (RewardAcnt (EraCrypto era)) Coin -> a
+  certFailure :: certFailure -> a
+
+instance
+  certFailure ~ PredicateFailure (EraRule "DELPL" era) =>
+  ShelleyDelegsFail certFailure (ShelleyDelegsPredFailure era) era
+  where
+  delegateeNotRegisteredDELEG = DelegateeNotRegisteredDELEG
+  withdrawalsNotInRewardsDELEGS = WithdrawalsNotInRewardsDELEGS
+  certFailure = DelplFailure
```
lehins: Remnants of a previous attempt
```suggestion
```

libs/cardano-ledger-core/src/Cardano/Ledger/CertState.hs
lehins: We can deal with this later no need to worry about deposits in this PR

lehins (approved):
Awesome work. Thank you!

aniketd (approved):
LGTM :raised_hands:

PR #3371, opened 2023-04-05
Replace UMapCompact with UMap and add tests

### LINKED ISSUE #3272: Remove `Data.UMap`
`Data.UMap` has moved to `Cardano.Ledger.UMapCompact`, but the old code is still there and is pretty dead. This ticket will track all of these related tasks:

* [x] - Remove `Data.UMap` from `cardano-data` package
* [x] - Remove dependency of `set-algebra` on `Data.UMap`
* [x] - Rename `Cardano.Ledger.UMapCompact` -> `Cardano.Ledger.UMap`, since there will be no other version of `UMap` in the repo, thus no ambiguity. Also it will help avoid confusion, since the type in the module is named `UMap`, not `UMapCompact`
* [x] - Adjust [old and unused tests located here](https://github.com/input-output-hk/cardano-ledger/blob/b2a76bf94bc6ac6d97efcad46a08ee3d66493f1a/libs/cardano-ledger-core/test/Test/Cardano/Ledger/UMapCompactSpec.hs) to use hspec and new version of `UMap`. Rename the module to `UMapSpec.hs`

aniketd:
# Description

Closes #3272 

# Checklist

- [x] Self-reviewed the diff


aniketd:
2 tests are failing still, unexpectedly. I'm looking into it :+1:

JaredCorduan (approved):
love it! thank you!

aniketd:
Those tests should now pass :crossed_fingers:

libs/cardano-ledger-core/src/Cardano/Ledger/UMap.hs
```diff
@@ -363,6 +364,10 @@ compactRewView x = Map.map rdReward $ unUnify (RewardDeposits x)
 depositView :: UMap c -> Map.Map (Credential 'Staking c) Coin
 depositView x = Map.map (fromCompact . rdDeposit) $ unUnify (RewardDeposits x)
 
+-- | Materialize the RDPairs Map from a 'UMap'
+rdPairView :: UMap c -> Map.Map (Credential 'Staking c) RDPair
+rdPairView x = unUnify (RewardDeposits x)
```
lehins, line 378: Could you deprecate `rewards` in favor of this function. `rdPairView` is a better name anyways.

libs/cardano-ledger-core/test/Test/Cardano/Ledger/UMapSpec.hs
lehins, line 1: My guess is you didn't do `git mv`? Cause the git history seems to have disappeared.

Can you please fix it, cause it will make the review easier and will help with history in general

libs/cardano-ledger-core/test/Test/Cardano/Ledger/UMapSpec.hs
```diff
@@ -0,0 +1,400 @@
...
+sizeTest (rdPairs, delegs, ptrs) =
+  let
+    umap = unify rdPairs delegs ptrs
+    rdPairsSize = size (RewardDeposits umap)
+    delegsSize = size (Delegations umap)
+    ptrsSize = size (Ptrs umap)
+   in
+    (Map.size rdPairs == rdPairsSize)
+      && (Map.size delegs == delegsSize)
+      && (Map.size ptrs == ptrsSize)
```
lehins: Let's improve failure reporting a bit
```suggestion
  Expectation
sizeTest (rdPairs, delegs, ptrs) = do
  let
    umap = unify rdPairs delegs ptrs
    rdPairsSize = size (RewardDeposits umap)
    delegsSize = size (Delegations umap)
    ptrsSize = size (Ptrs umap)
  Map.size rdPairs `shouldBe` rdPairsSize
  Map.size delegs `shouldBe` delegsSize
  Map.size ptrs `shouldBe` ptrsSize
```

libs/cardano-ledger-core/test/Test/Cardano/Ledger/UMapSpec.hs
```diff
@@ -0,0 +1,400 @@
...
+  ) ->
+  Bool
+unifyRoundTripTo (rdPairs, delegs, ptrs) =
+  let
+    umap = unify rdPairs delegs ptrs
+    rdPairs' = rdPairView umap
+    delegs' = delView umap
+    ptrs' = ptrView umap
+   in
+    rdPairs == rdPairs' && delegs == delegs' && ptrs == ptrs'
```
lehins: Same here:
```suggestion
  Expectation
unifyRoundTripTo (rdPairs, delegs, ptrs) = do
  let umap = unify rdPairs delegs ptrs
  rdPairView umap `shouldBe` rdPairs
  delView umap `shouldB` delegs
  ptrView umap `shouldBe` ptrs
```

libs/cardano-ledger-core/test/Test/Cardano/Ledger/UMapSpec.hs
```diff
@@ -0,0 +1,400 @@
...
+    rdPairs = rdPairView umap
+    delegs = delView umap
+    ptrs = ptrView umap
+   in
+    umap === unify rdPairs delegs ptrs
+
+spec :: Spec
+spec = do
+  describe "UMap" $ do
+    context "Invariant" $ do
```
lehins: Why `context` in one place and `describe` in another? I think using the same `describe` for consistency everywhere is a good practice.

libs/cardano-ledger-core/testlib/Test/Cardano/Ledger/Core/Arbitrary.hs
```diff
@@ -467,6 +472,62 @@ instance Crypto c => Arbitrary (UMap c) where
   arbitrary = UMap <$> arbitrary <*> arbitrary
   shrink = genericShrink
```
lehins: Could you remove this one. Generic should not be used with `Map` (consequently with `UMap` either)
```suggestion
```

libs/cardano-ledger-core/testlib/Test/Cardano/Ledger/Core/Arbitrary.hs
```diff
@@ -467,6 +472,62 @@ instance Crypto c => Arbitrary (UMap c) where
...
+-- | Used for testing UMap operations
+genValidTriples ::
+  Gen
+    ( Map.Map (Credential 'Staking StandardCrypto) RDPair
+    , Map.Map (Credential 'Staking StandardCrypto) (KeyHash 'StakePool StandardCrypto)
+    , Map.Map Ptr (Credential 'Staking StandardCrypto)
+    )
+genValidTriples = resize 50 $ do
+  nCreds <- chooseInt (1, 50)
+  creds :: [Credential 'Staking StandardCrypto] <- vectorOf nCreds arbitrary
```
lehins: Empty should not be excluded
```suggestion
  creds :: [Credential 'Staking StandardCrypto] <-  arbitrary
  let nCreds = length creds
```

libs/cardano-ledger-core/testlib/Test/Cardano/Ledger/Core/Arbitrary.hs
```diff
@@ -467,6 +472,62 @@ instance Crypto c => Arbitrary (UMap c) where
...
+  rdPairs :: [RDPair] <- vectorOf nCreds arbitrary
+  delegs :: [KeyHash 'StakePool StandardCrypto] <- vectorOf nCreds arbitrary
+  ptrs :: [Ptr] <- vectorOf nCreds arbitrary
```
lehins: Pointer map does not have to have the same number of elements as the credentials
```suggestion
  ptrs :: [Ptr] <- arbitrary
```

libs/cardano-ledger-core/testlib/Test/Cardano/Ledger/Core/Arbitrary.hs
```diff
@@ -467,6 +472,62 @@ instance Crypto c => Arbitrary (UMap c) where
...
   shrink = genericShrink
 
+-- | Used for testing UMap operations
+genValidTriples ::
+  Gen
+    ( Map.Map (Credential 'Staking StandardCrypto) RDPair
+    , Map.Map (Credential 'Staking StandardCrypto) (KeyHash 'StakePool StandardCrypto)
+    , Map.Map Ptr (Credential 'Staking StandardCrypto)
+    )
+genValidTriples = resize 50 $ do
```
lehins: Almost always it is better to `scale`, rather than `resize`, because if someone does `scale (*10) $ genValidTriples`, they will not get the size `50`, but the desired `500` (or rather `600` when `(* 2)` is used)
```suggestion
genValidTriples = scale (* 2) $ do
```

libs/cardano-ledger-core/testlib/Test/Cardano/Ledger/Core/Arbitrary.hs
```diff
@@ -467,6 +472,62 @@ instance Crypto c => Arbitrary (UMap c) where
...
+genValidUMap = do
+  (rdPairs, delegs, ptrs) <- genValidTriples
+  pure $ unify rdPairs delegs ptrs
+
+genExcluding :: (Eq a, Arbitrary a) => [a] -> Gen a
+genExcluding as = do
+  a <- arbitrary
+  if a `elem` as
+    then genExcluding as
+    else pure a
```
lehins: Avoid `elem` on lists because it is an O(n) operation, either use a `Set` or a `Map`. Looks like you are already using a Map, so:
```suggestion
genExcludingKey :: (Ord k, Arbitrary k) => Map k a -> Gen k
genExcludingKey as = do
  a <- arbitrary
  if a `Map.member` as
    then genExcludingKey as
    else pure a
```

libs/cardano-ledger-core/CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## 1.1.0.0
 
+* Rename `UMapCompact` to `UMap` and add tests for it.
```
lehins: `1.1.0.0` has now been released, so this PR should start a `1.2.0.0`. This will require a bump in cabal file as well.

You should record that the module name has changed. The way it is written now looks like a type rename. 
```suggestion
* Rename `Cardano.Ledger.UMapCompact` to `Cardano.Ledger.UMap` and add tests for it.
```
Also there is no mention of new function: `rdPairView`

Moreover, every package that has been affected by this rename, must have its lower bound be updated.

libs/cardano-ledger-core/testlib/Test/Cardano/Ledger/Core/Arbitrary.hs
```diff
@@ -467,6 +472,62 @@ instance Crypto c => Arbitrary (UMap c) where
...
+
+genInvariant :: Gen (Credential 'Staking StandardCrypto, Ptr, UMap StandardCrypto)
+genInvariant = do
+  umap@(UMap tripmap ptrmap) <- genValidUMap
+  let creds = Map.keys tripmap
+      ptrs = Map.keys ptrmap
+  cred <-
+    oneof
+      [ elements creds
+      , genExcluding creds
```
lehins:
```suggestion
      , genExcludingKey tripmap
```

libs/cardano-ledger-core/testlib/Test/Cardano/Ledger/Core/Arbitrary.hs
```diff
@@ -467,6 +472,62 @@ instance Crypto c => Arbitrary (UMap c) where
...
+      ]
+  ptr <-
+    oneof
+      [ elements ptrs
+      , genExcluding ptrs
```
lehins:
```suggestion
      , genExcludingKey ptrmap
```

libs/cardano-ledger-core/src/Cardano/Ledger/UMap.hs
```diff
@@ -17,7 +17,7 @@
 --   and a fourth one in the inverse direction.
 --   The advantage of using 'UMap' is that 'UMap' stores all the information compactly, by exploiting the
 --   the large amount of sharing in the 2 maps.
-module Cardano.Ledger.UMapCompact (
+module Cardano.Ledger.UMap (
```
lehins, line 20: It would be nice to reduce breakage and promote easier migration by adding a module in a separate commit `Cardano.Ledger.UMapCompact` that re-exports `Cardano.Ledger.UMap`, but is marked as deprecated.

libs/cardano-data/cardano-data.cabal
```diff
@@ -20,7 +20,6 @@ library
         Data.CanonicalMaps
         Data.Pulse
         Data.MapExtras
-        Data.UMap
```
lehins, line 23: Module has been removed. That is a breaking change, therefore it needs to be reflected in a cabal version and the changelog. 

Moreover, this removal affects older versions of `set-algebra`, so if you feel adventurous you could [add a revision](https://github.com/input-output-hk/cardano-haskell-packages#how-to-add-a-new-package-metadata-revision) on the already released [version of `set-algebra` ](https://github.com/input-output-hk/cardano-haskell-packages/blob/main/_sources/set-algebra/1.0.0.0/meta.toml) and add an upper bound on the `cardano-data`, but only after this PR is merged, of course.

libs/set-algebra/src/Control/Iterate/BaseTypes.hs
```diff
@@ -278,41 +270,6 @@ instance Iter Map.Map where
...
-  where
-  addkv (k, v) m comb = UM.insertWith' comb k v m
-  addpair = UM.insert'
-  removekey = UM.delete'
-  domain = UM.domain
-  range = UM.range
-
-instance
-  (Ord coin, Ord cred, Ord ptr) =>
-  Iter (UM.View coin cred pool ptr)
```
lehins, line 296: Removing instances is a breaking change. Must be reflected in the changelog and in a version bump.

lehins (changes requested):
Looking great. Some issues with versions and bounds, but this is a totally new practice for everyone on the team, so it will take a while to get used to.

libs/cardano-ledger-core/test/Test/Cardano/Ledger/UMapSpec.hs
aniketd: I did `git mv` but then squashed all into one. I will split it out and retain an individual commit with just that. Thanks!

libs/cardano-ledger-core/test/Test/Cardano/Ledger/UMapSpec.hs
aniketd: I thought that's how `context` was supposed to be used, as a sub-grouping of a main test that is `describe`d. But you are right, other than being a bit more readable it hardly adds any value. But it brings more potential confusion as `context` isn't that popular I guess. I will change it to just using `describe`. :+1:

libs/cardano-ledger-core/src/Cardano/Ledger/UMap.hs
aniketd: Taking a closer look, I realized `rewards` and `rdPairView` are slightly different. Do you mean `rewView` and `depositView`?

libs/cardano-ledger-core/testlib/Test/Cardano/Ledger/Core/Arbitrary.hs
aniketd: Noted!

libs/cardano-ledger-core/testlib/Test/Cardano/Ledger/Core/Arbitrary.hs
aniketd: Thanks for correcting me here again. :slightly_smiling_face:

libs/cardano-ledger-core/testlib/Test/Cardano/Ledger/Core/Arbitrary.hs
aniketd: Awesome! Thanks! :slightly_smiling_face:

libs/cardano-ledger-core/src/Cardano/Ledger/UMap.hs
aniketd: After thinking even further, do you mean I could rename `rewards` to `rdPairs` to reflect the function semantics?

libs/cardano-ledger-core/testlib/Test/Cardano/Ledger/Core/Arbitrary.hs
aniketd: Understood! The call to `elements` however, throws an exception that its argument can't be empty. This ends up agreeing with our intention to generate non-empty `UMap`s only, as the tests that use `genInvariant` indeed expect a non-empty `UMap`. So I have renamed it to `genInvariantNonEmpty`.

```haskell
genValidTriplesNonEmpty = scale (* 2) $ do
  Positive nCreds <- arbitrary
  nPtrs <- chooseInt (1, nCreds)
  creds :: [Credential 'Staking StandardCrypto] <- vectorOf nCreds arbitrary
  rdPairs :: [RDPair] <- vectorOf nCreds arbitrary
  delegs :: [KeyHash 'StakePool StandardCrypto] <- vectorOf nCreds arbitrary
  ptrs :: [Ptr] <- vectorOf nPtrs arbitrary
  pure
    ( Map.fromList $ zip creds rdPairs
    , Map.fromList $ zip creds delegs
    , Map.fromList $ zip ptrs creds
    )
```

aniketd:
@lehins This is ready for another review!

libs/cardano-ledger-core/CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Version history for `cardano-ledger-core`
 
+## 1.2.0.0
+
+- Deprecate `Cardano.Ledger.UMapCompact` in favor of `Cardano.Ledger.UMap` and add tests for it. #3371
+  - Add `Cardano.Ledger.UMap.rdPairView` to view the reward-deposits pair from the `UMap`.
+  - Rename `Cardano.Ledger.UMap.rewards` to `rdPairs`.
```
lehins: `rewards` function was never exported, so this change should not be added to the changelog
```suggestion
```

libs/cardano-ledger-core/src/Cardano/Ledger/UMap.hs
lehins: I also didn't realize until now that they are slightly different. However the name `rewards` was invalid, so renaming the function was a good call. Thank you for doing it.

lehins (approved):
Awesome work. Thank you.

libs/cardano-ledger-core/CHANGELOG.md
aniketd: Oh! Got it! Thanks!

lehins:
@aniketd Looks like there is a test failure

lehins (approved):
:+1:

PR #3375, opened 2023-04-08
Revert Tickf optimization

lehins:
# Description

This PR disables both of the optimizations implemented in #3209 Current implementation is causing problems. Reasons are not yet known and require investigation. That is why this PR only disables the optimizations instead of reverting them completely.

Important to note that both of the optimizations are causing problems, even when applied separately: 
* lazy computation of pool stake distribution
* and minimizing logic in TICKF rule 

That is why they are both being disabled.

Naturally, as all of my PRs go, this one also has a little bit of a cleanup included in it as a separate commit.

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/eras/shelley/impl/CHANGELOG.md
+++ b/eras/shelley/impl/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Version history for `cardano-ledger-shelley`
 
+## 1.1.1.0
+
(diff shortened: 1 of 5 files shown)


JaredCorduan (approved):
:+1:

PR #3377, opened 2023-04-13
Fix cabal warnings

aniketd:
This gets rid of the following warnings by cabal:
1. `missing default language`
2. `-threaded -*rts* does not apply to library component`


lehins (approved):
You are absolutely right

PR #3381, opened 2023-04-14
Remove upper bounds on all plutus dependencies

lehins:
# Description

These upper bounds only prevent us from upgrading to newer versions. Need to be removed.

# Checklist

- [x] Self-reviewed the diff


michaelpj:
I don't think this is a good idea (and it's against the [current recommendation of the CEH](https://input-output-hk.github.io/cardano-engineering-handbook/policy/haskell/packaging/versioning.html#speculative-upper-bounds) ). We were actually thinking of addressing this on our side by making you a PR every time we do a release to bump the major version pin. That way we can do "push-based" upgrading rather than "pull-based".

lehins:
@michaelpj There was already three releases of `plutus-ledger-api`, but we didn't see any PRs to ledger that updated the upper bounds. So, the current recommendation in the handbook is broken. Unless there will be somebody doing PRs to the repositories that depend on the package being released, upper bounds make no sense.

On the other hand. If there are no upper bounds, then a release engineer will not be able to make a release without fixing the upper bounds with revisions and submitting a PR or an issue to the source repo, but in only in case when there is a real problem building. 

So, I respectfully disagree with you on this issue

michaelpj:
> There was already three releases of plutus-ledger-api, but we didn't see any PRs to ledger that updated the upper bounds. So, the current recommendation in the handbook is broken.

I don't think that follows. It would be _better_ for us to push bumps to you, but what's wrong with you doing them when you need them? What's so broken? And I encourage you to engage with the CEH on the topic :)

> If there are no upper bounds, then a release engineer will not be able to make a release without fixing the upper bounds with revisions and submitting a PR or an issue to the source repo, but in only in case when there is a real problem building.

Ledger has numerous packages (10-ish). Suppose you release 3 or 4 versions of these with no upper bounds. Now we release a major version bump with a breaking change (it's a major version after all!). Now we need to do some 40 revisions if we wanted to actually fix things. That seems pretty bad.

michaelpj:
As I think we discussed before, I tend to agree that speculative upper bounds are generally not a good idea. I think that the Cardano packages are a special case: they break frequently; we have close contact with the people updating them; we're releasing a lot; it's just not the same as Hackage.

lehins:
> It would be better for us to push bumps to you

First of all, no one has ever done this. Call me pessimistic, I suspect no one will.

Moreover, I appreciate that you are willing to push updates to us. I am not willing to push updates to all of our downstream users.

> but what's wrong with you doing them when you need them?

Many things wrong. Unless we notice ourselves or someone tells us, we will never know that there is an update to our dependency. Thus not only we can't benefit from all of the bug fixes and improvements, we also can't make our code compatible with newer releases in timely manner, whenever there is an incompatibility. Same argument applies to the downstream users of ledger.

> What's so broken?

With this mantra of strict upper bounds every release to a CHaP will always succeed. This is great, but that is only true because none of the packages that depend on this new release are able to use it! I mean NONE of them! So, the only way to make a dependency of a package usable, for example, `cardano-ledger-binary` depending on `plutus-ledger-api`, is to go through all of the releases and make revisions. Naturally, no one is ever gonna do that, because there is no incentive. So the only other possible way for the a release of `plutus-ledger-api` to be used is if one will go through all of the downstream packages that use it and submit PRs with version bumps and then make release of those packages too. And I mean ALL of them, because if at least one of them still has an upper bound oris unreleased the package is unusable. Only then a new breaking release of a package like `plutus-ledger-api` will be available for use. That is the example in this PR, neither one of the last three plutus releases are being used, because of those upper bounds issue. And if you look in this PR, if it wasn't for upper bounds we would have noticed months ago that there is a breakage with `plutus-core-1.2`. It would have been fixed in a much more timely manner.

Going back to the maintenance burden. With your suggested push approach, it might be ok for plutus team, but whenever we release some or all of the dozen packages from ledger, we don't even have a mechanism of figuring out all of the downstream uses to submit PRs with fixed upper bounds. And certainly I don't have any desire to waste my time on doing this redundant work.

> Ledger has numerous packages (10-ish). Suppose you release 3 or 4 versions of these with no upper bounds. Now we release a major version bump with a breaking change (it's a major version after all!). Now we need to do some 40 revisions if we wanted to actually fix things. That seems pretty bad.

That is if they are indeed broken. Quite often major version bump will result in no breakage downstream whatsoever. A nice example of that is Plutus team releasing three breaking versions of plutus-ledger-api, which resulted in no breakage in ledger. So you wouldn't even have to do anything. On the other hand if there is a breakage in one of the dependencies, there is no need to go submit revisions for all old versions, just the latest one will be enough. That is because older versions will not be picked by cabal when constructing a build plan. Naturally, if really wanted to fix all older ones as well we could write a script that could handle all those 40 revisions for us. Note that this part could be 99% automated, all it would  take for a release engineer is to specify name of the package that needs to be fixed up. 

So, to answer your question, if we didn't use strict upper bounds, upon attempting to release 10ish packages, we would immediately identify all packages that break downstream (thanks to your efforts of adding CI to CHaP) and we would be able to make revisions for all those packages and submit issues with findings to the downstream users, but most importantly all of the downstream packages that didn't break, would be able to use the newest versions immediately.  With push approach, none of the downstream dependencies that were released prior to the 10ish releases from ledger would be able to use them, EVER!

> As I think we discussed before, I tend to agree that speculative upper bounds are generally not a good idea. I think that the Cardano packages are a special case: they break frequently; we have close contact with the people updating them; we're releasing a lot; it's just not the same as Hackage

In the last half a year we had some drastic changes in ledger code base that indeed caused a lot of breakage, but that should no longer be a norm going forward.

So, "they break frequently; " - this should not be the norm. For intra-repository packages, sure. Like in ledger, for example, packages will and should break often because they are very much intertwined with each other. Which is why I would 100% advocate on having strict bounds on packages that work very tightly together. However, for integration between Cardano components we should strike to not experience much breakage.

Let me give you a concrete example on why a push approach is horrible, especially for repos like ledger. Say we remove a function in `cardano-ledger-shelley`: that's a breaking change. Right away, due to such little change, none of the ledger packages are usable by any of the downstream components like consensus and node. However, most of the time a change like that will happen only because we need it in another era like `cardano-ledger-conway`. So, let me ask you. Why do I need to go into two separate repositories (consensus and node) and submit PRs for a change that has no affect whatsoever on those repositories? Mind you, I also need to find all of the packages that use `cardano-ledger-shelley`. What about all other users of ledger like hydra and stuff that I have no clue about. Do I need to go and update them too? That just makes no sense. 

On the other hand with pull approach, every package release, be it with major or minor version update, is immediately usable by all of the downstream packages, unless there is a breakage, which is spotted even before a release can happen, which is another great aspect on its own.

So, whoever wrote this part of CEH is utterly wrong:

> However, for any Cardano dependencies, a package MUST include an upper bound that excludes the next major version of the dependency pin, and SHOULD pin it to a single major version.
> ...
> Cardano packages themselves typically both a) make frequent breaking changes and b) have important behavioural differences between major versions.

So, if the person who wrote that section wants to do all the integration work, please be my guest. Me, having spent last month doing integrating cardano-base, cardano-ledger, ourboros-consensus and cardano-node, I just refuse to set upper bounds without knowing that package does not support it, unless it is an intra-repo dependency that is.

With respect to the rest of the page on versioning in CEH, I am in full agreement.

michaelpj:
(This has turned into a much wider discussion so I'd like to tag some people in who might have thoughts @zliu41 @abailly-iohk @coot @ch1bo @andreabedini. Maybe we need to have a call about it? )

> Unless we notice ourselves or someone tells us, we will never know that there is an update to our dependency. Thus not only we can't benefit from all of the bug fixes and improvements, we also can't make our code compatible with newer releases in timely manner, whenever there is an incompatibility. Same argument applies to the downstream users of ledger.

I think this is getting towards the heart of it. There are some underlying questions here: "who is responsible for driving integration work between upstream and downstream components?" and also "who cares?". I think there is no one-size-fits all answer here.

Let's consider some scenarios:
- When the Plutus team makes a significant change (say SOPs), at the moment we largely "own" the feature, and we are the ones who are keen to get it integrated into the node - that's what "done" means for us. So it makes sense for us to therefore take on the work of pushing it downstream along the path towards the node. I think the same applies to when ledger/network adds new features.
- However, when the Plutus team makes changes, that also affects various other downstream consumers of our libraries, e.g. Hydra. But there the motivation is reversed: they want to update to newer versions of our libraries in order to use the new features or to use the new node, so it makes sense for them to pull the updates.

So my question is, based on what you've said: why do _you_ care that you don't have Plutus bug fixes or you can't get ahead of incompatibility work? That's _our_ problem: if we don't solve those problems then _our_ features won't get released! That includes if there is work that we need your help to do: if we don't surface that soon enough, it's our fault.

Overall therefore I think whether a push- or pull-based workflow makes sense varies depending on the situation, but it seems to me that the "mainline" cardano-node dependency tree should be mostly push-based.

> So, let me ask you. Why do I need to go into two separate repositories (consensus and node) and submit PRs for a change that has no affect whatsoever on those repositories? Mind you, I also need to find all of the packages that use cardano-ledger-shelley. What about all other users of ledger like hydra and stuff that I have no clue about. Do I need to go and update them too? That just makes no sense.

Based on what I said above, there is a clear separation here:
- You _want_ to update consensus because that's what you will need to do in order for your team's changes to eventually get released into the node. Arguably updating the node is the consensus team's problem, but you probably at least want to pay attention to help with any problems, because it's on your critical path.
- You don't particularly _want_ to update Hydra. That's fine, _they_ want to update. So you can let that be pull-based.

> So, the only way to make a dependency of a package usable, for example, cardano-ledger-binary depending on plutus-ledger-api, is to go through all of the releases and make revisions. Naturally, no one is ever gonna do that, because there is no incentive. So the only other possible way for the a release of plutus-ledger-api to be used is if one will go through all of the downstream packages that use it and submit PRs with version bumps and then make release of those packages too.

This seems... correct to me? Don't forget that the recommendation we made was to have caret bounds, so you get lower bounds too. That means there's a fairly natural bubbling-up process that happens even if people only care about their direct downstream dependents:
- ledger updates to a new major version of plutus
- we try to update downstream packages to a newer major version of ledger... which _forces_ the newer version of plutus, so if there's a dependency that doesn't allow it, we have to fix it
- etc. until we reach the node

It does mean that these steps have to be _conscious_ and require actions, but that seems good to me? Although, again I think you should mostly only care for a limited set of downstream things that you care about pushing to.

> With push approach, none of the downstream dependencies that were released prior to the 10ish releases from ledger would be able to use them, EVER!

Again... why is this bad? They're old versions. Just release a new version? I think we're getting a little _too_ keen on revisions here :sweat_smile: 

> Quite often major version bump will result in no breakage downstream whatsoever. A nice example of that is Plutus team releasing three breaking versions of plutus-ledger-api, which resulted in no breakage in ledger.

The problem is that you are undermining us here. The reason the Plutus team are using this versioning strategy is because we don't want to guarantee that a release _won't_ break things at the moment, so we are conservatively releasing major versions. That way we won't accidentally release a breaking change that gets pulled by downstream automatically. If you start treating major versions as minor versions that all becomes useless and we might as well lie and release minor versions that sometimes break. What's the point of even having different kinds of version at that point?

> So, to answer your question, if we didn't use strict upper bounds, upon attempting to release 10ish packages, we would immediately identify all packages that break downstream

> most importantly all of the downstream packages that didn't break, would be able to use the newest versions immediately

I think these statements are off because of a crucial issue: not having an upper bound doesn't mean you _will_ use a newer version of the library, it just means that cabal _can_ pick it. You can't force the newer version to be used without a _lower_ bound, which requires editing the package and doing a release/revision. And if you don't do that, there's nothing that stops cabal from picking an earlier version and masking build issues until much later. (e.g. `plutus-ledger-api-X` breaks ledger, ledger has no upper bound so could theoretically be built with `plutus-ledger-api-X`, but for some reason the solver picks `plutus-ledger-api-(X-1)`, we put `plutus-ledger-api >= X` only at the node, _NOW_ we discover the build issue) (yes, the solver does seem to randomly just pick older versions if it feels like it, this happened a lot when I was trying to make cardano-node buildable from CHaP...)

So removing upper bounds doesn't prevent us from having to go up the chain making version tweaks, because we need to fix the lower bounds in order to be sure things aren't broken. At which point we might as well just have caret bounds and update those.

> So, whoever wrote this part of CEH is utterly wrong:

I based it on [the issue](https://github.com/input-output-hk/cardano-engineering-handbook/issues/26) where we discussed these topics but I guess you didn't end up looking at [the PR](https://github.com/input-output-hk/cardano-engineering-handbook/pull/37), that's a shame. The point I want to make about the CEH is that 1) it's a living document, and 2) I really want it to reflect sensible policy. If what's there doesn't make sense we should delete it, but ideally we should replace it with some sensible reasoned policy that people can follow. As I think is clear from this discussion it's not obvious what to do, and distilling that into a sane policy that everyone can follow is a huge help for the organization, versus anarchy!

------------

Here's my attempt at a synthesizing a policy that might make sense:

(to go in "release best practices")

Projects SHOULD identify a key set of downstream dependencies which are needed in order for the project's deliverables to be completed. Projects SHOULD proactively push dependency bumps to these downstream projects whenever they make a release.

Example: `plutus` is a dependency of the Cardano node, and work in `plutus` cannot be used until the node is released. `plutus` should therefore identify the downstream dependencies between it and the node, and proactively push dependency updates to them.

Projects do not need to proactively push updates to any other downstream projects. It is sufficient to make sure that release do not break these, which should be handled by following the versioning policy and the CHaP CI.

zliu41:
I'd prefer keeping the upper bounds, simply because it guarantees that a specific plutus minor version (x.y) is used in a specific node version (assuming cardano-node also uses upper bounds for cardano-ledger), and it is easy to tell which version it is.

Without the upper bounds, could it be the case that two node instances, with the same node version, built at different times and on different machines, depend on two different plutus versions? Probably not but I don't know for sure.

And this may also complicate the node release process. Say we have a node release candidate, which depends on plutus-core-1.5.0 and another package X-1.6.0. During node QA, an issue is found in X-1.6.0, and we need to update the index state in order to pull in X-1.6.1. But if there's no upper-bound on plutus-core, it may now be pulling in plutus-core-1.8.0. This may break things, and even if it doesn't, it still isn't supposed to happen, because when preparing a release candidate for a release, we are not supposed to pull in any unrelated changes other than bug fixes.

> First of all, no one has ever done this. Call me pessimistic, I suspect no one will.

I recently talked to @JaredCorduan about this. We'd likely to start doing this and see how it goes. I just made [this PR](https://github.com/input-output-hk/cardano-ledger/pull/3382) today; it's pretty effortless. By the way, this PR will conflict with https://github.com/input-output-hk/cardano-ledger/pull/3365 which bumps plutus-core to 1.5.

lehins:
> So my question is, based on what you've said: why do you care that you don't have Plutus bug fixes or you can't get ahead of incompatibility work? That's our problem: if we don't solve those problems then our features won't get released! That includes if there is work that we need your help to do: if we don't surface that soon enough, it's our fault.

You are looking at this problem from the perspective of Plutus team, which is slightly different from other packages. Let's forget ledger and plutus for now and think of a package from `cardano-base`, something like `cardano-strict-containers` for instance. Let's say someone identified and fixed a bug in that package, but for some reason it required a major version bump. Despite that it was a breaking change it was a change that was very unlikely do break things downstream. Be it as it may, it was still a major version bump. I understand that bug fixes usually are released with minor version bump, but what if it was something like a removal of dangerous type class instance, that was previously thought to be safe. This means that it requires fixing of upper bounds in all of the packages downstream.  This, of course, is a hypothetical example, but not an unrealistic one. So, who is responsible for fixing the bounds in such scenario? Is this is a responsibility of `cardano-base` maintainer or a person who fixed the bug to do this propagation downstream? IMHO neither. It should be individual maintainers of downstream components who should take care of this compatibility, not the maintainer of `cardano-strict-containers`. However, status quo today is that until **all** of the downstream uses of `cardano-strict-containers` have fixed their bounds, this bugfix will not propagate all the way to `cardano-node`

Believe it or not, but similar scenario does apply to ledger codebase as well. It is quite rare that someone from the ledger team does the integration of new features downstream. I guess this is where the confusion about the intentions comes from.

> When the Plutus team makes a significant change (say SOPs), at the moment we largely "own" the feature, and we are the ones who are keen to get it integrated into the node - that's what "done" means for us.

Significant changes yes, but in ledger we can have many insignificant changes cause major version bumps, that we would like to propagate. What I am trying to say is that it is not the same for ledger as it is for Plutus. Moreover, nowadays it is up to a release engineer on duty, which is rotated from different teams, to do the integration and release work, so it is also not necessarily a ledger team member that will be doing releases of ledger packages either. Luckily, our release process allows for any of the packages to be released from master at any time, so release engineer can just make a release from master and tackle all of the integration downstream. What if current downstream component of ledger did not cause any breakage? And for the sake of example, only `cardano-api`  had changes since the last time ledger was released. If we use strict upper bounds then all of the packages from consensus and node must be released again with upper bounds fixed up. This is a huge waste of time.

> Again... why is this bad? They're old versions. Just release a new version?  

They are not necessarily old versions, they can be current version. The only thing preventing them being used is the upper bounds on there dependency! That is my biggest beef with this upper bound nonsense. Right now, if I change a number of arguments in function in `cardano-ledger-shelley` that is only used in `cardano-ledger-conway` and I want to make a release of those two packages, because this functionality is important downstream. Without upper bounds downstream all I need to do is just make a release and the next time those downstream components update their index states they will get this functionality. However, with strict upper bounds, I need to:

* make two PRs:
  * one into `ouroboros-network` that updates 18 occurances of bounds
  * one into cardano-node that updates 11 occurrences of bounds
* Release a dozen packages from `ouroboros-network`

Only then I can use that functionality.

What happens if `ouroboros-network` diverged and now includes major breaking changes or even worse it in a state that can't be released to CHaPs? Ok, then I can solve it with a revision to all of those dozen packages `ouroboros-network`. 

Avoiding upper bounds gives us this functionality for free, this is why it is that bad.

> I think we're getting a little too keen on revisions here

Without revisions we'd be forced to use strict upper bounds and we would not be having this conversation. With pull approach that I am suggesting revisions allow us to fix a dependency, but only if it is indeed broken. It is easy to see when something is broken, because then a release is not possible. With strict upper bounds, revisions can be used to make something that is unusable, but not broken, usable again. It is not easy to see if something not broken, because you need to make a revision in order to find that out. So, if we have a tool, we should use it to the full potential.

> why do you care that you don't have Plutus bug fixes or you can't get ahead of incompatibility work? That's our problem: if we don't solve those problems then our features won't get released! 

You are right, we don't care about plutus fixes that much until we are forced to support a new language version. It would be nice to spot things that aren't compatible sooner, but this is not a huge deal. Precisely because there very little affect plutus changes have on ledger, consensus and node. 

> You want to update consensus because that's what you will need to do in order for your team's changes to eventually get released into the node. Arguably updating the node is the consensus team's problem, but you probably at least want to pay attention to help with any problems, because it's on your critical path.

Depending on the change, yes. Sometimes we will have changes that will require integration with consensus, in which case we will work closely with consensus and node to do the integration. Something like introduction of a new era. Until very recently this was exactly the process. We only started on integration when major features like that were introduced, which caused the work to only start happening when people needed those new features downstream or we as ledger wanted to propagate them downstream. Because we release new major features like a new era as often as once a year or so, the integration of new packages didn't start until downstream components started to worry about the new era, which would be on the order of months. But we constantly have other improvements that should be propagated, not only when major features are ready. That is one of the reasons why we have such huge intervals between proper integration. To make this example more concrete we barely merged my integration work last Friday and before that it was all the way in October. This is just horrible. There is currently an effort on having integration done and having a cardano-node release every two weeks. What I am alluding to is that all the upper bounds will really get in the way of this effort and will result in more work then necessary and more time wasted on those stupid bounds and redundant releases.

> You don't particularly want to update Hydra. That's fine, they want to update. So you can let that be pull-based.

They can't be pull based if other packages are push based, because they also depend on consensus and cardano-api, which forces them to be push base too. That's why I am trying to say, this pull based thing screws everyone up and prevents timely upgradeablility. 

> That means there's a fairly natural bubbling-up process that happens even if people only care about their direct downstream dependents

I understand, that this can work well with respect to plutus. If this is what you guys want and this is what you are happy with then that's fine. Especially if you promise to send us PRs or let us know in some other form when you have a major releases pushed to CHaPs then that's fine by us too and we can continue using caret bounds for Plutus. But, you better keep your promise :wink: cause latest plutus didn't get integrated into cardano-node last Friday precisely because of those upper bounds. 

What I am trying to say, is that what you think works for Plutus will probably not work for ledger and probably for other components as well. More often than not we have changes that will be internal to ledger, but because we have intra-era packages that depend on each other those changes will result in major version bumps. So this natural bubbling-up is not feasible for ledger.

> Overall therefore I think whether a push- or pull-based workflow makes sense varies depending on the situation, but it seems to me that the "mainline" cardano-node dependency tree should be mostly push-based.

This is precisely where I want to disagree with you. Push based approach will lead to more reliable CHaP, can't argue with that, but pull based will lead to much faster integration and development. I know you care more about the former, because of your involvement in it. Thank you by the way, it seems to be working out great. However, in my opinion, we are really suffering from the latter and we should utilize the power of CHaP to the fullest and promote breakage of packages upon a release attempt in order to speed up the integration process. Integration should be as frictionless as possible, but 

> I think these statements are off because of a crucial issue: not having an upper bound doesn't mean you will use a newer version of the library, it just means that cabal can pick it. You can't force the newer version to be used without a lower bound, which requires editing the package and doing a release/revision. And if you don't do that, there's nothing that stops cabal from picking an earlier version and masking build issues until much later.

This is where you are wrong. Cabal dependency solver will always use the latest version, unless there is an upper bound that prevents it from doing so. 

> yes, the solver does seem to randomly just pick older versions if it feels like it, this happened a lot when I was trying to make cardano-node buildable from CHaP...

This is why it feels random, it is precisely because of the upper bounds! It is hard to figure out where the upper bound is coming from. It is often some transitive package that is at fault, commonly even more than one. In my experience figuring out build failures is actually much easier then figuring out the dependency conflict. To make matters worse, this bug with sublibraries and nix, takes it to a totally new level of incomprehensiveness. I wasted three hours of Friday trying to figure it out. In that case it was my fault of not doing `cabal update`, but I've ran into the same problem before where upper bounds where actually at fault.

> Projects SHOULD identify a key set of downstream dependencies which are needed in order for the project's deliverables to be completed. Projects SHOULD proactively push dependency bumps to these downstream projects whenever they make a release.

IMHO, this is not gonna happen. This requires at least one person who is responsible for this on each team and who is intimately familiar with **all** of the downstream dependencies to be able to push such updates downstream. Having just done all of that,  I think it is good idea to do it sometimes, because it gives some exposure to the downstream components, but making a team be responsible for integration downstream always, is a bad idea.

lehins:
@zliu41 I appreciate your input. I can't leave it without a response :slightly_smiling_face: 

> I'd prefer keeping the upper bounds, simply because it guarantees that a specific plutus minor version (x.y) is used in a specific node version (assuming cardano-node also uses upper bounds for cardano-ledger), and it is easy to tell which version it is.

It is easy to see all dependencies with `cabal freeze`, there is no need to rely on upper or lower bounds for that. We want cardano-node using the latest version of plutus as soon as possible. If latest version can't be used for whatever reason, then the upper bound should be in place. Not the other way around. If plutus version is unusable for some reason, then it probably should not be released to CHaPs either, but if it is up there, then there should be not be any artificial guards like upper bounds preventing cardano-node from using it.

> Without the upper bounds, could it be the case that two node instances, with the same node version, built at different times and on different machines, depend on two different plutus versions? Probably not but I don't know for sure.

This is not possible because index-state ensures the exact versions of dependencies. Naturally different OSes (win vs linux) will have different dependencies, but they will be exactly the same, regardless if you build it in the morning or in the evening or a year later, as long as the index-state is the same.

> And this may also complicate the node release process. Say we have a node release candidate, which depends on plutus-core-1.5.0 and another package X-1.6.0. During node QA, an issue is found in X-1.6.0, and we need to update the index state in order to pull in X-1.6.1. But if there's no upper-bound on plutus-core, it may now be pulling in plutus-core-1.8.0. This may break things, and even if it doesn't, it still isn't supposed to happen, because when preparing a release candidate for a release, we are not supposed to pull in any unrelated changes other than bug fixes.

We have moved away from release candidates and we will be tagging releases from master every two weeks. The solution of upper bounds you describe actually becomes a problem, instead of a solution. Because we want to be able to get newest  dependencies  of node as easy as possible and upper bounds stand in the  of that.

> We'd likely to start doing this and see how it goes.

That's good that you wanna start doing that. I am always happy to see PRs from other teams.

What I don't want is our team to start doing that! That is my main problem. Let me show you why. This is Plutus dependency footprint:

```shell
~/iohk/cardano-ledger> git grep -e plutus-ledger-api --or -e plutus-core -- *.cabal | wc -l
15
~/iohk/ouroboros-network> git grep -e plutus-ledger-api --or -e plutus-core -- *.cabal | wc -l
0
~/iohk/cardano-node> git grep -e plutus-ledger-api --or -e plutus-core -- *.cabal | wc -l
3
```

And this is for ledger:
```shell
~/iohk/ouroboros-network> git grep -e cardano-ledger --or -e cardano-protocol-tpraos --or -e cardano-data --or -e small-steps --or -e vector-map -- *.cabal | wc -l
83
~/iohk/cardano-node> git grep -e cardano-ledger --or -e cardano-protocol-tpraos --or -e cardano-data --or -e small-steps --or -e vector-map -- *.cabal | wc -l
54
```

I don't even wanna talk about `cardano-base`, that I have to maintain mostly by myself.

zliu41:
> This is not possible because index-state ensures the exact versions of dependencies.

I'm not super familiar with the details of how cabal works but is it possible for cabal to pick up an older version of a dependency if one hasn't run `cabal update` for a while? Or does `index-state` 100% guarantee the exact versions?

> We have moved away from release candidates

This is news to me. It must have happened recently, as the latest node release (1.35.7) still used release candidates. I wish cardano-node had a documented release process.

I'm suspicious of the idea that cardano-node would want to pull in a new major version of plutus-core, simply because a bug in an unrelated library needs to be fixed. What if the new major version of plutus-core does have non-trivial breaking changes, breaking a number of libraries? Then you are forced to fix everything and re-release every broken package, which could delay the release.

Breaking changes are not _that_ uncommon and this goes against the practice that downstream packages should be able to upgrade the dependencies and fix the breakages at their own paces.

> What I don't want is our team to start doing that!

Again the question is can we 100% rely on index-state? Let's say you made a critical bug fix in cardano-ledger-alonzo-1.2, which must get into the next node release. If all the downstream packages have `cardano-ledger-alonzo >= 1.1`, would you still be sure that the next node release will indeed depend on 1.2, as long as the index-state is set correctly? And if so, why even use lower bounds, can we remove all bounds?

lehins:
> if one hasn't run cabal update for a while?

There is an implicit assumption here that cabal is made aware of latest index-state for CHaPs and hackage for that matter. Other nasty problems come to live when cabal is used with nix when cabal state is out of sync with nix. For those that don't use nix, if you don't do `cabal update` then you don't get the newest versions, so upper bounds are irrelevant for this problem.

> This is news to me. It must have happened recently

Yes, this is very new and will start with the next `8.0.0` release. Versioning scheme is changing too. Initial number will always indicate the major protocol version that cardano-node can handle

The plan is to make sure we have a release ready to go from master at any time. Some releases will just be marked as `-pred` if they are not ready to be released to the public. This also means that integration should happen as often as possible, the plan is two weeks intervals. This whole plan is also very new and I was not the one who came up with it. I am just relying what I've learned from Sam.

 > I wish cardano-node had a documented release process.

I hope it will be eventually. 

>  Then you are forced to fix everything and re-release every broken package, which could delay the release.

Nope, that is exactly when you should use the upper bound. Because you know that newer dependency is incompatible, and the plans to fix this compatibility are postponed to a later release, because it will take a lot of work. But this is a conscious decision, rather than a speculative one. Usually it is the opposite, fixing compatibility is very easy when it is done often. In which case it is better to fix compatibility with newer plutus-core and make a release with latest version of plutus.

> Breaking changes are not that uncommon and this goes against the practice that downstream packages should be able to upgrade the dependencies and fix the breakages at their own paces.

I am not saying to ban upper bounds, I am saying to not use speculative upper bounds. What you are describing is a known incompatibility which can be either temporarily solved with an upper bound (even for released packages with a revision) or with fixing the breakage. The pace is still decided by the downstream packages.

> Again the question is can we 100% rely on index-state? 

Yes. Because we use nix. Either using `niv` or `flake` (depending on the repo) all dependencies are set in stone in the repository together with the index-state update.

> Let's say you made a critical bug fix in cardano-ledger-alonzo-1.2, which must get into the next node release. If all the downstream packages have cardano-ledger-alonzo >= 1.1, would you still be sure that the next node release will indeed depend on 1.2, as long as the index-state is set correctly? 

Yes. 100%. Bounds in cabal file should be only used to figure out compatibility, not to predict which version will be used in a release. The good thing about setting `cardano-ledger-alonzo >= 1.1` in all dependencies is that all I need to make `cardano-ledger-alonzo-1.2`  usable is release it to CHaP and nothing else. If there is, for example, an incompatibility with consensus, I will not be able to make a release without:
* either fixing bound in consensus with a revision, in which case we record known incompatibility with upper bound, not a speculative one. Submitting an issue or a PR on consensus repo would be in order to notify about the incompatibility.
* or fix compatibility and release consensus together with  `cardano-ledger-alonzo-1.2`

 of `cardano-ledger-alonzo-1.2` and no other changes, because if it is incompatible 


> And if so, why even use lower bounds, can we remove all bounds?

Bounds describe compatibility of packages in a project for cabal when it does the dependency solving.  So, when we set the lower bound, we indicate that we know that the package is indeed incompatible with this dependency. Same goes for upper bounds. Thing about lower bounds is that we can always check ahead of time if it is true or not. While with speculative upper bounds we are trying to predict the future in a way that is too strict. This is sometimes fine if you are closely monitoring your dependencies and making revisions of upper bounds when new versions are released. For example that is the process for all of the core packages in Haskell. However, I can't see this process being adopted with CHaP and Cardano. The project is a lot less stable and much faster moving where major versions are updated very often. Also because we utilze index-state we do not need to worry about older packages that were released that aren't used to be incompatible, which is not true for hackage. 
For example, speculative upper bounds are evil when you are dealing with stackage. I am convinced that similar ideology applies to CHaP as well.

michaelpj:
On the specific topic of this PR: it sounds like you're happy with leaving in upper bounds on the plutus packages so long as we push updates to you. We're happy to do that, so please do leave the bounds in. 

------

> This is where you are wrong. Cabal dependency solver will always use the latest version, unless there is an upper bound that prevents it from doing so.

Suppose that `plutus-core-2` adds a `hashable > 2` lower bound but `cardano-ledger` has a `hashable < 2` upper bound. Then cabal will silently pick `plutus-core-1` forever and you won't know.

If I want to make sure that `plutus-core-X` is being used by downstream then I _must_ propagate lower bounds. Otherwise any change to (unrelated) dependency bounds could indirectly lead to an old version being used.

This can also happen very easily if a Cardano package _does_ have a breaking change, and some packages therefore need bounds to ensure that they do/don't use it. For example, when I was working on making `cardano-node` buildable from CHaP, many packages needed strict upper or lower bounds on `cardano-prelude` to deal with the `slotting` split. But that meant I got various surprising plan choices. The same could happen today if we released, say, a breaking change in `cardano-crypto-class-5`: `plutus-2` might adapt and have `cardano-crypto-class >=5`, ledger might not adapt yet and have `cardano-crypto-class < 5`, and then ledger will _silently never use plutus-2_.

> What happens if ouroboros-network diverged and now includes major breaking changes or even worse it in a state that can't be released to CHaPs? Ok, then I can solve it with a revision to all of those dozen packages ouroboros-network.

This sounds like a good time for revisions, yeah. You've released a major version because you have a potentially breaking change, and you need to get it through network but network is in a bad state. _If_ we are in a place where everyone is releasing frequently (which I think we should aim for), then this should be an exceptional circumstance, and doing some revisions seems like a good way of getting past it.

I think a lot of our disagreement comes about whether we think this situation is abnormal or normal.

> cause latest plutus didn't get integrated into cardano-node last Friday precisely because of those upper bounds.

Why do you think we're not fine with that? We didn't take the steps necessary to integrate it downstream, so it didn't get into newer versions of downstream. That's what we expect and it's on us.

> This is Plutus dependency footprint... And this is for ledger...

Trying to make the problem super explicit here: you don't want to do push-based dependency updates of ledger because there are too many bounds to be updated?

This seems solvable with some automation. We have a script in `plutus` to bump all our internal version bounds when we do a release. I don't see why we couldn't write an `update-chap-deps.sh` script that does something like: 
1. For all packages in CHaP (or a listed subset)
2. Find the latest released major version
3. Rewrite all bounds on that package to pin the latest major version instead

Would that make this less painful for you?

> There is currently an effort on having integration done and having a cardano-node release every two weeks. What I am alluding to is that all the upper bounds will really get in the way of this effort and will result in more work then necessary and more time wasted on those stupid bounds and redundant releases.

So I am 1000% on board with doing integration faster and more frequently. I don't see why what I'm proposing will get in the way of that. If ledger/networking/consensus all release every two weeks _and bump their upstream dependency bounds_ then updates will promptly and reliably work their way up to the top.

Having written that and read what you've said, I can see one source of problems: if you have a more distant downstream dependent, and the packages in between don't update, then your integration work with the distant dependant can come up much later and not be prompt. e.g. if network doesn't release and/or try to push that to the node, then the new ledger changes also won't get in. I cant definitely see that's a problem, and I don't have a perfect solution to it, although I think "everyone releases frequently" is a pretty decent one.

> [Hydra] can't be pull based if other packages are push based, because they also depend on consensus and cardano-api, which forces them to be push base too.

I don't understand this at all (or maybe you didn't understand me?). I'm not proposing all-or-nothing push-based model, just pushing to the dependants that are critical for your stuff getting released. So if we have a new release of `consensus` the consensus team would push that to the node, but do nothing for Hydra. When Hydra _wants_ to update to a new node release, then they also need to pull in newer consensus etc (cabal will force them to). So that's perfectly pull-based.

> either fixing bound in consensus with a revision, in which case we record known incompatibility with upper bound, not a speculative one. Submitting an issue or a PR on consensus repo would be in order to notify about the incompatibility.

You should definitely do a PR. Otherwise you're forcing work on the other team when they update CHaP, which is pretty rude (and perhaps they will start putting in upper bounds themselves if you do it too much :p ).

> This requires at least one person who is responsible for this on each team and who is intimately familiar with all of the downstream dependencies to be able to push such updates downstream.

That's not what I said! I said you should identify _some_ downstream dependencies that you care about. That could be the empty set! For `plutus` it will probably just be the ledger and the node.

For example, I think it's quite plausible that for `cardano-base` we should _not_ push dependency bumps downstream. Again, let's consider who cares: usually it's downstream who cares about getting the bug fix. So let them pull it in.

Also, just because the upstream team is pushing the work forwards, doesn't mean they have to be able to do it all. Hopefully you have a good relationship with your downstream, and they are on board with frequent, small integration pushes, and so they will be happy to help with such PRs. (Lots of what I'm saying is assuming that people are on board with this kind of frequent integration, which I think they generally are.)

-----

Let me propose something that cuts the knot in a different way.

Suppose we had an `update-chap-deps.sh` script as described above. Now suppose we write a GHA that runs this on people's repositories nightly. That would mean that when someone does a release, their downstreams will _automatically_ get a bump PR made. If, as you claim, most bumps are safe, then these will trivially go green and can be merged with little attention. If the bump is not safe, well, don't merge the PR. (aka dependabot for CHaP)

That seems to me to maybe get the best of both worlds. Even for `cardano-base` packages, you could get updates propagated with very little effort from the `cardano-base` maintainers.

lehins:
> so please do leave the bounds in.

:+1: 

> Suppose that plutus-core-2 adds a hashable > 2 lower bound but cardano-ledger has a hashable < 2 upper bound. 

That is exactly what I mean. It is because of an upper bound!!! So it is not random. 

> Then cabal will silently pick plutus-core-1 forever and you won't know.

That is why you should use lower bounds, to prevent this from happening, if this is indeed an undesired behavior. Most of the time, however, this is exactly what we want, we want an older package version to be chosen, because it works. This is exactly what I want to happen too and that is what I am trying to convey to you! We want an older plutus packages to be picked if a newer ones can't be used. If there is a serious need to enforce the new version of plutus to be used, because of some bug or whatever, then a lower bound should be updated either in ledger or in node for plutus, but that is ONLY if we want to ensure that the latest version is being used. 

> If I want to make sure that plutus-core-X is being used by downstream then I must propagate lower bounds

Correct. Propagating lower bound instead of a caret. That is exactly what I am suggesting. 

> ledger might not adapt yet and have cardano-crypto-class < 5, and then ledger will silently never use plutus-2.

Again, from perspective of ledger this is the desired behavior. Because if ledger is incompatible with latest plutus (because of this incompatibility on common dependency `cardano-crypto-class`), then then node is not compatible with latest plutus either. That's the way it suppose to be.

> I think a lot of our disagreement comes about whether we think this situation is abnormal or normal.

Not only. The disagreement is in who is responsible for doing the downstream integration and releases (with upper bounds fixes). You are saying for Plutus it is suppose to be Plutus team that does the integration work. That is your prerogative. What I am saying for ledger this is not the case. There is a new policy that there is a rotating release engineer that does this integration.

> Why do you think we're not fine with that?

What's the point of a release to CHaP then if no one is using it? You also said that you want integration to happen as often as possible.

> I don't see why we couldn't write an update-chap-deps.sh script that does  ... Would that make this less painful for you?

No! It still requires you knowing all of the downstream dependencies.  Still requires making releases for all of the downstream dependencies, despite that in some cases a release might not be possible from `master`. It does not address all of the already released packages, those are the packages I want to continue working, not the new releases!

So, this suggestion does not solve much.

> That's not what I said! I said you should identify some downstream dependencies that you care about. 

I personally care about all of the packages in consensus and node and that is a lot of dependencies! But that is not my responsibility to go and update them! That's my point that I am trying to make across. 

> usually it's downstream who cares about getting the bug fix.

Precisely my point! And downstream wants to get this fix pulled in automatically.

> So let them pull it in.

They won't pull it in, because they don't know that there was a bug and that there was a new release. So the fixed version will not be used. It is not possible to mix the pull and push approaches, so I am even surprised you are saying: "let them pull it in."

> Suppose we had an update-chap-deps.sh script as described above.

This will not work, because such script could only potentially update the bounds, but it can't make releases and it can't adjust bounds for already released packages with revisions. Part with revisions would be completely impossible with strict upper bounds, because you can't know which bounds are speculative and which ones are intentional.

> That seems to me to maybe get the best of both worlds. Even for cardano-base packages, you could get updates propagated with very little effort from the cardano-base maintainers.

I am not doing this, because this is redundant work and it is not the job as a maintainer of a package to go and update bounds downstream. If you believe this should be the case feel free to do so.

@michaelpj I think this is going nowhere, so we can stop this discussion here. I am not gonna be setting any speculative upper bounds either because this is just dumb. At the same time it will be a while before I have to deal with this again, since I was just on a job of a rotating release engineer. Remember, it is not a responsibility of a ledger engineer to deal with downstream uses of ledger. This is the part that you are missing. Our job is to integrate new Plutus versions, because that is what we depend on. But since Plutus team is volunteering to update bounds for us then this solves that problem for us. I am not gonna worry about this anymore.

Closing in favor of #3382

michaelpj:
I didn't feel like this was going nowhere, I actually felt like we were figuring out quite a lot of interesting stuff in the problem space, but I definitely appreciate your desire to not keep going on about it :D 

I'll make a PR to weaken the recommendation in the CEH since we clearly don't have consensus on what to do. I'll tag you in that if you don't mind since I do want it to reflect something that we all agree on, even if it's weak :)

lehins:
@michaelpj Sounds good. Thank you.

PR #3385, opened 2023-04-19
Haddocks on Github Pages

teodanciu:
# Description

This PR introduces a new github workflow, triggered on master, which: 
- builds haddock for every package, aggregating them as a site (listed on the landing page)
- deploys this site to github pages, so the haddocks will be accessible here, after the PR is merged: https://input-output-hk.github.io/cardano-ledger  (replacing the current content, which seems obsolete anyway).  

The links between the packages are working well (for example: when inspecting rules, clicking on anything from small-steps or cardano-core will take you there). 
I'm unsatisfied for not being able to link to hackage the packages for which hackage documentation exists (like base).  I was able to do this only at the cost of the links between packages not working, but I thought it's more important to have the latter working. 

This is a copy of the site with this PR on my own fork, to demonstrate how the pages will look like: https://bienpulenta.github.io/cardano-ledger/

 
# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/.github/workflows/gh-pages.yml
+++ b/.github/workflows/gh-pages.yml
@@ -1,0 +1,40 @@
+name: Github Pages
+
+on:
+  push:
+    branches: [ "master" ]
+  workflow_dispatch:
+
+jobs:
+  gh-pages:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+
+    steps:
+      - uses: actions/checkout@v3
+      - uses: cachix/install-nix-action@v18
+        with:
+          nix_path: nixpkgs=channel:nixos-unstable
+          extra_nix_config: |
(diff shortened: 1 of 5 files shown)


scripts/prolog
lehins, line 1: I suspect this was added accidentally, since that file is generated by the `mkprolog.sh` script

.github/workflows/gh-pages.yml
```diff
@@ -0,0 +1,40 @@
...
+            substituters = https://cache.iog.io https://cache.nixos.org/ https://cache.zw3rk.com
+      - name: Build projects and haddocks
+        run: nix-shell --command "cabal update && cabal build --enable-documentation all && ./scripts/haddocks.sh"
+      - name: Add files
+        run: |
+              git config --local user.name ${{ github.actor }}
+              git config --local user.email "${{ github.actor }}@users.noreply.github.com"
+              cp -r ./haddocks/* ./
+              rm -rf haddocks
+              git add -A
```
lehins: Just in case `*.html` gets added to `.gitignore` in this repo as well, which is likely to happen sooner or later.
```suggestion
              git add -A --force
```

lehins (approved):
This is really awesome! Thank you.

It might be a good idea to add a link in the `README.md` to the landing page with haddocks.

JaredCorduan (approved):
:heart: :heart: :heart:

scripts/prolog
teodanciu: Indeed, thanks!

PR #3386, opened 2023-04-19
MapMember and NotMapMember

aniketd:
# Description

Implementation of `MapMember` and `NotMapMember` predicates.

# Checklist

- [x] Self-reviewed the diff


libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/Ast.hs
```diff
@@ -483,6 +491,16 @@ runPred env (NotMember x y) = do
   x2 <- runTerm env x
   y2 <- runTerm env y
   pure (Set.notMember x2 y2)
+runPred env (MapMember x y z) = do
+  x2 <- runTerm env x
+  y2 <- runTerm env y
+  z2 <- runTerm env z
+  pure $ Set.member x2 (Map.keysSet z2) && Set.member y2 (Set.fromList (Map.elems z2))
```
Soupstraw: I initially assumed, that this predicate tests whether the map maps `k` to `v`, but here it seems it checks separately that `k` is _a key_ in the map and `v` is _an element_ of the map. I'm not sure if this is intentional.

Soupstraw: So the key `k` does not necessarily map to element `v` in this case.

Soupstraw: Although, it seems to be consistent with the rest of the code, so I assume this is how it's meant to be.

aniketd: Yes @Soupstraw :slightly_smiling_face: that's correct: it is a predicate to check if a given key and value are in or are not in a given map.

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/Ast.hs
```diff
@@ -489,6 +497,16 @@ runPred env (NotMember x y) = do
...
```
TimSheard: something like this
case Map.lookup x2 z2 of
  Nothing -> False
  Just w -> w== y2

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/Ast.hs
```diff
@@ -489,6 +497,16 @@ runPred env (NotMember x y) = do
...
+runPred env (NotMapMember x y z) = do
+  x2 <- runTerm env x
+  y2 <- runTerm env y
+  z2 <- runTerm env z
+  pure $ not (Set.member x2 (Map.keysSet z2)) || not (Set.member y2 (Set.fromList (Map.elems z2)))
```
TimSheard: something like this
case Map.lookup x2 z2 of
  Nothing -> True
  Just w -> w /= y2

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/Rewrite.hs
```diff
@@ -327,6 +327,8 @@ accumdep info answer c = case c of
   -- be too late to overide the projections with the values of the vars in 'cs'
   Member t cs -> mkDeps (vars t) (vars cs) answer
   NotMember t cs -> mkDeps (vars t) (vars cs) answer
+  MapMember k v m -> mkDeps (vars k) (vars v) (mkDeps (vars v) (vars m) answer)
+  NotMapMember k v m -> mkDeps (vars k) (vars v) (mkDeps (vars v) (vars m) answer)
```
TimSheard: Experience might show we need to change this. So keep an ope mind

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/Solver.hs
```diff
@@ -373,6 +373,18 @@ solveMap v1@(V _ r@(MapR dom rng) _) predicate = explain msg $ case predicate of
...
+    Refl <- sameRep rng rng'
+    k <- simplify exprKey
+    v <- simplify exprVal
+    mapSpec (SzLeast 1) (relSuperset dom (Set.singleton k)) (RngRel (relSuperset rng (Set.singleton v)))
+  (NotMapMember exprKey exprVal (Var v2@(V _ (MapR dom' rng') _))) | Name v1 == Name v2 -> do
+    Refl <- sameRep dom dom'
+    Refl <- sameRep rng rng'
+    k <- simplify exprKey
+    v <- simplify exprVal
+    mapSpec SzAny (relDisjoint dom (Set.singleton k)) (RngRel (relDisjoint rng (Set.singleton v)))
```
TimSheard: I am not sure this works.  We want (k,v) to be in the same pair.  I don't think this spec means that.

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/Solver.hs
```diff
@@ -710,16 +722,36 @@ dispatch v1@(V nam r1 _) preds = explain ("Solving for variable " ++ nam ++ show
...
   [NotMember (Var v2@(V _ r2 _)) expr] | Name v1 == Name v2 -> do
     Refl <- sameRep r1 r2
     set <- simplify expr
     let msgs = ("Solving for variable " ++ nam) : map show preds
     pure $ suchThatErr msgs (genRep r2) (`Set.notMember` set)
+  [MapMember (Var v2@(V _ r2 _)) _exprVal exprMap] | Name v1 == Name v2 -> do
+    Refl <- sameRep r1 r2
+    m <- simplify exprMap
+    let msgs = ("Solving for variable " ++ nam) : map show preds
+    pure (fst <$> itemFromSet msgs (Map.keysSet m))
```
TimSheard: Don't we want  _exprVal to be Map.Lookup k?

TimSheard (commented):
A few issues to deal with  we want the pair (k,v) to be in Map.toList m

TimSheard (commented):
we need to fix the solving for MapMember when the map is known. The answer then are then some sort of lookup.

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/Solver.hs
TimSheard: done

TimSheard (approved):

PR #3387, opened 2023-04-19
Change `DCert` into a type family

Soupstraw:
# Description

This PR adds the `EraDCert` typeclass and changes `DCert` into a type family. There are also typeclasses `ShelleyEraDCert` and `ConwayEraDCert` that add patterns for certificates added in those eras.

# Checklist

- [x] Self-reviewed the diff


eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxInfo.hs
```diff
@@ -522,8 +549,8 @@ alonzoTxInfo pp lang ei sysS utxo tx = do
           { PV1.txInfoInputs = mapMaybe (uncurry txInfoIn) txIns
           , PV1.txInfoOutputs = mapMaybe txInfoOut (foldr (:) [] txOuts)
           , PV1.txInfoFee = transValue (inject @(MaryValue (EraCrypto era)) fee)
-          , PV1.txInfoMint = transMintValue (txBody ^. mintTxBodyL)
-          , PV1.txInfoDCert = foldr (\c ans -> transDCert c : ans) [] (txBody ^. certsTxBodyG)
+          , PV1.txInfoMint = transMultiAsset (txBody ^. mintTxBodyL)
```
lehins: We need to switch it back to `transMintValue`. This is only fixed in with `PlutusV3`

I know I was the one who did it, that was before we figured out how to proceed about it
```suggestion
          , PV1.txInfoMint = transMintValue (txBody ^. mintTxBodyL)
```

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Tx.hs
```diff
@@ -367,20 +367,29 @@ totExUnits tx =
...
-  deriving (Eq, Show, Generic, NoThunks, NFData)
-
-instance (Crypto c) => EncCBOR (ScriptPurpose c) where
-  encCBOR (Minting x) = encode (Sum Minting 0 !> To x)
-  encCBOR (Spending x) = encode (Sum Spending 1 !> To x)
-  encCBOR (Rewarding x) = encode (Sum Rewarding 2 !> To x)
+data ScriptPurpose era
+  = Minting !(PolicyID (EraCrypto era))
+  | Spending !(TxIn (EraCrypto era))
+  | Rewarding !(RewardAcnt (EraCrypto era)) -- Not sure if this is the right type.
```
Soupstraw:
```suggestion
  | Rewarding !(RewardAcnt (EraCrypto era))
```
It's already in production, so I sure hope this turned out to be the right type

JaredCorduan: :facepalm: yes, that type is correct. and that comment should have been removed. thank you!!

eras/allegra/impl/CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
   * `DState`
   * `PState`
   * `VState`
+* Add `ShelleyEraDCert` instance for `AllegraEra`
```
lehins, line 9:
```suggestion
* Add `EraDCert` and `ShelleyEraDCert` instance for `AllegraEra`
```

This change requires this bounds update: `cardano-ledger-shelley >=1.2`

eras/alonzo/impl/cardano-ledger-alonzo.cabal
```diff
@@ -24,6 +24,7 @@ library
         Cardano.Ledger.Alonzo
         Cardano.Ledger.Alonzo.Core
         Cardano.Ledger.Alonzo.Data
+        Cardano.Ledger.Alonzo.Delegation
```
lehins: This should be a hidden module, since it only exports class instances

eras/allegra/impl/testlib/Test/Cardano/Ledger/Allegra/Arbitrary.hs
```diff
@@ -102,7 +102,13 @@ instance
   arbitrary = genericArbitraryU
 
 instance
-  (EraTxOut era, Era era, Arbitrary (TxOut era), Arbitrary (PParamsHKD StrictMaybe era)) =>
+  ( EraTxOut era
+  , EraDCert era
+  , Era era
```
lehins: Redundant
```suggestion
```

eras/alonzo/impl/CHANGELOG.md
```diff
@@ -9,6 +9,15 @@
   * `VState`
 * Added support for Plutus V3 in the types and functions that use `Language`.
   (Note that the Alonzo ledger era rules do not allow V3 scripts, however.).
+* Change parametrization of `ScriptPurpose`, `CollectError` and `TransactionScriptFailure`
+  from `c`rypto to `era`
+* Add `EraDCert` and `ShelleyEraDCert` instances to `AlonzoEra`
+* Rename `transMultiAsset` to `transMintValue`
```
lehins: This is incorrect. It was not renamed, `transMultiAsset` was a new function, but it was added in a previous release, so this entry should not be here
```suggestion
```

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Rules/Utxos.hs
```diff
@@ -266,6 +275,7 @@ scriptsNotValidateTransition ::
   , ScriptsNeeded era ~ AlonzoScriptsNeeded era
   , STS (AlonzoUTXOS era)
   , Script era ~ AlonzoScript era
+  , EraPlutusContext 'PlutusV1 era
```
lehins, line 278: This doesn't seem right. This function is used with other versions of Plutus, not just V1. How does it work?

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Rules/Utxow.hs
```diff
@@ -458,24 +448,25 @@ witsVKeyNeeded utxo' tx genDelegs =
...
       where
         accum (DCertPool (RegPool pool)) ans =
           Set.union
             (Set.map asWitness (ppOwners pool))
             ans
         accum _cert ans = ans
-    cwitness (DCertDeleg dc) = extractKeyHashWitnessSet [delegCWitness dc]
+    cwitness (DCertShelleyDeleg dc) = extractKeyHashWitnessSet [delegCWitness dc]
     cwitness (DCertPool pc) = extractKeyHashWitnessSet [poolCWitness pc]
-    cwitness (DCertGenesis gc) = Set.singleton (asWitness $ genesisCWitness gc)
```
lehins, line 472: Why does `DCertGenesis` pattern doesn't work here?

Functionally, it should be the same as what you have below with `getDCertGenesis`

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/PlutusScriptApi.hs
```diff
@@ -24,27 +26,24 @@ module Cardano.Ledger.Alonzo.PlutusScriptApi (
 )
 where
 
-import Cardano.Ledger.Alonzo.Core (MaryEraTxBody (..))
+import Cardano.Ledger.Alonzo.Core hiding (TranslationError)
```
JaredCorduan, line 29: I personally prefer to avoid open imports. But if I'm the only one, I won't object.

eras/conway/impl/src/Cardano/Ledger/Conway/Delegation/Certificates.hs
```diff
@@ -89,48 +130,59 @@ instance NFData (ConwayDelegCert c)
 
 instance NoThunks (ConwayDelegCert c)
 
-data ConwayDCert c
-  = ConwayDCertDeleg !(ConwayDelegCert c)
-  | ConwayDCertPool !(PoolCert c)
-  | ConwayDCertConstitutional !(ConstitutionalDelegCert c)
+data ConwayDCert era
+  = ConwayDCertNoOp !(StakeCredential (EraCrypto era))
```
JaredCorduan: should this be called `ConwayDCertRegistration`?

eras/conway/impl/src/Cardano/Ledger/Conway/Delegation/Certificates.hs
```diff
@@ -89,48 +130,59 @@ instance NFData (ConwayDelegCert c)
...
-        pure (3, ConwayDCertPool $ RetirePool a (EpochNo b))
-      5 -> do
-        a <- decCBOR
-        b <- decCBOR
-        c <- decCBOR
-        pure (4, ConwayDCertConstitutional $ ConstitutionalDelegCert a b c)
-      6 -> do
+      t | t <= 6 -> do
+        (l, c) <- shelleyDCertDecoder @era t
+        c' <- transShelleyDCert c
```
JaredCorduan: nice!

eras/conway/test-suite/cddl-files/conway.cddl
```diff
@@ -253,7 +253,9 @@ script_data_hash = $hash32
 ; 1001 - 1101: future formats
 
 certificate =
-  [ pool_registration
+  [ noop_cert
```
JaredCorduan: registration?

eras/conway/test-suite/cddl-files/conway.cddl
```diff
@@ -267,12 +269,16 @@ delegatee =
   // 2, pool_keyhash, voting_credential
   ]
 
+
+noop_cert = (0, stake_credential)
+stake_deregistration = (1, stake_credential)
+stake_delegation = (2, stake_credential, pool_keyhash)
 pool_registration = (3, pool_params)
 pool_retirement = (4, pool_keyhash, epoch)
 constitutional_deleg_cert = (5, genesishash, genesis_delegate_hash, vrf_keyhash)
```
JaredCorduan, line 278: what happened to `6`? why did these get bumped?

eras/conway/impl/src/Cardano/Ledger/Conway/Delegation/Certificates.hs
lehins: There is really no need to keep around the stake credential, because it is unused. So, my opinion is we have two options:

1. If we keep the credential in the type, then we should call it:
   ```haskell
     = ConwayDCertReg !(StakeCredential (EraCrypto era))
       -- ^ This type is unused and is here as part of deprecation process
   ```
2. Discard the credential completely and call this a noop:
   ```haskell
     = ConwayDCertNoOp
       -- ^ This type is here as part of deprecation process of
       -- of the registration certificate
   ```

lehins (commented):
My feedback so far, I'll finish up the rest later on tonight

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxInfo.hs
```diff
@@ -385,30 +385,50 @@ transMintValue m = transMultiAsset m <> justZeroAda
     justZeroAda = PV1.singleton PV1.adaSymbol PV1.adaToken 0
 
 transValue :: MaryValue c -> PV1.Value
-transValue (MaryValue n m) = justAda <> transMultiAsset m
+transValue (MaryValue n m) = justAda <> transMintValue m
```
lehins: Bad rebase on master
```suggestion
transValue (MaryValue n m) = justAda <> transMultiAsset m
```

eras/babbage/impl/src/Cardano/Ledger/Babbage/Rules/Utxos.hs
```diff
@@ -126,6 +131,7 @@ utxosTransition ::
   , EncCBOR (PPUPPredFailure era)
   , Eq (PPUPPredFailure era)
   , Show (PPUPPredFailure era)
+  , EraPlutusContext 'PlutusV1 era
```
lehins, line 134: I think I see what the problem is now with the language, it is because we don't have it available at the type level. I'll need to think about what can we do about this. I think  it would be ok to leave it like this for now

eras/conway/impl/src/Cardano/Ledger/Conway.hs
```diff
@@ -42,17 +43,30 @@ type Conway = ConwayEra StandardCrypto
...
+  ) =>
+  API.ApplyBlock (ConwayEra c)
 
 instance Crypto c => API.CanStartFromGenesis (ConwayEra c) where
   type AdditionalGenesisConfig (ConwayEra c) = ConwayGenesis c
   fromShelleyPParams =
     error "Unimplemented: Current interface is too limited and needs replacement for Conway to work"
 
-instance Crypto c => ExtendedUTxO (ConwayEra c) where
+instance (Crypto c, EraPlutusContext 'PlutusV2 (ConwayEra c), EraPlutusContext 'PlutusV3 (ConwayEra c)) => ExtendedUTxO (ConwayEra c) where
```
lehins: Switch to multiline?

eras/conway/impl/src/Cardano/Ledger/Conway/TxBody.hs
```diff
@@ -155,7 +155,7 @@ deriving instance
   Show (ConwayTxBodyRaw era)
 
 instance
-  (EraPParams era, DecCBOR (TxOut era)) =>
+  (EraPParams era, ShelleyEraDCert era, DecCBOR (TxOut era)) =>
```
lehins: This constraint shouldn't be here, since we dropped `MIR` certificates
```suggestion
  (EraPParams era, DecCBOR (TxOut era)) =>
```

eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState/RefundsAndDeposits.hs
```diff
@@ -85,34 +84,34 @@ totalTxDeposits ::
...
   where
     keyDeposit = pp ^. ppKeyDepositL
     accum (!regKeys, !totalRefunds) = \case
-      DCertDeleg (RegKey k) ->
+      DCertShelleyDeleg (RegKey k) ->
         -- Need to track new delegations in case that the same key is later deregistered in
         -- the same transaction.
         (Set.insert k regKeys, totalRefunds)
-      DCertDeleg (DeRegKey k)
+      DCertShelleyDeleg (DeRegKey k)
```
lehins: Convention is to put the era name as a prefix
```suggestion
      ShelleyDCertDeleg (DeRegKey k)
```

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/PlutusScriptApi.hs
Soupstraw: IIRC we agreed to allow open imports for `Core` libraries, but I agree in general

eras/conway/impl/src/Cardano/Ledger/Conway/Delegation/Certificates.hs
Soupstraw: ~~I think that not having the stake credential breaks the round-trip tests, since there's no way to reconstruct the original CBOR without that information~~ Seems to be working fine.

eras/conway/test-suite/cddl-files/conway.cddl
Soupstraw: `6` used to be the MIR certificates. In Conway trying to decode a MIR cert will give an error, but I'm not sure if we actually need to keep that around.

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Rules/Utxos.hs
Soupstraw: We should be able to get rid of these once we add `transScriptPurpose` to `EraPlutusContext`. I figured I'll leave it to a future PR.

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/PlutusScriptApi.hs
JaredCorduan: right, I do remember us saying that `Cardano.Ledger.Core` might be a reasonable middle ground. Let's try all the `Core` libraries and see how it goes.

eras/conway/test-suite/cddl-files/conway.cddl
JaredCorduan: ah yes, thanks! maybe we could leave a comment saying that `6` was MIR, which is now deprecated?

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/PlutusScriptApi.hs
Soupstraw: It might even be better to only allow open importing the Core module of the current era, since it will re-export all the previous Core modules and we also have better control over what not to export should there be something that clashes with things in that era.

eras/conway/test-suite/cddl-files/conway.cddl
Soupstraw: good catch!

eras/conway/impl/src/Cardano/Ledger/Conway/Delegation.hs
```diff
@@ -0,0 +1,36 @@
+{-# LANGUAGE TypeFamilies #-}
+{-# OPTIONS_GHC -Wno-orphans #-}
+
+module Cardano.Ledger.Allegra.Delegation () where
+
+import Cardano.Ledger.Allegra.Era
+import Cardano.Ledger.Crypto
+import Cardano.Ledger.Shelley.Delegation
+
+instance Crypto c => EraDCert (AllegraEra c) where
```
aniketd: I think we may have missed out renaming `Allegra` to `Conway` here.

Soupstraw: good catch, looks like this file isn't even referenced in the cabal file

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/PlutusScriptApi.hs
lehins: > we also have better control over what not to export should there be something that clashes with things in that era.

There should never be a case that something is not re-exported from the previous era in the Core module. So, I don't think we need to enforce import of Core module from the current era, but I do agree it would be a good practice to use an open import of `Cardano.Ledger.[Era].Core` for modules within the era.

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Rules/Utxos.hs
lehins: Fixing in a subsequent PR. Sounds good

eras/conway/impl/src/Cardano/Ledger/Conway/Delegation/Certificates.hs
```diff
@@ -28,19 +35,53 @@ import Cardano.Ledger.Binary.Coders (
...
+  getDCertShelleyDeleg (ConwayDCertDeleg (ConwayDeleg c (DelegStake kh) _)) =
+    Just $ Delegate (Delegation c kh)
+  getDCertShelleyDeleg _ = Nothing
+
+  mkDCertMir = notSupportedInThisEra
+  getDCertMir = const Nothing
+
+class ShelleyEraDCert era => ConwayEraDCert era where
+  mkDCertConwayDeleg :: ConwayDelegCert (EraCrypto era) -> DCert era
+  getDCertConwayDeleg :: DCert era -> Maybe (ConwayDelegCert (EraCrypto era))
```
lehins: Let's follow the convention here as well of prefix era:
```suggestion
  mkConwayDCertDeleg :: ConwayDelegCert (EraCrypto era) -> DCert era
  getConwayDCertDeleg :: DCert era -> Maybe (ConwayDelegCert (EraCrypto era))
```

eras/conway/impl/src/Cardano/Ledger/Conway/Delegation/Certificates.hs
lehins, line 1: Let's do a `git mv` on this module to `Cardano.Ledger.Conway.Delegation` (same for other eras)

There is a redundant level, IMHO.

Alternatively we could rename them in all eras to be `Cardano.Ledger.[Era].Certificates`, but that would require a little bit more work. I don't have a preference

eras/conway/impl/src/Cardano/Ledger/Conway/Delegation/Certificates.hs
```diff
@@ -89,48 +130,59 @@ instance NFData (ConwayDelegCert c)
 
 instance NoThunks (ConwayDelegCert c)
 
-data ConwayDCert c
-  = ConwayDCertDeleg !(ConwayDelegCert c)
-  | ConwayDCertPool !(PoolCert c)
-  | ConwayDCertConstitutional !(ConstitutionalDelegCert c)
+data ConwayDCert era
+  = ConwayDCertRegKey !(StakeCredential (EraCrypto era))
```
JaredCorduan: let's remove this in favor of a new `ConwayRegKey` constructor in `ConwayDelegCert`.

```
ConwayRegKey !(StakeCredential c) !Coin
```

JaredCorduan: also we need to make the `Coin`s a `Maybe` (with a null maybe decoder)

eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState/RefundsAndDeposits.hs
```diff
@@ -129,4 +128,4 @@ keyTxRefunds ::
   CertState era ->
   TxBody era ->
   Coin
-keyTxRefunds pp dpstate tx = keyCertsRefundsCertState pp dpstate (tx ^. certsTxBodyG)
+keyTxRefunds pp dpstate tx = keyCertsRefundsCertState pp dpstate (toList $ tx ^. certsTxBodyL)
```
lehins: There is no need to convert to a list, `keyCertsRefundsCertState` accepts a foldable
```suggestion
keyTxRefunds pp dpstate tx = keyCertsRefundsCertState pp dpstate (tx ^. certsTxBodyL)
```

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Deleg.hs
```diff
@@ -285,13 +283,13 @@ delegationTransition = do
...
       pure u4
-    DCertDeleg (Delegate (Delegation hk dpool)) -> do
+    DCertShelleyDeleg (Delegate (Delegation hk dpool)) -> do
       -- note that pattern match is used instead of cwitness and dpool, as in the spec
       -- (hk ∈ dom (rewards ds))
       UM.member hk (rewards ds) ?! StakeDelegationImpossibleDELEG hk
 
       pure (ds {dsUnified = delegations ds UM.⨃ Map.singleton hk dpool})
-    DCertGenesis (ConstitutionalDelegCert gkh vkh vrf) -> do
+    _ | Just (ConstitutionalDelegCert gkh vkh vrf) <- getDCertGenesis c -> do
```
lehins:
```suggestion
    DCertGenesis (ConstitutionalDelegCert gkh vkh vrf) -> do
```

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Deleg.hs
```diff
@@ -321,63 +319,68 @@ delegationTransition = do
           { dsFutureGenDelegs =
               eval (dsFutureGenDelegs ds ⨃ singleton (FutureGenDeleg s' gkh) (GenDelegPair vkh vrf))
           }
-    DCertMir (MIRCert targetPot mirTarget) -> do
```
lehins, line 324: What's wrong with using this pattern synonym?

Could you revert it back, please?

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Pool.hs
```diff
@@ -227,12 +234,16 @@ poolDelegationTransition = do
...
       pure $ ps {psRetiring = eval (psRetiring ps ⨃ singleton hk e)}
-    DCertDeleg _ -> do
+    DCertShelleyDeleg _ -> do
       failBecause $ WrongCertificateTypePOOL 0
       pure ps
-    DCertMir _ -> do
+    _ | isGenesisDelegation c -> do
+      failBecause $ WrongCertificateTypePOOL 2
+      pure ps
+    _ | isInstantaneousRewards c -> do
```
lehins: Could you revert here as well. All the pattern synonyms were created precisely so this sort of changes where unnecessary

eras/conway/impl/src/Cardano/Ledger/Conway/Delegation/Certificates.hs
```diff
@@ -89,48 +130,59 @@ instance NFData (ConwayDelegCert c)
...
 instance NFData (ConwayDCert c)
 
 instance NoThunks (ConwayDCert c)
 
-instance Crypto c => DecCBOR (ConwayDCert c) where
+instance
+  ( ShelleyEraDCert era
+  ) =>
+  FromCBOR (ConwayDCert era)
+  where
```
lehins:
```suggestion
instance ShelleyEraDCert era => FromCBOR (ConwayDCert era) where
```

eras/conway/impl/src/Cardano/Ledger/Conway/Delegation/Certificates.hs
```diff
@@ -89,48 +130,59 @@ instance NFData (ConwayDelegCert c)
...
+  fromCBOR = toPlainDecoder (eraProtVerLow @era) decCBOR
+
+instance
+  ( ShelleyEraDCert era
+  ) =>
+  DecCBOR (ConwayDCert era)
+  where
```
lehins:
```suggestion
instance ShelleyEraDCert era => DecCBOR (ConwayDCert era) where
```

eras/shelley/impl/src/Cardano/Ledger/Shelley/Delegation.hs
```diff
@@ -0,0 +1,349 @@
...
+  toCBOR = toEraCBOR @era
+
+instance Era era => FromCBOR (ShelleyDCert era) where
+  fromCBOR = fromEraCBOR @era
+
+instance Era era => DecCBOR (ShelleyDCert era) where
+  decCBOR = decodeRecordSum "DCert crypto" shelleyDCertDecoder
+
+shelleyDCertDecoder ::
+  (Crypto (EraCrypto era)) =>
```
lehins:
```suggestion
  Era era =>
```

eras/shelley/impl/src/Cardano/Ledger/Shelley/Delegation.hs
```diff
@@ -0,0 +1,349 @@
...
+  b <- decCBOR
+  pure (3, ShelleyDCertPool $ RetirePool a (EpochNo b))
+shelleyDCertDecoder 5 = do
+  a <- decCBOR
+  b <- decCBOR
+  c <- decCBOR
+  pure (4, ShelleyDCertGenesis $ ConstitutionalDelegCert a b c)
+shelleyDCertDecoder 6 = do
+  x <- decCBOR
+  pure (2, ShelleyDCertMir x)
```
lehins: `MIR` certificates should be handled separately in the `DecCBOR (ShelleyDCert era)` iunstance itself, since that is not suppose to be a possible case in any era after Babbage
```suggestion
```

eras/conway/impl/src/Cardano/Ledger/Conway/Delegation/Certificates.hs
lehins: This is wrong. This will allow MIR certificates to be deserialized.

eras/conway/impl/src/Cardano/Ledger/Conway/Delegation/Certificates.hs
```diff
@@ -80,57 +120,68 @@ instance NFData (Delegatee c)
...
-      5 -> do
-        a <- decCBOR
-        b <- decCBOR
-        c <- decCBOR
-        pure (4, ConwayDCertConstitutional $ ConstitutionalDelegCert a b c)
-      6 -> do
+      t | t <= 6 -> do
+        (l, c) <- shelleyDCertDecoder @era t
+        c' <- transShelleyDCert c
+        pure (l, c')
```
lehins: This is wrong. We don't handle MIR certificates in Conway, so we shouldn't even deserialize them
```suggestion
      t | t <= 5 -> shelleyDCertDecoder @era t
```

eras/shelley/impl/src/Cardano/Ledger/Shelley/Delegation.hs
```diff
@@ -0,0 +1,349 @@
...
+instance Era era => FromCBOR (ShelleyDCert era) where
+  fromCBOR = fromEraCBOR @era
+
+instance Era era => DecCBOR (ShelleyDCert era) where
+  decCBOR = decodeRecordSum "DCert crypto" shelleyDCertDecoder
+
+shelleyDCertDecoder ::
+  (Crypto (EraCrypto era)) =>
+  Word ->
+  Decoder s (Int, ShelleyDCert era)
```
lehins: It should be a type family, so it can be reused in Conway. This would mean that all type constructors will need to be switched to pattern synonyms
```suggestion
  Decoder s (Int, DCert era)
```

eras/shelley/impl/src/Cardano/Ledger/Shelley/Delegation.hs
```diff
@@ -0,0 +1,349 @@
...
+  DCert era ->
+  Bool
+isReservesMIRCert x = case getDCertMir x of
+  Just (MIRCert ReservesMIR _) -> True
+  _ -> False
+
+isTreasuryMIRCert ::
+  ShelleyEraDCert era =>
+  DCert era ->
+  Bool
```
lehins: If it fits into 80 characters, I advise against switching to multi-line
```suggestion
isTreasuryMIRCert :: ShelleyEraDCert era => DCert era -> Bool
```

eras/shelley/impl/src/Cardano/Ledger/Shelley/Delegation.hs
```diff
@@ -0,0 +1,349 @@
...
+isInstantaneousRewards ::
+  ShelleyEraDCert era =>
+  DCert era ->
+  Bool
+isInstantaneousRewards = isJust . getDCertMir
+
+isReservesMIRCert ::
+  ShelleyEraDCert era =>
+  DCert era ->
+  Bool
```
lehins:
```suggestion
isReservesMIRCert :: ShelleyEraDCert era => DCert era -> Bool
```

eras/shelley/impl/src/Cardano/Ledger/Shelley/Delegation.hs
```diff
@@ -0,0 +1,349 @@
...
+
+-- | Check for 'RetirePool' constructor
+isRetirePool :: EraDCert era => DCert era -> Bool
+isRetirePool (DCertPool (RetirePool _ _)) = True
+isRetirePool _ = False
+
+isInstantaneousRewards ::
+  ShelleyEraDCert era =>
+  DCert era ->
+  Bool
```
lehins:
```suggestion
isInstantaneousRewards :: ShelleyEraDCert era => DCert era -> Bool
```

eras/shelley/impl/src/Cardano/Ledger/Shelley/Delegation.hs
```diff
@@ -0,0 +1,349 @@
...
+      0 -> pure ReservesMIR
+      1 -> pure TreasuryMIR
+      k -> invalidKey k
+
+-- | MIRTarget specifies if funds from either the reserves
+-- or the treasury are to be handed out to a collection of
+-- reward accounts or instead transfered to the opposite pot.
+data MIRTarget c
+  = StakeAddressesMIR (Map (Credential 'Staking c) DeltaCoin)
+  | SendToOppositePotMIR Coin
```
lehins: Not something you changed, just something I noticed.
```suggestion
  = StakeAddressesMIR !(Map (Credential 'Staking c) DeltaCoin)
  | SendToOppositePotMIR !Coin
```

eras/shelley/impl/src/Cardano/Ledger/Shelley/Delegation.hs
```diff
@@ -0,0 +1,349 @@
...
+      _ -> SendToOppositePotMIR <$> decCBOR
+
+instance Crypto c => EncCBOR (MIRTarget c) where
+  encCBOR (StakeAddressesMIR m) = encCBOR m
+  encCBOR (SendToOppositePotMIR c) = encCBOR c
+
+-- | Move instantaneous rewards certificate
+data MIRCert c = MIRCert
+  { mirPot :: MIRPot
+  , mirRewards :: MIRTarget c
```
lehins:
```suggestion
  { mirPot :: !MIRPot
  , mirRewards :: !(MIRTarget c)
```

eras/shelley/test-suite/src/Test/Cardano/Ledger/Shelley/Rules/Deleg.hs
```diff
@@ -111,10 +112,10 @@ keyRegistration
 keyRegistration _ = property ()
 
 -- | Check stake key de-registration
-keyDeRegistration :: SourceSignalTarget (ShelleyDELEG era) -> Property
+keyDeRegistration :: (ShelleyEraDCert era) => SourceSignalTarget (ShelleyDELEG era) -> Property
 keyDeRegistration
   SourceSignalTarget
-    { signal = (DCertDeleg (DeRegKey hk))
+    { signal = (DCertShelleyDeleg (DeRegKey hk))
```
lehins: Might as well
```suggestion
    { signal = DCertShelleyDeleg (DeRegKey hk)
```

eras/shelley/test-suite/src/Test/Cardano/Ledger/Shelley/Rules/Deleg.hs
```diff
@@ -128,10 +129,10 @@ keyDeRegistration
 keyDeRegistration _ = property ()
 
 -- | Check stake key delegation
-keyDelegation :: SourceSignalTarget (ShelleyDELEG era) -> Property
+keyDelegation :: (ShelleyEraDCert era) => SourceSignalTarget (ShelleyDELEG era) -> Property
 keyDelegation
   SourceSignalTarget
-    { signal = (DCertDeleg (Delegate (Delegation from to)))
+    { signal = (DCertShelleyDeleg (Delegate (Delegation from to)))
```
lehins:
```suggestion
    { signal = DCertShelleyDeleg (Delegate (Delegation from to))
```

eras/conway/test-suite/cddl-files/conway.cddl
```diff
@@ -267,12 +270,16 @@ delegatee =
...
-deleg_cert = (6, stake_credential, delegatee, coin)
-redeleg_cert = (7, stake_credential, delegatee)
-undeleg_cert = (8, stake_credential, coin)
+; number 6 used to be the MIR certificate, which is deprecated in Conway
+deleg_cert = (7, stake_credential, delegatee, coin)
```
lehins, line 280: Coin need to switch to optionally being null

eras/conway/test-suite/cddl-files/conway.cddl
```diff
@@ -267,12 +270,16 @@ delegatee =
   // 2, pool_keyhash, voting_credential
   ]
 
+stake_registration = (0, stake_credential)
```
lehins, line 273: This one will also get an optional coin

lehins (changes requested):
This is a really good piece of work.

eras/conway/impl/src/Cardano/Ledger/Conway/Delegation/Certificates.hs
```diff
@@ -28,18 +37,52 @@ import Cardano.Ledger.Binary.Coders (
...
+  getDCertPool _ = Nothing
+
+  mkDCertGenesis = ConwayDCertConstitutional
+  getDCertGenesis (ConwayDCertConstitutional x) = Just x
+  getDCertGenesis _ = Nothing
+
+instance Crypto c => ShelleyEraDCert (ConwayEra c) where
+  mkDCertShelleyDeleg = transShelleyDeleg
+
+  getDCertShelleyDeleg (ConwayDCertDeleg (ConwayDeleg c (DelegStake kh) _)) =
```
lehins: I think this one needs a few more cases

eras/conway/impl/src/Cardano/Ledger/Conway/Delegation/Certificates.hs
```diff
@@ -1,23 +1,32 @@
 {-# LANGUAGE DataKinds #-}
 {-# LANGUAGE DeriveGeneric #-}
 {-# LANGUAGE LambdaCase #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeApplications #-}
+{-# LANGUAGE TypeFamilies #-}
+{-# OPTIONS_GHC -Wno-orphans #-}
```
lehins: For consistency sake it would be good to move this pragma below the language extensions

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Pool.hs
Soupstraw: I can revert it for DCertGenesis, but when I use this kind of pattern matching for certs that get removed in a later era, then I don't have to propagate the protocol version constraint, since such certs will never match anyways in those eras.

eras/shelley/impl/src/Cardano/Ledger/Shelley/Delegation.hs
```diff
@@ -0,0 +1,352 @@
...
+  getDCertPool _ = Nothing
+
+  mkDCertGenesis = ShelleyDCertGenesis
+
+  getDCertGenesis (ShelleyDCertGenesis c) = Just c
+  getDCertGenesis _ = Nothing
+
+class EraDCert era => ShelleyEraDCert era where
+  mkDCertMir :: ProtVerAtMost era 8 => MIRCert (EraCrypto era) -> DCert era
+  getDCertMir :: DCert era -> Maybe (MIRCert (EraCrypto era))
```
lehins, line 101:
```suggestion
  getDCertMir :: ProtVerAtMost era 8 => DCert era -> Maybe (MIRCert (EraCrypto era))
```

This one also needs a constraint. We want to prevent any code that relies on `MIR` certificates to be used in Conway. This will have to be done in a later era regardless, so might as well. tackle this now. That being said, it will be better if we do this in a separate PR, this one is already pretty big

eras/conway/test-suite/cddl-files/conway.cddl
Soupstraw: Not relevant anymore, since we decided to always decode the deposit in Conway and never encode it in Shelley.

eras/conway/test-suite/cddl-files/conway.cddl
Soupstraw: Ditto

JaredCorduan (approved):
This is really awesome stuff, @Soupstraw !! thanks for this massive effort!

lehins (approved):
Beautiful. Thank you!

Feel free to create some ticket for things you think can be done as follow up to this PR

PR #3389, opened 2023-04-21
Reduce gh pages size

teodanciu:
# Description

Only add the haddocks to the commit on gh-pages branch, rather than force-adding everything. The latter creates huge git objects, which can causes failures and delays when fetching etc. 

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/.github/workflows/gh-pages.yml
+++ b/.github/workflows/gh-pages.yml
@@ -25,10 +25,9 @@ jobs:
         run: |
               git config --local user.name ${{ github.actor }}
               git config --local user.email "${{ github.actor }}@users.noreply.github.com"
-              cp -r ./haddocks/* ./
+              git add -A --force ./haddocks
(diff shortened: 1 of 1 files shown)


.github/workflows/gh-pages.yml
```diff
@@ -27,6 +27,7 @@ jobs:
               git config --local user.email "${{ github.actor }}@users.noreply.github.com"
               cp -r ./haddocks/* ./
               rm -rf haddocks
+              rm -rf dist-newstyle
```
lehins: :grin: Yeah, we don't want to check in all of the artifacts

teodanciu:
I think this is a better solution: only adding what's necessary, instead of force-adding everything and then removing

lehins (approved):
Nice. Very clever trick! Love it! :heart:

PR #3390, opened 2023-04-24
reinstate our two tickf optimizations

JaredCorduan:
# Description

While preparing for a new release, we had to remove our two `TICKF` optimizations: https://github.com/input-output-hk/cardano-ledger/pull/3375

The problem was that a local network test in cardano-node was failing, namely `cardano-testnet`.

The problem, however, was not in the optimizations, but rather in the test setup:

https://github.com/input-output-hk/ouroboros-network/blob/8685fc4b08cd5f11735768d60da921628602b630/ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Node/TPraos.hs#L336-L344

The problem is that the inserted `mark` snapshot is now not enough to quick-start the network, the pool distribution also needs to quick-started:

```
SL.ssStakeMarkPoolDistr = SL.calculatePoolDistr initSnapShot
```
 
I verified this by running `cardano-testnet` with a modified ledger and consensus as described above.

We can now restore the optimizations in ledger, but we still need to either fix `registerGenesisStaking` as described above, or do something safer to prevent future mistakes like this.

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/NewEpoch.hs
+++ b/eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/NewEpoch.hs
@@ -168,9 +168,9 @@ newEpochTransition = do
       es''' <- trans @(EraRule "EPOCH" era) $ TRC ((), es'', eNo)
       let adaPots = totalAdaPotsES es'''
       tellEvent $ TotalAdaPotsEvent adaPots
-      -- let pd' = ssStakeMarkPoolDistr (esSnapshots es)
+      let pd' = ssStakeMarkPoolDistr (esSnapshots es)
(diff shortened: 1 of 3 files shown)


lehins (approved):
Awesome. Make sure that the version of `ouroboros-consensus-cardano` with a fix to DRAGONS, has a lower bound on `cardano-ledger-shelley >= 1.2`

JaredCorduan:
> Awesome. Make sure that the version of `ouroboros-consensus-cardano` with a fix to DRAGONS, has a lower bound on `cardano-ledger-shelley >= 1.2`

Sounds good! We will approach the dragon :dragon: with shields :shield: made of `1.2` :crossed_swords: .

PR #3393, opened 2023-04-26
Performance improvements

lehins:
# Description

This PR takes care of the performance regression we encountered recently with 8.0.0

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/libs/cardano-ledger-binary/src/Cardano/Ledger/Binary/Decoding/DecCBOR.hs
+++ b/libs/cardano-ledger-binary/src/Cardano/Ledger/Binary/Decoding/DecCBOR.hs
@@ -87,6 +87,7 @@ class Typeable a => DecCBOR a where
   decCBOR :: Decoder s a
   default decCBOR :: Plain.FromCBOR a => Decoder s a
   decCBOR = fromPlainDecoder Plain.fromCBOR
+  {-# INLINE decCBOR #-}
 
   -- | Validate decoding of a Haskell value, without the need to actually construct
   -- it. Coule be slightly faster than `decCBOR`, however it should respect this law:
@@ -100,11 +101,13 @@ class Typeable a => DecCBOR a where
 
 instance DecCBOR Version where
   decCBOR = decodeVersion
+  {-# INLINE decCBOR #-}
 
 -- | Convert a versioned `DecCBOR` instance to a plain `Plain.Decoder` using Byron protocol
 -- version.
 fromByronCBOR :: DecCBOR a => Plain.Decoder s a
 fromByronCBOR = toPlainDecoder byronProtVer decCBOR
+{-# INLINE fromByronCBOR #-}
 
 --------------------------------------------------------------------------------
 -- Primitive types
@@ -112,9 +115,11 @@ fromByronCBOR = toPlainDecoder byronProtVer decCBOR
 
 instance DecCBOR () where
   decCBOR = decodeNull
+  {-# INLINE decCBOR #-}
 
 instance DecCBOR Bool where
   decCBOR = decodeBool
+  {-# INLINE decCBOR #-}
 
 --------------------------------------------------------------------------------
 -- Numeric data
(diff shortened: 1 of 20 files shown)


eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/NewEpoch.hs
```diff
@@ -236,7 +236,7 @@ updateRewards ::
 updateRewards es e ru'@(RewardUpdate dt dr rs_ df _) = do
   let totRs = sumRewards (esPrevPp es ^. ppProtocolVersionL) rs_
   Val.isZero (dt <> (dr <> toDeltaCoin totRs <> df)) ?! CorruptRewardUpdate ru'
-  let !(!es', filtered) = applyRUpdFiltered ru' es
+  let (!es', filtered) = applyRUpdFiltered ru' es
```
JaredCorduan: does the tuple constructor itself not need to be marked strict as well?

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Snap.hs
```diff
@@ -109,7 +110,7 @@ snapTransition = do
   pure $
     SnapShots
       { ssStakeMark = istakeSnap
-      , ssStakeMarkPoolDistr = calculatePoolDistr istakeSnap
+      , ssStakeMarkPoolDistr = PoolDistr mempty -- calculatePoolDistr istakeSnap
```
JaredCorduan: guessing you didn't mean to leave this here

libs/cardano-ledger-binary/src/Cardano/Ledger/Binary/Decoding/DecCBOR.hs
```diff
@@ -404,31 +451,21 @@ instance DecCBOR UTCTime where
...
-      Nothing ->
-        fail $
-          "hash bytes wrong size, expected "
-            ++ show expected
-            ++ " but got "
-            ++ show actual
-        where
-          expected = sizeHash (Proxy :: Proxy h)
-          actual = BS.length bs
+instance (HashAlgorithm h, Typeable a) => DecCBOR (Hash h a)
```
JaredCorduan, line 468: whoa, how does this work (removing `decCBOR`)?

libs/cardano-ledger-binary/src/Cardano/Ledger/Binary/Decoding/Decoder.hs
```diff
@@ -915,16 +957,15 @@ decodeMapEnforceNoDuplicates decodeKey decodeValueFor = do
...
         else do
-          (k, v) <- decodeInlinedPair
-          when (k `Map.member` acc) $ fail "Duplicate key detected in the Map"
-          loop condition nextStep (Map.insert k v acc)
-    decodeInlinedPair = do
-      !key <- decodeKey
-      !value <- decodeValueFor key
-      pure (key, value)
+          !key <- decodeKey
+          !value <- decodeValueFor key
```
JaredCorduan, line 961: is this just a style change, or is there something substantive?

lehins: Instead of adding an `INLINE` pragma for the function to be inlined, I decided to literally inline the function. Also, note that allocating of a tuple is no longer necessary.

libs/cardano-ledger-binary/src/Cardano/Ledger/Binary/Decoding/DecCBOR.hs
lehins: Default implementation for `DecCBOR` falls back onto `FromCBOR`. In case of Hash they were identical.

JaredCorduan: ah, so I guess the size is checked in `FromCBOR`, thanks

libs/cardano-ledger-binary/src/Cardano/Ledger/Binary/Decoding/Decoder.hs
JaredCorduan: ah!

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Snap.hs
lehins: Yeah, I'll revert this one. It was an experiment that I did to make sure it is not causing a slow down in #3394

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/NewEpoch.hs
lehins: Yes, it is needed. That was the whole reason why I added it in the first place in #3375. However, I was curious if it is causing a slow down so I reverted this change in #3394 

I am gonna revert this one too.

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Snap.hs
lehins: Interestingly enough this change of setting `ssStakeMarkPoolDistr` had a positive affect on performance. I can't really explain why, considering it is not used anywhere. 

I will revert it regardless, because this is necessary for #3390

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/NewEpoch.hs
lehins: This had no affect on performance, so I can confirm that I can revert it back

JaredCorduan (approved):
LGTM!

PR #3402, opened 2023-05-05
Fix broken references in the Shelley spec

JaredCorduan:
I've fixed the broken references in the Shelley spec:

* Deleg MIR rule
* reference to a no longer existent decay function
* MetaTarksi proof for deposit decay (no longer relevant)

---

- [x] Self-reviewed the diff


WhatisRT (approved):
LGTM

PR #3404, opened 2023-05-08
Create TotalDeposits events during all eras

kderme:
# Description

`TotalDeposits` appeared as event only during the Shelley era, since next eras have a different `Event (EraRule "UTXO")` instance. This pr makes sure the event appears during all eras.

Would it be possible to have new releases with this? I've updated the Changelogs assuming that.

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/eras/allegra/impl/CHANGELOG.md
+++ b/eras/allegra/impl/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 * Introduction of `TxCert` and `EraTxCert`
 * Add `EraTxCert` and `ShelleyEraTxCert` instances to `AllegraEra`
+* Fix an issue where `TotalDeposits` didn't appear on Allegra and Mary era
 
 ## 1.1.1.0
 
--- a/eras/allegra/impl/src/Cardano/Ledger/Allegra/Rules/Utxo.hs
+++ b/eras/allegra/impl/src/Cardano/Ledger/Allegra/Rules/Utxo.hs
@@ -14,6 +14,7 @@
 
 module Cardano.Ledger.Allegra.Rules.Utxo (
   AllegraUTXO,
+  AllegraUtxoEvent (..),
   AllegraUtxoPredFailure (..),
   validateOutsideValidityIntervalUTxO,
 )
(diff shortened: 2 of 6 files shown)


eras/allegra/impl/CHANGELOG.md
```diff
@@ -7,6 +7,10 @@
   * `PState`
   * `VState`
 
+## 1.1.0.1
+
+* Fix an issue where `TotalDeposits` didn't appear on Allegra and Mary era
```
JaredCorduan: We have not released 1.1.0.0 yet, so you should add your changelog entry there. We recently documented our new release process: https://github.com/input-output-hk/cardano-ledger/blob/e15e3545814140c63a972b1938372a075c288f6a/RELEASING.md#changelogmd

JaredCorduan:
good catch @kderme ! thanks for making the fix. After node 8.0.0.0 is released I will be doing another release of the ledger libraries.

eras/allegra/impl/CHANGELOG.md
kderme: I was wondering if this bugfix could be backported to the existing release branches for the 3 packages it affects, as mentioned at https://github.com/input-output-hk/cardano-ledger/blob/e15e3545814140c63a972b1938372a075c288f6a/RELEASING.md#release-process?

lehins: > could be backported to the existing release branches

No, it is not possible, because adding a constructor to an existing type is a breaking change and can only be included in a next major version bump.

kderme: How much major? Should it bump the first or the second release number? I bumped the second for now.

lehins: According to PVP version `A.B.C.D`:
* D - patch: non-semantic changes
* C - minor version: non-breaking changes
* B - major version: breaking changes
* A - complete rewrite :slightly_smiling_face:

Addition of a constructor to a type is a breaking change, so `B` needs a bump

JaredCorduan (approved):
Looks great, thank you @kderme !

PR #3412, opened 2023-05-09
Conway CERT and DELEG Rules

aniketd:
# Description

Add CERT and DELEG rules to Conway

# Checklist

- [x] Self-reviewed the diff


eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Deleg.hs
```diff
@@ -0,0 +1,255 @@
...
+    , cState@CertState
+        { certVState = VState{vsDReps}
+        , certPState = _
+        , certDState = certDState@DState{dsUnified} 
+        } 
+    , c
+    ) <- judgmentContext
+  let pd = pp ^. ppKeyDepositL
+  case c of
+    ConwayTxCertDeleg (ConwayRegCert stakeCred sMayCoin) -> do
```
Soupstraw:
```suggestion
    ConwayTxCertDeleg (ConwayRegCert stakeCred sMayDeposit) -> do
```
I think it's clearer when we specify what the coin is used for

eras/conway/impl/src/Cardano/Ledger/Conway/Era.hs
```diff
@@ -39,6 +40,7 @@ instance Crypto c => Era (ConwayEra c) where
   type PreviousEra (ConwayEra c) = BabbageEra c
   type EraCrypto (ConwayEra c) = c
   type ProtVerLow (ConwayEra c) = 9
+  -- type ProtVerHigh (ConwayEra c) = 10
```
Soupstraw: I'm not sure if we want to have an upper bound to the protocol before the next era is out, @lehins probably knows more about this.

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Deleg.hs
```diff
@@ -0,0 +1,255 @@
...
+      sMayCoin == SNothing || sMayCoin == SJust pd
```
Soupstraw:
```suggestion
      sMayCoin /= SJust pd
```
I think for now we'll want to allow `SNothing` in the deposits so we can make the decoder backwards compatible. The deposit field should be optional in Conway. Also the SJust case should be an inequality (so if the deposit is present, then it must be the correct amount).

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Deleg.hs
```diff
@@ -0,0 +1,255 @@
...
+import Cardano.Ledger.Coin (Coin)
+import Cardano.Ledger.CertState (VState (VState, vsDReps))
+import qualified Data.Set as Set
+import Cardano.Ledger.UMap (UView(RewDepUView, DRepUView, SPoolUView))
+import Cardano.Ledger.Core (Era(EraCrypto), EraRule, EraTx, TxCert)
+import Cardano.Ledger.Conway.Core ({-ShelleyEraTxBody,-} ConwayEraTxBody, ppKeyDepositL)
+import qualified Data.Map.Strict as Map
+
+data ConwayDelegPredFailure era
+  = RegCertProvidedCoinNotEqualToExpectedPoolDepositDELEG
```
Soupstraw: Maybe `IncorrectDepositProvided` would be a better name.

eras/conway/impl/src/Cardano/Ledger/Conway/Era.hs
lehins: Protocol version 10 is not necessarily Conway. So this would be wrong

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Deleg.hs
lehins: You meant to write `/=`, right? :grin:

Soupstraw: yes, my bad

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Deleg.hs
```diff
@@ -0,0 +1,255 @@
...
+        } 
+    , c
+    ) <- judgmentContext
+  let pd = pp ^. ppKeyDepositL
+  case c of
+    ConwayTxCertDeleg (ConwayRegCert stakeCred sMayCoin) -> do
+      sMayCoin == SNothing || sMayCoin == SJust pd
+        ?! RegCertProvidedCoinNotEqualToExpectedPoolDepositDELEG sMayCoin
+      UM.notMember stakeCred (RewDepUView dsUnified) 
+        ?! RegCertStakeKeyAlreadyRegisteredDELEG stakeCred
```
Soupstraw: nice, we haven't yet added this to the spec, but we definitely should!

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Deleg.hs
```diff
@@ -0,0 +1,255 @@
...
+      let ppDeposit = UM.compactCoinOrError pd
+          newRDs = UM.RewDepUView dsUnified UM.∪ (stakeCred, UM.RDPair (UM.CompactCoin 0) ppDeposit)
+          newUMap = UM.PtrUView newRDs UM.∪ (ptr, stakeCred)
```
Soupstraw: I don't think we care about pointer addresses in Conway, so this is unnecessary

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Deleg.hs
```diff
@@ -0,0 +1,255 @@
...
+      pure $ 
+        cState 
+          { certDState = 
+              certDState 
+                { dsUnified = newUMap }
+          }
+    ConwayTxCertDeleg (ConwayUnRegCert stakeCred sMayCoin) -> do
+      sMayCoin == SNothing || sMayCoin == SJust pd
```
Soupstraw, line 172: same as above

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Deleg.hs
Soupstraw: Also it should be fine to use the same error for all of the certificates, having a copy of the same error for each certificate is a pain to maintain.

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Deleg.hs
Soupstraw: oh, wait it's an `unless` rather than `when`? if so, then the original is correct.

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Deleg.hs
Soupstraw: We haven't really put this in the spec yet, but I think the deposit should be checked against the deposit that was actually paid, not the deposit that's currently required by PParams.

Soupstraw: ~maybe leave a TODO so we can come back to this once we've figured it out in the spec.~

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Deleg.hs
```diff
@@ -0,0 +1,255 @@
...
+        ?! UnRegCertProvidedCoinNotEqualToExpectedPoolDepositDELEG sMayCoin
+      UM.member stakeCred (RewDepUView dsUnified) 
+        ?! UnRegCertStakeKeyNotRegisteredDELEG stakeCred
+      let mRDPair = UM.lookup stakeCred (RewDepUView dsUnified)
+          mRewardCoin = UM.rdReward <$> mRDPair
+          mDepositCoin = UM.rdDeposit <$> mRDPair
+      mRewardCoin == Just mempty 
+        ?! UnRegCertStakeKeyNonZeroAccountBalanceDELEG (UM.fromCompact <$> mRewardCoin)
+      maybeToStrictMaybe (UM.fromCompact <$> mDepositCoin) == sMayCoin
```
Soupstraw: This is probably the deposit check we'll actually want to have.

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Deleg.hs
Soupstraw:
```suggestion
      sMayCoin == SNothing
```
Since you already check the deposit against the deposit that was actually paid, you don't actually have to check against the deposit required by PParams

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Deleg.hs
Soupstraw:
```suggestion
      maybeToStrictMaybe (UM.fromCompact <$> mDepositCoin) == sMayCoin || sMayCoin == SNothing
```

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Delegs.hs
```diff
@@ -169,14 +178,16 @@ conwayDelegsTransition = do
       -- It is impossible to have 65535 number of certificates in a
       -- transaction, therefore partial function is justified.
       let ptr = Ptr slot txIx (mkCertIxPartial $ toInteger $ length gamma)
-      trans @(EraRule "CERT" era) $
+      newState <- trans @(EraRule "CERT" era) $
         TRC (DelplEnv slot ptr pp acnt, certState', c)
+      trans @(EraRule "DELEG" era) $
+        TRC (DelegEnv slot ptr acnt pp, newState, c)
```
Soupstraw: This is supposed to be called from the CERT rule. It would make much more sense to rename DELEGS rule to CERTS rule instead, but this can be left for another PR.

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Deleg.hs
```diff
@@ -0,0 +1,255 @@
...
+        ?! UnRegCertStakeKeyDepositDoesNotMatchExpectedDepositDELEG (UM.fromCompact <$> mDepositCoin)
+      let umRDRemoved = Set.singleton stakeCred UM.⋪ UM.RewDepUView dsUnified
+          umSPoolRemoved = Set.singleton stakeCred UM.⋪ UM.SPoolUView umRDRemoved
+          newUMap = UM.PtrUView umSPoolRemoved UM.⋫ Set.singleton stakeCred
```
Soupstraw: we can get rid of the Ptr map here, there's no need to update it

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Deleg.hs
```diff
@@ -0,0 +1,255 @@
...
+      pure $
+        cState
+          { certDState =
+              certDState
+                { dsUnified = newUMap }
+          }
+    ConwayTxCertDeleg (ConwayDelegCert stakeCred delegatee) -> do
+      UM.member stakeCred (RewDepUView dsUnified)
+        ?! DelegCertStakeKeyNotRegisteredDELEG stakeCred
+      case delegatee of
```
Soupstraw: I would extract this as a function and then make it so that the `DelegStakeVote` case just calls the same function with `DelegStake` and `DelegVote` and then combines the results together. That way we can get rid of some code duplication.

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Deleg.hs
```diff
@@ -0,0 +1,255 @@
...
+        DelegStake sPool -> do
+          let umSPoolRemoved = Set.singleton stakeCred UM.⋪ UM.SPoolUView dsUnified
```
Soupstraw: You don't actually have to remove the stake credential, since you'll override it when you insert the new delegation

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Deleg.hs
```diff
@@ -0,0 +1,255 @@
...
+      coin == pd 
+        ?! RegDelegCertProvidedCoinNotEqualToExpectedPoolDepositDELEG coin
+      UM.notMember stakeCred (RewDepUView dsUnified) 
+        ?! RegDelegCertStakeKeyAlreadyRegisteredDELEG stakeCred
+      UM.notMember stakeCred (SPoolUView dsUnified) 
+        ?! RegDelegCertStakeKeyAlreadyRegisteredDELEG stakeCred
+      UM.notMember stakeCred (DRepUView dsUnified) 
+        ?! RegDelegCertDRepAlreadyRegisteredDELEG stakeCred
+      Set.member dRep vsDReps 
+        ?! RegDelegCertStakeVoteDRepNotRegisteredDELEG dRep
```
Soupstraw: We can get rid of this check - we don't care whether the DRep is registered or not.

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Deleg.hs
```diff
@@ -0,0 +1,255 @@
...
+              umDRepRemoved = Set.singleton stakeCred UM.⋪ UM.DRepUView umSPoolRemoved
+              newSPool = UM.SPoolUView umDRepRemoved UM.⨃ Map.singleton stakeCred sPool
+              newUMap = UM.DRepUView newSPool UM.⨃ Map.singleton stakeCred dRep
+          pure $
+            cState
+              { certDState =
+                  certDState
+                    { dsUnified = newUMap }
+              }
+    ConwayTxCertDeleg (ConwayRegDelegCert stakeCred (DelegStakeVote sPool dRep) coin) -> do
```
Soupstraw: We also need cases for the other `Delegatee`s

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Deleg.hs
```diff
@@ -0,0 +1,255 @@
...
+        ?! RegDelegCertStakeKeyAlreadyRegisteredDELEG stakeCred
+      UM.notMember stakeCred (SPoolUView dsUnified) 
+        ?! RegDelegCertStakeKeyAlreadyRegisteredDELEG stakeCred
+      UM.notMember stakeCred (DRepUView dsUnified) 
+        ?! RegDelegCertDRepAlreadyRegisteredDELEG stakeCred
+      Set.member dRep vsDReps 
+        ?! RegDelegCertStakeVoteDRepNotRegisteredDELEG dRep
+      let ppDeposit = UM.compactCoinOrError pd
+          newRDs = UM.RewDepUView dsUnified UM.∪ (stakeCred, UM.RDPair (UM.CompactCoin 0) ppDeposit)
+          newPtr = UM.PtrUView newRDs UM.∪ (ptr, stakeCred)
```
Soupstraw: We can get rid of this line
```suggestion
```

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Deleg.hs
aniketd: Noted! Thanks for pointing this out! :sweat_smile:

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Deleg.hs
aniketd: Noted! Nice catch! :slightly_smiling_face:

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Deleg.hs
```diff
@@ -0,0 +1,255 @@
...
+              }
+    ConwayTxCertDeleg (ConwayRegDelegCert stakeCred (DelegStakeVote sPool dRep) coin) -> do
+      coin == pd 
+        ?! RegDelegCertProvidedCoinNotEqualToExpectedPoolDepositDELEG coin
+      UM.notMember stakeCred (RewDepUView dsUnified) 
+        ?! RegDelegCertStakeKeyAlreadyRegisteredDELEG stakeCred
+      UM.notMember stakeCred (SPoolUView dsUnified) 
+        ?! RegDelegCertStakeKeyAlreadyRegisteredDELEG stakeCred
+      UM.notMember stakeCred (DRepUView dsUnified) 
+        ?! RegDelegCertDRepAlreadyRegisteredDELEG stakeCred
```
Soupstraw: We'll still want to allow using this cert even if the stake credential is already registered. In that case it would work just like `ConwayDelegCert`

eras/conway/impl/src/Cardano/Ledger/Conway/Era.hs
aniketd: :memo:

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Deleg.hs
aniketd: It is indeed similar to an `unless` :+1:

Soupstraw (changes requested):
There are a lot of nuances to keep track of, but you're on the right track. Keep up the good work!

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Deleg.hs
aniketd: Got it! :raised_hands:

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Delegs.hs
aniketd: Since this PR requires implementation of the `DELEGS` rule anyway, in order to embed `DELEG` under it, I have done the rename too, as it was not not much work. :raised_hands:

aniketd: Nice suggestion by the way! :+1: :smiley:

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Deleg.hs
aniketd: Great suggestion!

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Deleg.hs
aniketd: Yes! Thanks! :heart:

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Deleg.hs
aniketd: Got it, I have commented out that part after refactoring, for now. Will remove it completely after finishing up tests.

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Certs.hs
```diff
@@ -71,99 +71,107 @@ import GHC.Generics (Generic)
...
-newtype ConwayDelegsEvent era = CertEvent (Event (EraRule "CERT" era))
+newtype ConwayCertsEvent era = CertEvent (Event (EraRule "CERT" era))
 
 instance
   ( EraTx era
   , ShelleyEraTxBody era
   , State (EraRule "CERT" era) ~ CertState era
+  , State (EraRule "DELEG" era) ~ CertState era
   , Signal (EraRule "CERT" era) ~ TxCert era
+  , Signal (EraRule "DELEG" era) ~ TxCert era
```
Soupstraw:
```suggestion
```
I don't think you're using this constraint anywhere

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Certs.hs
```diff
@@ -71,99 +71,107 @@ import GHC.Generics (Generic)
...
   , Environment (EraRule "CERT" era) ~ DelplEnv era
-  , EraRule "DELEGS" era ~ ConwayDELEGS era
-  , Embed (EraRule "CERT" era) (ConwayDELEGS era)
+  , Environment (EraRule "DELEG" era) ~ DelegEnv era
```
Soupstraw, line 128:
```suggestion
```

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Certs.hs
```diff
@@ -71,99 +71,107 @@ import GHC.Generics (Generic)
...
+  , EraRule "CERTS" era ~ ConwayCERTS era
+  , Embed (EraRule "CERT" era) (ConwayCERTS era)
+  , Embed (EraRule "DELEG" era) (ConwayCERTS era)
```
Soupstraw, line 131:
```suggestion
```

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Certs.hs
Soupstraw: or any of the other `DELEG` constraints

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Pool.hs
```diff
@@ -0,0 +1,74 @@
...
+  NoThunks (PredicateFailure (EraRule "POOL" era)) =>
+  NoThunks (ConwayPoolPredFailure era)
+deriving instance
+  Eq (PredicateFailure (EraRule "POOL" era)) =>
+  Eq (ConwayPoolPredFailure era)
+deriving instance
+  Show (PredicateFailure (EraRule "POOL" era)) =>
+  Show (ConwayPoolPredFailure era)
+
+-- newtype ConwayDelegEvent era = NewEpoch EpochNo
```
Soupstraw:
```suggestion
```

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/VDel.hs
```diff
@@ -0,0 +1,76 @@
...
+  VDelEnv (..),
+  ConwayVDelPredFailure,
+)
+where
+
+import Cardano.Ledger.BaseTypes (
+  ShelleyBase,
+ )
+import Cardano.Ledger.CertState (VState)
+import Cardano.Ledger.Conway.Core ({-ShelleyEraTxBody,-} ConwayEraTxBody)
```
Soupstraw:
```suggestion
import Cardano.Ledger.Conway.Core (ConwayEraTxBody)
```

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/VDel.hs
```diff
@@ -0,0 +1,76 @@
...
+  NoThunks (PredicateFailure (EraRule "VDEL" era)) =>
+  NoThunks (ConwayVDelPredFailure era)
+deriving instance
+  Eq (PredicateFailure (EraRule "VDEL" era)) =>
+  Eq (ConwayVDelPredFailure era)
+deriving instance
+  Show (PredicateFailure (EraRule "VDEL" era)) =>
+  Show (ConwayVDelPredFailure era)
+
+-- newtype ConwayDelegEvent era = NewEpoch EpochNo
```
Soupstraw:
```suggestion
```

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Certs.hs
aniketd: Thanks for this catch!

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/Trace.hs
```diff
@@ -236,7 +236,9 @@ pcSmallUTxO proof u txs = ppMap pcTxIn (shortTxOut proof) m
 
 raiseMockError ::
   forall era.
-  (Reflect era, PrettyA (TxCert era)) =>
+  ( Reflect era
+  , PrettyA (TxCert era)
+  ) =>
```
Soupstraw, line 241:
```suggestion
  ( Reflect era, PrettyA (TxCert era)) =>
```

Soupstraw:
```suggestion
  (Reflect era, PrettyA (TxCert era)) =>
```

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/Trace.hs
```diff
@@ -565,7 +567,8 @@ chainTest proof n gsize = testProperty message action
       -- Here is where we can add some properties for traces:
       pure (_traceInitState trace1 === initState)
 
-testTraces :: Int -> TestTree
+testTraces ::
+  Int -> TestTree
```
Soupstraw:
```suggestion
testTraces :: Int -> TestTree
```

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/VDel.hs
```diff
@@ -0,0 +1,65 @@
...
+  STS (ConwayVDEL era)
+  where
+  type State (ConwayVDEL era) = VState era
+  type Signal (ConwayVDEL era) = ConwayCommitteeCert (EraCrypto era)
+  type Environment (ConwayVDEL era) = VDelEnv era
+  type BaseM (ConwayVDEL era) = ShelleyBase
+  type PredicateFailure (ConwayVDEL era) = ConwayVDelPredFailure era
+  type Event (ConwayVDEL era) = ConwayVDelEvent era
+
+  transitionRules = undefined
```
Soupstraw:
```suggestion
  transitionRules = undefined -- TODO
```

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Pool.hs
```diff
@@ -0,0 +1,59 @@
...
+  STS (ConwayPOOL era)
+  where
+  type State (ConwayPOOL era) = PState era
+  type Signal (ConwayPOOL era) = PoolCert (EraCrypto era)
+  type Environment (ConwayPOOL era) = PoolEnv era
+  type BaseM (ConwayPOOL era) = ShelleyBase
+  type PredicateFailure (ConwayPOOL era) = ConwayPoolPredFailure era
+  type Event (ConwayPOOL era) = ConwayPoolEvent era
+
+  transitionRules = undefined
```
Soupstraw:
```suggestion
  transitionRules = undefined -- TODO
```

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Deleg.hs
```diff
@@ -0,0 +1,219 @@
...
+      case delegatee of
+        DelegStake sPool -> delegStake stakeCred sPool dsUnified
+        DelegVote dRep -> delegVote stakeCred dRep dsUnified
+        DelegStakeVote sPool dRep -> delegVote stakeCred dRep $ delegStake stakeCred sPool dsUnified
+    checkDepositAgainstPParams pd sMayDeposit =
+      sMayDeposit
+        == SNothing
+        || sMayDeposit
+        == SJust pd
+        ?! IncorrectDepositDELEG sMayDeposit
```
Soupstraw:
```suggestion
    checkDepositAgainstPParams pd (SJust dep) 
      = (dep == pd) ?! IncorrectDepositDELEG sMayDeposit
    checkDepositAgainstPParams _ _ = pure ()
```
I'm not completely sure that it's correct, but I think it would be more readable when written in this style

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Deleg.hs
```diff
@@ -0,0 +1,219 @@
...
+    checkDepositAgainstPayedDeposit stakeCred dsUnified sMayDeposit =
```
Soupstraw:
```suggestion
    checkDepositAgainstPaidDeposit stakeCred dsUnified sMayDeposit =
```

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Deleg.hs
```diff
@@ -0,0 +1,219 @@
...
+      ) <-
+    judgmentContext
+  let pd = pp ^. ppKeyDepositL
+  case c of
+    ConwayRegCert stakeCred sMayDeposit -> do
+      checkDepositAgainstPParams pd sMayDeposit
+      checkStakeKeyNotAlreadyRegistered stakeCred dsUnified
+      pure $ dState {dsUnified = acceptDepositForStakeKey stakeCred dsUnified pd}
+    ConwayUnRegCert stakeCred sMayDeposit -> do
+      checkStakeKeyIsAlreadyRegistered stakeCred dsUnified
```
Soupstraw, line 138: It's also possible to only delegate to a DRep and not delegate to an SPO, so it should check that either of these delegations is present in the UMap.

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Deleg.hs
```diff
@@ -0,0 +1,219 @@
...
+    ( DelegEnv _slot _ptr _acnt pp
+      , dState@DState {dsUnified}
+      , c
+      ) <-
+    judgmentContext
+  let pd = pp ^. ppKeyDepositL
+  case c of
+    ConwayRegCert stakeCred sMayDeposit -> do
+      checkDepositAgainstPParams pd sMayDeposit
+      checkStakeKeyNotAlreadyRegistered stakeCred dsUnified
```
Soupstraw, line 135: It might also be delegating to a DRep. I'm not completely sure what we want to do about the other things it might be delegated to (e.g. Ptrs).

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Deleg.hs
aniketd: The formatter made it ugly here. :+1:  Thanks for the suggestion!

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Deleg.hs
aniketd: :slightly_smiling_face: :facepalm:

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/TxGen.hs
```diff
@@ -1125,8 +1125,11 @@ instance
 -- in a way that is Era Agnostic
 
 applySTSByProof ::
-  forall era.
-  (Era era, GoodCrypto (EraCrypto era)) =>
+  forall era c.
+  ( Era era
+  , EraCrypto era ~ c
+  , GoodCrypto (EraCrypto era)
+  ) =>
```
aniketd:
```suggestion
  forall era.
  (Era era, GoodCrypto (EraCrypto era)) =>
```

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Deleg.hs
aniketd:
```suggestion
    checkDepositAgainstPParams pd = \case
      SNothing -> pure ()
      SJust deposit -> deposit == pd ?! IncorrectDepositDELEG $ SJust deposit
```

aniketd: I updated it by using a `case` instead of multiple function heads. :+1:

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Deleg.hs
```diff
@@ -0,0 +1,214 @@
...
+      pure $ dState {dsUnified = acceptDepositForStakeKey stakeCred dsUnified pd}
+    ConwayUnRegCert stakeCred sMayDeposit -> do
+      checkStakeKeyIsAlreadyRegistered stakeCred dsUnified
+      checkStakeKeyHasZeroBalance stakeCred dsUnified
+      unless (sMayDeposit == SNothing) $ checkDepositAgainstPaidDeposit stakeCred dsUnified sMayDeposit
+      let umRDRemoved = Set.singleton stakeCred UM.⋪ UM.RewDepUView dsUnified
+          umSPoolRemoved = Set.singleton stakeCred UM.⋪ UM.SPoolUView umRDRemoved
```
Soupstraw: Is it possible to remove the whole UMap entry? We shouldn't release the deposit unless all of the memory allocated by the delegation is freed.

Soupstraw: at the moment we don't remove the DRep delegation

Soupstraw: at the very least we should remove the DRep delegation as well

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Deleg.hs
```diff
@@ -0,0 +1,214 @@
...
+          newUMap = UM.PtrUView umSPoolRemoved UM.⋫ Set.singleton stakeCred -- Although we don't care about Ptrs in Conway, we remove them when we can.
```
Soupstraw:
```suggestion
          -- Although we don't care about Ptrs in Conway, we remove them when we can.
          newUMap = UM.PtrUView umSPoolRemoved UM.⋫ Set.singleton stakeCred 
```

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Deleg.hs
```diff
@@ -0,0 +1,214 @@
...
+ )
+import qualified Data.Map.Strict as Map
+import qualified Data.Set as Set
+import GHC.Generics (Generic)
+import Lens.Micro ((^.))
+import NoThunks.Class (NoThunks)
+
+data ConwayDelegPredFailure era
+  = IncorrectDepositDELEG
+      !(StrictMaybe Coin)
```
lehins: Deposit can't be invalid if it is not provided
```suggestion
data ConwayDelegPredFailure era
  = IncorrectDepositDELEG !Coin
```

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Deleg.hs
```diff
@@ -0,0 +1,214 @@
...
+  | StakeKeyAlreadyRegisteredDELEG
+      !(Credential 'Staking (EraCrypto era))
+  | StakeKeyNotRegisteredDELEG
+      !(Credential 'Staking (EraCrypto era))
+  | StakeKeyHasNonZeroAccountBalanceDELEG
+      !(Maybe Coin)
```
lehins: Same here, having a predicate failure with non-zero balance implies that there is soma balance. It can't be Nothing, right?
```suggestion
  | StakeKeyHasNonZeroAccountBalanceDELEG
      !(Maybe Coin)
```

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Deleg.hs
```diff
@@ -0,0 +1,214 @@
...
+  let pd = pp ^. ppKeyDepositL
+  case c of
+    ConwayRegCert stakeCred sMayDeposit -> do
+      checkDepositAgainstPParams pd sMayDeposit
+      checkStakeKeyNotAlreadyRegistered stakeCred dsUnified
+      pure $ dState {dsUnified = acceptDepositForStakeKey stakeCred dsUnified pd}
+    ConwayUnRegCert stakeCred sMayDeposit -> do
+      checkStakeKeyIsAlreadyRegistered stakeCred dsUnified
+      checkStakeKeyHasZeroBalance stakeCred dsUnified
+      unless (sMayDeposit == SNothing) $ checkDepositAgainstPaidDeposit stakeCred dsUnified sMayDeposit
```
lehins: This check only applies when deposit is provided, so there is no point of passing Maybe to the function.
```suggestion
      forM_ sMayDeposit $ checkDepositAgainstPaidDeposit stakeCred dsUnified
```

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Deleg.hs
```diff
@@ -0,0 +1,214 @@
...
+  TRC
+    ( DelegEnv _slot _ptr _acnt pp
+      , dState@DState {dsUnified}
+      , c
+      ) <-
+    judgmentContext
+  let pd = pp ^. ppKeyDepositL
+  case c of
+    ConwayRegCert stakeCred sMayDeposit -> do
+      checkDepositAgainstPParams pd sMayDeposit
```
lehins: Same here. Also you aren't passing pparams to `checkDepositAgainstPParams`, but the actual expected deposit, so a better name is in line (also more descriptive `ppKeyDeposit` is better than `pd`).
```suggestion
      forM_ sMayDeposit $ checkExpectedDeposit ppKeyDeposit
```

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Deleg.hs
```diff
@@ -0,0 +1,214 @@
...
+    ConwayDelegCert stakeCred delegatee -> do
+      checkStakeKeyIsAlreadyRegistered stakeCred dsUnified
+      pure $
+        dState
+          { dsUnified = processDelegation stakeCred delegatee dsUnified
+          }
+    ConwayRegDelegCert stakeCred delegatee coin -> do
+      coin
+        == pd
+        ?! IncorrectDepositDELEG (SJust coin)
```
lehins: Putting it all on one line is a bit more readable. See my other comment about the SMaybe
```suggestion
      coin == pd ?! IncorrectDepositDELEG coin
```

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Deleg.hs
aniketd:
```haskell
-- | Delete the stake credentials in the domain and all associated ranges from the `UMap`
-- This can be expensive when there are many pointers associated with the credential.
domDeleteAll :: Set (Credential 'Staking c) -> UMap c -> UMap c
domDeleteAll ks UMap {umElems, umPtrs} =
  UMap 
    { umElems = Map.withoutKeys umElems ks
    , umPtrs = Map.filter (`Set.notMember` ks) umPtrs
    }
```

I have added this to `UMap`, and used it in this place. I hope this is what you meant, and it solves the problem. Correct me if I'm wrong, please.

lehins: I agree, it would be best to add a specialized function in the UMap module for removing credentials. Also in the name of performance and future compatibility:

```haskell
delCredUMap :: Crypto c => Credential Staking' c -> UMap c -> UMap c
```

lehins (changes requested):
Here is some feedback from me

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Deleg.hs
aniketd: Yes! :100: Thanks!

Soupstraw (approved):
Awesome work!

lehins (approved):
LGTM. I'd like to get @Soupstraw approval too.

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Deleg.hs
```diff
@@ -0,0 +1,186 @@
...
+      deposit == ppKeyDeposit ?! IncorrectDepositDELEG deposit
+    checkDepositAgainstPaidDeposit stakeCred dsUnified deposit =
+      Just deposit == fmap (UM.fromCompact . UM.rdDeposit) (UM.lookup stakeCred $ UM.RewDepUView dsUnified) ?! IncorrectDepositDELEG deposit
+    checkStakeKeyNotAlreadyRegistered stakeCred dsUnified =
+      UM.notMember stakeCred (UM.RewDepUView dsUnified) ?! StakeKeyAlreadyRegisteredDELEG stakeCred
+    checkStakeKeyIsAlreadyRegistered stakeCred dsUnified =
+      UM.member stakeCred (UM.RewDepUView dsUnified) ?! StakeKeyNotRegisteredDELEG stakeCred
+    -- checkDRepNotAlreadyRegistered stakeCred dsUnified = -- TODO: @aniketd to confirm
+    --   UM.notMember stakeCred (DRepUView dsUnified)
+    --     ?! DRepAlreadyRegisteredForStakeKeyDELEG stakeCred
```
aniketd:
```suggestion
```

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Deleg.hs
```diff
@@ -0,0 +1,186 @@
...
+    ConwayDelegCert stakeCred delegatee -> do
+      checkStakeKeyIsAlreadyRegistered stakeCred dsUnified
+      pure $
+        dState
+          { dsUnified = processDelegation stakeCred delegatee dsUnified
+          }
+    ConwayRegDelegCert stakeCred delegatee deposit -> do
+      deposit == ppKeyDeposit ?! IncorrectDepositDELEG deposit
+      checkStakeKeyNotAlreadyRegistered stakeCred dsUnified
+      -- checkDRepNotAlreadyRegistered stakeCred dsUnified -- TODO: @aniketd to confirm
```
aniketd:
```suggestion
```

PR #3421, opened 2023-05-11
Rename `DCert` -> `TxCert`

lehins:
# Description

Recently in #3387 we introduced a type family. In order to keep it consistent with all other type families that go into a transaction it was decided to rename `DCert` to a `TxCert`

Modules that contain the classes, instances and type families where also renamed to `TxCert`

There are no semantic changes in this PR, just the renames.

Also all of the versions where bumped, since there is currently ledger release happening to CHaPs.

# Checklist

- [ ] Self-reviewed the diff


eras/allegra/impl/CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Version history for `cardano-ledger-allegra`
 
+## 1.2.0.0
+
+* Introduction of `TxCert` and `EraTxCert`
+* Add `EraTxCert` and `ShelleyEraTxCert` instances to `AllegraEra`
+
```
JaredCorduan, line 7: @lehins I realize now that by releasing a bunch of ledger packages on CHaPs that were not from master, the changelogs are now a bit wrong. For example, the `ShelleyEraDCert` addition mentioned on line 14 of this changelog should go in `1.2.0.0` now, since it was not added to `1.1.1.0`.

After we merge your PR here, I can clean it all up.

JaredCorduan (approved):
Looks great, thanks for continually getting us closer and closer to uniformity!

The changelogs are going to need some help, as I mentioned in another comment, but I can do that in a follow up PR.

eras/allegra/impl/CHANGELOG.md
lehins: Yeah, that's exactly what I was trying to fixup in this PR. Guess I missed some. I did fix what mentioned here, but if you can double check all changelog files after this PR is merged that will be great

PR #3422, opened 2023-05-12
Migrate to nix flake

### LINKED ISSUE #3401: [cardano-ledger] fix nix support following cleanup done in cardano-base
- [x] clean up all legacy nix code replacing it by a flake
- [x] add in stubs for shell.nix and default.nix (same as in cardano-base).

jbgi:
waiting for
- [x] https://github.com/input-output-hk/iohk-nix/pull/543
- [ ] ideally https://github.com/numtide/flake-utils/pull/97

being merged.


flake.nix
```diff
@@ -1,50 +1,268 @@
...
+    ];
+  in
+    inputs.flake-utils.lib.eachSystem supportedSystems (
+      system: let
+        # setup our nixpkgs with the haskell.nix overlays, and the iohk-nix
+        # overlays...
+        nixpkgs = import inputs.nixpkgs {
+          overlays =
+            [inputs.haskellNix.overlay]
+            ++ builtins.attrValues inputs.iohkNix.overlays
```
angerman: We will need to be careful here to import them in the right order. See https://github.com/input-output-hk/cardano-db-sync/commit/a7cd5db34782e70b0044ecb97c70654506690dc5

flake.nix
```diff
@@ -1,50 +1,268 @@
...
+          ];
+          shell.withHoogle = false;
+
+          # package customizations as needed. Where cabal.project is not
+          # specific enough, or doesn't allow setting these.
+          modules = [
+            ({pkgs, ...}: {
+              # Use our forked libsodium from iohk-nix crypto overlay.
+              packages.cardano-crypto-class.components.library.pkgconfig = lib.mkForce [[pkgs.libsodium-vrf pkgs.secp256k1]];
+              packages.cardano-crypto-praos.components.library.pkgconfig = lib.mkForce [[pkgs.libsodium-vrf]];
```
angerman: These should not be necessary anymore. iohk-nix should inject them appropriately.
```suggestion
```

angerman (dismissed):

flake.nix
jbgi: Though if we put `iohkNix.overlays.crypto` before haskell.nix overlays then https://github.com/input-output-hk/iohk-nix/blob/86421fdd89b3af43fa716ccd07638f96c6ecd1e4/overlays/crypto/default.nix#L18-L26 does not get injected right?
working on a fix.

angerman: Ohh god. This is a bit messy 💀 loading it twice might work. But is also not ideal. Maybe this needs to be split. But in general the order dependence is rather annoying.

jbgi: fix in https://github.com/input-output-hk/iohk-nix/pull/543

.github/workflows/haskell.yml
```diff
@@ -210,6 +210,4 @@ jobs:
             trusted-public-keys = hydra.iohk.io:f/Ea+s+dFdN+3Y/G+FDgSq+a5NEWhJGzdjvKNGv0/EQ= cache.nixos.org-1:6NCHdD59X431o0gWypbMrAURkbJ16ZPMQFGspcDShjY= loony-tools:pr9m4BkM/5/eSTZlkQyRt57Jz7OMBxNSUiMC4FkcNfk=
             substituters = https://cache.iog.io https://cache.nixos.org/ https://cache.zw3rk.com
       - name: nix-build
-        run: nix-build
-      - name: nix-shell
-        run: nix-shell shell.nix
+        run: nix build .#hydraJobs.required
```
teodanciu, line 213: Why not also `nix develop` ? Is it not possible that this will fail if `nix build` doesn't?

jbgi: the `required` job does include the `nix develop` shell so it should cover `nix develop` case.
But thinking of it, this job could be more useful by checking that hydra or cicero did their job properly, by checking that everything is in cache already (no job should be built locally). Adding a commit (664ece0e47cf711d3af055f0b5d9f784a83aaead) for this.

.github/workflows/haskell.yml
```diff
@@ -197,8 +197,8 @@ jobs:
     - name: Regenerate hie.yaml and confirm that it is in sync
       run: ./scripts/gen-hie.sh
 
-  nix-build:
-    name: Build nix derivations and enter nix-shell
+  nix-cache-check:
+    name: Check nix required derivations are cached (built by hydra or cicero)
```
teodanciu: I don't understand this bit :thinking: 
At the moment, because our main CI workflow is building and running via cabal (rather than nix), we have this job that is run nightly, checking that everything still builds with nix. 
What does it mean to check that the derivations are cached?  Are they being built and cached if not already cached, and not ran if cached?   Wouldn't it make more sense, nightly, to build them without cache?

doc/README.md
```diff
@@ -21,19 +21,10 @@ The direct dependencies are specified in `requirements.in`. This file is only us
...
-      python -m venv .venv && source .venv/bin/activate
-    ```
-  * ```
-      python -m pip install pip-tools
-    ```
+* In `doc`, enter a `nix develop` shell that includes pip-tools:
   * ```
       cd doc
+      nix develop
       pip-compile requirements.in
```
teodanciu, line 28: Great, thanks for improving this by adding it to flakes file :pray:

eras/byron/ledger/executable-spec/.ghcid
```diff
@@ -1 +1 @@
---command="cabal repl --repl-options='-isrc -fwarn-unused-binds -fwarn-unused-imports -fno-code -fobject-code -g2 -fno-break-on-exception -fno-break-on-error -ferror-spans -j -Wno-unused-packages'" --clear --no-height-limit --reverse-errors --reload=../../../../ --outputfile=/tmp/byron-spec-ledger-ghcid.txt
\ No newline at end of file
+--command="cabal repl --repl-options='-isrc -fwarn-unused-binds -fwarn-unused-imports -fno-code -fobject-code -g2 -fno-break-on-exception -fno-break-on-error -ferror-spans -j -Wno-unused-packages'" --clear --no-height-limit --reverse-errors --reload=../../../../ --outputfile=/tmp/byron-spec-ledger-ghcid.tx
```
teodanciu: Missing a `t` at the end?  `/tmp/byron-spec-ledger-ghcid.txt`

flake.nix
```diff
@@ -1,50 +1,290 @@
...
+          #
+          inputMap = {
+            "https://input-output-hk.github.io/cardano-haskell-packages" = inputs.CHaP;
+          };
+
+          # tools we want in our shell, from hackage
+          shell.tools = {
+            cabal = "3.10.1.0";
+            ghcid = "0.8.8";
+            haskell-language-server = "latest";
```
teodanciu: Is it better to use `latest` for the haskell-language-server version ?  I see before it was set to `1.9.1.0`

flake.nix
```diff
@@ -1,50 +1,290 @@
...
+            hlint = {};
+            fourmolu = "0.12.0.0";
```
teodanciu: Maybe it would be better to keep in this PR the version of formolu as it was (`0.10.1.0`), because this new version seems to trigger some reformattings on the current code. Perhaps better if I do this separately, both the upgrade and the reformatting?

flake.nix
```diff
@@ -1,50 +1,290 @@
...
+          inherit (inputs.haskellNix) config;
+        };
+        inherit (nixpkgs) lib;
+
+        # We use cabalProject' to ensure we don't build the plan for
+        # all systems.
+        cabalProject = nixpkgs.haskell-nix.cabalProject' {
+          src = ./.;
+          name = "cardano-ledger";
+          compiler-nix-name = lib.mkDefault "ghc928";
```
teodanciu: I think this was ghc927 before - is this change on purpose?

flake.nix
```diff
@@ -1,50 +1,290 @@
...
+              nix-systems
+            ]);
+          # disable Hoogle until someone request it
+          shell.withHoogle = false;
+
+          # package customizations as needed. Where cabal.project is not
+          # specific enough, or doesn't allow setting these.
+          modules = [
+            ({pkgs, ...}: {
+              packages.byron-spec-chain.configureFlags = ["--ghc-option=-Werror"];
```
teodanciu, line 105: I wonder if this (from haskell.nix) is no longer necessary:
`        packages.cardano-crypto-class.components.library.pkgconfig = lib.mkForce [ [ pkgs.libsodium-vrf pkgs.secp256k1 ] ];
`
And the same for  ` packages.cardano-crypto-praos.`

CONTRIBUTING.md
teodanciu, line 1: Is the section in this file about updating dependencies using niv still applicable? https://github.com/input-output-hk/cardano-ledger/pull/3422/files#diff-eca12c0a30e25b4b46522ebf89465a03ba72a03f540796c979137931d8f92055R62

Or maybe it should be `nix flake update` instead?

flake.nix
jbgi: yes it is no longer necessary, that configuration is now global via the `haskell-nix-crypto` overlay: https://github.com/input-output-hk/iohk-nix/blob/02f42375ee5c2bab1640f14c6389b7e91bbfec8b/overlays/haskell-nix-crypto/default.nix#L9-L13

flake.nix
jbgi: I'm going to rollback to original value, and let you update in separate PR if needed.
I thing `latest` is often better when updating `haskellNix` as whole. But when the usual process is to only update `hackage` (as currently documented in CONTRIBUTING) then using a fixed version is probably better to avoid breakage.

flake.nix
jbgi: that was to fix a warning, but probably better to update in a separate PR.

CONTRIBUTING.md
jbgi: indeed

.github/workflows/haskell.yml
jbgi: Now we have both cicero and hydra runing on PR and master branch so the idea was to check that they are indeed building master branch, and populating the cache, correctly. But now I understand that the aim was more to have easy access to build logs in case of nix shell failing. So I'll remove that commit.

flake.nix
jbgi: tried `1.9.1.0` but looks like it does not compile with ghc 9.2.7, so switch to 2.0.0.0.

flake.nix
teodanciu: Good to know, thank you

teodanciu:
Hm,  I can't seem to be able to build with cabal from within the shell. I think this is the standard way of working locally with the code base - entering a shell and then building with cabal. 

```
cabal build all
Resolving dependencies...
Error: cabal: Could not resolve dependencies:
[__0] trying: byron-spec-chain-1.0.0.0 (user goal)
[__1] trying: base-4.16.4.0/installed-4.16.4.0 (dependency of
byron-spec-chain)
[__2] trying: byron-spec-ledger-1.0.0.0 (user goal)
[__3] next goal: cardano-crypto-class (dependency of byron-spec-ledger)
[__3] rejecting: cardano-crypto-class-2.1.2.0 (conflict: pkg-config package
libblst-any, not found in the pkg-config database)
[__3] rejecting: cardano-crypto-class-2.1.1.0, cardano-crypto-class-2.1.0.2,
cardano-crypto-class-2.1.0.1, cardano-crypto-class-2.1.0.0,
cardano-crypto-class-2.0.0.1 (conflict: pkg-config package libsodium-any, not
found in the pkg-config database)
[__3] rejecting: cardano-crypto-class-2.0.0.0.1 (conflict:
base==4.16.4.0/installed-4.16.4.0, cardano-crypto-class => base>=4.14 &&
<4.15)
[__3] skipping: cardano-crypto-class-2.0.0 (has the same characteristics that
caused the previous version to fail: excludes 'base' version 4.16.4.0)
[__3] fail (backjumping, conflict set: base, byron-spec-ledger,
cardano-crypto-class)
After searching the rest of the dependency tree exhaustively, these were the
goals I've had most trouble fulfilling
```

I tried: cabal clean, update, removing caches.  It does work with [the commit before this branch](https://github.com/input-output-hk/cardano-ledger/commit/6ae2dc37b576a18afa288ea5df38e9a7440bccee)  (`b2fc5d0e4: Make it build with ghc-9.6`) 
Do we need to update some index state? But it's strange, since this migration doesn't seem to change any dependencies, no?

jbgi:
@teodanciu oh, I accidentally committed a debug line. removed in https://github.com/input-output-hk/cardano-ledger/compare/9af52357cc2da3f06ce08ddfa52a30b2d92f891a..7b9c1c1001a7dae7de773bac3e1ab1f426d3e7d0#diff-206b9ce276ab5971a2489d75eb1b12999d4bf3843b7988cbe8d687cfde61dea0L74
(also remove windows cross-compilation support from the shell, it was causing compilation issues)

teodanciu:
@jbgi thanks! 
Do you have an idea how to make vscode work with flakes?  Until now,  `Nix Environment Selector` plugin in vscode was working well with shell.nix, but it doesn't seem to work with flakes. 
Have you seen a way of flakes working with vscode?

teodanciu:
@jbgi Nevermind, I actually managed to make it work with direnv, sorry for the noise.

teodanciu (approved):
Looks good to me, I also checked the other workflows (like push-specs) on my fork, and it's all working. Thank you!

jbgi:
> @jbgi Nevermind, I actually managed to make it work with direnv, sorry for the noise.

maybe a (your) `.envrc` could be added to the repo?

PR #3423, opened 2023-05-12
Rename ShelleyDelegCerts constructors

lehins:
# Description

Rename `ShelleyDelegCert` constructors so they:
* Match `ConwayDelegCerts` better
* Avoid misleading `Key` suffix, where it actually contains credential
* Deprecate `Delegation` type that brings nothing but mental overhead.

In summary. Deprecate:
  * `RegKey` in favor of `ShelleyRegCert`
  * `DeRegKey` in favor of `ShelleyUnRegCert`
  * `Delegate` in favor of `ShelleyDelegCert`
  * `Delegation`

# Checklist

- [ ] Any changes are noted in the latest section of a `CHANGELOG.md` for the affected packages. **New section is never added with the code changes.** (See [RELEASING.md](https://github.com/input-output-hk/cardano-ledger/blob/master/RELEASING.md#changelogmd))
- [ ] Self-reviewed the diff

### DIFF
--- a/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxInfo.hs
+++ b/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxInfo.hs
@@ -6,6 +6,7 @@
 {-# LANGUAGE GADTs #-}
 {-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE MultiParamTypeClasses #-}
+{-# LANGUAGE NamedFieldPuns #-}
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE PolyKinds #-}
 {-# LANGUAGE ScopedTypeVariables #-}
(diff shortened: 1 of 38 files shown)


eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxInfo.hs
```diff
@@ -416,22 +417,21 @@ instance Crypto c => EraPlutusContext 'PlutusV1 (AlonzoEra c) where
...
+      ShelleyUnRegCert stakeCred ->
+        PV1.DCertDelegDeRegKey (PV1.StakingHash (transStakeCred stakeCred))
+      ShelleyDelegCert stakeCred keyHash ->
+        PV1.DCertDelegDelegate (PV1.StakingHash (transStakeCred stakeCred)) (transKeyHash keyHash)
+  ShelleyTxCertPool (RegPool PoolParams {ppId, ppVrf}) ->
+    PV1.DCertPoolRegister (transKeyHash ppId) (PV1.PubKeyHash (PV1.toBuiltin (transHash ppVrf)))
+  ShelleyTxCertPool (RetirePool keyHash (EpochNo i)) ->
+    PV1.DCertPoolRetire (transKeyHash keyHash) (fromIntegral i)
+  ShelleyTxCertGenesis _ -> PV1.DCertGenesis
+  ShelleyTxCertMir _ -> PV1.DCertMir
```
teodanciu, line 434: Easier to read :+1:

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Reports.hs
```diff
@@ -50,9 +50,9 @@ showCred (KeyHashObj (KeyHash x)) = show x
 
 synopsisCert :: ShelleyEraTxCert era => TxCert era -> String
 synopsisCert x = case x of
-  ShelleyTxCertDeleg (RegKey cred) -> "RegKey " ++ take 10 (showCred cred)
-  ShelleyTxCertDeleg (DeRegKey cred) -> "DeRegKey " ++ take 10 (showCred cred)
-  ShelleyTxCertDeleg (Delegate _) -> "Delegation"
```
teodanciu, line 55: Wonder why there was this difference in this case, of not showing the first 10, any idea?

eras/shelley/impl/src/Cardano/Ledger/Shelley/TxCert.hs
```diff
@@ -211,12 +212,12 @@ instance Era era => EncCBOR (ShelleyTxCert era) where
 
 encodeShelleyDelegCert :: Crypto c => ShelleyDelegCert c -> Encoding
 encodeShelleyDelegCert = \case
-  RegKey cred ->
+  ShelleyRegCert cred ->
     encodeListLen 2 <> encodeWord8 0 <> encCBOR cred
-  DeRegKey cred ->
+  ShelleyUnRegCert cred ->
```
teodanciu, line 217: I think `UnReg` is better than `DeReg`, since `De` is also the start of `Delegation` , but there's no ambiguity with `UnReg`   
(I guess unless you are reading this in French as "A Registration" :laughing:  )

eras/shelley/impl/src/Cardano/Ledger/Shelley/TxCert.hs
```diff
@@ -301,38 +302,58 @@ commonTxCertDecoder = \case
...
 
 data ShelleyDelegCert c
-  = -- | A stake key registration certificate.
-    RegKey !(StakeCredential c)
-  | -- | A stake key deregistration certificate.
-    DeRegKey !(StakeCredential c)
+  = -- | A stake credential registration certificate.
+    ShelleyRegCert !(StakeCredential c)
+  | -- | A stake credential deregistration certificate.
+    ShelleyUnRegCert !(StakeCredential c)
```
teodanciu, line 308: I think `UnReg` is better than `DeReg`, since `De` is also the start of `Delegation` , but there's no ambiguity with `UnReg`   
(I guess unless you are reading this in French as "A Registration" :laughing:  )

eras/shelley/impl/src/Cardano/Ledger/Shelley/UTxO.hs
```diff
@@ -67,11 +68,10 @@ scriptStakeCred ::
...
   Maybe (ScriptHash (EraCrypto era))
-scriptStakeCred (ShelleyTxCertDeleg (DeRegKey (KeyHashObj _))) = Nothing
-scriptStakeCred (ShelleyTxCertDeleg (DeRegKey (ScriptHashObj hs))) = Just hs
-scriptStakeCred (ShelleyTxCertDeleg (Delegate (Delegation (KeyHashObj _) _))) = Nothing
-scriptStakeCred (ShelleyTxCertDeleg (Delegate (Delegation (ScriptHashObj hs) _))) = Just hs
-scriptStakeCred _ = Nothing
+scriptStakeCred = \case
+  ShelleyTxCertDeleg (ShelleyUnRegCert cred) -> scriptCred cred
+  ShelleyTxCertDeleg (ShelleyDelegCert cred _) -> scriptCred cred
+  _ -> Nothing
```
teodanciu, line 77: Wouldn't it be better to explicitly mention the other case instead of the catchall, in case a new constructor will be added, so we have to explicitly define the behaviour for it?

teodanciu (approved):
Looks good to me
(and also nice parentheses purge)

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Reports.hs
lehins: Probably oversight. This is only used in tests when assertions fail.

JaredCorduan (approved):
LGTM

eras/shelley/impl/src/Cardano/Ledger/Shelley/UTxO.hs
lehins: Those types are unlikely to change the number of constructors, but it doesn't hurt to do it anyhow. I changed it to:
```diff
-  ShelleyTxCertDeleg (ShelleyUnRegCert cred) -> scriptCred cred
-  ShelleyTxCertDeleg (ShelleyDelegCert cred _) -> scriptCred cred
+  ShelleyTxCertDeleg delegCert ->
+    case delegCert of
+      ShelleyRegCert _ -> Nothing
+      ShelleyUnRegCert cred -> scriptCred cred
+      ShelleyDelegCert cred _ -> scriptCred cred
```

PR #3425, opened 2023-05-13
Improve witness needed logic

lehins:
# Description

## Abstract deciding which certificates require witnesses.

Certificates structure has changed in Conway, because of that old logic,
even with backwards compatibility translation, will not work correctly.
In particular, some new certificates did not require a witness, where
they should have required one. Abstracting the logic by adding
`getVKeyWitnessTxCert` and `getScriptWitnessTxCert` to `EraTxCert` we
can solve this problem with different implemnetation on per era basis.
This also opens up an opportunitiy for deduplication implemented in
a subsequent commit.

## Make registration certificates with deposits require witnesses

It was discusses during a meeting that it was an unfortunate requirement
in Shelley to not enforce a witness for staking credential registration.
It was implemented to allow ITIN rewards, which is not a feature anymore.
Therefore this is no longer a necessary requirement and it was decided
to change this by enforcing a wtiness for registration certificates, but
only for the ones that contain a deposit. This would mean that we can
keep backwards compatibility throughout Conway, but in the next era it
will become a proper requirement.

## Reduce duplication by consolidatiing two versions of witsVKeyNeeded

Implementations in Shelley and Alonzo were almost identical.

## Avoid usage of partial functions

In three different places `error` call was removed by improving the logic.

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/UTxO.hs
+++ b/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/UTxO.hs
@@ -2,9 +2,9 @@
 {-# LANGUAGE DataKinds #-}
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE FlexibleInstances #-}
-{-# LANGUAGE PatternSynonyms #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE StandaloneDeriving #-}
+{-# LANGUAGE TupleSections #-}
 {-# LANGUAGE TypeFamilies #-}
 {-# LANGUAGE UndecidableInstances #-}
 {-# OPTIONS_GHC -Wno-orphans #-}
@@ -22,17 +22,11 @@ import Cardano.Ledger.Alonzo.Scripts.Data (Datum (..))
 import Cardano.Ledger.Alonzo.Tx (ScriptPurpose (..), isTwoPhaseScriptAddressFromMap)
 import Cardano.Ledger.Alonzo.TxBody (AlonzoEraTxOut (..), MaryEraTxBody (..))
 import Cardano.Ledger.Core
-import Cardano.Ledger.Credential (Credential (ScriptHashObj))
+import Cardano.Ledger.Credential (credScriptHash)
 import Cardano.Ledger.Crypto
 import Cardano.Ledger.Mary.UTxO (getConsumedMaryValue)
 import Cardano.Ledger.Mary.Value (PolicyID (..))
 import Cardano.Ledger.Shelley.TxBody (Withdrawals (..), getRwdCred)
-import Cardano.Ledger.Shelley.TxCert (
-  ShelleyDelegCert (..),
-  ShelleyEraTxCert,
-  pattern ShelleyTxCertDeleg,
- )
-import Cardano.Ledger.Shelley.UTxO (scriptCred)
 import Cardano.Ledger.TxIn
 import Cardano.Ledger.UTxO (
   EraUTxO (..),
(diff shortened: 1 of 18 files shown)


eras/alonzo/impl/src/Cardano/Ledger/Alonzo/UTxO.hs
```diff
@@ -152,22 +146,11 @@ getAlonzoScriptsNeeded (UTxO u) txBody =
       where
         withdrawals = unWithdrawals $ txBody ^. withdrawalsTxBodyL
         fromRwd accnt = do
-          hash <- scriptCred $ getRwdCred accnt
+          hash <- credScriptHash $ getRwdCred accnt
```
JaredCorduan, line 149: how about `scriptCredHash`? since it is the hash of a script credential.

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Utxow.hs
```diff
@@ -454,14 +447,14 @@ validateVerifiedWits tx =
...
+  (EraTx era, ShelleyEraTxBody era) =>
   GenDelegs (EraCrypto era) ->
   UTxO era ->
   Tx era ->
   Set (KeyHash 'Witness (EraCrypto era)) ->
   Test (ShelleyUtxowPredFailure era)
-validateNeededWitnesses witsvkeyneeded genDelegs utxo tx witsKeyHashes =
-  let needed = witsvkeyneeded utxo tx genDelegs
+validateNeededWitnesses genDelegs utxo tx witsKeyHashes =
+  let needed = witsVKeyNeeded utxo tx genDelegs
```
JaredCorduan, line 458: nice reduction!

JaredCorduan:
nice clean up!

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/UTxO.hs
lehins: That's right, that's why `credScriptHash` is better, because it is a `ScriptHash`, not a `CredHash` :smile: 

Also this way it is consistent with `credKeyHashWitness`, which is quite similar, but for the `KeyHash`es

JaredCorduan: :laughing: yep, makes sense now.

eras/mary/impl/src/Cardano/Ledger/Mary/TxCert.hs
```diff
@@ -5,13 +5,17 @@ module Cardano.Ledger.Mary.TxCert () where
 
 import Cardano.Ledger.Crypto (Crypto, StandardCrypto)
 import Cardano.Ledger.Mary.Era (MaryEra)
-import Cardano.Ledger.Shelley.TxCert (EraTxCert (..), ShelleyEraTxCert (..), ShelleyTxCert (..))
+import Cardano.Ledger.Shelley.TxCert
```
JaredCorduan: non-Core open import?

JaredCorduan (approved):
Looks great, much cleaner and safer!

eras/mary/impl/src/Cardano/Ledger/Mary/TxCert.hs
lehins: There are three imports in this module. Only one of them is open, which means that everything that is not imported explicitly comes from `Cardano.Ledger.Shelley.TxCert` (aside from Prelude of course)

That being said, I know how much you like those explicit imports, so I'll add it back.

JaredCorduan: It's true I don't like them. but more I was just hoping for a consistent rule about when we use them. I can be flexible too :)

PR #3428, opened 2023-05-15
Get rid of Constitutional in favor of Genesis and Committee

lehins:
# Description

Constitutional committee certificate turns out to be a bit different from genesis delegation certificate:

* Genesis has a VRF field, which makes no sense for committee certificate
* Hashes in committee certificates will have different roles
* ~Committee certificates have credentials?~ Spec is a bit outdated and it was decided not to allow scripts in committee certificates
* Committee certificates are split in two: register and unregister, while genesis delegation, could not be undelegated

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/eras/conway/test-suite/cddl-files/conway.cddl
+++ b/eras/conway/test-suite/cddl-files/conway.cddl
@@ -258,7 +258,6 @@ certificate =
   // stake_delegation
   // pool_registration
   // pool_retirement
-  // constitutional_deleg_cert
   // reg_cert
   // unreg_cert
   // vote_deleg_cert
@@ -266,6 +265,8 @@ certificate =
   // stake_reg_deleg_cert
   // vote_reg_deleg_cert
   // stake_vote_reg_deleg_cert
+  // reg_committee_hot_key_cert
+  // unreg_committee_hot_key_cert
   ]
 
 stake_registration = (0, stake_credential)
@@ -273,8 +274,8 @@ stake_deregistration = (1, stake_credential)
 stake_delegation = (2, stake_credential, pool_keyhash)
 pool_registration = (3, pool_params)
 pool_retirement = (4, pool_keyhash, epoch)
-constitutional_deleg_cert = (5, genesishash, genesis_delegate_hash, vrf_keyhash)
-; number 6 used to be the MIR certificate, which is deprecated in Conway
+; numbers 5 and 6 used to be the Genesis and MIR certificates respectively,
+; which were deprecated in Conway
 reg_cert = (7, stake_credential, coin)
 unreg_cert = (8, stake_credential, coin)
 vote_deleg_cert = (9, stake_credential, voting_credential)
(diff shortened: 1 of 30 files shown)


eras/conway/test-suite/cddl-files/conway.cddl
```diff
@@ -258,30 +258,34 @@ certificate =
...
+; numbers 5 and 6 used to be the Genesis and MIR certificates respectively,
+; which were deprecated in Conway
 reg_cert = (7, stake_credential, coin)
 unreg_cert = (8, stake_credential, coin)
 vote_deleg_cert = (9, stake_credential, voting_credential)
 stake_vote_deleg_cert = (10, stake_credential, pool_keyhash, voting_credential)
 stake_reg_deleg_cert = (11, stake_credential, pool_keyhash, coin)
 vote_reg_deleg_cert = (12, stake_credential, voting_credential, coin)
 stake_vote_reg_deleg_cert = (13, stake_credential, pool_keyhash, voting_credential, coin)
+reg_committee_cert = (14, credential, addr_keyhash)
```
JaredCorduan: We can double check with @WhatisRT about his expectations, but I do not think we want a registration cert for the constitutional committee. The come into existence by a governance action. Unregistering, however, does make sense, we want the members to be able to step down at will.

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/TxGen.hs
```diff
@@ -824,7 +824,7 @@ getConwayTxCertCredential (ConwayTxCertDeleg (ConwayRegCert _ _)) = Nothing
 getConwayTxCertCredential (ConwayTxCertDeleg (ConwayUnRegCert cred _)) = Just cred
 getConwayTxCertCredential (ConwayTxCertDeleg (ConwayDelegCert cred _)) = Just cred
 getConwayTxCertCredential (ConwayTxCertDeleg (ConwayRegDelegCert cred _ _)) = Just cred
-getConwayTxCertCredential (ConwayTxCertConstitutional _) = Nothing
+getConwayTxCertCredential (ConwayTxCertCommittee _) = Nothing -- TODO Get the credential
```
teodanciu: Would it be better to have `error "Unimplemented"` here instead so we don't forget about it?

lehins: Yeah, `error` would be better, but I want to implement it in this PR. I just need to ask @WhatisRT some questions about this first.

aniketd (approved):
Noted! Thanks! :raised_hands: :memo: :blush:

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/TxGen.hs
lehins: As I suspected this case isn't relevant for `ConwayTxCertCommittee`, removed.

eras/conway/test-suite/cddl-files/conway.cddl
lehins: As far as I understand this, it is registration of the hot key, not the constitutional committee.

So, you can register and unregister the hot key, hence two certificates.

In any case semantics of the types will be defined by the rules, these are just types, which are translated from Agda spec and is called `ccreghot`:
https://github.com/input-output-hk/formal-ledger-specifications/blob/fb3722bfc945b5a2711fbc247f9e54fc73d4f326/src/Ledger/Deleg.lagda#L43

Let me know if you can suggest better names, maybe that is what makes this misleading?

Maybe `[de]reg_committee_hot_key_cert`?

JaredCorduan: ah! yes, makes total sense new. I do like `reg_committee_hot_key_cert`, that's nice and clear.

JaredCorduan (approved):

PR #3430, opened 2023-05-18
Switch to Plutus version 1.6

erikd:
No code changes needed.


cabal.project
```diff
@@ -12,14 +12,10 @@ repository cardano-haskell-packages
...
 -- update either of these.
--- NOTE: repeating the index-state for Hackage is a temporary workaround
--- for hackage.nix limited parsing ability.
--- Bump both the following dates if you need newer packages from Hackage
-index-state: 2023-04-28T00:00:00Z
 index-state:
   , hackage.haskell.org 2023-04-28T00:00:00Z
 -- Bump this if you need newer packages from CHaP
-  , cardano-haskell-packages 2023-04-27T14:03:58Z
+  , cardano-haskell-packages 2023-05-09T09:56:38Z
```
lehins, line 18: Updating `index-state` in ledger requires an update to sources. See [Updating dependencies](https://github.com/input-output-hk/cardano-ledger/blob/master/CONTRIBUTING.md#updating-dependencies) section of `CONTRIBUTING.md` doc
```shell
$ niv update cardano-haskell-packages
```

lehins (changes requested):
Nix sources need to be updated, otherwise looks great.

cabal.project
erikd: I think this really needs to be a CI test and when it fails it should print the commands that need to be run. I do not regularly use Nix, and I could easily get it wrong.

lehins: > I think this really needs to be a CI test

It is, except nix (cicero/hydra) CI has been quite unreliable, so that's why they aren't required. Hopefully the state of affairs will change in the future, in which case failure to updated flakes will be rejected by CI.

erikd: Done

lehins (approved):
Thank you!

PR #3457, opened 2023-05-25
updatet hackage index

JaredCorduan:
# Description

Updatet hackage index and set `cborg >=0.2.9 in` `cardano-ledger-binary`. Also remove the cborg SRP.

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -12,9 +12,9 @@ repository cardano-haskell-packages
 
 -- See CONTRIBUTING for some Nix commands you will need to run if you
 -- update either of these.
-index-state: 2023-04-28T00:00:00Z
+index-state: 2023-05-25T00:00:00Z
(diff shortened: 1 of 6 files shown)


lehins:
Looks like there is a new version of https://hackage.haskell.org/package/ansi-wl-pprint which is causing trouble

You could just set upper bound `ansi-wl-pprint < 1.0` for now and we can handle it later

That package is marked as deprecated anyways, so we might wanna get rid of

JaredCorduan:
> You could just set upper bound `ansi-wl-pprint < 1.0` for now and we can handle it later

will do, thanks!

lehins (approved):
:+1:

lehins (approved):
:+1:

PR #3459, opened 2023-05-26
Start API for consensus queries

lehins:
# Description

Create a module in the Ledger API for queries and a testlib with older implementations for testing and benchmarking.

This PR implements the actual `GetFilteredDelegationsAndRewardAccounts` query in `cardano-ledger-api` to be used by consensus at a later point.

This is related to [consensus](https://github.com/input-output-hk/ouroboros-consensus/pull/104)

Also reduces duplication with ledger-state package and switches the package to the latest Babbage era.

# Checklist

- [x] Self-reviewed the diff


libs/cardano-ledger-core/src/Cardano/Ledger/UMap.hs
```diff
@@ -510,50 +529,74 @@ unUnifyToVMap uview = case uview of
...
+domRestrictedStakeCredentials setk UMap {umElems, umPtrs} =
+  let umElems' = Map.restrictKeys umElems setk
+      ptrs = Map.mapMaybe umElemPtrs umElems'
+   in StakeCredentials
+        { scRewards = Map.mapMaybe (\e -> fromCompact . rdReward <$> umElemRDPair e) umElems'
+        , scDeposits = Map.mapMaybe (\e -> fromCompact . rdDeposit <$> umElemRDPair e) umElems'
+        , scSPools = Map.mapMaybe umElemSPool umElems'
+        , scDReps = Map.mapMaybe umElemDRep umElems'
+        , scPtrs = umPtrs `Map.restrictKeys` fold ptrs
+        , scPtrsInverse = ptrs
```
aniketd, line 597: This got me thinking that there is a `revPtrMap` somewhere in `UMap`, that should perhaps be called `invPtrMap` to keep it consistent between using `reverse` or `inverse` as the word to describe the map in the opposite direction? Just a thought.

aniketd: I was also wondering whether `UMap.domRestrict` could be used for constructing these. :thinking:

libs/cardano-ledger-core/testlib/Test/Cardano/Ledger/Core/Arbitrary.hs
```diff
@@ -531,7 +543,33 @@ genValidUMapNonEmpty = do
...
+  Set k ->
+  g ->
+  m (Set k)
+uniformSubset mSubSetSize inputSet gen = do
+  subSetSize <- case mSubSetSize of
+    Nothing -> uniformRM (0, Set.size inputSet) gen
+    Just n -> pure $ max 0 $ min (Set.size inputSet) n
+  go inputSet Set.empty subSetSize
+  where
+    go !s !acc !i
```
aniketd, line 560: :+1: :raised_hands: :slightly_smiling_face:

aniketd (approved):
Awesome! :raised_hands: :bow:

libs/cardano-ledger-core/src/Cardano/Ledger/UMap.hs
lehins: Good idea about renaming to `invPtrMap`. Reverse implies ordering, which isn't applicable here.

I don't think `domRestrict` is applicable to this problem.

The deal about this domain restriction in `domRestrictedStakeCredentials` is to do it only once, because it is an expensive computation, which `domRestrict` would not allow us to avoid. Also `domRestrict` requires a Map and here we have a Set.

libs/cardano-ledger-api/cardano-ledger-api.cabal
```diff
@@ -62,13 +63,29 @@ library
         plutus-ledger-api,
         text
 
+library testlib
+    exposed-modules:  Test.Cardano.Ledger.Api.State.Query
```
JaredCorduan, line 67: :tada:

JaredCorduan (approved):
looks great! thanks for getting this going!!

PR #3461, opened 2023-05-31
fix ShelleyGenesis' NoThunks instance to allow thunks inside the sgInitialFunds and sgStaking fields

fraser-iohk:
this was causing consensus' test suite to fail when built with strict-stm's checktvarinvariant flag enabled, as consensus has an invariant that all mutable cells must contain no unforced thunks (via NoThunks)

# Description

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/eras/shelley/impl/src/Cardano/Ledger/Shelley/Genesis.hs
+++ b/eras/shelley/impl/src/Cardano/Ledger/Shelley/Genesis.hs
@@ -101,7 +101,7 @@ import Data.Time.Clock (
 import Data.Unit.Strict (forceElemsToWHNF)
 import Data.Word (Word32, Word64)
 import GHC.Generics (Generic)
-import NoThunks.Class (NoThunks (..), AllowThunksIn(..))
+import NoThunks.Class (AllowThunksIn (..), NoThunks (..))
 
 -- | Genesis Shelley staking configuration.
 --
@@ -221,8 +221,10 @@ deriving instance Crypto c => Show (ShelleyGenesis c)
 
 deriving instance Crypto c => Eq (ShelleyGenesis c)
 
-deriving via AllowThunksIn '["sgInitialFunds", "sgStaking"] (ShelleyGenesis c)
-  instance Crypto c => NoThunks (ShelleyGenesis c)
+deriving via
+  AllowThunksIn '["sgInitialFunds", "sgStaking"] (ShelleyGenesis c)
+  instance
+    Crypto c => NoThunks (ShelleyGenesis c)
 
 sgActiveSlotCoeff :: ShelleyGenesis c -> ActiveSlotCoeff
 sgActiveSlotCoeff = mkActiveSlotCoeff . sgActiveSlotsCoeff
--- a/eras/shelley/impl/src/Cardano/Ledger/Shelley/Genesis.hs
+++ b/eras/shelley/impl/src/Cardano/Ledger/Shelley/Genesis.hs
@@ -3,6 +3,7 @@
 {-# LANGUAGE DeriveAnyClass #-}
 {-# LANGUAGE DeriveGeneric #-}
 {-# LANGUAGE DerivingStrategies #-}
+{-# LANGUAGE DerivingVia #-}
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE FlexibleInstances #-}
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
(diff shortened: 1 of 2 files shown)


fraser-iohk:
running `scripts/fourmolize.sh` makes a large number of changes in code that I haven't touched (or even gone near), so I intentionally have not committed those changes

JaredCorduan:
> running `scripts/fourmolize.sh` makes a large number of changes in code that I haven't touched (or even gone near), so I intentionally have not committed those changes

I wonder if you have a different version of fourmolu installed locally. if you run it from the nix shell, you should not see it make any changes.

fraser-iohk:
> I wonder if you have a different version of fourmolu installed locally. if you run it from the nix shell, you should not see it make any changes.

good idea, I'll give that a shot now!

JaredCorduan:
great catch, thank you for the help! I'm happy to approve once you've moved it out of draft and think that this solves the failing no thunks tests.

eras/shelley/impl/src/Cardano/Ledger/Shelley/Genesis.hs
```diff
@@ -207,7 +207,7 @@ data ShelleyGenesis c = ShelleyGenesis
   , sgMaxLovelaceSupply :: !Word64
   , sgProtocolParams :: !(PParams (ShelleyEra c))
   , sgGenDelegs :: !(Map (KeyHash 'Genesis c) (GenDelegPair c))
-  , sgInitialFunds :: LM.ListMap (Addr c) Coin
+  , sgInitialFunds :: !(LM.ListMap (Addr c) Coin)
```
lehins: This field is required to be lazy. It is crucial for benchmarking, otherwise they run out of memory when loading initial funds. It would be good to leave comments for fields like that. (We'll appreciate if you can add those :slightly_smiling_face: )
```suggestion
  , sgInitialFunds :: LM.ListMap (Addr c) Coin
  -- ^ This field is lazy on purpose. Needed for benchmarking and testing.
```

In order to make nothunks tests to pass the NoThunks instance need to be adjusted to allow thunks in the last two fields

lehins (changes requested):
Thank you for catching the problem, but unfortunately the solution is invalid. See my comment

eras/shelley/impl/src/Cardano/Ledger/Shelley/Genesis.hs
fraser-iohk: I'll add comments noting the intentional laziness in the `sgInitialFunds` and `sgStaking` fields -- would you be happy for me to add an explicit but redundant `~` annotation to both of those fields too, or is that overkill?

lehins: I wish we could do that, unfortunately lazy annotation is an error without `StrictData`. Except we don't want to use `StrictData`, since it can lead to more unexpected bugs than being lazy by default. Here is what you'll get if you try do: `data Foo a = Foo ~a`

```haskell
error: …
    • Lazy annotation (~) without StrictData
```

JaredCorduan: thanks for catching this @lehins , I wouldn't have realized the problem

fraser-iohk: ah, of course! I'm very much used to having `StrictData` on all the time lol

shame that the discussions around making GHC allow `~` without `StrictData` never went anywhere

fraser-iohk:
@lehins is this something that you think would benefit from additional tests? if so, where would be the best place for them?

fraser-iohk:
also, please let me know if there's an issue with the changelog entry, because I'm not sure I understood the section in `CONTRIBUTING.md`

lehins:
> is this something that you think would benefit from additional tests?

@fraser-iohk Do you mean the nothunks test, or test for ensuring that those fields stay lazy?
If it is the former than we have some nothunks tests, but I am not sure where they are, you can ask @teodanciu about. Nothunks tests are more of a consensus things anyways, but I would not aoppose adding more in ledger as well. If it is the latter however, than a comment will suffice, no tests are needed for laziness annotation.

Changelog entry indeed has a problem. If you look at the checklist in the PR template, there is point "New section is never added with the code changes. ". This is because the latest version in the changelog is always set to the unreleased version. See `RELEASING.md` doc for more info. So, just add that same entry to the 1.3.0.0 section instead.

FYI, CI is failing due to laziness annotation.

lehins (approved):
Perfect. Thank you

PR #3463, opened 2023-05-31
fix conway CDDL, Value allows larger values

JaredCorduan:
# Description

In conway, the number of any individual multi-asset in a transaction output can be any positive Word64. The CDDL, however, was specifying that the max was `4,294,967,295`.

See:
- https://github.com/input-output-hk/cardano-ledger/blob/c1603995da7c57d95371844191d235b05bb476b7/eras/mary/impl/src/Cardano/Ledger/Mary/Value.hs#L296-L303
- https://github.com/input-output-hk/cardano-ledger/blob/c1603995da7c57d95371844191d235b05bb476b7/libs/cardano-ledger-core/src/Cardano/Ledger/Coin.hs#L147-L152

The CDDL property tests that currently exist will ensure that expanding the values from the CDDL do not lead to deserialization errors.

Many thanks to @vsubhuman for catching this!

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/eras/conway/test-suite/cddl-files/conway.cddl
+++ b/eras/conway/test-suite/cddl-files/conway.cddl
@@ -487,11 +487,10 @@ asset_name = bytes .size (0..32)
 negInt64 = -9223372036854775808 .. -1
 posInt64 = 1 .. 9223372036854775807
 nonZeroInt64 = negInt64 / posInt64 ; this is the same as the current int64 definition but without zero
-natNum = 1 .. 4294967295 ; non-zero 32bit uint
 
 positive_coin = 1 .. 18446744073709551615
 
-value = positive_coin / [positive_coin,multiasset<natNum>]
+value = positive_coin / [positive_coin,multiasset<positive_coin>]
 
 mint = multiasset<nonZeroInt64>
 
--- a/eras/conway/test-suite/CHANGELOG.md
+++ b/eras/conway/test-suite/CHANGELOG.md
@@ -1,0 +1,6 @@
+# Version history for `cardano-ledger-conway-test`
+
+## 1.2.0.0
+
+- changes have been made
+


eras/conway/test-suite/cddl-files/conway.cddl
```diff
@@ -487,11 +487,10 @@ asset_name = bytes .size (0..32)
 negInt64 = -9223372036854775808 .. -1
 posInt64 = 1 .. 9223372036854775807
 nonZeroInt64 = negInt64 / posInt64 ; this is the same as the current int64 definition but without zero
-natNum = 1 .. 4294967295 ; non-zero 32bit uint
 
 positive_coin = 1 .. 18446744073709551615
 
-value = positive_coin / [positive_coin,multiasset<natNum>]
+value = positive_coin / [positive_coin,multiasset<positive_coin>]
```
vsubhuman, line 493: @JaredCorduan , I think the correct change here would be this:

```suggestion
value = positive_coin / [positive_coin,multiasset<nonZeroInt64>]
```

Because in the code, technically, the multi-asset decoding in the value is completely equal to that of the mint. See here: `decodeValuePair` uses `decodeMultiAssetMaps` internally: https://github.com/input-output-hk/cardano-ledger/blob/c1603995da7c57d95371844191d235b05bb476b7/eras/mary/impl/src/Cardano/Ledger/Mary/Value.hs#L325 and `decodeMultiAssetMint` uses same exact `decodeMultiAssetMaps` internally as well: https://github.com/input-output-hk/cardano-ledger/blob/c1603995da7c57d95371844191d235b05bb476b7/eras/mary/impl/src/Cardano/Ledger/Mary/Value.hs#L349

The mint is defined as 
```
mint = multiasset<nonZeroInt64>
```

So the multi-asset in the value should just be the same

vsubhuman: The `positive_coin`, used in your fix has maximum value of 18446744073709551615, which is not equal to the maximum value of the `posInt64` - 9223372036854775807, which I think is the limit to how much of an asset can be sent

zuzunker, line 493: Needs confirming that `uint64` is actually allowed for multiasset in a value. Same code seems to be used to parse multi-assets in the mint and in values, and mint allows type `int64`, because it can be negative. There is a suspicion that multiasset in values can also only have the 9223372036854775807 (`int64`) maximum limit and not the 18446744073709551615 (`uint64`).

So potentially the type `posInt64` should be used instead of the `positive_coin`.

P.S. Thanks to @lisicky for pointing out the new types in the Conway CDDL.

zuzunker: Although this line does suggest that asset amounts are supposed to be Word64 as well, so should be all good if it's really so: https://github.com/input-output-hk/cardano-ledger/blob/c1603995da7c57d95371844191d235b05bb476b7/eras/mary/impl/src/Cardano/Ledger/Mary/Value.hs#L494

zuzunker: And if those Word64s are treated is `uint64` and not as `int64`

JaredCorduan: for `Value` (which is used it transaction outputs), you can see from the code snippets in the PR notes that the Haskell is decoding a `Word64` and producing an error for zero.

```
$ ghci
GHCi, version 8.8.4: https://www.haskell.org/ghc/  :? for help
Prelude> import Data.Word
Prelude Data.Word> maxBound :: Word64
18446744073709551615
```

zuzunker: The PR is all correct, I have figured finally that `decodePositiveCoin` is used as the quantities decoder, when `decodeMultiAssetMaps` is called for a value, while a different `decodeIntegerBounded64` function is used as a decoder, when parsing a mint. So everything is correct and the `positive_coin` type is right to be used for a value multiasset 👍

zuzunker (approved):

lehins (approved):
Have no clue how it became a 32bit number. Great catch!

PR #3464, opened 2023-06-01
Add POOL rules to Conway

aniketd:
# Description

- [x] Replace the `TransitionRule` function entirely with just calling the Shelley version instead, as there are currently no significant differences between the two. @aniketd 

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Pool.hs
+++ b/eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Pool.hs
@@ -4,7 +4,6 @@
 {-# LANGUAGE FlexibleInstances #-}
 {-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE OverloadedStrings #-}
-{-# LANGUAGE PatternSynonyms #-}
 {-# LANGUAGE RankNTypes #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE StandaloneDeriving #-}
@@ -19,6 +18,7 @@ module Cardano.Ledger.Shelley.Rules.Pool (
   PoolEnv (..),
   PredicateFailure,
   ShelleyPoolPredFailure (..),
+  poolDelegationTransition,
 )
 where
 
@@ -51,11 +51,11 @@ import Cardano.Ledger.Shelley.TxBody (
   getRwdNetwork,
  )
 import Cardano.Ledger.Shelley.TxCert (
+  isDeRegKey,
+  isDelegation,
+  isGenesisDelegation,
   isInstantaneousRewards,
-  pattern DelegStakeTxCert,
-  pattern GenesisDelegTxCert,
-  pattern RegTxCert,
-  pattern UnRegTxCert,
+  isRegKey,
  )
 import Cardano.Ledger.Slot (EpochNo (..), SlotNo, epochInfoEpoch)
 import Control.DeepSeq
@@ -73,6 +73,7 @@ import Control.State.Transition (
   (?!),
  )
 import qualified Data.ByteString as BS
+import Data.Kind (Type)
 import Data.Word (Word8)
 import GHC.Generics (Generic)
 import Lens.Micro ((^.))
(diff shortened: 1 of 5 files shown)


eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Pool.hs
```diff
@@ -172,15 +173,24 @@ instance Era era => DecCBOR (ShelleyPoolPredFailure era) where
...
+  , PredicateFailure (ledger era) ~ ShelleyPoolPredFailure era
+  ) =>
+  TransitionRule (ledger era)
 poolDelegationTransition = do
   TRC (PoolEnv slot pp, ps, c) <- judgmentContext
   let stpools = psStakePoolParams ps
   let pv = pp ^. ppProtocolVersionL
-  case c of
-    RegPoolTxCert poolParam -> do
+  case (getRegPoolTxCert c, getRetirePoolTxCert c) of
```
Soupstraw: @lehins created [patterns synonyms](https://github.com/input-output-hk/cardano-ledger/blob/master/eras/shelley/impl/src/Cardano/Ledger/Shelley/TxCert.hs#L173) that don't assume a concrete TxCert. You should use these instead.

Soupstraw: it shouldn't be necessary to change this function at all.

Soupstraw: oh, I see you are turning the rule into a type parameter, that makes sense

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Pool.hs
```diff
@@ -172,15 +173,24 @@ instance Era era => DecCBOR (ShelleyPoolPredFailure era) where
       k -> invalidKey k
 
 poolDelegationTransition ::
-  forall era.
-  (ShelleyEraTxCert era, EraPParams era) =>
-  TransitionRule (ShelleyPOOL era)
+  forall (ledger :: Type -> Type) era.
+  ( ShelleyEraTxCert era
+  , EraPParams era
+  , Signal (ledger era) ~ TxCert era
```
Soupstraw, line 243: I think it'd be a good idea to break this transition function into two smaller functions, one that takes a TxCert that we can use in Shelley, and then another one that only looks at the PoolCert cases that we can use in Conway. That way we can use PoolCert as the signal in Conway.

JaredCorduan (approved):
I like @Soupstraw 's comment about breaking that function up. That said, the PR looks good to me!

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Pool.hs
aniketd: Do you mean we would still use a specific `judgmentContext` in each of the eras and then call this function from both eras?

Soupstraw: yeah, the Shelley transition would extract the pool cert from the certificate and then pass it to the function. Conway can just use the function directly as it uses PoolCert as the signal.

aniketd: Done!

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Pool.hs
```diff
@@ -171,86 +174,99 @@ instance Era era => DecCBOR (ShelleyPoolPredFailure era) where
...
+  ( EraCrypto era ~ c
+  , EraPParams era
+  , State (ledger era) ~ PState era
+  , STS (ledger era)
+  , Event (ledger era) ~ PoolEvent era
+  , BaseM (ledger era) ~ ShelleyBase
+  , PredicateFailure (ledger era) ~ ShelleyPoolPredFailure era
+  ) =>
+  PoolEnv era ->
+  PState era ->
```
Soupstraw, line 188: I don't think this is necessary, since you already have the `State (ledger era) ~ PState era` constraint. You can just look the state up via `judgmentContext`.

aniketd: Oh! Yes you are right! My bad!

aniketd: But wait! If I do use the `judgmentContext` it makes it a bit ugly I thought. The `TransitionRule (ledger era)` makes use of that function to bring into scope exactly those things that are necessary for that rule. But we are using the rule as a "sub-call". Isn't it still better to pass all arguments explicitly to it? Because otherwise you get one call in which the `PoolCert` needs to be used and another in which the `TxCert` needs to be used, and so in the second case, the certificate returned by `judgmentContext` is left unused. I will ping you on Slack to discuss.

Soupstraw (approved):
LGTM

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Pool.hs
Soupstraw: yeah, let's leave it like that, I don't think there's an obvious solution to this.

PR #3467, opened 2023-06-02
Add VDEL rules to Conway

aniketd:
# Description

- [ ] ~~Rename all `CommitteeCert`s to `GovCert`s~~
- [ ] ~~Rename `TALLY` rule to `GOV` cc. @lehins @Soupstraw @WhatisRT (perhaps in a separate issue and PR)~~
- [x] The renames may be done in another follow-up PR.

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Cert.hs
+++ b/eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Cert.hs
@@ -26,7 +26,7 @@ import Cardano.Ledger.Binary.Coders
 import Cardano.Ledger.Conway.Core
 import Cardano.Ledger.Conway.Era (ConwayCERT, ConwayDELEG, ConwayPOOL, ConwayVDEL)
 import Cardano.Ledger.Conway.Rules.Deleg (ConwayDelegPredFailure (..))
-import Cardano.Ledger.Conway.Rules.VDel (ConwayVDelPredFailure, VDelEnv (VDelEnv))
+import Cardano.Ledger.Conway.Rules.VDel (ConwayVDelPredFailure)
 import Cardano.Ledger.Conway.TxCert (ConwayCommitteeCert, ConwayDelegCert, ConwayTxCert (..))
 import Cardano.Ledger.Shelley.API (CertState (..), DState, DelegEnv (DelegEnv), DelplEnv (DelplEnv), PState, PoolEnv (PoolEnv), VState)
 import Cardano.Ledger.Shelley.Rules (ShelleyPoolPredFailure)
@@ -83,7 +83,7 @@ instance
   , State (EraRule "VDEL" era) ~ VState era
   , Environment (EraRule "DELEG" era) ~ DelegEnv era
   , Environment (EraRule "POOL" era) ~ PoolEnv era
-  , Environment (EraRule "VDEL" era) ~ VDelEnv era
+  , Environment (EraRule "VDEL" era) ~ PParams era
   , Signal (EraRule "DELEG" era) ~ ConwayDelegCert (EraCrypto era)
   , Signal (EraRule "POOL" era) ~ PoolCert (EraCrypto era)
   , Signal (EraRule "VDEL" era) ~ ConwayCommitteeCert (EraCrypto era)
@@ -110,7 +110,7 @@ certTransition ::
   , State (EraRule "VDEL" era) ~ VState era
   , Environment (EraRule "DELEG" era) ~ DelegEnv era
   , Environment (EraRule "POOL" era) ~ PoolEnv era
-  , Environment (EraRule "VDEL" era) ~ VDelEnv era
+  , Environment (EraRule "VDEL" era) ~ PParams era
   , Signal (EraRule "DELEG" era) ~ ConwayDelegCert (EraCrypto era)
   , Signal (EraRule "POOL" era) ~ PoolCert (EraCrypto era)
   , Signal (EraRule "VDEL" era) ~ ConwayCommitteeCert (EraCrypto era)
@@ -129,8 +129,8 @@ certTransition = do
     ConwayTxCertPool poolCert -> do
       newPState <- trans @(EraRule "POOL" era) $ TRC (PoolEnv slot pp, certPState, poolCert)
       pure $ cState {certPState = newPState}
-    ConwayTxCertCommittee committeeCert -> do
-      newVState <- trans @(EraRule "VDEL" era) $ TRC (VDelEnv, certVState, committeeCert)
+    ConwayTxCertCommittee vDelCert -> do
+      newVState <- trans @(EraRule "VDEL" era) $ TRC (pp, certVState, vDelCert)
       pure $ cState {certVState = newVState}
 
 instance
--- a/eras/conway/impl/src/Cardano/Ledger/Conway/Rules/VDel.hs
+++ b/eras/conway/impl/src/Cardano/Ledger/Conway/Rules/VDel.hs
@@ -4,7 +4,9 @@
 {-# LANGUAGE EmptyDataDeriving #-}
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE LambdaCase #-}
+{-# LANGUAGE NamedFieldPuns #-}
 {-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeApplications #-}
 {-# LANGUAGE TypeFamilies #-}
 {-# LANGUAGE TypeOperators #-}
 {-# LANGUAGE UndecidableInstances #-}
@@ -13,8 +15,7 @@
 module Cardano.Ledger.Conway.Rules.VDel (
   ConwayVDEL,
   ConwayVDelEvent (..),
-  VDelEnv (..),
-  ConwayVDelPredFailure,
+  ConwayVDelPredFailure (..),
 )
 where
 
@@ -23,12 +24,14 @@ import Cardano.Ledger.BaseTypes (
  )
 import Cardano.Ledger.Binary (DecCBOR (..), EncCBOR (..), encodeListLen)
 import Cardano.Ledger.Binary.Coders
-import Cardano.Ledger.CertState (VState)
+import Cardano.Ledger.CertState (VState (..))
 import Cardano.Ledger.Conway.Era (ConwayVDEL)
-import Cardano.Ledger.Conway.TxCert (ConwayCommitteeCert)
-import Cardano.Ledger.Core (Era (EraCrypto), EraRule)
+import Cardano.Ledger.Conway.TxCert (ConwayCommitteeCert (..))
+import Cardano.Ledger.Core (Era (EraCrypto), EraPParams, EraRule, PParams)
+import Cardano.Ledger.Credential (Credential)
+import Cardano.Ledger.Keys (KeyHash, KeyRole (CommitteeColdKey, Voting))
 import Control.DeepSeq (NFData)
-import Control.State.Transition (
+import Control.State.Transition.Extended (
   BaseM,
   Environment,
   Event,
@@ -36,15 +39,25 @@ import Control.State.Transition (
   STS,
   Signal,
   State,
+  TRC (TRC),
+  TransitionRule,
+  judgmentContext,
   transitionRules,
+  (?!),
  )
-import Data.Typeable (Typeable)
-import Data.Word (Word8)
 import GHC.Generics (Generic)
 import NoThunks.Class (NoThunks (..))
 
-data ConwayVDelPredFailure era = ConwayVDelPredFailure
-  deriving (Show, Eq, Generic, NoThunks, NFData)
+data ConwayVDelPredFailure era
+  = ConwayDRepAlreadyRegisteredVDEL !(Credential 'Voting (EraCrypto era))
+  | ConwayDRepNotRegisteredVDEL !(Credential 'Voting (EraCrypto era))
+  | ConwayCommitteeAlreadyRegisteredVDEL !(KeyHash 'CommitteeColdKey (EraCrypto era))
+  | ConwayCommitteeNotRegisteredVDEL !(KeyHash 'CommitteeColdKey (EraCrypto era))
+  deriving (Show, Eq, Generic)
+
+instance NoThunks (ConwayVDelPredFailure era)
+
+instance NFData (ConwayVDelPredFailure era)
 
 instance Typeable era => EncCBOR (ConwayVDelPredFailure era) where
   encCBOR = \case
@@ -56,15 +69,13 @@ instance Typeable era => DecCBOR (ConwayVDelPredFailure era) where
       0 -> pure (1, ConwayVDelPredFailure)
       k -> invalidKey k
 
-data VDelEnv era = VDelEnv
-
 newtype ConwayVDelEvent era = VDelEvent (Event (EraRule "VDEL" era))
 
 instance
-  ( Era era
+  ( EraPParams era
   , State (EraRule "VDEL" era) ~ VState era
   , Signal (EraRule "VDEL" era) ~ ConwayCommitteeCert (EraCrypto era)
-  , Environment (EraRule "VDEL" era) ~ VDelEnv era
+  , Environment (EraRule "VDEL" era) ~ PParams era
   , EraRule "VDEL" era ~ ConwayVDEL era
   , Eq (PredicateFailure (EraRule "VDEL" era))
   , Show (PredicateFailure (EraRule "VDEL" era))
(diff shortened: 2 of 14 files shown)


eras/conway/impl/src/Cardano/Ledger/Conway/Rules/VDel.hs
```diff
@@ -3,63 +3,110 @@
...
+  TRC
+    ( _pp
+      , vState@VState {vsDReps, vsCCHotKeys}
+      , c
+      ) <-
+    judgmentContext
+  case c of
+    ConwayVDelRegDRep cred -> do
+      Set.notMember cred vsDReps ?! ConwayDRepAlreadyRegisteredVDEL cred
+      -- NOTE: @Soupstraw: something about DRep Deposits seems to be missing here.
```
Soupstraw: We don't currently have a protocol parameter for DRep deposits, we should at some point add it and then check against that instead of using the poolDeposit parameter. I think it's a good idea to leave a TODO here and then revisit it once we've updated PParams

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/VDel.hs
```diff
@@ -3,63 +3,110 @@
...
+        vState
+          { vsDReps = Set.insert cred vsDReps
+          }
+    ConwayVDelUnRegDRep cred ->
+      pure $
+        vState
+          { vsDReps = Set.delete cred vsDReps
+          }
+    ConwayVDelRegCommittee coldK hotK -> do
+      -- NOTE: @Soupstraw: from the spec: "Should we check if KeyHash actually belongs to the CC?"
```
Soupstraw: Leave this as a TODO, we'll revisit this once we've figured it out in the spec.

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/VDel.hs
```diff
@@ -3,63 +3,110 @@
...
  )
+import qualified Data.Map.Strict as Map
+import qualified Data.Set as Set
 import GHC.Generics (Generic)
 import NoThunks.Class (NoThunks (..))
 
 data ConwayVDelPredFailure era
-  deriving (Show, Eq, Generic, NoThunks, NFData)
+  = ConwayDRepAlreadyRegisteredVDEL !(Credential 'Voting (EraCrypto era))
+  | ConwayCommitteeAlreadyRegisteredVDEL !(KeyHash 'CommitteeColdKey (EraCrypto era))
```
Soupstraw: I'm not sure we want to add `VDEL` to the end of every failure. Maybe @lehins can comment on the naming practices.

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/VDel.hs
```diff
@@ -3,63 +3,110 @@
...
+    judgmentContext
+  case c of
+    ConwayVDelRegDRep cred -> do
+      Set.notMember cred vsDReps ?! ConwayDRepAlreadyRegisteredVDEL cred
+      -- NOTE: @Soupstraw: something about DRep Deposits seems to be missing here.
+      pure $
+        vState
+          { vsDReps = Set.insert cred vsDReps
+          }
+    ConwayVDelUnRegDRep cred ->
```
Soupstraw: Here we should check that the credential is in the DReps map. I think the spec has a mistake at this place, I'll fix that.

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/VDel.hs
```diff
@@ -3,63 +3,110 @@
...
+      pure $
+        vState
+          { vsDReps = Set.delete cred vsDReps
+          }
+    ConwayVDelRegCommittee coldK hotK -> do
+      -- NOTE: @Soupstraw: from the spec: "Should we check if KeyHash actually belongs to the CC?"
+      Map.notMember coldK vsCCHotKeys ?! ConwayCommitteeAlreadyRegisteredVDEL coldK
```
Soupstraw: We might want to allow re-registering an already exisiting key, the spec is not very clear about this. I'll look into it.

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/VDel.hs
```diff
@@ -3,63 +3,110 @@
...
+      pure $
+        vState
+          { vsCCHotKeys = Map.insert coldK hotK vsCCHotKeys
+          }
+    ConwayVDelUnRegCommittee coldK ->
```
Soupstraw: Here we probably want to check that the cold key is registered.

libs/cardano-ledger-core/src/Cardano/Ledger/CertState.hs
```diff
@@ -257,8 +257,8 @@ data VState era = VState
   { vsDReps :: !(Set (Credential 'Voting (EraCrypto era)))
   , vsCCHotKeys ::
       !( Map
-          (KeyHash 'Voting (EraCrypto era))
-          (KeyHash 'Voting (EraCrypto era))
+          (KeyHash 'CommitteeColdKey (EraCrypto era))
+          (KeyHash 'CommitteeHotKey (EraCrypto era))
```
Soupstraw: nice!

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/VDel.hs
aniketd: I think this is the pattern in all the rules. It will soon be renamed to `GOVCERT` though, as per discussions on Slack :+1:

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/VDel.hs
aniketd: Cool! Let me know if I need to take this out.

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/VDel.hs
aniketd: Got you!

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/VDel.hs
aniketd: Got it!

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/VDel.hs
Soupstraw: yeah, it seems like we'll want to allow re-registering without having to unregister first, so this check can be removed.

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Cert.hs
```diff
@@ -125,8 +126,8 @@ certTransition = do
     ConwayTxCertPool poolCert -> do
       newPState <- trans @(EraRule "POOL" era) $ TRC (PoolEnv slot pp, certPState, poolCert)
       pure $ cState {certPState = newPState}
-    ConwayTxCertCommittee committeeCert -> do
-      newVState <- trans @(EraRule "VDEL" era) $ TRC (VDelEnv, certVState, committeeCert)
+    ConwayTxCertCommittee vDelCert -> do
```
Soupstraw, line 132: I think we can also rename the `ConwayTxCertCommittee` constructor.

aniketd: Can I do this along with the other the renames together in a follow up PR?

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/VDel.hs
```diff
@@ -12,44 +13,57 @@
...
+import qualified Data.Map.Strict as Map
+import qualified Data.Set as Set
 import GHC.Generics (Generic)
 import NoThunks.Class (NoThunks (..))
 
 data ConwayVDelPredFailure era
-  deriving (Show, Eq, Generic, NoThunks, NFData)
+  = ConwayDRepAlreadyRegisteredVDEL !(Credential 'Voting (EraCrypto era))
+  | ConwayDRepNotRegisteredVDEL !(Credential 'Voting (EraCrypto era))
+  | ConwayCommitteeAlreadyRegisteredVDEL !(KeyHash 'CommitteeColdKey (EraCrypto era))
```
WhatisRT: This isn't necessary. The point of the hot/cold key scheme is to be able to change the hot key, so being already registered cannot be a failure.

The error message right below also isn't right. The only way registering can fail is if that CC member has resigned. So one error named something like `ConwayCommitteeHasResignedVDEL` would be appropriate.

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/VDel.hs
```diff
@@ -58,9 +72,44 @@ instance
...
+          { vsDReps = Set.insert cred vsDReps
+          }
+    ConwayDRepUnReg cred -> do
+      Set.member cred vsDReps ?! ConwayDRepNotRegisteredVDEL cred
+      pure $
+        vState
+          { vsDReps = Set.delete cred vsDReps
+          }
+    ConwayAuthCommitteeHotKey coldK hotK -> do
+      -- TODO: implement after the spec is to be updated w.r.t. checking if the coldK belongs to CC.
```
WhatisRT: The two rules here should have identical logic. They fail iff the CC member is already resigned. So `vsCCHotKeys` needs to accommodate that somebody has resigned. In the spec this is done with inserting a `nothing`, so `Map.delete` below needs to be changed.

Also, this TODO comment can be removed - we won't check is the key is in the CC. Leaving that check out actually speeds up changing the CC by an epoch.

WhatisRT (commented):
I don't think I can say anything useful beyond what's in `VDel.hs`, so I'm not sure how useful my review is.

The details of this rule have changed a bit (see [here](https://github.com/input-output-hk/formal-ledger-specifications/pull/123/files#diff-2052333992aa192efb62901ebc339a8c4bcf6cb2001c9015d4ecdb4727210995)) but I don't know if you want to incorporate all of that here before that PR has even gotten a review.

eras/conway/impl/src/Cardano/Ledger/Conway/TxCert.hs
```diff
@@ -291,6 +329,7 @@ committeeKeyHashWitness :: ConwayCommitteeCert c -> KeyHash 'Witness c
 committeeKeyHashWitness = \case
   ConwayAuthCommitteeHotKey coldKeyHash _ -> asWitness coldKeyHash
   ConwayResignCommitteeColdKey coldKeyHash -> asWitness coldKeyHash
+  _ -> undefined
```
lehins: Due to changes to the cert type, this function now needs to return a Maybe
```suggestion
  _ -> Nothing
```

eras/conway/impl/src/Cardano/Ledger/Conway/TxCert.hs
```diff
@@ -279,7 +315,9 @@ instance NFData (ConwayDelegCert c)
 instance NoThunks (ConwayDelegCert c)
 
 data ConwayCommitteeCert c
-  = ConwayAuthCommitteeHotKey !(KeyHash 'CommitteeColdKey c) !(KeyHash 'CommitteeHotKey c)
+  = ConwayDRepReg !(Credential 'Voting c)
```
lehins: I think it would be good to add deposits to the reg/unreg certificates in this PR. This would reduce number of times we have to touch CBOR implementation and cddl spec

libs/cardano-ledger-core/src/Cardano/Ledger/CertState.hs
```diff
@@ -257,8 +257,8 @@ data VState era = VState
   { vsDReps :: !(Set (Credential 'Voting (EraCrypto era)))
   , vsCCHotKeys ::
```
lehins: This will be good for consistency
```suggestion
  , vsCommitteeHotKeys ::
```

lehins (changes requested):
Looks good. It's definitely ok, to merge this without it being 100% finished. All the discussed renames can be done in a follow up PR.

eras/conway/impl/src/Cardano/Ledger/Conway/TxCert.hs
aniketd: Got you!

eras/conway/impl/src/Cardano/Ledger/Conway/TxCert.hs
aniketd: @WhatisRT should we also store the deposits somewhere in the `VState`?

lehins: Yes, we'll need to store deposits in the VState, but we can take care of handling deposits with a later PR. I am only suggesting adding the `Coin` fields to `ConwayDRepReg` and `ConwayDRepUnReg`, just so at least the types are ready.

aniketd: Yes, thanks @lehins ! I have added them but not using them to make any checks etc. yet :+1:

lehins: Also for consistency sake they should be named:
`ConwayRegDRep` and `ConwayUnRegDRep`

aniketd: Got it!

aniketd:
@WhatisRT 

> The details of this rule have changed a bit (see [here](https://github.com/input-output-hk/formal-ledger-specifications/pull/123/files#diff-2052333992aa192efb62901ebc339a8c4bcf6cb2001c9015d4ecdb4727210995)) but I don't know if you want to incorporate all of that here before that PR has even gotten a review.

I have not incorporated those changes here :+1: . But, the changes you and @lehins  suggested have been done here now, and this is ready for your final review. :raised_hands:

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/VDel.hs
```diff
@@ -58,9 +75,34 @@ instance
...
+      pure $ vState {vsDReps = Set.insert cred vsDReps}
+    ConwayUnRegDRep cred _deposit -> do
+      -- TODO: check against a new PParam `drepDeposit`, once PParams are updated. -- someCheck ?! ConwayDRepIncorrectDeposit deposit
+      Set.member cred vsDReps ?! ConwayDRepNotRegisteredVDEL cred
+      pure $ vState {vsDReps = Set.delete cred vsDReps}
+    ConwayAuthCommitteeHotKey coldK hotK -> do
+      (isNothing <$> Map.lookup coldK vsCommitteeHotKeys) /= Just True ?! ConwayCommitteeHasResignedVDEL coldK
+      pure $ vState {vsCommitteeHotKeys = Map.insert coldK (Just hotK) vsCommitteeHotKeys}
+    ConwayResignCommitteeColdKey coldK -> do
+      Map.member coldK vsCommitteeHotKeys ?! ConwayCommitteeNotRegisteredVDEL coldK
```
WhatisRT: This line and the corresponding error message should be removed. There is no reason why someone shouldn't be allowed to immediately resign.

aniketd: Oh okay, I will remove it now. This would allow a person who hasn't ever authorized a hot key, to simply resign. Thanks @WhatisRT :raised_hands:

lehins: @WhatisRT What if committee member already resigned, double resign is likely a mistake, so checking for it might be good, right?
```suggestion
       validateCommitteeColdKeyHasNotResigneed coldK vsCommitteeHotKeys
```

where

```haskell
validateCommitteeColdKeyHasNotResigneed coldK vsCommitteeHotKeys =
 (isNothing <$> Map.lookup coldK vsCommitteeHotKeys) /= Just True ?! ConwayCommitteeHasResignedVDEL coldK
```

WhatisRT: Yeah, it doesn't really matter since it doesn't change the state. But the logic in the spec is the same, if you've resigned you can't publish any certs for that cold key anymore. So that check makes sense to keep.

aniketd: Updated! :raised_hands:

WhatisRT (approved):
LGTM

eras/conway/impl/src/Cardano/Ledger/Conway/TxCert.hs
```diff
@@ -347,6 +387,14 @@ conwayTxCertDelegDecoder = \case
...
     cred <- decCBOR
     pure (2, ResignCommitteeColdTxCert cred)
+  16 -> do
+    cred <- decCBOR
+    deposit <- decCBOR
+    pure (3, RegDRepTxCert cred deposit)
+  17 -> do
+    cred <- decCBOR
+    deposit <- decCBOR
+    pure (3, UnRegDRepTxCert cred deposit)
```
lehins, line 397: We also need to adjust cddl spec for conway certificates. It's ok to add it in a subsequent PR too.

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/VDel.hs
```diff
@@ -12,44 +13,60 @@
...
 import GHC.Generics (Generic)
 import NoThunks.Class (NoThunks (..))
 
 data ConwayVDelPredFailure era
-  deriving (Show, Eq, Generic, NoThunks, NFData)
+  = ConwayDRepAlreadyRegisteredVDEL !(Credential 'Voting (EraCrypto era))
+  | ConwayDRepNotRegisteredVDEL !(Credential 'Voting (EraCrypto era))
+  | ConwayDRepIncorrectDepositVDEL !Coin
+  | ConwayCommitteeHasResignedVDEL !(KeyHash 'CommitteeColdKey (EraCrypto era))
+  | ConwayCommitteeNotRegisteredVDEL !(KeyHash 'CommitteeColdKey (EraCrypto era))
```
lehins: You might wanna rebase on this PR and add Enc/DecCBOR instances for this type
https://github.com/input-output-hk/cardano-ledger/pull/3477

Soon above PR will get merged and you'll have to do it anyways, so it is just a heads up

lehins (approved):
Very nice!

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/VDel.hs
aniketd: Oh yes, I saw @teodanciu PR just now. Got it! I will rebase on top of it and update here. :raised_hands: Thanks Alexey!

eras/conway/impl/src/Cardano/Ledger/Conway/TxCert.hs
aniketd: Awesome catch! Hope I have done the right thing there now :crossed_fingers:

lehins (approved):
Beautiful!

PR #3482, opened 2023-06-15
Improve Conway certificate safety

lehins:
# Description

Deprecated certificates (MIR and GenesisDeleg) were still usable in Conway, as far as the type system is concerned. This PR fixes it with proper deprecation of those certificates

* Prevent using `getMirTxCert` from being used in eras after Babbage. This also affects
  all functions that use it
* Prevent using `mkGenesisDelegTxCert`, `getGenesisDelegTxCert` and `GenesisDelegTxCert`
  from being used in eras after Babbage. This also affects all functions that use those
  functions and patern synonym
* Change `WrongCertificateTypePOOL`. Instead of accepting a number tag of a type, it now
  contains the actual violating certificate. This was an impossible case, so it should not
  really affect downstream users
* Add a placeholder for `ConwayUTXOW` rule and replace the previous `"UTXOW"` for `ConwayEra`. The rule itself will be implemented in #3466 

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Pool.hs
+++ b/eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Pool.hs
@@ -52,13 +52,6 @@ import Cardano.Ledger.Shelley.TxBody (
   PoolParams (..),
   getRwdNetwork,
  )
-import Cardano.Ledger.Shelley.TxCert (
-  isDeRegKey,
-  isDelegation,
-  isGenesisDelegation,
-  isInstantaneousRewards,
-  isRegKey,
- )
 import Cardano.Ledger.Slot (EpochNo (..), SlotNo, epochInfoEpoch)
 import Control.DeepSeq
 import Control.Monad (forM_, when)
@@ -96,7 +89,7 @@ data ShelleyPoolPredFailure era
       !EpochNo -- The epoch listed in the Pool Retirement Certificate
       !EpochNo -- The first epoch that is too far out for retirement
   | WrongCertificateTypePOOL
-      !Word8 -- The disallowed certificate (this case should never happen)
+      !(TxCert era) -- The disallowed certificate (this case should never happen)
   | StakePoolCostTooLowPOOL
       !Coin -- The stake pool cost listed in the Pool Registration Certificate
       !Coin -- The minimum stake pool cost listed in the protocol parameters
@@ -107,11 +100,15 @@ data ShelleyPoolPredFailure era
   | PoolMedataHashTooBig
       !(KeyHash 'StakePool (EraCrypto era)) -- Stake Pool ID
       !Int -- Size of the metadata hash
-  deriving (Show, Eq, Generic)
+  deriving (Generic)
 
-instance NoThunks (ShelleyPoolPredFailure era)
+deriving instance Show (TxCert era) => Show (ShelleyPoolPredFailure era)
 
-instance NFData (ShelleyPoolPredFailure era)
+deriving instance Eq (TxCert era) => Eq (ShelleyPoolPredFailure era)
+
+instance NoThunks (TxCert era) => NoThunks (ShelleyPoolPredFailure era)
+
+instance NFData (TxCert era) => NFData (ShelleyPoolPredFailure era)
 
 instance (ShelleyEraTxCert era, EraPParams era) => STS (ShelleyPOOL era) where
   type State (ShelleyPOOL era) = PState era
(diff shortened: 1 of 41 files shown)


eras/shelley/impl/testlib/Test/Cardano/Ledger/Shelley/Arbitrary.hs
```diff
@@ -652,9 +652,8 @@ instance Era era => Arbitrary (ShelleyPpupPredFailure era) where
   arbitrary = genericArbitraryU
   shrink = genericShrink
 
-instance Era era => Arbitrary (ShelleyPoolPredFailure era) where
+instance (Arbitrary (TxCert era), Era era) => Arbitrary (ShelleyPoolPredFailure era) where
   arbitrary = genericArbitraryU
-  shrink = genericShrink
```
lehins: This resulted in a bizarre overlapping generic instances error. So, I chose to remove shrinking instead of enabling IncoherentInstances

libs/cardano-ledger-test/cardano-ledger-test.cabal
```diff
@@ -1,6 +1,6 @@
 cabal-version:      3.0
 name:               cardano-ledger-test
-version:            1.1.0.0
+version:            9.9.9.9
```
lehins, line 3: Can't have any comments in cabal files, due to cabal formatter. Hopefully such strange version as `9.9.9.9` will deter engineers from releasing it

eras/shelley/impl/src/Cardano/Ledger/Shelley/TxCert.hs
```diff
@@ -366,7 +366,7 @@ instance
       gen <- decCBOR
       genDeleg <- decCBOR
       vrf <- decCBOR
-      pure (4, GenesisDelegTxCert gen genDeleg vrf)
```
teodanciu, line 369: The pattern doesn't work here because this instnace is meant to work for all eras, right?

lehins: Nah, it would work. We'd just have to add a constraint: `ProtVerAtMost era 8`. Which would be fine, but I decided to avoid adding constraints by using the actual type constructor.

Both are equivalent in functionality and safety guarantees.

lehins: Technically not 100% equivalent, but for our purposes they would be. 

* With constraint `ProtVerAtMost era 8` we say: This DecCBOR instance for ShelleyTxCert instance can only be used in Shelley through Babbage

* Without constraint we say: This DecCBOR instance for ShelleyTxCert can be used in any era, in which `TxCert ~ ShelleyTxCert`. Which is Shelley through Babbage :slightly_smiling_face:

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Pool.hs
```diff
@@ -250,24 +247,15 @@ poolDelegationTransition ::
...
       poolCertTransition penv ps (RetirePool sPoolId epochN)
-    _ | isRegKey c || isDeRegKey c || isDelegation c -> do
-      failBecause $ WrongCertificateTypePOOL 0
-      pure ps
-    _ | isInstantaneousRewards c -> do
-      failBecause $ WrongCertificateTypePOOL 1
-      pure ps
-    _ | isGenesisDelegation c -> do
-      failBecause $ WrongCertificateTypePOOL 2
-      pure ps
```
teodanciu, line 269: So actually all these 3 cases  (0, 1, 2) are in fact impossible, not just the one below with the comment `impossible case` ? 
Or is it the case that the other changes have rendered them impossible?

eras/shelley/impl/src/Cardano/Ledger/Shelley/TxCert.hs
teodanciu: Makes perfect sense, thank you!

teodanciu (approved):
Looks good  to me

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Pool.hs
lehins: Yes indeed an impossible case. In the matter of fact, thank you for asking this question, because it is possible to completely remove that case. Which is done in this follow up commit 6887cf750ef57559dbfde69e5bbfd8e253501fa6

teodanciu: Awesome simplification!  And  the `shrink` is back in `Arbitrary (ShelleyPoolPredFailure era)`

PR #3483, opened 2023-06-15
Implement reward gate

### LINKED ISSUE #3447: Pre-defined DReps
CIP-1694 describes two pre-defined DReps: https://github.com/JaredCorduan/CIPs/blob/voltaire-v1/CIP-1694/README.md#pre-defined-dreps

This logic needs to be accounted for in both the DRep delegation (perhaps a separate map) and the ledger rules.

### LINKED ISSUE #3450: Conway reward gate
In the conway ledger era, it will be required for a registered stake credential to either:

* delegate to a DRep
* delegate the pre-defined "No confidence" or "Abstain" Drep

in order to be able to withdraw their rewards.

Soupstraw:
# Description

This PR adds a check to LEDGER rule, which only allows withdrawing rewards if the reward address is delegated to a DRep. It also adds the `DRep` datatype from the spec.

closes #3450
closes #3447

# Checklist

- [x] Self-reviewed the diff


libs/cardano-ledger-core/src/Cardano/Ledger/UMap.hs
```diff
@@ -151,6 +155,29 @@ instance DecCBOR RDPair where
   decCBOR =
     decodeRecordNamed "RDPair" (const 2) $ RDPair <$> decCBOR <*> decCBOR
 
+data VDeleg c
+  = CredVoter (Credential 'Staking c)
+  | AbstainRep
+  | NoConfidenceRep
+  deriving (Show, Eq, Ord, Generic, NoThunks, NFData, ToExpr, ToJSON)
```
lehins: In order to reduce memory overhead and reduce serialization size we should unwind the Credential. Also, lets fixup the names for consistency
```suggestion
data DRep c
  = DRepKeyHash !(KeyHash 'Voting c)
  | DRepScriptHash !(ScriptHash c)
  | DRepAlwaysAbstain
  | DRepAlwaysNoConfidence
  deriving (Show, Eq, Ord, Generic, NoThunks, NFData, ToExpr, ToJSON)

getDRepCredential :: Drep c -> Maybe (Credential 'Voting c)
getDRepCredential = \case ..
```



Note also it is not `Staking`, but `Voting` :slightly_smiling_face:

Soupstraw: ~why not have `DRepId !(Credential 'Voting c)` instead?~ Ok, I reread your comment :facepalm:

lehins: That will be an extra memory overhead. The idea is that we'll have a lot of them. And since not all of the DRepIds will be in the VState there is no guarantee of memory sharing.

lehins: In other words unrolling like that will save us a machine word (8 bytes) per DRep that never registered. It will also help with serialization.

Soupstraw: Do you think it's worth it to hide the unwound credential behind a pattern synonym?

lehins: Yeah, it might not be a bad idea.

libs/cardano-ledger-core/src/Cardano/Ledger/UMap.hs
```diff
@@ -151,6 +159,52 @@ instance DecCBOR RDPair where
...
+dRepToCred _ = Nothing
+
+pattern DRepCredential :: Credential 'Voting c -> DRep c
+pattern DRepCredential c <- (dRepToCred -> Just c)
+  where
+    DRepCredential c = case c of
+      ScriptHashObj sh -> DRepScriptHash sh
+      KeyHashObj kh -> DRepKeyHash kh
+
+{-# COMPLETE DRepCredential, DRepAlwaysAbstain, DRepAlwaysNoConfidence :: DRep #-}
```
aniketd: The `DRep` data type may be more suited to reside along with `Credential` in its module. Let me know what you think.

Soupstraw: I feel like the `UMap` module is more fitting, because it has to do with governance and it only appears in the `UMap`, whereas Credentials are used everywhere.

libs/cardano-ledger-core/src/Cardano/Ledger/UMap.hs
```diff
@@ -151,6 +159,52 @@ instance DecCBOR RDPair where
   decCBOR =
     decodeRecordNamed "RDPair" (const 2) $ RDPair <$> decCBOR <*> decCBOR
 
+data DRep c
+  = DRepKeyHash (KeyHash 'Voting c)
+  | DRepScriptHash (ScriptHash c)
```
lehins:
```suggestion
  = DRepKeyHash !(KeyHash 'Voting c)
  | DRepScriptHash !(ScriptHash c)
```

eras/conway/test-suite/cddl-files/conway.cddl
```diff
@@ -303,6 +303,13 @@ credential =
   // 1, scripthash
   ]
 
+drep =
+  [ 0, addr_keyhash
+  // 1, scripthash
+  // 2   ; always abstain
```
```suggestion
  // 2  ; always abstain
```

libs/cardano-ledger-core/src/Cardano/Ledger/UMap.hs
lehins: I am leaning towards Aniket's suggestion that `UMap` is not a good place for `DRep` definition, except I don't think `Credential` module is the right place either. `DRep` is used in `TxCert`, so maybe placing it in `Cardano.Ledger.Core.TxCert` would be more sensible

libs/cardano-ledger-core/src/Cardano/Ledger/UMap.hs
```diff
@@ -104,12 +103,21 @@ module Cardano.Ledger.UMap (
   findWithDefault,
   size,
   domDeleteAll,
+
+  -- * Governance
+  DRep (
+    DRepAlwaysNoConfidence,
+    DRepAlwaysAbstain
+  ),
+  pattern DRepCredential,
```
lehins: There is a nice feature that Haskell has of exporting pattern synonyms together with the actual type :wink: 
```suggestion
    DRepAlwaysAbstain,
    DRepCredential
  ),
```

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Ledger.hs
```diff
@@ -243,6 +256,13 @@ ledgerTransition = do
             )
       else pure certState
 
+  let wdrlAddrs = Map.keysSet . unWithdrawals $ tx ^. bodyTxL . withdrawalsTxBodyL
+  let wdrlCreds = Set.map getRwdCred wdrlAddrs
+  let delegatedAddrs = Map.keysSet . dRepMap . dsUnified $ certDState certState'
+
+  eval (wdrlCreds ⊆ delegatedAddrs)
+    ?! ConwayWdrlNotDelegatedToDRep (wdrlCreds Set.\\ delegatedAddrs)
```
lehins: If my in github coding is correct this will be much faster, since there is neither Map nor Set conversion unless there is a predicate failure.
```suggestion
  let dUnified = dsUnified $ certDState certState'
      delegatedAddrs = DRepUView dUnified

  all (`UMap.member` delegatedAddrs) wdrlCreds
    ?! ConwayWdrlNotDelegatedToDRep (wdrlCreds Set.\\ Map.keysSet (dRepMap 
 dUnified))
```

This is quite important, cause it is expected we'll have quite a few DReps and doing this conversion for every transaction will be costly.

lehins (approved):
Looks great! Just a few improvements could be applied before we merge it.

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Ledger.hs
lehins: I just realized something very important. We want to disable this functionality for now, until we have the delegation fully implemented in cardano-api. Because otherwise we will not be submit regular transactions with withdrawals, which can be very disruptive for testing. (So, adding it in this PR is great, but let's disable it with `when False $ do ...` and a comment about why it is disabled)

PR #3490, opened 2023-06-19
Bump cardano-ledger-shelley-test version

teodanciu:
since its bounds on cardano-ledger-shelley have changed from 1.3 to 1.4

# Description

I'm having[ issues releasing to CHaP,](https://github.com/input-output-hk/cardano-haskell-packages/actions/runs/5315773558/jobs/9625101920?pr=353) and I think it might be because we haven't bumped this version when changing the cardano-ledger-shelley bounds (or at least one of the issues!) 

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/eras/shelley/test-suite/CHANGELOG.md
+++ b/eras/shelley/test-suite/CHANGELOG.md
@@ -2,4 +2,4 @@
 
 ## 1.2.0.1
 
-*
+* Changed bounds on cardano-ledger-shelley-test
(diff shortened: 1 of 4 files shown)


teodanciu:
For `cardano-protocol-tpraos`,  the upper bound of the shelley dependency has changed from 1.4 to 1.5 - does this require a release?

EDIT: I assumed yes and added a commit to reflect this

lehins (approved):
Good catch

PR #3492, opened 2023-06-21
Drop Cicero from flake

angerman:
use hydra from now on.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -33,21 +33,6 @@
         "type": "github"
       }
     },
-    "blank": {
-      "locked": {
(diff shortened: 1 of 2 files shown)


dermetfan (approved):

teodanciu (approved):
Looks good to me. 
I suppose the `tullia` references  from `flake.lock` will get removed when we `nix flake lock` ?

dermetfan:
Good catch, I removed the input and updated `flake.lock`

teodanciu:
Thank you @dermetfan

PR #3495, opened 2023-06-21
Proposal deposit returns

### LINKED ISSUE #3494: Return proposal deposits
input-output-hk/formal-ledger-specifications#102

Soupstraw:
# Description

This PR handles the proposal deposits. It stores the paid proposal deposits in a map in the `EnactState` and when the proposal becomes inactive (it is either enacted or rejected), the proposal is paid out to the reward account marked in that proposal. I also updated the `consumed == produced` test in Conway to ensure that the required deposit for governance actions is accounted for in the UTXO rule.

resolves #3494

# Checklist

- [x] Self-reviewed the diff


Soupstraw:
Waiting on #3498

eras/conway/impl/src/Cardano/Ledger/Conway/Governance.hs
```diff
@@ -249,7 +248,14 @@ instance EraPParams era => NoThunks (EnactState era)
 
 data RatifyState era = RatifyState
   { rsEnactState :: !(EnactState era)
-  , rsFuture :: !(StrictSeq (GovernanceActionId (EraCrypto era), GovernanceActionState era))
+  , rsFuture ::
+      !( StrictSeq
+          (GovernanceActionId (EraCrypto era), GovernanceActionState era)
```
lehins, line 258: Do we actually care about the order of governance actions? Wouldn't it make more sense to place them into a Map?

If we do care about the order, then we probably want to either use a strict tuple here or create a separate type with fields that are strict for:
```
data GovActionSomthing = GoveActionSomthing !(GovernanceActionId (EraCrypto era)) !(GovernanceActionState era)
```

Don't have to do this in this PR, leaving a TODO for later will be just fine

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Epoch.hs
```diff
@@ -136,6 +149,37 @@ instance
   type Event (ConwayEPOCH era) = ConwayEpochEvent era
   transitionRules = [epochTransition]
 
+returnProposalDeposits' ::
+  ConwayGovernance era ->
+  StrictSeq (GovernanceActionId (EraCrypto era), GovernanceActionState era) ->
+  UMap (EraCrypto era) ->
+  UMap (EraCrypto era)
+returnProposalDeposits' ConwayGovernance {..} gaids m =
```
lehins: I think avoiding ticks is a goooood practice :wink: 
```suggestion
returnProposalDepositsUMap ::
  ConwayGovernance era ->
  StrictSeq (GovernanceActionId (EraCrypto era), GovernanceActionState era) ->
  UMap (EraCrypto era) ->
  UMap (EraCrypto era)
returnProposalDepositsUMap ConwayGovernance {..} gaids m =
```

eras/conway/impl/src/Cardano/Ledger/Conway/Governance.hs
```diff
@@ -249,7 +248,14 @@ instance EraPParams era => NoThunks (EnactState era)
...
+       )
+  , rsRemoved ::
+      !( StrictSeq
+          (GovernanceActionId (EraCrypto era), GovernanceActionState era)
+       )
```
lehins, line 263: Just to understand your PR better `rsFuture` is a sequence, because the order in which governance actions where proposed actually matters, right?

We don't have the same constraint, so it would make sense to switch `rsRemoved` to a Map, no?

```suggestion
  , rsRemoved ::
      !( Map
          (GovernanceActionId (EraCrypto era)) (GovernanceActionState era)
       )
```

eras/conway/impl/src/Cardano/Ledger/Conway/Governance.hs
```diff
@@ -287,15 +295,17 @@ instance EraPParams era => ToJSON (RatifyState era) where
...
   let RatifyState {..} = cg
    in [ "enactState" .= rsEnactState
       , "future" .= rsFuture
+      , "removed" .= rsRemoved
       ]
 
 data ConwayGovernance era = ConwayGovernance
   { cgGov :: !(ConwayGovState era)
   , cgRatify :: !(RatifyState era)
+  , cgPropDeposits :: !(Map (GovernanceActionId (EraCrypto era)) Coin)
```
lehins: I don't think saving a couple of characters is justified (especially since prop is quite overloaded in haskell with property)
```suggestion
  , cgProposalDeposits :: !(Map (GovernanceActionId (EraCrypto era)) Coin)
```

eras/conway/impl/src/Cardano/Ledger/Conway/Governance.hs
```diff
@@ -305,19 +315,24 @@ cgGovL = lens cgGov (\x y -> x {cgGov = y})
 cgRatifyL :: Lens' (ConwayGovernance era) (RatifyState era)
 cgRatifyL = lens cgRatify (\x y -> x {cgRatify = y})
 
+cgPropDepositsL :: Lens' (ConwayGovernance era) (Map (GovernanceActionId (EraCrypto era)) Coin)
+cgPropDepositsL = lens cgPropDeposits (\x y -> x {cgPropDeposits = y})
```
```suggestion
cgProposalDepositsL :: Lens' (ConwayGovernance era) (Map (GovernanceActionId (EraCrypto era)) Coin)
cgProposalDepositsL = lens cgProposalDeposits (\x y -> x {cgProposalDeposits = y})
```

eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Procedures.hs
```diff
@@ -260,15 +261,24 @@ data GovernanceProcedures era = GovernanceProcedures
...
 
+-- | Attaches indices to a sequence of proposal procedures. The indices grow
+-- from left to right.
+indexedGovProps ::
+  Seq (ProposalProcedure era) ->
+  Seq (GovernanceActionIx, ProposalProcedure era)
+indexedGovProps = enumerateProps 0
+  where
+    enumerateProps _ Empty = Empty
+    enumerateProps n (x :<| xs) = (n, x) :<| enumerateProps (succ n) xs
```
lehins: Since tuples are lazy, we need to force the index
```suggestion
    enumerateProps !n (x :<| xs) = (n, x) :<| enumerateProps (succ n) xs
```

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Utxos.hs
```diff
@@ -128,16 +162,21 @@ conwayEvalScriptsTxValid ::
...
 
   let !_ = traceEvent validBegin ()
   expectScriptsToPass pp tx utxo
   let !_ = traceEvent validEnd ()
-
-  pure $! updateUTxOState pp u txBody depositChange pup
+  let !newUTxOState = updateUTxOState pp u txBody depositChange gov
+  pure $!
+    newUTxOState
+      & utxosGovernanceL . cgPropDepositsL %~ updateProposalDeposits pp txBody
```
lehins: Are we collecting deposits here in the UTXOS rule because we should not collect deposits for transactions that have invalid plutus scripts?

Am I understanding this correctly?

eras/conway/impl/src/Cardano/Ledger/Conway/Governance.hs
```diff
@@ -90,7 +93,6 @@ data GovernanceActionState era = GovernanceActionState
   { gasCommitteeVotes :: !(Map (Credential 'CommitteeHotKey (EraCrypto era)) Vote)
   , gasDRepVotes :: !(Map (Credential 'Voting (EraCrypto era)) Vote)
   , gasStakePoolVotes :: !(Map (KeyHash 'StakePool (EraCrypto era)) Vote)
-  , gasDeposit :: !Coin
```
lehins: Why can't we keep the deposit in the `GovernanceActionState`? It seems like a logical place for it to be in: both from the perspective of the organization and efficiency. Adding an extra Map for tracking deposits comes with memory overhead.

eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Procedures.hs
```diff
@@ -260,15 +261,24 @@ data GovernanceProcedures era = GovernanceProcedures
...
+
 instance EraPParams era => NoThunks (GovernanceProcedures era)
 
 instance EraPParams era => NFData (GovernanceProcedures era)
 
 deriving instance EraPParams era => Show (GovernanceProcedures era)
 
 data ProposalProcedure era = ProposalProcedure
-  { pProcDeposit :: !Coin
```
lehins: @WhatisRT Why was it decided to remove the deposit from the proposal? I thought we need it for the same reason why we need it in certificates:

* people can verify on chain how much deposit was payed
* helps hard wallets somehow

Not that it bothers me, I just want to confirm the change and learn the reason why it was made, for my own sake.

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Epoch.hs
```diff
@@ -136,6 +149,37 @@ instance
...
+  RewDepUView m ∪+ foldl' addRew mempty gaids
```
lehins: `unionRewAgg`/`∪+` has an invariant:
> We presume that the domain of the `Map` on the right, is a subset of the domain of the `UView` on the left.

Which means if the reward account is removed from UMap before the epoch boundary or not even present to begin with in the UMap, then the deposit will be dropped on the floor.

I might be wrong of course, but I think it is a valid concern.

lehins (changes requested):
I have  whole bunch of questions, but looks great overall

eras/conway/impl/src/Cardano/Ledger/Conway/Governance.hs
WhatisRT: We do care, since we need to enact them in the order of the proposal, so we need to preserve the order here

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Utxos.hs
WhatisRT: This isn't in the spec yet since we don't have a UTXOS rule yet, but yes, all deposits are collected in UTXOS for this reason

eras/conway/impl/src/Cardano/Ledger/Conway/Governance.hs
WhatisRT: We used to have them in there but it adds another pot of money, which isn't all that nice.

eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Procedures.hs
WhatisRT: Basically so that we don't have to store the deposit twice. But I do agree that it'd be nice to have it part of the transaction. But then we'd have two different types for the transaction & what we store.

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Epoch.hs
Soupstraw: good catch, we have a similar operator in the spec which acts as `unionWith`, so that's why I made that mistake here

eras/conway/impl/src/Cardano/Ledger/Conway/Governance.hs
lehins: Yeah, but duplicating the whole Map with governance proposals is bad too:

* It is bad for memory efficiency. The whole Map structure becomes a redundant overhead. Danger of keys not being shared. We really don't know how many proposals there will be, so we gotta be careful
* Also there is more potential for bugs, since it is easy to get the Map and actual GovActions out of sync

eras/conway/impl/src/Cardano/Ledger/Conway/Governance.hs
Soupstraw: `rsRemoved` is very short-lived data. We just use it to tell the Epoch rule which actions are no longer active so we can pay back the deposits for those. We fold over the whole sequence anyways, so I think it makes sense to keep it as a sequence.

eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Procedures.hs
```diff
@@ -13,6 +13,7 @@
 {-# OPTIONS_GHC -Wno-unrecognised-pragmas #-}
 
 {-# HLINT ignore "Use record patterns" #-}
+{-# LANGUAGE BangPatterns #-}
```
lehins: Can you group it with other LANGUAGE extensions

eras/conway/impl/src/Cardano/Ledger/Conway/Governance.hs
lehins: You don't just folder over the data structure, you convert it to a list, then a Set, then a Map.
```haskell
Map.withoutKeys m . Set.fromList . toList $ fst <$> removedProposals
```

So if we keep it as a Map instead of a Seq, then we don't have to do any of these expensive conversions. It becomes:
```
m `Map.difference` removedProposals
```

eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Procedures.hs
lehins: I don't quite get it. We don't store the `ProposalProcedure` in memory. It is an ephemeral data type that is alive as long as the transaction is being validated.

It would be placed on chain, but that's a different story.

eras/conway/impl/src/Cardano/Ledger/Conway/Governance.hs
Soupstraw: I think fold+delete would also give a similar effect

lehins: But why not keep it as a Map? The order in which proposals are removed is irrelevant. At least we'll get the uniqueness property out of it

lehins: Anyways. We can keep it as a sequence for now. It's not terribly important. We can change it later if necessary

Soupstraw: We're not keeping it as a map, `rsRemoved` is essentially built by filtering the active proposals. I feel like turning it into a map would be even more overhead. I could see a point of it being a map if we indexed into it.

eras/conway/impl/src/Cardano/Ledger/Conway/Governance.hs
Soupstraw: Yeah, I'd also prefer if it was stored in the action itself. We'll have to add an extra check in that case though to make sure that the declared deposit amount is correct.

Soupstraw: ok, I restored the deposit fields and removed the map. Now we just have to add a check somewhere that the declared deposit matches the one in PParams.

eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Procedures.hs
Soupstraw: I restored the `pProcDeposit` field

eras/conway/impl/src/Cardano/Ledger/Conway/UTxO.hs
```diff
@@ -1,27 +1,40 @@
+{-# LANGUAGE DataKinds #-}
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE FlexibleInstances #-}
 {-# LANGUAGE NamedFieldPuns #-}
 {-# LANGUAGE TypeFamilies #-}
 {-# OPTIONS_GHC -Wno-orphans #-}
 
-module Cardano.Ledger.Conway.UTxO () where
+module Cardano.Ledger.Conway.UTxO (
```
lehins, line 8: We need to add `Cardano.Ledger.Conway.UTxO` to `exposed-modules` in the cabal file, since now we export something from it, besides orphan instances

eras/conway/test-suite/cddl-files/conway.cddl
```diff
@@ -79,8 +79,7 @@ voting_procedure =
   ]
 
 proposal_procedure =
-  [ deposit : coin
-  , $hash28
+  [ $hash28
```
lehins: we need to bring back the deposit in cddl too.
```suggestion
  [ deposit : coin
  , $hash28
```

libs/cardano-ledger-core/CHANGELOG.md
```diff
@@ -13,6 +13,15 @@
...
 * Remove default implementation for `spendableInputsTxBodyL`
+* Add lenses:
+  * `dsUnifiedL`
+  * `dsGenDelegsL`
+  * `dsIRewardsL`
+  * `dsFutureGenDelegsL`
+  * `certDStateL`
+  * `certPStateL`
+  * `certVStateL`
+* Add `getProducedValue` to `EraTxBody`
```
```suggestion
* Add `getProducedValue` to `EraUTxO`
```

lehins (approved):
Looks great. Found just a couple minor changes that need to be done.

PR #3497, opened 2023-06-27
Conway Utxos

### LINKED ISSUE #3501: Create UTXOS rule in Conway
UTXOS rule hasn't changed much since Babbage, but it can't be reused verbatim, so we need to create a new rule that reuses as much as possible from Babbage.

teodanciu:
# Description

Copied from Babbage Utxos and adapted to Conway

Fixes #3501

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/eras/babbage/impl/src/Cardano/Ledger/Babbage/Rules/Utxos.hs
+++ b/eras/babbage/impl/src/Cardano/Ledger/Babbage/Rules/Utxos.hs
@@ -14,6 +14,9 @@
 module Cardano.Ledger.Babbage.Rules.Utxos (
   BabbageUTXOS,
   utxosTransition,
+  expectScriptsToPass,
+  tellDepositChangeEvent,
+  scriptsNo,
 ) where
 
 import Cardano.Ledger.Alonzo.Language (Language (..))
@@ -42,8 +45,10 @@ import Cardano.Ledger.Babbage.Era (BabbageUTXOS)
 import Cardano.Ledger.Babbage.Tx
 import Cardano.Ledger.BaseTypes (ShelleyBase, epochInfo, strictMaybeToMaybe, systemStart)
 import Cardano.Ledger.Binary (EncCBOR (..))
+import Cardano.Ledger.Coin (Coin)
 import Cardano.Ledger.SafeHash (hashAnnotated)
 import Cardano.Ledger.Shelley.LedgerState (
+  CertState,
   PPUPPredFailure,
   UTxOState (..),
   keyTxRefunds,
@@ -143,6 +148,59 @@ utxosTransition =
 
 -- ===================================================================
 
+tellDepositChangeEvent ::
+  forall era s.
+  ( AlonzoEraTx era
+  , Event (s era) ~ AlonzoUtxosEvent era
+  ) =>
+  PParams era ->
+  CertState era ->
+  TxBody era ->
+  Rule (s era) 'Transition Coin
+tellDepositChangeEvent pp dpstate txBody = do
+  {- refunded := keyRefunds pp txb -}
+  let refunded = keyTxRefunds pp dpstate txBody
+  {- depositChange := (totalDeposits pp poolParams txcerts txb) − refunded -}
+  let depositChange = totalTxDeposits pp dpstate txBody <-> refunded
+  tellEvent $ TotalDeposits (hashAnnotated txBody) depositChange
+  pure depositChange
+
+expectScriptsToPass ::
+  forall era s.
+  ( AlonzoEraTx era
+  , EraPlutusContext 'PlutusV1 era
+  , EraUTxO era
+  , ExtendedUTxO era
+  , Script era ~ AlonzoScript era
+  , ScriptsNeeded era ~ AlonzoScriptsNeeded era
+  , Tx era ~ AlonzoTx era
+  , STS (s era)
+  , Event (s era) ~ AlonzoUtxosEvent era
+  , PredicateFailure (s era) ~ AlonzoUtxosPredFailure era
+  , BaseM (s era) ~ ShelleyBase
+  ) =>
+  PParams era ->
+  AlonzoTx era ->
+  UTxO era ->
+  Rule (s era) 'Transition ()
+expectScriptsToPass pp tx utxo = do
+  sysSt <- liftSTS $ asks systemStart
+  ei <- liftSTS $ asks epochInfo
+  {- sLst := collectTwoPhaseScriptInputs pp tx utxo -}
+  case collectTwoPhaseScriptInputs ei sysSt pp tx utxo of
+    Right sLst -> do
+      {- isValid tx = evalScripts tx sLst = True -}
+      let protVer = pp ^. ppProtocolVersionL
+      whenFailureFree $
+        when2Phase $ case evalScripts @era protVer tx sLst of
+          Fails _ fs ->
+            failBecause $
+              ValidationTagMismatch
+                (tx ^. isValidTxL)
+                (FailedUnexpectedly (scriptFailuresToPredicateFailure protVer fs))
+          Passes ps -> mapM_ (tellEvent . SuccessfulPlutusScriptsEvent) (nonEmpty ps)
+    Left info -> failBecause (CollectErrors info)
+
 scriptsYes ::
   forall era.
   ( ExtendedUTxO era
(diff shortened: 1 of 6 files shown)


eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Utxo.hs
```diff
@@ -591,12 +591,12 @@ validateMaxTxSizeUTxO pp tx =
...
-  (EraTxBody era, GovernanceState era ~ ShelleyPPUPState era) =>
+  EraTxBody era =>
   PParams era ->
   UTxOState era ->
   TxBody era ->
   Coin ->
-  ShelleyPPUPState era ->
+  GovernanceState era ->
   UTxOState era
 updateUTxOState pp UTxOState {utxosUtxo, utxosDeposited, utxosFees, utxosStakeDistr} txb depositChange ppups =
```
lehins: I think, it would be nice to also rename the argument from `ppup` to `govState` or something

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Utxos.hs
```diff
@@ -1,59 +1,143 @@
...
+  , Show (PPUPPredFailure era)
+  ) =>
+  TransitionRule (ConwayUTXOS era)
+utxosTransition =
+  judgmentContext >>= \(TRC (_, _, tx)) -> do
+    case tx ^. isValidTxL of
+      IsValid True -> scriptsYes
+      IsValid False -> scriptsNo
+
+scriptsYes ::
```
lehins: Historically it caused bugs and confusion when there were functions with the same names in rules in different eras. I suggest renaming them (both in Alonzo and in Conway). Also `Yes` and `No` suffixes are a bit dumb IMHO, so renaming all of them to reflect the meaning would be superb. Something like this would be good, I think:

```suggestion
conwayEvalScriptsTxValid ::
```

Same for: `babbageEvalScriptsTxValid` and `babbageEvalScriptsTxInvalid`

eras/babbage/impl/src/Cardano/Ledger/Babbage/Rules/Utxos.hs
```diff
@@ -146,6 +148,60 @@ utxosTransition =
...
+
+expectScriptsToPass ::
+  forall era s.
+  ( AlonzoEraTx era
+  , EraPlutusContext 'PlutusV1 era
+  , EraUTxO era
+  , ExtendedUTxO era
+  , Script era ~ AlonzoScript era
+  , ScriptsNeeded era ~ AlonzoScriptsNeeded era
+  , Tx era ~ AlonzoTx era
```
lehins: This is redundant
```suggestion
```

eras/babbage/impl/src/Cardano/Ledger/Babbage/Rules/Utxos.hs
```diff
@@ -146,6 +148,60 @@ utxosTransition =
...
+  , STS (s era)
+  , Event (s era) ~ AlonzoUtxosEvent era
+  , PredicateFailure (s era) ~ AlonzoUtxosPredFailure era
+  , BaseM (s era) ~ ShelleyBase
+  ) =>
+  PParams era ->
+  AlonzoTx era ->
```
```suggestion
  Tx era ->
```

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Utxos.hs
```diff
@@ -1,59 +1,143 @@
...
+  ( AlonzoEraTx era
+  , BabbageEraTxBody era
+  , EraGovernance era
+  , EraPlutusContext 'PlutusV1 era
+  , EraUTxO era
+  , ExtendedUTxO era
+  , GovernanceState era ~ ConwayGovernance era
+  , Script era ~ AlonzoScript era
+  , ScriptsNeeded era ~ AlonzoScriptsNeeded era
+  , Tx era ~ AlonzoTx era
```
lehins: This will require changing `scriptsNo` to work on `Tx` instead of `AlonzoTx`
```suggestion
```

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Utxos.hs
```diff
@@ -1,59 +1,143 @@
...
   , EraTxCert era
-  , PredicateFailure (EraRule "UTXOS" era) ~ AlonzoUtxosPredFailure era
+  , EraUTxO era
+  , ExtendedUTxO era
   , Event (EraRule "UTXOS" era) ~ AlonzoUtxosEvent era
+  , GovernanceState era ~ ConwayGovernance era
+  , PredicateFailure (EraRule "UTXOS" era) ~ AlonzoUtxosPredFailure era
+  , Script era ~ AlonzoScript era
+  , ScriptsNeeded era ~ AlonzoScriptsNeeded era
+  , Tx era ~ AlonzoTx era
```
lehins:
```suggestion
```

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Utxos.hs
```diff
@@ -1,59 +1,143 @@
...
+  , STS (ConwayUTXOS era)
+  ) =>
+  TransitionRule (ConwayUTXOS era)
+scriptsYes = do
+  TRC (UtxoEnv _ pp dpstate _, u@(UTxOState utxo _ _ pup _), tx) <-
+    judgmentContext
+  let txBody = body tx
```
lehins: It is better to use lenses, this lets us avoid constraints like this: `Tx era ~ AlonzoTx era`
```suggestion
  let txBody = tx ^. bodyTxL
```

eras/babbage/impl/src/Cardano/Ledger/Babbage/Rules/Utxos.hs
```diff
@@ -146,6 +148,60 @@ utxosTransition =
...
+  UTxO era ->
+  Rule (s era) 'Transition ()
+expectScriptsToPass pp tx utxo = do
+  sysSt <- liftSTS $ asks systemStart
+  ei <- liftSTS $ asks epochInfo
+  {- sLst := collectTwoPhaseScriptInputs pp tx utxo -}
+  case collectTwoPhaseScriptInputs ei sysSt pp tx utxo of
+    Right sLst ->
+      {- isValid tx = evalScripts tx sLst = True -}
+      do
```
lehins: Consistency
```suggestion
    Right sLst -> do
      {- isValid tx = evalScripts tx sLst = True -}
```

lehins (changes requested):
Looks great. I'll do a final review later, but this for sure already takes care of UTXOS rule for Conway for us. Nice work!

eras/babbage/impl/src/Cardano/Ledger/Babbage/Rules/Utxos.hs
```diff
@@ -168,14 +224,7 @@ scriptsYes = do
   TRC (UtxoEnv slot pp dpstate genDelegs, u@(UTxOState utxo _ _ pup _), tx) <-
     judgmentContext
   let txBody = body tx
```
```suggestion
  let txBody = tx ^. bodyTxL
```

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Utxos.hs
teodanciu: The only way I could make this go away is by replacing it with this one:
` Signal (ConwayUTXOS era) ~ Tx era` .  It's still more generic, so i'm guessing it's better than `Tx era ~ AlonzoTx era~

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Utxos.hs
teodanciu: Same as above, I could only remove it if adding  `Signal (ConwayUTXOS era) ~ Tx era `

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Utxos.hs
lehins: Yes, that's the right way to go!

lehins (approved):
Looks great. Thank you

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Examples/ConwayFeatures.hs
lehins, line 1: I think this file got added by mistake, no?

It is not included in any cabal file

PR #3500, opened 2023-06-28
TxCert API

lehins:
# Description

Add basic API for TxCerts

# Checklist

- [x] Self-reviewed the diff


libs/cardano-ledger-api/src/Cardano/Ledger/Api/Tx/Cert.hs
```diff
@@ -0,0 +1,35 @@
...
+  pattern DelegTxCert,
+  pattern RegDepositDelegTxCert,
+  pattern AuthCommitteeHotKeyTxCert,
+  pattern ResignCommitteeColdTxCert,
+  pattern RegDRepTxCert,
+  pattern UnRegDRepTxCert,
+)
+where
+
+import Cardano.Ledger.Conway.TxCert
```
teodanciu: Just a haskell question: does this mean that the data constructors corresponding to these patterns (eg: ConwayRegCert, ConwayUnRegCert..etc) are also exported, or only the patterns?

teodanciu (approved):
Looks good to me

libs/cardano-ledger-api/src/Cardano/Ledger/Api/Tx/Cert.hs
lehins: They are imported (due to open import), but they are not re-exported from the API. Not even the concrete type `ConwayTxCert` is exported from the API, because we don't want to promote multiple ways of doing the same thing. It is better to export the most general (thus more powerful) way to deal with things.

That question reminds me that I should probably do explicit imports as well, just to help people reading the code infer where things are coming from.

lehins: Add explicit imports and improved haddock a bit.

teodanciu: Awesome, thanks for the explanation!

PR #3508, opened 2023-07-01
Fix voting procedure

lehins:
# Description

Switch from using Voting role at the value level to the type level.
This PR fixes a number of issues related to the `VotingProcedure`.

First and foremost the `Voting` key role is dedicated for DReps only,
but it was being used for Stake Pools and Consitutional Committee
members as well. Using the appropriate key roles makes tarcking of
VoterRole at the value level, not only redundant, but also wrong.

Votes where tracked in a single Map with `(VoterRole, Credential 'Voting c)`
as a key, which was inefficient and wrong at the same time. It was wrong
because it required key role casting, which actually led to a bug of
checking StakePool ids against Stake Credential Map.

Counting up the StakePool votes now uses the Stake Pool Distribution Map
instead of Stake Credential, which fixes the bug that was mentioned above.

ScriptHashes are no longer allowed for StakePool votes. This might be
reverted in the future, but hopefully not. See https://github.com/input-output-hk/cardano-ledger/issues/3512 for more discussion
on this issue.

Fix CDDL for changes to the new VotingProcedure field in the TxBody.

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/eras/conway/impl/src/Cardano/Ledger/Conway/Governance.hs
+++ b/eras/conway/impl/src/Cardano/Ledger/Conway/Governance.hs
@@ -3,10 +3,12 @@
 {-# LANGUAGE DerivingStrategies #-}
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE RecordWildCards #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE StandaloneDeriving #-}
+{-# LANGUAGE TupleSections #-}
 {-# LANGUAGE TypeApplications #-}
 {-# LANGUAGE TypeFamilies #-}
 {-# LANGUAGE UndecidableInstances #-}
@@ -24,13 +26,12 @@ module Cardano.Ledger.Conway.Governance (
   -- Lenses
   cgTallyL,
   cgRatifyL,
-  cgVoterRolesL,
   GovernanceAction (..),
   GovernanceActionState (..),
   GovernanceActionIx (..),
   GovernanceActionId (..),
   govActionIdToText,
-  VoterRole (..),
+  Voter (..),
   Vote (..),
   VotingProcedure (..),
   ProposalProcedure (..),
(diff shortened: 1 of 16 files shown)


eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Ratify.hs
```diff
@@ -84,32 +85,27 @@ instance
 
 --- Constants
 
-ccThreshold :: Int
-ccThreshold = 0
+-- ccThreshold :: Int
+-- ccThreshold = 0
```
teodanciu, line 89: The  ccThreshold check removal is also temporary, no? Maybe it would be good to add a comment explaining that this is why this is commented out.  We could also leave that check commented with a TODO so we don't forget about it?

teodanciu (approved):
Looks good to me

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Ratify.hs
lehins: We won't forget about this, because those hard coded thresholds themselves are temporary. The actual thresholds will be coming from PParams in the future.

eras/conway/impl/src/Cardano/Ledger/Conway/Governance.hs
```diff
@@ -301,23 +302,40 @@ instance EraPParams era => NFData (GovernanceProcedure era)
...
+    CommitteeVoter (KeyHashObj keyHash) ->
+      encodeListLen 2 <> encodeWord8 0 <> encCBOR keyHash
+    CommitteeVoter (ScriptHashObj scriptHash) ->
+      encodeListLen 2 <> encodeWord8 1 <> encCBOR scriptHash
+    DRepVoter (KeyHashObj keyHash) ->
+      encodeListLen 2 <> encodeWord8 2 <> encCBOR keyHash
+    DRepVoter (ScriptHashObj scriptHash) ->
+      encodeListLen 2 <> encodeWord8 3 <> encCBOR scriptHash
+    StakePoolVoter keyHash ->
+      encodeListLen 2 <> encodeWord8 4 <> encCBOR keyHash
```
Soupstraw, line 334: Why not use the `Coders` library here?

lehins: For two reasons:

* It will be more convoluted, since we are unwrapping the Credentials.
* There is nothing wrong with not using Coders, it is just a convenience wrapper around the regular encoders/decoders.

PR #3510, opened 2023-07-03
Don't zip `governanceProcedures` with infinite list in Tally

teodanciu:
# Description

The test is hanging in the version before this change.

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Tally.hs
+++ b/eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Tally.hs
@@ -7,7 +7,6 @@
 {-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE MultiParamTypeClasses #-}
 {-# LANGUAGE NamedFieldPuns #-}
-{-# LANGUAGE OverloadedLists #-}
 {-# LANGUAGE RankNTypes #-}
 {-# LANGUAGE RecordWildCards #-}
 {-# LANGUAGE ScopedTypeVariables #-}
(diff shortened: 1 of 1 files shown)


eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Tally.hs
```diff
@@ -184,7 +184,7 @@ tallyTransition = do
                   st'
         updateState st'' xs
 
-  updateState st $ Seq.zip govProcedures [0 ..]
+  updateState st $ Seq.zip govProcedures (fmap GovernanceActionIx [0 .. fromIntegral (Seq.length govProcedures - 1)])
```
lehins: `GovernanceActionIx` has a `Num` instance, so no extra conversions is needed.
```suggestion
  updateState st $ Seq.zip govProcedures [0 .. fromIntegral (Seq.length govProcedures - 1)]
```

lehins (approved):
Yeah, good catch, using infinite lists for Seq is not a good idea :smile:

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Tally.hs
teodanciu: Would it be maybe better to do 
`Seq.zip govProcedures (Seq.fromList [0 .. fromIntegral (Seq.length govProcedures - 1)])` 
and remove the OverloadedLists ?

lehins: They are one and the same thing. This is exactly what `OverloadedLists` will translate it into. So it is up to you.

PR #3511, opened 2023-07-03
Move `RATIFY` to `EPOCH`

### LINKED ISSUE #3504: Move call to RATIFY rule into EPOCH rule, from NEWEPOCH
RATIFY rule makes more sense in the EPOCH rule.

Also, an update to PParams: `esPrevPp` and `esPp` should happen in the EPOCH rule after the enactment happens. In the future #3503 will implement a proper solution, but right now we need to at least make it correct.

Soupstraw:
# Description

closes #3504

# Checklist

- [x] Self-reviewed the diff


lehins (approved):
This looks good. Needs to be fourmolized, but otherwise is ready to be merged.

PR #3523, opened 2023-07-06
Treasury donations

### LINKED ISSUE #3436: Transfer to the treasury
We require a mechanism that allows the transfer of lovelace from the UTxO to the treasury.

The current preferred solution is to add a new field to the transaction body which specifies the number of lovelace to transfer to the treasury, and let the `consumed == produced` check account for it (it would count as a positive amount to the consumed portion). The potentially tricky/messy aspect to this solution is not giving the UTxO rule access to changing the treasury and somehow propagating this information downstream to a rule with access to it.

Soupstraw:
# Description

Fixes #3436 

# Checklist

- [x] Self-reviewed the diff


Soupstraw:
Depends on UTXO added in #3495

eras/conway/impl/src/Cardano/Ledger/Conway/Core.hs
```diff
@@ -66,6 +66,8 @@ class BabbageEraTxBody era => ConwayEraTxBody era where
   -- | Lens for getting and setting `ProposalProcedures`.
   proposalProceduresTxBodyL :: Lens' (TxBody era) (StrictSeq (ProposalProcedure era))
 
+  treasuryDonationTxBodyL :: Lens' (TxBody era) Coin
```
lehins, line 69: I think it is better to have an explicit information about presence of this field in the transaction, instead of relying on this value to be 0 when no donation happens. In fact, I would even go as far as disallowing 0 amount at the serialization level or with a predicate check.
```suggestion
  treasuryDonationTxBodyL :: Lens' (TxBody era) (StrictMaybe Coin)
```

Soupstraw: Why so? It seems to me it would only complicate the code and add more potential points of failure.

lehins: Because the intention becomes explicit.

Supplying `0` and not supplying field at all should not mean the same thing.

lehins: As per our discussion, let's prevent zeros on the wire and we can get rid of the Maybe

eras/conway/impl/src/Cardano/Ledger/Conway/TxBody.hs
```diff
@@ -527,6 +538,7 @@ encodeTxBodyRaw ConwayTxBodyRaw {..} =
         !> Omit (null . unVotingProcedures) (Key 19 (To ctbrVotingProcedures))
         !> Omit null (Key 20 (To ctbrProposalProcedures))
         !> encodeKeyedStrictMaybe 21 ctbrCurrentTreasuryValue
+        !> Omit (== mempty) (Key 22 $ To ctbrTreasuryDonation)
```
lehins, line 546: You forgot to add the field to the decoder :smile: 

Make sure zeros are not allowed with https://github.com/input-output-hk/cardano-ledger/blob/2371e4651b5e90747813244860f6e4758513aaea/libs/cardano-ledger-core/src/Cardano/Ledger/Coin.hs#L147
```haskell
bodyFields 22 = ofield (\x tx -> tx {ctbrTreasuryDonation = x}) (D decodePositiveCoin)
```

lehins (approved):
Aside from the decoder, this PR is ready

PR #3526, opened 2023-07-10
Quviq ledger test work

MaximilianAlgehed:
@TimSheard please have a look at this as soon as you can. Note the approach we take to generating valid instances of `NewEpochState` in `LedgerTests.hs`. The neat thing is that most things work "out of the box" here.

However, I'm worried that the implementation of the constraints that I've been poking around with (in `Examples.hs`) are not the same instances that are used in the staged approach that you were working on. Let's discuss.


eras/conway/impl/cardano-ledger-conway.cabal
```diff
@@ -36,7 +36,9 @@ library
         Cardano.Ledger.Conway.TxCert
 
     hs-source-dirs:   src
-    other-modules:
+    -- TODO: this is temporarily exposed so that we can write tests for this code in the cardano-ledger-test library
```
lehins, line 39: You can get all the instances from `Cardano.Ledger.Conway`. You should not need to expose these modules

libs/cardano-ledger-test/cardano-ledger-test.cabal
```diff
@@ -81,12 +89,15 @@ library
         bytestring,
         cardano-data,
         cardano-crypto-class,
+        cardano-crypto-wrapper,
+        cardano-crypto,
```
lehins, line 93: We don't do any Byron testing, so these packages should not be used

libs/cardano-ledger-test/cardano-ledger-test.cabal
```diff
@@ -81,12 +89,15 @@ library
...
         cardano-ledger-allegra,
         cardano-ledger-alonzo:{cardano-ledger-alonzo, testlib},
         cardano-ledger-alonzo-test,
         cardano-ledger-api,
         cardano-ledger-babbage:{cardano-ledger-babbage, testlib},
         cardano-ledger-babbage-test,
+        cardano-ledger-byron,
```
lehins, line 100: Same with this one

MaximilianAlgehed:
@TimSheard to avoid trouble with merging when I'm on holiday, could we merge this into your branch now?

lehins:
> @TimSheard to avoid trouble with merging when I'm on holiday, could we merge this into your branch now?

@MaximilianAlgehed It'll be better to do this as a separate PR into `master`. Tim's PR is already huge, we don't need to get it bigger.

We can merge it when you get back from vacation, no worries

MaximilianAlgehed:
@lehins either way you want it. What I'm worried about is efforts diverging. This introduces some changes to Tim's stuff that I'd rather not have to rebase again as it can get quite hairy. However, I understand the aversion.

PR #3527, opened 2023-07-14
Bump plutus deps to 1.8

michaelpj:
Routine bump


michaelpj:
Looks like we're missing a lower bound on cardano-crypto-class. Will revise and retry.

michaelpj:
In fact, I suspect we also need to install the BLST libraries in the CI, and that's why it picked an older version. Probably once I revise we'll see the pkgconfig issue turn up.

michaelpj:
Okay, now we're getting somewhere. I'm getting a test failure that I'm pretty sure is because we added new builtins and so the cost model is a different size. But this change should only take effect for PlutusV3, so I'm surprised to see the failure in the Babbage tests. Can someone give me a pointer?

michaelpj:
Right, so this version adds the BLS primitives. I foolishly assumed that that would be simple and it would all just line up automatically for V3, but apparently not. I'm not sure what I need to change.

lehins (approved):
This is great. Thank you!

PR #3531, opened 2023-07-17
Plutus type improvement

lehins:
# Description

This PR does not change any Plutus semantics for all eras except Conway. It brings some type safety to Plutus ledger types and cleans up the code a bit.

For Conway it prepares the code to make it clearer a bit, but only after we are in Conway (see the added comment in the code)

Major parts of this PR are:
* Adding `Plutus` data type and changing `PlutusScript !Language !ShortByteString` to `PlutusScript !Plutus`
* Adding `PlutusWithContext` to avoid gross usage of four tuples and reduce number of arguments to a few functions

Most of the changes where done with deprecation in mind.

See changelogs or individual commits for more info

# Checklist

- [x] Self-reviewed the diff


libs/cardano-ledger-core/src/Cardano/Ledger/Language.hs
```diff
@@ -37,12 +62,40 @@ import Data.Aeson (
...
 import Data.Text (Text)
 import Data.Typeable (Typeable)
 import GHC.Generics (Generic)
 import NoThunks.Class (NoThunks)
 
+data Plutus = Plutus
+  { plutusLanguage :: !Language
+  , plutusScript :: !BinaryPlutus
+  }
+  deriving stock (Eq, Ord, Show, Generic)
```
teodanciu, line 77: Just curious what's the reason for having this  and `BinaryPlutus` in core?   (`PlutusWithContext` has to stay in Alonzo I suppose, because of its dependency on plutus api)

teodanciu: Ah, I guess so we can use it in the `Phase2Script`  constructor

eras/alonzo/test-suite/cardano-ledger-alonzo-test.cabal
```diff
@@ -49,7 +49,7 @@ library
     build-depends:
         base >=4.14 && <4.19,
         bytestring,
-        cardano-ledger-alonzo:{cardano-ledger-alonzo, testlib} >=1.2,
+        cardano-ledger-alonzo:{cardano-ledger-alonzo, testlib} >=1.4,
         cardano-ledger-binary:{cardano-ledger-binary, testlib} >=1.0,
         cardano-ledger-core:{cardano-ledger-core, testlib} >=1.3,
```
teodanciu: Don't we want to restrict core to <1.6 like we did for the others?

teodanciu (approved):
Looks good to me.  Much easier to understand what's going on in PlutusScriptApi with the parameters grouped like this :+1:

libs/cardano-ledger-core/src/Cardano/Ledger/Language.hs
lehins: Yep, you got it :slightly_smiling_face:

eras/alonzo/test-suite/cardano-ledger-alonzo-test.cabal
lehins: Yeah, sure, why not.
Done.

PR #3551, opened 2023-07-19
Update tools and dependencies

lehins:
# Description

This PR updates:

* Index state for CHaPs
* Index state for hackage
* GHC versions in nix and Github Actions: 9.2.7 -> 9.2.8 and 9.6.1 -> 9.6.2
* hls to 2.0.0.0 -> 2.0.0.1 (needed for ghc update)
* hlint 3.5 -> 3.6.1
* fourmolu 0.10.1.0 -> 0.13.1.0 

Update to fourmolu requires a bunch of changes to source code due to @Soupstraw's awesome suggestion: https://github.com/fourmolu/fourmolu/issues/288

It is probably easier to review this PR one commit at a time, since it contains a bunch of reformatting in one of the commits

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -102,7 +102,7 @@
             // lib.optionalAttrs (config.compiler-nix-name == defaultCompiler) {
               # tools that work only with default compiler
               fourmolu = "0.10.1.0";
-              hlint = "3.5";
+              hlint = "3.6.1";
               haskell-language-server = { src = nixpkgs.haskell-nix.sources."hls-2.0"; };
             };
 
--- a/flake.nix
+++ b/flake.nix
@@ -103,7 +103,7 @@
               # tools that work only with default compiler
               fourmolu = "0.10.1.0";
               hlint = "3.5";
-              haskell-language-server = "2.0.0.0";
+              haskell-language-server = { src = nixpkgs.haskell-nix.sources."hls-2.0"; };
             };
 
           # and from nixpkgs or other inputs
--- a/flake.nix
+++ b/flake.nix
@@ -101,7 +101,7 @@
             }
             // lib.optionalAttrs (config.compiler-nix-name == defaultCompiler) {
               # tools that work only with default compiler
-              fourmolu = "0.10.1.0";
+              fourmolu = "0.13.1.0";
               hlint = "3.6.1";
               haskell-language-server = { src = nixpkgs.haskell-nix.sources."hls-2.0"; };
             };
--- a/flake.nix
+++ b/flake.nix
@@ -47,7 +47,7 @@
         inherit (nixpkgs) lib;
 
         # see flake `variants` below for alternative compilers
-        defaultCompiler = "ghc927";
+        defaultCompiler = "ghc928";
         # We use cabalProject' to ensure we don't build the plan for
         # all systems.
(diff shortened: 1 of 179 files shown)


flake.nix
```diff
@@ -103,7 +103,7 @@
               # tools that work only with default compiler
               fourmolu = "0.10.1.0";
               hlint = "3.5";
-              haskell-language-server = "2.0.0.0";
+              haskell-language-server = { src = nixpkgs.haskell-nix.sources."hls-2.0"; };
```
teodanciu, line 106: Just curious what the reason for this is?  Is it working better like this?

fourmolu.yaml
Soupstraw, line 1: Fourmolu now also supports [line length limits](https://github.com/fourmolu/fourmolu/pull/305), maybe we should configure that as well?

teodanciu:
For ghc 9.2.8 and 9.6.2 it fails to compile some dependency (dependent sum). I see that 
* this library  is pulled in by plutus, and 
* plutus doesn't seem to have upgraded to 9.2.8 and 9.6.2

 Wonder whether this is why we can't build this?

fourmolu.yaml
Soupstraw: the downside to that is that it might be necessary to run fourmolu multiple times before the CI is happy with it.

lehins:
That is a great idea. However, I'd rather do it in a separate PR, since it will also result in a huge diff

lehins:
Yeah, it is weird, I had the same problem locally and then it went away. I'll try to figure out what needs to be done to get it fixed

fourmolu.yaml
aniketd: > Line-length limits

In my opinion, this would be a very welcome addition.

> might be necessary to run ... multiple times

@Soupstraw I'm surprised by this. I don't know how it works and why it won't be able to do the "right thing" the first time.

lehins:
> Wonder whether this is why we can't build this?

@teodanciu Figured it out. I needed to set a slightly newer index-state to CHaPs, because this revision actually fixes the problem: https://github.com/input-output-hk/cardano-haskell-packages/blob/7ac3e03766d57b53ab1f0abe27d17ccff9840f5e/_sources/plutus-core/1.8.0.0/revisions/2.cabal#L276

flake.nix
lehins: Honestly, I have no idea. This is what @hamishmack did in this PR: https://github.com/input-output-hk/cardano-ledger/pull/3479

So, I just copied it over.

If I had to guess this will use the latest 2.0 version from haskell-nix-sources, instead of us bumping it manually for every non-breaking release. For example `2.0.0.1` is needed for ghc-9.2.8 and 9.6.2.

teodanciu (approved):
Looks good to me

fourmolu.yaml
lehins: I responded through email and it didn't get placed here as a comment: https://github.com/input-output-hk/cardano-ledger/pull/3551#issuecomment-1641951038

We do wanna do the auto line wrap, but later when we have some spare time and in a separate PR.

I just wanted to update all of the tools while I was at updating GHC versions, which we do need.

PR #3556, opened 2023-07-20
Add Script to Constitution

### LINKED ISSUE #3532: Add Script to Constitution
It has been decided that we will want an optional Constitution that is runnable.

Therefore GovernanceAction should look like this:

```haskell
 NewConstitution !(SafeHash (EraCrypto era) ByteString) !(StrictMaybe (ScriptHash (EraCrypto era)))
```

Until the feature is fully implemented any script in the `NewConstitution` GovAction will be rejected by the rules.

aniketd:
# Description

Resolves #3532 

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/eras/conway/impl/CHANGELOG.md
+++ b/eras/conway/impl/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## 1.7.0.0
 
+* Change EnactState.ensConstitution #3556
+  * from `SafeHash (EraCrypto era) ByteString`
+  * to `Constitution era`
+  * Use this datatype for GovernanceAction.NewConstitution
 * Change `ConwayCommitteeCert` to allow committee hot keys to be script-hashes #3552
   * from `ConwayAuthCommitteeHotKey (KeyHash 'CommitteeColdKey c) (KeyHash 'CommitteeHotKey c)`
   * to `ConwayAuthCommitteeHotKey (KeyHash 'CommitteeColdKey c) (Credential 'CommitteeHotKey c)`
--- a/eras/conway/impl/src/Cardano/Ledger/Conway/Governance.hs
+++ b/eras/conway/impl/src/Cardano/Ledger/Conway/Governance.hs
@@ -38,6 +38,10 @@ module Cardano.Ledger.Conway.Governance (
   AnchorDataHash,
   ensConstitutionL,
   rsEnactStateL,
+  Constitution (..),
+  ConstitutionData (..),
+  constitutionHashL,
+  constitutionScriptL,
 ) where
 
 import Cardano.Ledger.BaseTypes (EpochNo (..), ProtVer (..), StrictMaybe)
@@ -73,11 +77,9 @@ import Cardano.Ledger.Conway.Governance.Procedures (
 import Cardano.Ledger.Core
 import Cardano.Ledger.Credential (Credential (..))
 import Cardano.Ledger.Keys (KeyHash (..), KeyRole (..))
-import Cardano.Ledger.SafeHash (SafeHash)
 import Cardano.Ledger.Shelley.Governance
 import Control.DeepSeq (NFData)
 import Data.Aeson (KeyValue, ToJSON (..), object, pairs, (.=))
-import Data.ByteString (ByteString)
 import Data.Default.Class (Default (..))
 import Data.Map.Strict (Map)
 import Data.Sequence.Strict (StrictSeq)
@@ -174,7 +176,7 @@ instance EraPParams era => FromCBOR (ConwayGovState era) where
 data EnactState era = EnactState
   { ensCommittee :: !(StrictMaybe (Set (KeyHash 'Voting (EraCrypto era)), Rational))
   -- ^ Constitutional Committee
-  , ensConstitution :: !(SafeHash (EraCrypto era) ByteString)
+  , ensConstitution :: !(Constitution era)
   -- ^ Hash of the Constitution
   , ensProtVer :: !ProtVer
   , ensPParams :: !(PParams era)
@@ -183,7 +185,7 @@ data EnactState era = EnactState
   }
   deriving (Generic)
 
-ensConstitutionL :: Lens' (EnactState era) (SafeHash (EraCrypto era) ByteString)
+ensConstitutionL :: Lens' (EnactState era) (Constitution era)
 ensConstitutionL = lens ensConstitution (\x y -> x {ensConstitution = y})
 
 instance EraPParams era => ToJSON (EnactState era) where
@@ -344,4 +346,4 @@ toConwayGovernancePairs cg@(ConwayGovernance _ _) =
 
 instance EraPParams (ConwayEra c) => EraGovernance (ConwayEra c) where
   type GovernanceState (ConwayEra c) = ConwayGovernance (ConwayEra c)
-  getConstitutionHash g = Just $ g ^. cgRatifyL . rsEnactStateL . ensConstitutionL
+  getConstitutionHash g = Just $ g ^. cgRatifyL . rsEnactStateL . ensConstitutionL . constitutionHashL
--- a/eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Procedures.hs
+++ b/eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Procedures.hs
@@ -28,7 +28,7 @@ module Cardano.Ledger.Conway.Governance.Procedures (
   govActionIdToText,
 ) where
 
-import Cardano.Crypto.Hash (ByteString, hashToTextAsHex)
+import Cardano.Crypto.Hash (hashToTextAsHex)
 import Cardano.Ledger.BaseTypes (ProtVer, Url)
 import Cardano.Ledger.Binary (
   DecCBOR (..),
@@ -48,12 +48,14 @@ import Cardano.Ledger.Credential (Credential (..))
 import Cardano.Ledger.Crypto (Crypto)
 import Cardano.Ledger.Keys (KeyHash, KeyRole (..))
 import Cardano.Ledger.SafeHash (SafeHash, extractHash)
+import Cardano.Ledger.Shelley.Governance (Constitution)
+import Cardano.Ledger.Shelley.RewardProvenance ()
 import Cardano.Ledger.TxIn (TxId (..))
 import Control.DeepSeq (NFData (..), rwhnf)
 import Data.Aeson (KeyValue (..), ToJSON (..), ToJSONKey (..), object, pairs)
 import Data.Aeson.Types (toJSONKeyText)
 import Data.Map.Strict (Map)
-import Data.Maybe.Strict (StrictMaybe)
+import Data.Maybe.Strict (StrictMaybe (..))
 import Data.Sequence (Seq)
 import Data.Set (Set)
 import qualified Data.Text as Text
@@ -302,7 +304,7 @@ data GovernanceAction era
   | TreasuryWithdrawals !(Map (Credential 'Staking (EraCrypto era)) Coin)
   | NoConfidence
   | NewCommittee !(Set (KeyHash 'Voting (EraCrypto era))) !Rational
-  | NewConstitution !(SafeHash (EraCrypto era) ByteString)
+  | NewConstitution !(Constitution era)
   | InfoAction
   deriving (Generic)
 
@@ -338,5 +340,5 @@ instance EraPParams era => EncCBOR (GovernanceAction era) where
       enc (TreasuryWithdrawals ws) = Sum TreasuryWithdrawals 2 !> To ws
       enc NoConfidence = Sum NoConfidence 3
       enc (NewCommittee mems quorum) = Sum NewCommittee 4 !> To mems !> To quorum
-      enc (NewConstitution h) = Sum NewConstitution 5 !> To h
+      enc (NewConstitution c) = Sum NewConstitution 5 !> To c
       enc InfoAction = Sum InfoAction 6
--- a/eras/shelley/impl/CHANGELOG.md
+++ b/eras/shelley/impl/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Version history for `cardano-ledger-shelley`
 
+## 1.5.0.0
+
+* Make new `Constitution` datatype and `ConstitutionData` newtype #3556
+  * Adopt some Default instances for example SafeHash
+
 ## 1.4.2.1
 
 *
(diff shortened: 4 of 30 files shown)


aniketd:
Please especially review the CDDL change. I think I may have got it wrong.

eras/conway/impl/src/Cardano/Ledger/Conway/Governance.hs
```diff
@@ -176,7 +176,7 @@ instance EraPParams era => FromCBOR (ConwayGovState era) where
 data EnactState era = EnactState
   { ensCommittee :: !(StrictMaybe (Set (KeyHash 'Voting (EraCrypto era)), Rational))
   -- ^ Constitutional Committee
-  , ensConstitution :: !(SafeHash (EraCrypto era) ByteString)
+  , ensConstitution :: !(SafeHash (EraCrypto era) ByteString, StrictMaybe (ScriptHash (EraCrypto era)))
```
Soupstraw: I'd like it more if it was a separate field in the `EnactState`.

aniketd: Noted!

Soupstraw: Either that or we create a datatype called `Constitution` which holds both. I think this way is better if we want to throw out the constitution script together with the human-readable constitution, which I think is how it's intended.

lehins: Yes, let's create a separate `Constitution` type.

eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Procedures.hs
```diff
@@ -326,7 +327,7 @@ instance EraPParams era => DecCBOR (GovernanceAction era) where
       dec 2 = SumD TreasuryWithdrawals <! From
       dec 3 = SumD NoConfidence
       dec 4 = SumD NewCommittee <! From <! From
-      dec 5 = SumD NewConstitution <! From
+      dec 5 = SumD NewConstitution <! From <! From
```
lehins: In order to use Null, you need to use `decodeNullStrictMaybe`
```suggestion
      dec 5 = SumD NewConstitution <! From <! D (decodeNullStrictMaybe decCBOR)
```

eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Procedures.hs
```diff
@@ -338,5 +339,5 @@ instance EraPParams era => EncCBOR (GovernanceAction era) where
       enc (TreasuryWithdrawals ws) = Sum TreasuryWithdrawals 2 !> To ws
       enc NoConfidence = Sum NoConfidence 3
       enc (NewCommittee mems quorum) = Sum NewCommittee 4 !> To mems !> To quorum
-      enc (NewConstitution h) = Sum NewConstitution 5 !> To h
+      enc (NewConstitution h s) = Sum NewConstitution 5 !> To h !> To s
```
lehins: Same with encoding.
```suggestion
      enc (NewConstitution h s) = Sum NewConstitution 5 !> To h !> E (encodeNullStrictMaybe encCBOR) s
```

lehins (changes requested):
I like the idea for creating a separate `Constritution` type and reusing in all places in this PR.
Note that my comments about serialization will have to travel to the serialization of the `Constitution` type

eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Procedures.hs
aniketd: Super thanks! :+1:

eras/conway/impl/src/Cardano/Ledger/Conway/Governance.hs
```diff
@@ -183,23 +187,22 @@ data EnactState era = EnactState
...
-ensConstitutionL :: Lens' (EnactState era) (SafeHash (EraCrypto era) ByteString)
+ensConstitutionL :: Lens' (EnactState era) (Constitution era)
 ensConstitutionL = lens ensConstitution (\x y -> x {ensConstitution = y})
 
 instance EraPParams era => ToJSON (EnactState era) where
   toJSON = object . toEnactStatePairs
   toEncoding = pairs . mconcat . toEnactStatePairs
 
 toEnactStatePairs :: (KeyValue a, EraPParams era) => EnactState era -> [a]
-toEnactStatePairs cg@(EnactState _ _ _ _ _ _) =
```
lehins: Can you pleas bring this back.

This is important, because pattern match like this protects us against forgetting to adjust the implementation when new fields are added.

eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Procedures.hs
```diff
@@ -296,13 +304,66 @@ instance EraPParams era => EncCBOR (ProposalProcedure era) where
...
+
+constitutionScriptL :: Lens' (Constitution era) (StrictMaybe (ScriptHash (EraCrypto era)))
+constitutionScriptL = lens constitutionScript (\x y -> x {constitutionScript = y})
+
+instance Era era => ToJSON (Constitution era) where
+  toJSON = object . toConstitutionPairs
+  toEncoding = pairs . mconcat . toConstitutionPairs
+
+toConstitutionPairs :: (KeyValue a, Era era) => Constitution era -> [a]
+toConstitutionPairs Constitution {..} =
```
lehins: Just as in my previous comment, this approach allows us to rely on compiler to remind us about future changes to the type.
```suggestion
toConstitutionPairs c@(Constitution _ _) =
  let Constitution {..} = cg in
```

eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Procedures.hs
```diff
@@ -296,13 +304,66 @@ instance EraPParams era => EncCBOR (ProposalProcedure era) where
...
+  [ "constitutionHash" .= constitutionHash
+  , "constitutionScript" .= constitutionScript
+  ]
```
lehins: It might be better to omit the script hash if is not present
```suggestion
  ] ++ [ "constitutionScript" .= script | SJust script <- constitutionScript]
```

eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Procedures.hs
```diff
@@ -296,13 +304,66 @@ instance EraPParams era => EncCBOR (ProposalProcedure era) where
         !> To pProcGovernanceAction
         !> To pProcAnchor
 
+data Constitution era = Constitution
+  { constitutionHash :: !(SafeHash (EraCrypto era) ByteString)
```
lehins: Because we never do the actual hashing or verification we can set the phantom type to whatever we like (even if it is recursive like that). And ByteString is a lot worse than `Constitution`
```suggestion
  { constitutionHash :: !(SafeHash (EraCrypto era) Constitution)
```

lehins (changes requested):
Couple more minor changes is needed

eras/conway/impl/src/Cardano/Ledger/Conway/Governance.hs
aniketd: Oh, got it! Noted for henceforth!

eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Procedures.hs
aniketd: TIL! Thanks Alexey! :raised_hands:

eras/conway/impl/CHANGELOG.md
```diff
@@ -2,11 +2,15 @@
...
+  * Use this datatype for GovernanceAction.NewConstitution
 * Add `ConwayPParams` #3498
-* Add `UpgradeConwayPParams` #3498
-* Add `ConwayEraPParams` #3498
-* Add `PoolVotingThresholds` #3498
-* Add `DRepVotingThresholds` #3498
+  * Add `UpgradeConwayPParams`
+  * Add `ConwayEraPParams`
+  * Add `PoolVotingThresholds`
+  * Add `DRepVotingThresholds`
```
aniketd: This is just a formatting change. It groups the changes under one PR.

eras/conway/impl/CHANGELOG.md
```diff
@@ -46,7 +50,7 @@
   `gasStakePoolVotes` in `GovernanceActionState`
 * Removal of `reRoles` from `RatifyEnv` as no longer needed
 * Addition of `reStakePoolDistr` to `RatifyEnv`
-* Remove `VoterDoesNotHaveRole` as no longer needed from `ConwayTallyPredFailure`
+* Remove `VoterDoesNotHaveRole` as no longer needed from `ConwayGovPredFailure`
```
lehins: In that version it was still called Tally
```suggestion
* Remove `VoterDoesNotHaveRole` as no longer needed from `ConwayTallyPredFailure`
```

eras/shelley/impl/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Version history for `cardano-ledger-shelley`
 
+## 1.5.0.0
+
+* Make new `Constitution` data type #3556
+  * Move Default instance for SafeHash to the SafeHash module
```
lehins:
```suggestion
  * Move Default instance for SafeHash to the SafeHash module

```

eras/shelley/impl/src/Cardano/Ledger/Shelley/RewardProvenance.hs
```diff
@@ -190,9 +189,6 @@ instance Crypto e => Default (Credential r e) where
 instance Crypto c => Default (RewardAcnt c) where
   def = RewardAcnt def def
```
lehins: Can you move the rest of the orphans to their respective types as well? and get rid of this flag at the top of the module `{-# OPTIONS_GHC -Wno-orphans #-}`

eras/shelley/impl/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Version history for `cardano-ledger-shelley`
 
+## 1.5.0.0
```
lehins, line 3: Removing an instance is a major version change, so this is correct.

But, it also needs a bumped in the `cardano-ledger-shelley.cabal` and bounds updated in all of the dependencies.

libs/cardano-ledger-api/src/Cardano/Ledger/Api/State/Query.hs
```diff
@@ -52,6 +51,6 @@ getDState :: NewEpochState era -> DState era
 getDState = certDState . lsCertState . esLState . nesEs
 
 queryConstitutionHash ::
-  EraGovernance era => NewEpochState era -> Maybe (SafeHash (EraCrypto era) ByteString)
+  EraGovernance era => NewEpochState era -> Maybe (SafeHash (EraCrypto era) (Constitution era))
```
lehins: So, this is wrong for two reasons:

* This will break query serialization, thus breaking query interface
* Function is named `queryConstitutionHash`, but it returns the `Constitution`. 

What we want to do is leave the function type as it was before, make it return the hash and add a new query that returns the actual `Constitution`

lehins (changes requested):
Looks good. A few changes are needed however.

eras/shelley/impl/src/Cardano/Ledger/Shelley/RewardProvenance.hs
aniketd: Done! :raised_hands:

libs/cardano-ledger-api/src/Cardano/Ledger/Api/State/Query.hs
aniketd: The `Constitution` happens to be a phantom type here, very confusing to notice :slightly_smiling_face:. But we don't want to break the API :100:  Will a single `castSafeHash` save the day here? I have pushed this change. Let me know what you think :raised_hands:

eras/shelley/impl/src/Cardano/Ledger/Shelley/Governance.hs
```diff
@@ -63,7 +69,7 @@ class
   getProposedPPUpdates _ = Nothing
 
   -- | Returns `Nothing` for all era preceding Conway, otherwise returns the hash of the constitution
-  getConstitutionHash :: GovernanceState era -> Maybe (SafeHash (EraCrypto era) ByteString)
+  getConstitutionHash :: GovernanceState era -> Maybe (SafeHash (EraCrypto era) (Constitution era))
```
lehins: This also needs to return a Hash, not `Constitution`

eras/shelley/impl/src/Cardano/Ledger/Shelley/Governance.hs
```diff
@@ -119,3 +125,56 @@ toPPUPStatePairs ShelleyPPUPState {..} =
 
 instance Default (ShelleyPPUPState era) where
   def = ShelleyPPUPState emptyPPPUpdates emptyPPPUpdates
+
+data Constitution era = Constitution
+  { constitutionHash :: !(SafeHash (EraCrypto era) (Constitution era))
```
lehins:
```haskell
newtype ConstitutionData = ConstitutionData ByteString
```

```suggestion
  { constitutionHash :: !(SafeHash (EraCrypto era) ConstitutionData)
```

libs/cardano-ledger-api/src/Cardano/Ledger/Api/State/Query.hs
```diff
@@ -54,4 +54,4 @@ getDState = certDState . lsCertState . esLState . nesEs
 queryConstitutionHash ::
   EraGovernance era => NewEpochState era -> Maybe (SafeHash (EraCrypto era) ByteString)
 queryConstitutionHash nes =
-  getConstitutionHash $ nes ^. nesEpochStateL . esLStateL . lsUTxOStateL . utxosGovernanceL
+  castSafeHash <$> getConstitutionHash (nes ^. nesEpochStateL . esLStateL . lsUTxOStateL . utxosGovernanceL)
```
```suggestion
  getConstitutionHash (nes ^. nesEpochStateL . esLStateL . lsUTxOStateL . utxosGovernanceL)
```

libs/cardano-ledger-api/src/Cardano/Ledger/Api/State/Query.hs
```diff
@@ -54,4 +54,4 @@ getDState = certDState . lsCertState . esLState . nesEs
 queryConstitutionHash ::
   EraGovernance era => NewEpochState era -> Maybe (SafeHash (EraCrypto era) ByteString)
```
```suggestion
  EraGovernance era => NewEpochState era -> Maybe (SafeHash (EraCrypto era) ConstitutionData)
```

lehins (changes requested):
Some more feedback, as we discussed on a call. It is better to just introduce a `ConstituitonData` instead of using the `Constitution era` as the phantom type for the Hash

lehins (approved):
Awesome work!

PR #3557, opened 2023-07-20
Update `NewCommittee` action to use Credential and add more info

### LINKED ISSUE #3543: `NewCommittee` proposal needs more info
We need to add a set for the previous committee and the epoch number when the new committee member should expire. Also the key role needs fixing. So, instead of current

https://github.com/input-output-hk/cardano-ledger/blob/319f0f3e5bb9ee8411f059f72e31dcd30a92afc4/eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Procedures.hs#L304

```haskell
  | NewCommittee
    !(Set (Credential 'CommitteeCold (EraCrypto era))) -- Old Committee
    !(Map (Credential 'CommitteeCold (EraCrypto era)) EpochNo) -- New Committee with expiration
    !Rational -- Quorum
```

teodanciu:
namely: old committee and the expiration epoch for each new member. Also update `EnactState` to reflect these changes

Closes https://github.com/input-output-hk/cardano-ledger/issues/3543

Draft because I haven't figured out how to update scriptsNeeded for this

# Description

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/eras/conway/impl/src/Cardano/Ledger/Conway/Governance.hs
+++ b/eras/conway/impl/src/Cardano/Ledger/Conway/Governance.hs
@@ -21,6 +21,7 @@ module Cardano.Ledger.Conway.Governance (
   EnactState (..),
   RatifyState (..),
   ConwayGovernance (..),
+  Committee (..),
   GovernanceAction (..),
   GovernanceActionState (..),
   GovernanceActionIx (..),
@@ -66,6 +67,7 @@ import Cardano.Ledger.Conway.Era (ConwayEra)
 import Cardano.Ledger.Conway.Governance.Procedures (
   Anchor (..),
   AnchorDataHash,
+  Committee (..),
   GovernanceAction (..),
   GovernanceActionId (..),
   GovernanceActionIx (..),
@@ -87,7 +89,6 @@ import Data.Aeson (KeyValue, ToJSON (..), object, pairs, (.=))
 import Data.Default.Class (Default (..))
 import Data.Map.Strict (Map)
 import Data.Sequence.Strict (StrictSeq)
-import Data.Set (Set)
 import GHC.Generics (Generic)
 import Lens.Micro (Lens', lens, (^.))
 import NoThunks.Class (NoThunks)
@@ -178,7 +179,7 @@ instance EraPParams era => FromCBOR (ConwayGovState era) where
   fromCBOR = fromEraCBOR @era
 
 data EnactState era = EnactState
-  { ensCommittee :: !(StrictMaybe (Set (KeyHash 'DRepRole (EraCrypto era)), Rational))
+  { ensCommittee :: !(StrictMaybe (Committee era))
   -- ^ Constitutional Committee
   , ensConstitution :: !(Constitution era)
   -- ^ Hash of the Constitution
--- a/eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Procedures.hs
+++ b/eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Procedures.hs
@@ -5,6 +5,7 @@
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
 {-# LANGUAGE LambdaCase #-}
+{-# LANGUAGE NamedFieldPuns #-}
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE RecordWildCards #-}
 {-# LANGUAGE ScopedTypeVariables #-}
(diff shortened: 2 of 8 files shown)


eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Procedures.hs
```diff
@@ -301,7 +302,13 @@ data GovernanceAction era
   | HardForkInitiation !ProtVer
   | TreasuryWithdrawals !(Map (Credential 'Staking (EraCrypto era)) Coin)
   | NoConfidence
-  | NewCommittee !(Set (KeyHash 'Voting (EraCrypto era))) !Rational
+  | NewCommittee
+      -- | Old committee
+      !(Set (Credential 'CommitteeColdKey (EraCrypto era)))
+      -- | New Committee with expiration epoch
+      !(Map (Credential 'CommitteeColdKey (EraCrypto era)) EpochNo)
+      -- | Quorum
```
lehins: I think haddock will choke otherwise
```suggestion
      -- Old committee
      !(Set (Credential 'CommitteeColdKey (EraCrypto era)))
      -- New Committee with expiration epoch
      !(Map (Credential 'CommitteeColdKey (EraCrypto era)) EpochNo)
      -- Quorum
```

eras/conway/impl/src/Cardano/Ledger/Conway/Governance.hs
```diff
@@ -172,7 +171,7 @@ instance EraPParams era => FromCBOR (ConwayGovState era) where
   fromCBOR = fromEraCBOR @era
 
 data EnactState era = EnactState
-  { ensCommittee :: !(StrictMaybe (Set (KeyHash 'Voting (EraCrypto era)), Rational))
+  { ensCommittee :: !(StrictMaybe (Map (Credential 'CommitteeColdKey (EraCrypto era)) EpochNo, Rational))
```
lehins: Now looking at the code, makes me realize it would be a good idea to create a type for Committee (especially since tuples are lazy and we want our state to be fully strict:
```suggestion
  { ensCommittee :: !(StrictMaybe (Committe era)
```

where

```haskell
data Committee era = Committee 
  { committeeMembers :: !(Map (Credential 'CommitteeColdKey (EraCrypto era)) EpochNo)
  , committeeQuorum :: !Rational
  }
```

Then this type can also be used in the `NewCommittee` type:

```haskell
  | NewCommittee
      -- | Old committee
      !(Set (Credential 'CommitteeColdKey (EraCrypto era)))
      -- | New Committee with expiration epoch and quorum
      !(Committee era)
```

lehins (changes requested):
I should have said this in the ticket, but it was only after I looked at the code I realized that it would be good to add a whole new type for the `Committee`

eras/conway/impl/src/Cardano/Ledger/Conway/Governance.hs
teodanciu: Good idea, thanks

aniketd (commented):
Just FYI, there is WIP to make the `CommitteeColdKey` from a `KeyHash` to a `Credential` #3564

eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Procedures.hs
lehins: Nevermind, I was totally wrong. Looks like haddock is capable of handling haddock like this nowadays. Which is great!

lehins (approved):
:+1:

eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Procedures.hs
```diff
@@ -343,12 +343,53 @@ instance EraPParams era => EncCBOR (ProposalProcedure era) where
...
+
 data GovernanceAction era
   = ParameterChange !(PParamsUpdate era)
   | HardForkInitiation !ProtVer
   | TreasuryWithdrawals !(Map (Credential 'Staking (EraCrypto era)) Coin)
   | NoConfidence
-  | NewCommittee !(Set (KeyHash 'Voting (EraCrypto era))) !Rational
+  | NewCommittee
+      -- | Old committee
+      !(Set (Credential 'CommitteeColdKey (EraCrypto era)))
```
aniketd: I may not have the complete context here, so I'm assuming we are not keeping `Committee era` here for the old committee because the old `UnitInterval` (quorum) or the members' respective retirement `EpochNo`s are not relevant in the `GovernanceAction`.

aniketd (approved):
Just a single doubt (question), but otherwise looks good! :raised_hands:

eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Procedures.hs
lehins: Yes, that is correct, as far as I understand it as well.

eras/conway/impl/src/Cardano/Ledger/Conway/Governance.hs
lehins: Implemented

PR #3570, opened 2023-07-24
Changed validateCollateralContainsNonADA so its clearer.

TimSheard:
Changed validateCollateralContainsNonADA, so the logic is easier to follow.

Small change to just 1 file. The logic for validateCollateralContainsNonADA was complicated and hard to follow,
because of nested case statements, if statements, failureUnless, and  logical conjunction.
Now it is broken into a Case statement with exactly 5 cases, and each case is clear.
This should be a non breaking change, as the function computes the same answer.

- [ x] Commit sequence broadly makes sense and commits have useful messages
- [x] Self-reviewed the diff

### DIFF
--- a/eras/babbage/impl/src/Cardano/Ledger/Babbage/Rules/Utxo.hs
+++ b/eras/babbage/impl/src/Cardano/Ledger/Babbage/Rules/Utxo.hs
@@ -233,58 +233,32 @@ validateCollateralContainsNonADA ::
   Map.Map (TxIn (EraCrypto era)) (TxOut era) ->
   Test (AlonzoUtxoPredFailure era)
 validateCollateralContainsNonADA txBody utxoCollateral =
-    -- When we do not have any non-ada TxOuts we can short-circuit the more
-    -- expensive validation, which has to compute the full balance on the Values
-    -- not just the Coin. So the (True,True) case, is cheap, and the computations
-    -- of colInputsBal and totalBal in the where clause are not computed at all
-    case (areAllAdaOnly utxoCollateral , areAllAdaOnly (txBody ^. collateralReturnTxBodyL)) of
-       (True,True) -> pure ()
-       (False,False) ->
-           if (Val.isAdaOnly totalBal) 
-              then pure () -- Both have MultiSssets, and they cancel out.
-              else failureIf True $ CollateralContainsNonADA colInputBal
-       (True,False) -> case txBody ^. collateralReturnTxBodyL of
-           SNothing -> pure ()
-                       -- This should be impossible since the collateralReturn has some NonAda
-                       -- So it can't be Nothing. If by some miracle it ever happens then
-                       -- the (pure ()) is the right answer, since in this branch the
-                       -- collateral inputs, have no Ada too.
-           SJust retTxOut -> failureIf True $ CollateralContainsNonADA  (retTxOut ^. valueTxOutL)
-       (False,True) -> failureIf True $ CollateralContainsNonADA colInputBal
-   where
+  -- When we do not have any non-ada TxOuts we can short-circuit the more expensive
+  -- validation, which has to compute the full balance on the Values, not just the Coin.
+  -- By using 'areAllAdaOnly', the (True,True) case, this approach is both cheap and common,
+  -- and the computations of colInputsBal and totalBal, in the where clause, are not computed at all.
+  case (areAllAdaOnly utxoCollateral, areAllAdaOnly (txBody ^. collateralReturnTxBodyL)) of
+    (True, True) -> pure ()
+    (False, False) ->
+      if (Val.isAdaOnly totalBal)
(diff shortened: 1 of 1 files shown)


eras/babbage/impl/src/Cardano/Ledger/Babbage/Rules/Utxo.hs
```diff
@@ -233,24 +233,32 @@ validateCollateralContainsNonADA ::
...
-  -- not just the Coin.
-  if areAllAdaOnly utxoCollateral && areAllAdaOnly (txBody ^. collateralReturnTxBodyL)
-    then pure ()
-    else failureUnless (Val.isAdaOnly bal) $
-      case txBody ^. collateralReturnTxBodyL of
-        SJust retTxOut
-          | not (Val.isAdaOnly colbal) ->
-              CollateralContainsNonADA (retTxOut ^. valueTxOutL)
-        _ -> CollateralContainsNonADA colbal
+  -- When we do not have any non-ada TxOuts we can short-circuit the more expensive
```
teodanciu, line 236: "We do not have any non-ada TxOuts"  means we have ada-only TxOuts, correct?  Or maybe: we don't have multiassets?  Might be easier to understand without the double-negative, unless it's expressing something in particular

eras/babbage/impl/src/Cardano/Ledger/Babbage/Rules/Utxo.hs
```diff
@@ -233,24 +233,32 @@ validateCollateralContainsNonADA ::
...
+  -- validation, which has to compute the full balance on the Values, not just the Coin.
+  -- By using 'areAllAdaOnly', the (True,True) case, this approach is both cheap and common,
+  -- and the computations of colInputsBal and totalBal, in the where clause, are not computed at all.
+  case (areAllAdaOnly utxoCollateral, areAllAdaOnly (txBody ^. collateralReturnTxBodyL)) of
+    (True, True) -> pure ()
+    (False, False) ->
+      if (Val.isAdaOnly totalBal)
+        then pure () -- Both have MultiSssets, and they cancel out.
+        else failureIf True $ CollateralContainsNonADA colInputBal
```
teodanciu, line 245: I think we can use `failure $ CollateralContainsNonADA colInputBal` instead if we 
add `failure` to the imports above :
`import Validation (Validation, failure, failureIf, failureUnless,)`

eras/babbage/impl/src/Cardano/Ledger/Babbage/Rules/Utxo.hs
```diff
@@ -233,24 +233,32 @@ validateCollateralContainsNonADA ::
...
+    (False, True) -> failureIf True $ CollateralContainsNonADA colInputBal
```
teodanciu, line 246: Same as above: `failure $ CollateralContainsNonADA colInputBal`

eras/babbage/impl/src/Cardano/Ledger/Babbage/Rules/Utxo.hs
```diff
@@ -233,24 +233,32 @@ validateCollateralContainsNonADA ::
...
+    (True, False) -> case txBody ^. collateralReturnTxBodyL of
+      SJust retTxOut -> failureIf True $ CollateralContainsNonADA (retTxOut ^. valueTxOutL)
```
teodanciu, line 248: `failure $ CollateralContainsNonADA (retTxOut ^. valueTxOutL)`

eras/babbage/impl/src/Cardano/Ledger/Babbage/Rules/Utxo.hs
```diff
@@ -233,24 +233,32 @@ validateCollateralContainsNonADA ::
...
-              CollateralContainsNonADA (retTxOut ^. valueTxOutL)
-        _ -> CollateralContainsNonADA colbal
+  -- When we do not have any non-ada TxOuts we can short-circuit the more expensive
+  -- validation, which has to compute the full balance on the Values, not just the Coin.
+  -- By using 'areAllAdaOnly', the (True,True) case, this approach is both cheap and common,
+  -- and the computations of colInputsBal and totalBal, in the where clause, are not computed at all.
+  case (areAllAdaOnly utxoCollateral, areAllAdaOnly (txBody ^. collateralReturnTxBodyL)) of
+    (True, True) -> pure ()
+    (False, False) ->
+      if (Val.isAdaOnly totalBal)
```
teodanciu, line 243: Unnecessary parentheses

teodanciu (commented):
I added some superficial remarks on the code. 

I think it's clearer than in master, and it made me understand the logic, but it seems that there's a new version of this that only triggers computations when validation is enabled: https://github.com/input-output-hk/cardano-ledger/pull/3574

lehins:
Closing in favor of #3574

PR #3574, opened 2023-07-25
Improve clarity and performance of collateral Non-ADA validation:

lehins:
# Description

Predicate checks are lazy in their argument, because it is possible to run STS rules without doing any validation. Current Non-Ada validation ignored this capability and forced validation regardless of the options provided to the STS runner, which this PR fixes.

This commit also improves the clarity on the implementation of validation, by explaining all possible cases.

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/eras/babbage/impl/src/Cardano/Ledger/Babbage/Rules/Utxo.hs
+++ b/eras/babbage/impl/src/Cardano/Ledger/Babbage/Rules/Utxo.hs
@@ -218,7 +218,7 @@ validateTotalCollateral pp txBody utxoCollateral =
     ]
   where
     bal = collAdaBalance txBody utxoCollateral
-    fromAlonzoValidation x = first (fmap inject) x
+    fromAlonzoValidation = first (fmap inject)
 
 -- | This validation produces the same failure as in Alonzo, but is slightly
 -- different then the corresponding one in Alonzo, since it is possible to add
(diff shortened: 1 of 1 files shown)


eras/babbage/impl/src/Cardano/Ledger/Babbage/Rules/Utxo.hs
```diff
@@ -233,24 +233,39 @@ validateCollateralContainsNonADA ::
...
+    --   we report utxo balance
+    valueWithNonAda =
+      case txBody ^. collateralReturnTxBodyL of
+        SNothing -> collateralBalance
+        SJust retTxOut ->
+          if utxoCollateralHasOnlyAda
+            then retTxOut ^. valueTxOutL
+            else collateralBalance
+    -- This is the balance that is provided by the collateral inputs
+    collateralBalance = balance $ UTxO utxoCollateral
```
teodanciu, line 262: What do you think about having this function extracted next to the one that computes the ada-only collateral balance that is in the Collateral module? https://github.com/input-output-hk/cardano-ledger/blob/master/eras/babbage/impl/src/Cardano/Ledger/Babbage/Collateral.hs#L43
Maybe it would make it would keep the focus of this function if the balance logic is extracted, but on the other hand, too much fragmentation is not helpful either. 
Otherwise we can do it if it turns out we need to reuse it again at some point

lehins: The problem with the suggestion is that the helper function would have to return both the collateralBalance and totalCollateralBalance, cause we need both, which is just two Values and it would be very easy to mix them up.

It would be something like:
```
collBalance ::
  forall era.
  BabbageEraTxBody era =>
  TxBody era ->
  Map.Map (TxIn (EraCrypto era)) (TxOut era) ->
  (Value era, Value era)
collBalance txBody utxoCollateral = ...
```
Which without looking into the body of the function is not clear which is which and could lead to mistakes.

Also, it is the only place where this functionality appear, so moving it out into a different module, might not be that great.

teodanciu: Totally agree, I missed that detail :+1:

teodanciu (approved):
Looks good to me

PR #3579, opened 2023-07-26
Voting procedure map

### LINKED ISSUE #3542: Convert `votingProceduresTxBodyL` to a `Map`
Currently `votingProceduresTxBodyL` is defined as a sequence, which is problematic because it allows duplicate proposals.

Duplicate proposals don't make sense in general, but in the same transaction they will even cause issues when identifying a script with a `ScriptPurpose`.

For this reasons 
```haskell
  , ctbrVotingProcedures :: !(StrictSeq (VotingProcedure era))
```
must be switched to:
```haskell
  , ctbrVotingProcedures :: !(Map (VotingProcedureId (EraCrypto era)) (VotingProcedure era))
...

data VotingProcedureId c = VotingProcedureId
  { vpiGovActionId :: !(GovActionid c)
  , vpiVoter :: !(Voter c)
  }
```

This means that the `VotingProcedure` will loose the above two fields:
```haskell
data VotingProcedure era = VotingProcedure
  { vProcVote :: !Vote
  , vProcAnchor :: !(StrictMaybe (Anchor (EraCrypto era)))
  }
```

lehins:
# Description

Fixes #3542 

The fix is slightly different than in the ticket. The `votingProceduresTxBodyL` field is switched to a nested Map that is enforced to not have any empty Maps on the second level by the decoder. Arbitrary instance also reflects this invariant.

This approach reduced transaction size significantly when there many votes for the same Voter, which is expected to be a common case. Also simplifies rules in logic a bit.

Also bumps up versions for `small-steps-test` and `cardano-ledger-shelley-test`, since those provide support for GHC-9.6, but haven't been released yet.

# Checklist

- [x] Self-reviewed the diff


eras/conway/impl/CHANGELOG.md
```diff
@@ -23,6 +23,13 @@
    * `TALLY` -> `GOV`
    * `VDEL` -> `GOVCERT`
 * Make `Anchor` required in `ProposalProcedure`.
+* Add `VotingProcedures` type
+* Remove `vProcGovActionId` and `vProcVoter` from `VotingProcedure`
+* Change the type of `votingProceduresL` to return `VotingProcedures`, which is a nested Map
+  instead of a sequence, as before.
+* Change `GovernanceActionDoesNotExist` to `GovernanceActionsDoNotExist`, which now
+  reports all actions as a set, rather than one action per each individual failure.
+* Type of `gpVotingProcedures` in `GovernanceProcedures` was aslo changed to `GovernanceProcedures`
```
Soupstraw, line 42:
```suggestion
* Type of `gpVotingProcedures` in `GovernanceProcedures` was also changed to `GovernanceProcedures`
```

eras/conway/test-suite/cddl-files/conway.cddl
```diff
@@ -67,15 +67,15 @@ transaction_body =
   , ? 16 : transaction_output       ; collateral return
   , ? 17 : coin                     ; total collateral
   , ? 18 : set<transaction_input>   ; reference inputs
-  , ? 19 : [* voting_procedure]     ; New; Voting procedures
+  , ? 19 : voting_procedures        ; New; Voting procedures
   , ? 20 : [* proposal_procedure]   ; New; Proposal procedures
   }
 
+voting_procedures = { + voter => { + governance_action_id => voting_procedure } }
```
Soupstraw, line 74: It looks to me like the actual decoder would happily decode an empty outer map, so the CDDL might be too strict here
```suggestion
voting_procedures = { voter => { + governance_action_id => voting_procedure } }
```

lehins: No, empty Map in Haskell means that the field is missing in CBOR.

See decoder for the `mint` field. It is the exact same approach.

Soupstraw (approved):
LGTM

aniketd (approved):
LGTM! :raised_hands:

PR #3586, opened 2023-07-28
Add CurrentTreasuryValue to TxBody

### LINKED ISSUE #3536: Add current treasury amount information to the transaction
There are two options on how we can go about this:

* ~Add current treasury amount to the `TreasuryWithdrawals` proposal. Since that seems like the only place where it could be useful~ It was decided to use a separate field, since treasury amount could be useful for other operations, besides withdrawals (eg. donations)
* Add it as a separate field to the `TxBody`

One powerful use case for this feature is implementation of some treasure policy that can be enforced by the GovernancePolicy, which will be part of the constitution, see #3532

aniketd:
# Description

Resolves #3536 

# Checklist

- [x] Self-reviewed the diff


eras/conway/impl/src/Cardano/Ledger/Conway/TxBody.hs
```diff
@@ -378,6 +382,8 @@ instance Crypto c => ConwayEraTxBody (ConwayEra c) where
     lensMemoRawType ctbrVotingProcedures (\txb x -> txb {ctbrVotingProcedures = x})
   proposalProceduresTxBodyL =
     lensMemoRawType ctbrProposalProcedures (\txb x -> txb {ctbrProposalProcedures = x})
+  currentTreasuryValueTxBodyL =
+    lensMemoRawType ctbrCurrentTreasuryValue (\txb x -> txb {ctbrCurrentTreasuryValue = x})
```
aniketd, line 388: Wondering why we don't `INLINE` these like the other lenses in `*EraTxBody`.

eras/conway/impl/CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## 1.7.0.0
 
+* Add `*CurrentTreasuryValue :: !(StrictMaybe Coin)` as a new field to Conway TxBody #3586
```
lehins: The wildcard is a bit confusing.
```suggestion
* Add `currentTreasuryValue :: !(StrictMaybe Coin)` as a new field to Conway TxBody #3586
```

eras/conway/impl/src/Cardano/Ledger/Conway/TxBody.hs
```diff
@@ -512,6 +525,7 @@ encodeTxBodyRaw ConwayTxBodyRaw {..} =
         !> encodeKeyedStrictMaybe 15 ctbrTxNetworkId
         !> Omit (null . unVotingProcedures) (Key 19 (To ctbrVotingProcedures))
         !> Omit null (Key 20 (To ctbrProposalProcedures))
+        !> encodeKeyedStrictMaybe 21 ctbrTotalCollateral
```
lehins: Looks like a bad copy and paste
```suggestion
        !> encodeKeyedStrictMaybe 21 ctbrCurrentTreasuryValue
```

eras/conway/impl/src/Cardano/Ledger/Conway/TxBody.hs
```diff
@@ -491,8 +502,10 @@ encodeTxBodyRaw ::
 encodeTxBodyRaw ConwayTxBodyRaw {..} =
   let ValidityInterval bot top = ctbrVldt
    in Keyed
-        ( \i ifee ri o cr tc f t c w b rsh mi sh ah ni ga ->
-            ConwayTxBodyRaw i ifee ri o cr tc c w f (ValidityInterval b t) rsh mi sh ah ni ga
+        ( \i ifee ri o cr tc f t c w b rsh mi sh ah ni ga ctv ->
+            -- QUESTION: Why is there `ga` when that has already been split into
+            -- `VotingProcedures` and `ProposalProcedures`? Should it not be ` vp pp`?
+            ConwayTxBodyRaw i ifee ri o cr tc c w f (ValidityInterval b t) rsh mi sh ah ni ga ctv
```
lehins: Answer: Because it is eta reduced. So, in the current code above `ga` corresponds to `votingProcedures`, `ctv` corresponds to `proposalProcedures` and the new field you added is eta reduced. So, to avoid confusion and potential mistakes lets eta reduce as much as possible:
```suggestion
        ( \i ci ri o cr tc f t c w b ->
            ConwayTxBodyRaw i ci ri o cr tc c w f (ValidityInterval b t)
```

lehins (changes requested):
Found one mistake, otherwise looks good.

eras/conway/impl/src/Cardano/Ledger/Conway/TxBody.hs
lehins: Feel free to add them, it's good for performance. Whoever implemented these lenses simply have forgotten to add them.

eras/conway/impl/CHANGELOG.md
aniketd: I agree now :+1:

eras/conway/impl/src/Cardano/Ledger/Conway/TxBody.hs
aniketd: Thanks for a great catch!

eras/conway/impl/src/Cardano/Ledger/Conway/TxBody.hs
aniketd: Got it! Thanks! :raised_hands:

lehins (approved):
:+1:

Issue #3446, opened 2023-05-23
Compute the DRep stake distribution


JaredCorduan:
The DRep stake distribution is obtained by applying the DRep delegation relation to the per-stake-credential stake distribution, similiarly to how the per-stake-poool distribution is created. This distribution needs to be computed and stored in the ledger state for the governance ratification logic to use.

There are several ways to do this, with pros and cons. We will refine this issue when we have definitely made a decision.

lehins:
This will likely have to be done with a Pulser, in similar way how we compute StakePool distribution.

lehins:
Fixed by #3592

Issue #3590, opened 2023-07-31
Snapshot parts of ConwayGovState


lehins:
All of the relevant stuff necessary for voting has to be snapshotted.

The idea behind this ticket is that we have a full epoch to compute the results of a voting round, which means we need preserve the state of votes unchanged for that whole epoch and the accepted votes for governance actions in current epoch should not affect the votes in the previous epoch.

It is important to note that we only need information snapshotted for one epoch, so there is no need to add this information to the SnapShot type, which has three state (mark, set and go). We only need two: current epoch, that is changeable and previous epoch that is frozen.

lehins:
Something like this:

```haskell
data GovActionsState era = GovActionsState
  { curGovActionsState :: Map (GovActionId (EraCrypto era)) (GovActionState era)
  , prevGovActionsState :: Map (GovActionId (EraCrypto era)) (GovActionState era)
  }
```

PR #3592, opened 2023-07-31
Added pulsing incremental computation of the DRep distribution

TimSheard:
Added module Cardano.Ledger.DRepDistr, defines pulsing DRepDistr.
Integrated the DRepDistr into NEWEPOCH, EPOCH, and RATIFY rules.

# Description
Addresses issue #3446 

# Checklist

- [x] Self-reviewed the diff


eras/conway/impl/src/Cardano/Ledger/Conway/Rules/NewEpoch.hs
```diff
@@ -141,31 +143,33 @@ newEpochTransition ::
...
         SJust p@(Pulsing _ _) -> do
           (ans, event) <- liftSTS (completeRupd p)
           tellReward (DeltaRewardEvent (RupdEvent eNo event))
-          updateRewards es eNo ans
-        SJust (Complete ru') -> updateRewards es eNo ru'
-      es'' <- trans @(EraRule "EPOCH" era) $ TRC (pd, es', eNo)
-      let adaPots = totalAdaPotsES es''
+          updateRewards es0 eNo ans
+        SJust (Complete ru') -> updateRewards es0 eNo ru'
+      let _completedDRepPulser = extractDRepDistr (es1 ^. epochStateDRepDistrL)
```
lehins: This binding is ignored, because, I think, this part is being done in the `EPOCH` rule
```suggestion
```

eras/shelley/impl/src/Cardano/Ledger/Shelley/Lenses.hs
lehins, line 1: It is really not good to put lenses in separate module like that cause no-one will ever find them. 

Lenses should live right next to the types that they operate on. In fact, if you try to move them into the respectful modules, you will find that most of them already exist.

libs/cardano-ledger-core/src/Cardano/Ledger/DRepDistr.hs
```diff
@@ -0,0 +1,222 @@
...
+import Cardano.Ledger.UMap
+import qualified Cardano.Ledger.UMap as UMap
+import Data.Map.Strict (Map)
+import qualified Data.Map.Strict as Map
+import Data.Set (Set)
+import qualified Data.Set as Set
+
+-- import Cardano.Ledger.CertState(CertState(..),VState(..),PState(..),DState(..))
+-- import Cardano.Ledger.Shelley.LedgerState(NewEpochState(..))
+
```
lehins:
```suggestion
```

libs/cardano-ledger-core/src/Cardano/Ledger/DRepDistr.hs
```diff
@@ -0,0 +1,222 @@
...
+import Cardano.Ledger.Credential (Credential (..))
+import Cardano.Ledger.Keys (KeyRole (..))
+
+-- import Cardano.Ledger.Core(Era(EraCrypto))
+-- import Data.Maybe.Strict(StrictMaybe(..))
+
```
lehins:
```suggestion
```

libs/cardano-ledger-core/src/Cardano/Ledger/DRepDistr.hs
```diff
@@ -0,0 +1,222 @@
...
+import Control.Monad.Identity (Identity (..))
+import Data.Kind
+import Data.Pulse (Pulsable (..), complete, pulse)
+import Data.Typeable
+import Data.VMap (VB, VMap (..), VP, foldlWithKey)
+import qualified Data.VMap as VMap
+import GHC.Generics (Generic)
+import NoThunks.Class (NoThunks (..), allNoThunks)
+
+-- import Cardano.Ledger.EpochBoundary (SnapShot (..),  Stake (..))
```
lehins:
```suggestion
```

libs/cardano-ledger-core/src/Cardano/Ledger/DRepDistr.hs
```diff
@@ -0,0 +1,222 @@
...
+--   have 'm' and 'ans' as its last two parameters so we can make a Pulsable instance.
+--   We will always use this instantiation (DRepPulser c Identity (Map (DRep c) (CompactForm Coin)))
+data DRepPulser c (m :: Type -> Type) ans where
+  DRepPulser ::
+    (ans ~ Map (DRep c) (CompactForm Coin), m ~ Identity) =>
+    !Int ->
+    !(UMap c) ->
+    !(Set (Credential 'Voting c)) ->
+    -- purposely not strict
+    (VMap VB VP (Credential 'Staking c) (CompactForm Coin)) ->
```
lehins: Why is it lazy? We force this field at every pulse, so it seems superfluous to specifically keep it lazy.

libs/cardano-ledger-core/src/Cardano/Ledger/DRepDistr.hs
```diff
@@ -0,0 +1,222 @@
...
+    allNoThunks
+      [ noThunks ctxt n
+      , noThunks ctxt um
+      , --   , noThunks ctxt drep -- Here there be thunks
+        noThunks ctxt balance
+      , noThunks ctxt ans
+      ]
+
+instance NFData (DRepPulser c Identity (Map (DRep c) (CompactForm Coin))) where
+  rnf (DRepPulser n1 um drep bal ans) = seq (rnf n1) (seq (rnf um) (seq (rnf drep) (seq (rnf bal) (rnf ans))))
```
lehins: I think it is a bit easier to read it when it is in this form:
```suggestion
  rnf (DRepPulser n1 um drep bal ans) = n1 `deepseq` um `deepseq` drep `deepseq` bal `deepseq` rnf ans
```

libs/cardano-ledger-core/src/Cardano/Ledger/DRepDistr.hs
```diff
@@ -0,0 +1,222 @@
...
+  | cardano-ledger-core-1.5.0.0:Cardano.Ledger.Core.TxCert.DRepScriptHash !(ScriptHash c)
+  | DRepAlwaysAbstain
+  | DRepAlwaysNoConfidence
+
+data VState era
+  = VState {vsDReps :: !(Set (Credential 'Voting (EraCrypto era))),
+            vsCommitteeHotKeys :: !(Map
+                                      (KeyHash 'CommitteeColdKey (EraCrypto era))
+                                      (Maybe (Credential 'CommitteeHotKey (EraCrypto era))))
+-}
```
lehins: Some dead code
```suggestion
```

lehins (changes requested):
I think this looks great. Just some minor cleanup is needed and one minor question.

We still need to remember all of the votes at the epoch boundary as well, so they don't change, but that will be done as a separate ticket in #3590

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/NewEpoch.hs
TimSheard: I originally put it here, then I noticed  that this is computed from the EpochState, and could be just as well computed in the EPOCH rule. Then I forgot to remove it.

eras/shelley/impl/src/Cardano/Ledger/Shelley/Lenses.hs
TimSheard: Actually it is the Lenses that live in the files where the types are defined that you can never find. Here they are all in one place. An it is easy to see how they relate. If I need a lens, I need import just one file, and here they are all uniformly named.
It is also a good pace to put named composition of lenses. One can't put these in the files that the define the types, because for composed lenses, the types being composed are often defined in different files.

libs/cardano-ledger-core/src/Cardano/Ledger/DRepDistr.hs
TimSheard: done

libs/cardano-ledger-core/src/Cardano/Ledger/DRepDistr.hs
TimSheard: done

libs/cardano-ledger-core/src/Cardano/Ledger/DRepDistr.hs
```diff
@@ -0,0 +1,222 @@
...
+instance Crypto c => EncCBOR (DRepDistr c) where
+  encCBOR (DRComplete x) = encode (Rec DRComplete !> To x)
+  encCBOR (DRPulsing x) = encode (Rec DRComplete !> To (complete x))
+
+instance Crypto c => DecCBOR (DRepDistr c) where
+  decCBOR = decode (RecD DRComplete <! From)
+
+instance ToExpr (DRepDistr era) where
+  toExpr (DRComplete x) = App "DRComplete" [toExpr x]
+  toExpr (DRPulsing x) = App "DRComplete" [toExpr (complete x)]
```
teodanciu, line 194: Should this be App "DRPulsing" ?

libs/cardano-ledger-core/src/Cardano/Ledger/DRepDistr.hs
TimSheard: If it was strict, it would be forced when we constructed the DRepDistr (at the epoch boundary) where we definitely don't want to incur this cost.  In fact this is the same thunk that the TICKF (forecast) rule forces.  It will be force by the first pulse, or the first use of TICKF, and then never again.

eras/shelley/impl/src/Cardano/Ledger/Shelley/Lenses.hs
lehins: I appreciate your suggestion Tim, but it goes against a well established pattern, not only in our codebase, but in industry as well. So, could please move all of the lenses in this module next to their respective types.

libs/cardano-ledger-core/src/Cardano/Ledger/DRepDistr.hs
TimSheard: I like that too.

libs/cardano-ledger-core/src/Cardano/Ledger/DRepDistr.hs
TimSheard: Forgot that was there. It was nice to refer to these things when I first got started,

eras/shelley/impl/src/Cardano/Ledger/Shelley/Lenses.hs
Soupstraw: The issue with storing the lenses in a single module is that the module has to depend on every module that defines the data types that the lenses operate on. That makes it really hard to use the lenses inside the same package, since it's easy to introduce circular dependencies.

libs/cardano-ledger-core/src/Cardano/Ledger/DRepDistr.hs
TimSheard: Actually it should be "DRComplete".  Notice how it completes the pulser when turning it into an Expr.  This is a pattern we use a lot. Note that in (DRPulsing x) 'x' has a very complicated internal structure, with internal state, and what we really want is to compare two DRepDistr objects and report the subparts that are different.  That wouldn't really work, because two pulsers that have pulsed a different number of times, could have different internal structures, but the same final result. So we run every (DRPulsing x) to completion and compare the final results.

eras/shelley/impl/src/Cardano/Ledger/Shelley/Lenses.hs
TimSheard: I don't buy that argument here, Every single lens in this file is defined in the Shelley module (or the Core module), and the Shelley module depends on the Core module, so there are no extra dependencies. Even if you want just one lens from this file, you will get the same dependencies.

lehins: @TimSheard My argument is very simple: this is how everyone in Haskell defines lenses for types. We are not going change that convention for this codebase either. Remember what feels natural to you, does not mean everyone feels the same way.

libs/cardano-ledger-core/src/Cardano/Ledger/DRepDistr.hs
teodanciu: This make sense, thank you for the explanation

libs/cardano-ledger-core/src/Cardano/Ledger/DRepDistr.hs
TimSheard: done

eras/shelley/impl/src/Cardano/Ledger/Shelley/Lenses.hs
TimSheard: Lets get the facts straight. Of the 55 Lenses defined in Lenses.hs
only 15 are defined else where (Mostly in Cardano.Ledger.Shelley.LedgerState.Types)
of these 15 only these 6 are exported (from Cardano.Ledger.Shelley.LedgerState)
  esPpL,
  esLStateL,
  lsUTxOStateL,
  utxosFeesL,
  utxosGovernanceL,
  nesEpochStateL,
The other 9 are inaccessible because Types.hs is a hidden file.
Of these 6, the most widely used, 'nesEpochStateL', does not follow the naming
convention of being named after the field name it accesses. It should be 'nesEsL'

The missing 40 Lenses, if they existed would be spread across 2 other
modules (Cardano.Ledger.EpochBoundary and Cardano.Ledger.CertState, both in the Core, so 
the circularity problem won't occur).

No wonder no one can find the Lenses, and hence no one use them.
The argument that "this is the way other people do it" is silly.
Since when do we do things because other people do it that way? When doing it another
way serves our purpose better. We are defining the Ledger State. The Ledger state consists of 11 types.

DState
PState
VState
CertState
UTxOState
LedgerState
AccountState
SnapShots
SnapShot
EpochState
NewEpochState

For historical reasons we spread the definition of them across many
files. But they really are the heart of the LedgerState. We shoud treat them as a unit.
We have this opportunity to clean things up. To make them easy to find, and
widely used, and integrated as a unit (at least at the Lense level). Why don't we use the
Lenses in the STS rules? Mostly because they don't exist, and are hard to find.
We, the ledger team, are the ones we should accomodate. Not hypothetical 'other people'.
By the way, the other people that actually use our codebase are the downstream clients use the API.
An organized set of lenses would serve them as well. Why wouldn't we do this? This make sense. 

If I have to, I can make the changes, but it is a lot of work. It touches many files. And it doesn't serve us.

lehins: The whole lens situation for the ledger state types are a mess because they were created by different people in different places. But, I totally expected this mess and it is not a huge problem, until we make it a real public API (export from the cardano-ledegr-api)

That being said, I will personally fix it, after we are done with delivering SanchoNet.

When I fix it all lenses **will** live right next to their types. So, @TimSheard could you please help me out and move each individual lens next to the type which the lens is defined for. If you can find duplicate lenses I'd appreciate it if you can remove them. This will help me later in creating a proper ledger state lens API.

TimSheard: Like I said, I will do this. But I do not believe it serves our purpose. I guess we can agree to disagree.

lehins: I understand. Thank you for agreeing to apply this change! I do appreciate it. Hopefully you'll change your mind in the future once my plans do come to fruition for those lenses and you see what I had in mind.

libs/cardano-ledger-core/src/Cardano/Ledger/CertState.hs
```diff
@@ -253,8 +255,13 @@ toPStatePair PState {..} =
   , "deposits" .= psDeposits
   ]
 
+-- | The state that tracks the voting entities (DReps and Constituional Committe members)
+--   The 'vsDRepDistr' field is a pulser that incrementally computes the stake distribution of the DReps
+--   over the Epoch following the close of voting at end of the previous Epoch. The pulser is created
+--   at the Epoch boundary, but does no work until it is pulsed in the 'Tick' rule.
 data VState era = VState
   { vsDReps :: !(Set (Credential 'Voting (EraCrypto era)))
+  , vsDRepDistr :: DRepDistr (EraCrypto era)
```
lehins: We probably do want to make this field strict though, right?
```suggestion
  , vsDRepDistr :: !(DRepDistr (EraCrypto era))
```

lehins (approved):
Very nice. Thank you.

libs/cardano-ledger-core/src/Cardano/Ledger/CertState.hs
TimSheard: yes that is right

PR #3596, opened 2023-08-01
Upgradeable families

lehins:
# Description

Take care of part of #3618

Adds `upgradeTxOut`, `upgradeTxAuxData`, `upgradeScript` and `upgradeTxCert` as well as:

* Creates a test suite to each `cardano-ledger-[era]` package that didn't have one. See #3623 for more info on a long goal with respect to test suites.
* Add a test for each of the four type families that checks upgrading matches translation through CBOR.
* Add `eraName` that can be used in any place where textual representation of an era is needed (this can subsume similar functionality from consensus)
* Add `ToExpr` instances for more types. Needed for testing.
* For consistency moves `EraTxAuxData` instances into their own modules for Babbage and Conway 

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/eras/conway/impl/src/Cardano/Ledger/Conway/TxOut.hs
+++ b/eras/conway/impl/src/Cardano/Ledger/Conway/TxOut.hs
@@ -2,7 +2,6 @@
 {-# LANGUAGE DerivingVia #-}
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE FlexibleInstances #-}
-{-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE MultiParamTypeClasses #-}
 {-# LANGUAGE RankNTypes #-}
 {-# LANGUAGE ScopedTypeVariables #-}
--- a/eras/conway/impl/src/Cardano/Ledger/Conway/TxOut.hs
+++ b/eras/conway/impl/src/Cardano/Ledger/Conway/TxOut.hs
@@ -2,6 +2,7 @@
 {-# LANGUAGE DerivingVia #-}
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE FlexibleInstances #-}
+{-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE MultiParamTypeClasses #-}
 {-# LANGUAGE RankNTypes #-}
 {-# LANGUAGE ScopedTypeVariables #-}
@@ -14,8 +15,11 @@ module Cardano.Ledger.Conway.TxOut (
 )
 where
 
-import Cardano.Ledger.Alonzo.Scripts.Data (Datum (NoDatum))
+import Cardano.Ledger.Address (addrPtrNormalize)
+import Cardano.Ledger.Alonzo.Scripts.Data (translateDatum)
 import Cardano.Ledger.Babbage.TxBody (
+  BabbageTxOut (..),
+  Datum (..),
   addrEitherBabbageTxOutL,
   babbageMinUTxOValue,
   dataBabbageTxOutL,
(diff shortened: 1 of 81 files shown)


libs/cardano-ledger-core/testlib/Test/Cardano/Ledger/Core/Binary.hs
```diff
@@ -0,0 +1,128 @@
...
+specTxAuxDataUpgrade =
+  prop "upgradeTxAuxData is preserved through serialization" $ \prevTxAuxData -> do
+    case embedTripAnn (eraProtVerHigh @(PreviousEra era)) (eraProtVerLow @era) prevTxAuxData of
+      Left err ->
+        expectationFailure $
+          "Expected to deserialize: =======================================================\n"
+            ++ show err
+      Right (curTxAuxData :: t era) ->
+        -- We need to do all this MemoBytes trickery because underlying bytes and thus the
+        -- equality of the same type will no longer be the same, despite that the value will
```
nc6, line 83: Why is this the case for TxAux and not elsewhere?

libs/cardano-ledger-core/testlib/Test/Cardano/Ledger/Core/Binary.hs
```diff
@@ -0,0 +1,128 @@
...
+    case embedTripAnn (eraProtVerHigh @(PreviousEra era)) (eraProtVerLow @era) prevScript of
+      Left err ->
+        expectationFailure $
+          "Expected to deserialize: =======================================================\n"
+            ++ show err
+      Right (curScript :: Script era) ->
+        curScript `shouldBe` upgradeScript prevScript
+
+specUpgrade ::
+  forall era t.
```
nc6: Add a comment explaining what `t` is and why it needs to be specified explicitly?

libs/cardano-ledger-core/src/Cardano/Ledger/Core.hs
```diff
@@ -378,7 +382,16 @@ class
   EraTxAuxData era
   where
   type TxAuxData era = (r :: Type) | r -> era
+
+  -- | Every era, except Shelley, must be able to upgrade a `TxAuxData` from a previous
+  -- era.
+  --
+  -- /Warning/ - Important to note that any memoized binary representation will not be
```
nc6, line 389: The relevant distinction here being that this type contains its memoised bytes, but other things don't?

libs/cardano-ledger-core/src/Cardano/Ledger/Core/TxCert.hs
```diff
@@ -65,6 +67,18 @@ class
   where
   type TxCert era = (r :: Type) | r -> era
 
+  type TxCertUpgradeError era :: Type
+  type TxCertUpgradeError era = Void
+
+  -- | Every era, except Shelley, must be able to upgrade a `TxCert` from a previous
+  -- era. However, not all certificates can be upgraded, because some eras loose some of
```
nc6:
```suggestion
  -- era. However, not all certificates can be upgraded, because some eras lose some of
```

eras/conway/impl/src/Cardano/Ledger/Conway/TxOut.hs
```diff
@@ -2,6 +2,7 @@
 {-# LANGUAGE DerivingVia #-}
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE FlexibleInstances #-}
+{-# LANGUAGE LambdaCase #-}
```
nc6:
```suggestion
```

nc6 (approved):
LGTM, have left a few comments requesting mostly clarification

libs/cardano-ledger-core/testlib/Test/Cardano/Ledger/Core/Binary.hs
lehins: This case will also be in `TxBody`. It is because `TxAuxData` is based on MemoBytes and equality on memo bytes also checks the equality of underlying bytes, which might not be the same between the two eras. That's why we need to use `zipMemoRawType` in order to compare just the underlying raw type, while ignoring the bytes

libs/cardano-ledger-core/src/Cardano/Ledger/Core.hs
lehins: Yes, that's correct.

libs/cardano-ledger-core/testlib/Test/Cardano/Ledger/Core/Binary.hs
lehins: I am not quite sure how to best explain it, but I'll try to add something. In one sentence: otherwise we can't use MemoBytes interface to compare equality on the raw type.

> The true reason why we need to pass `t` as a type argument is because `TxAuxData` is a type family, so we don't know if the final type will be of the kind `Type -> Type` and can be used with MemoBytes, which requires `t` to be of such kind, because it is later applied to `era`.

^ @nc6 Do you think it is that a good enough explanation to add as a comment in the code?

lehins: Done.

PR #3597, opened 2023-08-02
Update `ProposalProcedure` return address to be a `RewardAcnt`

### LINKED ISSUE #3594: Change `pProcReturnAddr` to `Credential`
Currently it is set to just a `KeyHash`, which is only 50% correct.

It [the spec](https://github.com/input-output-hk/formal-ledger-specifications/blob/e49fe43cafce0833d7ca8caf8016cd3e4d1467b2/src/Ledger/GovernanceActions.lagda#L129) however it is set to `RewardAcnt`, which would mean we also need to worry about the `NetworkId`. This does not seem right to me, so it would be nice to get some clarification on this

teodanciu:
# Description

Closes https://github.com/input-output-hk/cardano-ledger/issues/3594

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/eras/conway/impl/CHANGELOG.md
+++ b/eras/conway/impl/CHANGELOG.md
@@ -50,6 +50,7 @@
 * Remove `DelegStakeTxCert` from the `COMPLETE` pragma for `TxCert`
 * Add `Committee` and adjust `NewCommittee` governance action
 * Add `ConwayUpdateDRep` constructor to `ConwayGovCert` type and corresponding pattern `UnRegDRepTxCert`
+* Update `ProposalProcedure` return address to be a `RewardAcnt`
 
 ## 1.6.3.0
 
--- a/eras/conway/impl/src/Cardano/Ledger/Conway/Governance.hs
+++ b/eras/conway/impl/src/Cardano/Ledger/Conway/Governance.hs
@@ -47,6 +47,7 @@ module Cardano.Ledger.Conway.Governance (
   constitutionScriptL,
 ) where
 
+import Cardano.Ledger.Address (RewardAcnt)
 import Cardano.Ledger.BaseTypes (EpochNo (..), ProtVer (..), StrictMaybe)
 import Cardano.Ledger.Binary (
   DecCBOR (..),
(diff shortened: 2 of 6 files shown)


teodanciu:
I assume `scriptsNeeded` doesn't have to be updated for this, please correct me if I'm wrong

eras/conway/impl/src/Cardano/Ledger/Conway/Governance.hs
```diff
@@ -98,7 +98,7 @@ data GovernanceActionState era = GovernanceActionState
   , gasDRepVotes :: !(Map (Credential 'DRepRole (EraCrypto era)) Vote)
   , gasStakePoolVotes :: !(Map (KeyHash 'StakePool (EraCrypto era)) Vote)
   , gasDeposit :: !Coin
-  , gasReturnAddr :: !(KeyHash 'Staking (EraCrypto era))
+  , gasReturnAddr :: !(Credential 'Staking (EraCrypto era))
```
lehins: Turns out it actually has to be a reward account, not just a `Credential`
```suggestion
  , gasReturnAddr :: !(RewardAcnt (EraCrypto era))
```

lehins (changes requested):
During a meeting yesterday we figured out a discrepancy. This has to be a RewardAcnt, not just a Credential

teodanciu:
I updated it to be a RewardAcnt, however, I haven't used the `Network` anywhere. 
I assume it will be needed (otherwise why have it), but not sure if at this point. Maybe in the way we insert in UMap in Epoch rule?

lehins:
We'll only need to validate that the NetworkId matches the network we are running. That's all.

See this ticket I created: https://github.com/input-output-hk/cardano-ledger/issues/3603

eras/conway/impl/CHANGELOG.md
```diff
@@ -50,6 +50,7 @@
 * Remove `DelegStakeTxCert` from the `COMPLETE` pragma for `TxCert`
 * Add `Committee` and adjust `NewCommittee` governance action
 * Add `ConwayUpdateDRep` constructor to `ConwayGovCert` type and corresponding pattern `UnRegDRepTxCert`
+* Update `ProposalProcedure` return address to be a `Credential`
```
```suggestion
* Update `ProposalProcedure` return address to be a `RewardAcnt`
```

lehins (approved):
Great. Thank you!

PR #3599, opened 2023-08-02
Add governance action well-formedness check

### LINKED ISSUE #3584: Add governance action well-formedness check
[see here](https://github.com/input-output-hk/formal-ledger-specifications/blob/master/src/Ledger/Gov.lagda#L84)

Soupstraw:
# Description

closes #3584 

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/eras/conway/impl/CHANGELOG.md
+++ b/eras/conway/impl/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## 1.7.0.0
 
+* Add `MalformedProposal` to `ConwayGovPredFailure`
+* Add `ppuWellFormed` to `ConwayEraPParams`
 * Filter out zero valued `TxOut`'s on Byron/Shelley boundary instead of on Babbage/Conway.
 * Switch GovernanceActionIx to `Word32` fro `Word64` and remove `Num` and `Enum`
   instances, which are dangerous due to potential overflows.
--- a/eras/conway/impl/src/Cardano/Ledger/Conway/PParams.hs
+++ b/eras/conway/impl/src/Cardano/Ledger/Conway/PParams.hs
@@ -9,6 +9,7 @@
 {-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE MultiParamTypeClasses #-}
 {-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE RankNTypes #-}
 {-# LANGUAGE RecordWildCards #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE StandaloneDeriving #-}
@@ -45,6 +46,7 @@ import Cardano.Ledger.Conway.Core hiding (Value)
 import Cardano.Ledger.Conway.Era (ConwayEra)
 import Cardano.Ledger.Crypto
 import Cardano.Ledger.HKD (HKD, HKDFunctor (..))
+import Cardano.Ledger.Val (Val (..))
 import Control.DeepSeq (NFData)
 import Data.Aeson hiding (Encoding, decode, encode)
 import qualified Data.Aeson as Aeson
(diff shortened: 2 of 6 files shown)


eras/conway/impl/src/Cardano/Ledger/Conway/PParams.hs
```diff
@@ -291,6 +291,26 @@ instance Crypto c => BabbageEraPParams (ConwayEra c) where
...
+      , ppuCommitteeTermLimitL
+      , ppuGovActionExpirationL
+      ]
+      && all
+        (\x -> ppu ^. x /= SJust mempty)
+        [ -- Coins
+          ppuPoolDepositL
+        , ppuGovActionDepositL
+        , ppuDRepDepositL
+        ]
```
lehins: I know this suggestion is slightly more complicated, but it would allow uniformly check different conditions, regardless of the type in the lens and the validation check needed. For example if we decide to check ranges, eg. `isValid (\p -> p > 0 && p < 7 % 10) ppuCollateralPercentageL`

Feel free to ignore my suggestion, cause technically there is nothing wrong with your implementation. Both of those implementations have the desired property of short-circuiting on the first validation failure.

```suggestion
    isSJust $ do
      let isValid :: (x -> Bool) -> Lens' (PParamsUpdate era) (StrictMaybe x) -> StrictMaybe ()
          isValid check l = do
            True <- (check <$> l ppu) <|> SJust True
            pure ()
      isValid (/= 0) ppuMaxBBSizeL
      isValid (/= 0) ppuMaxTxSizeL
      isValid (/= 0) ppuMaxBHSizeL
      isValid (/= 0) ppuMaxValSizeL
      isValid (/= 0) ppuCollateralPercentageL
      isValid (/= 0) ppuCommitteeTermLimitL
      isValid (/= 0) ppuGovActionExpirationL
      isValid (/= zero) ppuPoolDepositL
      isValid (/= zero) ppuGovActionDepositL
      isValid (/= zero) ppuDRepDepositL
```

Side note. Needs `zero` from `Cardano.Ledger.Val`. It is cleaner to understand that we are dealing with Coin this way, not any Monoid.

eras/conway/impl/CHANGELOG.md
```diff
@@ -49,6 +49,7 @@
 * Remove `DelegStakeTxCert` from the `COMPLETE` pragma for `TxCert`
 * Add `Committee` and adjust `NewCommittee` governance action
 * Add `ConwayUpdateDRep` constructor to `ConwayGovCert` type and corresponding pattern `UnRegDRepTxCert`
+* Add `MalformedProposal` to `ConwayGovPredFailure`
```
lehins: Could you also add the note about adding `ppuWellFormed`

lehins (approved):
Looks great. Thank you

eras/conway/impl/src/Cardano/Ledger/Conway/PParams.hs
Soupstraw: Yeah, I think your solution has better readability than mine. I switched to using `isValid` inspired by your suggestion

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Gov.hs
```diff
@@ -154,13 +157,21 @@ noSuchGovernanceActions (ConwayGovState st) gaIds =
    in failureUnless (Set.null unknownGovActionIds) $
         GovernanceActionsDoNotExist unknownGovActionIds
 
+actionWellFormed :: GovernanceAction era -> Test (ConwayGovPredFailure era)
+actionWellFormed ga = failureUnless isWellFormed $ MalformedProposal ga
+  where
+    isWellFormed = case ga of
+      ParameterChange ppd -> ppWellFormed ppd
```
lehins: It's failing to build.
```suggestion
      ParameterChange ppd -> ppuWellFormed ppd
```

eras/conway/impl/CHANGELOG.md.orig
lehins, line 1: Looks like an accidental file got committed

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Gov.hs
```diff
@@ -59,13 +61,14 @@ data GovEnv era = GovEnv
...
   deriving (Eq, Show, Generic)
 
-instance Era era => NFData (ConwayGovPredFailure era)
+instance EraPParams era => NFData (ConwayGovPredFailure era)
 
-instance Era era => NoThunks (ConwayGovPredFailure era)
+instance EraPParams era => NoThunks (ConwayGovPredFailure era)
 
 instance EraPParams era => DecCBOR (ConwayGovPredFailure era) where
   decCBOR = decode $ Summands "ConwayGovPredFailure" $ \case
```
lehins, line 74: We also need to fix the decoder for the type

PR #3601, opened 2023-08-02
Added DRep anchors

### LINKED ISSUE #3553: DRep anchors
The Anchor that's part of the DRep registration certificate should probably be kept in the `dreps` map in `GState`, since people will want to query it. So you probably want it to be a `Map Credential (Epoch, Anchor)`.

This isn't part of the spec since it's irrelevant for block validity and an implementation that doesn't do this is still correct.

Soupstraw:
# Description

closes #3553

# Checklist

- [x] Self-reviewed the diff


libs/cardano-ledger-core/src/Cardano/Ledger/CertState.hs
```diff
@@ -275,8 +279,76 @@ toPStatePair PState {..} =
   , "deposits" .= psDeposits
   ]
 
+data AnchorDataHash
+
+data Anchor c = Anchor
```
lehins: Maybe we should put into the BaseTypes module?

Soupstraw: I don't think it's that easy since it requires `SafeHash` which creates a cyclical dependency.

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/GovCert.hs
```diff
@@ -139,21 +138,21 @@ conwayGovCertTransition = do
     judgmentContext
   case c of
     ConwayRegDRep cred _deposit _mAnchor -> do
-      Set.notMember cred vsDReps ?! ConwayDRepAlreadyRegistered cred
+      Map.notMember cred vsDReps ?! ConwayDRepAlreadyRegistered cred
       -- TODO: check against a new PParam `drepDeposit`, once PParams are updated. -- someCheck ?! ConwayDRepIncorrectDeposit deposit
-      pure $ vState {vsDReps = Set.insert cred vsDReps}
+      pure $ vState {vsDReps = Map.insert cred undefined vsDReps}
```
teodanciu: Are we gonna leave this as undefined?  Shouldn't we build a `DRepState` using the given anchor?

Soupstraw: oops, nice catch!

aniketd (approved):
LGTM! :raised_hands: other than the comments already made.

teodanciu (approved):
Looks good to me

libs/cardano-ledger-core/src/Cardano/Ledger/CertState.hs
lehins: Let's fix the circular dependency. 

Move this instance from `Cardano.Ledger.Orphans`:
```
instance Default Network where
  def = Mainnet
```
into the place where it belongs: `Cardano.Ledger.BaseTypes`

This should make it possible to put `Anchor` into `Cardano.Ledger.BaseTypes` as well

eras/conway/impl/src/Cardano/Ledger/Conway/Gov/Procedures.hs.orig
lehins, line 1: Accidental file.

I'd suggest you add `*.orig` to you `.global_gitignore` file :wink:

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/TypeRep.hs
```diff
@@ -1123,6 +1130,7 @@ hasOrd rep xx = explain ("'hasOrd " ++ show rep ++ "' fails") (help rep xx)
     help BoolR v = pure $ With v
     help DRepR v = pure $ With v
     help (PoolMetadataR _) v = pure $ With v
+    help DRepStateR _ = failT ["DRepState does not have Ord instance"]
```
lehins, line 1133: Just a question. Why can't we add an Ord instance for DRepState?

Seems like Anchor can easily get an Ord instance

Soupstraw: I don't think it's impossible, I just didn't want to bother with it :sweat_smile:

Soupstraw: We can add it in a different PR

lehins: Postponed in #3624

PR #3609, opened 2023-08-07
Adjust thresholds for sanchonet

### LINKED ISSUE #3605: Temporary Hard Coded Thresholds
We temporarily want to hard-code the thresholds until we get parameters integrated. Here's what I suggest to make the experience more exciting for participants:

1. Thresholds (make actions pass with SPO majority alone):

  * DRep - 0%
  * CC - 0%
  * SPO - >50% (51% if that's easier)

2. Equation should be (yes / total stake delegated - abstain votes)
3. expiration of actions: 30 epochs
4. HF Action >100% (impossible)

lehins:
# Description

Fixes #3605 

* [x] - Thresholds (make actions pass with SPO majority alone):
  * DRep - 0% - these were already 0
  * CC - 0% - same here, also 0
  * SPO - >50% (51% if that's easier) - set to 51%
* [x] - Equation should be (yes / total stake delegated - abstain votes) - done.
* [x] - expiration of actions: 30 epochs - set to 30
* [x] - HF Action >100% (impossible) - set to 101%

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/libs/cardano-ledger-test/src/Test/Cardano/Ledger/Examples/ConwayFeatures.hs
+++ b/libs/cardano-ledger-test/src/Test/Cardano/Ledger/Examples/ConwayFeatures.hs
@@ -182,7 +182,7 @@ govActionStateWithVote ProposalProcedure {..} kh v =
     pProcGovernanceAction
     (EpochNo 0)
 
--- | Value for the actual threshold, plus a small epsilon for GT (>) reation
+-- | Value for the actual threshold, plus a small epsilon for GT (>) relation
 spoThreshold :: Rational
 spoThreshold = 51 % 100 + 1 % 100000000000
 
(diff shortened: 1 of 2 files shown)


libs/cardano-ledger-test/src/Test/Cardano/Ledger/Examples/ConwayFeatures.hs
```diff
@@ -182,8 +182,9 @@ govActionStateWithVote ProposalProcedure {..} kh v =
     pProcGovernanceAction
     (EpochNo 0)
 
+-- | Value for the actual threshold, plus a small epsilon for GT (>) reation
```
teodanciu: `relation` ?

teodanciu (approved):
Looks good to me

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Examples/ConwayFeatures.hs
lehins: I'll fix it in a subsequent PR, just so we don't have to wait for another CI run

lehins: Turns out had to run CI again anyways. Fixed in this PR :smile: 
Thank you for pointing it out.

PR #3625, opened 2023-08-09
Change constitution to use `Anchor`

### LINKED ISSUE #3613: Switch constitution hash to an Anchor
Current Constitution does not have a field for the Url, just the hash of the constitution.

We need to switch the hash to Anchor, that contains a hash and url

Depends on #3601, because it moves `Anchor` type into `cardano-ledger-core`

This will also mean that `ConstitutionData` type can be removed

Soupstraw:
# Description

This PR replaces the constitution has with `Anchor`. Replaced `getConstitutionHash` with `getConstitution` in `EraGov`. Added `ConwayEraGov`, which adds a lens for accessing the constitution.

closes #3613

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/eras/conway/impl/CHANGELOG.md
+++ b/eras/conway/impl/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## 1.7.0.0
 
+* Add `ConwayEraGov` with `constitutionGovStateL`
 * Add `PrevGovActionId` and `GovActionPurpose`
 * Add optional `PrevGovActionId` to `ParameterChange`, `HardForkInitiation`,
   `NoConfidence`, `NewCommittee` and `NewConstitution` governance actions.
--- a/libs/cardano-ledger-api/src/Cardano/Ledger/Api/State/Query.hs
+++ b/libs/cardano-ledger-api/src/Cardano/Ledger/Api/State/Query.hs
@@ -10,8 +10,9 @@ module Cardano.Ledger.Api.State.Query (
   queryConstitutionHash,
 ) where
 
+import Cardano.Ledger.Allegra.Core (Constitution (constitutionAnchor))
 import Cardano.Ledger.Coin (Coin)
-import Cardano.Ledger.Conway.Governance (ConstitutionData, EraGov (getConstitutionHash))
+import Cardano.Ledger.Conway.Governance (Anchor (..), AnchorData, EraGov (..))
 import Cardano.Ledger.Core
 import Cardano.Ledger.Credential (Credential)
 import Cardano.Ledger.Keys (KeyHash, KeyRole (StakePool, Staking))
@@ -50,7 +51,13 @@ queryStakePoolDelegsAndRewards nes = filterStakePoolDelegsAndRewards (dsUnified 
 getDState :: NewEpochState era -> DState era
 getDState = certDState . lsCertState . esLState . nesEs
 
+queryConstitution :: EraGov era => NewEpochState era -> Maybe (Constitution era)
+queryConstitution nes =
+  getConstitution (nes ^. nesEpochStateL . esLStateL . lsUTxOStateL . utxosGovStateL)
+
 queryConstitutionHash ::
-  EraGov era => NewEpochState era -> Maybe (SafeHash (EraCrypto era) ConstitutionData)
+  EraGov era =>
+  NewEpochState era ->
+  Maybe (SafeHash (EraCrypto era) (AnchorData (EraCrypto era)))
 queryConstitutionHash nes =
-  getConstitutionHash $ nes ^. nesEpochStateL . esLStateL . lsUTxOStateL . utxosGovStateL
+  anchorDataHash . constitutionAnchor <$> queryConstitution nes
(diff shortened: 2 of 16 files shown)


eras/conway/impl/CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## 1.7.0.0
 
+* Add `ConwayEraGov`
```
teodanciu: Should we also mention `constitutionGovStateL` ? 
Add `ConwayEraGov` with `constitutionGovStateL` or something like that?

eras/shelley/impl/src/Cardano/Ledger/Shelley/Gov.hs
```diff
@@ -76,10 +72,6 @@ class
   getProposedPPUpdates :: GovState era -> Maybe (ProposedPPUpdates era)
   getProposedPPUpdates _ = Nothing
 
-  -- | Returns `Nothing` for all era preceding Conway, otherwise returns the hash of the constitution
-  getConstitutionHash :: GovState era -> Maybe (SafeHash (EraCrypto era) ConstitutionData)
```
teodanciu: I wonder why this was defined here, since it doesn't seem used in Shelley?  
But anyway, the lens version is more aligned with everything else :+1:

teodanciu: Ah I see, because `queryConstitutionHash` is paramterized on Era, but now we've given it  `ConwayEraGov era` constraint, nevermind!

lehins: It is used for the query in consensus, so we might still have to bring it back. I am trying to figure out right now in consensus if we can get away with removing it in Shelley.

libs/cardano-ledger-core/src/Cardano/Ledger/BaseTypes.hs
```diff
@@ -812,6 +812,9 @@ instance Crypto c => ToJSON (Anchor c) where
 
 instance ToExpr (Anchor c)
 
+instance Crypto c => Default (Anchor c) where
+  def = Anchor (Url "") def
```
teodanciu, line 825: Mhm, not sure about this, but I understand the convenience of it.

eras/shelley/impl/src/Cardano/Ledger/Shelley/Gov.hs
lehins: @Soupstraw We need to bring this function back, otherwise we can't implement the consensus ledger state query.

libs/cardano-ledger-api/src/Cardano/Ledger/Api/State/Query.hs
```diff
@@ -51,6 +51,12 @@ getDState :: NewEpochState era -> DState era
 getDState = certDState . lsCertState . esLState . nesEs
 
 queryConstitutionHash ::
-  EraGov era => NewEpochState era -> Maybe (SafeHash (EraCrypto era) ConstitutionData)
+  ConwayEraGov era => NewEpochState era -> Anchor (EraCrypto era)
```
lehins: All queries must be stable, we can't change their return type if that type has a different binary representation. So this function must continue to return a `Maybe (SafeHash`

Another unfortunate restriction imposed on queries is that the must build for all eras, however `ConwayEraGov` prevents that from happening.

lehins (changes requested):

eras/conway/impl/src/Cardano/Ledger/Conway/Gov.hs
```diff
@@ -381,8 +382,13 @@ toConwayGovPairs cg@(ConwayGovState _ _) =
 
 instance EraPParams (ConwayEra c) => EraGov (ConwayEra c) where
   type GovState (ConwayEra c) = ConwayGovState (ConwayEra c)
-  getConstitutionHash g = Just $ g ^. cgRatifyL . rsEnactStateL . ensConstitutionL . constitutionHashL
 
   curPParamsGovStateL = curPParamsConwayGovStateL
 
   prevPParamsGovStateL = prevPParamsConwayGovStateL
+
+class EraGov era => ConwayEraGov era where
```
aniketd: Awesome! Thank you! @Soupstraw

eras/shelley/impl/src/Cardano/Ledger/Shelley/Gov.hs
aniketd: @teodanciu It was kept here only because the previous functionality was implemented in Shelley :slightly_frowning_face: . I didn't have the time to introduce a `ConwayEraGov` before our demonstration. This is clean-up. Although, I remember we decided that we will break downstream quickly after and move it from Shelley to Conway-only.

lehins: Replaced this function with `getConstitution`, since we do need something like that for consensus queries

lehins (approved):
I've made a few minor changes in order to get this PR over the line.

eras/conway/impl/CHANGELOG.md
lehins: Done

PR #3628, opened 2023-08-09
DELEG, POOL, GOVCERT conformance with Spec v0.8

### LINKED ISSUE #3442: Conway Certs
The main `DELEG` rule that has been largely unchanged since the Shelley era will see some changes in Conway:

- [x] It will allow for DRep registration, delegation, and retirement
- [ ] It will check the stake key registration deposit amount, if specified in the cert
- [x] It will add new CC hot key delegation and cold key retirements
- [x] It will not include the MIR cert support

Note that the DRep delegation sub-rule will be storing the new DRep delegation relation in the UMap.

aniketd:
# Description

Resolves #3442 , #3638, #3646, #3649

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/eras/conway/impl/CHANGELOG.md
+++ b/eras/conway/impl/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## 1.7.0.0
 
+* Make `DELEG`, `POOL` and `GOVCERT` conform to spec-v0.8 #3628
+  * Add `CertEnv` and `CertsEnv` to pass `EpochNo` down from `LEDGER` to sub-rules
+  * Add `drepDeposit` to `DRepState` to track deposits paid by `DRep`s
+  * Update `DRep` expiry in `LEDGER` for all `DRep`s who are voting in current `Tx`
 * Add `ConwayGovCertEnv`
 * Change the environment of `GOVCERT` to `ConwayGovCertEnv`
 * Add `ConwayEraGov` with `constitutionGovStateL`
--- a/eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Cert.hs
+++ b/eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Cert.hs
@@ -18,9 +18,10 @@ module Cardano.Ledger.Conway.Rules.Cert (
   ConwayCERT,
   ConwayCertPredFailure (..),
   ConwayCertEvent,
+  CertEnv (..),
 ) where
 
-import Cardano.Ledger.BaseTypes (ShelleyBase)
+import Cardano.Ledger.BaseTypes (EpochNo, ShelleyBase, SlotNo)
 import Cardano.Ledger.Binary (DecCBOR (..), EncCBOR (..))
 import Cardano.Ledger.Binary.Coders
 import Cardano.Ledger.Conway.Core
@@ -43,8 +44,6 @@ import Cardano.Ledger.Conway.TxCert (
 import Cardano.Ledger.Shelley.API (
   CertState (..),
   DState,
-  DelegEnv (DelegEnv),
-  DelplEnv (DelplEnv),
   PState,
   PoolEnv (PoolEnv),
   VState,
@@ -64,6 +63,12 @@ import Control.State.Transition.Extended (
 import Data.Typeable (Typeable)
 import GHC.Generics (Generic)
 import NoThunks.Class (NoThunks)
+
+data CertEnv era = CertEnv
+  { ceSlotNo :: !SlotNo
+  , cePParams :: !(PParams era)
+  , ceCurrentEpoch :: !EpochNo
+  }
 
 data ConwayCertPredFailure era
   = DelegFailure (PredicateFailure (EraRule "DELEG" era))
@@ -110,7 +115,7 @@ instance
   , State (EraRule "DELEG" era) ~ DState era
   , State (EraRule "POOL" era) ~ PState era
   , State (EraRule "GOVCERT" era) ~ VState era
-  , Environment (EraRule "DELEG" era) ~ DelegEnv era
+  , Environment (EraRule "DELEG" era) ~ PParams era
   , Environment (EraRule "POOL" era) ~ PoolEnv era
   , Environment (EraRule "GOVCERT" era) ~ ConwayGovCertEnv era
   , Signal (EraRule "DELEG" era) ~ ConwayDelegCert (EraCrypto era)
@@ -125,7 +130,7 @@ instance
   where
   type State (ConwayCERT era) = CertState era
   type Signal (ConwayCERT era) = TxCert era
-  type Environment (ConwayCERT era) = DelplEnv era
+  type Environment (ConwayCERT era) = CertEnv era
   type BaseM (ConwayCERT era) = ShelleyBase
   type PredicateFailure (ConwayCERT era) = ConwayCertPredFailure era
   type Event (ConwayCERT era) = ConwayCertEvent era
@@ -137,7 +142,7 @@ certTransition ::
   ( State (EraRule "DELEG" era) ~ DState era
   , State (EraRule "POOL" era) ~ PState era
   , State (EraRule "GOVCERT" era) ~ VState era
-  , Environment (EraRule "DELEG" era) ~ DelegEnv era
+  , Environment (EraRule "DELEG" era) ~ PParams era
   , Environment (EraRule "POOL" era) ~ PoolEnv era
   , Environment (EraRule "GOVCERT" era) ~ ConwayGovCertEnv era
   , Signal (EraRule "DELEG" era) ~ ConwayDelegCert (EraCrypto era)
@@ -150,17 +155,17 @@ certTransition ::
   ) =>
   TransitionRule (ConwayCERT era)
 certTransition = do
-  TRC (DelplEnv slot ptr pp acnt, cState, c) <- judgmentContext
+  TRC (CertEnv slot pp currentEpoch, cState, c) <- judgmentContext
   let CertState {certDState, certPState, certVState} = cState
   case c of
     ConwayTxCertDeleg delegCert -> do
-      newDState <- trans @(EraRule "DELEG" era) $ TRC (DelegEnv slot ptr acnt pp, certDState, delegCert)
+      newDState <- trans @(EraRule "DELEG" era) $ TRC (pp, certDState, delegCert)
       pure $ cState {certDState = newDState}
     ConwayTxCertPool poolCert -> do
       newPState <- trans @(EraRule "POOL" era) $ TRC (PoolEnv slot pp, certPState, poolCert)
       pure $ cState {certPState = newPState}
     ConwayTxCertGov govCert -> do
-      newVState <- trans @(EraRule "GOVCERT" era) $ TRC (ConwayGovCertEnv pp slot, certVState, govCert)
+      newVState <- trans @(EraRule "GOVCERT" era) $ TRC (ConwayGovCertEnv pp currentEpoch, certVState, govCert)
       pure $ cState {certVState = newVState}
 
 instance
--- a/eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Certs.hs
+++ b/eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Certs.hs
@@ -19,9 +19,10 @@ module Cardano.Ledger.Conway.Rules.Certs (
   ConwayCERTS,
   ConwayCertsPredFailure (..),
   ConwayCertsEvent (..),
+  CertsEnv (..),
 ) where
 
-import Cardano.Ledger.BaseTypes (Globals (..), ShelleyBase, mkCertIxPartial)
+import Cardano.Ledger.BaseTypes (EpochNo, Globals (..), ShelleyBase, SlotNo)
 import Cardano.Ledger.Binary (DecCBOR (..), EncCBOR (..))
 import Cardano.Ledger.Binary.Coders (
   Decode (..),
@@ -31,20 +32,29 @@ import Cardano.Ledger.Binary.Coders (
   (!>),
   (<!),
  )
+import Cardano.Ledger.CertState (certDStateL, certVStateL, vsDRepsL)
+import Cardano.Ledger.Conway.Core (
+  ConwayEraPParams,
+  ConwayEraTxBody (votingProceduresTxBodyL),
+  Era (EraCrypto),
+  EraRule,
+  EraTx (Tx, bodyTxL),
+  EraTxBody (withdrawalsTxBodyL),
+  EraTxCert (TxCert),
+  PParams,
+  ppDRepActivityL,
+ )
 import Cardano.Ledger.Conway.Era (ConwayCERT, ConwayCERTS)
-import Cardano.Ledger.Conway.Rules.Cert (ConwayCertEvent, ConwayCertPredFailure)
-import Cardano.Ledger.Core
+import Cardano.Ledger.Conway.Governance (Voter (DRepVoter), VotingProcedures (unVotingProcedures))
+import Cardano.Ledger.Conway.Rules.Cert (CertEnv (CertEnv), ConwayCertEvent, ConwayCertPredFailure)
+import Cardano.Ledger.DRepDistr (drepExpiryL)
 import Cardano.Ledger.Shelley.API (
   CertState (..),
   Coin,
-  DelegsEnv (DelegsEnv),
-  DelplEnv (DelplEnv),
   KeyHash,
   KeyRole (..),
-  Ptr (Ptr),
   RewardAcnt,
  )
-import Cardano.Ledger.Shelley.Core (ShelleyEraTxBody)
 import Cardano.Ledger.Shelley.Rules (
   drainWithdrawals,
   validateDelegationRegistered,
@@ -61,11 +71,18 @@ import Control.State.Transition.Extended (
   trans,
   validateTrans,
  )
-import Data.Map.Strict (Map)
+import qualified Data.Map.Strict as Map
 import Data.Sequence (Seq (..))
 import GHC.Generics (Generic)
-import Lens.Micro ((^.))
+import Lens.Micro
 import NoThunks.Class (NoThunks (..))
+
+data CertsEnv era = CertsEnv
+  { certsTx :: !(Tx era)
+  , certsPParams :: !(PParams era)
+  , certsSlotNo :: !SlotNo
+  , certsCurrentEpoch :: !EpochNo
+  }
 
 data ConwayCertsPredFailure era
   = -- | Target pool which is not registered
@@ -73,7 +90,7 @@ data ConwayCertsPredFailure era
       !(KeyHash 'StakePool (EraCrypto era))
   | -- | Withdrawals that are missing or do not withdrawal the entire amount
     WithdrawalsNotInRewardsCERTS
-      !(Map (RewardAcnt (EraCrypto era)) Coin)
+      !(Map.Map (RewardAcnt (EraCrypto era)) Coin)
   | -- | CERT rule subtransition Failures
     CertFailure !(PredicateFailure (EraRule "CERT" era))
   deriving (Generic)
@@ -118,17 +135,18 @@ instance
 
 instance
   ( EraTx era
-  , ShelleyEraTxBody era
+  , ConwayEraTxBody era
+  , ConwayEraPParams era
   , State (EraRule "CERT" era) ~ CertState era
   , Signal (EraRule "CERT" era) ~ TxCert era
-  , Environment (EraRule "CERT" era) ~ DelplEnv era
+  , Environment (EraRule "CERT" era) ~ CertEnv era
   , Embed (EraRule "CERT" era) (ConwayCERTS era)
   ) =>
   STS (ConwayCERTS era)
   where
   type State (ConwayCERTS era) = CertState era
   type Signal (ConwayCERTS era) = Seq (TxCert era)
-  type Environment (ConwayCERTS era) = DelegsEnv era
+  type Environment (ConwayCERTS era) = CertsEnv era
   type BaseM (ConwayCERTS era) = ShelleyBase
   type
     PredicateFailure (ConwayCERTS era) =
@@ -140,34 +158,49 @@ instance
 conwayCertsTransition ::
   forall era.
   ( EraTx era
-  , ShelleyEraTxBody era
+  , ConwayEraTxBody era
+  , ConwayEraPParams era
   , State (EraRule "CERT" era) ~ CertState era
   , Embed (EraRule "CERT" era) (ConwayCERTS era)
-  , Environment (EraRule "CERT" era) ~ DelplEnv era
+  , Environment (EraRule "CERT" era) ~ CertEnv era
   , Signal (EraRule "CERT" era) ~ TxCert era
   ) =>
   TransitionRule (ConwayCERTS era)
 conwayCertsTransition = do
-  TRC (env@(DelegsEnv slot txIx pp tx acnt), certState, certificates) <- judgmentContext
+  TRC
+    ( env@(CertsEnv tx pp slot currentEpoch)
+      , certState
+      , certificates
+      ) <-
+    judgmentContext
   network <- liftSTS $ asks networkId
 
   case certificates of
     Empty -> do
-      let dState = certDState certState
+      -- Update DRep expiry for all DReps that are voting in this transaction
+      let drepActivity = pp ^. ppDRepActivityL
+          updatedVSDReps =
+            Map.foldlWithKey'
+              ( \dreps voter _ -> case voter of
+                  DRepVoter cred -> Map.adjust (drepExpiryL .~ currentEpoch + drepActivity) cred dreps
+                  _ -> dreps
+              )
+              (certState ^. certVStateL . vsDRepsL)
+              (unVotingProcedures $ tx ^. bodyTxL . votingProceduresTxBodyL)
+          certStateWithDRepExpiryUpdated = certState & certVStateL . vsDRepsL .~ updatedVSDReps
+          dState = certStateWithDRepExpiryUpdated ^. certDStateL
           withdrawals = tx ^. bodyTxL . withdrawalsTxBodyL
+      -- Validate withdrawals and rewards and drain withdrawals
       validateTrans WithdrawalsNotInRewardsCERTS $
         validateZeroRewards dState withdrawals network
-      pure $ certState {certDState = drainWithdrawals dState withdrawals}
+      pure $ certStateWithDRepExpiryUpdated & certDStateL .~ drainWithdrawals dState withdrawals
     gamma :|> c -> do
       certState' <-
         trans @(ConwayCERTS era) $ TRC (env, certState, gamma)
       validateTrans DelegateeNotRegisteredDELEG $
         validateDelegationRegistered certState' c
-      -- It is impossible to have 65535 number of certificates in a
-      -- transaction, therefore partial function is justified.
-      let ptr = Ptr slot txIx (mkCertIxPartial $ toInteger $ length gamma)
       trans @(EraRule "CERT" era) $
-        TRC (DelplEnv slot ptr pp acnt, certState', c)
+        TRC (CertEnv slot pp currentEpoch, certState', c)
 
 instance
   ( Era era
--- a/eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Deleg.hs
+++ b/eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Deleg.hs
@@ -34,11 +34,10 @@ import Cardano.Ledger.Conway.TxCert (
   ConwayDelegCert (ConwayDelegCert, ConwayRegCert, ConwayRegDelegCert, ConwayUnRegCert),
   Delegatee (DelegStake, DelegStakeVote, DelegVote),
  )
-import Cardano.Ledger.Core (Era (EraCrypto), EraPParams, EraRule)
+import Cardano.Ledger.Core (Era (EraCrypto), EraPParams, EraRule, PParams)
 import Cardano.Ledger.Credential (Credential)
 import Cardano.Ledger.Keys (KeyRole (Staking))
 import Cardano.Ledger.Shelley.LedgerState (DState (..))
-import Cardano.Ledger.Shelley.Rules (DelegEnv (DelegEnv))
 import qualified Cardano.Ledger.UMap as UM
 import Control.DeepSeq (NFData)
 import Control.Monad (forM_)
@@ -64,9 +63,9 @@ import NoThunks.Class (NoThunks)
 
 data ConwayDelegPredFailure era
   = IncorrectDepositDELEG !Coin
-  | StakeKeyAlreadyRegisteredDELEG !(Credential 'Staking (EraCrypto era))
+  | StakeKeyRegisteredDELEG !(Credential 'Staking (EraCrypto era))
   | StakeKeyNotRegisteredDELEG !(Credential 'Staking (EraCrypto era))
-  | StakeKeyHasNonZeroAccountBalanceDELEG !Coin
+  | StakeKeyHasNonZeroRewardAccountBalanceDELEG !Coin
   | DRepAlreadyRegisteredForStakeKeyDELEG !(Credential 'Staking (EraCrypto era))
   | WrongCertificateTypeDELEG
   deriving (Show, Eq, Generic)
@@ -80,12 +79,12 @@ instance Era era => EncCBOR (ConwayDelegPredFailure era) where
     encode . \case
       IncorrectDepositDELEG mCoin ->
         Sum (IncorrectDepositDELEG @era) 1 !> To mCoin
-      StakeKeyAlreadyRegisteredDELEG stakeCred ->
-        Sum (StakeKeyAlreadyRegisteredDELEG @era) 2 !> To stakeCred
+      StakeKeyRegisteredDELEG stakeCred ->
+        Sum (StakeKeyRegisteredDELEG @era) 2 !> To stakeCred
       StakeKeyNotRegisteredDELEG stakeCred ->
         Sum (StakeKeyNotRegisteredDELEG @era) 3 !> To stakeCred
-      StakeKeyHasNonZeroAccountBalanceDELEG mCoin ->
-        Sum (StakeKeyHasNonZeroAccountBalanceDELEG @era) 4 !> To mCoin
+      StakeKeyHasNonZeroRewardAccountBalanceDELEG mCoin ->
+        Sum (StakeKeyHasNonZeroRewardAccountBalanceDELEG @era) 4 !> To mCoin
       DRepAlreadyRegisteredForStakeKeyDELEG stakeCred ->
         Sum (DRepAlreadyRegisteredForStakeKeyDELEG @era) 5 !> To stakeCred
       WrongCertificateTypeDELEG ->
@@ -94,9 +93,9 @@ instance Era era => EncCBOR (ConwayDelegPredFailure era) where
 instance Era era => DecCBOR (ConwayDelegPredFailure era) where
   decCBOR = decode $ Summands "ConwayDelegPredFailure" $ \case
     1 -> SumD IncorrectDepositDELEG <! From
-    2 -> SumD StakeKeyAlreadyRegisteredDELEG <! From
+    2 -> SumD StakeKeyRegisteredDELEG <! From
     3 -> SumD StakeKeyNotRegisteredDELEG <! From
-    4 -> SumD StakeKeyHasNonZeroAccountBalanceDELEG <! From
+    4 -> SumD StakeKeyHasNonZeroRewardAccountBalanceDELEG <! From
     5 -> SumD DRepAlreadyRegisteredForStakeKeyDELEG <! From
     6 -> SumD WrongCertificateTypeDELEG
     n -> Invalid n
@@ -107,14 +106,14 @@ instance
   ( EraPParams era
   , State (EraRule "DELEG" era) ~ DState era
   , Signal (EraRule "DELEG" era) ~ ConwayDelegCert (EraCrypto era)
-  , Environment (EraRule "DELEG" era) ~ DelegEnv era
+  , Environment (EraRule "DELEG" era) ~ PParams era
   , EraRule "DELEG" era ~ ConwayDELEG era
   ) =>
   STS (ConwayDELEG era)
   where
   type State (ConwayDELEG era) = DState era
   type Signal (ConwayDELEG era) = ConwayDelegCert (EraCrypto era)
-  type Environment (ConwayDELEG era) = DelegEnv era
+  type Environment (ConwayDELEG era) = PParams era
   type BaseM (ConwayDELEG era) = ShelleyBase
   type PredicateFailure (ConwayDELEG era) = ConwayDelegPredFailure era
   type Event (ConwayDELEG era) = ConwayDelegEvent era
@@ -124,7 +123,7 @@ instance
 conwayDelegTransition :: forall era. EraPParams era => TransitionRule (ConwayDELEG era)
 conwayDelegTransition = do
   TRC
-    ( DelegEnv _slot _ptr _acnt pp
+    ( pp
       , dState@DState {dsUnified}
       , c
       ) <-
@@ -133,34 +132,29 @@ conwayDelegTransition = do
   case c of
     ConwayRegCert stakeCred sMayDeposit -> do
       forM_ sMayDeposit $ checkDepositAgainstPParams ppKeyDeposit
-      checkStakeKeyNotAlreadyRegistered stakeCred dsUnified
-      pure $ dState {dsUnified = acceptDepositForStakeKey stakeCred dsUnified ppKeyDeposit}
+      dsUnified' <- checkAndAcceptDepositForStakeCred stakeCred ppKeyDeposit dsUnified
+      pure $ dState {dsUnified = dsUnified'}
     ConwayUnRegCert stakeCred sMayDeposit -> do
-      checkStakeKeyIsAlreadyRegistered stakeCred dsUnified
-      checkStakeKeyHasZeroBalance stakeCred dsUnified
+      checkStakeKeyIsRegistered stakeCred dsUnified
+      checkStakeKeyHasZeroRewardBalance stakeCred dsUnified
       forM_ sMayDeposit $ checkDepositAgainstPaidDeposit stakeCred dsUnified
-      pure $
-        dState
-          { dsUnified = UM.domDeleteAll (Set.singleton stakeCred) dsUnified
-          }
+      pure $ dState {dsUnified = UM.domDeleteAll (Set.singleton stakeCred) dsUnified}
     ConwayDelegCert stakeCred delegatee -> do
-      checkStakeKeyIsAlreadyRegistered stakeCred dsUnified
-      pure $
-        dState
-          { dsUnified = processDelegation stakeCred delegatee dsUnified
-          }
+      checkStakeKeyIsRegistered stakeCred dsUnified
+      pure $ dState {dsUnified = processDelegation stakeCred delegatee dsUnified}
     ConwayRegDelegCert stakeCred delegatee deposit -> do
-      deposit == ppKeyDeposit ?! IncorrectDepositDELEG deposit
-      checkStakeKeyNotAlreadyRegistered stakeCred dsUnified
-      pure $
-        dState
-          { dsUnified =
-              processDelegation stakeCred delegatee $
-                acceptDepositForStakeKey stakeCred dsUnified ppKeyDeposit
-          }
+      checkDepositAgainstPParams ppKeyDeposit deposit
+      dsUnified' <- checkAndAcceptDepositForStakeCred stakeCred deposit dsUnified
+      pure $ dState {dsUnified = processDelegation stakeCred delegatee dsUnified'}
   where
-    acceptDepositForStakeKey stakeCred dsUnified ppKeyDeposit =
-      UM.RewDepUView dsUnified UM.∪ (stakeCred, UM.RDPair (UM.CompactCoin 0) (UM.compactCoinOrError ppKeyDeposit))
+    -- Whenever we want to accept new deposit, we must always check if the stake credential isn't already registered.
+    checkAndAcceptDepositForStakeCred stakeCred deposit dsUnified = do
+      checkStakeKeyNotRegistered stakeCred dsUnified
+      -- This looks like it should have been a right-biased union, so that the (reward, deposit) pair would be inserted
+      -- (or overwritten) in the UMap. But since we are sure that the stake credential isn't a member yet
+      -- it will still work. The reason we cannot use a right-biased union here is because UMap treats deposits specially
+      -- in right-biased unions, and is unable to accept new deposits.
+      pure $ UM.RewDepUView dsUnified UM.∪ (stakeCred, UM.RDPair (UM.CompactCoin 0) (UM.compactCoinOrError deposit))
     delegStake stakeCred sPool dsUnified =
       UM.SPoolUView dsUnified UM.⨃ Map.singleton stakeCred sPool
     delegVote stakeCred dRep dsUnified =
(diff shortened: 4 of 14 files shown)


eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Deleg.hs
```diff
@@ -121,6 +121,34 @@ instance
...
+--   requiredDeposit pp (just x) = PParams.poolDeposit pp
+--   requiredDeposit pp nothing = 0
+--
+--   requiredVDelegDeposit : PParams → Maybe VDeleg → Coin
+--   requiredVDelegDeposit pp (just _) = PParams.poolDeposit pp
+--   requiredVDelegDeposit pp nothing = 0
+--
+--   insertIfPresent : ∀ {A B} → ⦃ DecEq A ⦄ → A → Maybe B → A ⇀ B → A ⇀ B
+--   insertIfPresent x nothing  m = m
+--   insertIfPresent x (just y) m = insert m x y
```
aniketd: This is `insertIfPresent` in the certificate, not in the UMap.

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Deleg.hs
```diff
@@ -136,31 +164,42 @@ conwayDelegTransition = do
...
     ConwayUnRegCert stakeCred sMayDeposit -> do
-      checkStakeKeyIsAlreadyRegistered stakeCred dsUnified
-      checkStakeKeyHasZeroBalance stakeCred dsUnified
+      checkStakeKeyIsRegistered stakeCred dsUnified
+      checkStakeKeyHasZeroBalanceReward stakeCred dsUnified
       forM_ sMayDeposit $ checkDepositAgainstPaidDeposit stakeCred dsUnified
       pure $
         dState
-          { dsUnified = UM.domDeleteAll (Set.singleton stakeCred) dsUnified
+          { dsUnified = UM.domDeleteAll (Set.singleton stakeCred) dsUnified -- QUESTION: @WhatisRT: We auto-un-register him as a DRep?
```
WhatisRT: This is stake credential deregistration, correct? If so, nothing should happen with the DRep registry.

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Deleg.hs
```diff
@@ -136,31 +164,42 @@ conwayDelegTransition = do
...
           }
     ConwayDelegCert stakeCred delegatee -> do
-      checkStakeKeyIsAlreadyRegistered stakeCred dsUnified
+      checkStakeKeyIsRegistered stakeCred dsUnified
+      -- QUESTION: @WhatisRT: One cannot delegate if nothing is staked? This seems wrong because `processDelegation` also
```
WhatisRT: I'm not super familiar with the UMap. In the spec there's the rewards map, and being present in the domain of that map means you've paid a deposit. If the UMap just adds a bunch of fields to the values then it should have the same semantics I guess. I hope that's documented somewhere?

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Deleg.hs
```diff
@@ -136,31 +164,42 @@ conwayDelegTransition = do
...
   where
-    acceptDepositForStakeKey stakeCred dsUnified ppKeyDeposit =
-      UM.RewDepUView dsUnified UM.∪ (stakeCred, UM.RDPair (UM.CompactCoin 0) (UM.compactCoinOrError ppKeyDeposit))
+    -- QUESTION: @WhatisRT: Why don't we track DRep deposits in VState like we track SPool deposits in DState,
+    -- to be able to return them on retirement after checking against original paid deposit? Both ppDRepDeposit
+    -- as well as ppKeyDeposit parameters are subject to change over time.
+
+    acceptDepositForStakeKey stakeCred dsUnified deposit =
+      UM.RewDepUView dsUnified UM.∪ (stakeCred, UM.RDPair (UM.CompactCoin 0) (UM.compactCoinOrError deposit))
+    -- \^ QUESTION: @WhatisRT: This seems wrong, because `unionL` will keep an old value if the key is present.
```
WhatisRT: I'm not sure I understand this question. The left-biased union in the spec guarantees that the values on the left end up in the map, but the ones on the right might not. But `DELEG-delegate` doesn't have that union.

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Deleg.hs
```diff
@@ -136,31 +164,42 @@ conwayDelegTransition = do
...
       pure $
         dState
           { dsUnified =
               processDelegation stakeCred delegatee $
                 acceptDepositForStakeKey stakeCred dsUnified ppKeyDeposit
           }
   where
-    acceptDepositForStakeKey stakeCred dsUnified ppKeyDeposit =
-      UM.RewDepUView dsUnified UM.∪ (stakeCred, UM.RDPair (UM.CompactCoin 0) (UM.compactCoinOrError ppKeyDeposit))
+    -- QUESTION: @WhatisRT: Why don't we track DRep deposits in VState like we track SPool deposits in DState,
```
WhatisRT: They should be tracked somewhere, there's a big map for them in the spec but I don't know where they'll end up in the code.

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Deleg.hs
aniketd: Oh got it! Sorry for the wrong question here. "unregistering" will result in not only removing the stake deposit and delegation, but also the DRep delegation. That's what I meant to ask about. `ConwayUnRegCert` currently removes even DRep delegation, because DRep delegation is also efficiently kept in the same UMap, that is keyed by stake credential.

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Deleg.hs
aniketd: > In the spec there's the rewards map, and being present in the domain of that map means you've paid a deposit.

Yes, it is so even in Haskell :+1: 

> If the UMap just adds a bunch of fields to the values then it should have the same semantics I guess. I hope that's documented somewhere?

Yes indeed. And it is documented in the `UMap` module :+1:. The fields added to just the values happen to be
1. The (reward, deposit) pair
2. The stake pool key
3. The drep (key, script, always-noconfidence, or always-abstain)
4. (inconsequential) The set of pointers.

About the `checkStakeKeyIsRegistered`, it results in a predicate failure unless the stake credential is already registered, that is, the stake credential already has made a staking deposit. Should this check be present or not over here? I'm guessing you mean that it _should_ be present, since the stake deposit is expected to be present for the stake credential at this point, and only delegation should be allowed here.

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Deleg.hs
aniketd: Got it! So this is indeed not completely correct. This needs to be a right-biased union here, so that the (reward, deposit) pair is inserted (or overwritten) in the `UMap`. But since we are sure that the stake credential isn't a member yet, it will still work. The right-biased union for `UMap` treats deposits specially, and is unable to accept new deposits.

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Deleg.hs
aniketd: Opened https://github.com/input-output-hk/cardano-ledger/issues/3638 to track this.

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/GovCert.hs
```diff
@@ -164,48 +167,57 @@ conwayGovCertTransition ::
...
-      pure $ vState & vsDRepsL %~ Map.update (Just . updateDRepState) cred
+      expiryEpoch <- calcDRepExpiryEpoch cgceCurrentSlot $ cgcePParams ^. ppDRepActivityL
+      pure
+        vState
+          { vsDReps =
+              Map.adjust
+                (\drepState -> drepState {drepExpiry = expiryEpoch, drepAnchor = mAnchor})
+                cred
+                vsDReps
+          }
```
aniketd, line 186: This change from using lenses to record updates, is only in favor of consistency. The lenses implementation is also correct.

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Certs.hs
```diff
@@ -166,6 +175,7 @@ conwayCertsTransition = do
       -- It is impossible to have 65535 number of certificates in a
       -- transaction, therefore partial function is justified.
       let ptr = Ptr slot txIx (mkCertIxPartial $ toInteger $ length gamma)
+      -- TODO: @aniketd: Update drepActivity for all DReps that are voting in this CERTS transition.
```
aniketd: TODO

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Deleg.hs
```diff
@@ -66,7 +66,7 @@ data ConwayDelegPredFailure era
   = IncorrectDepositDELEG !Coin
   | StakeKeyAlreadyRegisteredDELEG !(Credential 'Staking (EraCrypto era))
   | StakeKeyNotRegisteredDELEG !(Credential 'Staking (EraCrypto era))
-  | StakeKeyHasNonZeroAccountBalanceDELEG !Coin
+  | StakeKeyHasNonZeroAccountBalanceRewardDELEG !Coin
```
Soupstraw: I'm not sure if adding `Reward` at the end makes anything clearer. I'm not sure what the predicate failure means either, so maybe it's indeed necessary.

Soupstraw: hmm or maybe it'd be clearer if it was `StakeKeyHasNonZeroRewardAccountBalanceDELEG`

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Deleg.hs
```diff
@@ -135,22 +157,19 @@ conwayDelegTransition = do
...
       checkStakeKeyNotAlreadyRegistered stakeCred dsUnified
       pure $ dState {dsUnified = acceptDepositForStakeKey stakeCred dsUnified ppKeyDeposit}
+    -- TODO: https://github.com/input-output-hk/cardano-ledger/issues/3638
+    -- DELEG is called before UTXO. So removing the UMap entry for UnReg causes a problem in Conway.
+    -- The UTXO rule therefore does not have any information about how much to refund to retiring SPOs and DReps.
     ConwayUnRegCert stakeCred sMayDeposit -> do
-      checkStakeKeyIsAlreadyRegistered stakeCred dsUnified
-      checkStakeKeyHasZeroBalance stakeCred dsUnified
+      checkStakeKeyIsRegistered stakeCred dsUnified
+      checkStakeKeyHasZeroBalanceReward stakeCred dsUnified
```
Soupstraw: here as well, I think `checkStakeKeyHasZeroRewardBalance` would be clearer

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/GovCert.hs
```diff
@@ -164,48 +167,57 @@ conwayGovCertTransition ::
...
+    --   ────────────────────────────────
+    --   ⟦ e , pp , vs ⟧ᶜ ⊢ ⟦ dReps , ccKeys ⟧ᵛ ⇀⦇ regdrep c d an ,GOVCERT⦈
+    --                      ⟦ ❴ c , e +ᵉ' drepActivity ❵ᵐ ∪ᵐˡ dReps , ccKeys ⟧ᵛ
+    ConwayRegDRep cred deposit mAnchor -> do
+      (cgcePParams ^. ppDRepDepositL == deposit && Map.notMember cred vsDReps)
+        || (Coin 0 == deposit && Map.member cred vsDReps)
+          ?! ConwayDRepRegMalformedRequest deposit cred
+      expiryEpoch <- calcDRepExpiryEpoch cgceCurrentSlot $ cgcePParams ^. ppDRepActivityL
+      let drepState = DRepState expiryEpoch mAnchor
       pure $ vState {vsDReps = Map.insert cred drepState vsDReps}
```
Soupstraw: I think we should leave this as it was before, because we decided to handle the right side of `||` using the `ConwayUpdateDRep` certificate.

Soupstraw: The right hand case is used if a somebody wants to update the expiry of an already existing DRep.

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Pool.hs
```diff
@@ -175,6 +175,12 @@ poolCertTransition ::
   PoolCert c ->
   TransitionRule (ledger era)
 poolCertTransition (PoolEnv slot pp) ps@PState {psStakePoolParams, psFutureStakePoolParams, psRetiring} = \case
+  -- FOR CONWAY
+  --
+  --   POOL-regpool : let open PParams pp ; open PoolParams poolParams in
+  --     c ∉ dom (pools ˢ)
+  --     ────────────────────────────────
+  --     pp ⊢ ⟦ pools , retiring ⟧ᵖ ⇀⦇ regpool c poolParams ,POOL⦈ ⟦ ❴ c , poolParams ❵ᵐ ∪ᵐˡ pools , retiring ⟧ᵖ
```
Soupstraw: I think the logic is actually the same in Conway as it was in Babbage, the spec just doesn't reflect every detail from the previous eras.

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Pool.hs
```diff
@@ -196,7 +202,7 @@ poolCertTransition (PoolEnv slot pp) ps@PState {psStakePoolParams, psFutureStake
         tellEvent $ RegisterPool ppId
         pure $
           payPoolDeposit ppId pp $
-            ps {psStakePoolParams = eval (psStakePoolParams ∪ singleton ppId poolParams)}
+            ps {psStakePoolParams = eval (psStakePoolParams ⨃ singleton ppId poolParams)}
```
Soupstraw, line 199: nice!

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Pool.hs
```diff
@@ -211,6 +217,11 @@ poolCertTransition (PoolEnv slot pp) ps@PState {psStakePoolParams, psFutureStake
             { psFutureStakePoolParams = eval (psFutureStakePoolParams ⨃ singleton ppId poolParams)
             , psRetiring = eval (setSingleton ppId ⋪ psRetiring)
             }
+  -- FOR CONWAY
+  --
+  --   POOL-retirepool : let open PoolParams poolParams in
+  --     pp ⊢ ⟦ pools , retiring ⟧ᵖ ⇀⦇ retirepool c e ,POOL⦈
+  --          ⟦ pools , ❴ c , e ❵ᵐ ∪ᵐˡ retiring ⟧ᵖ
```
Soupstraw: This should also not change between Babbage and Conway.

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Pool.hs
Soupstraw: hmm, actually looking at it again, I don't think it changes anything, since we're making sure that the `ppId` isn't in the map before we add the pool params.

Soupstraw (changes requested):
Thanks for going over the code, it's always nice to have someone double-check these things. I think it would make more sense to turn some of the TODO comments in this PR into issues and then maybe merge some of the predicate failure improvements. The logic already seemed to be correct, so I don't think it should be changed.

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Pool.hs
Soupstraw: Ok, a few things might actually change, but they're not in the spec yet

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Deleg.hs
aniketd: :raised_hands:

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Pool.hs
aniketd: Got it! Should I leave this here and every time we update this to match an updated version of the spec, we paste the updated version in? What do you prefer?

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Pool.hs
aniketd: Yes, I agree it makes no difference, though considering just the single-line, the newer version would be the more appropriate operation, as it reflects the intention, irrespective of the pre-condition. I'm happy to revert this change back unless you are fine with it :sweat_smile:

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Pool.hs
aniketd: Noted! Do you want me to take this out or keep it though?

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/GovCert.hs
aniketd: Ah! Thank you! :raised_hands:

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Pool.hs
Soupstraw: I think it's better to just refer to the spec, since the comments can easily get outdated.

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Cert.hs
```diff
@@ -65,6 +67,14 @@ import Data.Typeable (Typeable)
 import GHC.Generics (Generic)
 import NoThunks.Class (NoThunks)
 
+data CertEnv era = CertEnv
+  { certSlotNo :: !SlotNo
+  , certPtr :: !Ptr
+  , certPParams :: !(PParams era)
+  , certAccountState :: !AccountState
+  , certCurrentEpoch :: !EpochNo
```
lehins: It is better to follow the same convention that we used with other types, namely prefix with just the first letters `[C]ert[E]nv` -> `ce`
```suggestion
  { ceSlotNo :: !SlotNo
  , cePtr :: !Ptr
  , cePParams :: !(PParams era)
  , ceAccountState :: !AccountState
  , ceCurrentEpoch :: !EpochNo
```

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/GovCert.hs
```diff
@@ -62,19 +56,24 @@ import Data.Maybe (isNothing)
 import Data.Typeable (Typeable)
 import Data.Word (Word8)
 import GHC.Generics (Generic)
-import Lens.Micro ((%~), (&), (.~), (^.))
+import Lens.Micro ((^.))
 import NoThunks.Class (NoThunks (..))
 
 data ConwayGovCertEnv era = ConwayGovCertEnv
   { cgcePParams :: !(PParams era)
   , cgceCurrentSlot :: !SlotNo
```
lehins: I believe that the slot number is no longer needed in the GOVCERT rule
```suggestion
```

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Certs.hs
```diff
@@ -137,20 +161,54 @@ instance
...
 
+-- | CERT-base :
+--     let open PParams pp; open CertState st; open GState gState
+--         refresh = mapPartial (λ v → let open GovVote v in case role of λ where
+--           GovRole.DRep → just credential
+--           _    → nothing) (fromList vs)
+--     in ⊤ -- TODO: check that the withdrawals are correct here
+--     ────────────────────────────────
+--     ⟦ e , pp , vs ⟧ᶜ ⊢ st ⇀⦇ _ ,CERTBASE⦈ record st { gState = record gState
+--                          { dreps = constMap refresh (e +ᵉ' drepActivity) ∪ᵐˡ dreps } }
```
lehins: I don't think we should embed parts of the spec into the code. Definitely not now, since the spec can still change.
```suggestion
```

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Certs.hs
```diff
@@ -137,20 +161,54 @@ instance
...
+              DRepVoter cred -> Just cred
+              _ -> Nothing
+          )
+          (Map.keys . unVotingProcedures $ tx ^. bodyTxL . votingProceduresTxBodyL)
+      certState =
+        certStateFresh
+          & certVStateL . vsDRepsL
+            %~ Map.mapWithKey
+              ( \cred state ->
+                  if cred `elem` votingDRepCreds
```
lehins: Never use `elem` with lists inside of a loop. This is an O(n) operation, which means inside a loop you get quadratic behavior: O(n*m)

`votingDRepCreds` should be a Set instead, which makes it O(m*log(n)) complexity

```suggestion
                  if cred `Set.member` votingDRepCreds
```

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Certs.hs
```diff
@@ -137,20 +161,54 @@ instance
...
 
+  -- Update DRep expiry for all DReps that are voting in this transaction
+  let drepActivity = pp ^. ppDRepActivityL
+      votingDRepCreds =
+        mapMaybe
+          ( \case
+              DRepVoter cred -> Just cred
+              _ -> Nothing
+          )
+          (Map.keys . unVotingProcedures $ tx ^. bodyTxL . votingProceduresTxBodyL)
```
lehins: This is for my previous comment. Making `votingDRepCreds` into a Set:
```suggestion
        Map.keysSet $ Map.mapMaybeWithKey
          ( \voter _ -> case voter of
              DRepVoter cred -> Just cred
              _ -> Nothing
          )
          (unVotingProcedures $ tx ^. bodyTxL . votingProceduresTxBodyL)
```

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Deleg.hs
```diff
@@ -121,6 +121,28 @@ instance
...
+--
+--   insertIfPresent : ∀ {A B} → ⦃ DecEq A ⦄ → A → Maybe B → A ⇀ B → A ⇀ B
+--   insertIfPresent x nothing  m = m
+--   insertIfPresent x (just y) m = insert m x y
+--
+--   DELEG-delegate :
+--     d ≡ requiredVDelegDeposit pp mv ⊔ requiredDeposit pp mc
+--     ────────────────────────────────
+--     pp ⊢ ⟦ vDelegs , sDelegs , rwds ⟧ᵈ ⇀⦇ delegate c mv mc d ,DELEG⦈
+--          ⟦ insertIfPresent c mv vDelegs , insertIfPresent c mc sDelegs , rwds ⟧ᵈ
```
lehins: Same here
```suggestion
```

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Deleg.hs
```diff
@@ -159,8 +175,14 @@ conwayDelegTransition = do
...
-    acceptDepositForStakeKey stakeCred dsUnified ppKeyDeposit =
-      UM.RewDepUView dsUnified UM.∪ (stakeCred, UM.RDPair (UM.CompactCoin 0) (UM.compactCoinOrError ppKeyDeposit))
+    acceptDepositForStakeKey stakeCred dsUnified deposit =
+      -- NOTE: Always use this for new stake credentials, and call `checkStakeKeyNotAlreadyRegistered` just before.
+      --
+      -- This looks like it should have been a right-biased union, so that the (reward, deposit) pair would be inserted
+      -- (or overwritten) in the UMap. But since we are sure that the stake credential isn't a member yet
+      -- it will still work. The reason we cannot use a right-biased union here is because UMap treats deposits specially
+      -- in right-biased unions, and is unable to accept new deposits.
+      UM.RewDepUView dsUnified UM.∪ (stakeCred, UM.RDPair (UM.CompactCoin 0) (UM.compactCoinOrError deposit))
```
lehins: Why make a comment what has to be always done, if we can actually do it here instead.

To reduce duplication even more we can add deposit validation here as well.

Also a minor nit: it is a stake credential, not just key
```suggestion
    acceptDepositForStakeCred pp stakeCred dsUnified maybeDeposit = do
      forM_ maybeDeposit $ checkDepositAgainstPParams (pp ^. ppKeyDepositL)
      checkStakeKeyNotAlreadyRegistered stakeCred dsUnified
      -- This looks like it should have been a right-biased union, so that the (reward, deposit) pair would be inserted
      -- (or overwritten) in the UMap. But since we are sure that the stake credential isn't a member yet
      -- it will still work. The reason we cannot use a right-biased union here is because UMap treats deposits specially
      -- in right-biased unions, and is unable to accept new deposits.
      pure $ UM.RewDepUView dsUnified UM.∪ (stakeCred, UM.RDPair (UM.CompactCoin 0) (UM.compactCoinOrError deposit))
```
All you need to do is move usage of `acceptDepositForStakeKey` into the monad.

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/GovCert.hs
```diff
@@ -145,67 +144,60 @@ instance
...
       , c
       ) <-
     judgmentContext
   case c of
-    ConwayRegDRep cred _deposit mAnchor -> do
+    -- GOVCERT-regdrep : let open PParams pp in
+    --   (d ≡ drepDeposit × c ∉ dom (dReps ˢ)) ⊎ (d ≡ 0 × c ∈ dom (dReps ˢ))
+    --   ────────────────────────────────
+    --   ⟦ e , pp , vs ⟧ᶜ ⊢ ⟦ dReps , ccKeys ⟧ᵛ ⇀⦇ regdrep c d an ,GOVCERT⦈
+    --                      ⟦ ❴ c , e +ᵉ' drepActivity ❵ᵐ ∪ᵐˡ dReps , ccKeys ⟧ᵛ
```
lehins:
```suggestion
```

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/GovCert.hs
```diff
@@ -145,67 +144,60 @@ instance
...
       pure $ vState {vsDReps = Map.insert cred drepState vsDReps}
-    ConwayUnRegDRep cred _deposit -> do
-      -- TODO: check against a new PParam `drepDeposit`, once PParams are updated. -- someCheck ?! ConwayDRepIncorrectDeposit deposit
+    -- GOVCERT-deregdrep :
+    --   c ∈ dom (dReps ˢ)
+    --   ────────────────────────────────
+    --   Γ ⊢ ⟦ dReps , ccKeys ⟧ᵛ ⇀⦇ deregdrep c ,GOVCERT⦈
+    --       ⟦ dReps ∣ ❴ c ❵ ᶜ , ccKeys ⟧ᵛ
+    ConwayUnRegDRep cred deposit -> do
+      cgcePParams ^. ppDRepDepositL == deposit ?! ConwayDRepIncorrectDeposit deposit
```
lehins: This is wrong. By the time a DRep decides to unregister, the deposit might have changed.
So we need to check it against the deposit in the DRep state, rather than the ppDRepDepositL pparam.
```suggestion
```

See #3645

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/GovCert.hs
```diff
@@ -145,67 +144,60 @@ instance
...
+      let expiryEpoch = cgceCurrentEpoch + cgcePParams ^. ppDRepActivityL
+          drepState = DRepState expiryEpoch mAnchor
       pure $ vState {vsDReps = Map.insert cred drepState vsDReps}
-    ConwayUnRegDRep cred _deposit -> do
-      -- TODO: check against a new PParam `drepDeposit`, once PParams are updated. -- someCheck ?! ConwayDRepIncorrectDeposit deposit
+    -- GOVCERT-deregdrep :
+    --   c ∈ dom (dReps ˢ)
+    --   ────────────────────────────────
+    --   Γ ⊢ ⟦ dReps , ccKeys ⟧ᵛ ⇀⦇ deregdrep c ,GOVCERT⦈
+    --       ⟦ dReps ∣ ❴ c ❵ ᶜ , ccKeys ⟧ᵛ
```
lehins:
```suggestion
```

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/GovCert.hs
```diff
@@ -145,67 +144,60 @@ instance
...
+    ConwayUnRegDRep cred deposit -> do
+      cgcePParams ^. ppDRepDepositL == deposit ?! ConwayDRepIncorrectDeposit deposit
       Map.member cred vsDReps ?! ConwayDRepNotRegistered cred
       pure $ vState {vsDReps = Map.delete cred vsDReps}
+    -- GOVCERT-ccreghot :
+    --   (kh , nothing) ∉ ccKeys ˢ
+    --   ────────────────────────────────
+    --   Γ ⊢ ⟦ dReps , ccKeys ⟧ᵛ ⇀⦇ ccreghot (inj₁ kh) mkh ,GOVCERT⦈
+    --       ⟦ dReps , singletonᵐ kh mkh ∪ᵐˡ ccKeys ⟧ᵛ
```
lehins:
```suggestion
```

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/GovCert.hs
```diff
@@ -145,67 +144,60 @@ instance
...
     ConwayAuthCommitteeHotKey coldK hotK -> do
       checkColdKeyHasNotResigned coldK vsCommitteeHotKeys
       pure $ vState {vsCommitteeHotKeys = Map.insert coldK (Just hotK) vsCommitteeHotKeys}
     ConwayResignCommitteeColdKey coldK -> do
       checkColdKeyHasNotResigned coldK vsCommitteeHotKeys
       pure $ vState {vsCommitteeHotKeys = Map.insert coldK Nothing vsCommitteeHotKeys}
+    -- Update a DRep Anchor and Expiry by overwriting them.
     ConwayUpdateDRep cred mAnchor -> do
       Map.notMember cred vsDReps ?! ConwayDRepNotRegistered cred
```
lehins: This isn't the line that was touched in this PR, however I spotted that it is wrong, so could you fix it as well:
```suggestion
      Map.member cred vsDReps ?! ConwayDRepNotRegistered cred
```

lehins (changes requested):
This is looking good. DRep deposits are missing, but that can be done as a follow up PR.

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/GovCert.hs
```diff
@@ -145,67 +144,60 @@ instance
...
-        expiryEpoch <-
-          calcDRepExpiryEpoch
-            cgceCurrentSlot
-            (cgcePParams ^. ppDRepActivityL)
-        pure
-          DRepState
-            { drepExpiry = expiryEpoch
-            , drepAnchor = mAnchor
-            }
+      cgcePParams ^. ppDRepDepositL == deposit ?! ConwayDRepIncorrectDeposit deposit
```
lehins: Let's also add the expected value for the deposit to the predicate failure
```suggestion
      cgcePParams ^. ppDRepDepositL == deposit ?! ConwayDRepIncorrectDeposit deposit (cgcePParams ^. ppDRepDepositL)
```

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Cert.hs
aniketd: :100: Thanks!

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Certs.hs
aniketd: My thinking behind doing this was that:
1. We would keep the version of spec that matched the code together
2. When the spec changes, we can diff the version documented in haskell comments and more easily understand what exactly changed in the spec, compared to the implementation
3. Update the spec comments in Haskell along with updating the implementation, this way comments and code in Haskell always correspond to each other.

I will remove it now though, since there is something I'm sure I'm not thinking of. :+1: cc. @Soupstraw

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Certs.hs
aniketd: :100: :bowing_man: Sorry! I had this in mind but forgot to update this.

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Certs.hs
aniketd: :100: Thank you!

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Deleg.hs
aniketd: :bow: This is great! Thank you!

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/GovCert.hs
aniketd: :raised_hands: Thank you!

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/GovCert.hs
aniketd: Great catch! Thanks :100:

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Certs.hs
```diff
@@ -140,17 +164,44 @@ instance
...
+                _ -> Nothing
+            )
+          $ unVotingProcedures
+          $ tx ^. bodyTxL . votingProceduresTxBodyL
+      certState =
+        certStateFresh
+          & certVStateL . vsDRepsL
+            %~ Map.mapWithKey
+              ( \cred state ->
+                  if DRepVoter cred `Set.member` votingDRepCreds
```
lehins, line 200: I don't see how it can type check?
```suggestion
                  if cred `Set.member` votingDRepCreds
```

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Certs.hs
```diff
@@ -140,17 +164,44 @@ instance
...
 
+  -- Update DRep expiry for all DReps that are voting in this transaction
+  let drepActivity = pp ^. ppDRepActivityL
+      votingDRepCreds =
+        Map.keysSet
+          $ Map.mapMaybeWithKey
+            ( \voter _ -> case voter of
+                DRepVoter cred -> Just cred
+                _ -> Nothing
+            )
```
lehins: That's what I get for in PR programming. What we want is a Set of just the DRep creds:
```suggestion
        Map.foldlWithKey'
          $ ( \acc voter _ -> case voter of
                DRepVoter cred -> Set.insert cred acc
                _ -> acc
            )
          Set.empty
```

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Certs.hs
lehins: Now we can get rid of extra indirection
```suggestion
                  if cred `Set.member` votingDRepCreds
```

lehins (commented):
My bad, my previous suggestion was slightly off.

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Certs.hs
aniketd: Awesome!!! :raised_hands: Super thanks @lehins ! :grin:

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Deleg.hs
```diff
@@ -176,8 +171,8 @@ conwayDelegTransition = do
       Just deposit == fmap (UM.fromCompact . UM.rdDeposit) (UM.lookup stakeCred $ UM.RewDepUView dsUnified) ?! IncorrectDepositDELEG deposit
     checkStakeKeyNotAlreadyRegistered stakeCred dsUnified =
       UM.notMember stakeCred (UM.RewDepUView dsUnified) ?! StakeKeyAlreadyRegisteredDELEG stakeCred
```
lehins: I just realized something. Shouldn't it be not yet registered? "Not already registered" does not sound right to me?

Let's remove this confusing naming. This will make it consistent with other predicate failures.
```suggestion
    checkStakeKeyNotRegistered stakeCred dsUnified =
      UM.notMember stakeCred (UM.RewDepUView dsUnified) ?! StakeKeyRegisteredDELEG stakeCredfied
```

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Ledger.hs
```diff
@@ -276,24 +280,25 @@ ledgerTransition = do
...
         let utxoState' = utxoState & utxosGovStateL . cgGovActionsStateL .~ govActionsState'
         pure (utxoState', certState')
       else pure (utxoState, certState)
 
   utxoState'' <-
     trans @(EraRule "UTXOW" era) $
       TRC
-        ( UtxoEnv @era slot pp certState' genCerts
+        -- Pass to UTXOW the unmodified CertState in its Environment, so it can process
+        -- refunds of deposits for deregistering stake credentials. The modified CertState
```
lehins:
```suggestion
        -- refunds of deposits for deregistering stake credentials and DReps. The modified CertState
```

lehins (approved):
I think it's good to go, once other minor comments that I made are addressed.

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Certs.hs
```diff
@@ -140,17 +163,44 @@ instance
...
-  , Environment (EraRule "CERT" era) ~ DelplEnv era
+  , Environment (EraRule "CERT" era) ~ CertEnv era
   , Signal (EraRule "CERT" era) ~ TxCert era
   ) =>
   TransitionRule (ConwayCERTS era)
 conwayCertsTransition = do
-  TRC (env@(DelegsEnv slot txIx pp tx acnt), certState, certificates) <- judgmentContext
+  TRC
+    ( env@(CertsEnv txIx tx acnt pp slot currentEpoch)
+      , certStateFresh
```
Soupstraw: I'd rather use `certState` here and `newCertState` later, because fresh makes it sound like it's a newly generated `CertState`

aniketd: Nice suggestion! I agree! :heart:

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Certs.hs
```diff
@@ -140,17 +163,44 @@ instance
...
+              DRepVoter cred -> Set.insert cred acc
+              _ -> acc
+          )
+          Set.empty
+          $ unVotingProcedures
+          $ tx ^. bodyTxL . votingProceduresTxBodyL
+      certState =
+        certStateFresh
+          & certVStateL . vsDRepsL
+            %~ Map.mapWithKey
```
Soupstraw: This is pretty inefficient since you'd have to map over all of the DReps every time someone submits a transaction, I'd rather build a map with all the credentials that voted in this transaction and then union it with the old DReps map

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/GovCert.hs
```diff
@@ -145,73 +146,59 @@ instance
...
-      let hotCredMap = csCommitteeCreds $ vsCommitteeState vState
-      ((isNothing <$> Map.lookup coldCred hotCredMap) /= Just True)
-        ?! ConwayCommitteeHasResigned coldCred
-      pure $
+    checkColdCredHasNotResigned coldCred csCommitteeCreds =
+      ((isNothing <$> Map.lookup coldCred csCommitteeCreds) /= Just True)
+        ?! ConwayCommitteeHasPreviouslyResigned coldCred
+    checkDepositAgainstPaidDeposit vsDReps cred deposit =
+      forM_ (Map.lookup cred vsDReps) $ \drepState ->
+        let paidDeposit = drepState ^. drepDepositL
```
Soupstraw:
```suggestion
      let drepState = Map.lookup cred vsDReps
          paidDeposit = drepState ^. drepDepositL
```

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Certs.hs
lehins: :100: It would be better to iterate over `votingDRepCreds` Set instead and update individual DReps state that way.

lehins: This logic could even go into the above `foldlWithKey'`. We'd even avoid constructing an extra `votingDRepCreds` set all together

aniketd: :metal:  Great idea! And thanks for the correction! :metal:

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/GovCert.hs
aniketd: `Map.lookup` gives us a `Maybe a`. Do you mean we should use the lens `_Just . _1 . drepDepositL` ?

aniketd:
Since `DRep` expiry update has been moved from `CERTS` to `LEDGER`, this definitely needs another review (at least that part does) @lehins @Soupstraw

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/GovCert.hs
Soupstraw: I'd rather use `do` notation here in that case, my first instinct when reading `forM_` was that it's traversing a list of some sort :grin:

Soupstraw: This should work (and it doesn't skip the test if the lookup fails)
```haskell
let paidDeposit = view drepDepositL <$> Map.lookup cred vsDReps
Just deposit == paidDeposit ?! ConwayDRepIncorrectDeposit deposit paidDeposit

Comment (aniketd) [reply eras/conway/impl/src/Cardano/Ledger/Conway/Rules/GovCert.hs]: Do you mean I should add a `do` at the top of the function?

Comment (Soupstraw) [reply eras/conway/impl/src/Cardano/Ledger/Conway/Rules/GovCert.hs]: I think it'd be even better to throw a different predicate failure if the lookup fails here

File: eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Certs.hs [hunk 34]
```diff
@@ -165,9 +185,8 @@ conwayCertsTransition = do
         validateDelegationRegistered certState' c
       -- It is impossible to have 65535 number of certificates in a
       -- transaction, therefore partial function is justified.
```
Soupstraw:
```suggestion
```

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/GovCert.hs
aniketd: How about I put the membership check predicate in the same function so they are inseparable when this function is called, like:

```haskell
    checkDepositAgainstPaidDeposit vsDReps cred deposit = do
      Map.member cred vsDReps ?! ConwayDRepNotRegistered cred
      forM_ (Map.lookup cred vsDReps) $ \drepState ->
        let paidDeposit = drepState ^. drepDepositL
         in deposit == paidDeposit ?! ConwayDRepIncorrectDeposit deposit paidDeposit
```

Soupstraw: ignore the `do` comment :)

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Certs.hs
aniketd: Nice catch!

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/GovCert.hs
Soupstraw: Yeah, I think that's a good idea

Soupstraw: but wait, why not just case match on the result of lookup?

aniketd: Yes, that's a better idea for sure :100:!

Soupstraw: getting a `Just x` from the lookup is a way better proof that the value is in the map than the `Bool` you get from `Map.member`

eras/conway/impl/CHANGELOG.md
```diff
@@ -2,6 +2,11 @@
 
 ## 1.7.0.0
 
+* Make `DELEG`, `POOL` and `GOVCERT` conform to spec-v0.8 #3628
+  * Add `CertEnv` and `CertsEnv` to pass `EpochNo` down from `LEDGER` to sub-rules
+  * Add `drepDeposit` to `DRepState` to track deposits paid by `DRep`s
+  * Update `DRep` expiry in `LEDGER` for all `DRep`s who are voting in current `Tx`
+  * Remove `calcDRepExpiryEpoch` since we only calculate it once in `LEDGER`
```
lehins: This was an internal (not exported) function, so no need to add it to the changelog
```suggestion
```

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/GovCert.hs
```diff
@@ -145,73 +146,59 @@ instance
...
-    ConwayUnRegDRep cred _deposit -> do
-      -- TODO: check against a new PParam `drepDeposit`, once PParams are updated. -- someCheck ?! ConwayDRepIncorrectDeposit deposit
+      deposit == ppDRepDeposit ?! ConwayDRepIncorrectDeposit deposit ppDRepDeposit
+      pure
+        vState
+          { vsDReps =
+              Map.insert cred (DRepState (cgceCurrentEpoch + ppDRepActivity) mAnchor ppDRepDeposit) vsDReps
+          }
+    ConwayUnRegDRep cred deposit -> do
       Map.member cred vsDReps ?! ConwayDRepNotRegistered cred
```
lehins:
```suggestion
      let mDRepState = Map.lookup cred vsDReps
      isJust mDRepState ?! ConwayDRepNotRegistered cred
```

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/GovCert.hs
```diff
@@ -145,73 +146,59 @@ instance
...
-      pure $ vState {vsDReps = Map.delete cred vsDReps}
+      checkDepositAgainstPaidDeposit vsDReps cred deposit
```
lehins:
```suggestion
      checkDepositAgainstPaidDeposit mDRepState cred deposit
```

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/GovCert.hs
```diff
@@ -145,73 +146,59 @@ instance
...
-    overwriteCommitteeHotCred vState coldCred hotCred = do
-      let hotCredMap = csCommitteeCreds $ vsCommitteeState vState
-      ((isNothing <$> Map.lookup coldCred hotCredMap) /= Just True)
-        ?! ConwayCommitteeHasResigned coldCred
-      pure $
+    checkColdCredHasNotResigned coldCred csCommitteeCreds =
+      ((isNothing <$> Map.lookup coldCred csCommitteeCreds) /= Just True)
+        ?! ConwayCommitteeHasPreviouslyResigned coldCred
+    checkDepositAgainstPaidDeposit vsDReps cred deposit =
+      forM_ (Map.lookup cred vsDReps) $ \drepState ->
```
lehins: Minor optimization. This way we can avoid double lookup of a DRep in the state.
```suggestion
    checkDepositAgainstPaidDeposit mDRepState cred deposit =
      forM_ mDRepState $ \drepState ->
```

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/GovCert.hs
lehins: Ignore this. I could not delete this and other comments because you updated the code. Your version is even better than what I suggested :wink:

eras/conway/impl/CHANGELOG.md
aniketd: Understood!

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/GovCert.hs
aniketd: :raised_hands: :metal:

aniketd: @Soupstraw is the star! :grin:

aniketd:
@Soupstraw need you approval here :crown:

Soupstraw (approved):
LGTM

PR #3630, opened 2023-08-09
Implemented DRep refreshing

### LINKED ISSUE #3621: Make it possible to refresh the DRep timeout
input-output-hk/formal-ledger-specifications#127 adds the possibility of submitting a ~`RegDRep`~ `UpdateDRepTxCert` certificate for an already existing DRep in order to reset the DRep expiration timer.

Soupstraw:
# Description

resolves #3621

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/eras/conway/impl/CHANGELOG.md
+++ b/eras/conway/impl/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## 1.7.0.0
 
+* Add `ConwayGovCertEnv`
+* Change the environment of `GOVCERT` to `ConwayGovCertEnv`
 * Add `ConwayEraGov` with `constitutionGovStateL`
 * Add `PrevGovActionId` and `GovActionPurpose`
 * Add optional `PrevGovActionId` to `ParameterChange`, `HardForkInitiation`,
--- a/eras/conway/impl/src/Cardano/Ledger/Conway/Rules/GovCert.hs
+++ b/eras/conway/impl/src/Cardano/Ledger/Conway/Rules/GovCert.hs
@@ -4,6 +4,7 @@
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE NamedFieldPuns #-}
+{-# LANGUAGE RecordWildCards #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE TypeApplications #-}
 {-# LANGUAGE TypeFamilies #-}
@@ -15,15 +16,19 @@ module Cardano.Ledger.Conway.Rules.GovCert (
   ConwayGOVCERT,
   ConwayGovCertEvent (..),
   ConwayGovCertPredFailure (..),
+  ConwayGovCertEnv (..),
 )
 where
 
 import Cardano.Ledger.BaseTypes (
+  EpochNo,
   ShelleyBase,
+  SlotNo,
+  epochInfoPure,
  )
 import Cardano.Ledger.Binary (DecCBOR (..), EncCBOR (..), encodeListLen)
 import Cardano.Ledger.Binary.Coders
-import Cardano.Ledger.CertState (DRepState (..), VState (..))
+import Cardano.Ledger.CertState (DRepState (..), VState (..), vsDRepsL)
 import Cardano.Ledger.Coin (Coin)
 import Cardano.Ledger.Conway.Core (ConwayEraPParams, ppDRepActivityL)
 import Cardano.Ledger.Conway.Era (ConwayGOVCERT)
@@ -31,19 +36,24 @@ import Cardano.Ledger.Conway.TxCert (ConwayGovCert (..))
 import Cardano.Ledger.Core (Era (EraCrypto), EraRule, PParams)
 import Cardano.Ledger.Credential (Credential)
 import Cardano.Ledger.Crypto (Crypto)
+import Cardano.Ledger.DRepDistr (drepAnchorL, drepExpiryL)
 import Cardano.Ledger.Keys (KeyRole (ColdCommitteeRole, DRepRole))
+import Cardano.Ledger.Slot (epochInfoEpoch)
 import Control.DeepSeq (NFData)
+import Control.Monad.Trans.Reader (asks)
 import Control.State.Transition.Extended (
   BaseM,
   Environment,
   Event,
   PredicateFailure,
+  Rule,
   STS,
   Signal,
   State,
   TRC (TRC),
   TransitionRule,
   judgmentContext,
+  liftSTS,
   transitionRules,
   (?!),
  )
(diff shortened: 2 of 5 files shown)


eras/conway/impl/src/Cardano/Ledger/Conway/Rules/GovCert.hs
```diff
@@ -161,7 +174,16 @@ conwayGovCertTransition = do
...
-      pure vState -- TODO: update anchor
+      epochNo <- liftSTS $ do
+        ei <- asks epochInfoPure
+        epochInfoEpoch ei cgceCurrentSlot
+      let newTimeout = epochNo + cgcePParams ^. ppDRepActivityL
+      pure $
+        vState
+          & vsDRepsL
+            %~ Map.update
+              (\x -> Just $ x & drepExpiryL .~ newTimeout)
```
lehins: Let's also update the anchor while we at it.
```suggestion
              (\x -> Just $ x & drepExpiryL .~ newTimeout
                              & drepAnchorL .~ mAnchor)
```

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/GovCert.hs
```diff
@@ -161,7 +174,16 @@ conwayGovCertTransition = do
       pure $ vState {vsCommitteeHotKeys = Map.insert coldK Nothing vsCommitteeHotKeys}
     ConwayUpdateDRep cred _mAnchor -> do
```
lehins:
```suggestion
    ConwayUpdateDRep cred mAnchor -> do
```

lehins (approved):
Looks great! Thank you!

teodanciu (approved):
Looks good to me

eras/conway/impl/CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## 1.7.0.0
 
+* Add `ConwayGovCertEnv`
```
aniketd, line 5: I also added a `GovCertEnv` in my local work, but named it without the `Conway` prefix. I like it better with the prefix, though I thought it was inconsistent with the other `Env`s we have. Nice! :raised_hands:

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/GovCert.hs
```diff
@@ -159,9 +172,23 @@ conwayGovCertTransition = do
...
+        vState
+          & vsDRepsL
+            %~ Map.update
+              ( \x ->
+                  Just $
+                    x
+                      & drepExpiryL .~ newTimeout
+                      & drepAnchorL .~ mAnchor
+              )
+              cred
```
aniketd: It looks like the `drepExpiry` is supposed to be the `Epoch` in which the DRep should expire. But during registration (a few lines above) the `drepExpiry` field is considererd to be the "duration" of epochs that the drep should not expire for.

I may be wrong, but I think that the drep expiration calculation with respect to the current epoch should happen during registration as well? And here the `drepActivity` duration should be added yet again to the current EpochNo here?

aniketd (changes requested):
@Soupstraw I have some questions :smile_cat:

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/GovCert.hs
aniketd:
```haskell
    -- Update a DRep Anchor and Expiry by overwriting them.
    ConwayUpdateDRep cred mAnchor -> do
      Map.notMember cred vsDReps ?! ConwayDRepNotRegistered cred
      expiryEpoch <- calcDRepExpiryEpoch gceSlotNo $ gcePP ^. ppDRepActivityL
      pure vState {vsDReps = Map.adjust (updateAnchor expiryEpoch mAnchor) cred vsDReps }
where
    updateAnchor newMAnchor expiryEpoch oldDRepState =
      oldDRepState { drepExpiry = expiryEpoch, drepAnchor = newMAnchor }

[...]

calcDRepExpiryEpoch slot drepActivity = do
  ei <- liftSTS $ asks epochInfoPure
  currentE <- liftSTS $ epochInfoEpoch ei slot
  pure $ currentE + drepActivity
```

Soupstraw: Yeah, you're right. According to the spec it should store the epoch after which the DRep is considered inactive. So this PR does it right, but the registration certificate should be changed. Nice catch!

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/GovCert.hs
```diff
@@ -159,9 +194,23 @@ conwayGovCertTransition = do
...
+        vState
+          & vsDRepsL
+            %~ Map.update
+              ( \x ->
+                  Just $
+                    x
+                      & drepExpiryL .~ expiryEpoch
+                      & drepAnchorL .~ mAnchor
+              )
+              cred
```
lehins: Just for clarity, I think it would be good to extract this update function into a separate let binding
```suggestion
      let updateDRepState drepState =
            drepState & drepExpiryL .~ expiryEpoch
                      & drepAnchorL .~ mAnchor
      pure $ vState & vsDRepsL %~ Map.update (Just . updateDRepState) cred
```

Soupstraw: wow, fourmolu absolutely massacred this

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/GovCert.hs
```diff
@@ -159,9 +194,17 @@ conwayGovCertTransition = do
...
       checkColdKeyHasNotResigned coldK vsCommitteeHotKeys
       pure $ vState {vsCommitteeHotKeys = Map.insert coldK Nothing vsCommitteeHotKeys}
-    ConwayUpdateDRep cred _mAnchor -> do
+    ConwayUpdateDRep cred mAnchor -> do
       Map.notMember cred vsDReps ?! ConwayDRepNotRegistered cred
-      pure vState -- TODO: update anchor
+      expiryEpoch <-
+        calcDRepExpiryEpoch
+          cgceCurrentSlot
+          (cgcePParams ^. ppDRepActivityL)
```
aniketd, line 202: This has been repeated once here after seeing it above. May be we can turn it into a top-level function, since it needs to be used in CERTS as well later, if I remember correctly from the spec.

aniketd: Oh wait, it is already made into a function! Sorry!

aniketd (approved):
Looks great! Awesome job! :heart:

PR #3633, opened 2023-08-10
Prevent state changes on phase2 failure

lehins:
# Description

This PR prevents phase1 validation and state changes when `isValid = False` This is necessary for correctness.

We should only validate the UTXO rules when transaction is marked as invalid, because we should only collect the collateral when phase2 is expected to fail.
    
There is no requirement for `CERTS` or `GOV` rule run validation if we only collect the collateral.

Also renames:
* `cgGov[L]` -> `cgGovActionsState[L]`
* `cgRatify[L]` -> `cgRatifyState[L]`
* 
# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Ledger.hs
+++ b/eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Ledger.hs
@@ -39,7 +39,7 @@ import Cardano.Ledger.Conway.Governance (
   ConwayGovState (..),
   GovActionsState,
   GovProcedures (..),
-  cgGovL,
+  cgGovActionsStateL,
  )
 import Cardano.Ledger.Conway.Rules.Certs (ConwayCertsEvent, ConwayCertsPredFailure)
 import Cardano.Ledger.Conway.Rules.Gov (ConwayGovPredFailure, GovEnv (..))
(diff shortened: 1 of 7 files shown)


eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Ledger.hs
teodanciu, line 1: Does this mean that it's wrong in the current version too? 
The Ledger rule in Alonzo (which is used in Babbage) is calling UTXOW after the  `IsValid` check

Soupstraw: Here we're still calling UTXOW after the check. I think the issue is with the other rules (GOV and CERTS)?

teodanciu: Sorry, I misread where the `else` finishes... :neutral_face: . Nevermind

Soupstraw (approved):
LGTM

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Ledger.hs
```diff
@@ -244,58 +246,58 @@ ledgerTransition ::
...
+        certState' <-
+          trans @(EraRule "CERTS" era) $
+            TRC
+              ( DelegsEnv slot txIx pp tx account
+              , certState
+              , StrictSeq.fromStrict $ txBody ^. certsTxBodyL
+              )
+        let wdrlAddrs = Map.keysSet . unWithdrawals $ tx ^. bodyTxL . withdrawalsTxBodyL
+        let wdrlCreds = Set.map getRwdCred wdrlAddrs
+        let dUnified = dsUnified $ certDState certState'
```
aniketd, line 266:
```suggestion
        let wdrlAddrs = Map.keysSet . unWithdrawals $ tx ^. bodyTxL . withdrawalsTxBodyL
            wdrlCreds = Set.map getRwdCred wdrlAddrs
            dUnified = dsUnified $ certDState certState'
```

Please ignore if unnecessary.

lehins: I like this change, but I don't want to rerun the CI again :grin: 

I'll apply it some other PR :wink:

PR #3634, opened 2023-08-10
Disable serialization of protocolVersion param update

### LINKED ISSUE #3632: Disable serialization of protocolVersion param update
We will need to eventually remove protocolVersion from PParamsUpdate, see #3539 

However, an easy approach to at least make it safe is to prevent [de]serialization of this field. This will also allow us to provide correct CDDL, which is important for downstream tooling

This will require a change to Arbitrary instance (always generate Nothing for that field), to ensure roundtripping still works correctly

Soupstraw:
# Description

closes #3632

# Checklist

- [x] Self-reviewed the diff


lehins (approved):
Perfect! Thanks!

PR #3635, opened 2023-08-10
Make snapshots of GovActionsState

### LINKED ISSUE #3590: Snapshot parts of ConwayGovState

Soupstraw:
# Description

This PR adds governance action state snapshotting and fixes some issues with governance action deposits. `RatifyState` is no longer stored inside `ConwayGovState` and is replaced with `EnactState`.

closes #3590

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/eras/conway/impl/src/Cardano/Ledger/Conway/Governance.hs
+++ b/eras/conway/impl/src/Cardano/Ledger/Conway/Governance.hs
@@ -42,6 +42,7 @@ module Cardano.Ledger.Conway.Governance (
   ConwayEraGov (..),
   -- Lenses
   cgGovActionsStateL,
+  cgEnactStateL,
   cgRatifyStateL,
   ensConstitutionL,
   rsEnactStateL,
@@ -49,6 +50,10 @@ module Cardano.Ledger.Conway.Governance (
   prevPParamsConwayGovStateL,
   constitutionScriptL,
   constitutionAnchorL,
+  curGovActionsStateL,
+  prevGovActionsStateL,
+  prevDRepsStateL,
+  prevCommitteeStateL,
 ) where
 
 import Cardano.Ledger.Address (RewardAcnt)
@@ -67,6 +72,7 @@ import Cardano.Ledger.Binary.Coders (
   (!>),
   (<!),
  )
+import Cardano.Ledger.CertState (CommitteeState, DRepState)
 import Cardano.Ledger.Coin (Coin (..))
 import Cardano.Ledger.Conway.Era (ConwayEra)
 import Cardano.Ledger.Conway.Governance.Procedures (
@@ -93,9 +99,12 @@ import Cardano.Ledger.Credential (Credential (..))
 import Cardano.Ledger.Crypto (Crypto)
 import Cardano.Ledger.Keys (KeyHash (..), KeyRole (..))
 import Cardano.Ledger.Shelley.Governance
+import Cardano.Ledger.TreeDiff (ToExpr)
 import Control.DeepSeq (NFData (..))
 import Data.Aeson (KeyValue, ToJSON (..), object, pairs, (.=))
 import Data.Default.Class (Default (..))
+import Data.Foldable (Foldable (..))
+import Data.Functor.Identity (Identity)
 import Data.Map.Strict (Map)
 import qualified Data.Map.Strict as Map
 import Data.Sequence.Strict (StrictSeq)
@@ -115,6 +124,8 @@ data GovActionState era = GovActionState
   , gasExpiresAfter :: !EpochNo
   }
   deriving (Generic)
+
+instance EraPParams era => ToExpr (GovActionState era)
 
 instance EraPParams era => ToJSON (GovActionState era) where
   toJSON = object . toGovActionStatePairs
@@ -170,32 +181,88 @@ instance EraPParams era => EncCBOR (GovActionState era) where
         !> To gasProposedIn
         !> To gasExpiresAfter
 
-newtype GovActionsState era = GovActionsState
-  { unGovActionsState :: Map (GovActionId (EraCrypto era)) (GovActionState era)
+data GovActionsState era = GovActionsState
+  { curGovActionsState :: !(Map (GovActionId (EraCrypto era)) (GovActionState era))
+  , prevGovActionsState :: !(Map (GovActionId (EraCrypto era)) (GovActionState era))
+  , prevDRepsState :: !(Map (Credential 'DRepRole (EraCrypto era)) (DRepState (EraCrypto era)))
+  , prevCommitteeState :: !(CommitteeState era)
   }
-  deriving (Generic, NFData, Semigroup, Monoid)
+  deriving (Generic)
+
+instance EraPParams era => ToExpr (GovActionsState era)
 
 insertGovActionsState ::
   GovActionState era ->
   GovActionsState era ->
   GovActionsState era
-insertGovActionsState v@GovActionState {gasId} (GovActionsState m) =
-  GovActionsState $ Map.insert gasId v m
+insertGovActionsState v@GovActionState {gasId} =
+  curGovActionsStateL %~ Map.insert gasId v
+
+curGovActionsStateL ::
+  Lens'
+    (GovActionsState era)
+    (Map (GovActionId (EraCrypto era)) (GovActionState era))
+curGovActionsStateL = lens curGovActionsState (\x y -> x {curGovActionsState = y})
+
+prevGovActionsStateL ::
+  Lens'
+    (GovActionsState era)
+    (Map (GovActionId (EraCrypto era)) (GovActionState era))
+prevGovActionsStateL = lens prevGovActionsState (\x y -> x {prevGovActionsState = y})
+
+prevDRepsStateL ::
+  Lens'
+    (GovActionsState era)
+    (Map (Credential 'DRepRole (EraCrypto era)) (DRepState (EraCrypto era)))
+prevDRepsStateL = lens prevDRepsState (\x y -> x {prevDRepsState = y})
+
+prevCommitteeStateL ::
+  Lens'
+    (GovActionsState era)
+    (CommitteeState era)
+prevCommitteeStateL = lens prevCommitteeState (\x y -> x {prevCommitteeState = y})
 
 deriving instance EraPParams era => Eq (GovActionsState era)
 
 deriving instance EraPParams era => Show (GovActionsState era)
 
-deriving instance EraPParams era => ToJSON (GovActionsState era)
+toGovActionsStatePairs :: (KeyValue a, EraPParams era) => GovActionsState era -> [a]
+toGovActionsStatePairs gas@(GovActionsState _ _ _ _) =
+  let GovActionsState {..} = gas
+   in [ "curGovActionsState" .= curGovActionsState
+      , "prevGovActionsState" .= prevGovActionsState
+      , "prevDRepsState" .= prevDRepsState
+      , "prevCommitteeState" .= prevCommitteeState
+      ]
+
+instance EraPParams era => ToJSON (GovActionsState era) where
+  toJSON = object . toGovActionsStatePairs
+  toEncoding = pairs . mconcat . toGovActionsStatePairs
+
+instance EraPParams era => NFData (GovActionsState era)
 
 instance EraPParams era => NoThunks (GovActionsState era)
 
 instance Default (GovActionsState era) where
-  def = GovActionsState mempty
+  def = GovActionsState mempty mempty mempty def
 
-deriving instance EraPParams era => EncCBOR (GovActionsState era)
+instance EraPParams era => EncCBOR (GovActionsState era) where
+  encCBOR GovActionsState {..} =
+    encode $
+      Rec GovActionsState
+        !> To curGovActionsState
+        !> To prevGovActionsState
+        !> To prevDRepsState
+        !> To prevCommitteeState
 
-deriving instance EraPParams era => DecCBOR (GovActionsState era)
+instance EraPParams era => DecCBOR (GovActionsState era) where
+  decCBOR =
+    decode $
+      RecD GovActionsState
+        <! From
+        <! From
+        <! From
+        <! From
 
 instance EraPParams era => ToCBOR (GovActionsState era) where
   toCBOR = toEraCBOR @era
@@ -224,6 +291,8 @@ ensCurPParamsL = lens ensPParams (\es x -> es {ensPParams = x})
 
 ensPrevPParamsL :: Lens' (EnactState era) (PParams era)
 ensPrevPParamsL = lens ensPrevPParams (\es x -> es {ensPrevPParams = x})
+
+instance ToExpr (PParamsHKD Identity era) => ToExpr (EnactState era)
 
 instance EraPParams era => ToJSON (EnactState era) where
   toJSON = object . toEnactStatePairs
@@ -292,7 +361,6 @@ instance EraPParams era => NoThunks (EnactState era)
 
 data RatifyState era = RatifyState
   { rsEnactState :: !(EnactState era)
-  , rsFuture :: !(StrictSeq (GovActionState era))
   , rsRemoved :: !(StrictSeq (GovActionState era))
   }
   deriving (Generic)
@@ -304,6 +372,8 @@ deriving instance EraPParams era => Show (RatifyState era)
 rsEnactStateL :: Lens' (RatifyState era) (EnactState era)
 rsEnactStateL = lens rsEnactState (\x y -> x {rsEnactState = y})
 
+instance EraPParams era => ToExpr (RatifyState era)
+
 instance EraPParams era => Default (RatifyState era)
 
 instance EraPParams era => DecCBOR (RatifyState era) where
@@ -312,14 +382,12 @@ instance EraPParams era => DecCBOR (RatifyState era) where
       RecD RatifyState
         <! From
         <! From
-        <! From
 
 instance EraPParams era => EncCBOR (RatifyState era) where
   encCBOR RatifyState {..} =
     encode $
       Rec RatifyState
         !> To rsEnactState
-        !> To rsFuture
         !> To rsRemoved
 
 instance EraPParams era => ToCBOR (RatifyState era) where
@@ -337,30 +405,36 @@ instance EraPParams era => ToJSON (RatifyState era) where
   toEncoding = pairs . mconcat . toRatifyStatePairs
 
 toRatifyStatePairs :: (KeyValue a, EraPParams era) => RatifyState era -> [a]
-toRatifyStatePairs cg@(RatifyState _ _ _) =
+toRatifyStatePairs cg@(RatifyState _ _) =
   let RatifyState {..} = cg
    in [ "enactState" .= rsEnactState
-      , "future" .= rsFuture
       , "removed" .= rsRemoved
       ]
 
 data ConwayGovState era = ConwayGovState
   { cgGovActionsState :: !(GovActionsState era)
-  , cgRatifyState :: !(RatifyState era)
+  , cgEnactState :: !(EnactState era)
   }
   deriving (Generic, Eq, Show)
 
 cgGovActionsStateL :: Lens' (ConwayGovState era) (GovActionsState era)
 cgGovActionsStateL = lens cgGovActionsState (\x y -> x {cgGovActionsState = y})
 
+cgEnactStateL :: Lens' (ConwayGovState era) (EnactState era)
+cgEnactStateL = lens cgEnactState (\x y -> x {cgEnactState = y})
+
+{-# DEPRECATED cgRatifyStateL "Use cgEnactStateL instead" #-}
 cgRatifyStateL :: Lens' (ConwayGovState era) (RatifyState era)
-cgRatifyStateL = lens cgRatifyState (\x y -> x {cgRatifyState = y})
+cgRatifyStateL =
+  lens
+    (\ConwayGovState {..} -> RatifyState cgEnactState mempty)
+    (\x RatifyState {..} -> x & cgEnactStateL .~ rsEnactState)
 
 curPParamsConwayGovStateL :: Lens' (ConwayGovState era) (PParams era)
-curPParamsConwayGovStateL = cgRatifyStateL . rsEnactStateL . ensCurPParamsL
+curPParamsConwayGovStateL = cgEnactStateL . ensCurPParamsL
 
 prevPParamsConwayGovStateL :: Lens' (ConwayGovState era) (PParams era)
-prevPParamsConwayGovStateL = cgRatifyStateL . rsEnactStateL . ensPrevPParamsL
+prevPParamsConwayGovStateL = cgEnactStateL . ensPrevPParamsL
 
 instance EraPParams era => DecCBOR (ConwayGovState era) where
   decCBOR =
@@ -374,7 +448,7 @@ instance EraPParams era => EncCBOR (ConwayGovState era) where
     encode $
       Rec ConwayGovState
         !> To cgGovActionsState
-        !> To cgRatifyState
+        !> To cgEnactState
 
 instance EraPParams era => ToCBOR (ConwayGovState era) where
   toCBOR = toEraCBOR @era
@@ -392,24 +466,29 @@ instance EraPParams era => ToJSON (ConwayGovState era) where
   toJSON = object . toConwayGovPairs
   toEncoding = pairs . mconcat . toConwayGovPairs
 
+instance EraPParams era => ToExpr (ConwayGovState era)
+
 toConwayGovPairs :: (KeyValue a, EraPParams era) => ConwayGovState era -> [a]
 toConwayGovPairs cg@(ConwayGovState _ _) =
   let ConwayGovState {..} = cg
    in [ "gov" .= cgGovActionsState
-      , "ratify" .= cgRatifyState
+      , "ratify" .= cgEnactState
       ]
 
 instance EraPParams (ConwayEra c) => EraGov (ConwayEra c) where
   type GovState (ConwayEra c) = ConwayGovState (ConwayEra c)
 
-  getConstitution g = Just $ g ^. cgRatifyStateL . rsEnactStateL . ensConstitutionL
+  getConstitution g = Just $ g ^. cgEnactStateL . ensConstitutionL
 
   curPParamsGovStateL = curPParamsConwayGovStateL
 
   prevPParamsGovStateL = prevPParamsConwayGovStateL
 
+  obligationGovState st =
+    foldMap' gasDeposit (st ^. cgGovActionsStateL . curGovActionsStateL)
+
 class EraGov era => ConwayEraGov era where
   constitutionGovStateL :: Lens' (GovState era) (Constitution era)
 
 instance Crypto c => ConwayEraGov (ConwayEra c) where
-  constitutionGovStateL = cgRatifyStateL . rsEnactStateL . ensConstitutionL
+  constitutionGovStateL = cgEnactStateL . ensConstitutionL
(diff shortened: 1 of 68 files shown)


eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Epoch.hs
```diff
@@ -255,7 +256,7 @@ epochTransition = do
         , reCurrentEpoch = eNo
         }
     govStateToSeq = Seq.fromList . Map.toList
-    ratSig = RatifySignal . govStateToSeq . unGovActionsState $ cgGovActionsState govSt
+    ratSig = RatifySignal . govStateToSeq . curGovActionsState $ cgGovActionsState govSt
```
lehins: When we ratify we should be using the previous state.
```suggestion
    ratSig = RatifySignal . govStateToSeq . prevGovActionsState $ cgGovActionsState govSt
```

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Epoch.hs
```diff
@@ -265,9 +266,16 @@ epochTransition = do
...
-      newGov = GovActionsState . Map.fromList . toList $ rsFuture
+      esGovernanceL :: Lens' (EpochState era) (GovState era)
       esGovernanceL = esLStateL . lsUTxOStateL . utxosGovStateL
+      -- TODO can we be more efficient with the list conversions?
+      newGov =
+        GovActionsState
+          { curGovActionsState = Map.fromList . toList $ rsFuture
+          , prevGovActionsState =
+              es
+                ^. esGovernanceL . cgGovActionsStateL . curGovActionsStateL
```
lehins: Slight reduction in number of lines :wink: 
```suggestion
              es ^. esGovernanceL . cgGovActionsStateL . curGovActionsStateL
```

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Examples/ConwayFeatures.hs
lehins, line 1: I think these tests will have to be adjusted, since proposals can't be ratified right after they've been proposed

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Examples/ConwayFeatures.hs
```diff
@@ -301,6 +302,12 @@ vote pf govActionId =
     , otherWitsFields = []
     }
 
+assertEqualEdiff :: (Eq a, ToExpr a) => String -> a -> a -> Assertion
+assertEqualEdiff msg x y
+  | x == y = pure ()
+  | otherwise = do
+      error $ msg <> ":\n" <> show (ansiWlEditExpr $ ediff x y)
```
lehins: We already have a function that does the same thing:
https://github.com/input-output-hk/cardano-ledger/blob/12161e72f8a718b012ec0b811f82431427f1fc94/libs/cardano-ledger-binary/testlib/Test/Cardano/Ledger/Binary/TreeDiff.hs#L140-L145

lehins (commented):
Looking good :+1:

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Epoch.hs
Soupstraw: I'm a bit confused by this. Wouldn't this change delay the ratification by one epoch?

lehins: Yes, that is exactly what we want, because we need the full epoch to compute DRep stake distribution.

lehins: That is also why we need to do this snapshotting in the first place :slightly_smiling_face:

eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState/Types.hs
```diff
@@ -742,3 +744,18 @@ freshDRepPulser n es =
...
+  Bool
+potEqualsObligation certState utxoSt = obligation == pot
+  where
+    obligation = totalObligation certState utxoSt
+    pot = utxoSt ^. utxosDepositedL
+
+totalObligation :: EraGov era => CertState era -> UTxOState era -> Coin
+totalObligation certState utxoSt =
+  obligationCertState certState
+    <> obligationGovState (utxoSt ^. utxosGovStateL)
```
aniketd: Would passing `utxoSt ^. utxosGovStateL` to this function read better, especially when considering that the `pot` is contained in the `utxoSt` and we are trying to assert that the `pot` equals `obligation`?

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Examples/ConwayFeatures.hs
```diff
@@ -368,36 +378,58 @@ testGov pf = do
...
             ]
         )
-    eitherEpochState1 = runEPOCH (EPOCH pf) epochState0 (EpochNo 2) poolDistr
-    epochState1 = expectRight "Error running runEPOCH: " eitherEpochState1
-    ledgerState2 = epochState1 ^. esLStateL
-    constitution = (ensConstitution . rsEnactState) (epochState1 ^. esLStateL . lsUTxOStateL . utxosGovStateL . cgRatifyStateL)
-  assertEqual "constitution after enactment" constitution (proposedConstitution @era)
-
+    -- Wait two epochs
+    !eitherEpochState1 = runEPOCH (EPOCH pf) epochState0 (EpochNo 2) poolDistr
```
teodanciu: Why all the bangs?

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Examples/ConwayFeatures.hs
```diff
@@ -12,6 +13,7 @@
 {-# LANGUAGE TypeFamilies #-}
 {-# LANGUAGE TypeOperators #-}
 {-# LANGUAGE UndecidableInstances #-}
+{-# OPTIONS_GHC -Wno-name-shadowing #-}
```
aniketd: We should refrain from this I think. It would be hard to debug these tests.

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Epoch.hs
```diff
@@ -238,12 +235,15 @@ epochTransition = do
...
           ss'
           nm
 
   let
-    utxoSt''' = utxoSt' {utxosDeposited = obligationCertState adjustedCertState}
+    utxoSt''' =
+      utxoSt'
+        { utxosDeposited = totalObligation adjustedCertState utxoSt'
+        }
     acnt'' = acnt' {asReserves = asReserves acnt'}
```
aniketd: I doubt whether this statement does anything meaningful and is a no-op.

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Epoch.hs
```diff
@@ -256,26 +256,47 @@ epochTransition = do
         , reCurrentEpoch = eNo
         }
     govStateToSeq = Seq.fromList . Map.toList
```
aniketd: Going via a `List` could be avoided by folding over.

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Epoch.hs
```diff
@@ -256,26 +256,47 @@ epochTransition = do
...
+      es ^. esGovernanceL . cgGovActionsStateL . curGovActionsStateL
+    newGovActionsState =
+      foldr'
+        (\(k, _) m -> Map.delete k m)
+        oldGovActionsState
+        rsRemoved
+    newGov =
+      GovActionsState
+        { curGovActionsState = newGovActionsState
+        , prevGovActionsState = newGovActionsState
```
aniketd, line 297: Should this be `prevGovActionsState = oldGovActionsState` instead? I'm not sure and don't have the full picture yet.

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Epoch.hs
```diff
@@ -256,26 +256,47 @@ epochTransition = do
...
-      newGov = GovActionsState . Map.fromList . toList $ rsFuture
-      esGovernanceL = esLStateL . lsUTxOStateL . utxosGovStateL
-      esDonationL :: Lens' (EpochState era) Coin
-      esDonationL = esLStateL . lsUTxOStateL . utxosDonationL
-      donations = es'' ^. esDonationL
+    ratSig = RatifySignal . govStateToSeq . prevGovActionsState $ cgGovActionsState govSt
+  RatifyState {rsRemoved, rsEnactState} <-
+    trans @(EraRule "RATIFY" era) $
+      TRC
+        ( ratEnv
```
lehins, line 265: I suspect that we need to supply `reCurrentEpoch = eNo - 1` to ratify instead of a current epochNo, otherwise all DReps and GovActions will expire one epoch sooner.

Moreover, once you rebase on master, you'll see that RatifyEnv now also includes DRep state, which means we need to snapshot that as well, otherwise DReps can deregister during the epoch. Same applies to ComitteeState, we'll have to snapshot that, otherwise votes for resigned CC members will not count.

We need to discuss it more with @WhatisRT

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Examples/ConwayFeatures.hs
Soupstraw: It makes it easier to debug when the test fails. This makes the let statements evaluate in sequence so I can just add `traceM` between them.

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Epoch.hs
Soupstraw: This is going to be changed soon anyways. We'll have to use sequences for governance states so we won't have to convert from a map.

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Epoch.hs
Soupstraw: The idea is that we set both of them to the value we got at the end of the current epoch and only the current state is going to be changed in the next epoch while the previous one stays the same so we can add up the votes.

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Examples/ConwayFeatures.hs
Soupstraw: I don't think it's as bad if we evaluate all the `let` statements strictly, then it sort of reads like python or C code :smiley: . It can be hard to give each variable a unique name and when something gets added in the middle we have to re-enumerate everything.

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Epoch.hs
Soupstraw: There are some changes in deposits after running the `POOLREAP` rule I think.

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Epoch.hs
teodanciu: Would you mind adding this as a comment?  Might be useful to remember  in the future, since it does seem a little peculiar if you don't know what's going on.

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Epoch.hs
Soupstraw: oh, right, I looked at the lines before it, you're right about `acnt''`

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Examples/ConwayFeatures.hs
lehins: @Soupstraw This is a totally invalid assumption. GHC will still evaluate in an order it prefers. Only monadic sequencing can enforce a specific order of evaluation, otherwise it is pointless.

The only thing bang patterns here do is say: regardless if the binding is used or not, evaluate it.

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Examples/ConwayFeatures.hs
lehins: Sorry, @Soupstraw but we only disable compiler warning when we really have to. So, I agree with Aniket. Could you please remove this GHC flag.

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Epoch.hs
lehins: Yes, I 100% agree with Teodora. Adding a comment would be super useful. It took me a while as well to figure out why both of them have to be set to the same value.

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Examples/ConwayFeatures.hs
aniketd: > I don't think it's as bad if we evaluate all the let statements strictly, then it sort of reads like python or C code 😃 . It can be hard to give each variable a unique name and when something gets added in the middle we have to re-enumerate everything.

Well, I would agree with you, but let me explain what I thought about this: all conventional haskell programmers are used to making all sorts of assumptions in their heads due to the default semantics of the language. When one changes the default semantics of the language, the indicator that some part of the code is to be treated specially is best left right next to the site where the semantics is changed, so other programmers can notice it. Here we are proposing to suppress name-shadowing warnings file-wide, and keeping the indicator at the top of the file. Any engineer reading the code 3 months from now, including you, has a high (read total) probability of being confused by this. So if and when we are sure that something will always be a problem, we should put it in bad-practice category. Bad practice or "code-smells" are typically exactly like this: they feel like a cool short-cut but bite us back in the long term.

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Examples/ConwayFeatures.hs
aniketd: > The only thing bang patterns here do is say: regardless if the binding is used or not, evaluate it.

TIL :raised_hands: :bow: :100:

Soupstraw: I agree that it's a bad assumption in the sense that it doesn't specify *when* the value has to be evaluated, but it seems to be a good enough assumption for testing and tracing, because GHC chose to evaluate them in the order they appear in the `do` block. See this: https://play.haskell.org/saved/EzsvMLPx

Soupstraw: But it's definitely a horrible idea to rely on that behaviour in the implementation :100:

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Epoch.hs
Soupstraw: I added previous `prevDRepState` and `prevCommitteeState` to `GovActionsState`

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Examples/ConwayFeatures.hs
Soupstraw: It's not the end of the world if this assumption were to not hold one day, but it's convenient for debugging while it does hold

lehins: @Soupstraw You are totally wrong about this!!

Without running the code, try to guess which error message you will see first:

```haskell
main :: IO ()
main = do
  let !x = error "x"
      !z = x + y
      !y = error "y"
  traceM "Still made it"
```

https://play.haskell.org/saved/YbjO28OE

So, please remove those bangs, so not to confuse anyone.

lehins: If you ever want to require specific ordering for debugging or testing, you should either:
* manually specify it with a `seq`, eg. 
```haskell
x `seq` y `seq` traceM "Tested"
```
* or use [`evaluate`](https://downloads.haskell.org/ghc/9.2.8/docs/html/libraries/base-4.16.4.0/Control-Exception.html#v:evaluate):
```haskell
x' <- evaluate x
y' <- evaluate y
traceM "Tested"
```

Soupstraw: Ok you convinced me :smile:

Soupstraw: but then maybe it's a good idea to use `evaluate` on `runEPOCH` and `runLEDGER` since those can fail

lehins: What I suggest is to change `expectRight` into:

```haskell
expectRight :: (HasCallStack, Show a) => String -> Either a b ->  IO b
expectRight msg = either (\x -> error (msg <> show x)) evaluate
```

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Examples/ConwayFeatures.hs
```diff
@@ -201,16 +210,21 @@ spoThreshold :: Rational
...
 
-proposalDeposit :: Integer
-proposalDeposit = 10
+proposalDeposit :: Coin
+proposalDeposit = Coin 10
 
-expectRight :: Show a => String -> Either a b -> b
-expectRight msg = either (\x -> error (msg <> show x)) id
+-- | Evaluates an `Either` value and fails immediately if the result is `Left`
+expectRight :: Show a => String -> Either a b -> IO b
```
lehins: Add a HasCallStack constraint, so we can see actually which occurrence fails
```suggestion
expectRight :: (HasCallStack, Show a) => String -> Either a b -> IO b
```

eras/conway/impl/src/Cardano/Ledger/Conway/Governance.hs
```diff
@@ -170,32 +181,88 @@ instance EraPParams era => EncCBOR (GovActionState era) where
         !> To gasProposedIn
         !> To gasExpiresAfter
 
-newtype GovActionsState era = GovActionsState
-  { unGovActionsState :: Map (GovActionId (EraCrypto era)) (GovActionState era)
+data GovActionsState era = GovActionsState
+  { curGovActionsState :: !(Map (GovActionId (EraCrypto era)) (GovActionState era))
+  , prevGovActionsState :: !(Map (GovActionId (EraCrypto era)) (GovActionState era))
+  , prevDRepState :: !(Map (Credential 'DRepRole (EraCrypto era)) (DRepState (EraCrypto era)))
```
lehins: Just to disambiguate with state for a single DRep
```suggestion
  , prevDRepsState :: !(Map (Credential 'DRepRole (EraCrypto era)) (DRepState (EraCrypto era)))
```

eras/conway/impl/src/Cardano/Ledger/Conway/Governance.hs
```diff
@@ -170,32 +181,88 @@ instance EraPParams era => EncCBOR (GovActionState era) where
...
+  Lens'
+    (GovActionsState era)
+    (Map (GovActionId (EraCrypto era)) (GovActionState era))
+prevGovActionsStateL = lens prevGovActionsState (\x y -> x {prevGovActionsState = y})
+
+prevDRepStateL ::
+  Lens'
+    (GovActionsState era)
+    (Map (Credential 'DRepRole (EraCrypto era)) (DRepState (EraCrypto era)))
+prevDRepStateL = lens prevDRepState (\x y -> x {prevDRepState = y})
```
lehins: Same here.
```suggestion
prevDRepsStateL ::
  Lens'
    (GovActionsState era)
    (Map (Credential 'DRepRole (EraCrypto era)) (DRepState (EraCrypto era)))
prevDRepsStateL = lens prevDRepsState (\x y -> x {prevDRepsState = y})
```

eras/conway/impl/src/Cardano/Ledger/Conway/Governance.hs
```diff
@@ -170,32 +181,88 @@ instance EraPParams era => EncCBOR (GovActionState era) where
...
 
 deriving instance EraPParams era => Show (GovActionsState era)
 
-deriving instance EraPParams era => ToJSON (GovActionsState era)
+toGovActionsStatePairs :: (KeyValue a, EraPParams era) => GovActionsState era -> [a]
+toGovActionsStatePairs gas@(GovActionsState _ _ _ _) =
+  let GovActionsState {..} = gas
+   in [ "curGovActionsState" .= curGovActionsState
+      , "prevGovActionsState" .= prevGovActionsState
+      , "prevDRepState" .= prevDRepState
```
lehins: And here
```suggestion
      , "prevDRepsState" .= prevDRepsState
```

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Epoch.hs
```diff
@@ -238,13 +233,18 @@ epochTransition = do
...
           acnt'
-          (returnProposalDeposits adjustedLState)
+          adjustedLState
           ss'
           nm
 
   let
-    utxoSt''' = utxoSt' {utxosDeposited = obligationCertState adjustedCertState}
-    acnt'' = acnt' {asReserves = asReserves acnt'}
+    utxoSt''' =
```
lehins: going from one tick to three  could lead to confusion.
```suggestion
    utxoSt'' =
```

eras/conway/impl/src/Cardano/Ledger/Conway/Governance.hs
```diff
@@ -170,32 +181,88 @@ instance EraPParams era => EncCBOR (GovActionState era) where
...
+      , "prevCommitteeState" .= prevCommitteeState
+      -- TODO Add prevDRepState
```
lehins: It is already added
```suggestion
```

eras/conway/impl/src/Cardano/Ledger/Conway/Governance.hs
```diff
@@ -170,32 +181,88 @@ instance EraPParams era => EncCBOR (GovActionState era) where
...
+  def = GovActionsState mempty mempty mempty def
 
-deriving instance EraPParams era => EncCBOR (GovActionsState era)
+instance EraPParams era => EncCBOR (GovActionsState era) where
+  encCBOR GovActionsState {..} =
+    encode $
+      Rec GovActionsState
+        !> To curGovActionsState
+        !> To prevGovActionsState
+        !> To prevDRepState
```
lehins: I though Coders protects us from mistakes like this? Serialization for `prevCommitteeState` is missing.

FYI, this is probably why CI is failing

eras/conway/impl/src/Cardano/Ledger/Conway/Governance.hs
```diff
@@ -392,24 +466,30 @@ instance EraPParams era => ToJSON (ConwayGovState era) where
...
-  getConstitution g = Just $ g ^. cgRatifyStateL . rsEnactStateL . ensConstitutionL
+  getConstitution g = Just $ g ^. cgEnactStateL . ensConstitutionL
 
   curPParamsGovStateL = curPParamsConwayGovStateL
 
   prevPParamsGovStateL = prevPParamsConwayGovStateL
 
+  obligationGovState st =
+    sumVal $
+      gasDeposit <$> (st ^. cgGovActionsStateL . curGovActionsStateL)
```
lehins: This will be slightly more efficient, since it avoids construction of another intermediate Map. Also `fmap` for Maps is lazy, so this `gasDeposit <$>` could result in spaceleaks
```suggestion
    foldMap' gasDeposit (st ^. cgGovActionsStateL . curGovActionsStateL)
```

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Epoch.hs
```diff
@@ -170,16 +169,12 @@ returnProposalDepositsUMap gaids m =
...
   forall era.
-  GovState era ~ ConwayGovState era =>
+  StrictSeq (GovActionState era) ->
   LedgerState era ->
   LedgerState era
-returnProposalDeposits ls@LedgerState {..} =
-  ls
-    & lsCertStateL . certDStateL . dsUnifiedL %~ dstate
+returnProposalDeposits removedProposals =
+  lsCertStateL . certDStateL . dsUnifiedL %~ dstate
```
lehins: I know you didn't name it, but since you touched this function, might as well fix it. I think `dstate` is a terrible name for the binding
```suggestion
  lsCertStateL . certDStateL . dsUnifiedL %~ updateUMap
```

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Epoch.hs
```diff
@@ -253,31 +253,62 @@ epochTransition = do
...
+        , RatifyState
+            { rsRemoved = mempty
+            , rsEnactState = govSt ^. cgEnactStateL
+            }
+        , ratSig
+        )
+  let
+    lState = returnProposalDeposits rsRemoved $ esLState epochState'
+    es'' =
+      epochState'
```
lehins: Let's make naming consistent. Either switch all to `es` or all to `epochState`. 
```suggestion
      es'
```

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Gov.hs
```diff
@@ -186,7 +186,7 @@ noSuchGovActions ::
   GovActionsState era ->
   Set.Set (GovActionId (EraCrypto era)) ->
   Test (ConwayGovPredFailure era)
-noSuchGovActions (GovActionsState st) gaIds =
+noSuchGovActions (GovActionsState st _ _ _) gaIds =
   let unknownGovActionIds = Set.filter (`Map.notMember` st) gaIds
```
lehins: Pattern matching becomes a little bit dangerous when we have two fields with the same type. Also switching to explicit record makes it a bit clearer which state we are using here: previous or current.
```suggestion
noSuchGovActions gas gaIds =
  let unknownGovActionIds = Set.filter (`Map.notMember` gasCurGovActionsState gas) gaIds
```

lehins (changes requested):
Just a few minor changes are needed, otherwise it looks great!!!

eras/conway/impl/src/Cardano/Ledger/Conway/Governance.hs
Soupstraw: Yeah, I remember pointing out this problem with the current `Decoder` and `Encoding` types. We should be able to fix this if we add a phantom type to those types that tracks which type we're actually coding. Right now the `encode` and `decode` functions forget the type of the value we're coding, so in this case it's actually encoding `CommitteeState era -> GovActionsState era`

lehins (approved):
Awesome work. Thank you.

PR #3636, opened 2023-08-11
Implement DRep ratification with an 'always passing' threshold

### LINKED ISSUE #3612: Implement RATIFY logic for DReps
At the moment, only SPO votes are counted. We need to consider also DRep votes

teodanciu:
# Description

Closes https://github.com/input-output-hk/cardano-ledger/issues/3612

An example of the computation:

For these Dreps:
```
D1 -> s1
D2 -> s2
D3 -> s3
D4 -> s4
Abstain -> sA
NoConfidence -> sN
```

and these votes:
```
D1 -> Yes
D2 -> No
D3 -> Abstain
D9 -> No  -- This is not in the drep distribution map
```

We need to compute this ratio in case of an action different from NoConfidence:
` s1/(s1 + s2 + s3 + s4 + sA + sN - s3 -sA)` 
and this one for NoConfidence:
` (s1 + sN)/(s1 + s2 + s3 + s4 + sA + sN - s3 -sA)` 

Because we don't have the ratios precomputed (as is the case  for spo stake distribution), instead of folding to get the total and then subtracting the abstains, we fold the distribution and only add what's necessary, depending on the vote and on the action. 

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Ratify.hs
+++ b/eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Ratify.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE BangPatterns #-}
 {-# LANGUAGE DataKinds #-}
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE FlexibleInstances #-}
@@ -16,6 +17,8 @@ module Cardano.Ledger.Conway.Rules.Ratify (
   RatifyState (..),
   RatifyEnv (..),
   RatifySignal (..),
+  dRepAccepted,
+  dRepAcceptedRatio,
 ) where
 
 import Cardano.Ledger.BaseTypes (ShelleyBase)
@@ -31,7 +34,7 @@ import Cardano.Ledger.Conway.Governance (
  )
 import Cardano.Ledger.Conway.Rules.Enact (EnactPredFailure, EnactState (..))
 import Cardano.Ledger.Core
-import Cardano.Ledger.Credential (Credential)
+import Cardano.Ledger.Credential (Credential (..))
 import Cardano.Ledger.Keys (KeyRole (..))
 import Cardano.Ledger.PoolDistr (PoolDistr (..), individualPoolStake)
 import Cardano.Ledger.Slot (EpochNo (..))
@@ -50,6 +53,7 @@ import Data.Monoid (Sum (..))
 import Data.Ratio ((%))
 import Data.Sequence.Strict (StrictSeq (..))
 import Data.Void (absurd)
+import Data.Word (Word64)
 
 data RatifyEnv era = RatifyEnv
   { reStakeDistr :: !(Map (Credential 'Staking (EraCrypto era)) Coin)
@@ -57,6 +61,7 @@ data RatifyEnv era = RatifyEnv
   , reDRepDistr :: !(Map (DRep (EraCrypto era)) (CompactForm Coin))
   , reCurrentEpoch :: !EpochNo
   }
+  deriving (Show)
 
 newtype RatifySignal era
   = RatifySignal
@@ -93,11 +98,14 @@ instance
 spoThreshold :: Rational
 spoThreshold = 51 % 100
 
+dRepThreshold :: Rational
+dRepThreshold = 0
+
 epochsToExpire :: EpochNo
 epochsToExpire = 30
 
-accepted :: RatifyEnv era -> GovActionState era -> Bool
-accepted RatifyEnv {reStakePoolDistr = PoolDistr poolDistr} gas =
+spoAccepted :: RatifyEnv era -> GovActionState era -> Bool
+spoAccepted RatifyEnv {reStakePoolDistr = PoolDistr poolDistr} gas =
   totalAcceptedStakePoolsRatio > getStakePoolThreshold gasAction
   where
     GovActionState {gasStakePoolVotes, gasAction} = gas
@@ -135,6 +143,49 @@ accepted RatifyEnv {reStakePoolDistr = PoolDistr poolDistr} gas =
       HardForkInitiation {} -> 101 % 100
       _ -> spoThreshold
 
+dRepAccepted :: forall era. RatifyEnv era -> GovActionState era -> Rational -> Bool
+dRepAccepted RatifyEnv {reDRepDistr} GovActionState {gasDRepVotes, gasAction} threshold =
+  dRepAcceptedRatio reDRepDistr gasDRepVotes gasAction >= threshold
+
+-- Compute the dRep ratio yes/(yes + no), where
+-- yes: is the total stake of
+--    - registered dReps that voted 'yes', plus
+--    - the AlwaysNoConfidence dRep, in case the action is NoConfidence
+-- no: is the total stake of
+--    - registered dReps that voted 'no', plus
+--    - registered dReps that did not vote for this action, plus
+--    - the AlwaysNoConfidence dRep
+-- In other words, the denominator `yes + no` is the total stake of all registered dReps, minus the abstain votes stake
+-- (both credential DReps and AlwaysAbstain)
+--
+-- We iterate over the dRep distribution, and incrementally construct the numerator and denominator.
+dRepAcceptedRatio ::
+  forall era.
+  Map (DRep (EraCrypto era)) (CompactForm Coin) ->
+  Map (Credential 'DRepRole (EraCrypto era)) Vote ->
+  GovAction era ->
+  Rational
+dRepAcceptedRatio reDRepDistr gasDRepVotes gasAction
+  | totalExcludingAbstainStake == 0 = 0
+  | otherwise = toInteger yesStake % toInteger totalExcludingAbstainStake
+  where
+    accumStake :: (Word64, Word64) -> DRep (EraCrypto era) -> CompactForm Coin -> (Word64, Word64)
+    accumStake (!yes, !tot) drep (CompactCoin stake) =
+      case drep of
+        DRepCredential cred ->
+          case Map.lookup cred gasDRepVotes of
+            Nothing -> (yes, tot + stake)
+            Just VoteYes -> (yes + stake, tot + stake)
+            Just Abstain -> (yes, tot)
+            Just VoteNo -> (yes, tot + stake)
+        DRepAlwaysNoConfidence ->
+          case gasAction of
+            NoConfidence _ -> (yes + stake, tot + stake)
+            _ -> (yes, tot + stake)
+        DRepAlwaysAbstain -> (yes, tot)
+
+    (yesStake, totalExcludingAbstainStake) = Map.foldlWithKey' accumStake (0, 0) reDRepDistr
+
 ratifyTransition ::
   forall era.
   ( Embed (EraRule "ENACT" era) (ConwayRATIFY era)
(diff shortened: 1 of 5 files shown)


eras/conway/impl/test/Test/Cardano/Ledger/Conway/RatifySpec.hs
```diff
@@ -0,0 +1,84 @@
+{-# LANGUAGE AllowAmbiguousTypes #-}
+{-# LANGUAGE DataKinds #-}
+{-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE FlexibleInstances #-}
+{-# LANGUAGE LambdaCase #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeApplications #-}
+{-# OPTIONS_GHC -Wno-type-defaults #-}
```
teodanciu: Note to self to sort this problem out
```
    • Defaulting the following constraints to type ‘Double’
        (RealFrac a0)
          arising from a use of ‘ceiling’
          at test/Test/Cardano/Ledger/Conway/RatifySpec.hs:76:21-27
        (Fractional a0)
          arising from a use of ‘fromRational’
```

eras/conway/impl/test/Test/Cardano/Ledger/Conway/RatifySpec.hs
```diff
@@ -0,0 +1,95 @@
...
+                | otherwise = Just $ stakeYes % (stakeYes + stakeNo + notVotedStake + stakeAlwaysNoConfidence)
+          actual `shouldBe` expectedRephrased
+
+          let actualNoConfidence = dRepAcceptedRatio @era dRepDistr votes (NoConfidence SNothing)
+              expectedNoConfidence
+                | totalStake == stakeAbstain + stakeAlwaysAbstain = Nothing
+                | otherwise = Just $ (stakeYes + stakeAlwaysNoConfidence) % (totalStake - stakeAbstain - stakeAlwaysAbstain)
+          actualNoConfidence `shouldBe` expectedNoConfidence
+  where
+    ratio pct tot = ceiling $ fromRational ((pct * tot) % 100)
```
lehins: This will allow you to remove the ghc warning flag at the top
```suggestion
    ratio pct tot = ceiling $ (pct * tot) % 100
```

teodanciu: Thanks!

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Ratify.hs
```diff
@@ -135,6 +140,51 @@ accepted RatifyEnv {reStakePoolDistr = PoolDistr poolDistr} gas =
       HardForkInitiation {} -> 101 % 100
       _ -> spoThreshold
 
+dRepAccepted :: forall era. RatifyEnv era -> GovActionState era -> Bool
+dRepAccepted RatifyEnv {reDRepDistr} GovActionState {gasDRepVotes, gasAction} =
+  case dRepAcceptedRatio reDRepDistr gasDRepVotes gasAction of
+    Nothing -> True -- TODO: change this when we are ready to consider dReps on sanchonet
+    Just ratio -> ratio > dRepThreshold
```
Soupstraw: According to the spec, it should also return True if the threshold is set to zero:
https://github.com/input-output-hk/formal-ledger-specifications/blob/master/src/Ledger/Ratify.lagda#L225

Soupstraw: and it looks like this is supposed to be a non-strict inequality
```suggestion
    Just ratio -> ratio >= dRepThreshold
```
this should also fix the zero threshold issue

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Ratify.hs
```diff
@@ -135,6 +140,51 @@ accepted RatifyEnv {reStakePoolDistr = PoolDistr poolDistr} gas =
       HardForkInitiation {} -> 101 % 100
       _ -> spoThreshold
 
+dRepAccepted :: forall era. RatifyEnv era -> GovActionState era -> Bool
```
Soupstraw: I think it'd be better if we could reuse some of the logic across all the `*Accepted` functions. Right now it looks like there's some repetition between `dRepAccepted` and `spoAccepted`

Soupstraw: So for example the difference between `dRepAccepted` and `spoAccepted` is the map they use for looking up votes and what the default vote is when somebody hasn't submitted a vote, so we could make a function that takes both the map and the default vote and then use that function for both DReps and SPOs.

teodanciu: The conditions are a bit different though, because for `spoAccepted` we already have the ratios, but here we have to compute them.  So I'm iterating the distribution and computing the ratios incrementally, whereas for spos we are iterating the votes, and then looking the corresponding stake up in the distribution. 
If we went with the same formula, we'd have to first add all the stake and then compute what needs to be removed and remove it.  But maybe there is a better way indeed

Soupstraw (changes requested):
Good work so far, could use some re-organizing before merge :+1:

teodanciu:
An example of the computation:

For these Dreps:
```
D1 -> s1
D2 -> s2
D3 -> s3
D4 -> s4
Abstain -> sA
NoConfidence -> sN
```

and these votes:
```
D1 -> Yes
D2 -> No
D3 -> Abstain
D9 -> No  -- This is not in the drep distribution map
```

We need to compute this ratio in case of an action different from NoConfidence:
` s1/(s1 + s2 + s3 + s4 + sA + sN - s3 -sA)` 
and this one for NoConfidence:
` (s1 + sN)/(s1 + s2 + s3 + s4 + sA + sN - s3 -sA)` 

Because we don't have the ratios precomputed (as is the case  for spo stake distribution), instead of folding to get the total and then subtracting the abstains, we fold the distribution and only add what's necessary, depending on the vote and on the action.

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Ratify.hs
teodanciu: Good catch, I missed that. Thanks

Soupstraw (approved):

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Ratify.hs
```diff
@@ -135,6 +142,53 @@ accepted RatifyEnv {reStakePoolDistr = PoolDistr poolDistr} gas =
...
+  Map (DRep (EraCrypto era)) (CompactForm Coin) ->
+  Map (Credential 'DRepRole (EraCrypto era)) Vote ->
+  GovAction era ->
+  Maybe (Ratio Word64)
+dRepAcceptedRatio reDRepDistr gasDRepVotes gasAction
+  | totalExcludingAbstainStake == 0 = Nothing
+  | otherwise = Just (yesStake % totalExcludingAbstainStake)
+  where
+    accumStake :: (Word64, Word64) -> DRep (EraCrypto era) -> CompactForm Coin -> (Word64, Word64)
+    accumStake (yes, tot) drep (CompactCoin stake) =
```
lehins: Need to force the accumulator in order to avoid space leaks
```suggestion
    accumStake (!yes, !tot) drep (CompactCoin stake) =
```

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Ratify.hs
```diff
@@ -135,6 +142,53 @@ accepted RatifyEnv {reStakePoolDistr = PoolDistr poolDistr} gas =
...
+-- In other words, the denominator `yes + no` is the total stake of all registered dReps, minus the abstain votes stake
+-- (both credential DReps and AlwaysAbstain)
+--
+-- We iterate over the dRep distribution, and incrementally construct the numerator and denominator.
+dRepAcceptedRatio ::
+  forall era.
+  Map (DRep (EraCrypto era)) (CompactForm Coin) ->
+  Map (Credential 'DRepRole (EraCrypto era)) Vote ->
+  GovAction era ->
+  Maybe (Ratio Word64)
```
lehins: Ratio should never be used with fixed width integral numbers. They are susceptible to overflows in very surprising ways. See my comments on this topic in this issue: https://github.com/input-output-hk/cardano-ledger/pull/2326
```suggestion
  Maybe (Ratio Integer)
```

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Ratify.hs
```diff
@@ -135,6 +142,53 @@ accepted RatifyEnv {reStakePoolDistr = PoolDistr poolDistr} gas =
...
+dRepAcceptedRatio reDRepDistr gasDRepVotes gasAction
+  | totalExcludingAbstainStake == 0 = Nothing
+  | otherwise = Just (yesStake % totalExcludingAbstainStake)
```
lehins: See my comment above
```suggestion
  | otherwise = Just (toInteger yesStake % toInteger totalExcludingAbstainStake)
```

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Ratify.hs
```diff
@@ -135,6 +142,53 @@ accepted RatifyEnv {reStakePoolDistr = PoolDistr poolDistr} gas =
       HardForkInitiation {} -> 101 % 100
       _ -> spoThreshold
 
+dRepAccepted :: forall era. RatifyEnv era -> GovActionState era -> Ratio Word64 -> Bool
+dRepAccepted RatifyEnv {reDRepDistr} GovActionState {gasDRepVotes, gasAction} threshold =
+  case dRepAcceptedRatio reDRepDistr gasDRepVotes gasAction of
+    Nothing
+      | threshold == 0 -> True
+      | otherwise -> False
+    Just ratio -> ratio >= threshold
```
lehins: I am not sure if returning Nothing and handling it specially makes it clearer or not. Instead of Nothing we could just return `0`, in which case this would just become
```suggestion
  dRepAcceptedRatio reDRepDistr gasDRepVotes gasAction >= threshold
```

Feel free to ignore this suggestion if you believe we should make this case explicit.

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Ratify.hs
```diff
@@ -135,6 +142,53 @@ accepted RatifyEnv {reStakePoolDistr = PoolDistr poolDistr} gas =
...
+            Just VoteYes -> (yes + stake, tot + stake)
+            Just Abstain -> (yes, tot)
+            Just VoteNo -> (yes, tot + stake)
+        DRepAlwaysNoConfidence ->
+          case gasAction of
+            NoConfidence _ -> (yes + stake, tot + stake)
+            _ -> (yes, tot + stake)
+        DRepAlwaysAbstain -> (yes, tot)
+
+    (yesStake, totalExcludingAbstainStake) = Map.foldlWithKey accumStake (0, 0) reDRepDistr
```
lehins: For performance it is always better to use a strict variant of a fold
```suggestion
    (yesStake, totalExcludingAbstainStake) = Map.foldlWithKey' accumStake (0, 0) reDRepDistr
```

lehins (approved):
Looks great! Thank you.

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Ratify.hs
teodanciu: I was a little hesitant because the ratio is undefined rather than zero, but it does simplify the logic and it's maybe easier to understand what's going on, so i changed it  to remove the intermediate `Maybe`

teodanciu:
Good Haskell lessons from your review @lehins , thank you

PR #3637, opened 2023-08-11
Avoid using sequence of tuples, by adding GovActionId to GovActionState

### LINKED ISSUE #3578: Store `GovernanceAction`s in a strict data structure
We want to either use a strict tuple here or create a separate type with fields that are strict for:
```
data GovActionSomthing = GoveActionSomthing !(GovernanceActionId (EraCrypto era)) !(GovernanceActionState era)
```

_Originally posted by @lehins in https://github.com/input-output-hk/cardano-ledger/pull/3495#discussion_r1273367801_

Soupstraw:
# Description

closes #3578

This PR fixes the above ticket in a different way. Namely, by adding `govActionId` to the `GovActionState`. This approach is very similar to how `PoolParams` are implemented.

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/eras/conway/impl/CHANGELOG.md
+++ b/eras/conway/impl/CHANGELOG.md
@@ -1,8 +1,11 @@
 # Version history for `cardano-ledger-conway`
 
-## 1.7.0.1
+## 1.8.0.0
 
-*
+* Add field `gasId` to `GovActionState`
+* Add `insertGovActionsState`
+* Change type of `rsRemoved` in `RatifyState` to use  `GovActionState` instead of a tuple
+* Change `RatifySignal` to use `GovActionsState` instead of a tuple
 
 ## 1.7.0.0
 
--- a/eras/conway/impl/CHANGELOG.md
+++ b/eras/conway/impl/CHANGELOG.md
@@ -6,7 +6,6 @@
 
 ## 1.7.0.0
 
-* Add `Network` validation for `ProposalProcedure` and `TreasuryWithdrawals` in GOV #3659
 * Make `DELEG`, `POOL` and `GOVCERT` conform to spec-v0.8 #3628
   * Add `CertEnv` and `CertsEnv` to pass `EpochNo` down from `LEDGER` to sub-rules
   * Add `drepDeposit` to `DRepState` to track deposits paid by `DRep`s
--- a/eras/conway/impl/src/Cardano/Ledger/Conway/Governance.hs
+++ b/eras/conway/impl/src/Cardano/Ledger/Conway/Governance.hs
@@ -17,7 +17,7 @@
 module Cardano.Ledger.Conway.Governance (
   EraGov (..),
   GovActionsState (..),
-  insertGovActionEntry,
+  insertGovActionsState,
   EnactState (..),
   RatifyState (..),
   ConwayGovState (..),
--- a/eras/conway/impl/src/Cardano/Ledger/Conway/Governance.hs
+++ b/eras/conway/impl/src/Cardano/Ledger/Conway/Governance.hs
@@ -3,6 +3,7 @@
 {-# LANGUAGE DerivingStrategies #-}
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE NamedFieldPuns #-}
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE RecordWildCards #-}
 {-# LANGUAGE ScopedTypeVariables #-}
@@ -16,6 +17,7 @@
 module Cardano.Ledger.Conway.Governance (
   EraGov (..),
   GovActionsState (..),
+  insertGovActionEntry,
   EnactState (..),
   RatifyState (..),
   ConwayGovState (..),
@@ -95,13 +97,15 @@ import Control.DeepSeq (NFData (..))
 import Data.Aeson (KeyValue, ToJSON (..), object, pairs, (.=))
 import Data.Default.Class (Default (..))
 import Data.Map.Strict (Map)
+import qualified Data.Map.Strict as Map
 import Data.Sequence.Strict (StrictSeq)
 import GHC.Generics (Generic)
 import Lens.Micro
 import NoThunks.Class (NoThunks)
 
 data GovActionState era = GovActionState
-  { gasCommitteeVotes :: !(Map (Credential 'HotCommitteeRole (EraCrypto era)) Vote)
+  { gasId :: !(GovActionId (EraCrypto era))
+  , gasCommitteeVotes :: !(Map (Credential 'HotCommitteeRole (EraCrypto era)) Vote)
   , gasDRepVotes :: !(Map (Credential 'DRepRole (EraCrypto era)) Vote)
   , gasStakePoolVotes :: !(Map (KeyHash 'StakePool (EraCrypto era)) Vote)
   , gasDeposit :: !Coin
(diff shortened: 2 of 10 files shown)


eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Epoch.hs
```diff
@@ -266,7 +267,7 @@ epochTransition = do
           & prevPParamsEpochStateL .~ pp
           & curPParamsEpochStateL .~ pp
       -- TODO can we be more efficient?
-      newGov = GovActionsState . Map.fromList . toList $ rsFuture
+      newGov = foldr' insertGovActionEntry mempty rsFuture
```
teodanciu: Is it this in particular that improves the strictness in this solution?

Soupstraw: The main improvement is that the fields in `GovActionEntry` are strict as opposed to the pair we had before. Here I managed to avoid a conversion to list which is also nice.

teodanciu: Makes sense, thank you for explaining.

eras/conway/impl/src/Cardano/Ledger/Conway/Governance.hs
```diff
@@ -272,18 +282,42 @@ instance EraPParams era => NFData (EnactState era)
...
-  , rsFuture ::
-      !( StrictSeq
-          (GovActionId (EraCrypto era), GovActionState era)
-       )
-  , rsRemoved ::
-      !( StrictSeq
-          (GovActionId (EraCrypto era), GovActionState era)
-       )
+  , rsFuture :: !(StrictSeq (GovActionEntry era))
+  , rsRemoved :: !(StrictSeq (GovActionEntry era))
```
lehins: While looking through the PR I realized, why don't we just add `GovActionId` to the `GovActionState`? It will simplify things. The normal argument of it being duplicated in memory, since it is also used as a key in the GovActionsState Map does't really apply, since values will be shared.

@Soupstraw What do you think?
```suggestion
  , rsFuture :: !(StrictSeq (GovActionState era))
  , rsRemoved :: !(StrictSeq (GovActionState era))
```

eras/conway/impl/src/Cardano/Ledger/Conway/Governance.hs
```diff
@@ -162,7 +169,14 @@ instance EraPParams era => EncCBOR (GovActionState era) where
...
   }
-  deriving (Generic, NFData)
+  deriving (Generic, NFData, Semigroup, Monoid)
+
+insertGovActionEntry ::
+  GovActionState era ->
+  GovActionsState era ->
+  GovActionsState era
+insertGovActionEntry v@GovActionState {gasId} (GovActionsState m) =
+  GovActionsState $ Map.insert gasId v m
```
```suggestion
insertGovActionsState ::
  GovActionState era ->
  GovActionsState era ->
  GovActionsState era
insertGovActionsState v@GovActionState {gasId} (GovActionsState m) =
  GovActionsState $ Map.insert gasId v m
```

eras/conway/impl/CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## 1.7.0.0
 
+* Add `GovActionEntry`, `insertGovActionEntry`
```
```suggestion
* Add `insertGovActionsState`
```

lehins (approved):
Very nice. Found some remnants of the previous Entry type, but otherwise it looks great!

Thank you.

libs/cardano-ledger-pretty/CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## 1.3.0.0
 
+* Add `PrettyA` instance for `GovActionEntry`
```
teodanciu: This line is unnecessary

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Examples/ConwayFeatures.hs
```diff
@@ -372,7 +383,12 @@ testGov pf = do
     secondGovActionId = GovActionId (txid (secondProposalTx ^. bodyTxL)) (GovActionIx 0)
     expectedGovActionsState2 =
       GovActionsState $
-        Map.fromList [(secondGovActionId, govActionState (anotherConstitutionProposal pf govActionId))]
+        Map.fromList
+          [
+            ( secondGovActionId
+            , govActionState govActionId (anotherConstitutionProposal pf govActionId)
```
teodanciu: This should be `secondGovActionId`  :

```
govActionState secondGovActionId (anotherConstitutionProposal pf govActionId)
```

(sorry for the stupid names)

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Examples/ConwayFeatures.hs
```diff
@@ -391,11 +407,15 @@ testGov pf = do
...
-    [(gId, gas')] ->
+    [gas'] ->
       assertEqual
         "un-enacted govAction is recorded in rsFuture"
-        (gId, gas')
-        (secondGovActionId, govActionState (anotherConstitutionProposal pf govActionId))
+        (gasId gas', gas')
+        ( secondGovActionId
+        , govActionState
+            govActionId
```
teodanciu: Same here, `secondGovActionId`

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Epoch.hs
aniketd: @Soupstraw should we remove the TODO now that you have make it efficient enough? Or does it still remain?

teodanciu (approved):
Looks good to me (except you left an import of  the inexistent` insertGovActionEntry,` in `Governance`)

lehins (approved):
Looks great. Thank you!

eras/conway/impl/CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 
 ## 1.7.0.0
 
+* Add `insertGovActionsState`
```
lehins: Just realized something.

This entry now needs to go to the next version above and it needs to be switched to `1.8.0.0` (both in cabal file and in changelog)

Change in this is breaking, since `GovActionState` and `RatifyState` types has changed, which should also go into the change log:

```markdown
* `GovActionState` now has new field `gasId`
* Type for `rsRemoved` in `RatifyState` was changed.
```

No need to mention `rsFuture`, since it will be removed in a subsequent PR

eras/conway/impl/CHANGELOG.md
```diff
@@ -1,12 +1,14 @@
 # Version history for `cardano-ledger-conway`
 
-## 1.7.0.1
+## 1.8.0.0
 
-*
+* Add field `gasId` to `GovActionState`
+* Add `insertGovActionsState`
+* Change type of `rsFuture` in `RatifyState` to use  `GovActionState` instead of a tuple
```
lehins: It is the `rsFuture` that will be removed :grin: 
```suggestion
* Change type of `rsRemoved` in `RatifyState` to use `GovActionState` instead of a tuple
```

libs/cardano-ledger-pretty/cardano-ledger-pretty.cabal
```diff
@@ -1,6 +1,6 @@
 cabal-version:      3.0
 name:               cardano-ledger-pretty
-version:            1.3.0.1
```
teodanciu, line 3: @lehins  - I had to make these changes to because of the conway version change

lehins: That's right, it was expected.

lehins: Thank you for doing the change!

eras/conway/impl/CHANGELOG.md
```diff
@@ -1,12 +1,14 @@
...
 
-*
+* Add field `gasId` to `GovActionState`
+* Add `insertGovActionsState`
+* Change type of `rsRemoved` in `RatifyState` to use  `GovActionState` instead of a tuple
+* Change `RatifySignal` to use `GovActionsState` instead of a tuple
 
 ## 1.7.0.0
 
-* Add `Network` validation for `ProposalProcedure` and `TreasuryWithdrawals` in GOV #3659
```
aniketd, line 9: Did this line just get removed? Should it be part of `1.8` instead? @Soupstraw

teodanciu: Oops, looks like I removed it by mistake. Yes, I think it should be part of 1.8 because 1.7 was released  at (bbdfc3fdb6487653f23385c3df1c9c575bf567cf) before this commit . Sorry!

Issue #3639, opened 2023-08-11
Does AnchorData need its crypto type param?


nfrisby:
It's odd to me that `AnchorData` carries a crypto phantom parameter.

https://github.com/input-output-hk/cardano-ledger/blob/bf2dc0fdde313723b4f7a35362839920670e67ea/libs/cardano-ledger-core/src/Cardano/Ledger/BaseTypes.hs#L783

I ~can't~ can imagine a good reason for this, but it's not obvious.

This Issue is to remove it if it's unnecessary or to add a short comment explaining that it's necessary.

nfrisby:
To clarify:

Right now the code is:

```
data Anchor c = Anchor
  { anchorUrl :: !Url
  , anchorDataHash :: !(SafeHash c (AnchorData c))
  }

newtype AnchorData c = AnchorData ByteString
```

I was instead expecting:

```
data Anchor c = Anchor
  { anchorUrl :: !Url
  , anchorDataHash :: !(SafeHash c AnchorData)
  }
```
newtype AnchorData = AnchorData ByteString```

PR #3640, opened 2023-08-14
Remove crypto parameterization from `AnchorData`

lehins:
# Description

Besides fixing up AnchorData (#3639) this PR also:

* Fixes names in CDDL
* Removes redundant definition of a lnes function

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/eras/conway/test-suite/cddl-files/conway.cddl
+++ b/eras/conway/test-suite/cddl-files/conway.cddl
@@ -309,9 +309,9 @@ stake_vote_reg_deleg_cert = (13, stake_credential, pool_keyhash, drep, coin)
 ; GOVCERT
 reg_committee_hot_key_cert = (14, committee_cold_credential, committee_hot_credential)
 unreg_committee_hot_key_cert = (15, committee_cold_credential)
-reg_drep_cert = (16, voting_credential, coin, anchor / null)
-unreg_drep_cert = (17, voting_credential, coin)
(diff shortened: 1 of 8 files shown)


eras/conway/test-suite/cddl-files/conway.cddl
```diff
@@ -105,11 +105,11 @@ treasury_withdrawals_action = (2, { $reward_account => coin })
 
 no_confidence = (3, gov_action_id / null)
 
-new_committee = (4, gov_action_id / null, set<$committee_coldkey_credential>, committee)
+new_committee = (4, gov_action_id / null, set<$committee_cold_credential>, committee)
```
teodanciu, line 108: Not specific to this change, but just something I wanted to ask in the past: 

We seem to be using the `$` prefix in a sort of ad-hoc way in  our cddl no?
In the spec it says 
`Per convention, CDDL extension points are marked with a leading
   dollar sign (types)`

Is there some rationale for it that I am not seeing?

teodanciu (approved):
Looks good to me

eras/conway/test-suite/cddl-files/conway.cddl
lehins: Honestly, I have no idea. It'd be awesome if you could research what is the correct way that we should be using :grin:

PR #3648, opened 2023-08-14
Governance queries

### LINKED ISSUE #3624: Add Ord instance for DRepState
Just a question. Why can't we add an Ord instance for DRepState?

Seems like Anchor can easily get an Ord instance

_Originally posted by @lehins in https://github.com/input-output-hk/cardano-ledger/pull/3601#discussion_r1288269412_

lehins:
# Description

This PR implements 5 queries for consensus:

* `GetGovState`
* `GetDRepState`
* `GetDRepStakeDistr`
* `GetCommitteeState`
* `GetConstitution`

Also it adds a newtype `CommitteeState`

And fixes #3624 

# Checklist

- [x] Self-reviewed the diff

(Edit: After the PR was merged, I updated this description to also mention the fifth added query, `GetConstitution`.)

### DIFF
--- a/libs/cardano-ledger-api/src/Cardano/Ledger/Api/State/Query.hs
+++ b/libs/cardano-ledger-api/src/Cardano/Ledger/Api/State/Query.hs
@@ -6,17 +6,35 @@ module Cardano.Ledger.Api.State.Query (
   filterStakePoolDelegsAndRewards,
   queryStakePoolDelegsAndRewards,
 
+  -- * @GetGovState@
+  queryGovState,
+
+  -- * @GetConstitution@
+  queryConstitution,
+
   -- * @GetConstitutionHash@
   queryConstitutionHash,
+
+  -- * @GetDRepState@
+  queryDRepState,
+
+  -- * @GetDRepStakeDistr@
+  queryDRepStakeDistr,
+
+  -- * @GetCommitteeState@
+  queryCommitteeState,
 ) where
 
 import Cardano.Ledger.Allegra.Core (Constitution (constitutionAnchor))
+import Cardano.Ledger.CertState
 import Cardano.Ledger.Coin (Coin)
-import Cardano.Ledger.Conway.Governance (Anchor (..), AnchorData, EraGov (..))
+import Cardano.Ledger.Compactible (fromCompact)
 import Cardano.Ledger.Core
 import Cardano.Ledger.Credential (Credential)
-import Cardano.Ledger.Keys (KeyHash, KeyRole (StakePool, Staking))
+import Cardano.Ledger.DRepDistr (extractDRepDistr)
+import Cardano.Ledger.Keys (KeyHash, KeyRole (..))
 import Cardano.Ledger.SafeHash (SafeHash)
+import Cardano.Ledger.Shelley.Governance (EraGov (GovState, getConstitution))
 import Cardano.Ledger.Shelley.LedgerState
 import Cardano.Ledger.UMap (
   StakeCredentials (scRewards, scSPools),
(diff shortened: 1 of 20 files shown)


libs/cardano-ledger-core/src/Cardano/Ledger/DRepDistr.hs
```diff
@@ -51,7 +51,7 @@ data DRepState c = DRepState
   { drepExpiry :: !EpochNo
   , drepAnchor :: !(StrictMaybe (Anchor c))
   }
-  deriving (Show, Eq, Generic)
+  deriving (Show, Eq, Ord, Generic)
```
teodanciu, line 54: This is automatically using the `drepExpiry`  instance?

lehins: Yep, this is regular stock deriving mechanism for Ord, just as it is done with Eq.

teodanciu: But then it's also taking the `drepAnchor` into account?  I guess it's not a problem

lehins: Yes, of course :slightly_smiling_face:

libs/cardano-ledger-api/src/Cardano/Ledger/Api/State/Query.hs
```diff
@@ -52,12 +71,48 @@ getDState :: NewEpochState era -> DState era
...
   anchorDataHash . constitutionAnchor <$> queryConstitution nes
+
+-- | This query returns all of the state related to governance
+queryGovState :: NewEpochState era -> GovState era
+queryGovState nes = nes ^. nesEpochStateL . esLStateL . lsUTxOStateL . utxosGovStateL
+
+-- | Query DRep state.
+queryDRepState ::
+  NewEpochState era ->
+  -- | Specify state of which DReps should be returned. When this set is empty,
```
teodanciu: Just a nitpick, i find the phrasing a little confusing (took me a few readings)
Maybe:  `Specify DReps whose state should be returned` ?

lehins: Thanks. Done

libs/cardano-ledger-api/src/Cardano/Ledger/Api/State/Query.hs
```diff
@@ -52,12 +71,48 @@ getDState :: NewEpochState era -> DState era
...
+  NewEpochState era ->
+  -- | Specify a set of DRep credentials whose state should be returned. When this set is
+  -- empty, states for all of the DReps will be returned.
+  Set (Credential 'DRepRole (EraCrypto era)) ->
+  Map (Credential 'DRepRole (EraCrypto era)) (DRepState (EraCrypto era))
+queryDRepState nes creds
+  | null creds = drepsState
+  | otherwise = drepsState `Map.restrictKeys` creds
+  where
+    drepsState = vsDReps $ certVState $ lsCertState $ esLState $ nesEs nes
```
teodanciu, line 98: A little inconsistent with the above queryGovState which uses lenses, so was expecting:

 drepsState = nes ^. nesEsL . esLStateL . lsCertStateL . certVStateL . vsDRepsL

But it's maybe overkill to use lenses if the code is almost identical without.

teodanciu (approved):
Looks good to me

libs/cardano-ledger-api/src/Cardano/Ledger/Api/State/Query.hs
lehins: I don't think we should worry about lens consistency for now. I have a good idea how we can implement a consistent way of dealing with ledger state with a nice lens interface. But that will come later, definitely after SanchoNet.

Issue #3645, opened 2023-08-14
DRep deposits and refunds


lehins:
Three parts required:

1. Record the deposit amount in the DRep state
2. Validate the deposit and refund in GOVCERT rule
3. `getConsumedValue` and `getProducedValue` will need to account for DReps certificates, in the same way StakingCredential certificates are handled in Shelley.

PR #3650, opened 2023-08-15
Change environment for `ShelleyPOOLREAP` rule

lehins:
# Description

Previous type for environment was `PParams`, which was unused. Current environment is set to `VState`, because it will be needed for correct implemention of `obligationCertState` used in the rule's assertion. It will be needed for DRep deposits: #3645

Also rename CDDL Committee certificates variables:
* `reg_committee_hot_key_cert` -> `auth_committee_hot_cert`
* `unreg_committee_hot_key_cert` -> `resign_committee_cold_cert`

# Checklist

- [x] Self-reviewed the diff


Soupstraw (approved):
LGTM, thanks for adding the VState!

PR #3659, opened 2023-08-17
Validate Network for ProposalProcedure and TreasuryWithdrawal

### LINKED ISSUE #3603: Validate NetworkIds
Ensure that all RewardAcnt that were added in ConwayEra have their NetworkIds verified:

* [x] - ProposalProcedure
* [x] - TreasuryWithdrawal

aniketd:
# Description

Resolves #3603 

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Enact.hs
+++ b/eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Enact.hs
@@ -50,7 +50,7 @@ instance EraGov era => STS (ConwayENACT era) where
   initialRules = []
   transitionRules = [enactmentTransition]
 
-enactmentTransition :: forall era. EraGov era => TransitionRule (ConwayENACT era)
+enactmentTransition :: forall era. EraPParams era => TransitionRule (ConwayENACT era)
 enactmentTransition = do
   TRC ((), st, act) <- judgmentContext
 
--- a/eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Enact.hs
+++ b/eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Enact.hs
@@ -50,9 +50,10 @@ instance EraGov era => STS (ConwayENACT era) where
   initialRules = []
   transitionRules = [enactmentTransition]
 
-enactmentTransition :: forall era. EraPParams era => TransitionRule (ConwayENACT era)
+enactmentTransition :: forall era. EraGov era => TransitionRule (ConwayENACT era)
 enactmentTransition = do
   TRC ((), st, act) <- judgmentContext
+
   case act of
     ParameterChange _prevGovActionId ppup -> pure $ st {ensPParams = newPP}
       where
@@ -60,8 +61,6 @@ enactmentTransition = do
     HardForkInitiation _prevGovActionId pv -> pure $ st {ensProtVer = pv}
     TreasuryWithdrawals wdrls -> do
       let wdrlsAmount = fold wdrls
-          -- TODO: validate NetworkId in all
-          -- We can drop NetworkIds after their validation
           wdrlsNoNetworkId = Map.mapKeys getRwdCred wdrls
       runTest
         . failureUnless (wdrlsAmount <= ensTreasury st)
--- a/eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Gov.hs
+++ b/eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Gov.hs
@@ -42,7 +42,6 @@ import Cardano.Ledger.Core
 import Cardano.Ledger.Rules.ValidationMode (Inject (..), Test, runTest)
 import Cardano.Ledger.Shelley.Tx (TxId (..))
 import Control.DeepSeq (NFData)
-import Control.Monad (forM_)
 import Control.Monad.Trans.Reader (asks)
 import Control.State.Transition.Extended (
   STS (..),
@@ -67,8 +66,8 @@ data GovEnv era = GovEnv
 data ConwayGovPredFailure era
   = GovActionsDoNotExist (Set.Set (GovActionId (EraCrypto era)))
   | MalformedProposal (GovAction era)
-  | ProposalProcedureNetworkIdMismatch !Network !Network -- First is supplied and second is the expected.
-  | TreasuryWithdrawalsNetworkIdMismatch !Network !Network -- First is supplied and second is the expected.
+  | ProposalProcedureNetworkIdMismatch (RewardAcnt (EraCrypto era)) Network
+  | TreasuryWithdrawalsNetworkIdMismatch (Set.Set (RewardAcnt (EraCrypto era))) Network
   deriving (Eq, Show, Generic)
 
 instance EraPParams era => NFData (ConwayGovPredFailure era)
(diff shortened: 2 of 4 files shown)


eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Enact.hs
```diff
@@ -50,19 +52,20 @@ instance EraGov era => STS (ConwayENACT era) where
...
-          -- TODO: validate NetworkId in all
-          -- We can drop NetworkIds after their validation
-          wdrlsNoNetworkId = Map.mapKeys getRwdCred wdrls
+      expectedNetworkId <- liftSTS $ asks networkId
+      let withdrawals = Map.filterWithKey (\rwrdAcnt _ -> getRwdNetwork rwrdAcnt == expectedNetworkId) wdrls
+          wdrlsAmount = fold withdrawals
+          wdrlsNoNetworkId = Map.mapKeys getRwdCred withdrawals
       runTest
         . failureUnless (wdrlsAmount <= ensTreasury st)
         $ EnactTreasuryInsufficientFunds @era wdrls (ensTreasury st)
```
aniketd: `wdrls` here is intentionally left unedited. Should this instead be just the filtered ones?

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Enact.hs
```diff
@@ -50,19 +52,20 @@ instance EraGov era => STS (ConwayENACT era) where
...
     HardForkInitiation _prevGovActionId pv -> pure $ st {ensProtVer = pv}
     TreasuryWithdrawals wdrls -> do
-      let wdrlsAmount = fold wdrls
-          -- TODO: validate NetworkId in all
-          -- We can drop NetworkIds after their validation
-          wdrlsNoNetworkId = Map.mapKeys getRwdCred wdrls
+      expectedNetworkId <- liftSTS $ asks networkId
+      let withdrawals = Map.filterWithKey (\rwrdAcnt _ -> getRwdNetwork rwrdAcnt == expectedNetworkId) wdrls
+          wdrlsAmount = fold withdrawals
+          wdrlsNoNetworkId = Map.mapKeys getRwdCred withdrawals
```
lehins: NetworkId for the withdrawal account need to be validated upon proposal, i.e. in the `GOV` rule, not during Enactment.

If it is validated upon proposal, then here, when we ENACT this withdrawal action we can simply ignore the NetworkId value

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Gov.hs
```diff
@@ -63,6 +66,7 @@ data GovEnv era = GovEnv
 data ConwayGovPredFailure era
   = GovActionsDoNotExist (Set.Set (GovActionId (EraCrypto era)))
   | MalformedProposal (GovAction era)
+  | WrongNetworkIdInReturnAddressOfProposalProcedure !Network !Network -- First is supplied and second is the expected.
```
lehins: This is way tooooooo long. :smile: How about?
```suggestion
  | ProposalProcedureNetworkIdMismatch 
    -- | Supplied 
    !Network
    -- | Expected
    !Network
```

lehins (changes requested):

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Enact.hs
aniketd: Got it! Thanks for the pointers! :memo:

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Gov.hs
```diff
@@ -170,9 +177,15 @@ govTransition = do
   -- TODO Check the signatures
```
teodanciu: I don't think anyone knows what this TODO refers to.. maybe we can remove it?

aniketd: @Soupstraw any ideas?

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Gov.hs
```diff
@@ -165,14 +177,32 @@ actionWellFormed ga = failureUnless isWellFormed $ MalformedProposal ga
...
+        pProcNetworkId
+          == expectedNetworkId
+            ?! ProposalProcedureNetworkIdMismatch pProcNetworkId expectedNetworkId
+        case pProcGovAction of
+          TreasuryWithdrawals wdrls ->
+            forM_ (Map.keys wdrls) $ \rwrdAcnt ->
+              let wdrlNetworkId = getRwdNetwork rwrdAcnt
+               in wdrlNetworkId
+                    == expectedNetworkId
+                      ?! TreasuryWithdrawalsNetworkIdMismatch wdrlNetworkId expectedNetworkId
```
lehins: We don't need to report a failure for every account, it is better to report a set of accounts that violates this predicate. Also it is better to keep predicate check lazy. This way there is no cost payed upon replay of chain without validation.
```suggestion
            let mismatchedNetworkIds = 
                  Set.filter ((/= expectedNetworkId) . getRwdNetwork) $ Map.keys wdrls
               in Set.null mismatchedNetworkIds
                      ?! TreasuryWithdrawalsNetworkIdMismatch mismatchedNetworkIds expectedNetworkId
```

lehins (changes requested):

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Gov.hs
aniketd: :100: Got it!

aniketd: I didn't notice the suggestion you made here and ended up unifying the `ProposalProcedure` and `TreasuryWithdrawals` failures together. My bad. I will revert that change and apply this suggestion, before re-requesting your feedback :disappointed:

aniketd: In my high-contrast GitHub UI both the suggestions and context-snippets for comments are the same color. I should have payed more attention.

aniketd: I also misread your comment to mean that we need to report `Credential`s for which the `Network` is wrong.

aniketd: :100: Now I got it.

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Gov.hs
aniketd: I will make sure to remove the comment before merging if no one yells by tomorrow. :raised_hands:

Soupstraw: Yeah I think it's not relevant

aniketd: Thansk @Soupstraw ! :raised_hands:

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Enact.hs
```diff
@@ -50,18 +50,17 @@ instance EraGov era => STS (ConwayENACT era) where
   initialRules = []
   transitionRules = [enactmentTransition]
 
-enactmentTransition :: forall era. EraPParams era => TransitionRule (ConwayENACT era)
+enactmentTransition :: forall era. EraGov era => TransitionRule (ConwayENACT era)
```
lehins: Reverting back to the original constraint
```suggestion
enactmentTransition :: forall era. EraPParams era => TransitionRule (ConwayENACT era)
```

lehins (approved):
Perfect! Thank you

PR #3662, opened 2023-08-18
Equality on raw types

lehins:
# Description

This PR adds a `EqRaw` class that can be used to compare the Haskell type ignoring the underlying binary reopresentation for memoized times, regardless if they use MemoBytes or not. Default implementation, however, uses the MemoBytes representation, which will work for flat types based on MemoBytes, which is the majority of them. This approach simplifies the testing a bit.

@nc6 I took your suggestion of creating a separate class for this concept, it works out slightly nicer, because it will make tests less complex.

As a bonus this PR also adds ToExpr instances for some of the core types.

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/eras/allegra/impl/src/Cardano/Ledger/Allegra/TxAuxData.hs
+++ b/eras/allegra/impl/src/Cardano/Ledger/Allegra/TxAuxData.hs
@@ -43,6 +43,7 @@ import Cardano.Ledger.Core (
 import Cardano.Ledger.Crypto (Crypto (HASH))
 import Cardano.Ledger.Hashes (EraIndependentTxAuxData)
 import Cardano.Ledger.MemoBytes (
+  EqRaw,
   Mem,
   MemoBytes,
   MemoHashIndex,
@@ -107,6 +108,8 @@ newtype AllegraTxAuxData era = AuxiliaryDataWithBytes (MemoBytes AllegraTxAuxDat
   deriving (Generic)
   deriving newtype (Eq, ToCBOR, SafeToHash)
 
+instance ToExpr (AllegraTxAuxData era)
+
 instance Memoized AllegraTxAuxData where
   type RawType AllegraTxAuxData = AllegraTxAuxDataRaw
 
@@ -122,6 +125,8 @@ deriving newtype instance
 deriving newtype instance Era era => NoThunks (AllegraTxAuxData era)
 
 deriving newtype instance NFData (AllegraTxAuxData era)
+
+instance EqRaw (AllegraTxAuxData era)
 
 pattern AllegraTxAuxData ::
   Era era =>
--- a/eras/alonzo/impl/cardano-ledger-alonzo.cabal
+++ b/eras/alonzo/impl/cardano-ledger-alonzo.cabal
@@ -1,6 +1,6 @@
 cabal-version:      3.0
 name:               cardano-ledger-alonzo
-version:            1.4.0.1
+version:            1.4.1.0
 license:            Apache-2.0
 maintainer:         operations@iohk.io
 author:             IOHK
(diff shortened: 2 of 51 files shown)


eras/alonzo/impl/cardano-ledger-alonzo.cabal
teodanciu, line 1: Shouldn't we make this depend on cardano-ledger-core >=1.6, since it's using now `EqRaw` which is introduced in 1.6?

eras/babbage/impl/cardano-ledger-babbage.cabal
```diff
@@ -60,7 +60,7 @@ library
         cardano-crypto-class,
         cardano-data >=1.0,
         cardano-ledger-allegra >=1.1,
-        cardano-ledger-alonzo ^>=1.4,
+        cardano-ledger-alonzo ^>=1.4.1,
         cardano-ledger-binary >=1.0,
         cardano-ledger-core >=1.5 && <1.7,
```
teodanciu: Same question as for alonzo cabal file, shouldn't this depend on cardano-ledger-core >= 1.6 ?

eras/alonzo/impl/cardano-ledger-alonzo.cabal
lehins: Good catch. Fixed it for shelley and babbage packages too

eras/babbage/impl/cardano-ledger-babbage.cabal
lehins: :100: 

Fixed

teodanciu (commented):
(Just to cancel the review that I started I think because of a github glitch)

eras/allegra/impl/src/Cardano/Ledger/Allegra/TxAuxData.hs
```diff
@@ -107,6 +108,8 @@ newtype AllegraTxAuxData era = AuxiliaryDataWithBytes (MemoBytes AllegraTxAuxDat
   deriving (Generic)
   deriving newtype (Eq, ToCBOR, SafeToHash)
 
+instance ToExpr (AllegraTxAuxData era)
```
teodanciu, line 111: Just curious, is this because it was missing, or is it relevant to implementing `EqRaw` ?

lehins: It is not relevant for `EqRaw`, but just like `EqRaw` it is relevant for testing. Ultimately we want to add `ToExpr` instances for all of our types, since that improves failure reporting significantly.

teodanciu: Makes sense, thanks!

eras/alonzo/impl/cardano-ledger-alonzo.cabal
```diff
@@ -67,7 +67,7 @@ library
         cardano-ledger-allegra >=1.1,
         cardano-crypto-class,
         cardano-ledger-binary >=1.0.1,
-        cardano-ledger-core >=1.5 && <1.7,
+        cardano-ledger-core >=1.6 && <1.7,
         cardano-ledger-mary >=1.1,
         cardano-ledger-shelley ^>=1.5,
```
teodanciu: Here we would need at least 1.5.1.0, right? Because of `shelleyEqTxRaw`. 
But I'm not sure how this `^>=` operator works, it seems to me as if it would accept 1.5.0 too

lehins: You are right on point again. Thank you for checking this!

libs/cardano-ledger-core/testlib/Test/Cardano/Ledger/Core/Binary.hs
```diff
@@ -78,10 +73,12 @@ specTxAuxDataUpgrade =
         expectationFailure $
           "Expected to deserialize: =======================================================\n"
             ++ show err
-      Right (curTxAuxData :: t era) ->
-        -- We need to do all this MemoBytes trickery because underlying bytes and thus the
-        -- equality of the same type will no longer be the same, despite that the value will
-        zipMemoRawType @t @t expectExprEqual curTxAuxData (upgradeTxAuxData prevTxAuxData)
+      Right (curTxAuxData :: TxAuxData era) -> do
+        let upgradedTxAuxData = upgradeTxAuxData prevTxAuxData
+        unless (eqRaw curTxAuxData upgradedTxAuxData) $
```
teodanciu, line 78: Awesome that we could improve on the 'trickery'!

teodanciu (approved):
Looks good to me

PR #3664, opened 2023-08-21
Track proposal expiry

### LINKED ISSUE #3642: Track Proposal Procedure Expiry
GovAction must expire after the number of epochs specified in PParams.

It is also important not to affect expiry of current proposals, when `govActionExpiration` param is updated. For this we need to record the epoch when action expires in GovActionsState whenever ProposalProcedure is being proposed

aniketd:
# Description

Closes #3642 

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/eras/conway/impl/CHANGELOG.md
+++ b/eras/conway/impl/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## 1.8.0.0
 
+* Implement expiry for governance proposals #3664
+  * Update `ppGovActionExpiration` to be an `EpochNo`
+  * Add `gasExpiresAfter :: !EpochNo` to `GovActionState`
+  * Add `gePParams` to `GovEnv`
 * Add `reDRepState` to `RatifyEnv`
 * Add field `gasId` to `GovActionState`
 * Add `insertGovActionsState`
--- a/eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Gov.hs
+++ b/eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Gov.hs
@@ -24,7 +24,7 @@ import Cardano.Ledger.BaseTypes (EpochNo (..), Network, ShelleyBase, networkId)
 import Cardano.Ledger.Binary (DecCBOR (..), EncCBOR (..), FromCBOR (..), ToCBOR (..))
 import Cardano.Ledger.Binary.Coders (Decode (..), Encode (..), decode, encode, (!>), (<!))
 import Cardano.Ledger.Coin (Coin (..))
-import Cardano.Ledger.Conway.Core (ConwayEraPParams (..))
+import Cardano.Ledger.Conway.Core (ConwayEraPParams (..), ppGovActionExpirationL)
 import Cardano.Ledger.Conway.Era (ConwayGOV)
 import Cardano.Ledger.Conway.Governance (
   GovActionId (..),
@@ -55,12 +55,14 @@ import qualified Data.Map.Strict as Map
 import Data.Sequence (Seq (..))
 import qualified Data.Set as Set
 import GHC.Generics (Generic)
+import Lens.Micro ((^.))
 import NoThunks.Class (NoThunks (..))
 import Validation (failureUnless)
 
 data GovEnv era = GovEnv
-  { teTxId :: !(TxId (EraCrypto era))
-  , teEpoch :: !EpochNo
+  { geTxId :: !(TxId (EraCrypto era))
+  , geEpoch :: !EpochNo
+  , gePParams :: !(PParams era)
   }
 
 data ConwayGovPredFailure era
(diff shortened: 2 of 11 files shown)


eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Gov.hs
```diff
@@ -59,8 +59,9 @@ import NoThunks.Class (NoThunks (..))
 import Validation (failureUnless)
 
 data GovEnv era = GovEnv
-  { teTxId :: !(TxId (EraCrypto era))
-  , teEpoch :: !EpochNo
+  { geTxId :: !(TxId (EraCrypto era))
+  , geEpoch :: !EpochNo
+  , geGovActionExpiry :: !EpochNo
```
lehins: I think it is better to just pass all of PParams, since we'll need some of the other ones in the GOV rule in the future. Namely `minCommitteSize` and `govActionDeposit`
```suggestion
  , gePParams :: !PParams era
```

libs/cardano-ledger-pretty/cardano-ledger-pretty.cabal
```diff
@@ -39,7 +39,7 @@ library
         cardano-ledger-alonzo >=1.2,
         cardano-ledger-babbage >=1.1,
         cardano-ledger-byron,
-        cardano-ledger-conway ^>=1.7,
+        cardano-ledger-conway >=1.7,
```
lehins: That bound is invalid. You made a breaking change that `cardano-ledger-pretty` now depends on.
```suggestion
        cardano-ledger-conway ^>=1.8,
```

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Ratify.hs
```diff
@@ -217,11 +213,13 @@ ratifyTransition = do
...
+          -- This action hasn't been ratified yet. Process the remaining actions.
           st' <- trans @(ConwayRATIFY era) $ TRC (env, st, RatifySignal sigs)
-          if expired
-            then -- Action expired, do not include it in the next epoch
+          -- Finally, filter out actions that are not processed.
+          if gasExpiresAfter >= reCurrentEpoch
+            then -- Action expires after current Epoch. Remove it.
               pure $ st' {rsRemoved = act :<| rsRemoved}
-            else -- Include this action in the next epoch
+            else -- Action neither expires, nor is enacted, yet. Carry it over to the next Epoch.
```
lehins: This comment can be removed, since `rsFuture` will be removed in another PR
```suggestion
            else
```

eras/conway/impl/CHANGELOG.md
```diff
@@ -1,9 +1,12 @@
...
 
-## 1.7.0.1
-
-*
+## 1.8.0.0
 
+* Implement expiry for governance proposals #3664
+  * Update `ppGovActionExpiration` to be an `EpochNo`
+  * Add `gasExpiresAfter :: !EpochNo` to `GovActionState`
+  * Add `geGovActionExpiry` to `GovEnv` from the `PParams`
```
```suggestion
  * Add `gePParams` to `GovEnv`.
```

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Ratify.hs
```diff
@@ -217,11 +213,13 @@ ratifyTransition = do
                   }
           trans @(ConwayRATIFY era) $ TRC (env, st', RatifySignal sigs)
         else do
+          -- This action hasn't been ratified yet. Process the remaining actions.
           st' <- trans @(ConwayRATIFY era) $ TRC (env, st, RatifySignal sigs)
-          if expired
-            then -- Action expired, do not include it in the next epoch
+          -- Finally, filter out actions that are not processed.
+          if gasExpiresAfter >= reCurrentEpoch
```
lehins: For consistency with DRep expiration, the spec and the name of the record (After) :slightly_smiling_face: 

Also, as @teodanciu pointed out, expiration should be smaller than the current epoch in order for it to expire. It starts out as the epoch in the future (or current epoch, if the `govActionExpiration` pparam is set to 0)
```suggestion
          if gasExpiresAfter < reCurrentEpoch
```

lehins (changes requested):
Looks good. Few important changes are needed though

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Ratify.hs
teodanciu: Wait, so if the currentEpoch is smaller than the expiration Epoch, this means that it's expired?  Is this right?

lehins: You are absolutely right :smile: Updated my suggestion.

libs/cardano-ledger-pretty/cardano-ledger-pretty.cabal
aniketd: Oh yes! What I did is wrong :+1: Thanks!

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Ratify.hs
aniketd: :face_with_head_bandage: This mistake made me feel a bit sick :face_with_spiral_eyes:. Thanks for catching this!

aniketd: Thanks for the awesome catch :raised_hands:

lehins (approved):
:+1:

Issue #3655, opened 2023-08-16
Validate previous governance action id upon proposal


lehins:
This is a nice to have extra safety check that would check that the previous governance action of a specific type is either the very last action that has been enacted or is already in the current proposals set.

This would allow us to:
* filter out invalid proposals right from the start. 
* reconstruct a full chain of how proposals would be enacted, if they would be all enacted that is.
* fully predict the expected updates
* prevent circular dependencies within proposals

PR #3670, opened 2023-08-23
Validate previously enacted govAction

### LINKED ISSUE #3656: Validate Previous Governance Action Id upon enactment
Every GovAction contains a `PrevGovActionId`, which must correspond to the very last action that was enacted.

For this we need to store the four of different governance action  ids that were enacted last and update them when new GovActions get enacted.

teodanciu:
# Description

Closes https://github.com/input-output-hk/cardano-ledger/issues/3656

* EnactState: store the last enacted GovId for each of the 4 types of actions for which it is relevant (corresponding to GovActionPurpose)  
* Ratify: check that the previous govId that we have from the proposal matches the one that we store in the enact state. 
If it doesn't match, we "drop" the signal  (we don't enact it and it will eventually expire)
* Enact:  when enacting, update the stored value for last govId 

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/eras/conway/impl/src/Cardano/Ledger/Conway/Governance.hs
+++ b/eras/conway/impl/src/Cardano/Ledger/Conway/Governance.hs
@@ -45,12 +45,15 @@ module Cardano.Ledger.Conway.Governance (
   cgGovActionsStateL,
   cgEnactStateL,
   cgRatifyStateL,
+  ensCommitteeL,
   ensConstitutionL,
+  ensCurPParamsL,
   ensPrevGovActionIdsL,
   ensPrevPParamUpdateL,
   ensPrevHardForkL,
   ensPrevCommitteeL,
   ensPrevConstitutionL,
+  ensProtVerL,
   rsEnactStateL,
   curPParamsConwayGovStateL,
   prevPParamsConwayGovStateL,
@@ -340,8 +343,14 @@ data EnactState era = EnactState
   }
   deriving (Generic)
 
+ensCommitteeL :: Lens' (EnactState era) (StrictMaybe (Committee era))
+ensCommitteeL = lens ensCommittee (\x y -> x {ensCommittee = y})
+
 ensConstitutionL :: Lens' (EnactState era) (Constitution era)
 ensConstitutionL = lens ensConstitution (\x y -> x {ensConstitution = y})
+
+ensProtVerL :: Lens' (EnactState era) ProtVer
+ensProtVerL = lens ensProtVer (\x y -> x {ensProtVer = y})
 
 ensCurPParamsL :: Lens' (EnactState era) (PParams era)
 ensCurPParamsL = lens ensPParams (\es x -> es {ensPParams = x})
--- a/eras/conway/impl/src/Cardano/Ledger/Conway/Governance.hs
+++ b/eras/conway/impl/src/Cardano/Ledger/Conway/Governance.hs
@@ -405,7 +405,7 @@ toEnactStatePairs cg@(EnactState _ _ _ _ _ _ _ _) =
       , "prevPParams" .= ensPParams
       , "treasury" .= ensTreasury
       , "withdrawals" .= ensWithdrawals
-      , "PrevGovActionIds" .= ensPrevGovActionIds
+      , "prevGovActionIds" .= ensPrevGovActionIds
       ]
 
 deriving instance Eq (PParams era) => Eq (EnactState era)
--- a/eras/conway/impl/src/Cardano/Ledger/Conway/Governance.hs
+++ b/eras/conway/impl/src/Cardano/Ledger/Conway/Governance.hs
@@ -27,6 +27,7 @@ module Cardano.Ledger.Conway.Governance (
   GovActionIx (..),
   GovActionId (..),
   GovActionPurpose (..),
+  PrevGovActionIds (..),
   PrevGovActionId (..),
   govActionIdToText,
   Voter (..),
@@ -45,6 +46,11 @@ module Cardano.Ledger.Conway.Governance (
   cgEnactStateL,
   cgRatifyStateL,
   ensConstitutionL,
+  ensPrevGovActionIdsL,
+  ensPrevPParamUpdateL,
+  ensPrevHardForkL,
+  ensPrevCommitteeL,
+  ensPrevConstitutionL,
   rsEnactStateL,
   curPParamsConwayGovStateL,
   prevPParamsConwayGovStateL,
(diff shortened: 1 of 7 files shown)


eras/conway/impl/src/Cardano/Ledger/Conway/Governance.hs
```diff
@@ -203,34 +212,119 @@ instance EraPParams era => ToCBOR (GovActionsState era) where
 instance EraPParams era => FromCBOR (GovActionsState era) where
   fromCBOR = fromEraCBOR @era
 
+data LastGovActions era = LastGovActions
```
teodanciu: What do you guys think about this name? I initially called it `PrevGovActions`, but then I thought that having this as "last" seems more to indicate the "state"  of the system as opposed to the `Prev` reference from the proposals. But maybe it's creating more confusion.. I'm happy to change it to `PrevGovActions` or anything else if anyone thinks that's better.

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Enact.hs
```diff
@@ -52,13 +69,19 @@ instance EraGov era => STS (ConwayENACT era) where
...
 enactmentTransition = do
-  TRC ((), st, act) <- judgmentContext
+  TRC ((), st, EnactSignal govActionId act) <- judgmentContext
 
   case act of
-    ParameterChange _prevGovActionId ppup -> pure $ st {ensPParams = newPP}
-      where
-        newPP = ensPParams st `applyPPUpdates` ppup
-    HardForkInitiation _prevGovActionId pv -> pure $ st {ensProtVer = pv}
+    ParameterChange _ ppup ->
```
teodanciu, line 75: The fact that we have access here to the `prevGovActionId` worries  me (because I'm not using it). Should we be using it for something?

libs/cardano-ledger-pretty/src/Cardano/Ledger/Pretty/Conway.hs
```diff
@@ -311,7 +312,7 @@ instance PrettyA (Constitution era) where
       ]
 
 instance PrettyA (PParams era) => PrettyA (EnactState era) where
-  prettyA ens@(EnactState _ _ _ _ _ _ _) =
+  prettyA ens@(EnactState _ _ _ _ _ _ _ _) =
```
teodanciu, line 323: I see this pattern often, whereas it could be replaced with: 
```
prettyA EnactState {..} = 
  ppRecord ...
```
But I wonder if it's kept like this on purpose so we get a warning when adding a new field?

teodanciu: Nevermind, confirmed in the meeting

eras/conway/impl/src/Cardano/Ledger/Conway/Governance.hs
lehins: Here is my suggestion `PrevGovActionIds`.

The reason for this ordering is consistency with `PrevGovActionId` type and Ids suffix is important too, in order to emphasize that this is just an id, not the action itself.

Alternative would be to rename `PrevGovActionId` to either:
* `LastGovActionId`
* or `LastEnactedGovActionId`

Which would allow us to name this type either:
* `LastGovActionIds`
* `LastEnactedGovActionIds`

respectively.

Not sure how much effort you wanna put into this, but consistency is more important that the actual name. A proper haddock explaining the meaning of the type would be more than sufficient.

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Ratify.hs
```diff
@@ -223,6 +229,20 @@ ratifyTransition = do
             else pure st' {rsFuture = ast :<| rsFuture}
     Empty -> pure st
 
+-- | Check that there were no new enacted actions of the same type since this action was proposed
```
lehins: This comment is a bit misleading, because previous governance action id is insignificant during proposal, it is only important during ratification. At least for now, maybe in the future we'll have some enforcement at the proposal time in the GOV rule, but that is is a separate issue: #3655 
```suggestion
-- | Check that the previous governance action id specified in the proposal does match the last one of the same purpose that was enacted. 
```

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Enact.hs
lehins: This is fine, previous gov action id is only important for ratification, not enactment, so don't worry that it is ignored here.

eras/conway/impl/src/Cardano/Ledger/Conway/Governance.hs
```diff
@@ -203,34 +212,119 @@ instance EraPParams era => ToCBOR (GovActionsState era) where
...
+  encCBOR LastGovActions {..} =
+    encode $
+      Rec (LastGovActions @era)
+        !> To lgaPParamUpdate
+        !> To lgaHardFork
+        !> To lgaCommittee
+        !> To lgaConstitution
+
+toLastGovActionsParis :: (KeyValue a, Era era) => LastGovActions era -> [a]
+toLastGovActionsParis LastGovActions {..} =
```
lehins: Adopting the good practice :wink: 
```suggestion
toLastGovActionsParis lga@(LastGovActions _ _ _ _) =
  let LastGovActions {..} = lga
  in
```

lehins (changes requested):
Perfect with the respect to logic. Let's converge on the naming and this will be good to be merged

eras/conway/impl/src/Cardano/Ledger/Conway/Governance.hs
teodanciu: I agree,  happy with `PrevGovActionIds`, thanks :+1:

eras/conway/impl/src/Cardano/Ledger/Conway/Governance.hs
```diff
@@ -308,6 +405,7 @@ toEnactStatePairs cg@(EnactState _ _ _ _ _ _ _) =
       , "prevPParams" .= ensPParams
       , "treasury" .= ensTreasury
       , "withdrawals" .= ensWithdrawals
+      , "PrevGovActionIds" .= ensPrevGovActionIds
```
lehins: Throughout ledger instances we never use first capital letter in JSON fields
```suggestion
      , "prevGovActionIds" .= ensPrevGovActionIds
```

lehins (approved):
Perfect! Thank you!

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Ratify.hs
```diff
@@ -203,11 +205,15 @@ ratifyTransition = do
 
   case rsig of
     ast :<| sigs -> do
-      let GovActionState {gasAction, gasExpiresAfter} = ast
-      if spoAccepted env ast && dRepAccepted env ast dRepThreshold
+      let GovActionState {gasId, gasAction, gasExpiresAfter} = ast
+      if spoAccepted env ast
+        && dRepAccepted env ast dRepThreshold
+        && prevActionAsExpected gasAction (ensPrevGovActionIds rsEnactState)
```
lehins: This should be the first check since it is the fastest one to verify. This way we will short-circuit the if branch without redundantly looking through all the votes whenever prevActionId doesn't match.
```suggestion
      if prevActionAsExpected gasAction (ensPrevGovActionIds rsEnactState)
        && spoAccepted env ast
        && dRepAccepted env ast dRepThreshold
```

PR #3674, opened 2023-08-24
Added RATIFY thresholds

### LINKED ISSUE #3524: RATIFY thresholds
Implement the changes to RATIFY.

https://github.com/input-output-hk/formal-ledger-specifications/pull/134

Soupstraw:
# Description

Implemented ratification thresholds. I used a free applicative to assign the fields of `PParams` into groups. The applicative adds some type safety in ensuring that all the fields are handled while also making it easy to gather all the modified groups in a `PParamsUpdate`.

closes #3524

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/eras/conway/impl/CHANGELOG.md
+++ b/eras/conway/impl/CHANGELOG.md
@@ -2,6 +2,19 @@
 
 ## 1.8.0.0
 
+* Add:
+  * `PParamGroup`
+  * `ParamGrouper`
+  * `pGroup`
+  * `pUngrouped`
+  * `modifiedGroups`
+  * `dvtPPNetworkGroupL`
+  * `dvtPPGovGroupL`
+  * `dvtPPTechnicalGroupL`
+  * `dvtPPEconomicGroupL`
+  * `threshold`
+  * `ensCommitteeL`
+* Add `pparamsGroups` to `ConwayEraPParams`
 * Add `PrevGovActionIds`
 * Change `EnactState` to add `ensPrevGovActionIds`
 * Add  `ensPrevGovActionIdsL`, `ensPrevPParamUpdateL`, `ensPrevHardForkL` `ensPrevCommitteeL`, `ensPrevConstitutionL`
--- a/eras/conway/impl/src/Cardano/Ledger/Conway/Governance.hs
+++ b/eras/conway/impl/src/Cardano/Ledger/Conway/Governance.hs
@@ -2,7 +2,7 @@
 {-# LANGUAGE DeriveGeneric #-}
 {-# LANGUAGE DerivingStrategies #-}
 {-# LANGUAGE FlexibleContexts #-}
-{-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE NamedFieldPuns #-}
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE RecordWildCards #-}
@@ -41,6 +41,9 @@ module Cardano.Ledger.Conway.Governance (
   indexedGovProps,
   Constitution (..),
   ConwayEraGov (..),
+  thresholdSPO,
+  thresholdDRep,
+  thresholdCC,
   -- Lenses
   cgGovActionsStateL,
   cgEnactStateL,
@@ -48,6 +51,7 @@ module Cardano.Ledger.Conway.Governance (
   ensCommitteeL,
   ensConstitutionL,
   ensCurPParamsL,
+  ensPrevPParamsL,
   ensPrevGovActionIdsL,
   ensPrevPParamUpdateL,
   ensPrevHardForkL,
@@ -66,7 +70,12 @@ module Cardano.Ledger.Conway.Governance (
 ) where
 
 import Cardano.Ledger.Address (RewardAcnt)
-import Cardano.Ledger.BaseTypes (EpochNo (..), ProtVer (..), StrictMaybe)
+import Cardano.Ledger.BaseTypes (
+  EpochNo (..),
+  ProtVer (..),
+  StrictMaybe (..),
+  UnitInterval,
+ )
 import Cardano.Ledger.Binary (
   DecCBOR (..),
   EncCBOR (..),
@@ -83,6 +92,17 @@ import Cardano.Ledger.Binary.Coders (
  )
 import Cardano.Ledger.CertState (CommitteeState, DRepState)
 import Cardano.Ledger.Coin (Coin (..))
+import Cardano.Ledger.Conway.Core (
+  ConwayEraPParams,
+  PParamGroup (..),
+  dvtPPEconomicGroupL,
+  dvtPPGovGroupL,
+  dvtPPNetworkGroupL,
+  dvtPPTechnicalGroupL,
+  modifiedGroups,
+  ppDRepVotingThresholdsL,
+  ppPoolVotingThresholdsL,
+ )
 import Cardano.Ledger.Conway.Era (ConwayEra)
 import Cardano.Ledger.Conway.Governance.Procedures (
   Anchor (..),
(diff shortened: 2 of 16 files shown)


eras/conway/impl/src/Cardano/Ledger/Conway/Governance.hs
```diff
@@ -68,6 +74,7 @@ import Cardano.Ledger.Binary.Coders (
   (<!),
  )
 import Cardano.Ledger.Coin (Coin (..))
+import Cardano.Ledger.Conway.Core (ConwayEraPParams, PParamGroup (..), dvtPPEconomicGroupL, dvtPPGovGroupL, dvtPPNetworkGroupL, dvtPPTechnicalGroupL, modifiedGroups, ppDRepVotingThresholdsL, ppPoolVotingThresholdsL)
```
lehins: Maybe switch to multiline?

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Ratify.hs
```diff
@@ -89,16 +95,16 @@ instance
...
-  totalAcceptedStakePoolsRatio > getStakePoolThreshold gasAction
+spoAccepted ::
+  ConwayEraPParams era =>
+  RatifyState era ->
+  RatifyEnv era ->
+  GovActionState era ->
+  Bool
+spoAccepted rs RatifyEnv {reStakePoolDistr = PoolDistr poolDistr} gas =
+  case threshold rs gasAction SPO of
+    SJust r -> totalAcceptedStakePoolsRatio > unboundRational r
```
lehins: It should be `>=`, like in case of DReps. Also if it is `>` the zero threshold would not work. In fact, let's short circuit on zero to avoid redundant computation
```suggestion
    SJust r -> r == minBound || totalAcceptedStakePoolsRatio >= unboundRational r
```

eras/conway/impl/src/Cardano/Ledger/Conway/Governance.hs
```diff
@@ -413,3 +427,65 @@ class EraGov era => ConwayEraGov era where
...
+        Set.map lookupGroupThreshold $
+          modifiedGroups @era ppu
+
+threshold ::
+  ConwayEraPParams era =>
+  RatifyState era ->
+  GovAction era ->
+  GovRole ->
+  StrictMaybe UnitInterval
+threshold rSt action role =
```
lehins: Let's get rid of `GovRole` and split one `threashold` function into three separate functions:
* `thresholdDRep`. Instead of passing the whole of `PParams`, pass `DRepThreasholds` for this function.
* `thresholdCommittee`. 
* `thresholdStakePool`

It will be much cleaner than this huge monstrosity. Also, instead of passing the whole `RatifyState` we should just pass the `StrictMaybe Committee` instead.

eras/conway/impl/src/Cardano/Ledger/Conway/Governance.hs
```diff
@@ -413,3 +427,65 @@ class EraGov era => ConwayEraGov era where
...
+  let pp = rSt ^. rsEnactStateL . ensCurPParamsL
+      DRepVotingThresholds {..} = pp ^. ppDRepVotingThresholdsL
+      PoolVotingThresholds {..} = pp ^. ppPoolVotingThresholdsL
```
lehins: It is better to avoid RecordWildcards in complex functions like this, it is unclear where the binding comes from. Either switch to NamedFieldPuns or to regular field accessors.

RecordWildCards are justified, IMHO, whenever it is used in a small function with only one main argument, like To/FromJSON for example. Under more complex conditions  it should be avoided, because when reading the code it is unclear which binding belongs to which type (in current function `pvt` vs `dvt` prefixes help very little, not only because those abbreviations are not immediately obvious, but also because they look so similar).

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Ratify.hs
```diff
@@ -130,14 +136,12 @@ spoAccepted RatifyEnv {reStakePoolDistr = PoolDistr poolDistr} gas =
...
-      HardForkInitiation {} -> 101 % 100
-      _ -> spoThreshold
 
-dRepAccepted :: forall era. RatifyEnv era -> GovActionState era -> Rational -> Bool
-dRepAccepted ratifyEnv GovActionState {gasDRepVotes, gasAction} threshold =
-  dRepAcceptedRatio ratifyEnv gasDRepVotes gasAction >= threshold
+dRepAccepted :: forall era. ConwayEraPParams era => RatifyEnv era -> RatifyState era -> GovActionState era -> Bool
+dRepAccepted re rs GovActionState {gasDRepVotes, gasAction} =
+  case threshold rs gasAction DRep of
+    SJust r -> dRepAcceptedRatio re gasDRepVotes gasAction >= unboundRational r
```
lehins: Let's short circuit on zero, otherwise we'll end up counting DRep vote distribution for all governance action, even those that DReps should not be voting for.
```suggestion
    -- Short circuit on zero threshold in order to avoid redundant computation.
    SJust r -> r == minBound || dRepAcceptedRatio re gasDRepVotes gasAction >= unboundRational r
```

lehins (changes requested):
This is a clever trick with `Ap`. There is one minor issue with it, which I already mentioned during a meeting that implementation of `pparamsGroup` depends on the correct order of arguments, but we can solve it at a later point.

Overall looks great. Changes I suggested will be good in order to improve clarity and performance, despite that they slightly diverge from the spec.

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Examples/ConwayWithdrawalDelay.hs
lehins, line 1: My guess is that you though about adding some tests, but then changed your mind? :slightly_smiling_face: 

In any case, we should not add empty modules like that.

libs/cardano-ledger-pretty/src/Cardano/Ledger/Pretty/Conway.hs
```diff
@@ -7,17 +8,20 @@
 {-# LANGUAGE RankNTypes #-}
 {-# LANGUAGE RecordWildCards #-}
 {-# LANGUAGE ScopedTypeVariables #-}
+{-# HLINT ignore "Use record patterns" #-}
```
lehins: Let's get rid of it
```suggestion
```

libs/cardano-ledger-pretty/src/Cardano/Ledger/Pretty/Conway.hs
```diff
@@ -7,17 +8,20 @@
...
+{-# LANGUAGE TypeApplications #-}
 {-# LANGUAGE TypeOperators #-}
 {-# LANGUAGE UndecidableInstances #-}
 {-# OPTIONS_GHC -Wno-orphans #-}
 {-# OPTIONS_GHC -Wno-unrecognised-pragmas #-}
```
lehins: Which will allow us getting rid of extra ghc flags
```suggestion
```

eras/conway/impl/src/Cardano/Ledger/Conway/Governance.hs
```diff
@@ -492,3 +522,93 @@ class EraGov era => ConwayEraGov era where
 
 instance Crypto c => ConwayEraGov (ConwayEra c) where
   constitutionGovStateL = cgEnactStateL . ensConstitutionL
+
+data GovRole
+  = SPO
+  | DRep
+  | CC
```
lehins: This can be removed now
```suggestion
```

eras/conway/impl/CHANGELOG.md
```diff
@@ -2,6 +2,20 @@
...
+  * `PParamGroup`
+  * `ParamGrouper`
+  * `pGroup`
+  * `pUngrouped`
+  * `modifiedGroups`
+  * `dvtPPNetworkGroupL`
+  * `dvtPPGovGroupL`
+  * `dvtPPTechnicalGroupL`
+  * `dvtPPEconomicGroupL`
+  * `GovRole`
```
lehins: Unused
```suggestion
```

lehins (approved):
Few minor adjustments are still needed, but other than that it looks great. Thank you.

Issue #3672, opened 2023-08-23
Delegated stake to DReps is not counted as active


lehins:
Current implementation of incremental stake distribution calculation only counts stake as active when it is delegated to a stake pool:

https://github.com/input-output-hk/cardano-ledger/blob/96947e71746e0053fb14c8d5edf8cafc5d6e2295/libs/cardano-ledger-core/src/Cardano/Ledger/UMap.hs#L263-L269

Therefore, when stake is delegated only to a DRep, but not a stake pool it will not be counted as an active stake:

https://github.com/input-output-hk/cardano-ledger/blob/96947e71746e0053fb14c8d5edf8cafc5d6e2295/eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState/IncrementalStake.hs#L242-L244

For that reason it will currently not going to show up in the DRep stake distribution

lehins:
This was fixed in #3676 
Follow-up task to improve performance was created: #3702

PR #3676, opened 2023-08-24
Fix inactive PoolStake not counting as Drep Stake

TimSheard:
Changed freshDRepPulser to use map form IncrementalStake rather than the Mark SnapShot

addresses issue #3672

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState/Types.hs
+++ b/eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState/Types.hs
@@ -50,11 +50,11 @@ import Cardano.Ledger.CertState (
   vsDRepDistrL,
   vsDRepsL,
  )
-import Cardano.Ledger.Coin (Coin (..))
+import Cardano.Ledger.Coin (Coin (..), CompactForm)
 import Cardano.Ledger.Credential (Credential (..), Ptr (..))
 import Cardano.Ledger.Crypto (Crypto)
 import Cardano.Ledger.DRepDistr (DRepDistr (..), startDRepDistr)
-import Cardano.Ledger.EpochBoundary (SnapShots (..))
+import Cardano.Ledger.EpochBoundary (SnapShots (..), ssStakeDistrL, ssStakeMarkL)
 import Cardano.Ledger.Keys (
   KeyHash (..),
   KeyPair,
@@ -76,6 +76,7 @@ import Data.Default.Class (Default, def)
 import Data.Group (Group, invert)
 import Data.Map.Strict (Map)
 import qualified Data.Map.Strict as Map
+import Data.VMap (VB, VMap, VP)
 import qualified Data.VMap as VMap
 import GHC.Generics (Generic)
 import Lens.Micro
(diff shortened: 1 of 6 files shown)


eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState/Types.hs
```diff
@@ -719,8 +727,8 @@ epochStateDRepDistrL = esLStateL . lsCertStateL . certVStateL . vsDRepDistrL
 epochStateStakeDistrL ::
   Lens'
     (EpochState era)
-    (VMap VB VP (Credential 'Staking (EraCrypto era)) (CompactForm Coin))
-epochStateStakeDistrL = esSnapshotsL . ssStakeMarkL . ssStakeDistrL
+    (Map (Credential 'Staking (EraCrypto era)) Coin)
+epochStateStakeDistrL = esLStateL . lsUTxOStateL . utxosStakeDistrL . credMapL
```
lehins: This change requires an entry in a changelog and a bump in cardano-ledger-shelley to `1.6.0.0`. Also cardano-ledger-conway needs a lower dependency bump on `cardano-ledger-shelley ^>= 1.6`, while other packages that currently specify `cardano-ledger-shelley >= x && < 1.6` as a dependency, can be switched to `cardano-ledger-shelley >=x && < 1.7`

eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState/Types.hs
```diff
@@ -741,4 +749,4 @@ freshDRepPulser n es =
     n
     (es ^. epochStateUMapL)
     (es ^. epochStateRegDrepL)
-    (es ^. epochStateStakeDistrL)
+    (VMap.fromMap (compactCoinOrError <$> (es ^. epochStateStakeDistrL)))
```
lehins: This is an expensive operation, since iterates over the huge Map, which we don't want to do on the epoch boundary. So, we either want to switch IncrementalStake to use CompactCoin, or switch DRepDistr to use Coin.

FYI. We don't have to fix it in this PR. I will create an issue, so we can figure it out later.

lehins (changes requested):
Awesome fix. Thank you Tim

eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState/Types.hs
TimSheard: Could we just create a new Lens, with a different name. This lens is mentioned only twice. Where it is defined, and where it us used in  freshDRepPulser.

eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState/Types.hs
TimSheard: I think using CompactCoin in IncrementalStake would be best.

eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState/Types.hs
TimSheard: I did this and update the min version for Conway

eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState/Types.hs
TimSheard: Save this for a new PR.

lehins (approved):
Perfect

PR #3677, opened 2023-08-24
Minor patch that fixes the DRep distribution computation

lehins:
# Description

This is a minor patch that fixes the DRep distribution computation

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/eras/conway/impl/CHANGELOG.md
+++ b/eras/conway/impl/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Version history for `cardano-ledger-conway`
 
+## 1.7.1.0
+
+* Fix DRep distribution computation.
+
 ## 1.7.0.0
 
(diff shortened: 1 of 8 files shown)


eras/conway/impl/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Version history for `cardano-ledger-conway`
 
+## 1.7.1.0
```
teodanciu, line 3: So, given that the actual code change is in master, does this mean that this changelog entry will  for 1.7.1.0 will never be past of master?

lehins: Yes, whenever a release happens not from HEAD of master, there needs to be an adjustment to versions and changelog on master

PR #3679, opened 2023-08-25
Remove unused ubuntu packages to avoid running out of disk space

teodanciu:
when building nix derivations

# Description

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/.github/workflows/haskell.yml
+++ b/.github/workflows/haskell.yml
@@ -46,6 +46,25 @@ jobs:
         os: [ubuntu-latest]
 
     steps:
+
+      # Update the local package list in order to ensure that the action freeing up space will succeed
+    - name: "Update package list"
+      if: ${{ github.event_name == 'schedule' }}
+      run: |
+        sudo apt-get update
+      # there is not enough disk space to build nix derivations in the nightly build,
+      # so we're removing some unneccesary software from the ubuntu image
+    - name: Free disk space
+      if: ${{ github.event_name == 'schedule' }}
+      uses: jlumbroso/free-disk-space@main
+      with:
+        android: true
+        dotnet: true
+        haskell: false
+        large-packages: false # takes a long time to free not so much space, so it's disabled at the moment
+        docker-images: true
+        swap-storage: false # disabled, otherwise running out of memory during tests
+
     - uses: actions/checkout@v3
     - name: Set NIGHTLY environment variable if the job was triggered by the scheduler
       if: ${{ github.event_name == 'schedule' }}


.github/workflows/haskell.yml
```diff
@@ -46,6 +46,19 @@ jobs:
         os: [ubuntu-latest]
 
     steps:
+
+    - run: |
```
lehins: As discussed during the meeting let's add a condition for this to run only for nightly

lehins (approved):

.github/workflows/haskell.yml
```diff
@@ -46,6 +46,19 @@ jobs:
...
+        sudo apt-get update
+    - name: Free disk space
```
angerman, line 57: Can we please add a comment here that explains the rational for this change?

If I'm just reading it as it is, I'm really confused as to why we do this here and how we end up running out of diskspace.

I'm not questioning that we are. But we should capture the analysis as to why here as well.

angerman (changes requested):

.github/workflows/haskell.yml
teodanciu: Sure, I'll add a comment in the file.
Basically we can't build nix derivations (which we only do in nightly builds) because `error: writing to file: No space left on device` 

For example here: https://github.com/teodanciu/cardano-ledger/actions/runs/6033766318/job/16370965979#step:4:5647

Admittedly there might be better ways to deal with this, going at the root of the problem (maybe cleaning unnecessary things that we create ourselves, though these don't seem to be that big)

aniketd:
I rebased this since the jobs had failed on account of some downloads failing (most likely during the time GitHub services had degraded recently). Hopefully now these jobs will pass. :crossed_fingers:

.github/workflows/haskell.yml
lehins, line 1: Don't merge until https://github.com/input-output-hk/cardano-ledger/pull/3697 is merged

teodanciu: Also, the requested changes have to be approved before merging - are you happy with the changes @angerman ?

angerman:
So am I understanding this right, we are building `.#hydraJobs.required` on GHA? Don't we build that exact derivation on hydra anyway for each commit already? So then the jiggly job would only end up pulling all the prebuilt stuff from hydra?

Sorry, I'm a bit confused what we do here 😵‍💫

teodanciu:
> So am I understanding this right, we are building `.#hydraJobs.required` on GHA? Don't we build that exact derivation on hydra anyway for each commit already? So then the jiggly job would only end up pulling all the prebuilt stuff from hydra?
> 
> Sorry, I'm a bit confused what we do here 😵‍💫

No, it's good you're asking, because you're probably right that it doesn't make much sense.
So, the initial idea was this: because our CI doesn't use nix, it sometimes happened that after a change that successfully passed CI and got merged, we couldn't enter nix-shell (because the change had broken the nix build). 
So we just wanted to add something that runs a `nix-shell` (as it was back then before flakes).  This was after the original hydra got decommissioned. 
Then after the transition to flakes, this job has evolved a little bit, and indeed it makes sense what you're saying, it doesn't make sense to just pull prebuilt hydra stuff.  

Does it still make sense to check that `nix develop` works, what do you think ? 
We're building the jobs on hydra now as well, I see, so maybe we can make that check mandatory for merging. I guess if `nix develop` would not work, then that build would maybe break too?

angerman:
> > So am I understanding this right, we are building `.#hydraJobs.required` on GHA? Don't we build that exact derivation on hydra anyway for each commit already? So then the jiggly job would only end up pulling all the prebuilt stuff from hydra?
> > Sorry, I'm a bit confused what we do here 😵‍💫
> 
> No, it's good you're asking, because you're probably right that it doesn't make much sense. So, the initial idea was this: because our CI doesn't use nix, it sometimes happened that after a change that successfully passed CI and got merged, we couldn't enter nix-shell (because the change had broken the nix build). So we just wanted to add something that runs a `nix-shell` (as it was back then before flakes). This was after the original hydra got decommissioned. Then after the transition to flakes, this job has evolved a little bit, and indeed it makes sense what you're saying, it doesn't make sense to just pull prebuilt hydra stuff.
> 
> Does it still make sense to check that `nix develop` works, what do you think ? We're building the jobs on hydra now as well, I see, so maybe we can make that check mandatory for merging. I guess if `nix develop` would not work, then that build would maybe break too?

Right, so pretty much every PR already runs the `required` job on hydra (you can see the required job being built on this PR as well. If it includes the `devShell` from the flake in the required property, then it will always be built.

You may also consider dropping the shell logic altogether from hydra, and using e.g. https://github.com/input-output-hk/devx as a way to obtain a development shell with IOG libs, and tools. That one, by being independent of the repo, doesn't risk getting accidentally broken by changes to the repo. The downside is that you potentially have to rebuild a tiny bit more as it doesn't come with any hydra dependencies pre-cached. (Yet; this is on the roadmap but likely will take quite a while, and cabal-cache is a good substitute).

lehins:
> We're building the jobs on hydra now as well, I see, so maybe we can make that check mandatory for merging

That's not going to happen until I have full confidence in hydra. Throughout my time at IOG every CI setup (hydra, cicero, BuildKite) that was on anything other than Github Actions has always had problems, sometimes only intermittent, but still problems, which often blocked PRs from being merged, potentially for days.

I'd be totally happy to not doing any nix stuff on GithubActions and run it on hydra only. Important part is that we still do get results from Hydra, so if there is some issue, we'll notice it. It is just not going to block us from moving forward.

teodanciu:
I'm going to close this PR, since it became clear that this is a questionable solution to a seemingly non-existent problem.
When the time comes for nightly builds to matter, I will remove the nix derivation building from it.
Good to know about input-output/devx, will look into into, thanks @angerman .

PR #3680, opened 2023-08-25
Bump Plutus deps to 1.11

bezirg:
# Description

# Checklist

- [ ] Self-reviewed the diff


lehins (changes requested):
Whenever bounds in a cabal file are changed the patch version of the package needs to be increased, otherwise the package will not get released next time we'll be going through a release process. So, could you please bump the last number in versions for these packages:
* cardano-ledger-alonzo-test
* cardano-ledger-binary

lehins (approved):
This will break some of the plutus functionality in Conway, but the breakage only affects Conway and PlutusV3, so this is totally fine

PR #3684, opened 2023-08-28
Add proposal delaying, remove predicate failure from ENACT

### LINKED ISSUE #3675: Remove failures from enactment
`EnactTreasuryInsufficientFunds` predicate failure should be turned into  ratification check that causes this GovAction to be delayed. Failing on the epoch boundary is not an option.

https://github.com/input-output-hk/cardano-ledger/blob/96947e71746e0053fb14c8d5edf8cafc5d6e2295/eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Enact.hs#L65-L67

Soupstraw:
# Description

Removed the predicate failure from ENACT. Made the enactment of some actions delay all the other actions. Treasury withdrawals are now delayed when there is not enough ADA in the treasury.

I also changed the logic of `RATIFY` so that actions with invalid previous governance action ID get delayed rather than dropped.

closes #3675 

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/eras/conway/impl/CHANGELOG.md
+++ b/eras/conway/impl/CHANGELOG.md
@@ -2,7 +2,14 @@
 
 ## 1.8.0.0
 
+* Change `PredicateFailure (ConwayENACT era)` to `Void`
+* Remove `EnactPredFailure`
+* Change `PredicateFailure (ConwayEPOCH era)` to `Void`
+* Remove `ConwayEpochPredFailure`
+* Remove `EpochFailure` and `RatifyFailure` from `ConwayNewEpochPredFailure`
+* Change `PredicateFailure (ConwayRATIFY era)` to `Void`
 * Add:
+  * `rsDelayed`
   * `PParamGroup`
   * `ParamGrouper`
   * `pGroup`
--- a/eras/conway/impl/src/Cardano/Ledger/Conway/Governance.hs
+++ b/eras/conway/impl/src/Cardano/Ledger/Conway/Governance.hs
@@ -487,6 +487,7 @@ instance EraPParams era => NoThunks (EnactState era)
 data RatifyState era = RatifyState
   { rsEnactState :: !(EnactState era)
   , rsRemoved :: !(StrictSeq (GovActionState era))
+  , rsDelayed :: !Bool
   }
   deriving (Generic)
 
@@ -507,6 +508,7 @@ instance EraPParams era => DecCBOR (RatifyState era) where
       RecD RatifyState
         <! From
         <! From
+        <! From
 
 instance EraPParams era => EncCBOR (RatifyState era) where
   encCBOR RatifyState {..} =
@@ -514,6 +516,7 @@ instance EraPParams era => EncCBOR (RatifyState era) where
       Rec RatifyState
         !> To rsEnactState
         !> To rsRemoved
+        !> To rsDelayed
 
 instance EraPParams era => ToCBOR (RatifyState era) where
   toCBOR = toEraCBOR @era
@@ -530,10 +533,11 @@ instance EraPParams era => ToJSON (RatifyState era) where
   toEncoding = pairs . mconcat . toRatifyStatePairs
 
 toRatifyStatePairs :: (KeyValue a, EraPParams era) => RatifyState era -> [a]
-toRatifyStatePairs cg@(RatifyState _ _) =
+toRatifyStatePairs cg@(RatifyState _ _ _) =
   let RatifyState {..} = cg
    in [ "enactState" .= rsEnactState
       , "removed" .= rsRemoved
+      , "delayed" .= rsDelayed
       ]
 
 data ConwayGovState era = ConwayGovState
@@ -552,7 +556,7 @@ cgEnactStateL = lens cgEnactState (\x y -> x {cgEnactState = y})
 cgRatifyStateL :: Lens' (ConwayGovState era) (RatifyState era)
 cgRatifyStateL =
   lens
-    (\ConwayGovState {..} -> RatifyState cgEnactState mempty)
+    (\ConwayGovState {..} -> RatifyState cgEnactState mempty False)
     (\x RatifyState {..} -> x & cgEnactStateL .~ rsEnactState)
 
 curPParamsConwayGovStateL :: Lens' (ConwayGovState era) (PParams era)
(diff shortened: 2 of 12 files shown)


Soupstraw:
Oops, I just noticed @teodanciu already had an open PR implementing the delaying and `GovId` checking #3670

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Epoch.hs
```diff
@@ -340,9 +340,8 @@ instance
   , STS (ConwayRATIFY era)
   , BaseM (ConwayRATIFY era) ~ ShelleyBase
   , Event (ConwayRATIFY era) ~ Void
-  , PredicateFailure (EraRule "RATIFY" era) ~ EnactPredFailure era
   ) =>
   Embed (ConwayRATIFY era) (ConwayEPOCH era)
   where
-  wrapFailed = ConwayRatifyFailure
+  wrapFailed = absurd
```
lehins, line 329: I think we can remove this instance completely. We can do that by removing this predicate failure: https://github.com/input-output-hk/cardano-ledger/blob/900c2a47b7cca0c00ae5dfed3ffff1af6b201012/eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Epoch.hs#L108

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Ratify.hs
```diff
@@ -184,35 +195,83 @@ dRepAcceptedRatio RatifyEnv {reDRepDistr, reDRepState, reCurrentEpoch} gasDRepVo
...
+      StrictMaybe (PrevGovActionId r (EraCrypto era)) ->
+      Bool
+    check h prev = hpId (ens ^. h) == prev
+
+delayed ::
+  GovAction era ->
+  EnactState era ->
+  Bool ->
+  Bool
+delayed a es d = d || not (verifyPrev a es) || not (withdrawalCanWithdraw a)
```
lehins: There is too much negation going on here, let's not do that. It is much cleaner the way it is right now were we check the conjunction of all checks that a required for acceptance, rather than everything is needed for delaying.

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Ratify.hs
```diff
@@ -184,35 +195,83 @@ dRepAcceptedRatio RatifyEnv {reDRepDistr, reDRepState, reCurrentEpoch} gasDRepVo
...
+    HardForkInitiation prev _ -> check ensProtVerHPL prev
+    ParameterChange prev _ -> check ensCurPParamsHPL prev
+    TreasuryWithdrawals {} -> True
+    InfoAction {} -> True
+  where
+    check ::
+      Lens' (EnactState era) (HashProtected r a (EraCrypto era)) ->
+      StrictMaybe (PrevGovActionId r (EraCrypto era)) ->
+      Bool
+    check h prev = hpId (ens ^. h) == prev
```
lehins: You are really not gaining any clarity or reduction in amount of code by adding a helper `check` function like this :slightly_smiling_face: : 
```suggestion
    NoConfidence prev -> hpId (ens ^. ensCommitteeHPL) == prev
    NewCommittee prev _ _ -> hpId (ens ^. ensCommitteeHPL) == prev
    NewConstitution prev _ -> hpId (ens ^. ensConstitutionHPL) == prev
    HardForkInitiation prev _ -> hpId (ens ^. ensProtVerHPL) == prev
    ParameterChange prev _ -> hpId (ens ^. ensCurPParamsHPL) == prev
    TreasuryWithdrawals {} -> True
    InfoAction {} -> True
```

eras/conway/impl/src/Cardano/Ledger/Conway/Governance.hs
```diff
@@ -270,28 +282,89 @@ instance EraPParams era => ToCBOR (GovActionsState era) where
 instance EraPParams era => FromCBOR (GovActionsState era) where
   fromCBOR = fromEraCBOR @era
 
+data HashProtected r a c
+  = Unprotected !a
+  | Protected !a !(PrevGovActionId r c)
```
lehins: Currently spec assumes that there is always governance action id available, which is incorrect and will be fixed in the spec in the future
```suggestion
  | Protected !a !(StrictMaybe (PrevGovActionId r c))
```

lehins (changes requested):
I chose to go with @teodanciu's approach implemented in #3670 because it is correct, while this implementation is not quite. More importantly, however, the approach take in #3670 is much cleaner and is more flexible, because we will likely need to update prevGovActionId for PParamsUpdate during ratification. Updating the PParams during Enactment will have to happen an epoch later due to #3691. HashProtected approach in this PR prevents us from doing so.

In any case I found the HashProtected approach in the spec a bit ugly to begin with, because it is more convoluted that it has to be.

So, could you extract only the parts that are relevant for #3675 and either submit it as a new PR or modify this one, whichever you prefer.

eras/conway/impl/src/Cardano/Ledger/Conway/Governance.hs
Soupstraw: `Proteced` constructor is for the case where we have a previous govaction ID, while `Unprotected` is used when we don't have one.

Soupstraw: Maybe the naming is a bit confusing

Soupstraw: But it doesn't matter since I'm going to scrap that anyways

eras/conway/impl/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## 1.8.0.0
 
+* Change `Environment (ConwayENACT era)` to `GovActionId (EraCrypto era)`
```
aniketd: `docs/diagrams` in the other PR needs a corresponding update :slightly_smiling_face:

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Epoch.hs
Soupstraw: I think we still need it if we want to call the `RATIFY` rule from `EPOCH`

Soupstraw: but we can definitely get rid of `ConwayRatifyFailure`

eras/conway/impl/CHANGELOG.md
Soupstraw: I accidentally left this one in. The only change is the predicate failure.

lehins (approved):
Very nice!!!! :muscle: 

Love all the predicate failure cleanup

PR #3686, opened 2023-08-28
Post patch release fixup

lehins:
# Description

This PR fixes:

* changelogs due to recent release #3676
* versions that didn't get updated in #3635
* changelogs that had versions bumped twice in #3661

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/eras/conway/impl/CHANGELOG.md
+++ b/eras/conway/impl/CHANGELOG.md
@@ -49,6 +49,10 @@
 * Change type of `rsRemoved` in `RatifyState` to use  `GovActionState` instead of a tuple
 * Change `RatifySignal` to use `GovActionsState` instead of a tuple
 
+## 1.7.1.0
+
+* Fix DRep distribution computation.
+
 ## 1.7.0.0
 
 * Add `Network` validation for `ProposalProcedure` and `TreasuryWithdrawals` in GOV #3659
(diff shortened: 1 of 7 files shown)


teodanciu:
Hm, seems like after  `Make snapshots of GovActionsState` , a few more packages needed their version updated,  because of version change of their dependency, no? 
 
cardano-ledger-allegra  to 1.2.2.1 
cardano-ledger-alonzo  to 1.4.1.1  
cardano-ledger-alonzo-test  to 1.1.2.5  
cardano-ledger-babbage to 1.4.4.1
cardano-ledger-babbage-test to 1.1.1.6
conway-ledger-conway  to 1.8.0.1
cardano-ledger-pretty to 1.3.0.3
cardano-protocol-tpraos to 1.0.3.6

But the changes in this PR are still valid, so we can add these afterwards. I'm happy to make a PR with these if you like

teodanciu (approved):
Looks good to me

lehins:
> Hm, seems like after Make snapshots of GovActionsState , a few more packages needed their version updated, because of version change of their dependency, no?

@teodanciu That is not correct. All of these packages already had a version bump. In fact I noticed thanks to your comment that `cardano-ledger-pretty` had been already bumped twice (which is what your comment suggests to do for all of the packages in the list). The released version is `cardanoledger-pretty-1.3.0.0`, but version on master in cabal file on master is `cardanoledger-pretty-1.3.0.2`. We should not skip versions like that without actual releases. Also versions in CHNAGELOG should always match, except when there are no changes in a package when compared to the currently released version.

PR #3687, opened 2023-08-29
Update `NoThunks` instances to reflect thunks in `PState`and `CertState`

teodanciu:
# Description

Changing `utxosDeposited` in `UTxOState` to  be lazy is causing the  (nightly run) `NoThunks` test to fail. 
Simply changing the NoThunks implementation in `UTxOState` to allow thunks in `utxosDeposited`  is not enough though, it seems that we have to allow them for  `psDeposits`  in `PState`. 
I don't understand exactly why this is the case, given that `PState` is part of `CertState`, which  is defined alongside with `UTxOState` inside `LedgerState`, but I guess it is a consequence of how the rules and assertions are invoked, 

This PR simply makes the `NoThunks` test pass again, but I'm not convinced :
1)  Is there a better way to make the test pass (instead of allowing thunks in PState) 
2) If we have to allow thunks in PState, do we still want to have `utxosDeposited` strict? 

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/libs/cardano-ledger-core/src/Cardano/Ledger/CertState.hs
+++ b/libs/cardano-ledger-core/src/Cardano/Ledger/CertState.hs
@@ -1,5 +1,6 @@
 {-# LANGUAGE DataKinds #-}
 {-# LANGUAGE DeriveGeneric #-}
+{-# LANGUAGE DerivingVia #-}
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
 {-# LANGUAGE NamedFieldPuns #-}
@@ -99,7 +100,7 @@ import qualified Data.Map.Strict as Map
 import Data.Typeable
 import GHC.Generics (Generic)
 import Lens.Micro (Lens', lens, (^.), _1, _2)
-import NoThunks.Class (NoThunks (..))
+import NoThunks.Class (AllowThunksIn (..), NoThunks (..))
 
 -- ======================================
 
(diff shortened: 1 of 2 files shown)


libs/cardano-ledger-core/src/Cardano/Ledger/CertState.hs
```diff
@@ -246,8 +247,10 @@ data PState era = PState
   }
   deriving (Show, Eq, Generic)
 
-instance NoThunks (PState era)
-
+deriving via
+  AllowThunksIn '["psDeposits"] (PState era)
```
Soupstraw, line 251: Why is `psDeposits` allowed to have thunks?

teodanciu: I wrote some explanation in the description of the PR: 

```
Changing utxosDeposited in UTxOState to be lazy is causing the (nightly run) NoThunks test to fail.
Simply changing the NoThunks implementation in UTxOState to allow thunks in utxosDeposited is not enough though, it seems that we have to allow them for psDeposits in PState.
I don't understand exactly why this is the case, given that PState is part of CertState, which is defined alongside with UTxOState inside LedgerState, but I guess it is a consequence of how the rules and assertions are invoked,

This PR simply makes the NoThunks test pass again, but I'm not convinced :

Is there a better way to make the test pass (instead of allowing thunks in PState)
If we have to allow thunks in PState, do we still want to have utxosDeposited strict?

```

Any suggestion welcome !

lehins:
We should not allow thunks in the PState. Instead we need to resolve the cause of the issue, i.e. why nothunks is complaining about thunks in the PState due to lazy `utxosDeposited`

PR #3688, opened 2023-08-29
DRep Refunds and update evalTransactionBalance

### LINKED ISSUE #3645: DRep deposits and refunds

aniketd:
# Description

Closes #3645 

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/eras/conway/impl/src/Cardano/Ledger/Conway/UTxO.hs
+++ b/eras/conway/impl/src/Cardano/Ledger/Conway/UTxO.hs
@@ -1,7 +1,12 @@
+{-# LANGUAGE BangPatterns #-}
 {-# LANGUAGE DataKinds #-}
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE FlexibleInstances #-}
+{-# LANGUAGE LambdaCase #-}
+{-# LANGUAGE PatternSynonyms #-}
+{-# LANGUAGE TypeApplications #-}
 {-# LANGUAGE TypeFamilies #-}
+{-# LANGUAGE TypeOperators #-}
 {-# OPTIONS_GHC -Wno-orphans #-}
 
 module Cardano.Ledger.Conway.UTxO (
@@ -13,12 +18,16 @@ import Cardano.Ledger.Alonzo.UTxO (
   getAlonzoScriptsHashesNeeded,
   getAlonzoScriptsNeeded,
  )
+import Cardano.Ledger.Coin (Coin (Coin))
 import Cardano.Ledger.Conway.Core (
+  ConwayEraPParams,
   ConwayEraTxBody (..),
   Era (..),
   EraTxBody (..),
+  MaryEraTxBody,
   PParams,
   Value,
+  ppDRepDepositL,
  )
 import Cardano.Ledger.Conway.Era (ConwayEra)
 import Cardano.Ledger.Conway.Governance.Procedures (
@@ -27,15 +36,24 @@ import Cardano.Ledger.Conway.Governance.Procedures (
   VotingProcedures (..),
  )
 import Cardano.Ledger.Conway.TxBody ()
-import Cardano.Ledger.Credential (credScriptHash)
+import Cardano.Ledger.Conway.TxCert (
+  ConwayEraTxCert,
+  pattern RegDRepTxCert,
+  pattern UnRegDRepTxCert,
+ )
+import Cardano.Ledger.Credential (Credential, credScriptHash)
 import Cardano.Ledger.Crypto (Crypto)
 import Cardano.Ledger.Keys (KeyHash, KeyRole (..))
+import Cardano.Ledger.Mary (MaryValue)
 import Cardano.Ledger.Mary.UTxO (getConsumedMaryValue)
 import Cardano.Ledger.Shelley.UTxO (shelleyProducedValue)
 import Cardano.Ledger.UTxO (EraUTxO (..), UTxO)
 import Cardano.Ledger.Val (Val (..))
+import Data.Foldable (foldMap', foldl')
 import qualified Data.Map.Strict as Map
 import Data.Maybe (mapMaybe)
+import Data.Semigroup (getSum)
+import qualified Data.Set as Set
 import Lens.Micro ((^.))
 
 getConwayScriptsNeeded ::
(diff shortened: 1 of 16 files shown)


eras/conway/impl/src/Cardano/Ledger/Conway/UTxO.hs
```diff
@@ -57,22 +72,59 @@ getConwayScriptsNeeded utxo txb = getAlonzoScriptsNeeded utxo txb <> voterScript
...
+      let go accum@(!drepRegsInTx, !totalDeposit) = \case
+            RegDRepTxCert c _ _
+              -- Re-registration is a no-op and not considered
+              | Just _ <- lookupDRepDeposit c -> accum
+              | otherwise -> (Set.insert c drepRegsInTx, totalDeposit <+> pp ^. ppDRepDepositL)
+            UnRegDRepTxCert c _
+              -- DRep previously registered in the same tx.
+              | Set.member c drepRegsInTx -> (Set.delete c drepRegsInTx, totalDeposit <-> pp ^. ppDRepDepositL)
+            _ -> accum
+       in snd $ foldl' go (Set.empty, Coin 0) $ txb ^. certsTxBodyL
```
lehins: This is incorrect. You totally mixed it up. Produced is what leaves the transaction (outputs, fees and deposits). While consumed is what adds to the transaction (inputs, withdrawals and refunds)

In this PR you are doing the opposite.

Implement it by example in the same way the Stake Delegation deposits are implemented here: 

* for produced: https://github.com/input-output-hk/cardano-ledger/blob/d14ff95a51e7c229cc64a07579976d9a2bdd9078/eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState/RefundsAndDeposits.hs#L60
* for consumed: https://github.com/input-output-hk/cardano-ledger/blob/d14ff95a51e7c229cc64a07579976d9a2bdd9078/eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState/RefundsAndDeposits.hs#L118C1-L118C16

libs/cardano-ledger-core/src/Cardano/Ledger/UTxO.hs
```diff
@@ -218,6 +220,8 @@ class EraTxBody era => EraUTxO era where
     PParams era ->
     -- | Check whether a pool with a supplied PoolStakeId is already registered.
     (KeyHash 'StakePool (EraCrypto era) -> Bool) ->
+    -- | Function that can lookup current drep deposits
+    (Credential 'DRepRole (EraCrypto era) -> Maybe Coin) ->
```
lehins: Produced should not have this argument. This should have been an indicator that something is wrong, cause I've mentioned this before that only one of the two functions needs a new argument.
```suggestion
```

lehins (changes requested):
I realize why you mixed it up. Consumed and produced are a bit weirdly named and unclear concepts.

Anyways, this PR needs some serious fixing.

eras/conway/impl/src/Cardano/Ledger/Conway/UTxO.hs
aniketd: Thanks for the pointers! :bowing_man:

libs/cardano-ledger-core/src/Cardano/Ledger/UTxO.hs
aniketd: Sorry I didn't note that mention, and went down the wrong rabbit-hole on this one.

aniketd:
I hope I have got it correct this time :cold_sweat:

libs/cardano-ledger-api/test/Test/Cardano/Ledger/Api/Tx/Body.hs
```diff
@@ -141,7 +219,5 @@ spec =
       prop "evalBalanceTxBody" $ propEvalBalanceTxBody @Alonzo
     describe "BabbageEra" $ do
       prop "evalBalanceTxBody" $ propEvalBalanceTxBody @Babbage
-
--- TODO: bring back when TxCerts become a type family
--- describe "ConwayEra" $ do
---   prop "evalBalanceTxBody" $ propEvalBalanceTxBody @Conway
+    describe "ConwayEra" $ do
+      prop "evalBalanceTxBodyConway" $ propEvalBalanceTxBodyConway @Conway
```
lehins: Get rid of `evalBalanceTxBodyConway` and `propEvalBalanceTxBodyConway`. Copy pasting functionality in order to test that functionality matches the copy is not a test.

All you needed to do was uncomment the piece of code with a TODO

```suggestion
      prop "evalBalanceTxBody" $ propEvalBalanceTxBody @Conway
```

eras/shelley/impl/src/Cardano/Ledger/Shelley/UTxO.hs
```diff
@@ -163,11 +166,12 @@ shelleyProducedValue pp isRegPoolId txBody =
 getConsumedCoin ::
   ShelleyEraTxBody era =>
   PParams era ->
-  (StakeCredential (EraCrypto era) -> Maybe Coin) ->
+  (Credential 'Staking (EraCrypto era) -> Maybe Coin) ->
+  (Credential 'DRepRole (EraCrypto era) -> Maybe Coin) ->
```
lehins: This function is not using the argument, thus it should not be even accepting that argument! This is always wrong to write functions like that. It is at the use site 
 that it should ignore the argument
```suggestion
```

It is in the instance declaration that you should ignore the argument:
```haskell
  getConsumedValue pp lookupRefund _ = getConsumedCoin pp lookupRefund
```

eras/mary/impl/src/Cardano/Ledger/Mary/UTxO.hs
```diff
@@ -54,11 +55,12 @@ instance Crypto c => EraUTxO (MaryEra c) where
 getConsumedMaryValue ::
   (MaryEraTxBody era, Value era ~ MaryValue (EraCrypto era)) =>
   PParams era ->
-  (StakeCredential (EraCrypto era) -> Maybe Coin) ->
+  (Credential 'Staking (EraCrypto era) -> Maybe Coin) ->
+  (Credential 'DRepRole (EraCrypto era) -> Maybe Coin) ->
```
lehins: Same comment applies here.
```suggestion
```

lehins (changes requested):
Now that looks much better! :smile: 

Few minor things need adjustment. Also ghc-9.6 needs TyepOperators extension, see the CI failure

eras/shelley/impl/src/Cardano/Ledger/Shelley/UTxO.hs
aniketd: Got it! :raised_hands:

aniketd: This is actually even worse. I had `getConsumedMaryValue` being used inside `conwayConsumedValue` and didn't realise the lookup function need not be provided there!

libs/cardano-ledger-api/test/Test/Cardano/Ledger/Api/Tx/Body.hs
```diff
@@ -2,6 +2,7 @@
 {-# LANGUAGE DataKinds #-}
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE FlexibleInstances #-}
+{-# LANGUAGE GADTs #-}
```
aniketd: Removing this line.

libs/cardano-ledger-api/test/Test/Cardano/Ledger/Api/Tx/Body.hs
```diff
@@ -31,6 +32,7 @@ import qualified Data.Set as Set
 import Lens.Micro
 import Test.Cardano.Ledger.Babbage.Arbitrary ()
 import Test.Cardano.Ledger.Common
+import Test.Cardano.Ledger.Conway.Arbitrary ()
```
aniketd: Removing this line too.

lehins (approved):
Looks great. Thank you!

PR #3690, opened 2023-08-30
Preserve the order of ProposalProcedures

### LINKED ISSUE #3641: Preserve the order of ProposalProcedures
Governance actions must be enacted in the same order, in which they were proposed. For this we need an extra sequence with `GovActionId`s in `GovActionsState`

Soupstraw:
# Description

closes #3641 

# Checklist

- [x] Self-reviewed the diff


eras/conway/impl/src/Cardano/Ledger/Conway/TxBody.hs
```diff
@@ -579,3 +580,16 @@ instance ConwayEraTxBody era => EncCBOR (ConwayTxBodyRaw era) where
 
 -- | Encodes memoized bytes created upon construction.
 instance Era era => EncCBOR (ConwayTxBody era)
+
+class BabbageEraTxBody era => ConwayEraTxBody era where
```
lehins, line 585: Looks like you need to add `UndecidableSuperClasses`

eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Snapshots.hs
```diff
@@ -0,0 +1,147 @@
...
+import Data.Map.Strict (Map)
+import Data.Sequence.Strict (StrictSeq (..))
+import GHC.Generics (Generic)
+import Lens.Micro (Lens', (%~))
+import NoThunks.Class (NoThunks)
+
+data ProposalsSnapshot era = ProposalsSnapshot
+  { psMap :: !(Map (GovActionId (EraCrypto era)) (GovActionState era))
+  , -- Newer actions are near the end
+    psSeq :: !(StrictSeq (GovActionId (EraCrypto era)))
```
lehins: It is bad practice to name record fields with data structures that they use, cause we might change it in the future. It should have some sort of semantic meaning. Something like:
```suggestion
  { psGovActionStates :: !(Map (GovActionId (EraCrypto era)) (GovActionState era))
  , -- | Newer actions are near the end
    psProposalOrder :: !(StrictSeq (GovActionId (EraCrypto era)))
```

Also it is good to turn comments like this into haddock

eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Snapshots.hs
```diff
@@ -0,0 +1,147 @@
...
+
+snapshotActions ::
+  ProposalsSnapshot era ->
+  StrictSeq (GovActionState era)
+snapshotActions ProposalsSnapshot {..} = helper psSeq
+  where
+    helper (x :<| xs) =
+      case Map.lookup x psMap of
+        Just v -> v :<| helper xs
+        Nothing -> error "ProposalsSnapshot invariant is not maintained"
```
lehins: Mark it as impossible and in case if it does happen, showing the offender is a good idea for debugging:
```suggestion
        Nothing -> 
          error $ "Impossible: ProposalsSnapshot invariant is not maintained: "
                <> Text.unpack (govActionIdToText gaId)
```

eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Snapshots.hs
```diff
@@ -0,0 +1,147 @@
...
+    }
+
+snapshotActions ::
+  ProposalsSnapshot era ->
+  StrictSeq (GovActionState era)
+snapshotActions ProposalsSnapshot {..} = helper psSeq
+  where
+    helper (x :<| xs) =
+      case Map.lookup x psMap of
+        Just v -> v :<| helper xs
```
lehins: It is common to name such function either `loop` or `go`. Also, not necessary, but a slight improvement to naming of bindings, so it is clear what we are looping over
```suggestion
    go (gaId :<| gaIds) =
      case Map.lookup x psMap of
        Just v -> gaId :<| go gaIds
```

eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Snapshots.hs
```diff
@@ -0,0 +1,147 @@
...
+snapshotIds ::
+  ProposalsSnapshot era ->
+  StrictSeq (GovActionId (EraCrypto era))
+snapshotIds = psSeq
+
+snapshotActionsUnordered ::
+  ProposalsSnapshot era ->
+  [GovActionState era]
+snapshotActionsUnordered ProposalsSnapshot {..} =
+  Map.elems psMap
```
lehins: It is ok to return a Map. The only thing we use it for is to fold over, which can be done over a Map. Also, we never know it might come in handy.
```suggestion
snapshotGovActionStates ::
  ProposalsSnapshot era ->
  Map (GovActionId GovActionState era]
snapshotGovActionStates = psMap
```

eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Snapshots.hs
```diff
@@ -0,0 +1,147 @@
...
+
+snapshotInsertGovAction ::
+  GovActionState era ->
+  ProposalsSnapshot era ->
+  ProposalsSnapshot era
+snapshotInsertGovAction gas@GovActionState {..} ProposalsSnapshot {..} =
+  ProposalsSnapshot
+    { psMap = Map.insert gasId gas psMap
+    , psSeq = psSeq :|> gasId
+    }
```
lehins: Whenever you can use NamedFIeldPuns, instead of RecordWildCards.

Also this function isn't quite correct, because it allows duplicate govActionIds in the sequence. This is not something that can happen in our system, if we implement it correctly, but nonetheless we should account for it somehow
```suggestion
snapshotInsertGovAction gas@GovActionState {gasId} ps@ProposalsSnapshot {..}
  | Map.member gasId psMap = 
    ps { psMap = Map.insert gasId gas psMap }
  | otherwise = 
    ProposalsSnapshot
      { psMap = Map.insert gasId gas psMap
      , psSeq = psSeq :|> gasId
      }
```

eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Snapshots.hs
```diff
@@ -0,0 +1,147 @@
...
+
+snapshotRemoveId ::
+  GovActionId (EraCrypto era) ->
+  ProposalsSnapshot era ->
+  ProposalsSnapshot era
+snapshotRemoveId gId (ProposalsSnapshot {..}) =
+  ProposalsSnapshot
+    { psMap = Map.delete gId psMap
+    , psSeq = foldl' (\s x -> if x == gId then s else x :<| s) mempty psSeq
+    }
```
lehins: This approach in the end results in (O(n*m)) operations, which is not good since it is quadratic.

It is better to convert it to plural removal.

```suggestion
snapshotRemoveIds ::
  Set (GovActionId (EraCrypto era)) ->
  ProposalsSnapshot era ->
  ProposalsSnapshot era
snapshotRemoveIds gIds (ProposalsSnapshot {..}) =
  ProposalsSnapshot
    { psMap = psMap `Map.withoutKeys` gIds
    , psSeq = foldl' (\s x -> if x `Set.member` gIds then s else x :<| s) mempty psSeq
    }
```

eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Snapshots.hs
```diff
@@ -0,0 +1,147 @@
...
+-- ID of each `GovActionState` in the sequence must be unique or otherwise the
+-- Snapshot will be malformed!
+unsafeFromGovActionStateSeq ::
+  StrictSeq (GovActionState era) ->
+  ProposalsSnapshot era
+unsafeFromGovActionStateSeq (x :<| xs) =
+  snapshotInsertGovAction
+    x
+    (unsafeFromGovActionStateSeq xs)
+unsafeFromGovActionStateSeq Empty = def
```
lehins: It was `snapshotInsertGovAction` that was unsafe. Because of my suggestion in  `snapshotInsertGovAction` that fixes it, this can be changed to:
```suggestion
fromGovActionStateSeq ::
  StrictSeq (GovActionState era) ->
  ProposalsSnapshot era
fromGovActionStateSeq (x :<| xs) =
  snapshotInsertGovAction
    x
    (unsafeFromGovActionStateSeq xs)
fromGovActionStateSeq Empty = def
```

eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Snapshots.hs
```diff
@@ -0,0 +1,147 @@
...
+
+-- | Internal function for checking if the invariants are maintained
+isConsistent_ :: ProposalsSnapshot era -> Bool
+isConsistent_ (ProposalsSnapshot {..}) =
+  sort (Map.keys psMap) == sort (toList psSeq)
```
lehins: `keys` already produces a sorted list.
```suggestion
  Map.keys psMap == sort (toList psSeq)
```

eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Snapshots.hs
```diff
@@ -0,0 +1,147 @@
...
+    && all (\(k, GovActionState {..}) -> k == gasId) (Map.toList psMap)
```
lehins: Please try not to abuse REcordWildCards so much
```suggestion
    && all (\(k, GovActionState {gasId}) -> k == gasId) (Map.toList psMap)
```

eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Snapshots.hs
```diff
@@ -0,0 +1,147 @@
...
+  ProposalsSnapshot era
+unsafeFromGovActionStateSeq (x :<| xs) =
+  snapshotInsertGovAction
+    x
+    (unsafeFromGovActionStateSeq xs)
+unsafeFromGovActionStateSeq Empty = def
+
+-- | Internal function for checking if the invariants are maintained
+isConsistent_ :: ProposalsSnapshot era -> Bool
+isConsistent_ (ProposalsSnapshot {..}) =
```
lehins: Same here
```suggestion
isConsistent_ (ProposalsSnapshot {psMap, psSeq}) =
```

lehins (changes requested):
I haven't fully finished reviewing it. 
I'll do a second pass on Monday

eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Snapshots.hs
```diff
@@ -0,0 +1,153 @@
...
+  ProposalsSnapshot era
+snapshotRemoveIds gIds (ProposalsSnapshot {..}) =
+  ProposalsSnapshot
+    { psGovActionStates = psGovActionStates `Map.withoutKeys` gIds
+    , psProposalOrder = foldl' (\s x -> if x `Set.member` gIds then s else x :<| s) mempty psProposalOrder
+    }
+
+-- | Converts a sequence of `GovActionState`s to a `ProposalsSnapshot`. NB: The
+-- ID of each `GovActionState` in the sequence must be unique or otherwise the
+-- Snapshot will be malformed!
```
lehins: It will no longer be malformed, duplicates will just get dropped.
```suggestion
-- | Converts a sequence of `GovActionState`s to a `ProposalsSnapshot`.
-- 
-- /Warning/ - This function expects `GovActionState`'s to have unique 
-- `GovActionId`s, because duplicate Ids will result in `GovActionStates` 
-- to be dropped.
```

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Epoch.hs
```diff
@@ -263,21 +263,20 @@ epochTransition = do
...
-    newGovActionsState =
-      foldr'
-        (Map.delete . gasId)
-        oldGovActionsState
-        rsRemoved
+    oldGovSnapshots =
+      es ^. esGovernanceL . cgGovSnapshotsL . curGovSnapshotsL
+    newGovSnapshots =
+      snapshotRemoveIds
+        (foldl' (\s x -> Set.insert (gasId x) s) mempty rsRemoved)
```
lehins: Potential easy optimization:

I think it would make sense to switch `rsRemoved` to have the type:
```haskell
rsRemoved :: !(Set (GovActionId (EraCrypto era)))
```
Instead of doing this conversion here. We don't really care in which order GovActions get removed.

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Gov.hs
```diff
@@ -208,11 +190,11 @@ addAction epoch gaExpiry gaid c addr act as =
         }
 
 noSuchGovActions ::
-  GovActionsState era ->
+  GovSnapshots era ->
   Set.Set (GovActionId (EraCrypto era)) ->
   Test (ConwayGovPredFailure era)
 noSuchGovActions gas gaIds =
-  let unknownGovActionIds = Set.filter (`Map.notMember` curGovActionsState gas) gaIds
+  let unknownGovActionIds = foldl' (flip Set.delete) gaIds (snapshotIds $ curGovSnapshots gas)
```
lehins: This is less efficient, because you iterate over the whole sequence of current governance actions, instead of a small number of proposed governance actions. This is optimal:
```suggestion
  let curGovActionIds = snapshotGovActionStates $ curGovSnapshots gas
      unknownGovActionIds = Set.filter (`Map.notMember` curGovActionIds) gaIds
```

Difference in complexity is `n * log m` vs `m * log n`, where `n` (number of current gov actions) is known to be always much bigger than `m` (number of proposed gov actions)

libs/cardano-ledger-pretty/cardano-ledger-pretty.cabal
```diff
@@ -1,6 +1,6 @@
 cabal-version:      3.0
 name:               cardano-ledger-pretty
-version:            1.3.1.0
+version:            1.3.2.0
```
lehins: `1.3.1.0` hasn't been released yet. You can learn this info from the CHANGELOG (latest version in it matches the current one), from git tag or from CHaP
```suggestion
version:            1.3.1.0
```

eras/conway/impl/CHANGELOG.md
```diff
@@ -11,6 +11,25 @@
 * Remove `ConwayEpochPredFailure`
 * Remove `EpochFailure` and `RatifyFailure` from `ConwayNewEpochPredFailure`
 * Change `PredicateFailure (ConwayRATIFY era)` to `Void`
+* Move `ConwayEraTxBody` to `Cardano.Ledger.Conway.TxBody`
+* Move `ConwayEraPParams` to `Cardano.Ledger.Conway.PParams` together with
```
lehins: Did you try to add something at the end? Together with something?
```suggestion
* Move `ConwayEraPParams` to `Cardano.Ledger.Conway.PParams`
```

eras/conway/impl/src/Cardano/Ledger/Conway/Core.hs
```diff
@@ -1,43 +1,18 @@
...
-  ppDRepDepositL,
-  ppDRepActivityL,
-  ppuPoolVotingThresholdsL,
-  ppuDRepVotingThresholdsL,
-  ppuMinCommitteeSizeL,
-  ppuCommitteeTermLimitL,
-  ppuGovActionExpirationL,
-  ppuGovActionDepositL,
-  ppuDRepDepositL,
-  ppuDRepActivityL,
```
lehins, line 40: Can you try re-export all of these? Since dependency direction got flipped, we should be able to import all these guys into this module.
This will reduce breakage downstream and continue providing ability for us to export all core types from `.Core` module. At least until we are ready to change it up a bit.

eras/conway/test-suite/test/Test/Cardano/Ledger/Conway/GovSnapshot.hs
lehins, line 1: Is there any chance you could move these tests as hspec tests into `cardano-ledger-conway:tests` as well? Just as in #3698

This will have a new module name `Test.cardano.Ledger.GovSnapshotSpec`

lehins (approved):
This looks good. 

General feedback:

* Nice job on creating a safe interface for GovSnapshots :muscle: 
* It would have been much easier to review the PR if all the non-semantic changes like code moving (`ConwayEraTxBody`, `Params stuff`, etc) was done in a separate commit. I know it requires work and sometimes such change hard to predict and separate from the reset of the work. I am just saying that whenever possible we should try and do that. No need to split this PR into multiple commits.

eras/conway/impl/src/Cardano/Ledger/Conway/Core.hs
Soupstraw: the PParams and UTxO module still depend on Core

lehins: :cry: That's alright. Not the end of the world, especially since we'll be getting rid of those Core modules in the future anyways.

eras/conway/test-suite/test/Test/Cardano/Ledger/Conway/GovSnapshot.hs
Soupstraw: I'll not do it this time since we might actually have to remove the tests anyways after the optimization is done.

PR #3697, opened 2023-09-01
Rewrite testEql using Typeable to make it impossible to forget cases

MaximilianAlgehed:
# Description

Everything is `Typeable` so we can use that to reduce some clutter.

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/TypeRep.hs
+++ b/libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/TypeRep.hs
@@ -102,7 +102,8 @@ import qualified Data.Map.Strict as Map
 import Data.Maybe
 import Data.Set (Set)
 import qualified Data.Set as Set
-import Data.Universe (Eql, Shape (..), Shaped (..), Singleton (..), cmpIndex, (:~:) (Refl))
+import Data.Typeable
+import Data.Universe (Eql, Shape (..), Shaped (..), Singleton (..), cmpIndex)
 import Data.Word (Word16, Word64)
 import Formatting (formatToString)
 import Lens.Micro
@@ -148,7 +149,6 @@ import Test.Cardano.Ledger.Generic.Functions (protocolVersion)
 import Test.Cardano.Ledger.Generic.PrettyCore (
   credSummary,
   keyHashSummary,
-  pcAddr,
   pcCoin,
   pcConwayTxCert,
   pcDRep,
@@ -182,6 +182,9 @@ import Test.Cardano.Ledger.Shelley.Serialisation.Generators ()
 import Test.Cardano.Ledger.ShelleyMA.Serialisation.Generators ()
 import Test.QuickCheck hiding (Fixed, total)
 
+data IsTypeable a where
+  IsTypeable :: Typeable a => IsTypeable a
+
 -- =======================================================================
 -- Special functions for dealing with SoftForks properties that
 -- depend upon the ProtVer. Whiach can be computed from a (Proof era)
@@ -203,6 +206,7 @@ data Rep era t where
   MapR :: Ord a => Rep era a -> Rep era b -> Rep era (Map a b)
   SetR :: Ord a => Rep era a -> Rep era (Set a)
   ListR :: Rep era a -> Rep era [a]
+  AddrR :: Rep era (Addr (EraCrypto era))
   CredR :: Rep era (Credential 'Staking (EraCrypto era))
   VCredR :: Rep era (Credential 'DRepRole (EraCrypto era))
   PoolHashR :: Rep era (KeyHash 'StakePool (EraCrypto era))
@@ -259,7 +263,6 @@ data Rep era t where
   ExUnitsR :: Rep era ExUnits
   TagR :: Rep era Tag
   DataHashR :: Rep era (DataHash (EraCrypto era))
-  AddrR :: Rep era (Addr (EraCrypto era))
   PCredR :: Rep era (Credential 'Payment (EraCrypto era))
   ShelleyTxCertR :: Rep era (ShelleyTxCert era)
   ConwayTxCertR :: Rep era (ConwayTxCert era)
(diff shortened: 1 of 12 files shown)


MaximilianAlgehed:
@lehins and @TimSheard it would be good to get this reviewed and merged.

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/TypeRep.hs
```diff
@@ -294,128 +297,103 @@ stringR = ListR CharR
...
+  GenDelegPairR{}        -> IsTypeable
+  FutureGenDelegR{}      -> IsTypeable
+  PPUPStateR{}           -> IsTypeable
+  PtrR{}                 -> IsTypeable
+  IPoolStakeR{}          -> IsTypeable
+  SnapShotsR{}           -> IsTypeable
+  RewardR{}              -> IsTypeable
+  SlotNoR{}              -> IsTypeable
+  SizeR{}                -> IsTypeable
+  (repTypeable -> IsTypeable) :-> (repTypeable -> IsTypeable) -> IsTypeable
```
Soupstraw, line 384: I don't think I've seen this kind of pattern matching before, what does it do?

MaximilianAlgehed: This matches the constructor and ignores any arguments that would be bound in the pattern. So `Just{}` is equivalent to `Just _`.

Soupstraw: Ah, I mean this part of the pattern match `(repTypeable -> IsTypeable)`

MaximilianAlgehed: That's a view pattern, it means roughly `Just (negate -> a) = e ~ Just _x = let a = negate _x in e`

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/TypeRep.hs
```diff
@@ -735,8 +716,8 @@ instance Shaped (Rep era) any where
   shape DRepStateR = Nullary 87
   shape CommitteeStateR = Nullary 87
 
-compareRep :: forall era t s. Rep era t -> Rep era s -> Ordering
-compareRep = cmpIndex @(Rep era)
+compareRep :: forall era t s. Era era => Rep era t -> Rep era s -> Ordering
+compareRep x y = cmpIndex @(Rep era) x y
```
lehins: What's the point of eta expanding it?
```suggestion
compareRep = cmpIndex @(Rep era)
```

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/TypeRep.hs
```diff
@@ -294,128 +297,106 @@ stringR = ListR CharR
...
-  testEql (LedgerStateR c) (LedgerStateR d) =
-    do Refl <- testEql c d; pure Refl
-  testEql StakeHashR StakeHashR = Just Refl
-  testEql BoolR BoolR = Just Refl
-  testEql DRepR DRepR = Just Refl
-  testEql (PoolMetadataR c) (PoolMetadataR d) =
-    do Refl <- testEql c d; pure Refl
-  testEql _ _ = Nothing
+repTypeable :: Era era => Rep era t -> IsTypeable t
+repTypeable r = case r of
```
lehins, line 301: Very nice trick! I love it. No more wildcard matches in pattern matching for mismatched types.

lehins (approved):
This is awesome. I love how `IsTypeable` solves this source of mistakes when adding new Rep constructors: https://github.com/input-output-hk/cardano-ledger/blob/9e2f8151e3b9a0dde9faeb29a7dd2456e854427c/libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/TypeRep.hs#L417

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/TypeRep.hs
MaximilianAlgehed: Indeed. I also have a suspicion that this can help elsewhere too but I haven't look at it in detail yet. That'll be a separate PR when I get to it.

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/TypeRep.hs
MaximilianAlgehed: I think this was just something that slipped between the cracks during a rebase. I've removed it and pushed. Nicely spotted.

MaximilianAlgehed:
@lehins could you please run the checks? I think this is the final version of the PR.

lehins:
@MaximilianAlgehed This PR needs to be rebases on master. Whenever you see this at the bottom of the PR this is necessary for merging:

> This branch is out-of-date with the base branch

MaximilianAlgehed:
@lehins done

MaximilianAlgehed:
@lehins something appears to be broken in CI.

lehins:
@MaximilianAlgehed Looks like some issue with Github since yesterday. Traballs don't get downloaded fully or corrupted. This happens on multiple repos that we have.
```
gzip: stdin: unexpected end of file
tar: Unexpected EOF in archive
tar: Unexpected EOF in archive
```
image

lehins:
@MaximilianAlgehed Looks like Github has finally back to normal operation since yesterday. Just in case I implemented a fallback workaround too in #3713

Do you mind rebasing again on master again? We've had a few PRs merged yesterday.

MaximilianAlgehed:
@lehins still broken?

lehins:
> @lehins still broken?

@MaximilianAlgehed Why do you say it is broken? CI hasn't finished yet. (Only "Required" ones are important)

MaximilianAlgehed:
Right... So the one thats called `ci` is irrelevant to CI?

lehins:
> Right... So the one thats called `ci` is irrelevant to CI?

It is relevant, but it just doesn't work for forked repos, I think. In any case it is not terribly reliable, but it is useful, since it runs CI in nix, where GithubActions is build with regular cabal setup without nix.

MaximilianAlgehed:
Perfect, thanks for clearing that up. I won't care about that one in the
future when working from our fork.

Den tors 7 sep. 2023 12:41Alexey Kuleshevich ***@***.***>
skrev:

Issue #3623, opened 2023-08-08
Consolidate all `cardano-ledger-[era]-test` package


lehins:
Every test package that is attached to each era no longer needs to be a separate package, since we have public `testlib` sub-libraries for exporting test functionality to downstream packages. This means test suites can migrate as well into the `cardano-ledger-[era]` package, which will reduce the release and maintenance burden and make it more organized.

The final goal with respect to tests is to:
* migrate all of the era specific tests into each era package
* abstract as much of era agnostic reusable test functionality into `cardano-ledger-core:testlib` and reuse it in all test suites
* switch all tests to use hspec, instead of tasty
* switch any left over tests from using hedgehog to using QuickCheck

This is a broad ticket that will require quite a bit of work and organization. It is an umbrella issue for the long term goal of organizing and improving our tests.

As far as this ticket is concerned, it is important to note that none of this applies to:
* Constraint based test framework
* most parts of `cardano-ledger-test` in general
* Byron era codebase.

PR #3698, opened 2023-09-01
Reordering of governance actions

### LINKED ISSUE #3692: Re-order GovActions during Ratification
Order of Governance Actions that are being considered for ratification should be ordered by importance: https://github.com/cardano-foundation/CIPs/tree/master/CIP-1694#enactment

Important to note that sorting should be stable, meaning that relative order of governance actions that are of the same importance must stay the same as when they were proposed.

Soupstraw:
# Description

Reorder governance actions based on the action type.

closes #3692

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/eras/conway/impl/CHANGELOG.md
+++ b/eras/conway/impl/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## 1.9.0.0
 
+* Add `reorderActions`, `actionPriority`
 * Remove `ensProtVer` field from `EnactState`: #3705
 * Move `ConwayEraTxBody` to `Cardano.Ledger.Conway.TxBody`
 * Move `ConwayEraPParams` to `Cardano.Ledger.Conway.PParams`
@@ -24,6 +25,17 @@
   * `gasStakePoolVotesL`
 * Add `FromJSON` instance for `Committee`
 * Add `constitution` and `committee` fields to `ConwayGenesis`
+
+### testlib
+
+* Add `genNewCommittee`
+* Add `genNoConfidence`
+* Add `genTreasuryWithdrawals`
+* Add `genHardForkInitiation`
+* Add `genParameterChange`
+* Add `genNewConstitution`
+* Add `genGovActionStateFromAction`
+* Add `govActionGenerators`
 
 ## 1.8.0.0
 
--- a/eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Ratify.hs
+++ b/eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Ratify.hs
@@ -20,6 +20,8 @@ module Cardano.Ledger.Conway.Rules.Ratify (
   RatifySignal (..),
   dRepAccepted,
   dRepAcceptedRatio,
+  reorderActions,
+  actionPriority,
 ) where
 
 import Cardano.Ledger.BaseTypes (BoundedRational (..), ShelleyBase, StrictMaybe (..))
@@ -51,12 +53,15 @@ import Control.State.Transition.Extended (
   judgmentContext,
   trans,
  )
+import Data.Foldable (Foldable (..))
+import Data.List (sortOn)
 import Data.Map.Strict (Map)
 import qualified Data.Map.Strict as Map
 import Data.Maybe (fromMaybe)
 import Data.Monoid (Sum (..))
 import Data.Ratio ((%))
 import Data.Sequence.Strict (StrictSeq (..))
+import qualified Data.Sequence.Strict as Seq
 import qualified Data.Set as Set
 import Data.Void (Void, absurd)
 import Data.Word (Word64)
@@ -202,6 +207,18 @@ delayingAction TreasuryWithdrawals {} = False
 delayingAction ParameterChange {} = False
 delayingAction InfoAction {} = False
 
+actionPriority :: GovAction era -> Int
+actionPriority NoConfidence {} = 0
+actionPriority NewCommittee {} = 1
+actionPriority NewConstitution {} = 2
+actionPriority HardForkInitiation {} = 3
+actionPriority ParameterChange {} = 4
+actionPriority TreasuryWithdrawals {} = 5
+actionPriority InfoAction {} = 6
+
+reorderActions :: StrictSeq (GovActionState era) -> StrictSeq (GovActionState era)
+reorderActions = Seq.fromList . sortOn (actionPriority . gasAction) . toList
+
 ratifyTransition ::
   forall era.
   ( Embed (EraRule "ENACT" era) (ConwayRATIFY era)
@@ -218,7 +235,8 @@ ratifyTransition = do
       , RatifySignal rsig
       ) <-
     judgmentContext
-  case rsig of
+  let reorderedActions = reorderActions rsig
+  case reorderedActions of
     ast :<| sigs -> do
       let gas@GovActionState {gasId, gasAction, gasExpiresAfter} = ast
           withdrawalCanWithdraw (TreasuryWithdrawals m) =
--- a/eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Arbitrary.hs
+++ b/eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Arbitrary.hs
@@ -7,7 +7,17 @@
 {-# LANGUAGE UndecidableInstances #-}
 {-# OPTIONS_GHC -Wno-orphans #-}
 
-module Test.Cardano.Ledger.Conway.Arbitrary (uniqueIdGovActions) where
+module Test.Cardano.Ledger.Conway.Arbitrary (
+  genNewCommittee,
+  genNoConfidence,
+  genTreasuryWithdrawals,
+  genHardForkInitiation,
+  genParameterChange,
+  genNewConstitution,
+  genGovActionStateFromAction,
+  govActionGenerators,
+  uniqueIdGovActions,
+) where
 
 import Cardano.Ledger.Alonzo.Scripts (AlonzoScript)
 import Cardano.Ledger.BaseTypes (StrictMaybe (..))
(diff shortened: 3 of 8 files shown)


libs/cardano-ledger-test/src/Test/Cardano/Ledger/Examples/ConwayGovActionReorder.hs
```diff
@@ -0,0 +1,40 @@
+{-# LANGUAGE TypeApplications #-}
+
+module Test.Cardano.Ledger.Examples.ConwayGovActionReorder (
```
teodanciu: This test seems like it could live in `cardano-ledger-conway`,  next to the new tests, so in 
`eras/conway/impl/test/Test/Cardano/Ledger/Conway`
In this way it could use `Test.Cardano.Ledger.Common` and HSpec instead of Tasty . 
Or if not, in cardano-ledger-conway-test, but from what I understand, the plan is to eventually move the tests from there into cardano-ledger-conway. 

But not sure, maybe you had a good reason to add it to `cardano-ledger-test` ?

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Ratify.hs
```diff
@@ -203,6 +208,23 @@ delayingAction TreasuryWithdrawals {} = False
 delayingAction ParameterChange {} = False
 delayingAction InfoAction {} = False
 
+actionPriority :: GovAction era -> Int
```
teodanciu, line 210: Why not implement `Ord`? are we using this value somewhere other than for reordering? Or maybe if we had an instance of `Ord`  there would be other - undesired - consequences?

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Ratify.hs
```diff
@@ -203,6 +208,23 @@ delayingAction TreasuryWithdrawals {} = False
...
+actionPriority TreasuryWithdrawals {} = 5
+actionPriority InfoAction {} = 6
+
+reorderActions :: StrictSeq (GovActionState era) -> StrictSeq (GovActionState era)
+reorderActions acts =
+  Seq.fromList . concat . IntMap.elems $
+    foldr'
+      (\x -> IntMap.insertWith (<>) (actionPriority $ gasAction x) [x])
+      IntMap.empty
+      acts
```
teodanciu: Is this more efficient than sorting, for example:
`Seq.fromList $ sortOn (actionPriority . gasAction)  (toList acts)`     ?

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Ratify.hs
lehins: Because it is priority, not `Ord`. 
`Ord` instance requires to use all of the contents of the type, otherwise you'll break referential transparency

teodanciu: Ah yes. makes sense ,thanks

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Ratify.hs
lehins: Yes. `sortOn` and `sort` are pretty slow. Using `Map` and `IntMap` and `Set` for sorting will always be much faster.

Note also an important fact that it would have to be `nubOn . sortOn` in order to match `IntSet.elems . IntMap.fromList`

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Examples/ConwayGovActionReorder.hs
Soupstraw: I thought we were moving our tests to `cardano-ledger-test`, but it could just as well live in `cardano-ledger-conway`

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Ratify.hs
Soupstraw: Also I don't think `sort` is guaranteed to preserve the order of equal elements.

lehins: @Soupstraw `Data.List.sort` implements a stable sorting algorithm, so equal elements are guaranteed to preserve relative order.

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Examples/ConwayGovActionReorder.hs
lehins: > I thought we were moving our tests to cardano-ledger-test

@Soupstraw This is quite the opposite. Only tests that cannot live in a test suite for specific era, should go into `cardano-ledger-test`. In fact we alredy have too many tests in `cardano-ledger-test` that could be moved into a specific era, but we'll deal with that later.

However, going forward, we should try our best to implement tests in `cardano-ledger-[era]:tests`, instead of `cardano-ledger-[era]-test` or `cardano-ledger-test`.

The ultimate goal is to get rid of `cardano-ledger-[era]-test` packages completely and make `cardano-ledger-test` as thin as possible, which means only tests that truly require the `Proof` or other things that are unique to that package, like constraint based testing, will live there.

See #3623 for a bit more context.

lehins: With respect to this particular module, I concur with @teodanciu, it should be switched to using hspec and placed into `cardano-ledger-conway:tests:Test.Cardano.Ledger.Conway.GovActionSpec`

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Ratify.hs
lehins: That being said, because of potential inefficiency: ` IntMap.insertWith (<>) (actionPriority $ gasAction x) [x]` I am not really sure that in this whole approach is really worth it. Using `sortOn` will be much simpler.

In the matter of fact this whole PR can be made much more efficient, but we don't have to deal with that right now.

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Arbitrary.hs
```diff
@@ -7,7 +7,16 @@
...
-module Test.Cardano.Ledger.Conway.Arbitrary () where
+module Test.Cardano.Ledger.Conway.Arbitrary (
+  genNewCommittee,
+  genNoConfidence,
+  genTreasuryWithdrawals,
+  genHardForkInitiation,
+  genParameterChange,
+  genNewConstitution,
+  genGovActionStateFromAction,
+  govActionGenerators,
```
lehins, line 18: These need to be added to the changelog under `testlib` section:
```
## 1.8.0.0

* Add `reorderActions`, `actionPriority`
* ...

### `testlib`

* Add `genNewCommittee`
* ...
```

lehins (changes requested):
Approach taken in this PR for keeping the order of proposed actions is quite inefficient, because it requires sorting the whole proposed sequence upon ratification. It can be made much more efficient by adding proposals in the correct order upon proposal instead.

That being said, the logic is correct in this PR, so we can work on optimizing it at a later time. I'll create a ticket about.

Moving the tests into the correct test suite and simplifying sorting are the only things left to do with respect to this PR.

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Ratify.hs
lehins: So, let's switch it to `sortOn`, so it clearer what is going on.

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Examples/ConwayGovActionReorder.hs
Soupstraw: `PrettyA` instances are often the reason I move tests to `cardano-ledger-test`, but I haven't used pretty printing here yet.

lehins: We should also be switching to ToExpr instances from PrettyA as well for testing

eras/conway/test-suite/src/Test/Cardano/Ledger/Conway/Examples/GovActionReorder.hs
lehins, line 1: I think you completely misunderstood my previous [comment](https://github.com/input-output-hk/cardano-ledger/pull/3698#discussion_r1314715259) and #3623 ticket. Could you please take a look at it again and let me know if it is still confusing for any reason, I'll be happy to elaborate. 

Could you please move this to `cardano-ledger-conway` package. Not `cardano-ledger-conway-test`

We should follow conventions of placing hspec specs into `*Spec` suffixed modules.

Also, we should not use "Examples" anymore, since tests have nothing to do with examples.

So, can you rename this module to `Test.Cardano.Ledger.Conway.GovActionSpec` and place it into `cardano-ledger-conway:tests` target

eras/conway/test-suite/cardano-ledger-conway-test.cabal
```diff
@@ -103,4 +105,5 @@ test-suite cardano-ledger-conway-test
         cardano-ledger-conway-test,
         cardano-ledger-core,
         cardano-ledger-shelley-test,
-        tasty
+        tasty,
+        tasty-hspec
```
lehins: No longer needed
```suggestion
        tasty
```

lehins (changes requested):

lehins (approved):
Beautiful! Thank you!

eras/conway/impl/CHANGELOG.md
```diff
@@ -78,6 +79,17 @@
 * Add `FromJSON` instance for `Committee`
 * Add `constitution` and `committee` fields to `ConwayGenesis`
 
+### testlib
```
lehins: Changes now need to be moved to `1.9.0.0` section, since `1.8.0.0` has already been released

lehins (approved):
Needs a rebase and some conflicts resolved

PR #3699, opened 2023-09-01
To/FromJSONKey instances for Credential and Voter 

### LINKED ISSUE #3580: Create a user friendly text representation for Credential and Voter
This is needed for ToJSON/ToJSONKey instances and potentially for other uses cases, such as error reporting.

In order to use a type as a Key in JSON it should be represented as a flat text.

teodanciu:
# Description

Closes https://github.com/input-output-hk/cardano-ledger/issues/3580

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Procedures.hs
+++ b/eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Procedures.hs
@@ -70,7 +70,7 @@ import Cardano.Ledger.Binary.Coders (
  )
 import Cardano.Ledger.Coin (Coin)
 import Cardano.Ledger.Core (Era (..), EraPParams (..), PParamsUpdate)
-import Cardano.Ledger.Credential (Credential (..))
+import Cardano.Ledger.Credential (Credential (..), credToText)
 import Cardano.Ledger.Crypto (Crypto)
 import Cardano.Ledger.Keys (KeyHash, KeyRole (..))
 import Cardano.Ledger.SafeHash (extractHash)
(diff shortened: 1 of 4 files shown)


eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Procedures.hs
```diff
@@ -162,9 +168,19 @@ data Voter c
...
 instance Crypto c => ToJSON (Voter c)
 
--- TODO: Make it nicer, eg. "DRep-ScriptHash<0xdeadbeef>..."
--- Will also need a ToJSONKey instance for Credential.
-instance Crypto c => ToJSONKey (Voter c)
+instance Crypto c => ToJSONKey (Voter c) where
+  toJSONKey = toJSONKeyText prefixVoter
+    where
+      prefixVoter (CommitteeVoter (KeyHashObj (KeyHash kh))) =
+        "committee-keyhash-" <> hashToTextAsHex kh
```
teodanciu: I can't tell if it's possible here to reuse the `ToJSONKey` instance of the credential, which is what I would like to do.
`toJSONKey` returns `ToJSONKeyFunctions` without taking any arguments, so not sure how I would apply this function to the Credential :thinking:

lehins: Yes you can create a separate function `credentialToText` and reuse it here and in the ToJSONKey instance for Credential

lehins: Like so. Also a nice approach to reduce amount of code on the screen with LambdaCase:
```suggestion
  toJSONKey = toJSONKeyText $ \case
    CommitteeVoter cred -> "committee-" <> credentialToText cred
```

teodanciu: that's a good idea. thanks

lehins (approved):
Besides duplication of code, which you mentioned yourself in Voter instance, this PR is ready to go. :+1:

PR #3700, opened 2023-09-04
Add conway-specific certs to deposit/refunds

### LINKED ISSUE #3696: Account for new delegation certificates in deposit/refund calculation
In Conway we have new certificates that affect deposits and refunds, when compared to Shelley through Babbage, in particular:
* `RegDepositDelegTxCert`
* `RegDepositTxCert`
* and `UnRegDepositTxCert`

Above certificates are currently not accounted for in consumed/produced calculation

aniketd:
# Description

Closes #3696

I suspect that this implementation may be incorrect, because:
1. although we track `RegDepositTxCert` and `UnRegDepositTxCert` (in `Conway.UTxO`) separately from `RegTxCert` and `UnRegTxCert` (in `Mary.UTxO`) they could be considered essentially the same, because these could be mixed up submissions in the same `Tx`, and need to be calculated together in Conway?
2. with respect to deregistering, we have a `RegDepositDelegTxCert` that has no equivalent `Un_` counterpart (as it does not need one), but then, the natural question to ask myself is, should this also be considered to be accounted for along with the group of certs being tracked in point 1?

In other words, should the deposit/refund calculation be the following:

```haskell
produced =
  getSum @Int 
    (  foldMap'
        (  \case
            RegTxCert {} -> 1
            RegDepositTxCert {} -> 1
            RegDepositDelegTxCert {} -> 1
            _ -> 0
        )
        (txBody ^. certsTxBodyL)
    )
    <x> pp 
    ^. ppKeyDepositL
```

and

```haskell
consumed = 
  let go accum@(!depositsInTx, !totalRefunds) = \case
        -- tracking registrations in the same tx
        RegTxCert c -> (Set.insert c depositsInTx, totalRefund)
        RegDepositTxCert c _ -> (Set.insert c depositsInTx, totalRefund)
        RegDepositDelegTxCert c _ _ -> (Set.insert c depositsInTx, totalRefund)
        -- cancelling out registrations in the same tx and refunding otherwise
        UnRegTxCert c
          | Set.member c depositsInTx -> (Set.delete c depositsInTx, totalRefund <+> pp ^. ppKeyDepositL)
          | Just deposit <- lookupKeyDeposit c -> (depositsInTx, totalRefund <+> deposit)
        UnRegDepositTxCert c _
          | Set.member c depositsInTx -> (Set.delete c depositsInTx, totalRefund <+> pp ^. ppKeyDepositL)
          | Just deposit <- lookupKeyDeposit c -> (depositsInTx, totalRefund <+> deposit)
   in snd $ foldl' go (Set.empty, Coin 0) $ txBody ^. certsTxBodyL
```

unified in conway and not partially separated out (for reuse) in `shelleyProducedValue` and `getConsumedMaryValue`?

Let me know what you guys think!

# Checklist

- [x] Self-reviewed the diff


eras/conway/impl/src/Cardano/Ledger/Conway/UTxO.hs
```diff
@@ -84,13 +88,35 @@ conwayProducedValue ::
   TxBody era ->
   Value era
 conwayProducedValue pp isStakePool txBody =
-  inject propDeposits
```
lehins: Revert it back to `where` instead of `let` in order to minimize the diff.

Arbitrary switches like that between equally accepted styles should not happen without a good enough reason.

eras/conway/impl/src/Cardano/Ledger/Conway/UTxO.hs
```diff
@@ -118,7 +144,19 @@ conwayConsumedValue pp lookupKeyDeposit lookupDRepDeposit utxo txBody =
                 | Just deposit <- lookupDRepDeposit c -> (drepRegsInTx, totalRefund <+> deposit)
               _ -> accum
          in inject $ snd $ foldl' go (Set.empty, Coin 0) $ txBody ^. certsTxBodyL
-   in maryRefunds <+> drepRefunds
+      conwayRegRefunds =
```
lehins: Yes, this approach is incorrect. The way you pointed out in the PR comment is the correct one. However, there is a much cleaner way to solve this without all the duplication

We can meet and talk about it.

lehins (changes requested):

eras/conway/impl/src/Cardano/Ledger/Conway/UTxO.hs
aniketd: Super thanks! :grin: :raised_hands:

lehins (approved):
Perfect!

PR #3706, opened 2023-09-05
Fix ConwayTxBodyRaw decoder, VotingProcedures functions

### LINKED ISSUE #3671: API improvements for VotingProcedure.
1. Deserializer should not allow empty Map. It is an `votingProceduresTxBodyL` is an optional field, so having an empty Map in an optional field makes no sense
2. Provide functions from cardano-ledeger-api (define them next to the VotingProcedures type though) to insert and union votes. (Potentially create a single api that can also be used in the mint field)

aniketd:
# Description

Fix the deserializer for `ConwayTxBodyRaw` to fail when the `VotingProcedures` field is an empty map. This is important because if it is an empty map we omit this field entirely during serialization.

Also, export functions to work with VotingProcedures, namely,
1. `emptyVotingProcedures`
2. `singletonVotingProcedures`
3. `unionLVotingProcedures`
4. `unionLVotingProceduresEither`
5. `insertVote`

Closes #3671 

# Checklist

- [x] Self-reviewed the diff


eras/conway/impl/src/Cardano/Ledger/Conway/TxBody.hs
```diff
@@ -208,6 +215,20 @@ instance
...
+  vpMap <-
+    decodeMapByKey decCBOR $ \voter -> do
+      subMap <- decCBOR
+      when (Map.null subMap) $
+        fail $
+          "TxBody: VotingProcedures require votes, but Voter: " <> show voter <> " didn't have any."
+      pure subMap
+  when (Map.null vpMap) $ fail "TxBody: VotingProcedures requires voters, but none found. The map is empty."
+  pure $ VotingProcedures vpMap
+{-# INLINE decodeNonEmptyTxBodyVotingProcedures #-}
```
lehins, line 230: This is duplication of code. You must reuse the `DecCBOR` instance

eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Procedures.hs
lehins, line 1: This is not what I had in mind for this task, but the good thing is that we have a list of functions that cardano-api needs.

I'll create a follow up task.

lehins (changes requested):
I created a separate ticket: https://github.com/input-output-hk/cardano-ledger/issues/3707

This PR can be closed for now.

PR #3708, opened 2023-09-05
Post release changes

lehins:
# Description

There was a release of ledger packages today: https://github.com/input-output-hk/cardano-haskell-packages/pull/472. Need to account for it.

Also workaround an intermittent issue with GitHub 

Also fix the wrong path in the cabal file

# Checklist

- [x] Self-reviewed the diff


eras/conway/impl/cardano-ledger-conway.cabal
```diff
@@ -11,7 +11,7 @@ description:
 
 category:           Network
 build-type:         Simple
-data-files:         data/*.json
+data-files:         test/data/*.json
```
teodanciu, line 14: Oops, sorry about that.  Did it cause issues, or you happened to notice?

eras/conway/test-suite/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Version history for `cardano-ledger-conway-test`
 
+## 1.2.1.0
+
+* Add `Crypto c` constraint to `exampleConwayGenesis`
```
teodanciu, line 5: TIL that this kind of change has to be added to CHANGELOG too, makes sense.

teodanciu (approved):
Looks good to me, it seems like this ledger release was quite smooth, with no fixes necessary for the Chap build to work, awesome :raised_hands:

eras/conway/impl/cardano-ledger-conway.cabal
lehins: hydra CI failed, that's how I noticed. No big deal.

eras/conway/test-suite/CHANGELOG.md
lehins: In a regular Haskell package this would also be a breaking change, since downstream users would be forced to add this constraint too, but no-one except us using this function and it can't really be used without Crypto constraint anyways, so this is no big deal and for us it is not really a breaking change. 
But, yes, the important part is that this is a semantic change, which should go into a changelog

PR #3712, opened 2023-09-06
Disallow empty fields in ConwayTxBodyRaw

### LINKED ISSUE #3707: Prevent empty values in CBOR of TxBody
It is redundant to supply empty values for these three fields:

* `votingProceduresTxBodyL`
* `proposalProceduresTxBodyL`
* `certsTxBodyL`

Implement a helper function that will fail whenever an optional field is being decoded that corresponds to an empty value (would accept a function as an argument to detect if decoded value is empty (eg. `(== mempty)`)

The idea behind not allowing empty fields is reduction in the transaction size. The whole reason for using sparse fields is to make sure we use less memory, but if we allow supplying empty fields this will result to even worse overhead than if we didn't use sparse representation to begin with.

aniketd:
# Disallow empty fields in ConwayTxBodyRaw

- Fix `ConwayTxBodyRaw` decoder to disallow empty
  - `certsTxBodyL`
  - `withdrawalsTxBodyL`
  - `mintTxBodyL`
  - `collateralInputsTxBodyL`
  - `reqSignerHashesTxBodyL`
  - `referenceInputsTxBodyL`
  - `votingProceduresTxBodyL`
  - `proposalProceduresTxBodyL`
- Add `fieldGuarded` to conditionally construct a `Field`
- Fix the related CDDL for Conway

Closes #3707 

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/eras/conway/impl/CHANGELOG.md
+++ b/eras/conway/impl/CHANGELOG.md
@@ -2,6 +2,15 @@
 
 ## 1.9.0.0
 
+* Fix `ConwayTxBodyRaw` decoder to disallow empty `Field`s #3712
+  * `certsTxBodyL`
+  * `withdrawalsTxBodyL`
+  * `mintTxBodyL`
+  * `collateralInputsTxBodyL`
+  * `reqSignerHashesTxBodyL`
+  * `referenceInputsTxBodyL`
+  * `votingProceduresTxBodyL`
+  * `proposalProceduresTxBodyL`
 * Add `reorderActions`, `actionPriority`
 * Remove `ensProtVer` field from `EnactState`: #3705
 * Move `ConwayEraTxBody` to `Cardano.Ledger.Conway.TxBody`
--- a/eras/conway/impl/cardano-ledger-conway.cabal
+++ b/eras/conway/impl/cardano-ledger-conway.cabal
@@ -69,7 +69,7 @@ library
         aeson,
         bytestring,
         cardano-crypto-class,
-        cardano-ledger-binary >=1.1,
+        cardano-ledger-binary >=1.1.3,
         cardano-ledger-allegra >=1.1,
         cardano-ledger-alonzo ^>=1.4.1,
         cardano-ledger-babbage >=1.4.1,
--- a/eras/conway/impl/src/Cardano/Ledger/Conway/TxBody.hs
+++ b/eras/conway/impl/src/Cardano/Ledger/Conway/TxBody.hs
@@ -70,6 +70,7 @@ import Cardano.Ledger.Binary.Coders (
   encode,
   encodeKeyedStrictMaybe,
   field,
+  fieldGuarded,
   ofield,
   (!>),
  )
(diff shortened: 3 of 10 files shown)


eras/conway/impl/cardano-ledger-conway.cabal
```diff
@@ -69,7 +69,7 @@ library
         aeson,
         bytestring,
         cardano-crypto-class,
-        cardano-ledger-binary >=1.1,
+        cardano-ledger-binary >=1.1.2.1,
```
lehins: This is an indicator that something is not right. You should never have a patch versions in bounds (neither upper nor lower), because patch version bump is reserved for non-semantic changes to the public interface. In cabal bounds we only rely on public interface of a dependency.

libs/cardano-ledger-binary/src/Cardano/Ledger/Binary/Decoding/Coders.hs
```diff
@@ -130,6 +132,22 @@ data Field t where
...
+  -- | The condition to guard against
+  (x -> Bool) ->
+  (x -> t -> t) ->
+  Decode ('Closed d) x ->
+  Field t
+fieldGuarded failMsg check update dec =
+  Field
+    update
+    ( decode dec >>= \x ->
+        x <$ unless (check x) (fail $ "Guarded Field does not match condition. " <> failMsg)
```
lehins: Also, I don't think this is a useful message, since people who will get this error message have no clue what Guarded Field is
```suggestion
        x <$ when (check x) (fail failMsg)
```

libs/cardano-ledger-binary/CHANGELOG.md
```diff
@@ -2,7 +2,7 @@
 
 ## 1.1.2.1
 
-*
+* Add `fieldGuarded` to be able to conditionally construct a `Field` #
```
lehins: This is a semantic change, addition of a new function, which means it must be a minor bump: `1.1.3.0`

Also, I am sure you were gonna do this anyways, but just in case :wink: 
```suggestion
* Add `fieldGuarded` to be able to conditionally construct a `Field` #3712 
```

lehins (changes requested):
Few minor issues

eras/conway/impl/cardano-ledger-conway.cabal
aniketd: Got it! I need to bump the `minor` part, since a new function was added.

eras/conway/impl/src/Cardano/Ledger/Conway/TxBody.hs
```diff
@@ -195,8 +202,18 @@ instance
...
-      bodyFields 20 = field (\x tx -> tx {ctbrProposalProcedures = x}) From
+      bodyFields 19 =
+        fieldGuarded
+          "TxBody: Cannot decode empty VotingProcedures"
+          (Map.null . unVotingProcedures)
+          (\x tx -> tx {ctbrVotingProcedures = x})
+          From
+      bodyFields 20 =
+        fieldGuarded
+          "TxBody: Cannot decode empty ProposalProcedures"
```
lehins: What do you say if we abstract the message into a separate function in order to reduce duplication and rephrase the error message a bit:
```suggestion
          emptyFailure "proposalProcedures" "non-empty"
```
```haskell
  where
    emptyFailure fieldName requirement =
      "TxBody: '" ++ fieldName ++ "' must be " ++ requirement ++ " when supplied"
```

This way `emptyFailure` can be used for all such fields including the `treasuryDonations`

aniketd: Great idea! Thanks! :raised_hand_with_fingers_splayed:

libs/cardano-ledger-binary/src/Cardano/Ledger/Binary/Decoding/Coders.hs
```diff
@@ -130,6 +132,22 @@ data Field t where
...
+  -- | The condition to guard against
+  (x -> Bool) ->
+  (x -> t -> t) ->
+  Decode ('Closed d) x ->
+  Field t
+fieldGuarded failMsg check update dec =
+  Field
+    update
+    ( decode dec >>= \x ->
+        x <$ unless (check x) (fail failMsg)
```
lehins: I made this suggestion already `unless` is wrong here. We need to fail when the field is empty, not the other way around
```suggestion
        x <$ when (check x) (fail failMsg)
```

aniketd: Ah, yes! I got it backwards :+1:  :bow:

eras/conway/impl/src/Cardano/Ledger/Conway/TxBody.hs
```diff
@@ -180,23 +181,63 @@ instance
...
+          From
       bodyFields 7 = ofield (\x tx -> tx {ctbrAuxDataHash = x}) From
       bodyFields 8 =
         ofield
           (\x tx -> tx {ctbrVldt = (ctbrVldt tx) {invalidBefore = x}})
           From
-      bodyFields 9 = field (\x tx -> tx {ctbrMint = x}) From
+      bodyFields 9 =
+        fieldGuarded
+          (emptyFailure "Mint" "non-empty or non-zero")
```
lehins: This decoder does not check for zeros. Presence of zeros will be reported by the MultiAsset decoder if a zero indeed is present 
```suggestion
          (emptyFailure "Mint" "non-empty")
```

eras/conway/impl/src/Cardano/Ledger/Conway/TxBody.hs
```diff
@@ -180,23 +181,63 @@ instance
...
-      bodyFields 14 = field (\x tx -> tx {ctbrReqSignerHashes = x}) From
+      bodyFields 13 =
+        fieldGuarded
+          (emptyFailure "Collateral Inputs" "non-empty")
+          null
+          (\x tx -> tx {ctbrCollateralInputs = x})
+          From
+      bodyFields 14 =
+        fieldGuarded
+          (emptyFailure "Required Signer (Witness) Hashes" "non-empty")
```
lehins:
```suggestion
          (emptyFailure "Required Signer Hashes" "non-empty")
```

eras/conway/impl/src/Cardano/Ledger/Conway/TxBody.hs
```diff
@@ -209,6 +250,8 @@ instance
         , (1, "outputs")
         , (2, "fee")
         ]
+      emptyFailure fieldName requirement =
```
lehins, line 253: Use this function for `TreasuryWithdrawals` as well.

eras/conway/impl/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## 1.9.0.0
 
+* Fix `ConwayTxBodyRaw` decoder to disallow empty `Field`s #3712
+  * `certsTxBodyL`
+  * `votingProceduresTxBodyL`
+  * `proposalProceduresTxBodyL`
```
lehins, line 13: Need to add the other 5 fields here as well

eras/conway/test-suite/cddl-files/conway.cddl
```diff
@@ -68,7 +68,7 @@ transaction_body =
   , ? 17 : coin                     ; total collateral
   , ? 18 : set<transaction_input>   ; reference inputs
```
lehins: Can you add a comment on the fields that you adjusted which are `<set>`s, that they can't be empty. I am not sure if there is a way to specify it with cddl.

I suspect we might have problems with CDDL tests because of this  :disappointed:

lehins (approved):
:+1:

eras/conway/impl/src/Cardano/Ledger/Conway/TxBody.hs
aniketd: Got it! I think you mean treasury donations, since withdrawals already uses it :+1: Will update!

lehins: That's what I mean. I actually got confused by the current failure message which is wrong :smile: 
https://github.com/input-output-hk/cardano-ledger/blob/a0db238724037e4875516ab87f0d1a1b61ce3ec0/eras/conway/impl/src/Cardano/Ledger/Conway/TxBody.hs#L204

eras/conway/test-suite/cddl-files/conway.cddl
aniketd: Added a `nonempty_set<a> = [+ a]` to `cddl-files/mock/extras.cddl` and a corresponding definition to `cddl-files/real/extras.cddl` as well. I also increased the test runs to `10` from `5`, hoping that at least some more base cases could be covered.

eras/conway/impl/src/Cardano/Ledger/Conway/TxBody.hs
```diff
@@ -180,35 +181,77 @@ instance
...
+          (emptyFailure "ProposalProcedures" "non-empty")
+          null
+          (\x tx -> tx {ctbrProposalProcedures = x})
+          From
       bodyFields 21 = ofield (\x tx -> tx {ctbrCurrentTreasuryValue = x}) From
       bodyFields 22 =
         ofield
           (\x tx -> tx {ctbrTreasuryDonation = fromSMaybe zero x})
-          (D (decodePositiveCoin "'treasuryWithdrawal' must be non-zero when supplied"))
+          (D (decodePositiveCoin $ emptyFailure "Treasury Donation" "non-empty"))
```
lehins:
```suggestion
          (D (decodePositiveCoin $ emptyFailure "Treasury Donation" "non-zero"))
```

lehins (approved):

eras/conway/impl/src/Cardano/Ledger/Conway/TxBody.hs
aniketd: Thank you! :raised_hands:

Issue #3685, opened 2023-08-28
Add PredicateFailure for invalid Voter-GovAction pairs


lehins:
Not all GovActions can be voted by all Voters. Eg. SPOs can't vote on UpdateConstitution governance action. Currently this is protected during ratification by setting thresholds to zero, thus not counting any invalid votes that are submitted (implemented in #3674)

We need to improve clarity for users by preventing invalid votes like that to be placed on chain by adding a predicate check in a GOV rule that rejects transactions with invalid votes.

Care should be taken not to duplicate logic that is implemented in the RATIFY rule.

PR #3716, opened 2023-09-07
Abstract threshold calculation

lehins:
# Description

We need to have ability to answer if voting is allowed for a particular entity by reusing the current threshold calculation logic, thus reducing duplication. This is needed for #3685

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/eras/conway/impl/src/Cardano/Ledger/Conway/Governance.hs
+++ b/eras/conway/impl/src/Cardano/Ledger/Conway/Governance.hs
@@ -42,9 +42,12 @@ module Cardano.Ledger.Conway.Governance (
   indexedGovProps,
   Constitution (..),
   ConwayEraGov (..),
-  thresholdSPO,
-  thresholdDRep,
-  thresholdCC,
+  votingStakePoolThreshold,
+  votingDRepThreshold,
+  votingCommitteeThreshold,
+  isStakePoolVotingAllowed,
+  isDRepVotingAllowed,
+  isCommitteeVotingAllowed,
   ProposalsSnapshot,
   snapshotInsertGovAction,
   snapshotActions,
@@ -85,6 +88,7 @@ import Cardano.Ledger.BaseTypes (
   ProtVer (..),
   StrictMaybe (..),
   UnitInterval,
+  isSJust,
  )
 import Cardano.Ledger.Binary (
   DecCBOR (..),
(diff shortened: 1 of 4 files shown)


aniketd (approved):
LGTM :+1: Thank you!

Just an idea: may be we could make `PParams` have `VotingThreshold` too.

eras/conway/impl/src/Cardano/Ledger/Conway/Governance.hs
```diff
@@ -593,69 +597,156 @@ pparamsUpdateThreshold pp ppu =
...
+  | -- | Does not have a threshold, therefore an action can not be ratified
+    NoVotingThreshold
+  | -- | Some GovActions are not alloweed to be voted by some entities
+    NoVotingAllowed
+
+toRatifyVotingThreshold :: VotingThreshold -> StrictMaybe UnitInterval
+toRatifyVotingThreshold = \case
+  VotingThreshold t -> SJust t -- concrete threshold
+  NoVotingThreshold -> SNothing -- no voting threshold prevents ratification
+  NoVotingAllowed -> SJust minBound -- votes should not count, set threshold to zero
```
Soupstraw, line 613: The naming is a bit confusing. I would've thought that `NoVotingThreshold` means that it's automatically accepted and `NoVotingAllowed` means it never passes

Soupstraw: Maybe it's better to rename to `AlwaysPass` and `NeverPass`

eras/conway/impl/src/Cardano/Ledger/Conway/Governance.hs
```diff
@@ -42,9 +42,12 @@ module Cardano.Ledger.Conway.Governance (
   indexedGovProps,
   Constitution (..),
   ConwayEraGov (..),
-  thresholdSPO,
-  thresholdDRep,
-  thresholdCC,
+  getStakePoolThreshold,
+  getDRepThreshold,
+  getCommitteeThreshold,
```
Soupstraw: I don't think that prepending `get` is necessary, makes it look like Java code :grin:

eras/conway/impl/src/Cardano/Ledger/Conway/Governance.hs
lehins: > I would've thought that NoVotingThreshold means that it's automatically accepted and NoVotingAllowed means it never passes

That's because it is not. It is the opposite.

* `NoVotingThreshold` -> it will never be accepted
* `NoVotingAllowed` -> it will always be accepted.

Problem with naming is that it is not only about accepting governance actions, but it is also about allowing for them to be voted on. 

If I name it:
* `NeverAccept` or `NeverPass`
* `AlwaysAccept` or `AlwaysAccept`

Then why `AlwaysPass` does not allow voting and `NeverPass` does allow voting?

It is a bit confusing, but I can't think of a better naming. Any other ideas?

eras/conway/impl/src/Cardano/Ledger/Conway/Governance.hs
lehins: Agreed. Renamed to `voting[Entity]Threshod`

eras/conway/impl/src/Cardano/Ledger/Conway/Governance.hs
Soupstraw: We can just as well prevent voting on both `AlwaysPass` and `NeverPass`. Returning `SNothing` as the threshold essentially means that `threshold > 1`, it doesn't have anything to do with preventing or allowing voting.

Soupstraw: I also just realized that `VotingThreshold minBound` and `NoVotingAllowed` are essentially the same thing

Soupstraw: Oh, ok, except for `isVotingAllowed`

Soupstraw: Maybe it's a good idea to somehow prevent constructing `VotingThreshold minBound`

teodanciu (commented):
There is something about this pattern that feels a little wrong.  
I think I understand where it's coming from:  for these `is.*Allowed` functions, we don't actually need to know anything other than whether `voting.*ThresholdInternal` returns `NoVotingAllowed`, while returning `NoVotingAllowed` in `voting.*ThresholdInternal` functions does not depend on the value of other parameters like `pparam`, `isElectedCommittee` or `committee`.  
So in theory, 
for `isStakePoolVotingAllowed` we could return True,
for `isCommitteeVotingAllowed` we could return True if the action is NoConfidence or NewCommittee
and 
for `isDRepVotingAllowed` we could return True  

,if I understand correctly the code. 
But then this would mean we duplicate the logic between determining whether the voting is allowed and computing the actual threshold. 

Semantically these parameters are optional (`pparam`, `isElectedCommittee` or `committee`), but I can see how making them optional is not gonna help. 
On the other hand, passing `emptyPparams` makes it seem like we might depend on the value of the default value of the parameters.

Could we somehow extract this information (i.e.: what makes different actions return NoVotingAllowed) and then use it from both places? 
I don't have a concrete idea for refactoring, just sharing my thoughts on this. Maybe we can think of a way to refactor it later if it's not the time for it now.

eras/conway/impl/src/Cardano/Ledger/Conway/Governance.hs
lehins: > Oh, ok, except for isVotingAllowed

That's the whole point of this PR. This function is necessary for #3718 

> Maybe it's a good idea to somehow prevent constructing VotingThreshold minBound

`VotingThreshold minBound` is a legitimate value, because nothing says any of those thresholds can't be set to 0 in the PParams.

lehins: > We can just as well prevent voting on both AlwaysPass and NeverPass

We can't. Those are two separate concepts:

* For one of them (`NoVotingAllowed`) voting is not counted by RATIFY rule and will be prevented in GOV rule in #3718
* For the other (`NoVotingThreshold`) we do allow votes, but we can't RATIFY them. Which makes sense for:
  * InfoAction
  * Committee votes when there is no committee. Note that, committee members can still submit votes, even if there is currently no committee, because a committee member might now that he'll be elected soon. Once he's elected his votes will be counted. 
  
So, for `NoVotingThreshold` we can't disallow voting, since that would prohibit those use cases.

lehins:
> There is something about this pattern that feels a little wrong.

@teodanciu Good thing that coding is about facts, not feelings :stuck_out_tongue_winking_eye:. I think you getting the same feeling some people get about the `Default` class. Cause that is what that whole type class is about: providing some value, when the value is not particularly relevant. 

> But then this would mean we duplicate the logic between determining whether the voting is allowed and computing the actual threshold.

I don't understand this comment. The whole point of this PR is to prevent duplication of logic in two places: figuring the thresholds and figuring out whether voting is allowed or not. Which, I believe, implementation in this PR successfully avoids duplication.

> Semantically these parameters are optional (pparam, isElectedCommittee or committee), but I can see how making them optional is not gonna help.

They are optional for figuring out whether voting is allowed or not, but they are required for figuring out the thresholds. In fact I could use `def` for those arguments in the former case, when they are optional.

> On the other hand, passing emptyPparams makes it seem like we might depend on the value of the default value of the parameters.

We could pass `undefined`, but that would be a lot worse in my opinion. In any case this is an internal function with a comment describing why this is argument is optional.

> Could we somehow extract this information (i.e.: what makes different actions return NoVotingAllowed) and then use it from both places?

Previous implementation did not allow us to reuse this logic at all for the two different use cases. This was my attempt at making it possible. I know of a way to make it possible to completely avoid using default values for arguments, namely use types that return functions that can be later either applied to arguments. This is an approach one would do in a strict language. However that would result in a more complex solution and more boiler plate, since it would require more types. Luckily for us Haskell is a lazy language, so we don't need to worry about creating more elaborate solutions in order to avoid redundant computation.

> Maybe we can think of a way to refactor it later if it's not the time for it now.

We can always try to improve the code later. Most importantly this whole logic is localized to one module, so we don't need to worry about affecting any rules in a breaking way.

Soupstraw (approved):

teodanciu (approved):
Looks good to me

PR #3719, opened 2023-09-08
Delay all enactments by one extra epoch

### LINKED ISSUE #3691: Postpone Enactment of all GovActions to next epoch
According to [CIP-1694](https://github.com/cardano-foundation/CIPs/tree/master/CIP-1694#lifecycle) `HardForkInitiation` and `ParameterChange` governance actions should be enacted a full epoch after ratification.

In order to avoid complex changes to the spec and implementation we decided to postpone enactment for one more epoch. This is necessary:
* for proper functioning HardFork combinator in consensus (2 stability windows)
* for block header validation (1 stability window)

In the end we postpone enactment for 2 epochs:
* 1 epoch to compute stake distributions
* 1 more epoch for allowing for safe consensus

In the future we will be able optimize it to one 1 epoch, but implementation will have to be a lot more complex

Soupstraw:
# Description

Delay all enactments by one extra epoch. The order of events for a governance action from proposal to enactment looks like this:

```
-- Epoch 0 --
Action P is proposed

<wait some epochs until P gets enough votes or expires>

-- Start of epoch n --
Take a snapshot and start calculating DRep distributions
  (we also do that on all previous epochs, but here we assume that this time the action is either ratified or expires)

-- Start of epoch n+1 --
If P has enough votes:
  - Apply P onto `cgFutureEnactState`
Remove P from the snapshots

-- Start of epoch n+2 --
Rotate `cgFutureEnactState` to `cgEnactState`
```

closes #3691

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/eras/conway/impl/src/Cardano/Ledger/Conway/Governance.hs
+++ b/eras/conway/impl/src/Cardano/Ledger/Conway/Governance.hs
@@ -59,6 +59,7 @@ module Cardano.Ledger.Conway.Governance (
   -- Lenses
   cgGovSnapshotsL,
   cgEnactStateL,
+  cgFutureEnactStateL,
   cgRatifyStateL,
   ensCommitteeL,
   ensConstitutionL,
@@ -473,6 +474,7 @@ instance EraPParams era => NoThunks (RatifyState era)
 data ConwayGovState era = ConwayGovState
   { cgGovSnapshots :: !(GovSnapshots era)
   , cgEnactState :: !(EnactState era)
+  , cgFutureEnactState :: !(EnactState era)
   }
   deriving (Generic, Eq, Show)
 
@@ -481,6 +483,9 @@ cgGovSnapshotsL = lens cgGovSnapshots (\x y -> x {cgGovSnapshots = y})
 
 cgEnactStateL :: Lens' (ConwayGovState era) (EnactState era)
 cgEnactStateL = lens cgEnactState (\x y -> x {cgEnactState = y})
+
+cgFutureEnactStateL :: Lens' (ConwayGovState era) (EnactState era)
+cgFutureEnactStateL = lens cgFutureEnactState (\x y -> x {cgFutureEnactState = y})
 
 {-# DEPRECATED cgRatifyStateL "Use cgEnactStateL instead" #-}
 cgRatifyStateL :: Lens' (ConwayGovState era) (RatifyState era)
@@ -500,7 +505,8 @@ instance EraPParams era => DecShareCBOR (ConwayGovState era) where
   decShareCBOR _ =
     decode $
       RecD ConwayGovState
-        <! (D decNoShareCBOR)
+        <! D decNoShareCBOR
+        <! From
         <! From
 
 instance EraPParams era => DecCBOR (ConwayGovState era) where
@@ -512,6 +518,7 @@ instance EraPParams era => EncCBOR (ConwayGovState era) where
       Rec ConwayGovState
         !> To cgGovSnapshots
         !> To cgEnactState
+        !> To cgFutureEnactState
 
 instance EraPParams era => ToCBOR (ConwayGovState era) where
   toCBOR = toEraCBOR @era
@@ -532,10 +539,11 @@ instance EraPParams era => ToJSON (ConwayGovState era) where
 instance EraPParams era => ToExpr (ConwayGovState era)
 
 toConwayGovPairs :: (KeyValue a, EraPParams era) => ConwayGovState era -> [a]
-toConwayGovPairs cg@(ConwayGovState _ _) =
+toConwayGovPairs cg@(ConwayGovState _ _ _) =
   let ConwayGovState {..} = cg
    in [ "gov" .= cgGovSnapshots
-      , "ratify" .= cgEnactState
+      , "curEnactState" .= cgEnactState
+      , "futureEnactState" .= cgEnactState
       ]
 
 instance EraPParams (ConwayEra c) => EraGov (ConwayEra c) where
--- a/eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Epoch.hs
+++ b/eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Epoch.hs
@@ -43,6 +43,7 @@ import Cardano.Ledger.Conway.Governance (
   GovSnapshots (..),
   RatifyState (..),
   cgEnactStateL,
+  cgFutureEnactStateL,
   cgGovSnapshotsL,
   curGovSnapshotsL,
   ensCommitteeL,
@@ -253,13 +254,14 @@ epochTransition = do
         , reDRepState = vstate ^. vsDRepsL
         }
     ratSig = RatifySignal . snapshotActions . prevGovSnapshots $ cgGovSnapshots govSt
+    enactState = govSt ^. cgFutureEnactStateL
   RatifyState {rsRemoved, rsEnactState} <-
     trans @(EraRule "RATIFY" era) $
       TRC
         ( ratEnv
         , RatifyState
             { rsRemoved = mempty
-            , rsEnactState = govSt ^. cgEnactStateL
+            , rsEnactState = enactState
             , rsDelayed = False
             }
         , ratSig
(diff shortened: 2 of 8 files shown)


teodanciu:
Thanks for writing the example! 

In the implementation, where is this part? 

```If P has enough votes:
  - Apply P onto `cgFutureEnactState`
Remove P from the snapshots
```

eras/conway/impl/src/Cardano/Ledger/Conway/Governance.hs
```diff
@@ -551,10 +558,11 @@ instance EraPParams era => ToJSON (ConwayGovState era) where
 instance EraPParams era => ToExpr (ConwayGovState era)
 
 toConwayGovPairs :: (KeyValue a, EraPParams era) => ConwayGovState era -> [a]
-toConwayGovPairs cg@(ConwayGovState _ _) =
+toConwayGovPairs cg@(ConwayGovState _ _ _) =
   let ConwayGovState {..} = cg
    in [ "gov" .= cgGovSnapshots
-      , "ratify" .= cgEnactState
+      , "curEnact" .= cgEnactState
+      , "futureEnact" .= cgEnactState
```
```suggestion
      , "futureEnactState" .= cgFutureEnactState
```

eras/conway/impl/src/Cardano/Ledger/Conway/Governance.hs
```diff
@@ -551,10 +558,11 @@ instance EraPParams era => ToJSON (ConwayGovState era) where
 instance EraPParams era => ToExpr (ConwayGovState era)
 
 toConwayGovPairs :: (KeyValue a, EraPParams era) => ConwayGovState era -> [a]
-toConwayGovPairs cg@(ConwayGovState _ _) =
+toConwayGovPairs cg@(ConwayGovState _ _ _) =
   let ConwayGovState {..} = cg
    in [ "gov" .= cgGovSnapshots
-      , "ratify" .= cgEnactState
+      , "curEnact" .= cgEnactState
```
lehins: Nice catch on `ratify`. Let's give it a more complete name
```suggestion
      , "curEnactState" .= cgEnactState
```

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Epoch.hs
```diff
@@ -242,13 +243,14 @@ epochTransition = do
...
     ratSig = RatifySignal . snapshotActions . prevGovSnapshots $ cgGovSnapshots govSt
+    futureEnactState = govSt ^. cgFutureEnactStateL
   RatifyState {rsRemoved, rsEnactState} <-
     trans @(EraRule "RATIFY" era) $
       TRC
         ( ratEnv
         , RatifyState
             { rsRemoved = mempty
-            , rsEnactState = govSt ^. cgEnactStateL
+            , rsEnactState = futureEnactState
```
lehins: Are you sure this is suppose to be `futureEncastState`? I could of course be wrong, but in my understanding it should be `govSt ^. cgEnactStateL`.

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Examples/ConwayFeatures.hs
```diff
@@ -426,25 +436,25 @@ testGov pf = do
...
 
   assertExprEqualWithMessage "govState after second proposal" govState2 expectedGovState2
 
-  -- Wait two epochs
+  -- Wait three epochs
+  let
+    epochState2 = epochState1'' & esLStateL .~ ledgerState3
+  epochState3 <- expectRight "Error running runEPOCH: " $ runEPOCH (EPOCH pf) epochState2 (EpochNo 4) poolDistr
+  epochState3' <- expectRight "Error running runEPOCH: " $ runEPOCH (EPOCH pf) epochState3 (EpochNo 4) poolDistr
+  epochState4 <- expectRight "Error running runEPOCH: " $ runEPOCH (EPOCH pf) epochState3' (EpochNo 5) poolDistr
```
lehins: This looks slightly wrong, why do we run EPOCH rule on the same epoch number, but different state? Also we should stick to one convention of either ticks or numbers, without mixing them in the same function:
```suggestion
  epochState4 <- expectRight "Error running runEPOCH: " $ runEPOCH (EPOCH pf) epochState3 (EpochNo 5) poolDistr
  epochState5 <- expectRight "Error running runEPOCH: " $ runEPOCH (EPOCH pf) epochState4 (EpochNo 6) poolDistr
```

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Examples/ConwayFeatures.hs
```diff
@@ -399,19 +403,25 @@ testGov pf = do
...
             ]
         )
-    -- Wait two epochs
+    -- Wait three epochs
     eitherEpochState1 = runEPOCH (EPOCH pf) epochState0 (EpochNo 2) poolDistr
   epochState1 <- expectRight "Error running runEPOCH: " eitherEpochState1
   let
     constitution = epochState1 ^. esLStateL . lsUTxOStateL . utxosGovStateL . cgEnactStateL . ensConstitutionL
   assertExprEqualWithMessage "constitution after one epoch" constitution def
+  epochState1' <-
```
lehins: Same here could you avoid mixing ticks with numbers. It is hard to follow the dependency order

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Examples/ConwayFeatures.hs
```diff
@@ -399,19 +403,25 @@ testGov pf = do
...
+    expectRight "Error running runEPOCH: " $
+      runEPOCH (EPOCH pf) epochState1 (EpochNo 3) poolDistr
+  let
+    constitution' = epochState1' ^. esLStateL . lsUTxOStateL . utxosGovStateL . cgEnactStateL . ensConstitutionL
+  assertExprEqualWithMessage "constitution two epochs" constitution' def
+  epochState1'' <-
+    expectRight "Error running runEPOCH: " $
+      runEPOCH (EPOCH pf) epochState1' (EpochNo 3) poolDistr
```
lehins: Again same EpochNo as in the previous call to runEPOCH

lehins (changes requested):
I am very suspicious of the EnactState argument to the RATIFY rule. We should do some more analysis and more tests in order to ensure we got it right.

WhatisRT (commented):
I'm not sure I can provide a meaningful review, as I've got no idea how the delay for the incremental computation works. To me the code that didn't get changed looks completely wrong already, since all the inputs to `RATIFY` need to be snapshotted (if they aren't known statically), but it looks to me like the ingredients of `ratEnv` are just computed (transitively) from `es`.

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Epoch.hs
Soupstraw: I think it should be `futureEnactState`, because that's where we stored the enactments last epoch. Now one epoch later this becomes the actual enact state.

lehins:
> since all the inputs to RATIFY need to be snapshotted

@WhatisRT They are snapshotted. Stake distribution for Pools and DReps is from the previous epoch. We just need to make sure that the starting EnactState for the RATIFY rules also comes from the previous epoch

Soupstraw:
@teodanciu 
> In the implementation, where is this part?
> 
> ```
>   - Apply P onto `cgFutureEnactState`
> Remove P from the snapshots
> ```

This is when we run `RATIFY` using `futureEnactState`

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Epoch.hs
```diff
@@ -242,13 +243,14 @@ epochTransition = do
         , reDRepState = vstate ^. vsDRepsL
         }
     ratSig = RatifySignal . snapshotActions . prevGovSnapshots $ cgGovSnapshots govSt
+    futureEnactState = govSt ^. cgFutureEnactStateL
```
teodanciu: As we said in the meeting, `enactState` would be a less confusing name than `futureEnactState`

lehins:
This can be closed, as we decided not use this approach and instead we'll run RATIFY rule two stability windows before the end of the epoch (implemented in #3759), thus allowing us to postpone enactment only for one epoch, rather than two.

PR #3720, opened 2023-09-08
Made HardFork action drop all other actions

### LINKED ISSUE #3693: Discard all proposals upon HardForkInitiation ratification.
Upon ratification of a new HardForkInitiation governance action we must discard all of the governance proposals that are currently being considered. 

This is necessary because some of the proposals will become invalidated upon transition to a new era, a good example from the past would be removal of a protocol parameter.

Moreover, HardFork does not get Enacted until one epoch after being ratified, therefore we need to also prevent any new proposal procedures being accepted.

```
                    .--Epoch of no new Proposals
                    |
E0----------E1------^-----E2-------------E3
             |             |
             `RatifyHF     `EnactHF
               |              |
      (drop all Proposals)   (start accepting new Proposals)
```

Soupstraw:
# Description

Fixes #3693 

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Ratify.hs
+++ b/eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Ratify.hs
@@ -34,6 +34,7 @@ import Cardano.Ledger.Conway.Governance (
   GovActionState (..),
   PrevGovActionIds (..),
   RatifyState (..),
+  RatifyStrategy (..),
   Vote (..),
   votingDRepThreshold,
   votingStakePoolThreshold,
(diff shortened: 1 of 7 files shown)


eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Arbitrary.hs
```diff
@@ -127,6 +127,13 @@ instance
       <$> arbitrary
       <*> arbitrary
 
+instance Arbitrary RatifyStrategy where
+  arbitrary = elements
+    [ Continue
+    , Delay
+    , Drop
+    ]
```
lehins: Derive Enum and Bounded and you can use this implementation:
```suggestion
  arbitrary = arbitraryBoundedEnum
  shrink = shrinkBoundedEnum
```

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Ratify.hs
```diff
@@ -198,14 +199,14 @@ dRepAcceptedRatio RatifyEnv {reDRepDistr, reDRepState, reCurrentEpoch} gasDRepVo
...
-delayingAction TreasuryWithdrawals {} = False
-delayingAction ParameterChange {} = False
-delayingAction InfoAction {} = False
+actionFutureStrategy :: GovAction era -> RatifyStrategy
+actionFutureStrategy NoConfidence {} = Continue
+actionFutureStrategy NewCommittee {} = Continue
+actionFutureStrategy NewConstitution {} = Continue
+actionFutureStrategy TreasuryWithdrawals {} = Delay
+actionFutureStrategy ParameterChange {} = Delay
+actionFutureStrategy InfoAction {} = Delay
```
lehins: I think you mixed them up
```suggestion
actionFutureStrategy NoConfidence {} = Delay
actionFutureStrategy NewCommittee {} = Delay
actionFutureStrategy NewConstitution {} = Delay
actionFutureStrategy TreasuryWithdrawals {} = Continue
actionFutureStrategy ParameterChange {} = Continue
actionFutureStrategy InfoAction {} = Continue
```

lehins (changes requested):
Except one potential mistake, this looks good.

Soupstraw:
I'm thinking this wouldn't probably work the way I did it. We also somehow have to prevent adding new GAs after HF is ratified, probably by storing the RatifyStrategy for future epochs.

lehins:
> I'm thinking this wouldn't probably work the way I did it. We also somehow have to prevent adding new GAs after HF is ratified, probably by storing the RatifyStrategy for future epochs.

@Soupstraw It would work, you just haven't implemented it all the way

We need another flag in the enact state that would prevent new proposals whenever Ratify ended with a Drop strategy

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Ratify.hs
Soupstraw: nice catch!

lehins:
I think it will be better to clear out all of the proposals during translation, which will be implemented in the next era after Conway.

It would make a little more sense if we were leaning towards delaying for two epochs, but since now this is changing, dropping proposals at the same time we do a hardfork makes the rule a bit more complicated than it should be.

PR #3721, opened 2023-09-08
Committee expiration, validation and modification

### LINKED ISSUE #3443: Constitution Committee Member expiry
Each constitutional committee member will have their own specified epoch in which they are retired. This needs to be added to the CC cold/hot key map, something like: `Map (KeyHash 'CommitteeCold c) (KeyHash 'CommitteeHot c, Epoch)`.

The ledger rules need to enforce that the term limit is not past the protocol parameter that set the maximum.

~We have a decision still to make, the governance action that introduces a new committee member can either specify and epoch that the member retires or the duration (in epochs) until they retire.~ Duration is specified by a `CommitteeTermLimit` protocol parameter, which is specified in number of epochs.

teodanciu:
# Description

Closes https://github.com/input-output-hk/cardano-ledger/issues/3443

* In Gov rule, check the `NewCommittee` action failing if : 
   * the members to be removed are not part of the current committee (for this, the current committee from EnactState had to be added to GovEnv) 
   * the expiration epoch of the new members are not exceeding the ppCommitteeTermLimitL, given the current epoch
* Rename `NewCommittee` to `UpdateCommittee` and change its semantics to  "update committee", rather than "replace committee".  This means: in Enact rule: adding and removing members correctly, and in Gov rule: checking correctly the minimum committee size, considering the current committee.
* Update CommitteeState when members get removed

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Procedures.hs
+++ b/eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Procedures.hs
@@ -523,7 +523,7 @@ data GovAction era
       -- | Previous governance action id of `UpdateCommittee` or `NoConfidence` type, which
       -- corresponds to `CommitteePurpose`
       !(StrictMaybe (PrevGovActionId 'CommitteePurpose (EraCrypto era)))
-      -- | Old committee
+      -- | Constitutional Committe members to be removed
       !(Set (Credential 'ColdCommitteeRole (EraCrypto era)))
       -- | Constitutional committee members to be added
       !(Map (Credential 'ColdCommitteeRole (EraCrypto era)) EpochNo)
--- a/eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Procedures.hs
+++ b/eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Procedures.hs
@@ -516,17 +516,19 @@ data GovAction era
       -- | Proposed treasury withdrawals
       !(Map (RewardAcnt (EraCrypto era)) Coin)
   | NoConfidence
-      -- | Previous governance action id of `NoConfidence` or `NewCommittee` type, which
+      -- | Previous governance action id of `NoConfidence` or `UpdateCommittee` type, which
       -- corresponds to `CommitteePurpose`
       !(StrictMaybe (PrevGovActionId 'CommitteePurpose (EraCrypto era)))
-  | NewCommittee
-      -- | Previous governance action id of `NewCommittee` or `NoConfidence` type, which
+  | UpdateCommittee
+      -- | Previous governance action id of `UpdateCommittee` or `NoConfidence` type, which
       -- corresponds to `CommitteePurpose`
       !(StrictMaybe (PrevGovActionId 'CommitteePurpose (EraCrypto era)))
       -- | Old committee
       !(Set (Credential 'ColdCommitteeRole (EraCrypto era)))
-      -- | New Committee
-      !(Committee era)
+      -- | Constitutional committee members to be added
+      !(Map (Credential 'ColdCommitteeRole (EraCrypto era)) EpochNo)
+      -- New quorum
+      !UnitInterval
   | NewConstitution
       -- | Previous governance action id of `NewConstitution` type, which corresponds to
       -- `ConstitutionPurpose`
@@ -554,7 +556,7 @@ instance EraPParams era => DecCBOR (GovAction era) where
       1 -> SumD HardForkInitiation <! D (decodeNullStrictMaybe decCBOR) <! From
       2 -> SumD TreasuryWithdrawals <! From
       3 -> SumD NoConfidence <! D (decodeNullStrictMaybe decCBOR)
-      4 -> SumD NewCommittee <! D (decodeNullStrictMaybe decCBOR) <! From <! From
+      4 -> SumD UpdateCommittee <! D (decodeNullStrictMaybe decCBOR) <! From <! From <! From
       5 -> SumD NewConstitution <! D (decodeNullStrictMaybe decCBOR) <! From
       6 -> SumD InfoAction
       k -> Invalid k
@@ -571,8 +573,8 @@ instance EraPParams era => EncCBOR (GovAction era) where
         Sum TreasuryWithdrawals 2 !> To ws
       NoConfidence gid ->
         Sum NoConfidence 3 !> E (encodeNullStrictMaybe encCBOR) gid
-      NewCommittee gid old new ->
-        Sum NewCommittee 4 !> E (encodeNullStrictMaybe encCBOR) gid !> To old !> To new
+      UpdateCommittee gid old new q ->
+        Sum UpdateCommittee 4 !> E (encodeNullStrictMaybe encCBOR) gid !> To old !> To new !> To q
       NewConstitution gid c ->
         Sum NewConstitution 5 !> E (encodeNullStrictMaybe encCBOR) gid !> To c
       InfoAction ->
--- a/eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Enact.hs
+++ b/eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Enact.hs
@@ -18,6 +18,7 @@ import Cardano.Ledger.BaseTypes
 import Cardano.Ledger.Conway.Core
 import Cardano.Ledger.Conway.Era (ConwayENACT)
 import Cardano.Ledger.Conway.Governance (
+  Committee (..),
   EnactState (..),
   GovAction (..),
   GovActionId (..),
@@ -31,6 +32,8 @@ import Cardano.Ledger.Conway.Governance (
   ensPrevPParamUpdateL,
   ensProtVerL,
  )
+import Cardano.Ledger.Credential (Credential)
+import Cardano.Ledger.Keys (KeyRole (..))
 import Cardano.Ledger.Val (Val (..))
 import Control.State.Transition.Extended (
   STS (..),
@@ -40,6 +43,7 @@ import Control.State.Transition.Extended (
  )
 import Data.Foldable (fold)
 import qualified Data.Map.Strict as Map
+import qualified Data.Set as Set
 import Data.Void (Void)
 import Lens.Micro
 
@@ -86,10 +90,10 @@ enactmentTransition = do
         st
           & ensCommitteeL .~ SNothing
           & ensPrevCommitteeL .~ SJust (PrevGovActionId govActionId)
-    NewCommittee _ _ committee ->
+    UpdateCommittee _ membersToRemove membersToAdd newQuorum -> do
       pure $
         st
-          & ensCommitteeL .~ SJust committee -- TODO: check old members
+          & ensCommitteeL %~ SJust . updatedCommittee membersToRemove membersToAdd newQuorum
           & ensPrevCommitteeL .~ SJust (PrevGovActionId govActionId)
     NewConstitution _ c ->
       pure $
@@ -97,3 +101,21 @@ enactmentTransition = do
           & ensConstitutionL .~ c
           & ensPrevConstitutionL .~ SJust (PrevGovActionId govActionId)
     InfoAction -> pure st
+
+updatedCommittee ::
+  Set.Set (Credential 'ColdCommitteeRole (EraCrypto era)) ->
+  Map.Map (Credential 'ColdCommitteeRole (EraCrypto era)) EpochNo ->
+  UnitInterval ->
+  StrictMaybe (Committee era) ->
+  Committee era
+updatedCommittee membersToRemove membersToAdd newQuorum committee =
+  case committee of
+    SNothing -> Committee membersToAdd newQuorum
+    SJust (Committee currentMembers _) ->
+      let newCommitteeMembers =
+            Map.union
+              membersToAdd
+              (currentMembers `Map.withoutKeys` membersToRemove)
+       in Committee
+            newCommitteeMembers
+            newQuorum
--- a/eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Epoch.hs
+++ b/eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Epoch.hs
@@ -24,12 +24,20 @@ where
 
 import Cardano.Ledger.Address (RewardAcnt (getRwdCred))
 import Cardano.Ledger.BaseTypes (ShelleyBase)
-import Cardano.Ledger.CertState (certDStateL, certVStateL, dsUnifiedL, vsCommitteeStateL, vsDRepsL)
+import Cardano.Ledger.CertState (
+  CommitteeState (..),
+  certDStateL,
+  certVStateL,
+  dsUnifiedL,
+  vsCommitteeStateL,
+  vsDRepsL,
+ )
 import Cardano.Ledger.Coin (Coin)
 import Cardano.Ledger.Compactible (Compactible (..))
 import Cardano.Ledger.Conway.Core
 import Cardano.Ledger.Conway.Era (ConwayEPOCH, ConwayRATIFY)
 import Cardano.Ledger.Conway.Governance (
+  Committee,
   ConwayGovState (..),
   GovActionState (..),
   GovSnapshots (..),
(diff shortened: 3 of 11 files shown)


eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Procedures.hs
```diff
@@ -565,8 +567,8 @@ instance EraPParams era => EncCBOR (GovAction era) where
         Sum TreasuryWithdrawals 2 !> To ws
       NoConfidence gid ->
         Sum NoConfidence 3 !> E (encodeNullStrictMaybe encCBOR) gid
-      NewCommittee gid old new ->
-        Sum NewCommittee 4 !> E (encodeNullStrictMaybe encCBOR) gid !> To old !> To new
+      NewCommittee gid old new q ->
```
teodanciu: Could we change the name as well? to `UpdateCommitte`, since we're changing the cddl anyway?  Would that cause too big discrepancies with what is already integrated?

lehins: I think it is a great idea. Please, go ahead, change it in the cddl spec as well as Haskell code. This will reflect better the actual semantics that are described in the CIP-1694

teodanciu: I changed it without deprecation, I suppose they will have to reset sanchonet if this action was already submitted, but I assume that's find at this point

lehins: Yes, we are allowed to break SanchoNet :grin:

teodanciu: :fire:

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Epoch.hs
```diff
@@ -336,3 +352,9 @@ instance
   where
   wrapFailed = absurd
   wrapEvent = absurd
+
+updateCommitteeState :: StrictMaybe (Committee era) -> CommitteeState era -> CommitteeState era
+updateCommitteeState committee (CommitteeState creds) =
+  CommitteeState $ Map.restrictKeys creds (Map.keysSet members)
```
lehins: No need to convert it to a Set
```suggestion
  CommitteeState $ Map.difference creds members
```

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Epoch.hs
```diff
@@ -336,3 +352,9 @@ instance
...
+  where
+    members = foldMap' committeeMembers committee
```
lehins, line 361: Very cute way of converting Nothing to an empty Map :slightly_smiling_face: :heart:

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Gov.hs
```diff
@@ -232,17 +258,32 @@ govTransition = do
...
                   ?! TreasuryWithdrawalsNetworkIdMismatch mismatchedAccounts expectedNetworkId
-          NewCommittee _mPrevGovActionId _oldColdCreds newCommittee ->
-            let minCommitteeSize = pp ^. ppMinCommitteeSizeL
-                newCommitteeSize = fromIntegral . Map.size $ newCommittee ^. committeeMembersL
-             in newCommitteeSize
-                  >= minCommitteeSize
-                    ?! NewCommitteeSizeTooSmall newCommitteeSize minCommitteeSize
+          UpdateCommittee _mPrevGovActionId membersToRemove membersToAdd qrm -> do
+            _ <- checkCommitteeSize
+            _ <- checkUnexpectedMembers
```
lehins: These two actions return unit, so ignoring output should not be necessary?
```suggestion
            checkCommitteeSize
            checkUnexpectedMembers
```

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Gov.hs
```diff
@@ -232,17 +258,32 @@ govTransition = do
...
+                      >= minCommitteeSize
+                        ?! NewCommitteeSizeTooSmall committeeSize minCommitteeSize
+              checkUnexpectedMembers =
+                let currentMembers = Map.keysSet $ foldMap' (^. committeeMembersL) committee
+                    unexpectedMembers = membersToRemove `Set.difference` currentMembers
+                 in Set.null unexpectedMembers ?! UnexpectedCommitteeMembers unexpectedMembers
+              checkExpirationEpoch =
+                let maxTermLimit = pp ^. ppCommitteeTermLimitL
+                    invalidMembers =
+                      Map.filter (> EpochNo (fromIntegral maxTermLimit) + currentEpoch) $ membersToAdd
```
lehins: This check is too restrictive, since it will take a few epochs to vote on such proposal. In the matter of fact, I would prefer for the `UpdateCommittee` action to contain number of epochs that a committee member will be active for, rather than the absolute Epoch number of when a committee member expires:

@WhatisRT I couldn't find in the spec how we enforce [`ccTermLimit`](https://github.com/input-output-hk/formal-ledger-specifications/blob/7e2e805a9b4db18e73fff75aa6865d65213f004d/src/Ledger/PParams.lagda#L74), but that's how I would implement it. Does this align with your plans?

Also, redundant `$` and `EpochNo` has a `Num` instance, so this can be simplified a bit:
```suggestion
                      Map.filter (> fromIntegral maxTermLimit) membersToAdd
```

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Gov.hs
```diff
@@ -262,3 +303,19 @@ govTransition = do
 
 instance Inject (ConwayGovPredFailure era) (ConwayGovPredFailure era) where
   inject = id
+
+updatedCommittee ::
+  Set.Set (Credential 'ColdCommitteeRole (EraCrypto era)) ->
+  Map.Map (Credential 'ColdCommitteeRole (EraCrypto era)) EpochNo ->
+  UnitInterval ->
+  StrictMaybe (Committee era) ->
+  StrictMaybe (Committee era)
```
lehins: This function always returns new committee, so if SNOthing case is not possible, this function should not have the `StrictMaybe` return type.
```suggestion
  Committee era
```

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Gov.hs
```diff
@@ -262,3 +303,19 @@ govTransition = do
...
+updatedCommittee membersToRemove membersToAdd newQuorum committee =
+  case committee of
+    SNothing -> SJust (Committee membersToAdd newQuorum)
+    SJust (Committee currentMembers _) ->
+      SJust
+        ( Committee
+            (currentMembers `Map.withoutKeys` membersToRemove `Map.union` membersToAdd)
+            newQuorum
+        )
```
lehins: Firstly, `union` is left biased, therefore we want to use the new EpochNo from the `membersToAdd`, rather than the values from the current state of `currentMembers`
Secondly, using a let binding improves formatting a bit :slightly_smiling_face: 
```suggestion
      let newCommitteeMembers = 
            Map.union membersToAdd (currentMembers `Map.withoutKeys` membersToRemove)
        in Committee newCommitteeMembers newQuorum
```

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Gov.hs
```diff
@@ -232,17 +258,32 @@ govTransition = do
...
+                        ?! NewCommitteeSizeTooSmall committeeSize minCommitteeSize
+              checkUnexpectedMembers =
+                let currentMembers = Map.keysSet $ foldMap' (^. committeeMembersL) committee
+                    unexpectedMembers = membersToRemove `Set.difference` currentMembers
+                 in Set.null unexpectedMembers ?! UnexpectedCommitteeMembers unexpectedMembers
+              checkExpirationEpoch =
+                let maxTermLimit = pp ^. ppCommitteeTermLimitL
+                    invalidMembers =
+                      Map.filter (> EpochNo (fromIntegral maxTermLimit) + currentEpoch) $ membersToAdd
+                 in Map.null invalidMembers ?! InvalidCommitteeExpirationEpoch invalidMembers
```
lehins: It's good to expected value of the term limit to the predicate failure. Also renaming it could make sense to reflect the semantics a bit better
```suggestion
                 in Map.null invalidMembers ?! InvalidCommitteeTermLimit maxTermLimit invalidMembers
```

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Gov.hs
```diff
@@ -232,17 +288,32 @@ govTransition = do
...
+            _ <- checkUnexpectedMembers
+            checkExpirationEpoch
+            where
+              checkCommitteeSize =
+                let minCommitteeSize = pp ^. ppMinCommitteeSizeL
+                    newCommittee = updatedCommittee membersToRemove membersToAdd qrm committee
+                    committeeSize = fromIntegral . Map.size $ foldMap' (^. committeeMembersL) newCommittee
+                 in committeeSize
+                      >= minCommitteeSize
+                        ?! NewCommitteeSizeTooSmall committeeSize minCommitteeSize
```
lehins: I do like this check, however we can't apply it this way, because there is a chance, that there is a previous `UpdateCommitte` governance action, that this one relies on, which haven't been enacted yet.

We would only be able to implement this check if we also implement #3655 and resolve all of the proposals that are currently in-flight.

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Enact.hs
```diff
@@ -86,10 +87,10 @@ enactmentTransition = do
         st
           & ensCommitteeL .~ SNothing
           & ensPrevCommitteeL .~ SJust (PrevGovActionId govActionId)
-    NewCommittee _ _ committee ->
+    UpdateCommittee _ membersToRemove membersToAdd newQuorum -> do
       pure $
         st
-          & ensCommitteeL .~ SJust committee -- TODO: check old members
+          & ensCommitteeL %~ updatedCommittee membersToRemove membersToAdd newQuorum
```
lehins: We need to do all the checks (expiration and members to remove) during ratification. In other words, predicate checks during proposal in GOV rule do not guarantee that upon enactment they are still valid. Here are reasons why they might become invalid:

* `ppCommitteeTermLimitL` could have been updated between proposal and unactment
* another `UpdateCommittee` got enacted before this one, which affected the current committee state. (I've mentioned this case in another comment of mine)

I would suggest creating `verifyUpdateCommittee` and a custom failure type that indicates what is wrong, which could then potentially be used in the GOV rule as a predicate failure, as well as in RATIFY rule to prevent invalid `UpdateCommittee`. This whole logic could potentially live in a single `updateCommittee` function, if that would improve performance, but that is definitely not a requirement for this GOvAction, especially not for the initial version of it.

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Epoch.hs
```diff
@@ -302,6 +316,8 @@ epochTransition = do
       & esGovernanceL . cgGovSnapshotsL .~ newGov
       -- Move donations to treasury
       & esAccountStateL . asTreasuryL <>~ donations
+      -- remove hot keys of commitee members that were removed
```
lehins:
```suggestion
      -- remove hot keys of committee members that were removed
```

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Epoch.hs
```diff
@@ -289,9 +304,8 @@ epochTransition = do
         , prevDRepsState =
             es'
               ^. esLStateL . lsCertStateL . certVStateL . vsDRepsL
-        , prevCommitteeState =
-            es'
-              ^. esLStateL . lsCertStateL . certVStateL . vsCommitteeStateL
+        , -- committee state snapshot will not contain members that were removed in Enact
+          prevCommitteeState = updatedCommitteeState
```
lehins: I believe previous implementation was actually correct. We need to retain the committee state from the previous epoch, in order for committee votes to be valid
```suggestion
        , prevCommitteeState =
            es'
              ^. esLStateL . lsCertStateL . certVStateL . vsCommitteeStateL
```

If I am wrong we'll fix it later, since it should match the DRep implementation above (which is currently unused).


In any case we'll reevaluate the whole time travel thing later, especially after #3719 is merged. For now I think it would be best to revert the implementation, even if I am wrong in my understanding of this flow.

lehins (changes requested):
This is a very nice piece of work. Some fixes are needed, but overall looks great. 

Also needs a rebase on master.

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Gov.hs
WhatisRT: I agree that it's a little awkward that it's an absolute epoch instead of a number of epochs, but it is that way for a good reason: you want full predictability here. For example, if there's going to be a schedule where one third of the CC is replaced each year it's much better to align the schedule exactly, instead of having a few weeks where some time out but others don't.

Two notes about `ccTermLimit`:
- First, in my PR to the CIP I've renamed it to `ccMaxTermLength`, because that's what it actually is.
- My plan is to enforce it in `ENACT`. The reason is that this check can become successful over time, and we only know when it'll be successful in `ENACT`.

WhatisRT:
`NewCommittee` shouldn't fail in `GOV` if the removed members aren't in the CC. The reason for this is simple: you only know the current `EnactState`, but not necessarily the one that this GA is actually supposed to apply to. You could argue that this logic should then go into `ENACT`, and I originally wanted to put it there, but I've removed it again. The reasoning is here: https://github.com/input-output-hk/formal-ledger-specifications/pull/206#discussion_r1322786893

Long story short, I think we should only do this check if we keep track of in-flight GAs & validate new ones against the current state and the ones in the queue.

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Gov.hs
lehins: > if there's going to be a schedule where one third of the CC is replaced each year it's much better to align the schedule exactly

Wouldn't it be as easy as submitting an `UpdateCommittee` GovAction that only updates the lifetime of 1/3 of the committee to the same duration?

lehins:
> NewCommittee shouldn't fail in GOV if the removed members aren't in the CC. The reason for this is simple: you only know the current EnactState, but not necessarily the one that this GA is actually supposed to apply to.

@WhatisRT Yes, I brought up the same point here: https://github.com/input-output-hk/cardano-ledger/pull/3721#discussion_r1323240959

One more reason in favor of #3655

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Epoch.hs
teodanciu: It might actually not even matter so much, since it's only removing information that shouldn't be used anyway  (since the vote of the removed members shouldn't even get to be looked up in this map), but  I agree, better to do less if unsure and add to it later :+1:

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Epoch.hs
teodanciu: I need to remember that the value type don't have to coincide for `Map.difference` ! Thanks

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Gov.hs
teodanciu: 💯

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Gov.hs
teodanciu: Eeek!  Thank you

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Gov.hs
teodanciu: Oops, scala!

teodanciu:
@lehins I think the latest covers the items that came up after the meeting yesterday, if you would like to take another look. 

* Prevent expiry epochNo less or equal to current epochNo in GOV
* Accept expiry from the past on Enactment
* Delay expiry that is too long
* Prevent intersection of removed and added/updated in GOV

lehins (approved):
Beautiful! :+1:

PR #3725, opened 2023-09-11
Fix mistaken use of dollar sign in cddl files

teodanciu:
# Description

At the moment , `cddl generate` that we use for cddl roundtrips does NOT generate values for these types that we prefixed with `$`.
The types for which values ARE generated correctly that are prefixed with $ (eg: $hash28, $vkey), are also defined in the same way, prefixed by dollar.

It's not very clear from the spec that this should be the case, but running `cddl generate` makes it  clear that it's simply ignoring $required_signers,  $treasury_withdrawals_action and $new_committee when defined in the way they are currently in the cddl. 

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/eras/alonzo/test-suite/cddl-files/alonzo.cddl
+++ b/eras/alonzo/test-suite/cddl-files/alonzo.cddl
@@ -68,7 +68,7 @@ transaction_body =
  , ? 15 : network_id             ; New
  }
 
-required_signers = set<$addr_keyhash>
+required_signers = set<addr_keyhash>
 
 transaction_input = [ transaction_id : $hash32
                     , index : uint
--- a/eras/babbage/test-suite/cddl-files/babbage.cddl
+++ b/eras/babbage/test-suite/cddl-files/babbage.cddl
@@ -70,7 +70,7 @@ transaction_body =
   , ? 18 : set<transaction_input> ; reference inputs; New
   }
 
-required_signers = set<$addr_keyhash>
+required_signers = set<addr_keyhash>
 
 transaction_input = [ transaction_id : $hash32
                     , index : uint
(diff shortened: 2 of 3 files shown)


teodanciu:
If you want to test this yourself, do this:

1.  Create a file a.cddl with this content:
```
required_signers = nonempty_set<$addr_keyhash>
addr_keyhash          = $hash28
$hash28 /= bytes .size 28
nonempty_set<a> = [+ a]
```
2. run:  `nix-shell -p  cddl`
3. run: `cddl a.cddl generate 1`
4. observe the error
5.  change first line in a.cddl to `required_signers = nonempty_set<addr_keyhash>` 
6. run: `cddl a.cddl generate 1`

The reason we don't get an error in tests is  because it's optional this in transaction_body:
` , ? 14 : required_signers`

lehins (approved):
If only CDDL had a type checker :wink: 

Very nice find! Thank you.

teodanciu:
Or for GovAction:
```
gov_action =
  [
   new_committee
  //  treasury_withdrawals_action
  ]

set<a> = [a]
$hash28 /= bytes .size 28
$hash32 /= bytes .size 32
reward_account = bytes
coin = uint
unit_interval = #6.30([uint, uint])

epoch = uint

addr_keyhash           = $hash28
scripthash            = $hash28

credential =
  [  0, addr_keyhash
  // 1, scripthash
  ]
committee_cold_credential = credential
committee = [{ committee_cold_credential => epoch }, unit_interval]


gov_action_id =
  [ transaction_id   : $hash32
  , gov_action_index : uint
  ]

new_committee = (4, gov_action_id / null, set<committee_cold_credential>, committee)
treasury_withdrawals_action = (2, { reward_account => coin })
```

With  the current version in master, which is:
```
new_committee = (4, gov_action_id / null, set<$committee_cold_credential>, committee)
treasury_withdrawals_action = (2, ${ reward_account => coin })
```

`cddl b.cddl generate` throws an error.
If we remove the $ as my PR does, then values are generated as expected

JaredCorduan:
wow, great catch!

PR #3726, opened 2023-09-12
Conway and other eras roundtrip tests

### LINKED ISSUE #3158: Conway serialization tests
Write round-trip property tests for the new transaction body serialization.

lehins:
# Description

This PR organizes CBOR RoundTrip tests into a nice hierarchy

It captures all of the Conway types and most of the types from other eras, but not all of them. Fixes: #3158 

One important one is still missing is CBOR RoundTrip tests for PredicateFailures, but we can take care of it as a separate task.

Besides tests there other notable changes:

* Fix roundtrip serialization for `ProposalsSnapshot`
* Improve `ToJSON/FromJSON` implementation of `Credential`
* Improvements to RoundTrip tests failure reporting, namely addition of diffing for re-serialization testing

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Snapshots.hs
+++ b/eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Snapshots.hs
@@ -17,7 +17,7 @@ module Cardano.Ledger.Conway.Governance.Snapshots (
   snapshotGovActionStates,
 ) where
 
-import Cardano.Ledger.Binary (DecCBOR (..), EncCBOR (..))
+import Cardano.Ledger.Binary (DecCBOR (..), DecShareCBOR (..), EncCBOR (..))
 import Cardano.Ledger.Conway.Core (Era (..), EraPParams)
 import Cardano.Ledger.Conway.Governance.Procedures (
   GovActionId,
@@ -65,8 +65,9 @@ instance Default (ProposalsSnapshot era) where
 instance EraPParams era => EncCBOR (ProposalsSnapshot era) where
   encCBOR = encCBOR . snapshotActions
 
-instance EraPParams era => DecCBOR (ProposalsSnapshot era) where
-  decCBOR = fromGovActionStateSeq <$> decCBOR
+-- TODO: Implement Sharing: https://github.com/input-output-hk/cardano-ledger/issues/3486
+instance EraPParams era => DecShareCBOR (ProposalsSnapshot era) where
+  decShareCBOR _ = fromGovActionStateSeq <$> decCBOR
 
 snapshotInsertGovAction ::
   GovActionState era ->
--- a/eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Snapshots.hs
+++ b/eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Snapshots.hs
@@ -36,6 +36,7 @@ import Data.Foldable (Foldable (..))
 import Data.List (sort)
 import qualified Data.Map as Map
 import Data.Map.Strict (Map)
+import Data.Maybe (fromMaybe)
 import Data.Sequence.Strict (StrictSeq (..))
 import Data.Set (Set)
 import qualified Data.Set as Set
@@ -83,16 +84,12 @@ snapshotInsertGovAction gas@GovActionState {gasId} ps@ProposalsSnapshot {..}
 snapshotActions ::
   ProposalsSnapshot era ->
   StrictSeq (GovActionState era)
-snapshotActions ProposalsSnapshot {..} = go psProposalOrder
+snapshotActions ProposalsSnapshot {..} = toGovAction <$> psProposalOrder
   where
-    go (gaId :<| gaIds) =
-      case Map.lookup gaId psGovActionStates of
-        Just v -> v :<| go gaIds
-        Nothing ->
-          error $
-            "Impossible: ProposalsSnapshot invariant is not maintained: "
-              <> show gaId
-    go Empty = Empty
+    toGovAction gaId =
+      fromMaybe
+        (error $ "Impossible: ProposalsSnapshot invariant is not maintained: " <> show gaId)
+        (Map.lookup gaId psGovActionStates)
 
 snapshotIds ::
   ProposalsSnapshot era ->
@@ -152,9 +149,7 @@ snapshotLookupId gId ProposalsSnapshot {psGovActionStates} =
 fromGovActionStateSeq ::
   StrictSeq (GovActionState era) ->
   ProposalsSnapshot era
-fromGovActionStateSeq (x :<| xs) =
-  snapshotInsertGovAction x (fromGovActionStateSeq xs)
-fromGovActionStateSeq Empty = def
+fromGovActionStateSeq = foldl' (flip snapshotInsertGovAction) def
 
 -- | Internal function for checking if the invariants are maintained
 isConsistent_ :: ProposalsSnapshot era -> Bool
--- a/libs/cardano-ledger-core/src/Cardano/Ledger/Core/Era.hs
+++ b/libs/cardano-ledger-core/src/Cardano/Ledger/Core/Era.hs
@@ -30,6 +30,7 @@ module Cardano.Ledger.Core.Era (
   eraProtVerHigh,
   toEraCBOR,
   fromEraCBOR,
+  fromEraShareCBOR,
   eraDecoder,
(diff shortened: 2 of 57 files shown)


lehins:
Still needs a changeklog update and git history cleanup, but other than that is ready for review.

libs/cardano-ledger-core/src/Cardano/Ledger/Credential.hs
```diff
@@ -96,20 +96,20 @@ instance NoThunks (Credential kr c)
 instance Crypto c => ToJSON (Credential kr c) where
   toJSON (ScriptHashObj hash) =
     Aeson.object
-      [ "script hash" .= hash
+      [ "scripthash" .= hash
```
teodanciu: This will make the existing genesis file for shelley (at least) incompatible, wont' that be a problem?

lehins: Why do you think that? There aren't any credentials in ShelleyGenesis.
https://github.com/input-output-hk/cardano-node/blob/master/configuration/cardano/mainnet-shelley-genesis.json

teodanciu: Ah interesting.  In the genesis file in our repo there is a "staking" object with "stake" and "pools":  https://github.com/input-output-hk/cardano-ledger/blob/master/eras/shelley/test-suite/test/Golden/ShelleyGenesis#L44

They are coming from `ShelleyGenesisStaking`: https://github.com/input-output-hk/cardano-ledger/blob/master/eras/shelley/impl/src/Cardano/Ledger/Shelley/Genesis.hs#L115

I just checked the config in sanchonet and both of these appear empty: 
```
 "staking": {
        "pools": {},
        "stake": {}
    },
```
I guess the cardano-node repo is the most relevant source.

lehins: Aha. That's because those two fields are hacks :smile: they aren't really part of real ShelleyGenesis. They are used for testing and benchmarking and I am working on extracting those two fields (`initialFunds` and `staking`) out of ShelleyGenesis

lehins: That being said, I think it will be good for backwards compatibility to support both versions: with space and without space when parsing fields

I'll make this change

teodanciu: Ah I see. Good to know, thanks for the explanation :+1:

libs/cardano-ledger-core/src/Cardano/Ledger/Credential.hs
```diff
@@ -118,22 +118,23 @@ instance Crypto c => FromJSONKey (Credential kr c) where
...
-            (badHex hash)
-            (pure . ScriptHashObj . ScriptHash)
-            (hashFromTextAsHex hash)
-        ["keyhash", hash] ->
-          maybe
-            (badHex hash)
-            (pure . KeyHashObj . KeyHash)
-            (hashFromTextAsHex hash)
+        [key, hash] -- for backwards compatibility we support with space and without
+          | key == "scriptHash" || key == "script hash" ->
```
teodanciu: Actually, to keep it compatible it would have to be 
`| key == "scriptHash" || key == "scripthash" ->` 
 because `script hash`  wasn't the  *key* implementation, though it was the regular json implementation.
What a mess, I'm afraid I didn't make it better when I wrote this `fromJSONKey` version which is different from the  `fromJSON` - I had to chose between consistent but bad (with the space in it) or inconsistent but better. 
Anyway, glad you're cleaning it up and unifying it

libs/cardano-ledger-core/src/Cardano/Ledger/Core/Era.hs
```diff
@@ -207,8 +208,14 @@ fromEraCBOR :: forall era t s. (Era era, DecCBOR t) => Plain.Decoder s t
 fromEraCBOR = eraDecoder @era decCBOR
 {-# INLINE fromEraCBOR #-}
 
+-- | Convert a type that implements `DecCBOR` to plain `Plain.Decoder` using the lowest
```
teodanciu: Seems not in sync with the signature, where the constraint is a type that implements `DecShareCBOR`, unless the distinction is not relevant here

teodanciu: Nevermind, I just saw it's fixed in a later commit!

eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Snapshots.hs
```diff
@@ -153,9 +150,7 @@ snapshotLookupId gId ProposalsSnapshot {psGovActionStates} =
 fromGovActionStateSeq ::
   StrictSeq (GovActionState era) ->
   ProposalsSnapshot era
-fromGovActionStateSeq (x :<| xs) =
-  snapshotInsertGovAction x (fromGovActionStateSeq xs)
```
teodanciu, line 156: Isn't foldl' processing the list in a similar way like this recursive function:  the last one in the list would be the first one to be inserted?

libs/cardano-ledger-core/src/Cardano/Ledger/Credential.hs
lehins: Ah, crap, I meant to do this for FromJSON. Thank you for catching it.

We don't really need to worry about backwards compatibility for FromJSONKey, since "keyhash" version was barely released and no-one really relies on it

eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Snapshots.hs
lehins: Yes, except, `snapshotInsertGovAction` appends elements at the end of the sequence, instead of at the beginning, which resulted in sequence being reversed every time it was deserialized :grin:

libs/cardano-ledger-core/src/Cardano/Ledger/Credential.hs
lehins: Fixed

eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Snapshots.hs
teodanciu: Ah yeah, the original function looks more like the shape of `foldr`, which explains the difference in order.   Nice subtle bug :smile:

libs/cardano-ledger-core/testlib/Test/Cardano/Ledger/Core/Binary/RoundTrip.hs
```diff
@@ -0,0 +1,197 @@
...
+  , Arbitrary (TxWits era)
+  , Arbitrary (TxAuxData era)
+  , Arbitrary (Value era)
+  , Arbitrary (CompactForm (Value era))
+  , Arbitrary (Script era)
+  , Arbitrary (PParams era)
+  , Arbitrary (PParamsUpdate era)
+  ) =>
+  Spec
+roundTripCoreEraTypesSpec = do
```
teodanciu, line 176: Is this function meant to be invoked in each era (in the respective RoundTrip test) ?

teodanciu (approved):
Great to have a systematic way of doing these roundtrip tests. 

I don't understand well the part about the Sharing, but maybe one Watercooler we could go briefly over it, if we have time .

Looks good to me

libs/cardano-ledger-core/testlib/Test/Cardano/Ledger/Core/Binary/RoundTrip.hs
lehins: Yes. And it is

teodanciu: Ah yes! Via roundTripShelleyCommonSpec and roundTripAlonzoCommonSpec. Sorry, I missed that :+1:

eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Snapshots.hs
Soupstraw: Nice catch! :sweat_smile:

PR #3727, opened 2023-09-13
Add DRep activity check to the PParams wellformedness checks

### LINKED ISSUE #3724: Add a PParam check for DRep expiry
`ppDRepActivity` should be greater or equal to `ppGovActionExpiration`

This check can only be implemented if we enforce a linear history of proposals and we can predict the value of PParams before the enactment of a proposal. 

Requires #3655 and #4038

This potentially can be enforced by the guardrails script. Which will need support #4038 and the value of current PParams values that will be during the enactment of the proposal passed to the script context

Soupstraw:
# Description

closes  #3724

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/eras/conway/impl/CHANGELOG.md
+++ b/eras/conway/impl/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## 1.9.0.0
 
+* Replace `ppuWellFormed` with `ppWellFormed`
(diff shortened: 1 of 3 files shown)


lehins:
This WellFormedness check is no a true check, since it requires knowledge of the current state. This will be removed from the spec.

Moreover, we can't enforce such check unless we implement #3655, since there might be proposals that haven't been enacted yet that affect current PParams.

PR #3728, opened 2023-09-14
STS tests based on constraints

MaximilianAlgehed:
# Description

This PR introduces the test setup for testing STS rules using the constraint library. It's a WIP effort but we have a running test for a simple STS rule: GOVCERT.

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/libs/cardano-ledger-test/cardano-ledger-test.cabal
+++ b/libs/cardano-ledger-test/cardano-ledger-test.cabal
@@ -76,6 +76,7 @@ library
         Test.Cardano.Ledger.Generic.Updaters
         Test.Cardano.Ledger.NoThunks
         Test.Cardano.Ledger.Rational
+        Test.Cardano.Ledger.STS
         Test.Cardano.Ledger.TestableEra
         Test.Cardano.Ledger.ValueFromList
         Test.Cardano.Ledger.Tickf
--- a/libs/cardano-ledger-test/src/Test/Cardano/Ledger/STS.hs
+++ b/libs/cardano-ledger-test/src/Test/Cardano/Ledger/STS.hs
@@ -1,0 +1,370 @@
+{-# LANGUAGE DataKinds #-}
+{-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE FlexibleInstances #-}
+{-# LANGUAGE GADTs #-}
+{-# LANGUAGE MultiParamTypeClasses #-}
+{-# LANGUAGE RankNTypes #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeApplications #-}
+{-# LANGUAGE TypeFamilies #-}
+{-# LANGUAGE UndecidableInstances #-}
+{-# LANGUAGE ViewPatterns #-}
+
+module Test.Cardano.Ledger.STS where
+
+import Cardano.Crypto.Hash.Class (sizeHash)
+import Cardano.Ledger.Address
+import Cardano.Ledger.BaseTypes
+import Cardano.Ledger.CertState
+import Cardano.Ledger.Conway
+import Cardano.Ledger.Conway.Core
+import Cardano.Ledger.Conway.PParams
+import Cardano.Ledger.Conway.Rules
+import Cardano.Ledger.Conway.TxCert
+import qualified Cardano.Ledger.Crypto as CC (Crypto (HASH))
+import Cardano.Ledger.PoolParams
+import Cardano.Ledger.Shelley.Rules hiding (epochNo, slotNo)
+import qualified Cardano.Ledger.UMap as UM
+import Control.State.Transition.Extended
+import qualified Data.ByteString as BS
+import Data.Default.Class (Default (def))
+import Data.Foldable
+import qualified Data.Map as Map
+import Data.Maybe
+import qualified Data.Set as Set
+import Lens.Micro ((^.))
+import Test.Cardano.Ledger.Alonzo.AlonzoEraGen ()
+import Test.Cardano.Ledger.Alonzo.EraMapping ()
+import Test.Cardano.Ledger.Constrained.Ast
+import Test.Cardano.Ledger.Constrained.Classes
+import Test.Cardano.Ledger.Constrained.Env
+import Test.Cardano.Ledger.Constrained.Monad
+import Test.Cardano.Ledger.Constrained.Preds.CertState
+import Test.Cardano.Ledger.Constrained.Preds.PParams
+import Test.Cardano.Ledger.Constrained.Preds.Universes
+import Test.Cardano.Ledger.Constrained.Rewrite
+import Test.Cardano.Ledger.Constrained.Shrink
+import Test.Cardano.Ledger.Constrained.Solver
+import Test.Cardano.Ledger.Constrained.TypeRep
+import Test.Cardano.Ledger.Constrained.Vars hiding (delegations, rewards)
+import Test.Cardano.Ledger.Generic.Proof
+import Test.Cardano.Ledger.Shelley.Utils
+import Test.QuickCheck
+import Test.Tasty
+import Test.Tasty.QuickCheck
+
+------------------------------------------------------------------------
+-- Generator Setup
+------------------------------------------------------------------------
+
+applySubst :: Subst era -> [Pred era] -> [Pred era]
+applySubst = map . substPredWithVarTest
+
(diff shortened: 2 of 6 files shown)


eras/conway/impl/cardano-ledger-conway.cabal
```diff
@@ -56,7 +57,6 @@ library
         Cardano.Ledger.Conway.Rules.Tickf
         Cardano.Ledger.Conway.Rules.Utxos
         Cardano.Ledger.Conway.Rules.Utxow
-        Cardano.Ledger.Conway.TxAuxData
```
lehins: Why the move? It breaks the alphabetical order

libs/cardano-ledger-test/cardano-ledger-test.cabal
```diff
@@ -138,7 +139,8 @@ library
         time,
         transformers,
         unordered-containers,
-        vector
+        vector,
+        lens
```
lehins: We don't use `lens`. We use `microlens` on this project
```suggestion
```

libs/cardano-ledger-test/src/Test/Cardano/Ledger/STS.hs
```diff
@@ -0,0 +1,220 @@
...
+genUniverse = generate (genSubstFromConstraints (Conway Standard) standardOrderInfo (universePreds def (Conway Standard)))
+
+-- NOTE: these tests can be slow (~10 seconds) because it takes time to generate
+-- the initial universe. Reducing the number of tests will NOT help.
+tests_STS :: TestTree
+tests_STS =
+  withResource genUniverse (const $ pure ()) $ \ io_subst ->
+  testGroup "STS property tests"
+    [ testProperty "prop_GOVCERT" $ ioProperty (prop_GOVCERT <$> io_subst)
+    ]
```
MaximilianAlgehed: @TimSheard this is where we re-use the universe we generate. The nice thing here is that `io_subst` can be re-used in all the different properties.

libs/cardano-ledger-test/src/Test/Cardano/Ledger/STS.hs
```diff
@@ -0,0 +1,219 @@
...
+{-# LANGUAGE FlexibleInstances #-}
+{-# LANGUAGE GADTs #-}
+{-# LANGUAGE MultiParamTypeClasses #-}
+{-# LANGUAGE RankNTypes #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeApplications #-}
+{-# LANGUAGE TypeFamilies #-}
+{-# LANGUAGE UndecidableInstances #-}
+{-# LANGUAGE ViewPatterns #-}
+{-# OPTIONS_GHC -fno-warn-orphans #-}
```
lehins: I don't see any instances in this module
```suggestion
```

libs/cardano-ledger-test/src/Test/Cardano/Ledger/STS.hs
```diff
@@ -0,0 +1,219 @@
...
+
+module Test.Cardano.Ledger.STS where
+
+import Control.Lens ((^.))
```
lehins:
```suggestion
import Lens.Micro ((^.))
```

libs/cardano-ledger-test/src/Test/Cardano/Ledger/STS.hs
```diff
@@ -0,0 +1,219 @@
...
+    , Right v <- runTyped (runTerm env tm) ->
+        Just (Name x, Payload rep v acc)
+  _ -> Nothing
+  where
+    known = isJust . flip findName env
+    unknown = not . known
+    knownTerm :: forall t. Term era t -> Bool
+    knownTerm = all known . varsOfTerm mempty
+    knownSums :: forall r. [Sum era r] -> Bool
+    knownSums = all known . foldl varsOfSum mempty
```
lehins: Avoiding space leaks even in tests is a good idea :slightly_smiling_face: 
```suggestion
    knownSums = all known . foldl' varsOfSum mempty
```

lehins (changes requested):
That's a good start with STS. Only minor changes are needed.

eras/conway/impl/cardano-ledger-conway.cabal
MaximilianAlgehed: Whoops. Probably a rebase artifact.

MaximilianAlgehed:
@lehins I think this should be ready to merge

libs/cardano-ledger-test/src/Test/Cardano/Ledger/STS.hs
```diff
@@ -0,0 +1,219 @@
...
+  case result of
+    Left errs -> error $ unlines errs
+    Right subst -> pure subst
+
+shrinkFromConstraints :: Era era => OrderInfo -> [Pred era] -> RootTarget era t t -> t -> [t]
+shrinkFromConstraints order cs target val = do
+  let env = saturateEnv (getTarget val target emptyEnv) cs
+  graph <- monadTyped $ compile order $ rewrite cs
+  env' <- shrinkEnv graph env
+  monadTyped $ runTarget env' target
```
TimSheard: Be nice if we had some comments here, something like
  let env = saturateEnv (getTarget val target emptyEnv) cs
  -- Build a new env, with just the variables in the target, where the vars are bound to the values extracted from 'val'
 graph <- monadTyped $ compile order $ rewrite cs
-- Build a new graph from the original constraints
  env' <- shrinkEnv graph env
-- Shrink each of the values bound in the env, by using the graph to determine what other variables they depended on
 monadTyped $ runTarget env' target
-- Rebuild the 'val' using the target and the new env with the shrunk  values.

libs/cardano-ledger-test/src/Test/Cardano/Ledger/STS.hs
```diff
@@ -0,0 +1,219 @@
...
+saturateEnv env0 preds = go env0 preds
+  where
+    go env [] = env
+    go env (p : ps)
+      | Just (x, v) <- solveUnknown env p = saturateEnv (storeName x v env) preds
+      | otherwise = go env ps
+
+solveUnknown :: forall era. Era era => Env era -> Pred era -> Maybe (Name era, Payload era)
+solveUnknown env p = case p of
+  SumsTo _ (Var x@(V _ rep acc)) EQL sums
```
TimSheard: lots of missing cases when we pattern match against 'p'.  Do we have a plan to cover these as well?

libs/cardano-ledger-test/src/Test/Cardano/Ledger/STS.hs
```diff
@@ -0,0 +1,219 @@
...
+        Left errors -> counterexample ("lookupTerm: " ++ show errors) False
+        Right val -> property $ cont val
+
+------------------------------------------------------------------------
+-- Properties
+------------------------------------------------------------------------
+
+prop_GOVCERT :: Subst (ConwayEra StandardCrypto) -> Property
+prop_GOVCERT sub =
+  forAllShrink (pParamsGen sub) (pParamsShrink sub) $ \(unPParams -> pParams) ->
```
TimSheard: Try the DELEG, POOL, or UTXO rule. They will have some interesting properties

TimSheard (commented):
The goal of this PR is to prototype the use of the shrinking capabilities by testing that a rules maintain the constraints we used to generates a 'legitimate' state. We should add another rule, as suggested in the comments, where there are some real constraints. This way we get a better test of the capabilities.

Personally, I'd like to better understand how we shrink an environment using the constraints

libs/cardano-ledger-test/src/Test/Cardano/Ledger/STS.hs
MaximilianAlgehed: Yes, my plan is to implement these as they start to appear.

libs/cardano-ledger-test/src/Test/Cardano/Ledger/STS.hs
MaximilianAlgehed: Yes, I have them in my todo file :)

lehins (approved):
:+1: Thank you

MaximilianAlgehed:
@TimSheard I think I've addressed your comments

MaximilianAlgehed:
@lehins I added the `POOL` and `DELEG` rules here too. It might be worth while to merge this soon to avoid things changing beneath our feet.

PR #3729, opened 2023-09-14
DRep expiry update after a contiguous set of epochs with no proposals to vote on

### LINKED ISSUE #3723: Avoid DRep expiration when there are no governance proposals
Whenever there is not a single proposal a DRep can vote on, it is only fair that DReps should not expire.

The way to solve this would be to keep a counter in the state for how many epochs there was no proposals.

This counter would:
* be incremented every epoch when there is not a single proposal
* added to every DRep's expiry value upon addition of any proposal, whenever there previously was no proposal at all
* Reset to zero after it was added to each DRep's expiry
* When non-zero it should also be added to every DRep's expiry during an API query: https://github.com/input-output-hk/cardano-ledger/blob/8753f69f6ab2d9712d3fca6cce78a12ac14d6f1e/libs/cardano-ledger-api/src/Cardano/Ledger/Api/State/Query.hs#L88-L94

aniketd:
# Description

Closes #3723

1. Add `vsNumDormantEpochs` field to `VState`.
    1. This field counts the number of contiguous epochs in which the number of active governance proposals that can be voted on remained zero. It will be incremented in every `EPOCH` if the number of active governance actions that can be voted on continues to remain zero. It will be reset when a new governance proposal is submitted successfully
1. Prevent DRep expiry when no proposals to vote on.
    1. In EPOCH, we make sure to update the dormant-epochs counter every time we see that the previous epoch didn't have any governance proposals to  vote on.
    1. When there is a new governance proposal in a transaction AND the dormant-epochs counter is greater than zero, we
        1. Update the expiry for all DReps by adding the counter
        1. Reset the dormant-epochs counter
    1. We update Api.State.Query.queryDRepState to report the effective DRep expiry information.
    2. Add snapshotsGovStateL to ConwayEraGov
    3. Move asks out of loop in GO
1. Add ConwayFeatures test for preventing DRep  expiry

# Checklist

- [x] Self-reviewed the diff


eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Epoch.hs
```diff
@@ -162,6 +164,15 @@ returnProposalDeposits removedProposals =
   where
     updateUMap = returnProposalDepositsUMap removedProposals
 
+-- | When there have been zero governance proposals to vote on in the previous epoch
+-- increase the dormant-epoch counter by one.
+updateNumDormantEpochs :: GovState era ~ ConwayGovState era => LedgerState era -> LedgerState era
```
Soupstraw: Could this be replaced with `ConwayEraGov`?
```suggestion
updateNumDormantEpochs :: ConwayEraGov era => LedgerState era -> LedgerState era
```

libs/cardano-ledger-core/src/Cardano/Ledger/CertState.hs
```diff
@@ -312,6 +313,14 @@ data VState era = VState
        )
   , vsDRepDistr :: !(DRepDistr (EraCrypto era))
   , vsCommitteeState :: !(CommitteeState era)
+  -- | Number of contiguous epochs in which there are exactly zero
+  -- active governance proposals to vote on. It is incremented in every
+  -- EPOCH rule if the number of active governance proposals to It
```
teodanciu: This sentence looks like it's missing the last part?
`if the number of active governance proposals remains zero` ?

libs/cardano-ledger-core/src/Cardano/Ledger/CertState.hs
```diff
@@ -312,6 +313,14 @@ data VState era = VState
...
+  -- is reset to zero when a new governance action is successfully
+  -- proposed. vote on continues to be zero.
```
teodanciu: maybe `vote on continues to be zero.`  this is the continuation of the sentence from before :laughing:

libs/cardano-ledger-core/src/Cardano/Ledger/CertState.hs
```diff
@@ -312,6 +313,14 @@ data VState era = VState
...
+  -- We need this counter in order to bump DRep expiries through dormant
+  -- periods whenre DReps do not have an oppurtunity to vote on anything.
```
teodanciu: `when` 
`opportunity`

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Gov.hs
```diff
@@ -100,6 +108,17 @@ data GovEnv era = GovEnv
   , gePParams :: !(PParams era)
   }
 
+data GovRuleState era = GovRuleState
+  { grsGovSnapshots :: GovSnapshots era
+  , grsVState :: VState era
```
Soupstraw: Are we now storing `VState` in two places?

Soupstraw: ah, I suppose it's an ephemeral state

Soupstraw (approved):
Good job!

eras/conway/impl/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## 1.9.0.0
 
+* Prevent `DREP` expiry when there are no active governance proposals to vote on. #3729
+  * Add `updateNumDormantEpochs` function in `ConwayEPOCH` to update the dormant-epochs counter
+  * Add `VState` to the State for `ConwayGOV` by introducing a new `GovRuleState` type, and update `ConwayLEDGER` accordingly.
+  * Add a call to a new `updateDRepExpiry` function in the existing `addAction` function in `ConwayGOV`, that updates `DRep` expiries and resets the dormant-epochs counter to zero as soon as a new governance proposal is successfully submitted to vote on, after a contiguous set of dormant-epochs.
```
teodanciu: Not sure if this last entry  has to be in the CHANGELOG, since it doesn't affect the "surface"

eras/allegra/impl/cardano-ledger-allegra.cabal
```diff
@@ -52,7 +52,7 @@ library
         bytestring,
         cardano-crypto-class,
         cardano-ledger-binary >=1.0,
-        cardano-ledger-core >=1.6 && <1.7,
+        cardano-ledger-core >=1.6 && <1.8,
```
teodanciu: We need to bump the version of the package to `1.2.2.1` because of this

eras/alonzo/impl/cardano-ledger-alonzo.cabal
```diff
@@ -67,7 +67,7 @@ library
         cardano-ledger-allegra >=1.1,
         cardano-crypto-class,
         cardano-ledger-binary >=1.0.1,
-        cardano-ledger-core >=1.6 && <1.7,
+        cardano-ledger-core >=1.6 && <1.8,
```
teodanciu: We need to bump the version of the package to `1.4.1.1`

eras/allegra/impl/cardano-ledger-allegra.cabal
teodanciu: Oh, this already is done on master, your branch is out of date. Nevermind!

eras/alonzo/impl/cardano-ledger-alonzo.cabal
teodanciu: Sorry, haven't noticed the branch is out of date. The problem will get resolved by rebasing

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Epoch.hs
aniketd: Nope, since `ConwayEraGov` does not cover this constraint. We essentially need this constraint to be able to use `utxosGovStateL`, as `GovState` changes since `Shelley`.

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Gov.hs
aniketd: Yes, it is just yet another state-change, just like another other rule.

eras/conway/impl/CHANGELOG.md
aniketd: Hmm, I wonder myself :thinking:

lehins: No we don't need to add it. `addAction` is not even an exported function.

I understand that it chagnes the semantics of the RULE, but if we were to add every semantic change to every rule we make changelog would be huge :grin: 

We only need to add things into the changelog that would be useful to the downstream users, i.e. the "surface" people :smile:

aniketd: Noted! :memo:

Thanks @teodanciu and @lehins :raised_hands:

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Gov.hs
```diff
@@ -240,23 +261,42 @@ actionWellFormed ga = failureUnless isWellFormed $ MalformedProposal ga
...
       _ -> True
 
+-- | Called from within `addAction` to check if the proposal being added is the first one
+-- after a set of contiguous epochs (we need to check only the previous epoch to be sure)
+-- that had no active governance proposals to vote on, so that `DRep` expiry can be updated
+-- and the dormant-epochs counter can be reset to zero.
+-- Although this function is called on each proposal, it makes sure to update the `DRep`
+-- expiry only the first time that a new proposal is accepted in an epoch.
+updateDRepExpiry :: GovRuleState era -> GovRuleState era
+updateDRepExpiry grs@(GovRuleState govSnapshots vState) =
```
teodanciu: Just checking... it's not possible to do this in `Certs` rule, where we also update the expiry of Dreps after voting? 
If that was possible, then we wouldn't have to have the VState in the GovState (not saying that the latter is necessarily bad btw).

aniketd: > it's not possible to do this in Certs rule, where we also update the expiry of Dreps after voting?

I don't think so, because we don't have access to the `GovSnapshots` in `CERTS`. We'd have to make some unavailable state available to the target rule either way, if I'm not mistaken.

teodanciu: But do we necessarily need the GovSnapshots?   We have access to the proposals there, so couldn't we keep track that there is none ? 
Apologies if this doesn't make sense, it's likely I'm not understanding the solution (or problem) well enough , so feel free to ignore if it's absurd.

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Gov.hs
```diff
@@ -240,23 +261,42 @@ actionWellFormed ga = failureUnless isWellFormed $ MalformedProposal ga
...
+  let wasPrevEpochDormant = Seq.null . snapshotIds $ govSnapshots ^. prevGovSnapshotsL
+      isCurEpochDormantSoFar = Seq.null . snapshotIds $ govSnapshots ^. curGovSnapshotsL
+      numDormantEpochs = vState ^. vsNumDormantEpochsL
+      isNumDormantEpochsNonZero = numDormantEpochs /= 0
+   in if isNumDormantEpochsNonZero && isCurEpochDormantSoFar && wasPrevEpochDormant
```
lehins: We need to reset the counter and update DRep's expiry upon addition of a first proposal, which means previous state was empty and current is not.
```suggestion
   in if isNumDormantEpochsNonZero && not isCurEpochDormantSoFar && wasPrevEpochDormant
```

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Gov.hs
```diff
@@ -240,23 +261,42 @@ actionWellFormed ga = failureUnless isWellFormed $ MalformedProposal ga
...
+      applyProps grs' ((idx, ProposalProcedure {..}) :<| ps) = do
         let expectedDeposit = pp ^. ppGovActionDepositL
          in pProcDeposit
               == expectedDeposit
                 ?! ProposalDepositIncorrect pProcDeposit expectedDeposit
 
         runTest $ actionWellFormed pProcGovAction
 
-        expectedNetworkId <- liftSTS $ asks networkId
+        expectedNetworkId <- liftSTS $ asks networkId -- TODO: move this out of the recursion if possible
```
lehins: Good catch. It is possible. Just move it right below of were we get the `judgementContext` and before definition of `applyProps` (currently line 289)

lehins (approved):
Very nice!!! :+1:

lehins:
@aniketd Needs golden test adjustment, since we added a new field and some other test failures. Not sure what they are about

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Gov.hs
aniketd: Please correct me if I'm wrong, but my thought process behind this was: this is triggered _just before_ the processing of the first proposal, and not _after_, and so, the current epoch is still expected to be empty. It will become non-empty as soon as any proposal is added and continue to be non-empty, so if we check that it is non-empty, it will mean that it is applicable for all proposals except the first one. `isCurEpochDormantSoFar` will be `True` when it is empty, and we receive the first proposal action to add to the snapshot. It will be `False` for every proposal action that is submitted after the first one in the current epoch. We only want to act upon the first submission in an epoch, and so we check its truthiness.

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Gov.hs
aniketd: Will do! :raised_hands:

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Epoch.hs
Soupstraw: that doesn't sound right, `utxosGovState` just returns the `GovState` of the current era. As long as you use lenses that come from the `*EraGov` classes then it shouldn't be necessary to constrain the `GovState` to a concrete datatype.

Soupstraw: I think the constraint comes from `cgGovSnapshotsL`, which is specific to ConwayGovState. I think it's better to add this lens as a method of the `ConwayEraGov` typeclass and use it from there.

aniketd: > I think it's better to add this lens as a method of the ConwayEraGov typeclass and use it from there.

That's an awesome idea!

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Gov.hs
aniketd: @teodanciu and @lehins suggested significant improvements in this implementation and the result has been pushed. This conversation can be resolved, as it concerns a stale part of the implementation that has been removed completely.

lehins (approved):
Beautiful! :slightly_smiling_face:

PR #3742, opened 2023-09-19
Update fourmolu, ghcid and hls. Update haskellNix and iohkNix flakes

lehins:
# Description

Update tools

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/.github/workflows/haskell.yml
+++ b/.github/workflows/haskell.yml
@@ -142,8 +142,9 @@ jobs:
 
     - name: Install fourmolu
       run: |
+        FOURMOLU_VERSION="0.14.0.0"
         mkdir -p "$HOME/.local/bin"
-        curl -sL https://github.com/fourmolu/fourmolu/releases/download/v0.13.1.0/fourmolu-0.13.1.0-linux-x86_64 -o "$HOME/.local/bin/fourmolu"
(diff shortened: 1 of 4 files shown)


.github/workflows/haskell.yml
```diff
@@ -142,8 +142,9 @@ jobs:
 
     - name: Install fourmolu
       run: |
+        FOURMOLU_VERSION="0.14.0.0"
```
aniketd, line 145: I'm surprised this didn't require an `export` :slightly_smiling_face:

aniketd (approved):
Grand! :raised_hands:

.github/workflows/haskell.yml
lehins: Exporting is needed only if you want the environment variable be available in the subprocess. Here, however, we just do interpolation directly in the same bash session. See the difference:

```shell
> FOO="foo"
> echo "$FOO"
foo
> bash -c 'echo "$FOO"'

> export BAR="bar"
> echo "$BAR"
bar
> bash -c 'echo "$BAR"'
bar
```

aniketd: Ah! Got it! Thank you @lehins :bow:

PR #3743, opened 2023-09-19
Move DRepDistr from VState to ConwayGovState

TimSheard:
# Description

# Checklist

- [ ] Self-reviewed the diff


eras/conway/impl/src/Cardano/Ledger/Conway/Governance.hs
```diff
@@ -470,12 +503,20 @@ instance EraPParams era => NFData (RatifyState era)
...
 instance EraPParams era => NoThunks (RatifyState era)
 
+-- | The 'cgDRepDistr' field is a pulser that incrementally computes the stake distribution of the DReps
+--   over the Epoch following the close of voting at end of the previous Epoch. The pulser is created
+--   at the Epoch boundary, but does no work until it is pulsed in the 'NEWEPOCH' rule, whenever the
+--   system is NOT at the epoch boundary.
 data ConwayGovState era = ConwayGovState
   { cgGovSnapshots :: !(GovSnapshots era)
   , cgEnactState :: !(EnactState era)
+  , cgDRepDistr :: !(DRepDistr (EraCrypto era))
```
lehins: We can document individual fields
```suggestion
data ConwayGovState era = ConwayGovState
  { cgGovSnapshots :: !(GovSnapshots era)
  , cgEnactState :: !(EnactState era)
  -- | The 'cgDRepDistr' field is a pulser that incrementally computes the stake distribution of the DReps
  --   over the Epoch following the close of voting at end of the previous Epoch. The pulser is created
  --   at the Epoch boundary, but does no work until it is pulsed in the 'NEWEPOCH' rule, whenever the
  --   system is NOT at the epoch boundary.
  , cgDRepDistr :: !(DRepDistr (EraCrypto era))
```

eras/conway/impl/src/Cardano/Ledger/Conway/Governance.hs
```diff
@@ -727,3 +774,53 @@ votingDRepThresholdInternal pp isElectedCommittee action =
...
+epochStateIncrStakeDistrL = esLStateL . lsUTxOStateL . utxosStakeDistrL . credMapL
+
+epochStateRegDrepL ::
+  Lens'
+    (EpochState era)
+    (Map (Credential 'DRepRole (EraCrypto era)) (DRepState (EraCrypto era)))
+epochStateRegDrepL = esLStateL . lsCertStateL . certVStateL . vsDRepsL
+
+epochStateUMapL :: Lens' (EpochState era) (UMap (EraCrypto era))
+epochStateUMapL = esLStateL . lsCertStateL . certDStateL . dsUnifiedL
```
lehins: This lenses should not have been moved from: `Cardano.Ledger.Shelley.LedgerState.Types`, since they work directly with `EpochState`:

* epochStateStakeDistrL
* epochStateIncrStakeDistrL
* epochStateRegDrepL
* epochStateUMapL

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/NewEpoch.hs
```diff
@@ -148,8 +156,9 @@ newEpochTransition = do
       ) <-
     judgmentContext
   if eNo /= eL + 1
-    then pure nes
-    else do
+    then pure (nes & newEpochStateDRepDistrL %~ pulseDRepDistr)
+    else -- pure nes
+    do
```
lehins:
```suggestion
    else do
```

libs/cardano-ledger-api/src/Cardano/Ledger/Api/State/Query.hs
```diff
@@ -100,6 +101,7 @@ queryDRepState nes creds
 -- | Query DRep stake distribution. Note that this can be an expensive query because there
 -- is a chance that current distribution has not been fully computed yet.
 queryDRepStakeDistr ::
+  ConwayEraGov era =>
```
lehins: Unfortunately we can't use `ConwayEraGov` constraint here.

All queries must work for all eras starting with Shelley, even if they return empty results for some of them.

Which I think can be solved by adding a function that gets the distribution to the `EraGov` type class: 
```haskell
  getDRepDistr :: GovState era -> Map (DRep (EraCrypto era)) (CompactForm Coin)
```

It would return an empty `Map` for all eras and in conway it could use the new `drepDistrGovStateL` lens to get the distr

```suggestion
  EraGov era =>
```

lehins (commented):
My preliminary feedback. Other than that I think we are on the right track

eras/conway/impl/src/Cardano/Ledger/Conway/Governance.hs
TimSheard: ok

lehins (approved):
Perfect!!!

PR #3745, opened 2023-09-20
Constitutional Committee Ratification

### LINKED ISSUE #3444: CC vote always No, when the committee is too small
The constitutional committee will contain a group of members and a threshold for the proportion of yes votes required to pass governance actions. When a committee member retires (either by expiration or by actively submitting a retirement certificate), we still assume the same denominator as before the retirement for the purposes of determining if the threshold is met. If ever it becomes impossible to meet this threshold, the system automatically enters a state of no confidence.

For example, with four committee members and a threshold of 51%, if two members retire, the system enters a state of no confidence.

This logic needs to be present in the ratification rules.

### LINKED ISSUE #3701: Count Constitutional Committee votes during Ratification
Constitutional Committee votes are currently not counted.

Same as with DReps and StakePool votes we need to make sure GovAction has enough votes from ConstitutionalCommittee for ratification

teodanciu:
# Description

Adds constitutional committee ratification as a mandatory check in `Ratify`. 
Closes https://github.com/input-output-hk/cardano-ledger/issues/3701
Closes https://github.com/input-output-hk/cardano-ledger/issues/3444

The structure of the code is very similar to drep` ratification, so we could abstract it. 

Missing: 
* ~Fix ConwayFeatures test~  EDIT: Fixed  
* voting no when the committee is too small  (https://github.com/input-output-hk/cardano-ledger/issues/3444) 
*  ~there is a contradiction between[ a comment on the issue](https://github.com/input-output-hk/cardano-ledger/issues/3701#issuecomment-1706790738) and (what I think I see in ) the spec~  EDIT: clarified
   > This means that expired and resigned committee members still contribute to the ratio calculation, despite that their votes do not count. 
* ~some unseemly duplication in the test, should find a better way to set expired or resigned members with less noise~ EDIT: refactored to remove the duplication!

Might be just a problem of semantics, but in this PR, I wouldn't say that expired and resigned "contribute to the ratio" (unless we interpret that as:  because they're expired or resigned we don't count them at all..)

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/eras/conway/impl/CHANGELOG.md
+++ b/eras/conway/impl/CHANGELOG.md
@@ -1,7 +1,9 @@
 # Version history for `cardano-ledger-conway`
 
 ## 1.9.0.0
-
+* Add to `Ratify`:
+  * `committeeAccepted`
+  * `committeeAcceptedRatio`
 * Add `reCommitteeState` to `RatifyEnv`
 * Apply enacted `TreasuryWithdrawals` in `ConwayEPOCH` #3748
   * Add lenses `ensWithdrawalsL` and `ensTreasuryL`
--- a/eras/conway/impl/CHANGELOG.md
+++ b/eras/conway/impl/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## 1.9.0.0
 
+* Add `reCommitteeState` to `RatifyEnv`
 * Apply enacted `TreasuryWithdrawals` in `ConwayEPOCH` #3748
   * Add lenses `ensWithdrawalsL` and `ensTreasuryL`
 * Add PredicateFailure for current treasury value mismatch in tx body in LEDGER #3749
--- a/eras/conway/impl/src/Cardano/Ledger/Conway/Governance.hs
+++ b/eras/conway/impl/src/Cardano/Ledger/Conway/Governance.hs
@@ -166,6 +166,7 @@ import Cardano.Ledger.Conway.PParams (
   PParamGroup (..),
   PoolVotingThresholds (..),
   modifiedGroups,
+  ppCommitteeMinSizeL,
   ppDRepVotingThresholdsL,
   ppPoolVotingThresholdsL,
  )
@@ -696,28 +697,32 @@ votingStakePoolThresholdInternal pp isElectedCommittee action =
         TreasuryWithdrawals {} -> NoVotingAllowed
         InfoAction {} -> NoVotingThreshold
 
-isCommitteeVotingAllowed :: GovAction era -> Bool
+isCommitteeVotingAllowed :: ConwayEraPParams era => GovAction era -> Bool
 isCommitteeVotingAllowed =
-  isVotingAllowed . votingCommitteeThresholdInternal committee
+  isVotingAllowed . votingCommitteeThresholdInternal def committee
   where
     -- Information about presence of committe is irrelevant for knowing if voting is
     -- allowed or not
     committee = SNothing
 
 votingCommitteeThreshold ::
+  ConwayEraPParams era =>
   RatifyState era ->
   GovAction era ->
   StrictMaybe UnitInterval
 votingCommitteeThreshold ratifyState =
-  toRatifyVotingThreshold . votingCommitteeThresholdInternal committee
+  toRatifyVotingThreshold . votingCommitteeThresholdInternal pp committee
   where
     committee = ratifyState ^. rsEnactStateL . ensCommitteeL
+    pp = ratifyState ^. rsEnactStateL . ensCurPParamsL
 
 votingCommitteeThresholdInternal ::
+  ConwayEraPParams era =>
+  PParams era ->
   StrictMaybe (Committee era) ->
   GovAction era ->
   VotingThreshold
-votingCommitteeThresholdInternal committee = \case
+votingCommitteeThresholdInternal pp committee = \case
   NoConfidence {} -> NoVotingAllowed
   UpdateCommittee {} -> NoVotingAllowed
   NewConstitution {} -> threshold
@@ -728,8 +733,12 @@ votingCommitteeThresholdInternal committee = \case
   where
     threshold =
       case committeeQuorum <$> committee of
-        SJust t -> VotingThreshold t
-        SNothing -> NoVotingThreshold
+        -- if the committee size is smaller than the mnimimum given in pparams,
+        -- we treat it as if we had no committe
+        SJust t | committeeSize >= minSize -> VotingThreshold t
+        _ -> NoVotingThreshold
+    minSize = pp ^. ppCommitteeMinSizeL
+    committeeSize = fromIntegral $ Map.size $ foldMap' committeeMembers committee
 
 isDRepVotingAllowed ::
   ConwayEraPParams era =>
--- a/eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Epoch.hs
+++ b/eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Epoch.hs
@@ -405,6 +405,6 @@ instance
 
 updateCommitteeState :: StrictMaybe (Committee era) -> CommitteeState era -> CommitteeState era
 updateCommitteeState committee (CommitteeState creds) =
-  CommitteeState $ Map.difference creds members
+  CommitteeState $ Map.intersection creds members
   where
     members = foldMap' committeeMembers committee
--- a/eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Epoch.hs
+++ b/eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Epoch.hs
@@ -299,6 +299,7 @@ epochTransition = do
         , reDRepDistr = drepDistr
         , reCurrentEpoch = eNo - 1
         , reDRepState = vstate ^. vsDRepsL
+        , reCommitteeState = vstate ^. vsCommitteeStateL
         }
     ratSig = RatifySignal . snapshotActions . prevGovSnapshots $ cgGovSnapshots govSt
   RatifyState {rsRemoved, rsEnactState = rsEnactState'} <-
--- a/eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Ratify.hs
+++ b/eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Ratify.hs
@@ -120,13 +120,15 @@ instance
 -- We iterate over the committee, and incrementally construct the numerator and denominator,
 -- based on the votes and the committee state.
 committeeAccepted ::
+  ConwayEraPParams era =>
   RatifyState era ->
   RatifyEnv era ->
   GovActionState era ->
   Bool
 committeeAccepted rs RatifyEnv {reCommitteeState, reCurrentEpoch} GovActionState {gasCommitteeVotes, gasAction} =
   case votingCommitteeThreshold rs gasAction of
-    SNothing -> False -- this happens if we have no committee, in which case the committee vote is `no`
+    SNothing -> False -- this happens if we have no committee, or if the committee is too small,
+    -- in which case the committee vote is `no`
     SJust r ->
       -- short circuit on zero threshold, in which case the committee vote is `yes`
       r == minBound
--- a/eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Ratify.hs
+++ b/eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Ratify.hs
@@ -25,7 +25,7 @@ module Cardano.Ledger.Conway.Rules.Ratify (
 ) where
 
 import Cardano.Ledger.BaseTypes (BoundedRational (..), ShelleyBase, StrictMaybe (..))
-import Cardano.Ledger.CertState (DRepState (..))
+import Cardano.Ledger.CertState (CommitteeState, DRepState (..))
 import Cardano.Ledger.Coin (Coin (..), CompactForm (..))
 import Cardano.Ledger.Conway.Core
 import Cardano.Ledger.Conway.Era (ConwayENACT, ConwayRATIFY)
@@ -81,6 +81,7 @@ data RatifyEnv era = RatifyEnv
   , reDRepDistr :: !(Map (DRep (EraCrypto era)) (CompactForm Coin))
   , reDRepState :: !(Map (Credential 'DRepRole (EraCrypto era)) (DRepState (EraCrypto era)))
   , reCurrentEpoch :: !EpochNo
+  , reCommitteeState :: !(CommitteeState era)
   }
 
 deriving instance Show (RatifyEnv era)
--- a/eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Ratify.hs
+++ b/eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Ratify.hs
@@ -18,6 +18,8 @@ module Cardano.Ledger.Conway.Rules.Ratify (
   RatifyState (..),
   RatifyEnv (..),
   RatifySignal (..),
+  committeeAccepted,
+  committeeAcceptedRatio,
   dRepAccepted,
   dRepAcceptedRatio,
   reorderActions,
@@ -25,7 +27,7 @@ module Cardano.Ledger.Conway.Rules.Ratify (
 ) where
 
 import Cardano.Ledger.BaseTypes (BoundedRational (..), ShelleyBase, StrictMaybe (..))
-import Cardano.Ledger.CertState (CommitteeState, DRepState (..))
+import Cardano.Ledger.CertState (CommitteeState (csCommitteeCreds), DRepState (..))
 import Cardano.Ledger.Coin (Coin (..), CompactForm (..))
 import Cardano.Ledger.Conway.Core
 import Cardano.Ledger.Conway.Era (ConwayENACT, ConwayRATIFY)
@@ -36,8 +38,10 @@ import Cardano.Ledger.Conway.Governance (
   PrevGovActionIds (..),
   RatifyState (..),
   Vote (..),
+  ensCommitteeL,
   ensTreasuryL,
   rsEnactStateL,
+  votingCommitteeThreshold,
   votingDRepThreshold,
   votingStakePoolThreshold,
  )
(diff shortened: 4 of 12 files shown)


lehins:
> there is a contradiction betweenhttps://github.com/input-output-hk/cardano-ledger/issues/3701#issuecomment-1706790738 and (what I think I see in ) the spec:

Yeah, that comment on the ticket was wrong. I've fixed it. All expired and resigned CC members are considered as if they abstained:

https://github.com/input-output-hk/formal-ledger-specifications/blob/5feb58cd4f43aa331748503a2bfc9d2a04a1f015/src/Ledger/Ratify.lagda#L258-L261

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Epoch.hs
```diff
@@ -372,6 +372,6 @@ instance
 
 updateCommitteeState :: StrictMaybe (Committee era) -> CommitteeState era -> CommitteeState era
 updateCommitteeState committee (CommitteeState creds) =
-  CommitteeState $ Map.difference creds members
+  CommitteeState $ Map.restrictKeys creds (Map.keysSet members)
```
teodanciu: `Map.difference` doesn't do the right thing here, but can't think of a better way to do this : only keep in `creds` the  keys that are in `members` .. Is there a better way?

lehins: You are looking for `Map.intersection`

teodanciu: Yes, great! didn't realize this worked for different value types of Map. Thanks!

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Ratify.hs
```diff
@@ -103,10 +109,56 @@ instance
...
+  RatifyState era ->
+  RatifyEnv era ->
+  GovActionState era ->
+  Bool
+committeeAccepted rs RatifyEnv {reCommitteeState, reCurrentEpoch} GovActionState {gasCommitteeVotes, gasAction} =
+  case votingCommitteeThreshold rs gasAction of
+    SNothing -> False -- this happens if we have no committee, in which case the committee vote is `no`
+    SJust r ->
+      -- short circuit on zero threshold.
+      -- This happen if the action is NoConfidence or UpdateCommittee, in which case the committee vote is 'yes'
```
lehins: The case you mention is actually impossible, so we wouldn't need to short circuit on it.
```suggestion
      -- This can happen when quorum is set to 0, then committee vote is always 'yes'
```

eras/conway/impl/test/Test/Cardano/Ledger/Conway/CommitteeRatifySpec.hs
```diff
@@ -0,0 +1,330 @@
...
+  let
+    pairs = Map.toList members
+    CommitteeState {csCommitteeCreds} = committeeState
+    updatedMembers =
+      [ (ck, if expire then newEpochNo else epoch)
+      | (ck, epoch) <- pairs
+      , let expire = (maybeToList . join) (Map.lookup ck csCommitteeCreds) `List.isInfixOf` hotCreds
+      ]
+   in
+    TestData (Map.fromList updatedMembers) votes committeeState
```
lehins: `elem` for lists is bad, even for tests. Also conversion to a list can be avoided
```suggestion
  let
    hotCredsSet = Set.fromList hotCreds
   in
    TestData members votes $ committeeState {
        csCommitteeCreds = 
          Map.map (\mhk -> mhk >>= \hk -> hk <$ guard (hk `Set.notMember` hotCredSet)) updatedState
      }
```

eras/conway/impl/test/Test/Cardano/Ledger/Conway/CommitteeRatifySpec.hs
```diff
@@ -0,0 +1,330 @@
...
+
+totalVotes :: Votes era -> Map (Credential 'HotCommitteeRole (EraCrypto era)) Vote
+totalVotes Votes {votedYes, votedNo, votedAbstain} =
+  Map.union
+    (Map.fromList [(cred, VoteYes) | cred <- votedYes])
+    ( Map.union
+        (Map.fromList [(cred, VoteNo) | cred <- votedNo])
+        (Map.fromList [(cred, Abstain) | cred <- votedAbstain])
+    )
```
lehins: I suspect all of the tests will become slightly simpler if votes were actually sets, rather than lists
```suggestion
    Map.unions
      [ Map.fromSet (const VoteYes) votedYes
      , Map.fromSet (const VoteNo) votedNo
      , Map.fromSet (const Abstain) votedAbstain
      ]
```

eras/conway/impl/test/Test/Cardano/Ledger/Conway/CommitteeRatifySpec.hs
```diff
@@ -0,0 +1,330 @@
...
+  listOf1 (arbitrary @(Credential 'ColdCommitteeRole (EraCrypto era)))
+
+genNonResignedCommitteeState ::
+  forall era.
+  Era era =>
+  [Credential 'ColdCommitteeRole (EraCrypto era)] ->
+  Gen (CommitteeState era)
+genNonResignedCommitteeState coldCreds = do
+  hotCreds <- vectorOf (length coldCreds) (arbitrary @(Credential 'HotCommitteeRole (EraCrypto era)))
+  pure $ CommitteeState $ Map.fromList $ zip coldCreds (Just <$> hotCreds)
```
lehins: Same here. It is better to operate directly on sets and maps and avoid lists.
Also, I think it would be a good idea to occasionally test a case when a different coldCred would map to the same hotCred (not an unreasonable situation when two CC member use the same script for some reason)
```suggestion
  Set (Credential 'ColdCommitteeRole (EraCrypto era)) ->
  Gen (CommitteeState era)
genNonResignedCommitteeState coldCreds = do
  hotCredsMap <- sequence $ Map.fromSet (\_ -> Just <$> arbitrary) coldCreds
  frequency
    [ (9, pure $ CommitteeState hotCredsMap)
    , (1, CommitteeState <$> overwriteWithDuplicate hotCredsMap)
    ]
  where
    overwriteWithDuplicate m
      | Map.size < 2 = m
      | otherwise = do
        fromIx <- choose (0, Map.size m - 1)
        toIx <- choose (0, Map.size m - 1)
        pure $ Map.updateAt (\_ _ -> Just $ Map.elemAt fromIx m) toIx m
```

eras/conway/impl/test/Test/Cardano/Ledger/Conway/CommitteeRatifySpec.hs
```diff
@@ -0,0 +1,330 @@
...
+        (Map.fromList [(cred, VoteNo) | cred <- votedNo])
+        (Map.fromList [(cred, Abstain) | cred <- votedAbstain])
+    )
+
+genNonEmptyColdCreds ::
+  forall era.
+  Era era =>
+  Gen [Credential 'ColdCommitteeRole (EraCrypto era)]
+genNonEmptyColdCreds =
+  listOf1 (arbitrary @(Credential 'ColdCommitteeRole (EraCrypto era)))
```
lehins: Type signature for generator will inferred, also producing a set right away will be better, since we never care about duplicate cold keys (and if I remember correctly duplicates won't actually be generated, not with any sort of reasonable probability that is)
```suggestion
genNonEmptyColdCreds ::
  Era era => Gen (Set (Credential 'ColdCommitteeRole (EraCrypto era))
genNonEmptyColdCreds = Set.fromList <$> listOf1 arbitrary
```

eras/conway/impl/test/Test/Cardano/Ledger/Conway/CommitteeRatifySpec.hs
```diff
@@ -0,0 +1,331 @@
...
+  forall era.
+  Era era =>
+  Ratios ->
+  Gen (TestData era)
+genTestData ratios = do
+  coldCreds <- genNonEmptyColdCreds @era
+  committeeState@(CommitteeState {csCommitteeCreds}) <- genNonResignedCommitteeState @era coldCreds
+  let hotCreds = catMaybes $ Map.elems csCommitteeCreds
+      votes = distributeVotes @era ratios hotCreds
+      members = Map.fromList $ zip coldCreds (repeat (EpochNo 10))
```
lehins: It's usually better to generate values, rather than hard code specific ones in generators:
```suggestion
   members <- 
     Map.fromList . zip coldCreds <$> vectorOf (length coldCreds) (EpochNo <$> choose (10, 100))
```

eras/conway/impl/test/Test/Cardano/Ledger/Conway/CommitteeRatifySpec.hs
```diff
@@ -0,0 +1,331 @@
...
+        , notVoted = notVotedCreds
+        }
+  where
+    splitByPct :: Rational -> Rational -> Rational -> [a] -> ([a], [a], [a], [a])
+    splitByPct x y z l =
+      let
+        size = fromIntegral $ length l
+        (xs, rest) = splitAt (ceiling (x * size)) l
+        (ys, rest') = splitAt (ceiling (y * size)) rest
+        (zs, rest'') = splitAt (ceiling (z * size)) rest'
```
lehins: I believe rounding will produce a bit more fare.
```suggestion
        (xs, rest) = splitAt (round (x * size)) l
        (ys, rest') = splitAt (round (y * size)) rest
        (zs, rest'') = splitAt (round (z * size)) rest'
```

lehins (approved):
Majority of my comments are about how the test suite can be improved. 

From what I can see nothing is blocking this PR from being merged.

Great work!!!

eras/conway/impl/src/Cardano/Ledger/Conway/Governance.hs
```diff
@@ -677,10 +683,16 @@ votingCommitteeThresholdInternal committee = \case
...
+    threshold
+      -- if the committee size is smaller than the mnimimum given in pparams,
+      -- we treat it as if we had no committe
+      | committeeSize < minSize = NoVotingThreshold
+      | otherwise =
+          case committeeQuorum <$> committee of
+            SJust t -> VotingThreshold t
+            SNothing -> NoVotingThreshold
+    minSize = fromIntegral $ pp ^. ppCommitteeMinSizeL
+    committeeSize = Map.size $ foldMap' committeeMembers committee
```
lehins: Now this is wrong.

`NoVotingThreshold` means that action will be accepted.

This is a wrong place to handle this case of committee too small

lehins (changes requested):
Spoke too soon :grin: 

~Latest changes made it wrong~ After looking a bit closer, they are actually fine.

eras/conway/impl/src/Cardano/Ledger/Conway/Governance.hs
teodanciu: Yes, I wasn't sure about this at all actually

lehins: Ignore my comment, you are totally right!!!! :smile:

```suggestion
    minSize = pp ^. ppCommitteeMinSizeL
    committeeSize = fromIntegral $ Map.size $ foldMap' committeeMembers committee
```

eras/conway/impl/test/Test/Cardano/Ledger/Conway/CommitteeRatifySpec.hs
teodanciu: Love this version, thank you

eras/conway/impl/test/Test/Cardano/Ledger/Conway/CommitteeRatifySpec.hs
teodanciu: Great suggestion!  Many things are easier now, with the votes in Sets.

eras/conway/impl/test/Test/Cardano/Ledger/Conway/CommitteeRatifySpec.hs
teodanciu: Hm this is interesting... The tests are failing if the committeeState contains duplicate values, so what you said: different cold keys mapping to the same hot key. 
The reason is that the votes are indexed by the hot key, so now instead of two votes, we'll have only 1..
I'm not sure yet if it's because of how the test data is generated - I'm sure that's at least one problem, but the the votes is still a `Map (Credential 'HotCommitteeRole (EraCrypto era)) Vote` .  
Is that what we expect, just one vote , if two cold keys map to the same hot key?

teodanciu: For now I pushed a version that doesn't duplicate the hot keys, will change it back to your suggestion after figuring out the relationship of duplicated hot keys with the voting

lehins: The only two ways it can work is one vote can be counted either once or at least once, when the same hot credential maps to many cold credentials. If I understand the spec correctly, and as I would implement it too, I would count it as many times as there are active cold credentials that match that hot credential.

So, I would suggest you try and make sure it works that way and test are passing with this too, but we should certainly discuss it on Monday meeting and see what @WhatisRT thinks about it, since I am sure he thought about this scenario.

eras/conway/impl/CHANGELOG.md
```diff
@@ -1,7 +1,10 @@
 # Version history for `cardano-ledger-conway`
 
 ## 1.9.0.0
-
+* Add to `Ratify`:
+  * `committeeAccepted`
+  * `committeeAcceptedRatio`
+* Add `reCommitteeState` to `RatifyEnv``
```
```suggestion
* Add `reCommitteeState` to `RatifyEnv`
```

eras/conway/impl/src/Cardano/Ledger/Conway/Governance.hs
```diff
@@ -690,28 +691,32 @@ votingStakePoolThresholdInternal pp isElectedCommittee action =
         TreasuryWithdrawals {} -> NoVotingAllowed
         InfoAction {} -> NoVotingThreshold
 
-isCommitteeVotingAllowed :: GovAction era -> Bool
-isCommitteeVotingAllowed =
-  isVotingAllowed . votingCommitteeThresholdInternal committee
+isCommitteeVotingAllowed :: ConwayEraPParams era => PParams era -> GovAction era -> Bool
+isCommitteeVotingAllowed pparams =
+  isVotingAllowed . votingCommitteeThresholdInternal pparams committee
```
lehins: Pparams don't actually affect if the voting by CC memebrs is allowed or not, so we can pass any value for this argument (same as with stake pools and dreps)
```suggestion
isCommitteeVotingAllowed :: GovAction era -> Bool
isCommitteeVotingAllowed =
  isVotingAllowed . votingCommitteeThresholdInternal def committee
```

eras/conway/impl/src/Cardano/Ledger/Conway/Governance.hs
```diff
@@ -720,10 +725,16 @@ votingCommitteeThresholdInternal committee = \case
...
-      case committeeQuorum <$> committee of
-        SJust t -> VotingThreshold t
-        SNothing -> NoVotingThreshold
+    threshold
+      -- if the committee size is smaller than the mnimimum given in pparams,
+      -- we treat it as if we had no committe
+      | committeeSize < minSize = NoVotingThreshold
+      | otherwise =
+          case committeeQuorum <$> committee of
+            SJust t -> VotingThreshold t
```
lehins: Slight simplification that actually matches the comment a little better :wink: 
```suggestion
    threshold =
          case committeeQuorum <$> committee of
            -- if the committee size is smaller than the minimum given in pparams,
            -- we treat it as if we had no committee
            SJust t | committeeSize >= minSize -> VotingThreshold t
```

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Gov.hs
```diff
@@ -221,16 +221,17 @@ noSuchGovActions govSnapshots gaIds =
 checkVotesAreValid ::
   forall era.
   ConwayEraPParams era =>
+  PParams era ->
```
lehins: No longer needed
```suggestion
```

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Ratify.hs
```diff
@@ -255,12 +307,12 @@ ratifyTransition = do
           withdrawalCanWithdraw _ = True
           notDelayed = not rsDelayed
       if prevActionAsExpected gasAction ensPrevGovActionIds
-        && validCommitteeSize ensCommittee ensPParams
         && validCommitteeTerm ensCommittee ensPParams reCurrentEpoch
         && notDelayed
         && withdrawalCanWithdraw gasAction
         && spoAccepted st env ast
         && dRepAccepted env st gas
+        && committeeAccepted st env gas
```
lehins: I think committee will be the fastest among SPOs and DReps, so I think we should do this first, thus short circuit earlier
```suggestion
        && committeeAccepted st env gas
        && spoAccepted st env ast
        && dRepAccepted env st gas
```

lehins (approved):
Looks great. Thank you for doing this work!

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Ratify.hs
teodanciu: Good point!

eras/conway/impl/src/Cardano/Ledger/Conway/Governance.hs
teodanciu: Nice simplification

eras/conway/impl/src/Cardano/Ledger/Conway/Governance.hs
teodanciu: Love it.  I didn't know you could "or" in a case match :grimacing:  Thank you

eras/conway/impl/test/Test/Cardano/Ledger/Conway/CommitteeRatifySpec.hs
teodanciu: I adjusted the tests to support duplicate hotkeys. 
It presented some surprising challenges in terms of generating data, but it was fun to implement and I took the opportunity to also remove the duplication between generating expired and resigned members,. 
All in all I think it was a great suggestion and worth to have this case tested.  Thank you !

PR #3751, opened 2023-09-25
Add Conway Ratify Event

kderme:
# Description

I included the whole `RatifyEnv` even though I'm not sure everything in needed.
Also the `reStakeDistr` part of the environment is not used by the ledger at all. Will it be used or removed in the future?
Initially I wanted to define an `Event (ConwayRATIFY era)`, however the recursive implementation of the ratify rules doesn't make it that easy.

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Epoch.hs
+++ b/eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Epoch.hs
@@ -42,6 +42,7 @@ import Cardano.Ledger.Conway.Governance (
   ConwayEraGov,
   ConwayGovState (..),
   EnactState,
+  GovActionId,
   GovActionState (..),
   GovSnapshots (..),
   RatifyState (..),
@@ -113,6 +114,7 @@ import Control.State.Transition (
   TRC (..),
   TransitionRule,
   judgmentContext,
+  tellEvent,
   trans,
  )
 import Data.Foldable (Foldable (..))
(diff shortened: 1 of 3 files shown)


eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Epoch.hs
```diff
@@ -121,13 +123,15 @@ import Data.Maybe (fromMaybe)
...
 import qualified Data.Sequence.Strict as Seq
+import Data.Set (Set)
 import qualified Data.Set as Set
 import Data.Void (Void, absurd)
 import Lens.Micro (Lens', (%~), (&), (+~), (.~), (<>~), (^.))
 
 data ConwayEpochEvent era
   = PoolReapEvent (Event (EraRule "POOLREAP" era))
   | SnapEvent (Event (EraRule "SNAP" era))
+  | RatifyEvent (RatifyEnv era) (EnactState era) (Set (GovActionId (EraCrypto era)))
```
lehins, line 134: Let's call it `EnactEvent` instead, because actual ratification will happen at an arbitrary point within the epoch, while act of enactment will happen at known time onthe epoch boundary. Also, if you really need something specific from the `RatifyEnv` we could potentially include it in the event, but I'd suggest we do not include the whole thing in the event, since its contents are quite complicated and will soon change. 
```suggestion
  | EnactEvent (EnactState era) (Set (GovActionId (EraCrypto era)))
```

lehins (changes requested):
RATIFY rule is about to get a little bit more complicated, in a sense that it will not be called on the epoch boundary. So, I wanna be a bit careful about the events that we emit on this rule.

However, application of EnactState will always happen on the epoch boundary, so I'd rather emit Enact events, than Ratify.

Also, in general I recommend only including things into events that you need. It is less invasive to add new events, rather than remove unused data from existing events.

kderme:
> Let's call it EnactEvent instead, because actual ratification will happen at an arbitrary point within the epoch, while act of enactment will happen at known time onthe epoch boundary

Sure, I can include only enactment for now. I'd be interested to know how ratify will work eventually and at which point in the epoch it will be done.

lehins:
Closing in favor #3856 

@kderme Could you take a look at the above PR and let me know if introduced events in it have enough information for you

PR #3756, opened 2023-09-26
Cardano-Perf Regression: Future Proof

doyougnu:
This is the `Future Proof` patch for the cardano-perf investigation project.

What it does:

It's my best guess to future proof `Cardano.Ledger.Address` by reading the Core output. In general:

- If a function was exported we set it to INLINEABLE so that it can be optimized at call sites in other modules and libraries
- If a function is relatively simple then I INLINE'd it. I also INLINE'd a few that are in a hot loop in `decodeAddrStateLeftoverT`
- If a function used `MonadFail`, I SPECIALIZED it. This will result in larger Core, but 9.2 generally specializes against these MTL type classes less aggressively than 8.10. From reading Core a lot of these failed specializations were from `FailT`.

NOTE:

This is not intended to be merged. This patch is intended to be tested on either `db-analyzer` or `beacon`. cc @dnadales


lehins:
Same as #3757

PR #3757, opened 2023-09-26
Cardano-perf Regression: Status Quo

doyougnu:
This is the 2nd patch in the Cardano perf regression project. Its meant to test the hypothesis that 9.2 is under-specializing. It marks all exported functions in `Cardano.Ledger.Address` as `INLINEABLE` and uses module level flags to let GHC specialize more aggressively.

NOTE

This is not intended to be merged. This patch is intended to be tested on either db-analyzer or beacon. cc @dnadales

### DIFF
--- a/libs/cardano-ledger-core/src/Cardano/Ledger/Address.hs
+++ b/libs/cardano-ledger-core/src/Cardano/Ledger/Address.hs
@@ -12,6 +12,8 @@
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE TypeApplications #-}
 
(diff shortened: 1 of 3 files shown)


doyougnu:
I've decided to forgo the `True Baseline` patch. This patch regression the benchmarks in `cardano-ledger-core` so much that I deemed it wasn't worth to test it further.

lehins:
I'll close this PR as stale. 

FYI, I ran a simple benchmark of decoding bunch of random addresses on this approach and showed a 20% slowdown.

In any case, we might continue the investigation, but I don't want those stale unmergable PRs hanging around.

doyougnu:
Totally fine with me. I also found that this regressed with db-analyzer and beacon.

PR #3758, opened 2023-09-26
Cardano-Perf regression: UMap.size regression fix

doyougnu:
This patch is just an assemblage of performance tweaks that I have found by auditing the code base, inspecting performance and tickyticky profiles. Most of the tweaks are inlining some key functions so that GHC can further optimize them at their respective call sites with `case-of-known-constructor` which leads to better Core.

cc @dnadales


lehins (approved):

doyougnu:
I'm interested in if these changes change performance of the ledger. Is there a way we could get this data?

PR #3765, opened 2023-09-28
Improve native script handling

lehins:
# Description

* Remove redundant script hash verification. Script Hashes are not transmitted over the wire, they are computed upon deserialization. This means there is no point in recomputing and verifying those hashes as a predicate check in the rules
* Short circuit multisig and timelock verification for `RequireMOf` as soon as necessary number of scripts have been validated. Previously it was validating all scripts, regardless of how many was needed.
* Remove the whole Phased distinction, since it was not used nor is it needed for phase2 scripts. It was only necessary for native, phase1 scripts. That is why a simpler interface was added that deals with native scripts only. Addition of `NativeScript`, `getNativeScript` and `validateNativeScript`

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/eras/allegra/impl/src/Cardano/Ledger/Allegra/Scripts.hs
+++ b/eras/allegra/impl/src/Cardano/Ledger/Allegra/Scripts.hs
@@ -81,7 +81,8 @@ import Control.DeepSeq (NFData (..))
 import Data.ByteString.Lazy (fromStrict)
 import Data.ByteString.Short (fromShort)
 import Data.Sequence.Strict as Seq (StrictSeq (Empty, (:<|)), fromList)
-import Data.Set (Set, member)
+import qualified Data.Sequence.Strict as SSeq
+import qualified Data.Set as Set (Set, member)
 import GHC.Generics (Generic)
 import NoThunks.Class (NoThunks (..))
 
@@ -193,13 +194,12 @@ instance Memoized Timelock where
 
 deriving instance HashAlgorithm (HASH (EraCrypto era)) => Show (Timelock era)
 
-type instance SomeScript 'PhaseOne (AllegraEra c) = Timelock (AllegraEra c)
-
 -- | Since Timelock scripts are a strictly backwards compatible extension of
 -- MultiSig scripts, we can use the same 'scriptPrefixTag' tag here as we did
 -- for the ValidateScript instance in MultiSig
 instance Crypto c => EraScript (AllegraEra c) where
   type Script (AllegraEra c) = Timelock (AllegraEra c)
+  type NativeScript (AllegraEra c) = Timelock (AllegraEra c)
 
   upgradeScript = \case
     Shelley.RequireSignature keyHash -> RequireSignature keyHash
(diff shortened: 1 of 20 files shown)


eras/allegra/impl/src/Cardano/Ledger/Allegra/Scripts.hs
```diff
@@ -274,21 +273,24 @@ ltePosInfty (SJust i) j = i <= j
...
+    isValidMOf n SSeq.Empty = n <= 0
+    isValidMOf n (ts SSeq.:<| tss) =
+      n <= 0 || if go ts then isValidMOf (n - 1) tss else isValidMOf n tss
+    go = \case
+      RequireTimeStart lockStart -> lockStart `lteNegInfty` txStart
+      RequireTimeExpire lockExp -> txExp `ltePosInfty` lockExp
+      RequireSignature hash -> hash `Set.member` vhks
+      RequireAllOf xs -> all go xs
+      RequireAnyOf xs -> any go xs
+      RequireMOf m xs -> isValidMOf m xs
```
TimSheard, line 293: could we call this something like 'validateTimeloxk'

eras/shelley/impl/src/Cardano/Ledger/Shelley/Scripts.hs
```diff
@@ -191,3 +195,29 @@ eqMultiSigRaw t1 t2 = go (getMemoRawType t1) (getMemoRawType t2)
...
+    -- The important part of this validator is that it will stop as soon as it reaches the
+    -- required number of valid scripts
+    isValidMOf n [] = n <= 0
+    isValidMOf n (msig : msigs) =
+      n <= 0 || if go msig then isValidMOf (n - 1) msigs else isValidMOf n msigs
+    go = \case
+      RequireSignature hk -> Set.member hk vhks
+      RequireAllOf msigs -> all go msigs
+      RequireAnyOf msigs -> any go msigs
+      RequireMOf m msigs -> isValidMOf m msigs
```
TimSheard, line 217: How about 'evalMultiSig'  instead of go

TimSheard (approved):
Seems like a better approach.  using 'go' seems a bit too obtuse though.

eras/shelley/impl/src/Cardano/Ledger/Shelley/Scripts.hs
lehins: `evalMultiSig` is the name of the function. :smile: 

Are you talking about the local `isValidMOf`? This one only works for `MOf`, hence the name `isValidMOf`

eras/allegra/impl/src/Cardano/Ledger/Allegra/Scripts.hs
lehins: Same here. `evalTimelock` is the name of the function. `isValidMOf` is only applicable to the `RequireMOf`

lehins: `validateTimelock` is the top level function right here that calls `evalTimelock`:

https://github.com/input-output-hk/cardano-ledger/blob/71bf6ae648fdead3bf37d5ba69d46f9a35d29dc2/eras/allegra/impl/src/Cardano/Ledger/Allegra/Tx.hs#L72-L77

PR #3766, opened 2023-09-28
Prevent updating protocol version with `PParamUpdate`

Soupstraw:
# Description

Added type safety for the protocol version field of `ConwayPParams` so that is not possible to change that field with a `PParamsUpdate`

# Checklist

- [x] Self-reviewed the diff


libs/cardano-ledger-core/src/Cardano/Ledger/Core/Era.hs
```diff
@@ -1,7 +1,10 @@
 {-# LANGUAGE AllowAmbiguousTypes #-}
 {-# LANGUAGE ConstraintKinds #-}
 {-# LANGUAGE DataKinds #-}
+{-# LANGUAGE ExistentialQuantification #-}
 {-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE GADTs #-}
+{-# LANGUAGE RankNTypes #-}
```
lehins: Remnants of a previous attempt
```suggestion
{-# LANGUAGE FlexibleContexts #-}
```

libs/cardano-ledger-core/src/Cardano/Ledger/Core/PParams.hs
```diff
@@ -265,7 +267,7 @@ class
 
   -- | Upgrade PParams from previous era to the current one
   upgradePParamsHKD ::
-    (HKDFunctor f, EraPParams (PreviousEra era)) =>
+    (HKDFunctor f, EraPParams (PreviousEra era), Typeable f) =>
```
lehins: This was also probably from the previous attempt using typeable, right?
```suggestion
    (HKDFunctor f, EraPParams (PreviousEra era)) =>
```

libs/cardano-ledger-core/src/Cardano/Ledger/Core/PParams.hs
```diff
@@ -327,7 +329,20 @@ class
   hkdExtraEntropyL :: (HKDFunctor f, ProtVerAtMost era 6) => Lens' (PParamsHKD f era) (HKD f Nonce)
 
   -- | Protocol version
-  hkdProtocolVersionL :: HKDFunctor f => Lens' (PParamsHKD f era) (HKD f ProtVer)
+  hkdProtocolVersionL :: (HKDFunctor f, ProtVerAtMost era 8) => Lens' (PParamsHKD f era) (HKD f ProtVer)
+
+  -- | Protocol version getter
+  hkdProtocolVersionNoUpdateG :: HKDFunctor f => SimpleGetter (PParamsHKD f era) (HKDNoUpdate f ProtVer)
+  default hkdProtocolVersionNoUpdateG :: forall f. (HKDFunctor f, ProtVerAtMost era 8) => SimpleGetter (PParamsHKD f era) (HKDNoUpdate f ProtVer)
+  hkdProtocolVersionNoUpdateG = hkdProtocolVersionL @era @f . Lens.to (toNoUpdate @f @ProtVer)
```
lehins: I don't think we need this. The only use place for this is in cardano-ledger-pretty, which is not a package I particularly care about, so it should be solved there.

Let's remove this.
```suggestion
```

eras/shelley/impl/src/Cardano/Ledger/Shelley/PParams.hs
```diff
@@ -434,13 +437,14 @@ shelleyPParamsUpdatePairs pp =
...
   forall f era.
-  (HKDFunctor f, EraPParams era, ProtVerAtMost era 4, ProtVerAtMost era 6) =>
+  (HKDFunctor f, EraPParams era, ProtVerAtMost era 4, ProtVerAtMost era 6, ProtVerAtMost era 8) =>
   Proxy f ->
   PParamsHKD f era ->
   [(Key, HKD f Aeson.Value)]
 shelleyPParamsHKDPairs px pp =
   shelleyCommonPParamsHKDPairs px pp
     ++ shelleyCommonPParamsHKDPairsV6 px pp
+    ++ shelleyCommonPParamsHKDPairsV8 px pp
```
lehins, line 447: You need to add this also to Alonzo PParams here:
https://github.com/input-output-hk/cardano-ledger/blob/48d661803c40306dc8f58ce77ab077aecf873338/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/PParams.hs#L627-L630

lehins (changes requested):
This looks great!

lehins (approved):
:+1:

PR #3768, opened 2023-09-28
Fail proposal on invalid prevGovActionId

### LINKED ISSUE #3655: Validate previous governance action id upon proposal

aniketd:
# Description

Closes #3655 

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Gov.hs
+++ b/eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Gov.hs
@@ -48,7 +48,7 @@ import Cardano.Ledger.Conway.Governance (
   GovActionState (..),
   GovProcedures (..),
   GovSnapshots (..),
-  PrevGovActionId (unPrevGovActionId),
+  PrevGovActionId (PrevGovActionId),
   PrevGovActionIds (..),
   ProposalProcedure (..),
   Voter (..),
@@ -61,6 +61,7 @@ import Cardano.Ledger.Conway.Governance (
   isStakePoolVotingAllowed,
   snapshotAddVote,
   snapshotInsertGovAction,
+  snapshotLookupId,
  )
 import Cardano.Ledger.Conway.Governance.Procedures (
   GovAction (..),
@@ -265,31 +266,33 @@ checkProposalsHaveAValidPrevious ::
   GovProcedures era ->
   Test (ConwayGovPredFailure era)
 checkProposalsHaveAValidPrevious prevGovActionIds snapshots procedures =
-  let isValidMember ::
+  let isValidPrevGovActionId ::
         StrictMaybe (PrevGovActionId p (EraCrypto era)) ->
         (GovAction era -> Bool) ->
         Bool
-      isValidMember prev snapshotCond =
-        case prev of
-          SNothing -> True
-          SJust prev' -> case Map.lookup (unPrevGovActionId prev') (snapshotGovActionStates $ snapshots ^. curGovSnapshotsL) of
-            Nothing -> False
-            -- lookup has to succeed _and_ purpose of looked-up action has to match condition
-            Just found -> snapshotCond . gasAction $ found
+      isValidPrevGovActionId prevGovActionId snapshotCond =
+        case prevGovActionId of
+          -- The case of having an SNothing as valid, for the very first proposal ever, is handled in `prevActionAsExpected`
+          SNothing -> False
+          SJust (PrevGovActionId govActionId) ->
+            case snapshotLookupId govActionId $ snapshots ^. curGovSnapshotsL of
+              Nothing -> False
+              -- lookup has to succeed _and_ purpose of looked-up action has to match condition
+              Just found -> snapshotCond $ gasAction found
       isValid proposal =
         prevActionAsExpected (proposal ^. pProcGovActionL) prevGovActionIds
           || case proposal ^. pProcGovActionL of
             ParameterChange prev _ ->
-              isValidMember prev $ \case ParameterChange {} -> True; _ -> False
+              isValidPrevGovActionId prev $ \case ParameterChange {} -> True; _ -> False
             HardForkInitiation prev _ ->
-              isValidMember prev $ \case HardForkInitiation {} -> True; _ -> False
+              isValidPrevGovActionId prev $ \case HardForkInitiation {} -> True; _ -> False
             TreasuryWithdrawals _ -> True
             NoConfidence prev ->
-              isValidMember prev $ \case NoConfidence {} -> True; UpdateCommittee {} -> True; _ -> False
+              isValidPrevGovActionId prev $ \case NoConfidence {} -> True; UpdateCommittee {} -> True; _ -> False
             UpdateCommittee prev _ _ _ ->
-              isValidMember prev $ \case NoConfidence {} -> True; UpdateCommittee {} -> True; _ -> False
+              isValidPrevGovActionId prev $ \case NoConfidence {} -> True; UpdateCommittee {} -> True; _ -> False
             NewConstitution prev _ ->
-              isValidMember prev $ \case NewConstitution {} -> True; _ -> False
+              isValidPrevGovActionId prev $ \case NewConstitution {} -> True; _ -> False
             InfoAction -> True
       invalidProposals = Seq.filter (not . isValid) $ procedures ^. govProceduresProposalsL
    in failureUnless (Seq.null invalidProposals) $ InvalidPrevGovActionIdsInProposals invalidProposals
--- a/eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Gov.hs
+++ b/eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Gov.hs
@@ -24,6 +24,7 @@ import Cardano.Ledger.BaseTypes (
   EpochNo (..),
   Network,
   ShelleyBase,
+  StrictMaybe (..),
   networkId,
  )
 import Cardano.Ledger.Binary (
@@ -47,6 +48,8 @@ import Cardano.Ledger.Conway.Governance (
   GovActionState (..),
   GovProcedures (..),
   GovSnapshots (..),
+  PrevGovActionId (unPrevGovActionId),
+  PrevGovActionIds (..),
   ProposalProcedure (..),
   Voter (..),
   VotingProcedure (..),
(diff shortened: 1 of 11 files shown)


eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Gov.hs
```diff
@@ -243,12 +254,49 @@ actionWellFormed ga = failureUnless isWellFormed $ MalformedProposal ga
...
+      isValidMember prev snapshotCond prevActionAccessor =
+        all
+          (== Just True) -- lookup has to succeed _and_ purpose of lookedup action has to match condition
+          ( fmap (snapshotCond . gasAction)
+              . (flip Map.lookup . snapshotGovActionStates $ snapshots ^. curGovSnapshotsL @era)
+              . unPrevGovActionId
+              <$> prev
+          )
+          || prev == prevActionAccessor prevGovActionIds
+      isValid proposal = case proposal ^. pProcGovActionL of
```
lehins: You can reuse this function here: 
https://github.com/input-output-hk/cardano-ledger/blob/b8c6c136657dc7282dc6fc899b790c28f6393a58/eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Ratify.hs#L340-L351
```suggestion
      isValid proposal = 
        prevActionAsExpected (proposal ^. pProcGovActionL) prevGovActionIds ||
        case proposal ^. pProcGovActionL of
```

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Gov.hs
```diff
@@ -243,12 +254,49 @@ actionWellFormed ga = failureUnless isWellFormed $ MalformedProposal ga
...
+        (PrevGovActionIds era -> StrictMaybe (PrevGovActionId p (EraCrypto era))) ->
+        Bool
+      isValidMember prev snapshotCond prevActionAccessor =
+        all
+          (== Just True) -- lookup has to succeed _and_ purpose of lookedup action has to match condition
+          ( fmap (snapshotCond . gasAction)
+              . (flip Map.lookup . snapshotGovActionStates $ snapshots ^. curGovSnapshotsL @era)
+              . unPrevGovActionId
+              <$> prev
+          )
```
lehins: This looks correct, but it is written in a very obscure way. I suggest rewriting it with regular pattern matching and clarifying the logic a bit

lehins (changes requested):
Looks right, but I am not 100% sure. Besides addressing the comments could I as you to write a couple of unit test using the DSL that was implemented in #3733

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Gov.hs
```diff
@@ -246,12 +258,48 @@ actionWellFormed ga = failureUnless isWellFormed $ MalformedProposal ga
...
+        StrictMaybe (PrevGovActionId p (EraCrypto era)) ->
+        (GovAction era -> Bool) ->
+        Bool
+      isValidMember prev snapshotCond =
+        case prev of
+          SNothing -> True
+          SJust prev' -> case Map.lookup (unPrevGovActionId prev') (snapshotGovActionStates $ snapshots ^. curGovSnapshotsL) of
+            Nothing -> False
+            -- lookup has to succeed _and_ purpose of looked-up action has to match condition
+            Just found -> snapshotCond . gasAction $ found
```
lehins: This version is much clearer, however there are a few things that need to be changed:

* Whenever previous id is not provided it is invalid, unless it is the very first proposed gov action of that type. Therefore it is only valid for `SNothing` case when `prevActionAsExpected` allows it.
* `isValidMember` is a confusing name, since we have CC members and could easily be confused as related concept. Also it is not just a Map membership test, there is also `snapshotCond` application.
* `foo . bar $ baz` uses two more characters and one more function when compared to `foo $ bar baz`
* There is a specialized function for looking into snapshots: `snapshotLookupId`
```suggestion
      isValidPrevGovActionId prevGovActionId snapshotCond =
        case prevGovActionId of
          SNothing -> False
          SJust (PrevGovActionId govActionId) -> 
            case snapshotLookupId govActionId (snapshots ^. curGovSnapshotsL) of
              Nothing -> False
              -- lookup has to succeed _and_ purpose of looked-up action has to match condition
              Just found -> snapshotCond $ gasAction found
```

eras/conway/impl/test/Test/Cardano/Ledger/Conway/GovSpec.hs
```diff
@@ -0,0 +1,128 @@
...
+                    )
+                    SNothing
+                )
+        gaidConstitutionProp <- submitProposal constitutionAction
+        modifyNES $ \nes ->
+          nes
+            & nesEsL . esLStateL . lsUTxOStateL . utxosGovStateL . cgEnactStateL . ensPrevConstitutionL
+              .~ SJust (PrevGovActionId gaidConstitutionProp) -- Add the wrong one to PrevGovActionIds
+            & nesEsL . esLStateL . lsUTxOStateL . utxosGovStateL . cgGovSnapshotsL
+              .~ def -- Remove from snapshots
```
lehins: You should not need to modify the new epoch state for these tests
 ```suggestion
```

File: eras/conway/impl/test/Test/Cardano/Ledger/Conway/GovSpec.hs [hunk 5]
```diff
@@ -0,0 +1,128 @@
...
+              NewConstitution
+                (SJust $ PrevGovActionId (gaidConstitutionProp {gaidGovActionIx = GovActionIx 1})) -- make it wrong
+                ( Constitution
+                    ( Anchor
+                        (fromJust $ textToUrl "constitution.1")
+                        constitutionHash
+                    )
+                    SNothing
+                )
+        submitFailingProposal constitutionActionNext
```
lehins:
```suggestion
        let noConfidence = NoConfidence (SJust $ PrevGovActionId gaidConstitutionProp)
        submitFailingProposal noConfidence
```

eras/conway/impl/test/Test/Cardano/Ledger/Conway/GovSpec.hs
```diff
@@ -0,0 +1,128 @@
...
+      itM @Conway "Lookup snapshots" $ do
```
lehins: This test wasn't particularly useful, since it was failing for the same reason as the test above, namely missing govActionId. we could instead make it more interesting and expect it to fail for the reason of mismatched purpose

eras/conway/impl/test/Test/Cardano/Ledger/Conway/GovSpec.hs
```diff
@@ -0,0 +1,128 @@
...
+              NewConstitution
+                SNothing
+                ( Constitution
+                    ( Anchor
+                        (fromJust $ textToUrl "constitution.0")
+                        constitutionHash
+                    )
+                    SNothing
+                )
+        gaidConstitutionProp <- submitProposal constitutionAction
```
lehins, line 194: You should not have it in the proposals if you are testing that the id matches the previously enacted one. In other words in this test you have an impossible state: govActionid is both: the previously enacted one and one of the currently proposed gov actions

You should just generate `gaidConstitutionProp` randomly
```suggestion
```

lehins (changes requested):
There is one important change that is needed and a few other minor improvements

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Ratify.hs
```diff
@@ -345,17 +346,14 @@ ratifyTransition = do
...
-  prev == pgaConstitution
-prevActionAsExpected _ _ = True -- for the other actions, the previous action is not relevant
+prevActionAsExpected = \case
+  ParameterChange prev _ -> (prev ==) . pgaPParamUpdate
+  HardForkInitiation prev _ -> (prev ==) . pgaHardFork
+  TreasuryWithdrawals _ -> const True
+  NoConfidence prev -> (prev ==) . pgaCommittee
+  UpdateCommittee prev _ _ _ -> (prev ==) . pgaCommittee
+  NewConstitution prev _ -> (prev ==) . pgaConstitution
+  InfoAction -> const True
```
teodanciu, line 356: Nice refactoring :+1:

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Gov.hs
aniketd: I could not agree more. Thanks for the insightful catch on the `SNothing` part too! :+1:

eras/conway/impl/test/Test/Cardano/Ledger/Conway/GovSpec.hs
aniketd: The thought process behind this was the following:
1. In the first test, the initial proposal has been only added to the snapshots and hasn't been enacted yet. We propose an invalid proposal, and expect it to fail.
2. In the second test, the initial proposal has also been only added to the snapshots and hasn't been enacted yet. We propose a valid proposal this time, but we want to make sure that if a previous proposal does not appear in the snapshots but has been enacted before, we still want it to pass the validation check, unless I'm completely misunderstanding this.

I also think I named the tests badly in this case. I will make this more explicit if you are okay with this thought process for testing.

1. Reject when id is not present in either snapshot or enacted state
2. Reject when id is present but purpose does not match in either snapshot or enacted state
3. Accept when id and purpose match in snapshots
4. Accept when id and purpose match in enacted state.

eras/conway/impl/test/Test/Cardano/Ledger/Conway/GovSpec.hs
aniketd: I have updated the tests to be exhaustive in the cases mentioned in a previous comment. Let me know your thoughts on the update. I think the update removes this case.

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Ratify.hs
aniketd: Thank you! :raised_hands:

eras/conway/impl/test/Test/Cardano/Ledger/Conway/GovSpec.hs
aniketd: I have also now added the test for "first ever proposal" while all other proposals with same purpose have to fail.

lehins (approved):
:+1: Very nice.

PR #3769, opened 2023-09-28
Add roundtripping by validating FlatTerm

lehins:
# Description

This PR adds:

* validating produced encoding is valid by converting it to `FlatTerm` and checking `validFlatTerm`
* Round tripping through FlatTerm
* Verifying that value decoded through FlatTerm is the same when a regular decoder is used

# Checklist

- [x] Self-reviewed the diff


eras/conway/impl/src/Cardano/Ledger/Conway/Genesis.hs
```diff
@@ -45,17 +47,11 @@ instance Crypto c => NoThunks (ConwayGenesis c)
...
 instance Crypto c => DecCBOR (ConwayGenesis c) where
-  decCBOR =
-    ConwayGenesis
-      <$> decCBOR
-      <*> decCBOR
-      <*> decCBOR
+  decCBOR = decode $ RecD ConwayGenesis <! From <! From <! From
 
-instance Crypto c => EncCBOR (ConwayGenesis c) where
+instance forall c. Crypto c => EncCBOR (ConwayGenesis c) where
```
lehins: `forall` in class instance header is not needed for `ScopedTypeVariables`, all type variables are readily available when this extension is turned on. `forall` is only needed for function type signatures.
```suggestion
instance Crypto c => EncCBOR (ConwayGenesis c) where
```

eras/conway/impl/test/Test/Cardano/Ledger/Conway/BinarySpec.hs
```diff
@@ -28,4 +30,9 @@ spec = do
     roundTripCborSpec @(Voter StandardCrypto)
     roundTripConwayCommonSpec @Conway
     -- ConwayGenesis only makes sense in Conway era
-    roundTripEraSpec @Conway @(ConwayGenesis StandardCrypto)
+    context "Genesis" $ do
+      roundTripCborSpec @(UpgradeConwayPParams Identity)
+      roundTripCborSpec @(Constitution (ConwayEra StandardCrypto))
```
lehins: This type is tested in Shelley, where it is defined

eras/conway/impl/test/Test/Cardano/Ledger/Conway/BinarySpec.hs
```diff
@@ -28,4 +30,9 @@ spec = do
...
+      roundTripCborSpec @(Committee (ConwayEra StandardCrypto))
```
lehins: This type is tested by `roundTripConwayCommonSpec`

eras/conway/impl/test/Test/Cardano/Ledger/Conway/BinarySpec.hs
```diff
@@ -28,4 +30,9 @@ spec = do
...
+      roundTripCborSpec @(ConwayGenesis StandardCrypto)
```
lehins: This test is totally superfluous, since it will test roundtrip for all protocol versions, which is unnecessary. See the comment above:
> ConwayGenesis only makes sense in Conway era

Therefore we only test `ConwayGenesis` for Conway era, which is done below

eras/conway/impl/test/Test/Cardano/Ledger/Conway/BinarySpec.hs
```diff
@@ -28,4 +30,9 @@ spec = do
     roundTripCborSpec @(Voter StandardCrypto)
     roundTripConwayCommonSpec @Conway
     -- ConwayGenesis only makes sense in Conway era
-    roundTripEraSpec @Conway @(ConwayGenesis StandardCrypto)
+    context "Genesis" $ do
+      roundTripCborSpec @(UpgradeConwayPParams Identity)
```
lehins: This type is tested as part of the ConwayGenesis serialization. It is not meant to be serialized as a standalone type, and the only reason it even has the Enc/DecCBOR instances defined is because it is used in ConwayGenesis.

So, this test is also not needed. And if it was needed, it would have been redundant to use `roundTripCborSpec`, because `UpgradeConwayPParams` can only be used in Conway, but `roundTripCborSpec` would test it for all protocol versions.

lehins (commented):
@aniketd Thank you for fixing the ConwayGenesis serialization instance.

Addition of new tests was unnecessary. I've removed them, but I made comments to clarify why I did that.

eras/conway/impl/src/Cardano/Ledger/Conway/Genesis.hs
aniketd: TIL. Thanks! :heart:

eras/conway/impl/test/Test/Cardano/Ledger/Conway/BinarySpec.hs
aniketd: :memo: got it! Thanks! :+1:

eras/conway/impl/test/Test/Cardano/Ledger/Conway/BinarySpec.hs
aniketd: Got it. Yes, this one needs to be removed. I was testing the parts since the `ConwayGenesis` type was failing. :raised_hands:

eras/conway/impl/test/Test/Cardano/Ledger/Conway/BinarySpec.hs
aniketd: Noted! :slightly_smiling_face:

aniketd (approved):
LGTM :raised_hands:

PR #3770, opened 2023-09-29
cabal.project: Bump index states

erikd:
This is the first step in making this build with `ghc-9.8`. Only the `index-state` values were changed.

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -12,9 +12,9 @@ repository cardano-haskell-packages
 
 index-state:
   -- Bump this if you need newer packages from Hackage
-  , hackage.haskell.org 2023-07-17T00:00:00Z
+  , hackage.haskell.org 2023-09-29T00:20:27Z
   -- Bump this if you need newer packages from CHaP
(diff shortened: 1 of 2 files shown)


lehins:
Whenever index state is updated we also need to update nix flakes

erikd:
@lehins Updating the CHaP index-state was easy:
```
nix flake lock --update-input CHaP
```
However, I tried every combination I could think of for
```
nix flake lock --update-input hackage.nix
```
and none of them worked.

michaelpj:
@erikd it's written down: https://github.com/input-output-hk/cardano-ledger/blob/master/CONTRIBUTING.md?plain=1#L64

erikd:
> @erikd it's written down: https://github.com/input-output-hk/cardano-ledger/blob/master/CONTRIBUTING.md?plain=1#L64

So why is it different between projects? I have touched 10 different input-output-hk projects in the last week. Every project doing it in a unique snowflake way is a HUGE pain in the neck.

lehins (approved):
:+1:

PR #3775, opened 2023-10-03
Refactor and improve constraint based STS tests

### LINKED ISSUE #3785: Constraints and properties for GOVCERT rule
Constraint based testing framework is mature enough to support Conway.

We need to add properties for the GOVCERT rule

MaximilianAlgehed:
# Description

This PR refactors the STS-based tests and introduces some pretty-printing and a functioning common HOF for writing the tests.

We introduce some minor prettyprinters and make the prettyprinter for cost models slightly less verbose in the case where the costmodel is irrelevant (i.e. when it's all 0s).

We also refactor some constraints (the ones we needed for the example) into separate "generation" and "checking" constraints to aide shrinking and checking properties.

Finally, we also add checking to the `prop_DELEG` and `prop_POOL` properties to check that the new state conforms to the checking constraints.

@TimSheard the `prop_DELEG` property fails because we have a constraint `Dom rewards :=: Rng ptrs` and the `ptrs` aren't updated by the `DELEG` rule. As far as I can tell the `ptrs` don't exist in the agda spec for conway - should the constraint be a subset constraint or is this a bug in the `DELEG` rule?

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/libs/cardano-ledger-pretty/CHANGELOG.md
+++ b/libs/cardano-ledger-pretty/CHANGELOG.md
@@ -2,7 +2,12 @@
 
 ## 1.3.3.0
 
-* Add `PrettyA` instance for `NoUpdate`
+* Add `PrettyA` instances for:
+  * `NoUpdate`
+  * `ConwayGovCert`
+  * `ConwayGovCertEnv`
+  * `PoolEnv`
+  * `EnactSignal`
 
 ## 1.3.2.0
 
--- a/libs/cardano-ledger-pretty/src/Cardano/Ledger/Pretty/Conway.hs
+++ b/libs/cardano-ledger-pretty/src/Cardano/Ledger/Pretty/Conway.hs
@@ -48,9 +48,11 @@ import Cardano.Ledger.Conway.Rules (
   ConwayCertPredFailure (..),
   ConwayCertsPredFailure (..),
   ConwayDelegPredFailure (..),
+  ConwayGovCertEnv (..),
   ConwayGovCertPredFailure,
   ConwayGovPredFailure,
   ConwayLedgerPredFailure (..),
+  EnactSignal (..),
   EnactState (..),
   PredicateFailure,
   RatifyState (..),
@@ -85,6 +87,7 @@ import Cardano.Ledger.Pretty (
   ppWithdrawals,
  )
 import Cardano.Ledger.Pretty.Babbage ()
+import Cardano.Ledger.Shelley.Rules (PoolEnv (..))
 import Data.Text (Text)
 import Lens.Micro ((^.))
 import Numeric.Natural (Natural)
@@ -143,6 +146,15 @@ instance PrettyA (ConwayDelegCert c) where
       , ("Delegatee", prettyA delegatee)
       , ("Deposit", prettyA deposit)
       ]
+
+instance PrettyA (ConwayGovCert c) where
+  prettyA = ppConwayGovCert
+
+instance PrettyA (PParams c) => PrettyA (ConwayGovCertEnv c) where
+  prettyA (ConwayGovCertEnv pp ce) = ppSexp "ConwayGovCertEnv" [prettyA pp, prettyA ce]
+
+instance PrettyA (PParams c) => PrettyA (PoolEnv c) where
+  prettyA (PoolEnv sn pp) = ppSexp "PoolEnv" [prettyA sn, prettyA pp]
 
 ppConwayTxCert :: ConwayTxCert era -> PDoc
 ppConwayTxCert = \case
@@ -429,6 +441,14 @@ instance PrettyA (PParams era) => PrettyA (EnactState era) where
           , ("Withdrawals", prettyA ensWithdrawals)
           , ("PrevGovActionIds", prettyA ensPrevGovActionIds)
           ]
+
+instance PrettyA (PParamsUpdate era) => PrettyA (EnactSignal era) where
+  prettyA EnactSignal {..} =
+    ppRecord
+      "EnactSignal"
+      [ ("Gov Action Id", prettyA esGovActionId)
+      , ("Gov Action", prettyA esGovAction)
+      ]
 
 instance PrettyA (PrevGovActionIds era) where
   prettyA PrevGovActionIds {..} =
--- a/libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/Monad.hs
+++ b/libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/Monad.hs
@@ -17,6 +17,7 @@ module Test.Cardano.Ledger.Constrained.Monad (
   generateWithSeed,
 ) where
 
+import GHC.Stack
 import Test.QuickCheck.Gen
 import Test.QuickCheck.Random
 
@@ -63,13 +64,13 @@ requireAll xs answer = if null bad then answer else failT (concat msgs)
 -- runTyped :: Typed t => Either [String] x
 
 -- | Pushes the (Left msgs) into a call to 'error'
-errorTyped :: Typed t -> t
+errorTyped :: HasCallStack => Typed t -> t
 errorTyped t = case runTyped t of
   Right x -> x
   Left xs -> error (unlines ("\nSolver-time error" : xs))
 
 -- | Pushes the (Left msgs) into a call to 'error', then injects into a Monad
-monadTyped :: Monad m => Typed t -> m t
+monadTyped :: (HasCallStack, Monad m) => Typed t -> m t
 monadTyped t = pure $! errorTyped t
 
 -- ================================================================
--- a/libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/Preds/CertState.hs
+++ b/libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/Preds/CertState.hs
@@ -25,6 +25,7 @@ import Test.Cardano.Ledger.Constrained.Size (Size (..), genFromSize)
 import Test.Cardano.Ledger.Constrained.Solver
 import Test.Cardano.Ledger.Constrained.TypeRep
 import Test.Cardano.Ledger.Constrained.Vars
+import Test.Cardano.Ledger.Generic.Functions (protocolVersion)
 import Test.Cardano.Ledger.Generic.PrettyCore (
   pcCertState,
   pcDState,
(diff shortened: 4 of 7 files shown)


libs/cardano-ledger-pretty/src/Cardano/Ledger/Pretty/Conway.hs
```diff
@@ -144,6 +147,15 @@ instance PrettyA (ConwayDelegCert c) where
       , ("Deposit", prettyA deposit)
       ]
 
+instance PrettyA (ConwayGovCert c) where
```
lehins, line 150: Could you please list those new instances in the changelog for cardano-ledger-pretty package

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/Preds/CertState.hs
```diff
@@ -240,6 +245,19 @@ certStatePreds _p =
...
     gdKeyHashList = var "gdKeyHashList" (ListR GenDelegPairR)
 
+certStateCheckPreds :: Era era => Proof era -> [Pred era]
+certStateCheckPreds p =
+  [ NotMember (Lit CoinR (Coin 0)) (Rng stakeDeposits)
+  , Dom rewards :=: Dom stakeDeposits
+  , Dom delegations :⊆: Dom rewards
+  ]
+    ++
+    -- TODO: ideas welcome for making this nicer
```
lehins: As far as I  can see it, this is as nice as it gets :smile: 
```suggestion
```

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/Preds/LedgerState.hs
```diff
@@ -32,6 +32,24 @@ import Test.Tasty (TestTree, defaultMain)
 
 -- =========================================
 
+enactStateGenPreds :: Reflect era => Proof era -> [Pred era]
+enactStateGenPreds p =
+  [ Random committeeVar
+  , Random constitution
+  , prevPParams p :<-: (Constr "id" id ^$ (pparams p))
+  , currPParams p :<-: (Constr "id" id ^$ (pparams p))
```
lehins: Redundant parenthesis
```suggestion
  , prevPParams p :<-: (Constr "id" id ^$ pparams p)
  , currPParams p :<-: (Constr "id" id ^$ pparams p)
```

libs/cardano-ledger-test/src/Test/Cardano/Ledger/STS.hs
```diff
@@ -286,64 +312,111 @@ genDelegCert pp ds =
 -- Properties
 ------------------------------------------------------------------------
 
+-- Setup ------------------------------------------------------------------
+
+data PropEnv era = PropEnv
+  { net :: Network
+  , pParams :: PParams era
+  , epochNo :: EpochNo
+  , slotNo :: SlotNo
```
lehins: I recommend following the same convention for record names as we do in the rest of the codebase (Use abbreviation as prefix). Otherwise it is easy to clash names like these "slotNo" with local bindings
```suggestion
  { peNetwork :: Network
  , pePParams :: PParams era
  , peEpochNo :: EpochNo
  , peSlotNo :: SlotNo
```

libs/cardano-ledger-test/src/Test/Cardano/Ledger/STS.hs
```diff
@@ -286,64 +312,111 @@ genDelegCert pp ds =
...
+  , BaseM (EraRule r era) ~ ReaderT Globals Identity
+  , PrettyA st
+  , PrettyA sig
+  , PrettyA env
+  ) =>
+  Subst era ->
+  (PropEnv era -> env) ->
+  (PropEnv era -> GenShrink st) ->
+  (PropEnv era -> st -> GenShrink sig) ->
+  (PropEnv era -> env -> st -> sig -> st -> Property) ->
```
lehins, line 349: Would it maybe make sense to turn this into a type class?

```haskell
class STS (EraRule r era) => EraRuleProperty (r :: Symbol) era where
  ruleMkEnv :: PropEnv era -> env)
  ruleGenShrinkState :: PropEnv era -> GenShrink (State (EraRule r era))
  ruleGenShrinkSignal :: PropEnv era -> State (EraRule r era) -> GenShrink (Signal (EraRule r era))
  ruleProperty :: 
       Testable prop
    => PropEnv era
    -> env
    -> State (EraRule r era)
    -- ^ Initial state
    -> Signal (EraRule r era)
    -- ^ Signal to be used for STS rule application
    -> State (EraRule r era)
    -- ^ State after the signal was applied
    -> prop
```

libs/cardano-ledger-test/src/Test/Cardano/Ledger/STS.hs
```diff
@@ -286,64 +312,111 @@ genDelegCert pp ds =
...
+
+type GenShrink a = (Gen a, a -> [a])
+
+stsProperty ::
+  forall r era env st sig.
+  ( Environment (EraRule r era) ~ env
+  , State (EraRule r era) ~ st
+  , Signal (EraRule r era) ~ sig
+  , STS (EraRule r era)
+  , era ~ ConwayEra StandardCrypto
```
lehins: I suspect this is a temporary restriction of the era, right? I would assume function like this should work for all eras
```suggestion
```

lehins (approved):
Looks good. Few suggestions, but nothing blocking this PR from being merged. @TimSheard, could you take a look at it as well?

libs/cardano-ledger-test/src/Test/Cardano/Ledger/STS.hs
MaximilianAlgehed: I would like to avoid this as it makes it hard to put custom generators in place for specific properties. For example, when you want to do negative testing you might want to say something like "please always give me an invalid state" or "please give me environments with this specific feature" - stuff like that. The type class approach would force you to put all such variations in the same place, effectively forcing you to fuse a bunch of properties into one. This in turn would make the implementation of `ruleProperty` ugly as it would now need to inspect the values given by the generators to figure out what specific property we are testing now (i.e. what negative property we are testing). From experience I can say that this approach is a bit error prone and difficult to read.

libs/cardano-ledger-test/src/Test/Cardano/Ledger/STS.hs
MaximilianAlgehed: Yes, it's temporary but fixing it might touch a bunch of other stuff - or it's a leftover from the rewrite that never got removed. I'll take a stab at fixing it today.

MaximilianAlgehed:
This PR closes #3785

libs/cardano-ledger-pretty/cardano-ledger-pretty.cabal
```diff
@@ -1,6 +1,6 @@
 cabal-version:      3.0
 name:               cardano-ledger-pretty
-version:            1.3.3.0
+version:            1.3.4.0
```
lehins: Sorry, I should have been more clear. You didn't need to update the version, just the changelog.

Version `1.3.3.0` has not been released yet. I think it would be useful if you checked out our release process: https://github.com/input-output-hk/cardano-ledger/blob/master/RELEASING.md
It will explain why this is the case
```suggestion
version:            1.3.3.0
```

libs/cardano-ledger-pretty/CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
 # Changelog for `cardano-ledger-pretty`
 
+## 1.3.4.0
+
+* Add `PrettyA` instances for:
+  * `ConwayGovCert`
+  * `ConwayGovCertEnv`
+  * `PoolEnv`
+  * `EnactSignal`
```
lehins: These need to go under `1.3.3.0` and section `1.3.4.0` needs to be removed.

lehins (changes requested):
Sorry, but the version bump was wrong, could you fix that. Otherwise it will cause problems with our release process

MaximilianAlgehed:
@lehins fixed

MaximilianAlgehed:
Sorry, I missed this part "[CHANGELOG.md] will always have the top most entry in the CHANGELOG.md set to an unreleased version." - I'm used to it being e.g. `UNREALEASED` at the top or something like that... Anway, fixed now :)

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/Monad.hs
```diff
@@ -63,13 +64,13 @@ requireAll xs answer = if null bad then answer else failT (concat msgs)
...
-errorTyped :: Typed t -> t
+errorTyped :: HasCallStack => Typed t -> t
 errorTyped t = case runTyped t of
   Right x -> x
   Left xs -> error (unlines ("\nSolver-time error" : xs))
 
 -- | Pushes the (Left msgs) into a call to 'error', then injects into a Monad
-monadTyped :: Monad m => Typed t -> m t
+monadTyped :: (HasCallStack, Monad m) => Typed t -> m t
 monadTyped t = pure $! errorTyped t
```
TimSheard, line 74: Great idea.  Wish I thought of it.

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/Preds/CertState.hs
```diff
@@ -240,6 +245,16 @@ certStatePreds _p =
...
     gdKeyHashList = var "gdKeyHashList" (ListR GenDelegPairR)
 
+certStateCheckPreds :: Era era => Proof era -> [Pred era]
+certStateCheckPreds p =
+  [ NotMember (Lit CoinR (Coin 0)) (Rng stakeDeposits)
+  , Dom rewards :=: Dom stakeDeposits
+  , Dom delegations :⊆: Dom rewards
+  , case p of
+      Conway {} -> Sized (ExactSize 0) ptrs
+      _ -> Dom rewards :=: Rng ptrs
```
TimSheard: I think this should be some thing like this
```
if protocolVersion p >= protocolVersion (Conway Standard)
   then  Sized (ExactSize 0) ptrs
   else Dom rewards :=: Rng ptrs
```
This way it will hold in all Eras greater than Conway
By the way one can find protocolVersion   by  import  Test.Cardano.Ledger.Generic.Functions(protocolVersion )

libs/cardano-ledger-test/src/Test/Cardano/Ledger/STS.hs
```diff
@@ -157,17 +183,18 @@ conwayProof = Conway Standard
...
+pParamsPs :: Reflect era => Subst era -> [Pred era]
+pParamsPs s = applySubst s $ pParamsPreds reify
 
-pParamsGen :: Subst (ConwayEra StandardCrypto) -> Gen (PParamsF (ConwayEra StandardCrypto))
-pParamsGen s = genFromConstraints conwayProof standardOrderInfo (pParamsPs s) pParamsT
+pParamsGen :: Reflect era => Subst era -> Gen (PParamsF era)
+pParamsGen s = genFromConstraints reify standardOrderInfo (pParamsPs s) pParamsT
 
 -- TODO: this is currently broken because it fails to find `maxTxSize`. This can be fixed
 -- by building a proper target for PParams but I don't yet know how to do that.
```
TimSheard, line 200: Can't we use the 
```haskell
 Proj :: Lens' b t -> Rep era t -> Term era b -> Term era t
```
on var      
```haskell
currPParams :: Era era => Proof era -> Term era (PParamsF era)
```
and the var
```haskell
maxTxSize :: Era era => Proof era -> Term era Natural
```
to do something?

TimSheard (approved):
This looks great.  I approve

lehins (approved):
Perfect. Thank you

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/Preds/CertState.hs
MaximilianAlgehed: Thanks for the pointer!

libs/cardano-ledger-test/src/Test/Cardano/Ledger/STS.hs
MaximilianAlgehed: I think we can do something like that. I've put a note in the file because it isn't blocking us from merging now.

PR #3778, opened 2023-10-04
Bump plutus to 1.14

zliu41:
Added a helper function `deserialiseAndEvaluateScript` due to the breaking changes in https://github.com/input-output-hk/plutus/pull/5538

cc @lehins @michaelpj


zliu41:
I don't think I understand the test failure. What exactly is the error, and is it possibly unrelated to my diff?

michaelpj:
Looks like a random property failure, I restarted the jobs.

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxInfo.hs
```diff
@@ -770,18 +771,39 @@ debugPlutus version db =
...
                         (_, Right ex) -> DebugSuccess ex
                     SPlutusV3 ->
-                      case PV3.evaluateScriptRestricting pv v cm eu script d of
+                      case deserialiseAndEvaluateScript PlutusV3 pv v cm eu script d of
                         (logs, Left e) -> DebugInfo logs (PlutusErrorV3 e) pd
                         (_, Right ex) -> DebugSuccess ex
 
+-- TODO: this function is supposed to exist temporarily to cope with the breaking changes
+-- in https://github.com/input-output-hk/plutus/pull/5538, until script deserialisation
+-- is properly fixed in the ledger.
```
michaelpj, line 788: Can we be more specific? Do we have a JIRA ticket for this? I think the specific issue is "until the ledger changes to deserialize scripts as part of transaction deserialisation", right?

michaelpj (approved):

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxInfo.hs
lehins: We have a related ticket in ledger: https://github.com/input-output-hk/cardano-ledger/issues/3717

I think it is enough for the context.

> until the ledger changes to deserialize scripts as part of transaction deserialisation

Just a side comment. We'll not be able to make plutus script deserialization as part of transaction deserialization, because plutus script deserialization is allowed to change for intra-era hardforks, which is not supported by the versioned serialization interface used by ledger.

However, this doesn't mean we can't avoid redundant deserialization in the rules, so this work will provide a nice optimization regardless.

lehins (approved):
Looks great. Thank you

Issue #3540, opened 2023-07-18
Prevent duplicates in `proposalProceduresTxBodyL `


lehins:
Currently `proposalProceduresTxBodyL` is defined as a sequence, which is problematic because it allows duplicate proposals.

Duplicate proposals don't make sense in general, but in the same transaction it will cause an issue with identifying a script with a ScriptPurpose, when it is implemented in #3533

Related ticket #3541, which could share the same solution.

aniketd:
#3779

PR #3779, opened 2023-10-05
Prevent duplicate certs and proposals

### LINKED ISSUE #3541: Prevent duplicate `TxCert`s in `certsTxBodyL`
There should be no identical certificates in the same transaction. It has caused us problem in the past and we should finally fix it in Conway.

This causes a weird predicate failure whenever a script is present. So we should explicitly disable duplicates, in order to avoid confusion.

Potentially implement an unordered `StrictUSet` that is backed by a `StrictSeq` and does not require an `Ord` instance. Note that conversion from a list can still be implemented in O(n*log n) time whenever an `Ord` instance is available, which is what we mostly need. This way we would be able to enforce this invariant at the type level, rather than with a predicate check

aniketd:
# Description

Closes #3541 and #3540 

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Gov.hs
+++ b/eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Gov.hs
@@ -93,8 +93,9 @@ import Control.State.Transition.Extended (
  )
 import qualified Data.Map.Merge.Strict as Map (dropMissing, merge, zipWithMaybeMatched)
 import qualified Data.Map.Strict as Map
-import Data.Sequence (Seq (..))
+import qualified Data.OSet.Strict as OSet
 import qualified Data.Sequence as Seq
+import qualified Data.Sequence.Strict as SSeq
 import qualified Data.Set as Set
 import GHC.Generics (Generic)
 import Lens.Micro ((^.))
@@ -128,7 +129,7 @@ data ConwayGovPredFailure era
   | ExpirationEpochTooSmall
       -- | Members for which the expiration epoch has already been reached
       (Map.Map (Credential 'ColdCommitteeRole (EraCrypto era)) EpochNo)
-  | InvalidPrevGovActionIdsInProposals (Seq (ProposalProcedure era))
+  | InvalidPrevGovActionIdsInProposals (Seq.Seq (ProposalProcedure era))
   deriving (Eq, Show, Generic)
 
 instance EraPParams era => ToExpr (ConwayGovPredFailure era)
@@ -293,7 +294,7 @@ checkProposalsHaveAValidPrevious prevGovActionIds snapshots procedures =
             NewConstitution prev _ ->
               isValidPrevGovActionId prev $ \case NewConstitution {} -> True; _ -> False
             InfoAction -> True
-      invalidProposals = Seq.filter (not . isValid) $ procedures ^. govProceduresProposalsL
+      invalidProposals = Seq.filter (not . isValid) $ SSeq.fromStrict $ OSet.toStrictSeq $ procedures ^. govProceduresProposalsL
    in failureUnless (Seq.null invalidProposals) $ InvalidPrevGovActionIdsInProposals invalidProposals
 
 govTransition ::
@@ -304,8 +305,8 @@ govTransition = do
   TRC (GovEnv txid currentEpoch pp prevGovActionIds, st, gp) <- judgmentContext
   expectedNetworkId <- liftSTS $ asks networkId
 
-  let applyProps st' Empty = pure st'
-      applyProps st' ((idx, ProposalProcedure {..}) :<| ps) = do
+  let applyProps st' Seq.Empty = pure st'
+      applyProps st' ((idx, ProposalProcedure {..}) Seq.:<| ps) = do
         let expectedDeposit = pp ^. ppGovActionDepositL
          in pProcDeposit
               == expectedDeposit
@@ -348,7 +349,7 @@ govTransition = do
                 pProcGovAction
                 st'
         applyProps st'' ps
-  stProps <- applyProps st $ indexedGovProps $ gpProposalProcedures gp
+  stProps <- applyProps st $ indexedGovProps $ SSeq.fromStrict $ OSet.toStrictSeq $ gpProposalProcedures gp
 
   let VotingProcedures votingProcedures = gpVotingProcedures gp
       -- Inversion of the keys in VotingProcedures, where we can find the voter for every
--- a/eras/conway/impl/src/Cardano/Ledger/Conway/TxBody.hs
+++ b/eras/conway/impl/src/Cardano/Ledger/Conway/TxBody.hs
@@ -112,6 +112,7 @@ import Control.Arrow (left)
 import Control.DeepSeq (NFData)
 import Control.Monad (when)
 import Data.Maybe.Strict (StrictMaybe (..))
+import qualified Data.OSet.Strict as SOS
 import Data.Sequence.Strict (StrictSeq, (|>))
 import qualified Data.Sequence.Strict as StrictSeq
 import Data.Set (Set)
@@ -131,7 +132,7 @@ data ConwayTxBodyRaw era = ConwayTxBodyRaw
   , ctbrOutputs :: !(StrictSeq (Sized (TxOut era)))
   , ctbrCollateralReturn :: !(StrictMaybe (Sized (TxOut era)))
   , ctbrTotalCollateral :: !(StrictMaybe Coin)
-  , ctbrCerts :: !(StrictSeq (ConwayTxCert era))
+  , ctbrCerts :: !(SOS.OSet (ConwayTxCert era))
   , ctbrWithdrawals :: !(Withdrawals (EraCrypto era))
   , ctbrTxfee :: !Coin
   , ctbrVldt :: !ValidityInterval
@@ -141,7 +142,7 @@ data ConwayTxBodyRaw era = ConwayTxBodyRaw
   , ctbrAuxDataHash :: !(StrictMaybe (AuxiliaryDataHash (EraCrypto era)))
   , ctbrTxNetworkId :: !(StrictMaybe Network)
   , ctbrVotingProcedures :: !(VotingProcedures era)
-  , ctbrProposalProcedures :: !(StrictSeq (ProposalProcedure era))
+  , ctbrProposalProcedures :: !(SOS.OSet (ProposalProcedure era))
   , ctbrCurrentTreasuryValue :: !(StrictMaybe Coin)
   , ctbrTreasuryDonation :: !Coin
   }
@@ -189,7 +190,7 @@ instance
       bodyFields 4 =
         fieldGuarded
           (emptyFailure "Certificates" "non-empty")
-          null
+          SOS.null
           (\x tx -> tx {ctbrCerts = x})
           From
       bodyFields 5 =
@@ -240,7 +241,7 @@ instance
       bodyFields 20 =
         fieldGuarded
           (emptyFailure "ProposalProcedures" "non-empty")
-          null
+          SOS.null
           (\x tx -> tx {ctbrProposalProcedures = x})
           From
       bodyFields 21 = ofield (\x tx -> tx {ctbrCurrentTreasuryValue = x}) From
@@ -318,7 +319,7 @@ basicConwayTxBodyRaw =
     StrictSeq.empty
     SNothing
     SNothing
-    StrictSeq.empty
+    SOS.empty
     (Withdrawals mempty)
     mempty
     (ValidityInterval SNothing SNothing)
@@ -328,7 +329,7 @@ basicConwayTxBodyRaw =
     SNothing
     SNothing
     (VotingProcedures mempty)
-    mempty
+    SOS.empty
     SNothing
     mempty
 
@@ -371,7 +372,7 @@ instance Crypto c => EraTxBody (ConwayEra c) where
   withdrawalsTxBodyL = lensMemoRawType ctbrWithdrawals (\txb x -> txb {ctbrWithdrawals = x})
   {-# INLINE withdrawalsTxBodyL #-}
 
-  certsTxBodyL = lensMemoRawType ctbrCerts (\txb x -> txb {ctbrCerts = x})
+  certsTxBodyL = lensMemoRawType (SOS.toStrictSeq . ctbrCerts) (\txb x -> txb {ctbrCerts = SOS.fromStrictSeq x})
   {-# INLINE certsTxBodyL #-}
 
   upgradeTxBody btb = do
@@ -404,7 +405,7 @@ instance Crypto c => EraTxBody (ConwayEra c) where
               <$> btbCollateralReturn btb
         , ctbTotalCollateral = btbTotalCollateral btb
         , ctbCurrentTreasuryValue = SNothing
-        , ctbProposalProcedures = mempty
+        , ctbProposalProcedures = SOS.empty
         , ctbVotingProcedures = VotingProcedures mempty
         , ctbTreasuryDonation = Coin 0
         }
@@ -543,7 +544,7 @@ pattern ConwayTxBody ::
   StrictMaybe (AuxiliaryDataHash (EraCrypto era)) ->
   StrictMaybe Network ->
   VotingProcedures era ->
-  StrictSeq (ProposalProcedure era) ->
+  SOS.OSet (ProposalProcedure era) ->
   StrictMaybe Coin ->
   Coin ->
   ConwayTxBody era
@@ -576,7 +577,7 @@ pattern ConwayTxBody
         , ctbrOutputs = ctbOutputs
         , ctbrCollateralReturn = ctbCollateralReturn
         , ctbrTotalCollateral = ctbTotalCollateral
-        , ctbrCerts = ctbCerts
+        , ctbrCerts = (SOS.toStrictSeq -> ctbCerts)
         , ctbrWithdrawals = ctbWithdrawals
         , ctbrTxfee = ctbTxfee
         , ctbrVldt = ctbVldt
@@ -620,7 +621,7 @@ pattern ConwayTxBody
             outputsX
             collateralReturnX
             totalCollateralX
-            certsX
+            (SOS.fromStrictSeq certsX)
             withdrawalsX
             txfeeX
             vldtX
@@ -658,7 +659,7 @@ encodeTxBodyRaw ConwayTxBodyRaw {..} =
         !> encodeKeyedStrictMaybe 17 ctbrTotalCollateral
         !> Key 2 (To ctbrTxfee)
         !> encodeKeyedStrictMaybe 3 top
-        !> Omit null (Key 4 (To ctbrCerts))
+        !> Omit SOS.null (Key 4 (To ctbrCerts))
         !> Omit (null . unWithdrawals) (Key 5 (To ctbrWithdrawals))
         !> encodeKeyedStrictMaybe 8 bot
         !> Omit null (Key 14 (To ctbrReqSignerHashes))
@@ -667,7 +668,7 @@ encodeTxBodyRaw ConwayTxBodyRaw {..} =
         !> encodeKeyedStrictMaybe 7 ctbrAuxDataHash
         !> encodeKeyedStrictMaybe 15 ctbrTxNetworkId
         !> Omit (null . unVotingProcedures) (Key 19 (To ctbrVotingProcedures))
-        !> Omit null (Key 20 (To ctbrProposalProcedures))
+        !> Omit SOS.null (Key 20 (To ctbrProposalProcedures))
         !> encodeKeyedStrictMaybe 21 ctbrCurrentTreasuryValue
         !> Omit (== mempty) (Key 22 $ To ctbrTreasuryDonation)
 
@@ -686,6 +687,6 @@ class BabbageEraTxBody era => ConwayEraTxBody era where
   votingProceduresTxBodyL :: Lens' (TxBody era) (VotingProcedures era)
 
   -- | Lens for getting and setting `ProposalProcedures`.
-  proposalProceduresTxBodyL :: Lens' (TxBody era) (StrictSeq (ProposalProcedure era))
+  proposalProceduresTxBodyL :: Lens' (TxBody era) (SOS.OSet (ProposalProcedure era))
 
   treasuryDonationTxBodyL :: Lens' (TxBody era) Coin
--- a/eras/conway/test-suite/cddl-files/conway.cddl
+++ b/eras/conway/test-suite/cddl-files/conway.cddl
@@ -55,7 +55,7 @@ transaction_body =
   , 1 : [* transaction_output]
   , 2 : coin                               ; fee
   , ? 3 : uint                             ; time to live
-  , ? 4 : [+ certificate]
+  , ? 4 : certificates
   , ? 5 : withdrawals
   , ? 7 : auxiliary_data_hash
   , ? 8 : uint                             ; validity interval start
@@ -68,7 +68,7 @@ transaction_body =
   , ? 17 : coin                            ; total collateral
   , ? 18 : nonempty_set<transaction_input> ; reference inputs
   , ? 19 : voting_procedures               ; New; Voting procedures
-  , ? 20 : [+ proposal_procedure]          ; New; Proposal procedures
+  , ? 20 : proposal_procedures             ; New; Proposal procedures
   , ? 21 : coin                            ; New; current treasury value
   , ? 22 : positive_coin                   ; New; donation
   }
@@ -86,6 +86,10 @@ proposal_procedure =
   , gov_action
   , anchor
   ]
+
+proposal_procedures = nonempty_oset<proposal_procedure>
+
+certificates = nonempty_oset<certificate>
 
 gov_action =
   [ parameter_change_action
--- a/eras/conway/test-suite/cddl-files/mock/extras.cddl
+++ b/eras/conway/test-suite/cddl-files/mock/extras.cddl
@@ -3,6 +3,8 @@ set<a> = [a]
 
 nonempty_set<a> = [+ a]
 
+nonempty_oset<a> = #6.258([+ a])
+
 unit_interval = #6.30([1, 2])
   ; real unit_interval is: #6.30([uint, uint])
   ; but this produces numbers outside the unit interval
--- a/libs/cardano-data/CHANGELOG.md
+++ b/libs/cardano-data/CHANGELOG.md
@@ -1,8 +1,8 @@
 # Version history for `cardano-data`
 
-## 1.1.0.1
+## 1.1.1.0
 
-*
+- Add Data.OSet.Strict #3779
 
 ## 1.1.0.0
 
--- a/libs/cardano-data/cardano-data.cabal
+++ b/libs/cardano-data/cardano-data.cabal
@@ -1,6 +1,6 @@
 cabal-version:      3.0
 name:               cardano-data
-version:            1.1.0.0
+version:            1.1.1.0
 license:            Apache-2.0
 maintainer:         operations@iohk.io
 author:             IOHK
@@ -22,6 +22,7 @@ library
         Data.MapExtras
         Data.ListMap
         Data.Universe
+        Data.OSet.Strict
 
     hs-source-dirs:   src
     default-language: Haskell2010
@@ -32,7 +33,8 @@ library
     build-depends:
         base >=4.14 && <4.19,
         aeson,
-        cardano-ledger-binary >=1.0,
+        cardano-ledger-binary >=1.2,
+        cardano-strict-containers >=0.1.2.1,
         containers,
         deepseq,
         mtl,
@@ -40,7 +42,10 @@ library
         vector
 
 library testlib
-    exposed-modules:  Test.Cardano.Data
+    exposed-modules:
+        Test.Cardano.Data
+        Test.Cardano.Data.Arbitrary
+
     visibility:       public
     hs-source-dirs:   testlib
     default-language: Haskell2010
@@ -52,13 +57,18 @@ library testlib
         base,
         containers,
         hspec,
-        QuickCheck
+        QuickCheck,
+        cardano-ledger-binary:testlib,
+        cardano-data
 
 test-suite cardano-data-tests
     type:             exitcode-stdio-1.0
     main-is:          Main.hs
     hs-source-dirs:   test
-    other-modules:    Test.Cardano.Data.MapExtrasSpec
+    other-modules:
+        Test.Cardano.Data.MapExtrasSpec
+        Test.Cardano.Data.OSet.StrictSpec
+
     default-language: Haskell2010
     ghc-options:
         -Wall -Wcompat -Wincomplete-record-updates
@@ -70,5 +80,8 @@ test-suite cardano-data-tests
         containers,
         hspec,
         cardano-data,
+        cardano-ledger-binary:testlib,
+        cardano-strict-containers,
         testlib,
-        QuickCheck
+        QuickCheck,
+        quickcheck-classes-base
--- a/libs/cardano-data/src/Data/OSet/Strict.hs
+++ b/libs/cardano-data/src/Data/OSet/Strict.hs
@@ -1,0 +1,256 @@
+{-# LANGUAGE BangPatterns #-}
+{-# LANGUAGE DeriveAnyClass #-}
+{-# LANGUAGE DeriveGeneric #-}
+{-# LANGUAGE DerivingStrategies #-}
+{-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE ImportQualifiedPost #-}
+{-# LANGUAGE LambdaCase #-}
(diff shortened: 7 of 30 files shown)


libs/cardano-data/test/Test/Cardano/Data/OSet/Strict/Spec.hs
```diff
@@ -0,0 +1,55 @@
...
+import Test.Cardano.Data.Arbitrary ()
+import Test.Hspec
+import Test.Hspec.QuickCheck
+
+spec :: Spec
+spec =
+  describe "OSet.Strict" $ do
+    prop "membership checks work" $ \(s :: OSet Int) -> do
+      case uncons s of
+        Nothing -> True `shouldBe` True
```
lehins: There is no need to use expectations that are trivially true. Same goes for all other places in this module that use this
```suggestion
        Nothing -> pure ()
```

libs/cardano-data/src/Data/OSet/Strict.hs
```diff
@@ -0,0 +1,152 @@
...
+import Data.Sequence.Strict qualified as SSeq
+import NoThunks.Class (NoThunks (..))
+import Prelude hiding (lookup, null)
+
+-- | A general-purpose finite, ordered, set that is strict in its values.
+--
+-- The strictness is enforced by the underlying `StrictSeq` from base, and the
+-- uniqueness of the values is enforced in this module, by the constructing
+-- functions rather than at the type-level.
+newtype OSet a = OSet {unOSet :: SSeq.StrictSeq a}
```
lehins:
```suggestion
data OSet a = OSet { oSeq :: !(SSeq.StrictSeq a)
                   , oSet :: Set a
                   }
```

libs/cardano-data/src/Data/OSet/Strict.hs
```diff
@@ -0,0 +1,152 @@
...
+singleton :: a -> OSet a
+singleton !x = OSet $ SSeq.singleton x
+
+lookup :: Int -> OSet a -> Maybe a
+lookup i (OSet xs) = SSeq.lookup i xs
+
+(!?) :: OSet a -> Int -> Maybe a
+(!?) = flip lookup
+
+-- | \(O(n)\). Checks membership before cons'ing with the new OSet.
```
lehins: This is BAD!
```suggestion
-- | \(O(n^2)\). Checks membership before cons'ing with the new OSet.
```

lehins (commented):
I haven't looked into the PR too deeply, but overall looks like a  nice initial implementation. 

Let's now make it more efficient by adding a `Set` to the `OSet`

eras/conway/impl/src/Cardano/Ledger/Conway/TxBody.hs
```diff
@@ -371,7 +372,7 @@ instance Crypto c => EraTxBody (ConwayEra c) where
   withdrawalsTxBodyL = lensMemoRawType ctbrWithdrawals (\txb x -> txb {ctbrWithdrawals = x})
   {-# INLINE withdrawalsTxBodyL #-}
 
-  certsTxBodyL = lensMemoRawType ctbrCerts (\txb x -> txb {ctbrCerts = x})
+  certsTxBodyL = lensMemoRawType ctbrCerts (\txb x -> txb {ctbrCerts = SOS.toStrictSeq $ SOS.fromStrictSeq x})
```
lehins: This is the desired semantics.
We want to use the OSet in the `ConwayTxBody` type
```suggestion
  certsTxBodyL = lensMemoRawType (SOS.toStrictSeq . ctbrCerts) (\txb x -> txb {ctbrCerts = SOS.fromStrictSeq x})
```

libs/cardano-data/src/Data/OSet/Strict.hs
```diff
@@ -0,0 +1,152 @@
...
+-- The strictness is enforced by the underlying `StrictSeq` from base, and the
+-- uniqueness of the values is enforced in this module, by the constructing
+-- functions rather than at the type-level.
+newtype OSet a = OSet {unOSet :: SSeq.StrictSeq a}
+  deriving stock (Eq, Ord, Show)
+  deriving newtype
+    ( Serialise
+    , NFData
+    , ToCBOR
+    , FromCBOR
```
lehins: These should not be derived. We need decoders that fail when duplicates are encoded
```suggestion
```

aniketd: Noted! :raised_hands: Thank you! :bow:

libs/cardano-data/src/Data/OSet/Strict.hs
```diff
@@ -0,0 +1,203 @@
...
+-- De-duplicates the StrictSeq without overwriting.
+-- Starts from the left or head, using `foldl'`
+fromStrictSeq :: Ord a => SSeq.StrictSeq a -> OSet a
+fromStrictSeq = foldl' snoc empty
+
+-- | \(O(n \log n)\). Checks membership before cons'ing.
+-- Calls `error` when duplicates are present.
+-- Starts from the left or head, using `foldl'`
+fromStrictSeqWithError :: (Show a, Ord a) => SSeq.StrictSeq a -> OSet a
+fromStrictSeqWithError = foldl' (flip cons') empty
```
aniketd:
```suggestion
fromStrictSeqWithError = foldl' snoc' empty
```

libs/cardano-data/src/Data/OSet/Strict.hs
```diff
@@ -0,0 +1,203 @@
...
+{-# COMPLETE Empty, (:|>:) #-}
+{-# COMPLETE Empty, (:<|:) #-}
+
+-- | \( O(n\log(m*n)) \). For every uncons-ed element from the sequence on the right,
+-- checks its membership in the sequence on the left, before snoc'ing it.
+-- Preserves order. Remove duplicates from sequence on the right.
+(|><) :: Ord a => OSet a -> OSet a -> OSet a
+osetl |>< (OSet seqr setr) = case seqr of
+  SSeq.Empty -> osetl
+  r' SSeq.:<| rs -> (osetl |> r') |>< OSet rs (r' `Set.delete` setr)
```
aniketd: Why not
```suggestion
osetl |>< osetr = case osetr of
  Empty -> osetl
  r :<|: rs -> (osetl |> r) |>< rs
```
and similarly for `><|` too?

libs/cardano-data/src/Data/OSet/Strict.hs
```diff
@@ -0,0 +1,203 @@
...
+uncons :: Ord a => OSet a -> Maybe (a, OSet a)
+uncons (OSet seq set) = case Seq.viewl $ SSeq.fromStrict seq of
+  Seq.EmptyL -> Nothing
+  x Seq.:< xs' -> Just (x, OSet (SSeq.forceToStrict xs') (x `Set.delete` set))
+
+-- | \( O(\log n) \).
+unsnoc :: Ord a => OSet a -> Maybe (OSet a, a)
+unsnoc (OSet seq set) = case Seq.viewr $ SSeq.fromStrict seq of
+  Seq.EmptyR -> Nothing
+  xs' Seq.:> x -> Just (OSet (SSeq.forceToStrict xs') (x `Set.delete` set), x)
```
lehins: We need to avoid a call to `SSeq.forceToStrict` because it will iterate the whole structure, but all elements are already in WHNF, so this is redundant.
```suggestion
  xs' Seq.:> x -> Just (OSet xs' (x `Set.delete` set), x)
```

These are the functions that you probably missed:
https://github.com/input-output-hk/cardano-base/blob/420858fe28d977b44b649e94b3968ae898b86d4b/cardano-strict-containers/src/Data/Sequence/Strict.hs#L141-L150

libs/cardano-data/src/Data/OSet/Strict.hs
```diff
@@ -0,0 +1,203 @@
...
+(|>) :: Ord a => OSet a -> a -> OSet a
+(|>) = snoc
+
+infixl 5 |>
+
+-- | \( O(\log n) \).
+uncons :: Ord a => OSet a -> Maybe (a, OSet a)
+uncons (OSet seq set) = case Seq.viewl $ SSeq.fromStrict seq of
+  Seq.EmptyL -> Nothing
+  x Seq.:< xs' -> Just (x, OSet (SSeq.forceToStrict xs') (x `Set.delete` set))
```
lehins: Same here
```suggestion
  x Seq.:< xs' -> Just (x, OSet xs' (x `Set.delete` set))
```

libs/cardano-data/src/Data/OSet/Strict.hs
```diff
@@ -0,0 +1,203 @@
...
+-- | \(O(n \log n)\). Checks membership before cons'ing.
+-- De-duplicates the StrictSeq without overwriting.
+-- Starts from the left or head, using `foldl'`
+fromStrictSeq :: Ord a => SSeq.StrictSeq a -> OSet a
+fromStrictSeq = foldl' snoc empty
+
+-- | \(O(n \log n)\). Checks membership before cons'ing.
+-- Calls `error` when duplicates are present.
+-- Starts from the left or head, using `foldl'`
+fromStrictSeqWithError :: (Show a, Ord a) => SSeq.StrictSeq a -> OSet a
```
lehins: Producing error is not a very useful. `Either` is better, it is more powerful, since we can always convert it to bottom, but `error` to `Either` is not possible outside IO. So `Left a` would indicate the very first duplicate that was encountered
```suggestion
fromStrictEither :: (Show a, Ord a) => SSeq.StrictSeq a -> Either a (OSet a)
```
However, we can do even better. Let's also add or replace this one with a function that extracts duplicates and returns those duplicates as a Set along side the OSet without duplicates. We can always turn it into an `error` whenever the Set is non-empty
```haskell
fromStrictDuplicates :: (Show a, Ord a) => SSeq.StrictSeq a -> (Set a, OSet a)
```

Make sure to add some properties for these function since current implementation is not correct, but there are no tests for it

libs/cardano-data/src/Data/OSet/Strict.hs
lehins: Yes `flip cons'` will reverse the sequence

libs/cardano-data/src/Data/OSet/Strict.hs
```diff
@@ -0,0 +1,203 @@
...
+-- constructing functions by leveraging an accompanying `Set`.
+data OSet a = OSet
+  { osSSeq :: !(SSeq.StrictSeq a)
+  , osSet :: Set.Set a
+  }
+  deriving stock (Eq, Ord, Show, Generic)
+  deriving (ToExpr, NoThunks, NFData)
+
+instance EncCBOR a => EncCBOR (OSet a) where
+  encCBOR = encCBOR . toStrictSeq
```
lehins: Where is the DecCBOR?

libs/cardano-data/src/Data/OSet/Strict.hs
```diff
@@ -0,0 +1,203 @@
...
+
+-- | A general-purpose finite, ordered, set that is strict in its
+-- values.
+--
+-- The strictness is enforced by the underlying `StrictSeq` from base,
+-- and the uniqueness of the values is enforced in this module, by the
+-- constructing functions by leveraging an accompanying `Set`.
+data OSet a = OSet
+  { osSSeq :: !(SSeq.StrictSeq a)
+  , osSet :: Set.Set a
```
lehins: On a second thought, let's make it strict. There is not a single function in this module, which modifies the OSet without forcing this set. Which means there is no point in keeping it lazy.
```suggestion
  , osSet :: !(Set.Set a)
```

eras/conway/impl/src/Cardano/Ledger/Conway/TxBody.hs
```diff
@@ -188,9 +189,9 @@ instance
           From
       bodyFields 4 =
         fieldGuarded
-          (emptyFailure "Certificates" "non-empty")
-          null
-          (\x tx -> tx {ctbrCerts = x})
+          (emptyFailure "Certificates" "non-empty and without duplicates")
+          (\x -> null x || length x /= SOS.size (SOS.fromStrictSeq x))
```
lehins: This will result in a data structure to be iterated twice: once when we construct StrictSeq and the other time when we convert it to OSet

The proper way to do this is to have a decoder that fails on the very first duplicate when the OSet is being constructed.

libs/cardano-data/test/Test/Cardano/Data/OSet/Strict/Spec.hs
```diff
@@ -0,0 +1,67 @@
...
+import Test.Cardano.Data.Arbitrary ()
+import Test.Hspec
+import Test.Hspec.QuickCheck
+
+spec :: Spec
+spec =
+  describe "OSet.Strict" $ do
+    prop "membership checks work" $ \(s :: OSet Int) -> do
+      case uncons s of
+        Nothing -> pure ()
```
aniketd: Change to use the patterns here, so we don't have to export `uncons` and `unsnoc`.

libs/cardano-data/CHANGELOG.md
```diff
@@ -1,8 +1,8 @@
 # Version history for `cardano-data`
 
-## 1.1.0.1
+## 1.2.0.0
```
lehins: Addition of a module is not a breaking change:
```suggestion
## 1.1.2.0
```

libs/cardano-data/cardano-data.cabal
```diff
@@ -1,6 +1,6 @@
 cabal-version:      3.0
 name:               cardano-data
-version:            1.1.0.0
+version:            1.2.0.0
```
lehins: Ditto
```suggestion
version:            1.1.2.0
```

lehins (changes requested):
This looks great, some changes are still needed though

eras/conway/impl/src/Cardano/Ledger/Conway/TxBody.hs
lehins: In other words, it should be using the `DecCBOR` instance for `OSet`, which in turn should do the checks for duplicates

libs/cardano-data/src/Data/OSet/Strict.hs
aniketd: :100: :bow: Thank you!

libs/cardano-data/src/Data/OSet/Strict.hs
aniketd: Now we have it, based on `fail`, for when there are duplicates present :+1:

libs/cardano-data/CHANGELOG.md
aniketd: Oh, sorry! Noted! :memo:

aniketd: @lehins do you mean
```suggestion
## 1.1.1.0
```
instead of the double bump from `0` to `2` directly in the `minor` number?

lehins: Yes, good catch, I skipped a version somehow in my suggestion :facepalm:

aniketd:
This is still missing some tests, like CBOR round trip and for `fromStrictSeqDuplicates`

> Make sure to add some properties for these function since current implementation is not correct, but there are no tests for it

eras/conway/impl/src/Cardano/Ledger/Conway/TxBody.hs
```diff
@@ -576,7 +577,7 @@ pattern ConwayTxBody
         , ctbrOutputs = ctbOutputs
         , ctbrCollateralReturn = ctbCollateralReturn
         , ctbrTotalCollateral = ctbTotalCollateral
-        , ctbrCerts = ctbCerts
+        , ctbrCerts = (SOS.toStrictSeq -> ctbCerts)
```
Soupstraw, line 580: wouldn't a simple function application work here?

Soupstraw: Ah, ok so this is a pattern match, not a constructor, got it

lehins: No, it wouldn't. Notice that this is actually a pattern matching here on `ConwayTxBodyRaw`, so the only other way to handle this is to bind it to something else  like `stbCertsOSet` and then add another binding in the `where` clause:
```haskell
  where ctbCerts = SOS.toStrictSeq stbCertsOSet
```
However, since we are already using `ViewPatterns`, I agree with Aniket, that this is a much cleaner approach

libs/cardano-data/src/Data/OSet/Strict.hs
```diff
@@ -0,0 +1,209 @@
...
+null :: OSet a -> Bool
+null (OSet _ set) = Set.null set
+
+-- | \( O(1) \).
+size :: OSet a -> Int
+size (OSet _ set) = Set.size set
+
+-- | \( O(1) \).
+empty :: OSet a
+empty = OSet SSeq.Empty Set.empty
```
Soupstraw, line 127: maybe it's better to make `OSet` a `Monoid` instance instead

libs/cardano-data/src/Data/OSet/Strict.hs
```diff
@@ -0,0 +1,209 @@
...
+(<|) = cons
+
+infixr 5 <|
+
+-- | \( O(\log n) \). Checks membership before snoc'ing.
+snoc :: Ord a => OSet a -> a -> OSet a
+snoc oset@(OSet seq set) x =
+  if x `Set.member` set
+    then oset
+    else OSet (seq SSeq.|> x) (x `Set.insert` set)
```
Soupstraw, line 160: you can keep it like it is, but I'd do it this way
```suggestion
snoc oset@(OSet seq set) x 
  | x `Set.member` set = oset
  | otherwise =  OSet (seq SSeq.|> x) (x `Set.insert` set)
```

libs/cardano-data/src/Data/OSet/Strict.hs
aniketd: There are trivially two ways to `mappend`, in the case of `OSet`, since we can have a `left-preserving` and a `right-preserving` one. And so, I don't think it is a `Monoid`, since it is not a `Semigroup`. But please correct me if I'm wrong.

libs/cardano-data/src/Data/OSet/Strict.hs
```diff
@@ -0,0 +1,209 @@
...
+instance EncCBOR a => EncCBOR (OSet a) where
+  encCBOR = encCBOR . toStrictSeq
+
+instance (Show a, Ord a, DecCBOR a) => DecCBOR (OSet a) where
+  decCBOR =
+    let failOnDuplicates (duplicates, oset) =
+          if Set.null duplicates
+            then pure oset
+            else fail $ "Duplicates found while decoding OSet: " <> show duplicates
+     in failOnDuplicates . fromStrictSeqDuplicates =<< decCBOR
```
lehins: This still has the same problem as before, it will iterate all of the elements twice.

I suggest adjusting this function to make it work for Set and OSet:
https://github.com/input-output-hk/cardano-ledger/blob/d7a12f52b33700015d033b4c56ab68aadca03754/libs/cardano-ledger-binary/src/Cardano/Ledger/Binary/Decoding/Decoder.hs#L829-L851 

All you need to do is make it accept an `insert` and `member` functions as arguments

libs/cardano-data/src/Data/OSet/Strict.hs
aniketd: I like your suggestion! Thanks! Will change it. :+1:

libs/cardano-data/src/Data/OSet/Strict.hs
```diff
@@ -0,0 +1,209 @@
...
+
+-- | \(O(n \log n)\). Checks membership before snoc-ing.
+-- Returns a 2-tuple, with `fst` as a `Set` of duplicates found
+-- and the `snd` as the de-duplicated `OSet` without overwriting.
+-- Starts from the left or head, using `foldl'`
+fromStrictSeqDuplicates :: Ord a => SSeq.StrictSeq a -> (Set.Set a, OSet a)
+fromStrictSeqDuplicates = foldl' snoc' (Set.empty, empty)
+  where
+    snoc' :: Ord a => (Set.Set a, OSet a) -> a -> (Set.Set a, OSet a)
+    snoc' (duplicates, oset@(OSet seq set)) x =
```
lehins: Pattern matches are strict to WHNF only, so it is always best to force the elements as well. Here the space leak would not be too terrible, since `member` function would force the `oset`, but still it is a good habit to get into to ensure accumulator is as strict as possible for foldls
```suggestion
    snoc' (!duplicates, !oset@(OSet seq set)) x =
```

libs/cardano-data/src/Data/OSet/Strict.hs
aniketd: In particular for `(x <> y) <> z = x <> (y <> z)`, if I understand correctly, there are two possible `Semigroup` implementations of an `OSet`?

eras/conway/impl/src/Cardano/Ledger/Conway/TxBody.hs
```diff
@@ -512,7 +513,7 @@ instance Crypto c => ConwayEraTxBody (ConwayEra c) where
     lensMemoRawType ctbrVotingProcedures (\txb x -> txb {ctbrVotingProcedures = x})
   {-# INLINE votingProceduresTxBodyL #-}
   proposalProceduresTxBodyL =
-    lensMemoRawType ctbrProposalProcedures (\txb x -> txb {ctbrProposalProcedures = x})
+    lensMemoRawType (SOS.toStrictSeq . ctbrProposalProcedures) (\txb x -> txb {ctbrProposalProcedures = SOS.fromStrictSeq x})
```
lehins: Let's change proposals to `OSet` in a public interface, since this is a new field not present in previous eras, we are not introducing any breakage.
```suggestion
    lensMemoRawType (ctbrProposalProcedures) (\txb x -> txb {ctbrProposalProcedures = x})
```

libs/cardano-data/src/Data/OSet/Strict.hs
aniketd: What do you think @Soupstraw . Also, please correct me if I'm making any wrong assumptions, or any of my assumptions "smell" wrong to you :smiley: . Today is a fine day for learning, like every other day :raised_hands:

libs/cardano-data/src/Data/OSet/Strict.hs
```diff
@@ -0,0 +1,209 @@
...
+-- The strictness is enforced by the underlying `StrictSeq` from base,
+-- and the uniqueness of the values is enforced in this module, by the
+-- constructing functions by leveraging an accompanying `Set`.
+data OSet a = OSet
+  { osSSeq :: !(SSeq.StrictSeq a)
+  , osSet :: !(Set.Set a)
+  }
+  deriving stock (Eq, Ord, Show, Generic)
+  deriving (ToExpr, NoThunks, NFData)
+
```
lehins, line 69: Let's also add a Foldble instance. It should be very simple to do. All that is necessary for all functions in Foldable to be used with `osSSeq`, eg.
```haskell
  foldl f acc = foldl f acc . osSSeq

**Review by lehins — CHANGES_REQUESTED** [review]
This is looking much better :slightly_smiling_face:

File: libs/cardano-data/src/Data/OSet/Strict.hs (hunk 18 above)
Comment (aniketd) [reply libs/cardano-data/src/Data/OSet/Strict.hs:127]: Alexey suggested making a `Foldable` instance too, which will require a `Monoid` instance. Do you think it is okay to pick either of the possible `mappends` as _the_ `mappend` for `Semigroup` and make `Monoid` possible?

File: libs/cardano-data/src/Data/OSet/Strict.hs (hunk 20 above)
Comment (aniketd) [reply libs/cardano-data/src/Data/OSet/Strict.hs]: I'm sorry I forgot that you had mentioned for me to take a look at this function during our first discussion about this implementation. :+1:

File: libs/cardano-data/src/Data/OSet/Strict.hs (hunk 18 above)
Comment (aniketd) [reply libs/cardano-data/src/Data/OSet/Strict.hs:127]: TODO:
1. Use `left-preserving` to match semantics of `Set` w.r.t. `Semigroup` instance. 
2. `Foldable` requires a `Monoid` instance on the type-parameter, not on the class itself.
3. Change CDDL to reflect this usage, but aliasing `OSet` to a `Set` and use the `{Enc,Dec}CBOR for `Set`, if they are order preserving.

Comment (Soupstraw) [reply libs/cardano-data/src/Data/OSet/Strict.hs:127]: Yeah, both left-biased and right-biased union are valid semigroups, since they are associative. I think using the left-biased union is more consistent w.r.t. other biased unions that we have in our code.

Comment (lehins) [reply libs/cardano-data/src/Data/OSet/Strict.hs:127]: > Foldable requires a Monoid instance on the type-parameter, not on the class itself.

This is not the case. Replicate the Foldable instance for Sets, do not invent something new

File: libs/cardano-data/cardano-data.cabal [hunk 24]
```diff
@@ -52,13 +57,18 @@ library testlib
...
+        cardano-data
 
 test-suite cardano-data-tests
     type:             exitcode-stdio-1.0
     main-is:          Main.hs
     hs-source-dirs:   test
-    other-modules:    Test.Cardano.Data.MapExtrasSpec
+    other-modules:
+        Test.Cardano.Data.MapExtrasSpec
+        Test.Cardano.Data.OSet.Strict.Spec
```
lehins: Convention is adding a suffix, rather a separate module. (See above example `MapExtraSpec`). We should try to be consistent
```suggestion
        Test.Cardano.Data.OSet.StrictSpec
```

libs/cardano-ledger-binary/src/Cardano/Ledger/Binary/Decoding/Decoder.hs
```diff
@@ -65,6 +66,7 @@ module Cardano.Ledger.Binary.Decoding.Decoder (
   decodeVMap,
   decodeSeq,
   decodeStrictSeq,
+  decodeSetTag,
```
lehins, line 69: Changelog entry

libs/cardano-ledger-binary/src/Cardano/Ledger/Binary/Decoding/Decoder.hs
```diff
@@ -39,6 +39,7 @@ module Cardano.Ledger.Binary.Decoding.Decoder (
 
   -- ** Compatibility tools
   binaryGetDecoder,
+  allowTag,
```
lehins, line 42: Ditto

libs/cardano-ledger-binary/src/Cardano/Ledger/Binary/Encoding/Encoder.hs
```diff
@@ -39,6 +38,7 @@ module Cardano.Ledger.Binary.Encoding.Encoder (
   encodeMap,
   encodeVMap,
   encodeVector,
+  variableListLenEncoding,
```
lehins, line 41: Ditto

eras/conway/test-suite/cddl-files/conway.cddl
```diff
@@ -87,6 +87,8 @@ proposal_procedure =
   , anchor
   ]
 
+proposal_procedures = oset<proposal_procedure>
```
lehins: Same for certificates

libs/cardano-data/src/Data/OSet/Strict.hs
```diff
@@ -0,0 +1,259 @@
...
+instance Ord a => IsList (OSet a) where
+  type Item (OSet a) = a
+  fromList = fromStrictSeq . SSeq.fromList
+  toList = F.toList . osSSeq
+
+instance Foldable OSet where
+  foldr f z = foldr f z . osSSeq
+
+instance EncCBOR a => EncCBOR (OSet a) where
+  encCBOR oset = encodeTag setTag <> variableListLenEncoding (Set.size $ osSet oset) (foldMap' encCBOR oset)
```
lehins: Let's add some round trip testing for serialization.

libs/cardano-data/src/Data/OSet/Strict.hs
```diff
@@ -0,0 +1,259 @@
...
+instance Ord a => Monoid (OSet a) where
+  mempty = empty
+
+instance Ord a => IsList (OSet a) where
+  type Item (OSet a) = a
+  fromList = fromStrictSeq . SSeq.fromList
+  toList = F.toList . osSSeq
+
+instance Foldable OSet where
+  foldr f z = foldr f z . osSSeq
```
lehins: It is better to bypass the derived instance for StrictSeq and go directly to Seq
```suggestion
  foldr f z = foldr f z . SSeq.fromStrict . osSSeq
```

Do the same for all other functions from Foldable that are custom for Seq, otherwise default implementation will be used, which will likely not going tot be as efficient.
Also add `INLINE` pragmas for all of them. This will ensure there is no overhead in foldable due to our abstraction.

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Gov.hs
```diff
@@ -304,8 +304,8 @@ govTransition = do
   TRC (GovEnv txid currentEpoch pp prevGovActionIds, st, gp) <- judgmentContext
   expectedNetworkId <- liftSTS $ asks networkId
 
-  let applyProps st' Empty = pure st'
-      applyProps st' ((idx, ProposalProcedure {..}) :<| ps) = do
+  let applyProps st' OSet.Empty = pure st'
+      applyProps st' ((idx, ProposalProcedure {..}) OSet.:<|: ps) = do
```
lehins: Revert back as discussed in the call

eras/conway/test-suite/cddl-files/mock/extras.cddl
```diff
@@ -3,6 +3,8 @@ set<a> = [a]
 
 nonempty_set<a> = [+ a]
 
+oset<a> = #6.258([+ a])
```
lehins:
```suggestion
nonempty_oset<a> = #6.258([+ a])
```

libs/cardano-data/src/Data/OSet/Strict.hs
```diff
@@ -0,0 +1,259 @@
...
+  , osSet :: !(Set.Set a)
+  }
+  deriving stock (Eq, Ord, Show, Generic)
+  deriving (ToExpr, NoThunks, NFData)
+
+instance Ord a => Semigroup (OSet a) where
+  (<>) = (|><)
+
+instance Ord a => Monoid (OSet a) where
+  mempty = empty
```
lehins, line 74: Let's add laws checks with something like this: https://github.com/input-output-hk/cardano-ledger/blob/70ee1a2c2b717b6a57817464afdd7babf65b48f0/libs/vector-map/test/Test/Common.hs#L26-L30

libs/cardano-data/src/Data/OSet/Strict.hs
```diff
@@ -0,0 +1,259 @@
...
+    loop :: (t -> Decoder s (t, Bool)) -> t -> OSet a -> Decoder s (OSet a)
+    loop condition prevStep !acc = do
+      (nextStep, shouldStop) <- condition prevStep
+      if shouldStop
+        then pure acc
+        else do
+          a <- decodeValue
+          when (a `member` acc) $ fail "Duplicate value found while decoding OSet"
+          loop condition nextStep (acc |> a)
+{-# INLINE decodeOSet #-}
```
lehins: Let's avoid duplication by creating a helper function in cardano-ledger-binary

libs/cardano-data/cardano-data.cabal
```diff
@@ -33,14 +34,18 @@ library
         base >=4.14 && <4.19,
         aeson,
         cardano-ledger-binary >=1.0,
```
lehins: Update the lower bound

lehins (changes requested):
Looks awesome!!! :smile:

lehins (approved):
Very nice!

libs/cardano-ledger-binary/src/Cardano/Ledger/Binary/Decoding/Decoder.hs
```diff
@@ -833,22 +836,33 @@ decodeSetEnforceNoDuplicates ::
   Ord a =>
   Decoder s a ->
   Decoder s (Set.Set a)
-decodeSetEnforceNoDuplicates decodeElement = do
+decodeSetEnforceNoDuplicates decodeElement =
+  decodeDeduplicate decodeElement Set.member Set.insert -- do
+{-# INLINE decodeSetEnforceNoDuplicates #-}
+
+decodeDeduplicate ::
```
lehins, line 843: This is actually a bad name for this decoder, since deduplication means removal of duplicates, where this decoder will fail on duplicates (that is why it was called enforceNoDuplicates)

I've renamed it in my PR #3801 instead of blocking this PR from being merged.

aniketd: Thanks for fixing it! :+1: :heart:

PR #3781, opened 2023-10-06
CommitteeState query

### LINKED ISSUE #3750: Replace CommitteeState with CommitteMembersState query
Current implementation of `queryCommitteeState` is confusing at best, but at the worst provides incomplete story. There is also a place in the EnactState where store the actual cold keys for the current committee and their respective expiration epoch numbers.

https://github.com/input-output-hk/cardano-ledger/blob/15ccf21728173060576aca9799c06eef24b143df/libs/cardano-ledger-api/src/Cardano/Ledger/Api/State/Query.hs#L121-L123

We need to change this query to this semantics:

```haskell

-- | Query committee members
queryCommitteeMembersState ::
  Set (Credential 'ColdCommitteeRole c) ->
  -- ^ When empty, do not filter by the cold credential, otherwise return only members
  -- that are in this set
  Set (Credential 'HotCommitteeRole c) ->
  -- ^ When empty, do not filter by the cold credential, otherwise return only members
  -- that are in this set
  Maybe MemberStatus ->
  -- ^ Only return members with this status. Useful for discovering only active members,
  -- for example. When Nothing no filtering by status will occur.
  NewEpochState era ->
  CommitteeMembersState era
queryCommitteeMembersState = ...

data HotCredAuthStatus c
  = MemberHaveNotAuthorized
  -- ^ Member have been Enacted, but no hot credential for voting have been authorized
  | MemberResigned
  | Authorized (Credential 'HotCommitteeRole c)

data MemberStatus
  = -- | This can happen when someone has authorized a hot credential for an unknown cold
    -- credential. Such Committee member will either be removed from the state at the next
    -- epoch boundary or be enacted as a new member.
    Unrecognized
  | -- | This is an expired member
    Expired
  | -- | VOtes of this Commitee member wi
    Active

data NextEpochChange
  = -- | This committee member has not been enacted yet, but it will be at the next epoch.
    ToBeEnacted
  | -- | This committee member has will be removed from the state. Either because it wa
    ToBeRemoved
  | -- | Nothing expected to change for the committee member
    NoChangeExpected

data CommitteeMembersState c = CommitteeMemberState
  { cmsHotCredAuthStatus :: HotCredAuthStatus c
  , cmsStatus :: MemberStatus
  , cmsExpiration :: Maybe EpochNo
  -- ^ Absolute epoch number when the member will or have expired
  , csNextEpochChange :: NextEpochChange
  -- ^ Are there any changes planned for this Constitutional Committee member at the next
  -- epoch
  }

data CommitteeState c = CommitteState
  { csCommittee :: Map (Credential 'ColdCommitteeRole c) (CommitteeMemberState c)
  , csEpochNo :: EpochNo
  -- ^ Current epoch number. This is necessary to interpret states of some of the
  -- Committee members
  }
```

Result type will need a stable ToCBOR/FromCBOR instances as well as To/FromJSON instances.

teodanciu:
# Description

Closes https://github.com/input-output-hk/cardano-ledger/issues/3750 

~Still draft because I"m uncertain about a few things, to be discussed tomorrow.~

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/libs/cardano-ledger-api/cardano-ledger-api.cabal
+++ b/libs/cardano-ledger-api/cardano-ledger-api.cabal
@@ -69,7 +69,10 @@ library
         text
 
 library testlib
-    exposed-modules:  Test.Cardano.Ledger.Api.State.Query
+    exposed-modules:
+        Test.Cardano.Ledger.Api.State.Query
+        Test.Cardano.Ledger.Api.Arbitrary
+
     visibility:       public
     hs-source-dirs:   testlib
     default-language: Haskell2010
@@ -81,7 +84,8 @@ library testlib
     build-depends:
         base,
         containers,
-        cardano-ledger-core
+        cardano-ledger-api,
+        cardano-ledger-core:{cardano-ledger-core, testlib}
 
 test-suite cardano-ledger-api-test
     type:             exitcode-stdio-1.0
@@ -107,7 +111,7 @@ test-suite cardano-ledger-api-test
         cardano-ledger-babbage:testlib,
         cardano-ledger-core:{cardano-ledger-core, testlib},
         cardano-ledger-conway:{cardano-ledger-conway, testlib},
-        cardano-ledger-shelley,
+        cardano-ledger-shelley:{cardano-ledger-shelley, testlib},
         cardano-strict-containers,
         containers,
         microlens
--- a/libs/cardano-ledger-api/cardano-ledger-api.cabal
+++ b/libs/cardano-ledger-api/cardano-ledger-api.cabal
@@ -40,7 +40,10 @@ library
         Cardano.Ledger.Api.UTxO
 
     hs-source-dirs:   src
-    other-modules:    Cardano.Ledger.Api.Scripts.ExUnits
+    other-modules:
+        Cardano.Ledger.Api.Scripts.ExUnits
+        Cardano.Ledger.Api.State.Query.CommitteeMembersState
+
     default-language: Haskell2010
     ghc-options:
         -Wall -Wcompat -Wincomplete-record-updates
@@ -49,6 +52,7 @@ library
 
     build-depends:
         base >=4.14 && <4.19,
+        aeson,
         array,
         cardano-ledger-allegra >=1.1,
         cardano-ledger-alonzo >=1.4,
--- a/libs/cardano-ledger-api/src/Cardano/Ledger/Api/State/Query.hs
+++ b/libs/cardano-ledger-api/src/Cardano/Ledger/Api/State/Query.hs
@@ -1,5 +1,7 @@
 {-# LANGUAGE DataKinds #-}
 {-# LANGUAGE GADTs #-}
+{-# LANGUAGE LambdaCase #-}
+{-# LANGUAGE ScopedTypeVariables #-}
 
 module Cardano.Ledger.Api.State.Query (
   -- * @GetFilteredDelegationsAndRewardAccounts@
@@ -23,9 +25,24 @@ module Cardano.Ledger.Api.State.Query (
 
   -- * @GetCommitteeState@
   queryCommitteeState,
+
+  -- * @GetCommitteeState@
+  queryCommitteeMembersState,
+  CommitteeMemberState (..),
+  CommitteeMembersState (..),
+  HotCredAuthStatus (..),
+  MemberStatus (..),
+  NextEpochChange (..),
 ) where
 
 import Cardano.Ledger.Allegra.Core (Constitution (constitutionAnchor))
+import Cardano.Ledger.Api.State.Query.CommitteeMembersState (
+  CommitteeMemberState (..),
+  CommitteeMembersState (..),
+  HotCredAuthStatus (..),
+  MemberStatus (..),
+  NextEpochChange (..),
+ )
 import Cardano.Ledger.CertState
 import Cardano.Ledger.Coin (Coin)
 import Cardano.Ledger.Compactible (fromCompact)
@@ -34,17 +51,27 @@ import Cardano.Ledger.Credential (Credential)
 import Cardano.Ledger.DRepDistr (drepExpiryL)
 import Cardano.Ledger.Keys (KeyHash, KeyRole (..))
 import Cardano.Ledger.SafeHash (SafeHash)
-import Cardano.Ledger.Shelley.Governance (EraGov (GovState, getConstitution, getDRepDistr))
+import Cardano.Ledger.Shelley.Governance (
+  EraGov (
+    GovState,
+    getCommitteeMembers,
+    getConstitution,
+    getDRepDistr
+  ),
+ )
 import Cardano.Ledger.Shelley.LedgerState
 import Cardano.Ledger.UMap (
   StakeCredentials (scRewards, scSPools),
   UMap,
   domRestrictedStakeCredentials,
  )
+import Cardano.Slotting.Slot (EpochNo)
 import Data.Map (Map)
 import qualified Data.Map.Strict as Map
+import Data.Maybe (maybeToList)
 import Data.Set (Set)
-import Lens.Micro
+import qualified Data.Set as Set
+import Lens.Micro ((%~), (&), (<&>), (^.))
 
 -- | Filter out stake pool delegations and rewards for a set of stake credentials
 filterStakePoolDelegsAndRewards ::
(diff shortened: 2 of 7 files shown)


libs/cardano-ledger-api/test/Test/Cardano/Ledger/Api/State/QuerySpec.hs
```diff
@@ -16,3 +44,212 @@ spec = do
...
+        Map.fromList $
+          zipWith
+            (\(k1, _) (_, v2) -> (k1, v2))
+            (Map.assocs chosen)
+            (Map.assocs comStateMembers)
+  pure $ CommitteeState $ x `Map.union` (Map.drop s comStateMembers)
+
+propEmpty :: NewEpochState Conway -> Expectation
+propEmpty nes =
+  do
```
lehins: Could you please avoid this style?
Makes it look like we are using 4 space indentation
```suggestion
propEmpty nes = do
```

libs/cardano-ledger-api/test/Test/Cardano/Ledger/Api/State/QuerySpec.hs
```diff
@@ -16,3 +44,212 @@ spec = do
...
+    let (comMembers, comStateMembers) = committeeInfo nes
+    let (CommitteeMembersState result _) = queryCommitteeMembersStateNoFilters nes
```
lehins: Redundant parenthesis
```suggestion
    let CommitteeMembersState result _ = queryCommitteeMembersStateNoFilters nes
```

libs/cardano-ledger-api/test/Test/Cardano/Ledger/Api/State/QuerySpec.hs
```diff
@@ -16,3 +44,212 @@ spec = do
...
+    let (_, comStateMembers) = committeeInfo nes
+    let (CommitteeMembersState result _) = queryCommitteeMembersStateNoFilters nes
+    let notAuthorized =
+          Map.filter
+            ( \case
+                CommitteeMemberState MemberNotAuthorized _ _ _ -> True
+                _ -> False
+            )
+            result
+    -- if the member is NotAuthorized, it should not have an associated hot credential in the commiteSTate
```
lehins:
```suggestion
    -- if the member is NotAuthorized, it should not have an associated hot credential in the commiteState
```

libs/cardano-ledger-api/src/Cardano/Ledger/Api/State/CommitteeMembersState.hs
```diff
@@ -0,0 +1,213 @@
...
+instance DecCBOR MemberStatus where
+  decCBOR = decodeEnumBounded
+
+instance ToCBOR MemberStatus where
+  -- toCBOR = toPlainEncoding (eraProtVerLow @Conway) . encCBOR
+  toCBOR = toEraCBOR @Conway
+
+instance FromCBOR MemberStatus where
+  -- fromCBOR = toPlainDecoder (eraProtVerLow @Conway) decCBOR
+  fromCBOR = fromEraCBOR @Conway
```
lehins: These instances can't exist, since there is no access to `era` parameter. At the same time they are not needed either. At the place where they will be serialized/deserialized `era` will be available and that is where protocol version can be inferred. Here is an example how it works:
https://github.com/input-output-hk/ouroboros-consensus/blob/903ba3c7d2f8d8c015b59f0c664cf75c0f280c25/ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Query.hs#L812
```suggestion
```

libs/cardano-ledger-api/src/Cardano/Ledger/Api/State/CommitteeMembersState.hs
```diff
@@ -0,0 +1,213 @@
...
+      0 -> SumD MemberAuthorized <! From
+      1 -> SumD MemberNotAuthorized
+      2 -> SumD MemberResigned
+      k -> Invalid k
+
+instance Crypto c => ToCBOR (HotCredAuthStatus c) where
+  toCBOR = toEraCBOR @Conway
+
+instance Crypto c => FromCBOR (HotCredAuthStatus c) where
+  fromCBOR = fromEraCBOR @Conway
```
lehins: Same here
```suggestion
```

libs/cardano-ledger-api/src/Cardano/Ledger/Api/State/CommitteeMembersState.hs
```diff
@@ -0,0 +1,213 @@
...
+        <! From
+        <! From
+        <! From
+        <! From
+
+instance Crypto c => ToCBOR (CommitteeMemberState c) where
+  toCBOR = toEraCBOR @Conway
+
+instance Crypto c => FromCBOR (CommitteeMemberState c) where
+  fromCBOR = fromEraCBOR @Conway
```
lehins: Ditto
```suggestion
```

libs/cardano-ledger-api/src/Cardano/Ledger/Api/State/CommitteeMembersState.hs
```diff
@@ -0,0 +1,213 @@
...
+    decode $
+      RecD CommitteeMembersState
+        <! From
+        <! From
+
+instance Crypto c => ToCBOR (CommitteeMembersState c) where
+  toCBOR = toEraCBOR @Conway
+
+instance Crypto c => FromCBOR (CommitteeMembersState c) where
+  fromCBOR = fromEraCBOR @Conway
```
lehins: Ditto
```suggestion
```

libs/cardano-ledger-api/src/Cardano/Ledger/Api/State/CommitteeMembersState.hs
```diff
@@ -0,0 +1,213 @@
...
+  encCBOR = encodeEnum
+
+instance DecCBOR NextEpochChange where
+  decCBOR = decodeEnumBounded
+
+instance ToCBOR NextEpochChange where
+  toCBOR = toEraCBOR @Conway
+
+instance FromCBOR NextEpochChange where
+  fromCBOR = fromEraCBOR @Conway
```
lehins: Ditto
```suggestion
```

libs/cardano-ledger-api/src/Cardano/Ledger/Api/State/CommitteeMembersState.hs
```diff
@@ -0,0 +1,213 @@
...
+      , "expiration" .= cmsExpiration
+      , "nextEpochChange" .= cmsNextEpochChange
+      ]
+
+data CommitteeMembersState c = CommitteeMembersState
+  { csCommittee ::
+      Map
+        (Credential 'ColdCommitteeRole c)
+        (CommitteeMemberState c)
+  , csEpochNo :: EpochNo
```
lehins: Let's make 'em strict and this line isn't too long yet
```suggestion
  { csCommittee :: !(Map (Credential 'ColdCommitteeRole c) (CommitteeMemberState c))
  , csEpochNo :: !EpochNo
```

libs/cardano-ledger-api/src/Cardano/Ledger/Api/State/CommitteeMembersState.hs
```diff
@@ -0,0 +1,213 @@
...
+instance FromCBOR NextEpochChange where
+  fromCBOR = fromEraCBOR @Conway
+
+data CommitteeMemberState c = CommitteeMemberState
+  { cmsHotCredAuthStatus :: HotCredAuthStatus c
+  , cmsStatus :: MemberStatus
+  , cmsExpiration :: Maybe EpochNo
+  -- ^ Absolute epoch number when the member expires
+  , cmsNextEpochChange :: NextEpochChange
+  -- ^ Changes to the member at the next epoch
```
lehins:
```suggestion
  { cmsHotCredAuthStatus :: !(HotCredAuthStatus c)
  , cmsStatus :: !MemberStatus
  , cmsExpiration :: !(Maybe EpochNo)
  -- ^ Absolute epoch number when the member expires
  , cmsNextEpochChange :: !NextEpochChange
  -- ^ Changes to the member at the next epoch
```

libs/cardano-ledger-api/src/Cardano/Ledger/Api/State/Query.hs
```diff
@@ -121,3 +143,108 @@ queryDRepStakeDistr nes creds
...
+          else comMembers `Map.restrictKeys` coldCredsFilter
+
+      -- Start from the committee and lookup the hot credential in the Committee State.
+      -- This will miss 'Unrecognized' members (since we're iterating over the "recognized" ones).
+      recognized :: [(Credential 'ColdCommitteeRole (EraCrypto era), CommitteeMemberState (EraCrypto era))]
+      recognized =
+        Map.assocs filteredComMembers >>= \(ck, expirationEpoch) ->
+          let mbHk = Map.lookup ck comStateMembers
+              mbCommitteeMemberState =
+                do
```
```suggestion
              mbCommitteeMemberState = do
```

libs/cardano-ledger-api/src/Cardano/Ledger/Api/State/Query.hs
```diff
@@ -121,3 +143,108 @@ queryDRepStakeDistr nes creds
...
+                      CommitteeMemberState
+                        hkStatus
+                        Unrecognized
+                        Nothing
+                        nextEpochChange
+                pure (ck, cms)
+           in maybeToList mbCommitteeMemberState
+   in CommitteeMembersState
+        { csCommittee = Map.fromList (recognized ++ unrecognized)
+        , csEpochNo = nes ^. nesELL - 1
```
lehins: Why is this change to previous epoch? This query suppose to be for the current state.
```suggestion
        , csEpochNo = nes ^. nesELL
```

libs/cardano-ledger-api/src/Cardano/Ledger/Api/State/Query.hs
```diff
@@ -121,3 +143,108 @@ queryDRepStakeDistr nes creds
...
+                  pure (ck, cms)
+           in maybeToList mbCommitteeMemberState
+
+      filteredUnrecognizedMembers =
+        if null coldCredsFilter
+          then Map.difference comStateMembers comMembers
+          else
+            Map.restrictKeys
+              (Map.difference comStateMembers comMembers)
+              coldCredsFilter
```
lehins: Slight reduction in duplication. Also usage of guards as in binding below
```suggestion
      unrecognizedMembers = Map.difference comStateMembers comMembers
      filteredUnrecognizedMembers
        | null coldCredsFilter = unrecognizedMembers
        | otherwise = Map.restrictKeys unrecognizedMembers coldCredsFilter
```

libs/cardano-ledger-api/src/Cardano/Ledger/Api/State/Query.hs
```diff
@@ -121,3 +143,108 @@ queryDRepStakeDistr nes creds
...
+  -- | filter by cold credentials (don't filter when empty)
+  Set (Credential 'ColdCommitteeRole (EraCrypto era)) ->
+  -- | filter by hot credentials (don't filter when empty)
+  Set (Credential 'HotCommitteeRole (EraCrypto era)) ->
+  -- | filter by status (don't filter when empty)
+  -- (useful, for discovering, for example, only active members)
+  Set MemberStatus ->
+  NewEpochState era ->
+  CommitteeMembersState (EraCrypto era)
+queryCommitteeMembersState coldCredsFilter hotCredsFilter statusFilter nes =
```
lehins, line 165: Current implementation of this query is a bit complicated and it is hard to follow. I suggest that it will be easier to start with a set of relevant cold keys and convert it to a Map with some filtering, something along these lines:
```haskell

      relevantHotCreds :: Set (Credential 'ColdCommitteeRole (EraCrypto era))
      relevantHotCreds
        | null hotCredsFilter = comStateMembers
        | otherwise = Map.filter (maybe False (`Set.member` hotCredsFilter)) comStateMembers

      relevantColdCreds :: Set (Credential 'ColdCommitteeRole (EraCrypto era))
      relevantColdCreds
        -- we only look for Unrecognized members when it is contained in the filter, or when there is no filtering
        | Set.null statusFilter || Set.member Unrecognized statusFilter =
            Map.keysSet comMembers `Set.union` Map.keysSet relevantHotCredsMap
        | otherwise = Map.keysSet comMembers

      filteredColdCreds
        | null coldCredsFilter = relevantColdCreds
        | otherwise = relevantColdCreds `Map.restrictKeys` coldCredsFilter

      committeMemeberState = 
        Map.mapMaybe id $ Map.fromSet mkMaybeMemberState filteredColdCreds

      mkMaybeMemberState :: Credential 'ColdCommitteeRole (EraCrypto era) ->
                            Maybe (CommitteeMemberState (EraCrypto era))
      mkMaybeMemberState coldCred = do
        let status =
              case Map.lookup coldCred comMembers of
                Nothing -> Unrecognized
                Just expiry | nes ^. nesELL > expiry -> Expired
                            | otherwise -> Active
        guard (null statusFilter || status `Set.member` statusFilter)
        ...
```
This way you can focus on a single cold credential at a time in `mkMaybeMemberState`, which only needs to filter by status and you can use `do` syntax for Maybe to keep it readable.

lehins (changes requested):
> I went down the wrong path

No worries, it happens.

This PR is on the right path :wink:

libs/cardano-ledger-api/cardano-ledger-api.cabal
```diff
@@ -36,6 +36,7 @@ library
         Cardano.Ledger.Api.Tx.In
         Cardano.Ledger.Api.Tx.Out
         Cardano.Ledger.Api.Tx.Wits
+        Cardano.Ledger.Api.State.CommitteeMembersState
```
lehins: This should be a hidden module with all types re-exported from `Cardano.Ledger.Api.State.Query`

libs/cardano-ledger-api/src/Cardano/Ledger/Api/State/Query.hs
teodanciu: I wrote it like this because of [how we create the RatifyEnv in Ledger](https://github.com/input-output-hk/cardano-ledger/blob/master/eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Epoch.hs#L300), by passing it `currentEpoch - 1`, but I suppose, from the perspective of the caller of the Query, the epochNo is the one from the LedgerState.

libs/cardano-ledger-api/src/Cardano/Ledger/Api/State/Query.hs
teodanciu: Nice suggestion, thank you. 
I shuffled it around, because in this form it's  a little wrong (the `union` after the `filter` overwrites the filter), but I kept the same idea.

libs/cardano-ledger-api/cardano-ledger-api.cabal
```diff
@@ -40,7 +40,10 @@ library
         Cardano.Ledger.Api.UTxO
 
     hs-source-dirs:   src
-    other-modules:    Cardano.Ledger.Api.Scripts.ExUnits
+    other-modules:
+        Cardano.Ledger.Api.Scripts.ExUnits
+        Cardano.Ledger.Api.State.CommitteeMembersState
```
lehins: Could you move this module into a Query subfolder, since we'll have a lot more State, but not query related modules in the `Cardano/Ledger/Api/State` path
```suggestion
        Cardano.Ledger.Api.State.Query.CommitteeMembersState
```

libs/cardano-ledger-api/test/Test/Cardano/Ledger/Api/State/QuerySpec.hs
```diff
@@ -16,3 +45,226 @@ spec = do
...
+
+queryCommitteeMembersStateNoFilters :: NewEpochState Conway -> CommitteeMembersState StandardCrypto
+queryCommitteeMembersStateNoFilters =
+  queryCommitteeMembersState @Conway
+    Set.empty
+    Set.empty
+    Set.empty
+
+instance Arbitrary MemberStatus where
+  arbitrary = arbitraryBoundedEnum
```
lehins: Can you add a module to the `testlib`: `Test.Cardano.Ledger.Api.Arbitrary` and move this instance in there.

lehins (approved):
Looks great. Thank you!

libs/cardano-ledger-api/test/Test/Cardano/Ledger/Api/State/QuerySpec.hs
teodanciu: I actually did that at first, and when I saw it's only this that needs an instance ,I thought it's "not worth it", but I agree, I prefer it in a different module, even if it's two lines.  That's how big messes start, line by line :laughing: 
Moved it now, thanks for suggesting .

libs/cardano-ledger-api/cardano-ledger-api.cabal
teodanciu: Yes, sorry for forgetting, I know you told me yesterday. Done

PR #3782, opened 2023-10-06
Move `ImpTest` to Shelley testlib

Soupstraw:
# Description

Moved `ImpTest` to core and added the `EraImpTest` class together with an instance in the `Conway` era. Moved some tests into `ImpTestSpec` in Shelley and made it run in every era.

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/eras/alonzo/impl/CHANGELOG.md
+++ b/eras/alonzo/impl/CHANGELOG.md
@@ -15,6 +15,9 @@
 
 ### `testlib`
 
+* Add `emptyAlonzoImpNES`
+* Add `Test.Cardano.Ledger.Alonzo.ImpTest`
+* Add `EraImpTest` instance for `AlonzoEra`
 * Add `ToExpr` instances for:
   * `CollectError`
   * `AlonzoUtxoPredFailure`
--- a/eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/ImpTest.hs
+++ b/eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/ImpTest.hs
@@ -1,682 +1,414 @@
-{-# LANGUAGE BangPatterns #-}
 {-# LANGUAGE DataKinds #-}
-{-# LANGUAGE DeriveFunctor #-}
-{-# LANGUAGE DeriveGeneric #-}
 {-# LANGUAGE FlexibleContexts #-}
-{-# LANGUAGE GADTs #-}
-{-# LANGUAGE GeneralizedNewtypeDeriving #-}
 {-# LANGUAGE NamedFieldPuns #-}
-{-# LANGUAGE OverloadedStrings #-}
-{-# LANGUAGE RankNTypes #-}
+{-# LANGUAGE PatternSynonyms #-}
 {-# LANGUAGE RecordWildCards #-}
 {-# LANGUAGE ScopedTypeVariables #-}
-{-# LANGUAGE StandaloneDeriving #-}
 {-# LANGUAGE TypeApplications #-}
-{-# LANGUAGE TypeOperators #-}
+{-# LANGUAGE TypeFamilies #-}
 {-# LANGUAGE UndecidableInstances #-}
+{-# OPTIONS_GHC -Wno-orphans #-}
 
 module Test.Cardano.Ledger.Conway.ImpTest (
-  ImpTestM,
-  ImpException (..),
-  passEpoch,
-  passTick,
-  itM,
-  freshKeyHash,
-  freshSafeHash,
-  lookupKeyPair,
-  submitTx,
-  submitFailingTx,
-  submitBasicConwayTx,
-  logEntry,
-  modifyNES,
-  getsNES,
-  impIO,
-  impIOMsg,
+  module ImpTest,
   submitProposal,
   submitFailingProposal,
+  voteForProposal,
+  registerDRep,
+  setupSingleDRep,
+  getEnactState,
+  lookupGovActionState,
+  getRatifyEnv,
+  calculateDRepAcceptedRatio,
+  calculateCommitteeAcceptedRatio,
+  logAcceptedRatio,
+  isGovActionAccepted,
+  logRatificationChecks,
+  registerCCHotKey,
 ) where
 
-import Cardano.Crypto.DSIGN (DSIGNAlgorithm (..), seedSizeDSIGN)
-import Cardano.Crypto.Hash (Hash)
-import Cardano.Crypto.Seed (mkSeedFromBytes)
-import qualified Cardano.Crypto.VRF.Class as VRF
-import Cardano.Ledger.Address (Addr (..), RewardAcnt (RewardAcnt))
-import Cardano.Ledger.Babbage.TxBody (BabbageTxOut (..), Datum (..))
-import Cardano.Ledger.BaseTypes (
-  BlocksMade (..),
-  EpochSize (..),
-  Globals (..),
-  Network (..),
-  ShelleyBase,
-  SlotNo,
-  mkActiveSlotCoeff,
- )
+import Cardano.Crypto.DSIGN.Class (Signable)
+import Cardano.Crypto.Hash.Class (Hash)
+import Cardano.Ledger.Address (RewardAcnt (..))
+import Cardano.Ledger.BaseTypes (Network (..), StrictMaybe (..))
+import Cardano.Ledger.CertState (DRep (..))
 import Cardano.Ledger.Coin (Coin (..))
-import Cardano.Ledger.Conway (Conway)
+import Cardano.Ledger.Conway (ConwayEra)
 import Cardano.Ledger.Conway.Core (
   Era (..),
   EraIndependentTxBody,
-  EraPParams (..),
-  EraRule,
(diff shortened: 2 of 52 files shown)


libs/cardano-ledger-core/src/Cardano/Ledger/UTxO.hs
```diff
@@ -252,7 +253,8 @@ class EraTx era => EraUTxO era where
   -- for the transaction to be valid, once those scripts are evaluated
   getScriptsNeeded :: UTxO era -> TxBody era -> ScriptsNeeded era
 
-  -- | Extract the set of all script hashes that are needed for script validation.
+  witsVKeyNeeded :: UTxO era -> TxBody era -> GenDelegs (EraCrypto era) -> Set (KeyHash 'Witness (EraCrypto era))
```
lehins: Why is this change needed? Especially how is this relevant for the work on a test suite?

I am against adding such function to this type class, unless there is a really strong reason for it. Problem with placing this function into a type class is that it suppose work for all eras, however, starting with Conway we have no notion of `GenDelegs`.

libs/cardano-ledger-core/src/Cardano/Ledger/UTxO.hs
```diff
@@ -252,7 +253,8 @@ class EraTx era => EraUTxO era where
   -- for the transaction to be valid, once those scripts are evaluated
   getScriptsNeeded :: UTxO era -> TxBody era -> ScriptsNeeded era
 
-  -- | Extract the set of all script hashes that are needed for script validation.
```
lehins: Why was the haddock deleted?

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/UTxO.hs
```diff
@@ -56,6 +57,9 @@ newtype AlonzoScriptsNeeded era
   = AlonzoScriptsNeeded [(ScriptPurpose era, ScriptHash (EraCrypto era))]
   deriving (Semigroup, Monoid)
 
+alonzoScriptsSet :: AlonzoScriptsNeeded era -> Set.Set (ScriptHash (EraCrypto era))
+alonzoScriptsSet (AlonzoScriptsNeeded sn) = Set.fromList $ snd <$> sn
```
lehins: Why recreating this function and replacing with the identical implementation of `getAlonzoScriptsHashesNeeded`?

eras/alonzo/impl/CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## 1.5.0.0
 
+* Add `alonzoScriptsSet`
```
lehins: Again, `getAlonzoScriptsHashesNeeded` is already exported and available for use outside of `cardano-ledger-alonzo`
```suggestion
```

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Utxow.hs
```diff
@@ -433,6 +420,7 @@ validateMissingScripts pp (ShelleyScriptsNeeded sNeeded) scriptsprovided =
         MissingScriptWitnessesUTXOW (sNeeded `Set.difference` sProvided)
   where
     sProvided = Map.keysSet $ unScriptsProvided scriptsprovided
+    sNeeded = getScriptsHashesNeeded sn
```
lehins: That's a nice improvement of consistency. :+1:

eras/shelley/impl/testlib/Test/Cardano/Ledger/Shelley/ImpTest.hs
```diff
@@ -0,0 +1,604 @@
...
+impRootTxIdL :: Lens' (ImpTestState era) (TxId (EraCrypto era))
+impRootTxIdL = lens impRootTxId (\x y -> x {impRootTxId = y})
+
+class
+  ( Show (NewEpochState era)
+  , ToExpr (NewEpochState era)
+  , Signal (EraRule "LEDGER" era) ~ Tx era
+  , BaseM (EraRule "LEDGER" era) ~ ReaderT Globals Identity
+  , STS (EraRule "LEDGER" era)
+  , Era era
```
lehins: implied by `EraGov` and `EraUTxO`
```suggestion
```

libs/cardano-ledger-core/testlib/Test/Cardano/Ledger/Core/Utils.hs
```diff
@@ -1,9 +1,28 @@
+{-# LANGUAGE RankNTypes #-}
```
lehins: It's not used in this module 
```suggestion
```

libs/cardano-ledger-core/CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## 1.8.0.0
 
+* Add `witsVKeyNeeded` to `EraTx`
```
lehins: This function should not be part of `EraUTxO` (nor should it be part of `EraTx` :smile:)
```suggestion
```

eras/shelley/impl/testlib/Test/Cardano/Ledger/Shelley/ImpTest.hs
```diff
@@ -0,0 +1,604 @@
...
+    impIOMsg "Failed to run TICK" $
+      pure $
+        runShelleyBase (applySTSTest @(EraRule "TICK" era) trc)
+  case res of
+    Right !x -> do
+      impLastTickL += 1
+      impNESL .= x
+    Left e ->
+      impIO . error $
+        "Failed to run TICK:\n" <> unlines (show <$> e)
```
```suggestion
        unlines $ "Failed to run TICK:" : map show e
```

lehins (changes requested):
Overall this looks good. I don't really like reshuffling of `witsVKeyNeeded`.

libs/cardano-ledger-core/src/Cardano/Ledger/UTxO.hs
Soupstraw: I used it to construct the set of witnesses for a transaction in the `mkTxWits` function. I also noticed that we had copied the whole `Utxow` rule just to swap out the `witsVKeyNeeded` function in Conway. I figured that it makes sense to have this as a class method anyways so we can better reuse code across eras.

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/ImpTest.hs
```diff
@@ -1,682 +1,375 @@
...
+
+-- | Logs the ratios of accepted votes per category
+logAcceptedRatio :: ConwayEraGov era => GovActionId (EraCrypto era) -> ImpTestM era ()
+logAcceptedRatio aId = do
+  dRepRatio <- calculateDRepAcceptedRatio aId
+  committeeRatio <- calculateCommitteeAcceptedRatio aId
+  logEntry "----- ACCEPTED RATIOS -----"
+  logEntry $ "DRep accepted ratio:\t\t" <> show dRepRatio
+  logEntry $ "Committee accepted ratio:\t" <> show committeeRatio
+  logEntry ""
```
lehins, line 347: Why not add a function that can be applied to lists?
```suggestion
  logEntries 
    [ "----- ACCEPTED RATIOS -----"
    , "DRep accepted ratio:\t\t" <> show dRepRatio
    , "Committee accepted ratio:\t" <> show committeeRatio
    ]
```

lehins: Or even better takes a special `header` argument and adds the dashes around it:
```haskell
logEntries header xs = ...
```
would result in: 
```
----- Header ------
```

libs/cardano-ledger-core/src/Cardano/Ledger/UTxO.hs
lehins: As we discussed in a call, this is far from being the only reason why the Utxow was recreated from scratch in almost every era. There are major differences in all of them.

Since this functionality is needed in tests it should be solved in tests. So, adding such function to a type class in a test suite is a totally sensible solution.

Also a request for the future. Could you please separate any serious changes like that, which affect the rules, into a separate smaller PR. Mixing restructures like that to the rules with refactoring of test suites could obfuscate potential bugs.

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/ImpTest.hs
Soupstraw: I think an even better solution would be to format logs with the prettyprinter library.

lehins: Sure, that could be. We'll solve this problem at a different time. This was just a mere suggestion.

Soupstraw: I switched the logging type to `Doc ()`, this should add more flexibility to how we want to format the logs in the future.

lehins (approved):
Beautiful! Thank you!

PR #3791, opened 2023-10-12
Use a Data.OMap.Strict to replace ProposalsSnapshot

### LINKED ISSUE #3789: OMap for ProposalsSnapshot
Make an `OMap` on similar lines as the `OSet` and use that for `ProposalsSnapshot`.

aniketd:
# Description

Closes #3789

- [x] Merge #3779 first.

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Snapshots.hs
+++ b/eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Snapshots.hs
@@ -4,6 +4,7 @@
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
 {-# LANGUAGE MultiParamTypeClasses #-}
 {-# LANGUAGE RankNTypes #-}
+{-# LANGUAGE TypeOperators #-}
 {-# LANGUAGE UndecidableInstances #-}
 {-# OPTIONS_GHC -Wno-orphans #-}
 
@@ -79,28 +80,35 @@ instance EraPParams era => DecCBOR (ProposalsSnapshot era) where
 instance EraPParams era => DecShareCBOR (ProposalsSnapshot era) where
   decShareCBOR _ = fromGovActionStateSeq <$> decCBOR
 
+-- | Insert a `GovActionState`, overwriting an entry of it if the
+-- corresponding `GovActionId` already exists.
 snapshotInsertGovAction ::
   GovActionState era ->
   ProposalsSnapshot era ->
   ProposalsSnapshot era
 snapshotInsertGovAction gas (ProposalsSnapshot omap) =
-  ProposalsSnapshot (omap OMap.|> gas)
+  ProposalsSnapshot (omap OMap.||> gas)
 
+-- | Get the sequence of `GovActionState`s
 snapshotActions ::
   ProposalsSnapshot era ->
   StrictSeq (GovActionState era)
 snapshotActions (ProposalsSnapshot omap) = OMap.toStrictSeq omap
 
+-- | Get the sequence of `GovActionId`s
 snapshotIds ::
   ProposalsSnapshot era ->
   StrictSeq (GovActionId (EraCrypto era))
 snapshotIds (ProposalsSnapshot omap) = OMap.toStrictSeqOKeys omap
 
+-- | Get the unordered map of `GovActionId`s and `GovActionState`s
 snapshotGovActionStates ::
   ProposalsSnapshot era ->
   Map (GovActionId (EraCrypto era)) (GovActionState era)
 snapshotGovActionStates (ProposalsSnapshot omap) = OMap.toMap omap
 
+-- | Add a vote to an existing `GovActionState` This is a no-op if the .
+-- provided `GovActionId` does not already exist                       .
 snapshotAddVote ::
   Voter (EraCrypto era) ->
   Vote ->
@@ -122,6 +130,7 @@ snapshotAddVote voter vote gai (ProposalsSnapshot omap) =
       StakePoolVoter kh -> insertVote gasStakePoolVotesL kh
       CommitteeVoter c -> insertVote gasCommitteeVotesL c
 
+-- | Extract `GovActionState`s for the given set of `GovActionId`s from the `Proposals`
 snapshotRemoveIds ::
   Set (GovActionId (EraCrypto era)) ->
   ProposalsSnapshot era ->
@@ -145,5 +154,5 @@ fromGovActionStateSeq :: StrictSeq (GovActionState era) -> ProposalsSnapshot era
 fromGovActionStateSeq = ProposalsSnapshot . OMap.fromStrictSeq
 
 -- | Internal function for checking if the invariants are maintained
-isConsistent_ :: EraPParams era => ProposalsSnapshot era -> Bool
+isConsistent_ :: ProposalsSnapshot era -> Bool
 isConsistent_ (ProposalsSnapshot omap) = OMap.invariantHolds' omap
--- a/eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Snapshots.hs
+++ b/eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Snapshots.hs
@@ -84,7 +84,7 @@ snapshotInsertGovAction ::
   ProposalsSnapshot era ->
   ProposalsSnapshot era
 snapshotInsertGovAction gas (ProposalsSnapshot omap) =
-  ProposalsSnapshot (omap OMap.:|>: gas)
+  ProposalsSnapshot (omap OMap.|> gas)
 
 snapshotActions ::
   ProposalsSnapshot era ->
@@ -127,7 +127,7 @@ snapshotRemoveIds ::
   ProposalsSnapshot era ->
   (ProposalsSnapshot era, Map.Map (GovActionId (EraCrypto era)) (GovActionState era))
 snapshotRemoveIds gais (ProposalsSnapshot omap) =
-  let (removed, retained) = OMap.extractKeys gais omap
+  let (retained, removed) = OMap.extractKeys gais omap
    in (ProposalsSnapshot retained, removed)
 
 snapshotLookupId ::
--- a/eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Snapshots.hs
+++ b/eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Snapshots.hs
@@ -1,7 +1,11 @@
 {-# LANGUAGE DeriveGeneric #-}
-{-# LANGUAGE NamedFieldPuns #-}
+{-# LANGUAGE DerivingStrategies #-}
+{-# LANGUAGE GADTs #-}
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE MultiParamTypeClasses #-}
 {-# LANGUAGE RankNTypes #-}
-{-# LANGUAGE RecordWildCards #-}
+{-# LANGUAGE UndecidableInstances #-}
+{-# OPTIONS_GHC -Wno-orphans #-}
 
 module Cardano.Ledger.Conway.Governance.Snapshots (
   ProposalsSnapshot,
@@ -32,25 +36,27 @@ import Cardano.Ledger.TreeDiff (ToExpr)
 import Control.DeepSeq (NFData)
 import Data.Aeson (ToJSON)
 import Data.Default.Class (Default (..))
-import Data.Foldable (Foldable (..))
-import Data.List (sort)
 import qualified Data.Map as Map
 import Data.Map.Strict (Map)
-import Data.MapExtras (extractKeys)
-import Data.Maybe (fromMaybe)
+import qualified Data.OMap.Strict as OMap
 import Data.Sequence.Strict (StrictSeq (..))
 import Data.Set (Set)
-import qualified Data.Set as Set
 import GHC.Generics (Generic)
-import Lens.Micro (Lens', (%~))
+import Lens.Micro (Lens', lens, (%~))
 import NoThunks.Class (NoThunks)
 
-data ProposalsSnapshot era = ProposalsSnapshot
-  { psGovActionStates :: !(Map (GovActionId (EraCrypto era)) (GovActionState era))
-  , psProposalOrder :: !(StrictSeq (GovActionId (EraCrypto era)))
-  -- ^ Newer actions are near the end
-  }
-  deriving (Generic, Eq, Show)
+-- Ref: https://gitlab.haskell.org/ghc/ghc/-/issues/14046
+instance
+  c ~ EraCrypto era =>
+  OMap.HasOKey (GovActionId c) (GovActionState era)
+  where
+  okeyL = lens gasId $ \gas gi -> gas {gasId = gi}
+
+newtype ProposalsSnapshot era
+  = ProposalsSnapshot
+      (OMap.OMap (GovActionId (EraCrypto era)) (GovActionState era))
+  deriving newtype (Show, Eq)
+  deriving stock (Generic)
 
 instance EraPParams era => ToExpr (ProposalsSnapshot era)
 
@@ -61,7 +67,7 @@ instance EraPParams era => NFData (ProposalsSnapshot era)
 instance EraPParams era => NoThunks (ProposalsSnapshot era)
 
 instance Default (ProposalsSnapshot era) where
-  def = ProposalsSnapshot def def
+  def = ProposalsSnapshot def
 
 instance EraPParams era => EncCBOR (ProposalsSnapshot era) where
   encCBOR = encCBOR . snapshotActions
@@ -77,34 +83,23 @@ snapshotInsertGovAction ::
   GovActionState era ->
   ProposalsSnapshot era ->
   ProposalsSnapshot era
-snapshotInsertGovAction gas@GovActionState {gasId} ps@ProposalsSnapshot {..}
-  | Map.member gasId psGovActionStates =
-      ps {psGovActionStates = Map.insert gasId gas psGovActionStates}
-  | otherwise =
-      ProposalsSnapshot
-        { psGovActionStates = Map.insert gasId gas psGovActionStates
-        , psProposalOrder = psProposalOrder :|> gasId
-        }
+snapshotInsertGovAction gas (ProposalsSnapshot omap) =
+  ProposalsSnapshot (omap OMap.:|>: gas)
 
 snapshotActions ::
   ProposalsSnapshot era ->
   StrictSeq (GovActionState era)
-snapshotActions ProposalsSnapshot {..} = toGovAction <$> psProposalOrder
-  where
-    toGovAction gaId =
-      fromMaybe
-        (error $ "Impossible: ProposalsSnapshot invariant is not maintained: " <> show gaId)
-        (Map.lookup gaId psGovActionStates)
+snapshotActions (ProposalsSnapshot omap) = OMap.toStrictSeq omap
 
 snapshotIds ::
   ProposalsSnapshot era ->
   StrictSeq (GovActionId (EraCrypto era))
-snapshotIds = psProposalOrder
+snapshotIds (ProposalsSnapshot omap) = OMap.toStrictSeqOKeys omap
 
 snapshotGovActionStates ::
   ProposalsSnapshot era ->
   Map (GovActionId (EraCrypto era)) (GovActionState era)
-snapshotGovActionStates = psGovActionStates
+snapshotGovActionStates (ProposalsSnapshot omap) = OMap.toMap omap
 
 snapshotAddVote ::
   Voter (EraCrypto era) ->
@@ -112,10 +107,8 @@ snapshotAddVote ::
   GovActionId (EraCrypto era) ->
   ProposalsSnapshot era ->
   ProposalsSnapshot era
-snapshotAddVote voter vote gId ps@ProposalsSnapshot {..} =
-  ps
-    { psGovActionStates = Map.update (Just . updateVote) gId psGovActionStates
-    }
+snapshotAddVote voter vote gai (ProposalsSnapshot omap) =
+  ProposalsSnapshot $ OMap.adjust updateVote gai omap
   where
     insertVote ::
       Ord k =>
@@ -133,35 +126,24 @@ snapshotRemoveIds ::
   Set (GovActionId (EraCrypto era)) ->
   ProposalsSnapshot era ->
   (ProposalsSnapshot era, Map.Map (GovActionId (EraCrypto era)) (GovActionState era))
-snapshotRemoveIds gIds (ProposalsSnapshot {..}) = (retainedProposals, removedGovActionStates)
-  where
-    (retainedGovActionStates, removedGovActionStates) = psGovActionStates `extractKeys` gIds
-    retainedProposals =
-      ProposalsSnapshot
-        { psGovActionStates = retainedGovActionStates
-        , psProposalOrder =
-            foldl' (\s x -> if x `Set.member` gIds then s else x :<| s) mempty psProposalOrder
-        }
+snapshotRemoveIds gais (ProposalsSnapshot omap) =
+  let (removed, retained) = OMap.extractKeys gais omap
+   in (ProposalsSnapshot retained, removed)
 
 snapshotLookupId ::
   GovActionId (EraCrypto era) ->
   ProposalsSnapshot era ->
   Maybe (GovActionState era)
-snapshotLookupId gId ProposalsSnapshot {psGovActionStates} =
-  Map.lookup gId psGovActionStates
+snapshotLookupId gai (ProposalsSnapshot omap) = OMap.lookup gai omap
 
 -- | Converts a sequence of `GovActionState`s to a `ProposalsSnapshot`.
 --
 -- /Warning/ - This function expects `GovActionState`'s to have unique
 -- `GovActionId`s, because duplicate Ids will result in `GovActionStates`
 -- to be dropped.
-fromGovActionStateSeq ::
-  StrictSeq (GovActionState era) ->
-  ProposalsSnapshot era
-fromGovActionStateSeq = foldl' (flip snapshotInsertGovAction) def
+fromGovActionStateSeq :: StrictSeq (GovActionState era) -> ProposalsSnapshot era
+fromGovActionStateSeq = ProposalsSnapshot . OMap.fromStrictSeq
 
 -- | Internal function for checking if the invariants are maintained
-isConsistent_ :: ProposalsSnapshot era -> Bool
-isConsistent_ (ProposalsSnapshot {psGovActionStates, psProposalOrder}) =
-  Map.keys psGovActionStates == sort (toList psProposalOrder)
-    && all (\(k, GovActionState {gasId}) -> k == gasId) (Map.toList psGovActionStates)
+isConsistent_ :: EraPParams era => ProposalsSnapshot era -> Bool
+isConsistent_ (ProposalsSnapshot omap) = OMap.invariantHolds' omap
--- a/eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Snapshots.hs
+++ b/eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Snapshots.hs
@@ -4,9 +4,7 @@
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
 {-# LANGUAGE MultiParamTypeClasses #-}
 {-# LANGUAGE RankNTypes #-}
-{-# LANGUAGE TypeOperators #-}
 {-# LANGUAGE UndecidableInstances #-}
-{-# OPTIONS_GHC -Wno-orphans #-}
 
 module Cardano.Ledger.Conway.Governance.Snapshots (
   ProposalsSnapshot,
@@ -43,15 +41,8 @@ import qualified Data.OMap.Strict as OMap
 import Data.Sequence.Strict (StrictSeq (..))
 import Data.Set (Set)
 import GHC.Generics (Generic)
-import Lens.Micro (Lens', lens, (%~))
+import Lens.Micro (Lens', (%~))
 import NoThunks.Class (NoThunks)
-
--- Ref: https://gitlab.haskell.org/ghc/ghc/-/issues/14046
-instance
-  c ~ EraCrypto era =>
-  OMap.HasOKey (GovActionId c) (GovActionState era)
-  where
-  okeyL = lens gasId $ \gas gi -> gas {gasId = gi}
 
 newtype ProposalsSnapshot era
   = ProposalsSnapshot
@@ -151,7 +142,7 @@ snapshotLookupId gai (ProposalsSnapshot omap) = OMap.lookup gai omap
 -- `GovActionId`s, because duplicate Ids will result in `GovActionStates`
 -- to be dropped.
 fromGovActionStateSeq :: StrictSeq (GovActionState era) -> ProposalsSnapshot era
-fromGovActionStateSeq = ProposalsSnapshot . OMap.fromStrictSeq
+fromGovActionStateSeq = ProposalsSnapshot . OMap.fromFoldable
 
 -- | Internal function for checking if the invariants are maintained
 isConsistent_ :: ProposalsSnapshot era -> Bool
--- a/libs/cardano-data/src/Data/OMap/Strict.hs
+++ b/libs/cardano-data/src/Data/OMap/Strict.hs
@@ -1,0 +1,300 @@
+{-# LANGUAGE BangPatterns #-}
+{-# LANGUAGE DerivingStrategies #-}
+{-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE FunctionalDependencies #-}
+{-# LANGUAGE ImportQualifiedPost #-}
+{-# LANGUAGE PatternSynonyms #-}
+{-# LANGUAGE RankNTypes #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE StandaloneDeriving #-}
+{-# LANGUAGE TypeFamilies #-}
+{-# LANGUAGE ViewPatterns #-}
+
+module Data.OMap.Strict (
+  HasOKey (okeyL),
+  OMap (Empty, (:<|:), (:|>:)),
+  null,
+  size,
+  empty,
+  singleton,
+  lookup,
+  member,
+  (!?),
+  fromSet,
+  fromStrictSeq,
+  fromStrictSeqDuplicates,
+  toStrictSeq,
+  toStrictSeqOKeys,
+  toStrictSeqOfPairs,
+  invariantHolds,
+  invariantHolds',
+  (|>),
+  (<|),
+  (|><),
+  (><|),
+  elem,
+  filter,
+)
+where
+
+import Cardano.Ledger.Binary (DecCBOR, EncCBOR (encCBOR), decodeDeduplicate, encodeStrictSeq, encodeTag, setTag)
+import Cardano.Ledger.Binary.Decoding (DecCBOR (decCBOR))
+import Data.Foldable qualified as F
+import Data.Map.Strict qualified as Map
+import Data.Sequence.Strict qualified as SSeq
+import Data.Set qualified as Set
+import Data.Typeable (Typeable)
+import GHC.Exts (IsList (..))
+import Lens.Micro
+import Prelude hiding (elem, filter, lookup, null, seq)
+
+-- | Class of types that can be mapped by a lens or a projection to a
+-- Ord type.
+--
+-- For a type @V@, defines a lens from @V@ to and Ord type @K@.
+class HasOKey k v | v -> k where
+  okeyL :: Lens' v k
+
+-- | A general-purpose finite, insert-ordered, map that is strict in its
+-- keys and values.
+--
+-- The strictness is enforced by the underlying strict `Map` that can
+-- be looked-up by a projection or lens. and the ordering is maintained
+-- by the constructing functions, leveraging `StrictSeq` to hold the
+-- insert-order of the values.
+--
+-- TODO: DecShareCBOR instance
+data OMap k v = OMap
+  { omSSeq :: !(SSeq.StrictSeq v)
+  , _omMap :: !(HasOKey k v => Map.Map k v)
+  }
+
+deriving instance (Show k, Show v, HasOKey k v) => Show (OMap k v)
+
+instance (Eq v, HasOKey k v) => Eq (OMap k v) where
+  (OMap seql _kvl) == (OMap seqr _kvr) = seql == seqr
+
+instance (HasOKey k v, Ord k) => Semigroup (OMap k v) where
+  (<>) = (|><)
+
+instance (HasOKey k v, Ord k) => Monoid (OMap k v) where
+  mempty = empty
+
+instance (HasOKey k v, Ord k) => IsList (OMap k v) where
+  type Item (OMap k v) = v
+  fromList = fromStrictSeq . SSeq.fromList
+  toList = F.toList . omSSeq
+
+instance Foldable (OMap k) where
+  foldMap f = F.foldMap f . SSeq.fromStrict . omSSeq
+  {-# INLINEABLE foldMap #-}
+  foldr f z = F.foldr f z . SSeq.fromStrict . omSSeq
+  {-# INLINEABLE foldr #-}
+  foldl f z = F.foldl f z . SSeq.fromStrict . omSSeq
+  {-# INLINEABLE foldl #-}
+  foldr' f z = F.foldr' f z . SSeq.fromStrict . omSSeq
+  {-# INLINEABLE foldr' #-}
+  foldl' f z = F.foldl' f z . SSeq.fromStrict . omSSeq
+  {-# INLINEABLE foldl' #-}
+  foldr1 f = F.foldr1 f . SSeq.fromStrict . omSSeq
+  {-# INLINEABLE foldr1 #-}
+  foldl1 f = F.foldl1 f . SSeq.fromStrict . omSSeq
+  {-# INLINEABLE foldl1 #-}
+  length = F.length . SSeq.fromStrict . omSSeq
+  {-# INLINE length #-}
+  null = F.null . SSeq.fromStrict . omSSeq
+  {-# INLINE null #-}
+
+instance (Typeable k, EncCBOR v) => EncCBOR (OMap k v) where
+  encCBOR (OMap seq _kv) = encodeTag setTag <> encodeStrictSeq encCBOR seq
+
+instance (Typeable k, HasOKey k v, DecCBOR v, Ord k) => DecCBOR (OMap k v) where
+  decCBOR = decodeDeduplicate decCBOR isMember insert
+    where
+      isMember = member . (^. okeyL)
+      insert v omap = omap |> (v ^. okeyL, v)
+
+-- | \(O(1)\). Shallow invariant using just `length` and `size`.
+invariantHolds :: HasOKey k v => OMap k v -> Bool
+invariantHolds (OMap seq kv) = SSeq.length seq == Map.size kv
+
+-- | \(O(n^2)\). Deep, costly invariant using membership check for each
+-- value. By the pigeon-hole principle, this check is exhaustive.
+invariantHolds' :: (HasOKey k v, Eq v) => OMap k v -> Bool
+invariantHolds' omap@(OMap seq kv) =
+  let vs = Map.elems kv
+   in invariantHolds omap && all (`F.elem` vs) seq
+
+-- | \(O(1)\).
+null :: OMap k v -> Bool
+null (OMap seq _) = SSeq.null seq
+
+-- | \(O(1)\).
+size :: OMap k v -> Int
+size (OMap seq _) = SSeq.length seq
+
+-- | \(O(1)\).
+empty :: OMap k v
+empty = OMap SSeq.Empty Map.empty
+
+-- | \(O(1)\). Strict in its arguments.
+singleton :: k -> v -> OMap k v
+singleton !k !v = OMap (SSeq.singleton v) (Map.singleton k v)
+
+-- | \(O(\log n)\). The element at the specified position, counting from
+-- 0. If the specified position is negative or at least the length of
+-- the sequence, 'lookup' returns 'Nothing'.
+lookup :: (HasOKey k v, Ord k) => k -> OMap k v -> Maybe v
+lookup k (OMap _seq kv) = Map.lookup k kv
+
+-- | `flip`ed version of `lookup`
+(!?) :: (HasOKey k v, Ord k) => OMap k v -> k -> Maybe v
+(!?) = flip lookup
+
+-- | \(O(\log n)\). Checks membership before cons'ing.
+cons :: (HasOKey k v, Ord k) => (k, v) -> OMap k v -> OMap k v
+cons (k, v) omap@(OMap seq kv) =
+  case Map.lookup k kv of
+    Just _ -> omap
+    Nothing -> OMap (v SSeq.<| seq) (Map.insert k v kv)
+
+(<|) :: (HasOKey k v, Ord k) => (k, v) -> OMap k v -> OMap k v
+(<|) = cons
+
+infixr 5 <|
+
+-- | \(O(\log n)\). Checks membership before snoc'ing.
+snoc :: (HasOKey k v, Ord k) => OMap k v -> (k, v) -> OMap k v
+snoc omap@(OMap seq kv) (k, v) =
+  case Map.lookup k kv of
+    Just _ -> omap
+    Nothing -> OMap (seq SSeq.|> v) (Map.insert k v kv)
+
+(|>) :: (HasOKey k v, Ord k) => OMap k v -> (k, v) -> OMap k v
+(|>) = snoc
+
+infixl 5 |>
+
+-- | \(O(\log n)\).
+uncons :: (HasOKey k v, Ord k) => OMap k v -> Maybe ((k, v), OMap k v)
+uncons (OMap seq kv) = case seq of
+  SSeq.Empty -> Nothing
+  v SSeq.:<| vs ->
+    let k = v ^. okeyL
+     in if Map.member k kv
+          then Just ((k, v), OMap vs (Map.delete k kv))
+          else error "Invariant falsified! In OMap, element from sequence not found in corresponding map"
+
+-- | \(O(\log n)\).
+unsnoc :: (HasOKey k v, Ord k) => OMap k v -> Maybe (OMap k v, (k, v))
+unsnoc (OMap seq kv) = case seq of
+  SSeq.Empty -> Nothing
+  vs SSeq.:|> v ->
+    let k = v ^. okeyL
+     in if Map.member k kv
+          then Just (OMap vs (Map.delete k kv), (k, v))
+          else error "Invariant falsified! In OMap, element from sequence not found in corresponding map"
+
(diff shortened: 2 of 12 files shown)


libs/cardano-data/test/Test/Cardano/Data/OMap/Strict/Spec.hs
```diff
@@ -0,0 +1,67 @@
...
+import Data.Sequence.Strict (StrictSeq, fromList)
+import Data.Set (Set, elems)
+import Test.Cardano.Data.Arbitrary ()
+import Test.Hspec
+import Test.Hspec.QuickCheck
+
+spec :: Spec
+spec =
+  describe "OMap.Strict" $ do
+    context "membership check works" $ do
```
aniketd:
```suggestion
    context "membership checks work" $ do
```

eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Snapshots.hs
```diff
@@ -37,19 +42,25 @@ import Data.List (sort)
...
-data ProposalsSnapshot era = ProposalsSnapshot
-  { psGovActionStates :: !(Map (GovActionId (EraCrypto era)) (GovActionState era))
-  , psProposalOrder :: !(StrictSeq (GovActionId (EraCrypto era)))
-  -- ^ Newer actions are near the end
-  }
-  deriving (Generic, Eq, Show)
+-- Ref: https://gitlab.haskell.org/ghc/ghc/-/issues/14046
+instance
+  eracrypto ~ EraCrypto era =>
+  OMap.HasOKey (GovActionId eracrypto) (GovActionState era)
```
lehins:
```suggestion
  c ~ EraCrypto era =>
  OMap.HasOKey (GovActionId c) (GovActionState era)
```

eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Snapshots.hs
```diff
@@ -133,35 +134,24 @@ snapshotRemoveIds ::
...
   ProposalsSnapshot era ->
   (ProposalsSnapshot era, Map.Map (GovActionId (EraCrypto era)) (GovActionState era))
-snapshotRemoveIds gIds (ProposalsSnapshot {..}) = (retainedProposals, removedGovActionStates)
-  where
-    (retainedGovActionStates, removedGovActionStates) = psGovActionStates `extractKeys` gIds
-    retainedProposals =
-      ProposalsSnapshot
-        { psGovActionStates = retainedGovActionStates
-        , psProposalOrder =
-            foldl' (\s x -> if x `Set.member` gIds then s else x :<| s) mempty psProposalOrder
```
aniketd, line 143: I suspect that `x :<| s` here would reverse the `psProposalOrder`

libs/cardano-data/src/Data/OMap/Strict.hs
```diff
@@ -0,0 +1,360 @@
...
+    combine accum v =
+      if f v
+        then accum |> v
+        else accum
+
+-- | \(O(n)\). Given a `Set` of @k@s, and an `OMap` @k@ @v@ return
+-- a pair of `Map` and `OMap` where the @ks@ in the `Set` have been
+-- removed from the `OMap` and presented as a separate `Map`.
+extractKeys :: (HasOKey k v, Ord k) => Set.Set k -> OMap k v -> (Map.Map k v, OMap k v)
+extractKeys ks (OMap seq kv) =
```
lehins: Having two functions with the same name and similar functionality that have opposite return type is dangerous!

```suggestion
extractKeys :: (HasOKey k v, Ord k) => Set.Set k -> OMap k v -> (OMap k v, Map.Map k v)
extractKeys ks (OMap seq kv) =
```

eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Snapshots.hs
lehins: Yeah, I think you are right :smile:

eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Snapshots.hs
```diff
@@ -77,45 +83,32 @@ snapshotInsertGovAction ::
...
-snapshotInsertGovAction gas@GovActionState {gasId} ps@ProposalsSnapshot {..}
-  | Map.member gasId psGovActionStates =
-      ps {psGovActionStates = Map.insert gasId gas psGovActionStates}
-  | otherwise =
-      ProposalsSnapshot
-        { psGovActionStates = Map.insert gasId gas psGovActionStates
-        , psProposalOrder = psProposalOrder :|> gasId
-        }
+snapshotInsertGovAction gas (ProposalsSnapshot omap) =
+  ProposalsSnapshot (omap OMap.:|>: gas)
```
lehins: Note that you changed semantics here. 

Previously it would overwrite the value whenever the element is already present, while `:|>:` simply ignores it.

I don't think it matters for our use case, since we never deal with duplicates, but I prefer the previous semantics

libs/cardano-data/src/Data/OMap/Strict.hs
```diff
@@ -0,0 +1,360 @@
...
+-- | \(O(1)\). Shallow invariant using just `length` and `size`.
+invariantHolds :: HasOKey k v => OMap k v -> Bool
+invariantHolds (OMap seq kv) = SSeq.length seq == Map.size kv
+
+-- | \(O(n^2)\). Deep, costly invariant using membership check for each
+-- value. By the pigeon-hole principle, this check is exhaustive.
+invariantHolds' :: (HasOKey k v, Eq v) => OMap k v -> Bool
+invariantHolds' omap@(OMap seq kv) =
+  let vs = Map.elems kv
+   in invariantHolds omap && all (`F.elem` vs) seq
```
lehins: This is not 100% correct nor is a very efficient invariant checker. With this approach if I were to insert into a `kv` Map a value `v` for which `k /= (v ^. okeyL)`, this checker would not catch it.

This way is better and faster:
```suggestion
  invariantHolds omap && all (\v -> Map.lookup (v ^. okeyL) kv == Just v) seq
```

libs/cardano-data/src/Data/OMap/Strict.hs
```diff
@@ -0,0 +1,360 @@
...
+--
+-- TODO: DecShareCBOR instance
+data OMap k v = OMap
+  { omSSeq :: !(SSeq.StrictSeq v)
+  , omMap :: !(HasOKey k v => Map.Map k v)
+  }
+
+instance (HasOKey k v, ToExpr v) => ToExpr (OMap k v) where
+  listToExpr = listToExpr . F.toList
+  toExpr = listToExpr . F.toList
```
lehins:
```suggestion
  toExpr = toExpr . F.toList
```

libs/cardano-data/src/Data/OMap/Strict.hs
```diff
@@ -0,0 +1,360 @@
...
+
+instance (HasOKey k v, NFData v) => NFData (OMap k v) where
+  rnf = rnf . omSSeq
+
+instance (HasOKey k v, NoThunks v) => NoThunks (OMap k v) where
+  noThunks ctx omap = noThunks ctx $ omSSeq omap
+  wNoThunks ctx omap = wNoThunks ctx $ omSSeq omap
+  showTypeOf = const "OMap"
```
lehins: Why didn't you use generics to derive it? Probably because of existential? That's why I am not big fan of them, usually they just limit functionality.

Current implementation only looks at the SSeq, but we do want to ensure that there are no thunks in the Map too. This class is used for testing, so I suggest removing existential and derive the instance using Generics

libs/cardano-data/src/Data/OMap/Strict.hs
```diff
@@ -0,0 +1,360 @@
...
+--
+-- The strictness is enforced by the underlying strict `Map` that can
+-- be looked-up by a projection or lens. and the ordering is maintained
+-- by the constructing functions, leveraging `StrictSeq` to hold the
+-- insert-order of the values.
+--
+-- TODO: DecShareCBOR instance
+data OMap k v = OMap
+  { omSSeq :: !(SSeq.StrictSeq v)
+  , omMap :: !(HasOKey k v => Map.Map k v)
```
lehins: We don't gain much by using it
```suggestion
  , omMap :: !(Map.Map k v)
```

libs/cardano-data/src/Data/OMap/Strict.hs
```diff
@@ -0,0 +1,360 @@
...
+size :: OMap k v -> Int
+size (OMap seq _) = SSeq.length seq
+
+-- | \(O(1)\).
+empty :: OMap k v
+empty = OMap SSeq.Empty Map.empty
+
+-- | \(O(1)\). Strict in its arguments.
+singleton :: k -> v -> OMap k v
+singleton !k !v = OMap (SSeq.singleton v) (Map.singleton k v)
```
lehins: This violates the invariant! Key should never be supplied separately when inserting into the map
```suggestion
singleton :: HasOKey k v => v -> OMap k v
singleton !v = OMap (SSeq.singleton v) (Map.singleton (v ^. okeyL) v)
```

libs/cardano-data/src/Data/OMap/Strict.hs
```diff
@@ -0,0 +1,360 @@
...
+-- `OMap` invariant is not sustained after the operation.
+adjust :: (HasOKey k v, Ord k) => (v -> v) -> k -> OMap k v -> OMap k v
+adjust f k omap@(OMap seq kv) =
+  case Map.lookup k kv of
+    Nothing -> omap
+    Just v ->
+      let v' = f v
+       in if v' ^. okeyL == k
+            then OMap seq (Map.insert k v' kv)
+            else omap
```
lehins: We have a big problem here.

The value is only adjusted in the Map, not in the SSeq. Which will result in zero votes being counted, since we are using `adjust` in `snapshotAddVote`

I am pretty sure this is the reason why you are getting test failures.

What are our options here:

* If we are to modify the SSeq as well, this will become an `O(n)` operation
* If we change `SSeq` to contain keys only then this operation will be `O(log n)` as expected, but then folding over becomes `O(n * log n)`. 

I think the latter choice is safer.

One more issue with this function is that it modifies the value, which means it can also modify the key inside the value. So this property can be violated due to this function:

```haskell
Just val = Map.lookup k (adjust (& okeyL .~ k') k m)
in val .^ okeyL /= k
```

I don't think there is a graceful solution to this problem, so at the very least this should be documented.

lehins (changes requested):
There is an unfortunate problem with the decision of using a SSeq of `v` instead of `SSeq` of `k`. I commented on the actual location where the problem is. We can discuss tomorrow some possible solutions

libs/cardano-data/src/Data/OMap/Strict.hs
aniketd: > I am pretty sure this is the reason why you are getting test failures.

Alexey! :heart: :bow: what a great catch! I was also suspecting the voting to be wrong, but I could not spot the problem. Thank you so much for spotting it!

libs/cardano-data/src/Data/OMap/Strict.hs
aniketd: :100:

libs/cardano-data/src/Data/OMap/Strict.hs
aniketd: :100: :+1: :bow:

libs/cardano-data/src/Data/OMap/Strict.hs
aniketd: :100: Awesome catch! :raised_hands:

libs/cardano-data/src/Data/OMap/Strict.hs
aniketd: Given the usages for the respective operations is the following (correct me if I'm wrong @TimSheard @lehins )

Assuming `k` for keys (`GovActionId`)  and `v` for values (`GovActionState`)
1. Updates/adjustments to `v` are expected to happen quite often, given that this operation is how voting would be performed.
2. The pulsar needs to have an ordered sequence of `v`s given to its closure, only once per epoch.

Therefore, we need more efficient updates/adjustments and we _can do_ with a less efficient `O(n * log n)` ordered sequence extraction operation. 

If we keep the `OMap k v = { x :: StrictSeq v, y :: Map k v }` we get inefficient adjustments/updates since we have to keep the `v`s updated and maintain the invariant `O(n + log n)`, but efficient ordered sequence extraction for the pulsar `O(1)`.

If we revert it back to the original `OMap k v = { x :: StrictSeq k, y :: Map k v }` we get inefficient ordered sequence extraction for the pulsar `O(n * log n)`, but efficient adjustments/updates since we have to update/adjust `v` only in the underlying `Map` in `O(log n)`, and updates and adjustments are going to be many times in an epoch.

I think you are totally correct @lehins and I would like to revert this.

eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Snapshots.hs
aniketd: Noted! :raised_hands: and fixed! :raised_hands:

libs/cardano-data/src/Data/OMap/Strict.hs
```diff
@@ -0,0 +1,430 @@
...
+
+-- TODO: Add this to cardano-base as soon as possible.
+-- This is currently a very expensive operation, unnecessarily.
+updateStrictSeq :: Int -> a -> SSeq.StrictSeq a -> SSeq.StrictSeq a
+updateStrictSeq i a = SSeq.forceToStrict . Seq.update i a . SSeq.fromStrict
+
+-- TODO: Add this to cardano-base as soon as possible.
+-- This is currently a very expensive operation, unnecessarily.
+deleteAtStrictSeq :: Int -> SSeq.StrictSeq a -> SSeq.StrictSeq a
+deleteAtStrictSeq i = SSeq.forceToStrict . Seq.deleteAt i . SSeq.fromStrict
```
lehins: This can be made simpler and no need to change cardano-base: :wink: 
```suggestion
       in if k' == k
            then OMap seq (Map.insert k v' kv)
            else
            let kv' = Map.insert k' v' $ Map.delete k kv
                (lseq, rseq) = 
                  case SSeq.spanl (/= k) seq of
                    (l, _ :<| r) -> (l, r)
                    _ -> error "Impossible: supplied key expected to be in the sequence"
            in case Map.lookup k' kv of
              Nothing -> OMap (lseq <> k' :<| rseq) kv'
              Just _ -> OMap (lseq <> rseq) kv'
```

libs/cardano-data/src/Data/OMap/Strict.hs
```diff
@@ -0,0 +1,430 @@
...
+--     1. The modified value's `okeyL` is unaltered
+--            - we return omap with the adjusted value,
+--     2. The modified value's `okeyL` is altered, but not a duplicate
+--            - we return the omap with adjusted key (in place) and value
+--     3. The modified value's `okeyL` is altered and is a duplicate
+--            - we return the omap with the old key deleted from the sequence but
+--              without inserting the new key since it is a duplicate, and
+--              deleting old value and inserting the new value in place of its duplicate.
+adjust :: (HasOKey k v, Ord k) => (v -> v) -> k -> OMap k v -> OMap k v
+adjust f k omap@(OMap seq kv) =
```
lehins: I'd suggest going through this module and changing all `seq` bindings to `sseq`, because `seq` is a function from Prelude, which could be misleading for readers of code.
```suggestion
adjust f k omap@(OMap sseq kv) =
```

libs/cardano-data/src/Data/OMap/Strict.hs
```diff
@@ -0,0 +1,430 @@
...
+-- three possible cases:
+--
+--     1. The modified value's `okeyL` is unaltered
+--            - we return omap with the adjusted value,
+--     2. The modified value's `okeyL` is altered, but not a duplicate
+--            - we return the omap with adjusted key (in place) and value
+--     3. The modified value's `okeyL` is altered and is a duplicate
+--            - we return the omap with the old key deleted from the sequence but
+--              without inserting the new key since it is a duplicate, and
+--              deleting old value and inserting the new value in place of its duplicate.
```
lehins, line 294: I suggest going into ghci and cooking up some examples and adding them to haddock in doctest style.

libs/cardano-data/src/Data/OMap/Strict.hs
```diff
@@ -0,0 +1,430 @@
...
+member :: Ord k => k -> OMap k v -> Bool
+member k (OMap _seq kv) = Map.member k kv
+
+-- | \(O(\log n)\). Value membership check, via key membership.
+elem :: (HasOKey k v, Ord k) => v -> OMap k v -> Bool
+elem v = member (v ^. okeyL)
+
+-- | \(O(n)\). Expensive value membership check.
+elem' :: Eq v => v -> OMap k v -> Bool
+elem' v (OMap _seq kv) = F.elem v $ Map.elems kv
```
lehins: Problems with current code:

1. `elem` did not check for equality of the element, so it was no different from a `member` in that respect.
2. `elem'` does not have to be that expensive
3. There is no use for both `elem` and `elem'`

```suggestion
-- | \(O(\log n)\). Value membership check.
elem :: (HasOKey k v, Eq v) => v -> OMap k v -> Bool
elem v = maybe False (== v) . lookup (v ^. okeyL)
```

libs/cardano-data/src/Data/OMap/Strict.hs
```diff
@@ -0,0 +1,430 @@
...
+toStrictSeq :: Ord k => OMap k v -> SSeq.StrictSeq v
+toStrictSeq (OMap seq kv) = fromJust $ traverse (`Map.lookup` kv) seq
+
+-- | \(O(1)\).
+toStrictSeqOKeys :: OMap k v -> SSeq.StrictSeq k
+toStrictSeqOKeys = omSSeq
+
+-- | \(O(n \log n)\). Uses partial `fromJust`.
+toStrictSeqOfPairs :: Ord k => OMap k v -> SSeq.StrictSeq (k, v)
+toStrictSeqOfPairs (OMap seq kv) = fromJust $ traverse (\k -> (k,) <$> Map.lookup k kv) seq
```
lehins:
```suggestion
toStrictSeqOfPairs (OMap seq kv) = (\k -> let !v = kv M.! k in (k, v)) <$> seq
```

libs/cardano-data/src/Data/OMap/Strict.hs
```diff
@@ -0,0 +1,430 @@
...
+fromSet :: (HasOKey k v, Ord k) => Set.Set v -> OMap k v
+fromSet = fromStrictSeq . SSeq.fromList . Set.elems
+
+-- | \(O(1)\).
+toMap :: OMap k v -> Map.Map k v
+toMap = omMap
+
+-- | \(O(n \log n)\). Uses partial `fromJust`.
+toStrictSeq :: Ord k => OMap k v -> SSeq.StrictSeq v
+toStrictSeq (OMap seq kv) = fromJust $ traverse (`Map.lookup` kv) seq
```
lehins:
```suggestion
toStrictSeq (OMap sseq kv) = (kv `Map.!`) <$> sseq
```

libs/cardano-data/src/Data/OMap/Strict.hs
```diff
@@ -0,0 +1,430 @@
...
+-- Starts from the left or head, using `foldl'`
+fromStrictSeq :: (HasOKey k v, Ord k) => SSeq.StrictSeq v -> OMap k v
+fromStrictSeq = F.foldl' snoc_ empty
+  where
+    snoc_ :: (HasOKey k v, Ord k) => OMap k v -> v -> OMap k v
+    snoc_ omap@(OMap seq kv) v =
+      let k = v ^. okeyL
+       in if Map.member k kv
+            then omap
+            else OMap (seq SSeq.|> k) (Map.insert k v kv)
```
lehins: Why do you redefine `snoc` from scratch?
```suggestion
fromStrictSeq = F.foldl' snoc empty
```

libs/cardano-data/src/Data/OMap/Strict.hs
```diff
@@ -0,0 +1,430 @@
...
+  SSeq.Empty -> Nothing
+  ks SSeq.:|> k ->
+    case Map.lookup k kv of
+      Just v -> Just (OMap ks (Map.delete k kv), v)
+      Nothing -> error "Invariant falsified! In OMap, key from sequence not found in corresponding map"
+
+-- | \(O(n \log n)\). Checks membership before snoc'ing.
+-- De-duplicates the StrictSeq without overwriting.
+-- Starts from the left or head, using `foldl'`
+fromStrictSeq :: (HasOKey k v, Ord k) => SSeq.StrictSeq v -> OMap k v
```
lehins: Why not call it `fromFoldable? You are forcing every value in it anyways
```suggestion
fromFoldable :: (Foldable f, HasOKey k v, Ord k) => f v -> OMap k v
```

libs/cardano-data/src/Data/OMap/Strict.hs
```diff
@@ -0,0 +1,430 @@
...
+    snoc_ omap@(OMap seq kv) v =
+      let k = v ^. okeyL
+       in if Map.member k kv
+            then omap
+            else OMap (seq SSeq.|> k) (Map.insert k v kv)
+
+-- | \(O(n \log n)\). Checks membership before snoc'ing.
+-- De-duplicates the StrictSeq and collects and returns the duplicates found.
+-- Starts from the left or head, using `foldl'`
+fromStrictSeqDuplicates :: (HasOKey k v, Ord k, Ord v) => SSeq.StrictSeq v -> (Set.Set v, OMap k v)
```
lehins: Ditto
```suggestion
fromFoldableDuplicates :: (Forldable f, HasOKey k v, Ord k, Ord v) => f v -> (Set.Set v, OMap k v)
```

libs/cardano-data/src/Data/OMap/Strict.hs
```diff
@@ -0,0 +1,430 @@
...
+infixl 5 |>
+
+-- | \(O(\log n)\). Checks membership before snoc'ing. Overwrites a
+-- duplicate.
+snoc' :: (HasOKey k v, Ord k) => OMap k v -> v -> OMap k v
+snoc' (OMap seq kv) v =
+  let k = v ^. okeyL
+   in case Map.lookup k kv of
+        Just _ -> OMap seq (Map.insert k v kv)
+        Nothing -> OMap (seq SSeq.|> k) (Map.insert k v kv)
```
lehins: Could you do the same for `cons`es

Swap `lookup` for `member`
```suggestion
snoc' (OMap seq kv) v
  | Map.member k kv = OMap seq kv'
  | otherwise = OMap (seq SSeq.|> k) kv'
  where 
    k = v ^. okeyL
    kv' = Map.insert k v kv
```

libs/cardano-data/src/Data/OMap/Strict.hs
```diff
@@ -0,0 +1,430 @@
...
+
+infixr 5 <||
+
+-- | \(O(\log n)\). Checks membership before snoc'ing.
+snoc :: (HasOKey k v, Ord k) => OMap k v -> v -> OMap k v
+snoc omap@(OMap seq kv) v =
+  let k = v ^. okeyL
+   in case Map.lookup k kv of
+        Just _ -> omap
+        Nothing -> OMap (seq SSeq.|> k) (Map.insert k v kv)
```
lehins:
```suggestion
snoc omap@(OMap seq kv) v =
  | Map.member k kv = omap
  | otherwise = OMap (seq SSeq.|> k) (Map.insert k v kv)
  where
    k = v ^. okeyL
```

libs/cardano-data/src/Data/OMap/Strict.hs
```diff
@@ -0,0 +1,430 @@
...
+fromStrictSeqDuplicates = F.foldl' snoc_ (Set.empty, empty)
+  where
+    snoc_ :: (HasOKey k v, Ord k, Ord v) => (Set.Set v, OMap k v) -> v -> (Set.Set v, OMap k v)
+    snoc_ (duplicates, omap@(OMap seq kv)) v =
+      let k = v ^. okeyL
+       in if Map.member k kv
+            then (Set.insert v duplicates, omap)
+            else (duplicates, OMap (seq SSeq.|> k) (Map.insert k v kv))
+
+-- | \(O(n + n \log n)\).
```
lehins: O(n + n) === O(n)

```suggestion
-- | \(O(n \log n)\).
```

libs/cardano-data/src/Data/OMap/Strict.hs
```diff
@@ -0,0 +1,430 @@
...
+fromSet :: (HasOKey k v, Ord k) => Set.Set v -> OMap k v
+fromSet = fromStrictSeq . SSeq.fromList . Set.elems
```
lehins: Not sure if this function is no longer useful since we have conversion from Foldable now
```suggestion
fromSet :: (HasOKey k v, Ord k) => Set.Set v -> OMap k v
fromSet = fromFoldable
```

libs/cardano-data/src/Data/OMap/Strict.hs
```diff
@@ -0,0 +1,430 @@
...
+(><|) :: (HasOKey k v, Ord k) => OMap k v -> OMap k v -> OMap k v
+omapl ><| omapr = case omapl of
+  Empty -> omapr
+  ls :|>: l -> ls ><| (l <| omapr)
+
+infixr 5 ><|
+
+instance (HasOKey k v, Ord k) => IsList (OMap k v) where
+  type Item (OMap k v) = v
+  fromList = fromStrictSeq . SSeq.fromList
```
lehins:
```suggestion
  fromList = fromFoldable
```

libs/cardano-data/src/Data/OMap/Strict.hs
```diff
@@ -0,0 +1,430 @@
...
+import GHC.Generics (Generic)
+import Lens.Micro
+import NoThunks.Class (NoThunks (..))
+import Prelude hiding (elem, filter, lookup, null, seq)
+
+-- | Class of types that can be mapped by a lens or a projection to an
+-- Ord type.
+--
+-- For a type @V@, defines a lens from @V@ to and Ord type @K@.
+class HasOKey k v | v -> k where
```
lehins: This will reduce number of constraints you have to write on all functions
```suggestion
class Ord k => HasOKey k v | v -> k where
```

libs/cardano-data/src/Data/OMap/Strict.hs
```diff
@@ -0,0 +1,430 @@
...
+
+-- | \(O(n + n \log n)\).
+fromSet :: (HasOKey k v, Ord k) => Set.Set v -> OMap k v
+fromSet = fromStrictSeq . SSeq.fromList . Set.elems
+
+-- | \(O(1)\).
+toMap :: OMap k v -> Map.Map k v
+toMap = omMap
+
+-- | \(O(n \log n)\). Uses partial `fromJust`.
```
lehins: This sort of stuff does not belong in haddock, since it is not a user facing concern
```suggestion
-- | \(O(n \log n)\).
```

libs/cardano-data/src/Data/OMap/Strict.hs
```diff
@@ -0,0 +1,430 @@
...
+
+instance (Typeable k, EncCBOR v, Ord k) => EncCBOR (OMap k v) where
+  encCBOR omap = encodeTag setTag <> encodeStrictSeq encCBOR (toStrictSeq omap)
+
+instance (Typeable k, HasOKey k v, DecCBOR v, Ord k) => DecCBOR (OMap k v) where
+  decCBOR = decodeSetLikeEnforceNoDuplicates isMember insert decCBOR
+    where
+      isMember = elem
+      insert v omap = omap |> v
+
```
lehins: It is wrong to use `decodeSetLikeEnforceNoDuplicates`, because it allows a Set 258 tag, where Map does not have any tags.

You need to do a similar thing for Map as you did for Set, reorganize a function a bit

```suggestion
  decCBOR = decodeMapLikeEnforceNoDuplicates member snoc decCBOR
```

libs/cardano-data/src/Data/OMap/Strict.hs
```diff
@@ -0,0 +1,430 @@
...
+  foldl' f z = F.foldl' f z . SSeq.fromStrict . toStrictSeq
+  {-# INLINEABLE foldl' #-}
+  foldr1 f = F.foldr1 f . SSeq.fromStrict . toStrictSeq
+  {-# INLINEABLE foldr1 #-}
+  foldl1 f = F.foldl1 f . SSeq.fromStrict . toStrictSeq
+  {-# INLINEABLE foldl1 #-}
+  length = F.length . SSeq.fromStrict . toStrictSeq
+  {-# INLINE length #-}
+  null = F.null . SSeq.fromStrict . toStrictSeq
+  {-# INLINE null #-}
```
lehins: This made sense for `OSet`. For `OMap` it does not make sense to construct an `SSeq` first and fold over that. Folding should happen over the data structure without constructing intermediate ones. 

Think how expensive `null` is! You convert the whole data structure to a strict Seq just to answer if there are any elements.

Try and  figure out yourself on how to implement an efficient fold over the OMap

libs/cardano-data/src/Data/OMap/Strict.hs
```diff
@@ -0,0 +1,430 @@
...
+omapl ><| omapr = case omapl of
+  Empty -> omapr
+  ls :|>: l -> ls ><| (l <| omapr)
+
+infixr 5 ><|
+
+instance (HasOKey k v, Ord k) => IsList (OMap k v) where
+  type Item (OMap k v) = v
+  fromList = fromStrictSeq . SSeq.fromList
+  toList = F.toList . toStrictSeq
```
lehins:
```suggestion
  toList = F.toList
```

libs/cardano-data/testlib/Test/Cardano/Data/Arbitrary.hs
```diff
@@ -1,10 +1,16 @@
...
 import Test.Cardano.Ledger.Binary.Arbitrary ()
 import Test.QuickCheck
 
-instance (Arbitrary a, Ord a) => Arbitrary (OSet a) where
-  arbitrary = fromSet <$> arbitrary
+instance (Arbitrary a, Ord a) => Arbitrary (OSet.OSet a) where
+  arbitrary = OSet.fromSet <$> arbitrary
+
+instance (Ord k, Ord v, Arbitrary v, OMap.HasOKey k v) => Arbitrary (OMap.OMap k v) where
+  arbitrary = OMap.fromSet <$> arbitrary
```
lehins: We care about the equality of keys of the Map, not the actual equality of elements, so we should not be generating a `Set`

```suggestion
instance (Arbitrary k, Arbitrary v, OMap.HasOKey k v) => Arbitrary (OMap.OMap k v) where
  arbitrary = 
    OMap.fromFoldable . shuffle . Map.elems . Map.mapWithKey (flip (set okeyL)) <$> arbitrary
```

libs/cardano-data/test/Test/Cardano/Data/OMap/StrictSpec.hs
```diff
@@ -0,0 +1,182 @@
...
+    context "fromStrictSeqDuplicates preserves order" $ do
+      prop "with duplicates" $
+        \(set :: Set.Set Int) ->
+          let sseq = SSeq.fromList $ Set.elems set
+              omap = fromStrictSeq sseq
+              result = fromStrictSeqDuplicates (sseq SSeq.>< sseq)
+           in do
+                (toStrictSeq . snd) result `shouldBe` sseq
+                result `shouldBe` (set, omap)
+                snd result `shouldSatisfy` invariantHolds'
```
lehins: Why this crazy style of mixing `let in` and `do`? 
Also this is just ugly `(foo . bar) baz`
```suggestion
      prop "with duplicates" $
        \(set :: Set.Set Int) -> do
          let sseq = SSeq.fromList $ Set.elems set
              omap = fromStrictSeq sseq
              result = fromStrictSeqDuplicates (sseq SSeq.>< sseq)
          toStrictSeq (snd result) `shouldBe` sseq
          result `shouldBe` (set, omap)
          snd result `shouldSatisfy` invariantHolds'
```

libs/cardano-data/test/Test/Cardano/Data/OMap/StrictSpec.hs
```diff
@@ -0,0 +1,182 @@
...
+      prop "without duplicates" $
+        \(set :: Set.Set Int) ->
+          let sseq = SSeq.fromList $ Set.elems set
+              omap = fromStrictSeq sseq
+              result = fromStrictSeqDuplicates sseq
+           in do
+                (toStrictSeq . snd) result `shouldBe` sseq
+                result `shouldBe` (Set.empty, omap)
+                snd result `shouldSatisfy` invariantHolds'
```
lehins:
```suggestion
        \(set :: Set.Set Int) -> do
          let sseq = SSeq.fromList $ Set.elems set
              omap = fromStrictSeq sseq
              result = fromStrictSeqDuplicates sseq
          toStrictSeq (snd result) `shouldBe` sseq
          result `shouldBe` (Set.empty, omap)
          snd result `shouldSatisfy` invariantHolds'
```

libs/cardano-data/test/Test/Cardano/Data/OMap/StrictSpec.hs
```diff
@@ -0,0 +1,182 @@
...
+      prop "snoc' - (||>)" $
+        \((omap, i) :: (OMap Int OMapTest, OMapTest)) ->
+          let snoced = omap ||> i
+           in if i `elem` omap
+                then
+                  let adjusted = adjust (const i) (i ^. okeyL) omap
+                   in do
+                        snoced `shouldBe` adjusted
+                else do
+                  snoced `shouldBe` omap |> i
```
lehins:
```suggestion
        \((omap, i) :: (OMap Int OMapTest, OMapTest)) -> do
          let snoced = omap ||> i
          if i `elem` omap
            then snoced `shouldBe` adjust (const i) (i ^. okeyL) omap
            else snoced `shouldBe` omap |> i
```

libs/cardano-data/test/Test/Cardano/Data/OMap/StrictSpec.hs
```diff
@@ -0,0 +1,182 @@
...
+         in do
+              result `shouldSatisfy` (all (`Set.notMember` set) . fst)
+              result `shouldSatisfy` (all (`Set.member` set) . snd)
+              result `shouldSatisfy` invariantHolds' . fst
+    prop "filter" $
+      \((omap, i) :: (OMap Int Int, Int)) ->
+        let result = filter (< i) omap
+         in do
+              result `shouldSatisfy` all (< i)
+              result `shouldSatisfy` invariantHolds'
```
lehins:
```suggestion
      \((omap, i) :: (OMap Int Int, Int)) -> do
        let result = filter (< i) omap
        result `shouldSatisfy` all (< i)
        result `shouldSatisfy` invariantHolds'
```

libs/cardano-data/test/Test/Cardano/Data/OMap/StrictSpec.hs
```diff
@@ -0,0 +1,182 @@
...
+               in do
+                    result `shouldBe` i ><| (i' <| j)
+                    result `shouldSatisfy` invariantHolds'
+    prop "extractKeys should satisfy membership" $
+      \((omap, set) :: (OMap Int Int, Set.Set Int)) ->
+        let result = extractKeys set omap
+         in do
+              result `shouldSatisfy` (all (`Set.notMember` set) . fst)
+              result `shouldSatisfy` (all (`Set.member` set) . snd)
+              result `shouldSatisfy` invariantHolds' . fst
```
lehins:
```suggestion
      \((omap, set) :: (OMap Int Int, Set.Set Int)) -> do
        let (result, extracted) = extractKeys set omap
        result `shouldSatisfy` all (`Set.notMember` set)
        extracted `shouldSatisfy` all (`Set.member` set)
        result `shouldSatisfy` invariantHolds'
```

libs/cardano-data/test/Test/Cardano/Data/OMap/StrictSpec.hs
```diff
@@ -0,0 +1,182 @@
...
+    context "mappend preserves uniqueness" $ do
+      prop "mappending with itself should be a no-op" $
+        \(i :: OMap Int Int) ->
+          let il = i |>< i
+              ir = i ><| i
+           in do
+                il `shouldBe` i
+                ir `shouldBe` i
+                il `shouldSatisfy` invariantHolds'
+                ir `shouldSatisfy` invariantHolds'
```
lehins:
```suggestion
        \(i :: OMap Int Int) -> do
          let il = i |>< i
              ir = i ><| i
          il `shouldBe` i
          ir `shouldBe` i
          il `shouldSatisfy` invariantHolds'
          ir `shouldSatisfy` invariantHolds'
```

libs/cardano-data/test/Test/Cardano/Data/OMap/StrictSpec.hs
```diff
@@ -0,0 +1,182 @@
...
+      prop "mappending with duplicates: left-preserving" $
+        \((i, j) :: (OMap Int Int, OMap Int Int)) -> do
+          case j of
+            Empty -> i `shouldBe` i |>< j
+            j' :<|: _js ->
+              let result = i |>< j
+               in do
+                    result `shouldBe` (i |> j') |>< j
+                    result `shouldSatisfy` invariantHolds'
```
lehins:
```suggestion
            j' :<|: _js -> do
              let result = i |>< j
              result `shouldBe` (i |> j') |>< j
              result `shouldSatisfy` invariantHolds'
```

libs/cardano-data/test/Test/Cardano/Data/OMap/StrictSpec.hs
```diff
@@ -0,0 +1,182 @@
...
+      prop "mappending with duplicates: right-preserving" $
+        \((i, j) :: (OMap Int Int, OMap Int Int)) -> do
+          case i of
+            Empty -> i ><| j `shouldBe` j
+            _is :|>: i' ->
+              let result = i ><| j
+               in do
+                    result `shouldBe` i ><| (i' <| j)
+                    result `shouldSatisfy` invariantHolds'
```
lehins:
```suggestion
              let result = i ><| j
              result `shouldBe` i ><| (i' <| j)
              result `shouldSatisfy` invariantHolds'
```

libs/cardano-data/test/Test/Cardano/Data/OMap/StrictSpec.hs
```diff
@@ -0,0 +1,182 @@
...
+      prop "cons' - (<||)" $
+        \((omap, i) :: (OMap Int OMapTest, OMapTest)) ->
+          let consed = i <|| omap
+           in if i `elem` omap
+                then
+                  let adjusted = adjust (const i) (i ^. okeyL) omap
+                   in do
+                        consed `shouldBe` adjusted
+                else do
+                  consed `shouldBe` i <| omap
```
lehins:
```suggestion
        \((omap, i) :: (OMap Int OMapTest, OMapTest)) -> do
          let consed = i <|| omap
          if i `elem` omap
            then consed `shouldBe` adjust (const i) (i ^. okeyL) omap
            else consed `shouldBe` i <| omap
```

libs/cardano-data/test/Test/Cardano/Data/OMap/StrictSpec.hs
```diff
@@ -0,0 +1,182 @@
...
+                        snoced `shouldBe` adjusted
+                else do
+                  snoced `shouldBe` omap |> i
+    prop "fromStrictSeq preserves order" $
+      \(set :: Set.Set Int) ->
+        let sseq = SSeq.fromList $ Set.elems set
+            omap = fromStrictSeq sseq
+         in do
+              toStrictSeq omap `shouldBe` sseq
+              omap `shouldSatisfy` invariantHolds'
```
lehins:
```suggestion
      \(set :: Set.Set Int) -> do
        let sseq = SSeq.fromList $ Set.elems set
            omap = fromStrictSeq sseq
        toStrictSeq omap `shouldBe` sseq
        omap `shouldSatisfy` invariantHolds'
```

libs/cardano-data/testlib/Test/Cardano/Data/Arbitrary.hs
```diff
@@ -1,10 +1,16 @@
...
-import Data.OSet.Strict
+import Data.OMap.Strict qualified as OMap
+import Data.OSet.Strict qualified as OSet
 import Test.Cardano.Ledger.Binary.Arbitrary ()
 import Test.QuickCheck
 
-instance (Arbitrary a, Ord a) => Arbitrary (OSet a) where
-  arbitrary = fromSet <$> arbitrary
+instance (Arbitrary a, Ord a) => Arbitrary (OSet.OSet a) where
+  arbitrary = OSet.fromSet <$> arbitrary
```
lehins: I just realized that relying directly on the Set will make the values always appear in a sorted order, which is not good for testing.
```suggestion
  arbitrary = OSet.fromSet . shuffle . Set.toList <$> arbitrary
```

eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Snapshots.hs
```diff
@@ -1,7 +1,12 @@
...
-{-# LANGUAGE NamedFieldPuns #-}
+{-# LANGUAGE DerivingStrategies #-}
+{-# LANGUAGE GADTs #-}
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE MultiParamTypeClasses #-}
 {-# LANGUAGE RankNTypes #-}
-{-# LANGUAGE RecordWildCards #-}
+{-# LANGUAGE TypeOperators #-}
+{-# LANGUAGE UndecidableInstances #-}
+{-# OPTIONS_GHC -Wno-orphans #-}
```
lehins: Why is orphans warning being turned off? There is no reason why `HasOKey` instance should not live right next to the type definition
```suggestion
```

lehins (changes requested):
Sorry, I didn't look closely enough last time. A whole bunch more comments.

libs/cardano-data/src/Data/OMap/Strict.hs
aniketd: :metal: :100: :metal: :exploding_head:

libs/cardano-data/src/Data/OMap/Strict.hs
aniketd: :bow: :100: :metal:

libs/cardano-data/src/Data/OMap/Strict.hs
aniketd: Awesome! Applying the same to `toStrictSeq` as well :raised_hands:

libs/cardano-data/src/Data/OMap/Strict.hs
aniketd: I have used the same strict-data trick with a `let` in this, as in `toStrictSeqOfPairs`. I hope that's good.

libs/cardano-data/src/Data/OMap/Strict.hs
aniketd: Thanks for this catch! This is a remnant from a past version where the two were slightly different.

libs/cardano-data/src/Data/OMap/Strict.hs
aniketd: Awesome! Thank you! :raised_hands:

libs/cardano-data/src/Data/OMap/Strict.hs
aniketd: Yes! Done!

libs/cardano-data/src/Data/OMap/Strict.hs
aniketd: Yes! :metal:

libs/cardano-data/src/Data/OMap/Strict.hs
aniketd: Noted!

eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Snapshots.hs
aniketd: Got it!

libs/cardano-data/test/Test/Cardano/Data/OMap/StrictSpec.hs
aniketd: I will fix this for the whole file :+1:

libs/cardano-data/test/Test/Cardano/Data/OMap/StrictSpec.hs
aniketd: Thank you!

libs/cardano-data/src/Data/OMap/Strict.hs
aniketd: Thanks for catching this!

Sorry! This was a leftover from when `toStrictSeq` use to be `O(1)`. I have fixed `null` and `length`. Thought the `fold`s probably still need to be doing the new `toStrictSeq` since that would allow us to fold over the values in the correct order. Thank you! :heart:

libs/cardano-data/src/Data/OMap/Strict.hs
aniketd: :raised_hands: Now it looks like this
```haskell
decodeSetEnforceNoDuplicates = decodeSetLikeEnforceNoDuplicates Set.member Set.insert

decodeListLikeEnforceNoDuplicates isMember insert decodeElement = do
  decodeListLenOrIndef >>= \case
    Just len -> loop (\x -> pure (x - 1, x <= 0)) len mempty
    Nothing -> loop (\x -> (,) x <$> decodeBreakOr) () mempty
  where
    loop :: (t -> Decoder s (t, Bool)) -> t -> f a -> Decoder s (f a)
    loop condition prevStep !acc = [...]

decodeSetLikeEnforceNoDuplicates isMember insert decodeElement =
  allowTag setTag >> decodeListLikeEnforceNoDuplicates isMember insert decodeElement
```

libs/cardano-data/src/Data/OMap/Strict.hs
```diff
@@ -0,0 +1,429 @@
...
+-- be looked-up by a projection or lens. and the ordering is maintained
+-- by the constructing functions, leveraging `StrictSeq` to hold the
+-- insert-order of the keys.
+--
+-- TODO: DecShareCBOR instance
+data OMap k v = OMap
+  { omSSeq :: !(SSeq.StrictSeq k)
+  , omMap :: !(Map.Map k v)
+  }
+  deriving (Generic, Show, Eq)
```
lehins: Let's change the show instance. Current one has too much information that is not relevant for the user
```suggestion
  deriving (Generic, Eq)
  
instance Show (OMap k v) where
  show = show . toStrictSeqOfPairs
```

libs/cardano-data/src/Data/OMap/Strict.hs
```diff
@@ -0,0 +1,429 @@
...
+-- Examples:
+--
+-- >>> import GHC.Exts
+-- >>> import Lens.Micro
+-- >>> import Data.OMap.Strict
+-- >>> data OMapTest = OMapTest {omFst :: Int, omSnd :: Int} deriving (Eq, Show, Ord)
+-- >>> instance HasOKey Int OMapTest where okeyL = lens omFst $ \om u -> om {omFst = u}
+-- >>> let adjustingFn omt@OMapTest {omSnd} = omt {omSnd = omSnd + 1}
+-- >>> let overwritingAdjustingFn omt@OMapTest {omFst} = omt {omFst = omFst + 1} -- Changes the `okeyL`.
+-- >>> let m = fromList $ zipWith OMapTest [1,2] [1,2] :: OMap Int OMapTest
```
lehins: Examples in haddock should be simple
```suggestion
-- >>> import Lens.Micro
-- >>> import Data.OMap.Strict
-- >>> instance HasOKey k (k, v) where okeyL = _1
-- >>> let adjustingFn (k, v) = (k, succ v) -- Changes the value
-- >>> let overwritingAdjustingFn (k,v) = (succ k, v) -- Changes the `okeyL`.
-- >>> let m = fromFoldable $ zip [1,2] ['a','b'] :: OMap Int (Int, Char)
```

libs/cardano-data/src/Data/OMap/Strict.hs
```diff
@@ -0,0 +1,429 @@
...
+  mempty = empty
+
+instance Ord k => Foldable (OMap k) where
+  foldMap f = F.foldMap f . toStrictSeq
+  {-# INLINEABLE foldMap #-}
+  foldr f z = F.foldr f z . toStrictSeq
+  {-# INLINEABLE foldr #-}
+  foldl f z = F.foldl f z . toStrictSeq
+  {-# INLINEABLE foldl #-}
+  foldr' f z = F.foldr' f z . toStrictSeq
```
lehins: You missed part of my comment about the Foldable implementation. It applied to all functions in Foldable instance

Unlike `OSet`, going through StrictSeq is less efficient then implementing it manually, because you are constructing the whole extra `StrictSeq` for no reason other than simplicity of implementation.

This implementation will be faster and less memory intensive:
```suggestion
  foldr' f z OMap {omSSeq,omMap} = F.foldr' (\k -> f (omMap Map.! k)) z omSSeq
```

This is because we no longer need to construct a whole extra `StrictSeq` and we only iterate all of the data once, rather than twice


Also, last but not least, with such implementation it will short circuit operations like `foldr (\x acc -> acc || someExpensiveFunction x) True` and will only iterate potentially some very small part of the OMap.

libs/cardano-ledger-binary/src/Cardano/Ledger/Binary/Decoding/Decoder.hs
```diff
@@ -837,19 +838,18 @@ decodeSetEnforceNoDuplicates ::
...
-decodeSetLikeEnforceNoDuplicates ::
+-- | Decode a collection of values either as a definite or indefinite list. Duplicates are not
+-- allowed.
+decodeListLikeEnforceNoDuplicates ::
   forall s a f.
   Monoid (f a) =>
   -- | Check for membership. Decoder will fail if this predicate returns True
   (a -> f a -> Bool) ->
-  -- | Add an aelement into the decoded Set like data structure
+  -- | Add an aelement into the decoded List like data structure
```
lehins:
```suggestion
  -- | Add an element into the decoded List like data structure
```

lehins (changes requested):
Dammit, I reviewed the PR last night, but forgot to hit Submit Review

Looks much better. Couple of minor points still exist, but they should be easy to fix

libs/cardano-data/src/Data/OMap/Strict.hs
aniketd: Thanks for the awesome suggestions @lehins! I have tried the same for `foldr1` and `foldl1` without success however, since they both use a type signature that does not work for a similar implementation. `foldl1 :: (a -> a -> a) -> t a -> a`. It throws a `rigid type variable` error because in our implementation `k` and `v` are not expected to unify. Should we be skipping those two, so they get expressed automatically in terms of `foldr` and `foldl`?

lehins: Yes, you can provide the minimal implementation and use the default ones for the others.

lehins (approved):
Perfect! :+1:

PR #3792, opened 2023-10-13
RATIFY and GOV constraint tests

### LINKED ISSUE #3784: Constraints and properties for GOV rule
Constraint based testing framework is mature enough to support Conway.

We need to add properties for the GOV rule

MaximilianAlgehed:
# Description

A bunch more work on constraint-based STS testing. This time for running the RATIFY and GOV rules. The GOV rule doesn't work yet because we need to extend the constraint language a bit to handle the map-of-maps in the GOV signal.

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/Preds/CertState.hs
+++ b/libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/Preds/CertState.hs
@@ -50,24 +50,15 @@ manyCoin size = do
 
 -- ======================================
 
--- | Predicates for Voting State, only work in Conway or later Eras.
-vstatePreds :: forall era. Era era => Proof era -> [Pred era]
-vstatePreds p = case whichPParams p of
+vstatePreds :: Era era => Proof era -> [Pred era]
+vstatePreds p = vstateGenPreds p ++ vstateCheckPreds p
+
+vstateGenPreds :: Era era => Proof era -> [Pred era]
+vstateGenPreds p = case whichPParams p of
   PParamsConwayToConway ->
     [ Sized (Range 10 20) currentDRepState
     , Sized (Range 5 7) (Dom committeeState)
-    , ForEach
-        (Range 25 25) -- It is possible we create duplicates, so we make a few more than 20 (the size of drepState)
-        drepStateSet
-        (Pat DRepStateR [Arg expire, Arg anchor, Arg deposit])
-        [ Random (fieldToTerm anchor)
-        , GenFrom (fieldToTerm expire) (Constr "+200To500" (\(EpochNo n) -> EpochNo <$> choose (n + 200, n + 500)) ^$ currentEpoch)
-        , drepDeposit p :=: (fieldToTerm deposit)
-        ]
     , Subset (Dom currentDRepState) voteUniv
-    , Subset (Rng currentDRepState) drepStateSet
-    , SumsTo (Right (Coin 1)) totalDRepDeposit EQL [ProjMap CoinR drepDepositL currentDRepState]
-    , SumsTo (Left (Coin 1)) totalDRepDeposit EQL [SumList (Elems drepDeposits)]
     , Subset (Dom committeeState) voteCredUniv
     , Random numDormantEpochs
     ]
@@ -77,6 +68,23 @@ vstatePreds p = case whichPParams p of
     , Lit EpochR (EpochNo 0) :=: numDormantEpochs
     , Random currentDRepState
     ]
+
+vstateCheckPreds :: forall era. Era era => Proof era -> [Pred era]
+vstateCheckPreds p = case whichPParams p of
+  PParamsConwayToConway ->
+    [ ForEach
+        (Range 25 25) -- It is possible we create duplicates, so we make a few more than 20 (the size of drepState)
+        drepStateSet
+        (Pat DRepStateR [Arg expire, Arg anchor, Arg deposit])
+        [ Random (fieldToTerm anchor)
+        , GenFrom (fieldToTerm expire) (Constr "+200To500" (\(EpochNo n) -> EpochNo <$> choose (n + 200, n + 500)) ^$ currentEpoch)
+        , drepDeposit p :=: (fieldToTerm deposit)
+        ]
+    , Subset (Rng currentDRepState) drepStateSet
+    , SumsTo (Right (Coin 1)) totalDRepDeposit EQL [ProjMap CoinR drepDepositL currentDRepState]
+    , SumsTo (Left (Coin 1)) totalDRepDeposit EQL [SumList (Elems drepDeposits)]
+    ]
+  _ -> []
   where
     drepStateSet = Var $ pV p "drepStateSet" (SetR DRepStateR) No
     deposit = Field @era "deposit" CoinR DRepStateR drepDepositL
--- a/libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/Preds/CertState.hs
+++ b/libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/Preds/CertState.hs
@@ -53,14 +53,25 @@ manyCoin size = do
 vstatePreds :: Era era => Proof era -> [Pred era]
 vstatePreds p = vstateGenPreds p ++ vstateCheckPreds p
 
-vstateGenPreds :: Era era => Proof era -> [Pred era]
+vstateGenPreds :: forall era. Era era => Proof era -> [Pred era]
 vstateGenPreds p = case whichPParams p of
   PParamsConwayToConway ->
-    [ Sized (Range 10 20) currentDRepState
+    [ MetaSize (SzRng 5 15) currentDRepStateSize
+    , Sized currentDRepStateSize currentDRepState
     , Sized (Range 5 7) (Dom committeeState)
     , Subset (Dom currentDRepState) voteUniv
     , Subset (Dom committeeState) voteCredUniv
-    , Random numDormantEpochs
+    , Sized (Range 0 5) numDormantEpochs
+    , ForEach
+        currentDRepStateSize
+        drepStateList
+        (Pat DRepStateR [Arg expire, Arg anchor, Arg deposit])
+        [ Random (fieldToTerm anchor)
+        , GenFrom (fieldToTerm expire) (Constr "+200To500" (\(EpochNo n) -> EpochNo <$> choose (n + 200, n + 500)) ^$ currentEpoch)
+        , drepDeposit p :=: (fieldToTerm deposit)
+        ]
+    , drepStateList :=: (Elems currentDRepState)
+    , SumsTo (Left (Coin 1)) totalDRepDeposit EQL [ProjMap CoinR drepDepositL currentDRepState]
     ]
   _ ->
     [ Sized (Range 0 0) currentDRepState
@@ -68,29 +79,16 @@ vstateGenPreds p = case whichPParams p of
     , Lit EpochR (EpochNo 0) :=: numDormantEpochs
     , Random currentDRepState
     ]
-
-vstateCheckPreds :: forall era. Era era => Proof era -> [Pred era]
-vstateCheckPreds p = case whichPParams p of
-  PParamsConwayToConway ->
-    [ ForEach
-        (Range 25 25) -- It is possible we create duplicates, so we make a few more than 20 (the size of drepState)
-        drepStateSet
-        (Pat DRepStateR [Arg expire, Arg anchor, Arg deposit])
-        [ Random (fieldToTerm anchor)
-        , GenFrom (fieldToTerm expire) (Constr "+200To500" (\(EpochNo n) -> EpochNo <$> choose (n + 200, n + 500)) ^$ currentEpoch)
-        , drepDeposit p :=: (fieldToTerm deposit)
-        ]
-    , Subset (Rng currentDRepState) drepStateSet
-    , SumsTo (Right (Coin 1)) totalDRepDeposit EQL [ProjMap CoinR drepDepositL currentDRepState]
-    , SumsTo (Left (Coin 1)) totalDRepDeposit EQL [SumList (Elems drepDeposits)]
-    ]
-  _ -> []
   where
-    drepStateSet = Var $ pV p "drepStateSet" (SetR DRepStateR) No
+    drepStateList = Var $ pV p "drepStateList" (ListR DRepStateR) No
     deposit = Field @era "deposit" CoinR DRepStateR drepDepositL
-    totalDRepDeposit = Var $ pV p "totalDRepDeposit" CoinR No
     anchor = Field @era "anchor" (MaybeR AnchorR) DRepStateR (drepAnchorL . strictMaybeToMaybeL)
     expire = Field @era "expire" EpochR DRepStateR drepExpiryL
+    totalDRepDeposit = Var $ pV p "totalDRepDeposit" CoinR No
+    currentDRepStateSize = Var $ pV p "currentDRepStateSize" SizeR No
+
+vstateCheckPreds :: Proof era -> [Pred era]
+vstateCheckPreds _p = []
 
 vstateStage ::
   Reflect era =>
--- a/libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/Preds/LedgerState.hs
+++ b/libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/Preds/LedgerState.hs
@@ -34,6 +34,17 @@ import Test.Tasty (TestTree, defaultMain)
 
 -- =========================================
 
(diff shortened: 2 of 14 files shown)


libs/cardano-ledger-test/src/Test/Cardano/Ledger/STS.hs
```diff
@@ -315,6 +323,132 @@ genDelegCert pp ds =
...
+  Gen (GovProcedures era)
+genGovProceedure PropEnv{} GovEnv{} _snapshot =
+  GovProcedures <$> arbitrary <*> arbitrary
+
+------------------------------------------------------------------------
+-- Constraints
+------------------------------------------------------------------------
+
+-- TODO: I'm sure this is some form of Iso thing in la-la-lens land
+toCompF :: Compactible v => Map.Map k v -> Map.Map k (UM.CompactForm v)
```
lehins: You probably wanna see where compaction fails, if it does fail.
```suggestion
toCompF :: (HasCallStack, Compactible v) => Map.Map k v -> Map.Map k (UM.CompactForm v)
```

lehins (approved):
Looks great. Thank you!

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/Preds/CertState.hs
```diff
@@ -47,14 +47,20 @@ manyCoin size = do
 -- ======================================
 
 vstatePreds :: Era era => Proof era -> [Pred era]
-vstatePreds _p =
+vstatePreds p = vstateGenPreds p ++ vstateCheckPreds p
+
+vstateGenPreds :: Era era => Proof era -> [Pred era]
+vstateGenPreds _p =
```
TimSheard: You probably want to break this up, because in pre conway eras some VState things are empty

```
vstatePreds p = case whichPParams p of
  PParamsConwayToConway ->
    [ Sized (Range 10 20) currentDRepState
    , Sized (Range 5 7) (Dom committeeState)
.. the deposit field of the currentDRepState is the drepDeposit
   ]
  _ ->
    [ Sized (Range 0 0) currentDRepState
    , Sized (Range 0 0) committeeState
    ]
```

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/Preds/Tx.hs
```diff
@@ -488,7 +488,7 @@ minusMultiValue p v1 v2 = case whichValue p of
 -- Using constraints to generate a TxBody
 -- ==============================================================
 
-txBodyPreds :: forall era. (HasCallStack, Reflect era) => UnivSize -> Proof era -> [Pred era]
+txBodyPreds :: forall era. (HasCallStack, Show (TxF era), Show (TxBodyF era), Reflect era) => UnivSize -> Proof era -> [Pred era]
```
TimSheard: Just curious, where do the  Show (TxF era), Show (TxBodyF era)  arise?

TimSheard (changes requested):
Why do the PrettyA constraints pop up everywhere? I will help figure out how### to make them go away.

MaximilianAlgehed:
@TimSheard most of the `PrettyA` constraints come from this line:
```haskell
instance (EraPParams era, PrettyA (PParamsUpdate era)) => PrettyA (ProposalProcedure era) where
```

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/Preds/CertState.hs
MaximilianAlgehed: Which of these are generation constraints and which are checking constraints?

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/Preds/Tx.hs
MaximilianAlgehed: They come from these two lines:
```haskell
synopsis (TxBodyR _) x = show x
...
synopsis (TxR _) x = show x
```

MaximilianAlgehed: I wish we could get rid of them but the problem is that they (and `PParamsUpdates`) are era-indexed

TimSheard: Do this


Change these
synopsis (TxBodyR _) x = show x
synopsis (TxR _) x = show x

To
synopsis (TxBodyR p) x = show (pcTxBody p x)
synopsis (TxR p) x = show (pcTx p x)

add these imports
import Test.Cardano.Ledger.Generic.PrettyCore (pcTxBody,pcTx)

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/Preds/CertState.hs
TimSheard: I guess
these two
 [ Sized (Range 0 0) currentDRepState
    , Sized (Range 0 0) committeeState
    ]
are checking, since they need to be empty before Conway Era

There is some checking constrain, implied by this comment
... he deposit field of the currentDRepState is the drepDeposit
But I am not sure how to best express this. It goes something like this. Add up all the deposit fields of the 'currentDRepState' this must be less then the utxosDeposit field in the utxoState. It is more prcise than that, since there are 4 places deposits are stored, and the sum of all 4 must be equal to the variable 'deposits'

TimSheard:
We can fix this
instance (EraPParams era, PrettyA (PParamsUpdate era)) => PrettyA (ProposalProcedure era) where
by adding a name function for pretty printing PParams as is done here in this PR
https://github.com/input-output-hk/cardano-ledger/tree/ts-constrained-trace
I wish it didn't take so long to get PRs merged

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/Preds/Tx.hs
MaximilianAlgehed: I get the following from the compiler if I do this:
```
src/Test/Cardano/Ledger/Constrained/TypeRep.hs:588:43: error:
    ΓÇó Could not deduce: Core.TxBody e ~ TxBodyF e
      from the context: (t ~ TxBodyF e, Era e)
        bound by a pattern with constructor:
                   TxBodyR :: forall era.
                              Era era =>
                              Proof era -> Rep era (TxBodyF era),
                 in an equation for ΓÇÿsynopsisΓÇÖ
        at src/Test/Cardano/Ledger/Constrained/TypeRep.hs:588:11-19
      Expected: Core.TxBody e
        Actual: t
    ΓÇó In the second argument of ΓÇÿpcTxBodyΓÇÖ, namely ΓÇÿxΓÇÖ
      In the first argument of ΓÇÿshowΓÇÖ, namely ΓÇÿ(pcTxBody p x)ΓÇÖ
      In the expression: show (pcTxBody p x)
    ΓÇó Relevant bindings include
        p :: Proof e
          (bound at src/Test/Cardano/Ledger/Constrained/TypeRep.hs:588:19)
        synopsis :: Rep e t -> t -> String
          (bound at src/Test/Cardano/Ledger/Constrained/TypeRep.hs:508:1)
    |
588 | synopsis (TxBodyR p) x = show (pcTxBody p x)
    |                                           ^

src/Test/Cardano/Ledger/Constrained/TypeRep.hs:593:35: error:
    ΓÇó Could not deduce: Core.Tx e ~ TxF e
      from the context: (t ~ TxF e, Era e)
        bound by a pattern with constructor:
                   TxR :: forall era. Era era => Proof era -> Rep era (TxF era),
                 in an equation for ΓÇÿsynopsisΓÇÖ
        at src/Test/Cardano/Ledger/Constrained/TypeRep.hs:593:11-15
      Expected: Core.Tx e
        Actual: t
    ΓÇó In the second argument of ΓÇÿpcTxΓÇÖ, namely ΓÇÿxΓÇÖ
      In the first argument of ΓÇÿshowΓÇÖ, namely ΓÇÿ(pcTx p x)ΓÇÖ
      In the expression: show (pcTx p x)
    ΓÇó Relevant bindings include
        p :: Proof e
          (bound at src/Test/Cardano/Ledger/Constrained/TypeRep.hs:593:15)
        synopsis :: Rep e t -> t -> String
          (bound at src/Test/Cardano/Ledger/Constrained/TypeRep.hs:508:1)
    |
593 | synopsis (TxR p) x = show (pcTx p x)
    |                                   ^
```

MaximilianAlgehed: If I change it to `synopsis (TxBodyR p) x = show (pcTxBody p (unTxBodyF x))` and `synopsis (TxR p) x = show (pcTx p (unTxF x))` I instead get:
```
src/Test/Cardano/Ledger/Constrained/TypeRep.hs:589:32: error:
    ΓÇó Could not deduce (Cardano.Ledger.Pretty.PrettyA
                          (Core.PParamsUpdate e))
        arising from a use of ΓÇÿpcTxBodyΓÇÖ
      from the context: (t ~ TxBodyF e, Era e)
        bound by a pattern with constructor:
                   TxBodyR :: forall era.
                              Era era =>
                              Proof era -> Rep era (TxBodyF era),
                 in an equation for ΓÇÿsynopsisΓÇÖ
        at src/Test/Cardano/Ledger/Constrained/TypeRep.hs:589:11-19
    ΓÇó In the first argument of ΓÇÿshowΓÇÖ, namely
        ΓÇÿ(pcTxBody p (unTxBodyF x))ΓÇÖ
      In the expression: show (pcTxBody p (unTxBodyF x))
      In an equation for ΓÇÿsynopsisΓÇÖ:
          synopsis (TxBodyR p) x = show (pcTxBody p (unTxBodyF x))
    |
589 | synopsis (TxBodyR p) x = show (pcTxBody p (unTxBodyF x))
    |                                ^^^^^^^^

src/Test/Cardano/Ledger/Constrained/TypeRep.hs:594:28: error:
    ΓÇó Could not deduce (Cardano.Ledger.Pretty.PrettyA
                          (Core.PParamsUpdate e))
        arising from a use of ΓÇÿpcTxΓÇÖ
      from the context: (t ~ TxF e, Era e)
        bound by a pattern with constructor:
                   TxR :: forall era. Era era => Proof era -> Rep era (TxF era),
                 in an equation for ΓÇÿsynopsisΓÇÖ
        at src/Test/Cardano/Ledger/Constrained/TypeRep.hs:594:11-15
    ΓÇó In the first argument of ΓÇÿshowΓÇÖ, namely ΓÇÿ(pcTx p (unTxF x))ΓÇÖ
      In the expression: show (pcTx p (unTxF x))
      In an equation for ΓÇÿsynopsisΓÇÖ:
          synopsis (TxR p) x = show (pcTx p (unTxF x))
    |
594 | synopsis (TxR p) x = show (pcTx p (unTxF x))
    |                            ^^^^
Failed, 34 modules loaded.
```

MaximilianAlgehed: Which is less than helpful, we are back in the PParamsUpdate quagmire

MaximilianAlgehed:
> We can fix this instance (EraPParams era, PrettyA (PParamsUpdate era)) => PrettyA (ProposalProcedure era) where by adding a name function for pretty printing PParams as is done here in this PR https://github.com/input-output-hk/cardano-ledger/tree/ts-constrained-trace I wish it didn't take so long to get PRs merged

Can you please provide a pointer to the relevant code?

MaximilianAlgehed:
@TimSheard another solution to this is to add a `type TestableEra era = (Reflect era, PrettyA (Core.PParamsUpdate era))` constraint that we have everywhere. I think this would be fairly non-intrusive (all eras that you actually want to use will satisfy this constraint) and it would let us add things here that we need without having tonnes of ugly constraints floating around the codebase - instead all our testing functions would require `TestableEra era` and that would be it.

lehins:
> type TestableEra era = (Reflect era, PrettyA (Core.PParamsUpdate era))

I'd advise against using constraint type synonyms. Almost every time we used them it lead to problems, since compiler cannot advise you on redundant constraints, leading to over constraining in one place or another.

That being said this is a self contained package that no-one depends on and it is a pretty benign synonym, so I am not gonna block you from using it.

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/Preds/CertState.hs
MaximilianAlgehed: I've done this but I think we'll leave the details of implementing the constraints for the next PR - maybe after we've gotten some more input from the team on the constraints.

MaximilianAlgehed:
@tim

MaximilianAlgehed:
@TimSheard do you have a pointer to the PParamsUpdate fix?

MaximilianAlgehed:
@TimSheard do you mind rebasing this? I'm not up to speed with all the edits so it would be good to have your eye on the rebase so I don't screw something up.

lehins:
@MaximilianAlgehed We prefer rebase instead of commits that merge master into ledger

This results in a nice linear history.

In particular I am talking about this commit: https://github.com/input-output-hk/cardano-ledger/pull/3792/commits/344ca7d157d80647af6a54138cdfeb41a9c68b04

MaximilianAlgehed:
@lehins this will be squashed before merging anyway.

MaximilianAlgehed:
The problem is that rebasing when the history is a mess of rebases causes more problems than solutions.

lehins:
> The problem is that rebasing when the history is a mess of rebases causes more problems than solutions.


I prefer to squash and then rebase. It is easier to resolve conflicts that way. Whatever works for you though. I just wanted to make sure we don't get those kind of merge commits into master, that's all.

MaximilianAlgehed:
> > The problem is that rebasing when the history is a mess of rebases causes more problems than solutions.
> 
> I prefer to squash and then rebase. It is easier to resolve conflicts that way. Whatever works for you though. I just wanted to make sure we don't get those kind of merge commits into master, that's all.

Yes, that's what I usually do too. But here this was rebased on top of another PR before that PR was merged - which causes the history to be a bit messy...

MaximilianAlgehed:
@lehins I think there is a CI bug in 8.10.7 causing the build failure - can you restart it?

lehins:
> I think there is a CI bug in 8.10.7 causing the build failure - can you restart it?

Most certainly. We gotta wait for CI to finish before GithubActions provides the ability to restart failed builds.

>  But here this was rebased on top of another PR before that PR was merged - which causes the history to be a bit messy

You might know about this, but I found this pretty cool, so I'll share it with you. If I understand the problem you mentioned correctly then there is a simple solution in general.

Let's say you have this history:
```
master:
A--B--C

branch-a:
A--D

branch-b (eg. branch in this PR):
A--D--E--F
````

PR for `branch-a` gets rebased and merged into master:
```
master (with branch-a merged in):
A--B--C--D'

branch-b (eg. branch in this PR):
A--D--E--F
```

Then what you do on `branch-b` is `git rebase -i origin/master` and simply drop the `D`  commit and you'll get the expected history with no conflicts that would normally be related to rebasing `D` on commit `D'`:

```
branch-b:
A--B--C--D'--E'--F'
```

**Comment by MaximilianAlgehed** [comment]
The problem is when the other branch has changed _before_ being merged but that change hasn't been merged into your branch...

**Comment by lehins** [comment]
> I think there is a CI bug in 8.10.7 causing the build failure - can you restart it?

@MaximilianAlgehed Looks like tests are not passing. So, there is no point in restarting

**Comment by MaximilianAlgehed** [comment]
@TimSheard any clue what's going on with that test? I don't think this PR should change anything there, but that's probably some rebase artifact.

**Comment by TimSheard** [comment]
Some how the rebase lost a constraint suming the Drep deposits. Fixed it up

File: libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/Preds/LedgerState.hs [hunk 4]
```diff
@@ -80,7 +92,12 @@ ledgerStatePreds usize p =
   , Subset (Rng colUtxo) (colTxoutUniv p)
   , NotMember feeTxIn (Dom colUtxo)
   , NotMember feeTxOut (Rng colUtxo)
-  , SumsTo (Right (Coin 1)) deposits EQL [SumMap stakeDeposits, SumMap poolDeposits, One proposalDeposits, SumMap drepDeposits]
+  , -- , SumsTo (Right (Coin 1)) drepDepositView EQL [ProjMap CoinR drepDepositL currentDRepState]
```
TimSheard: I fixed this commented out code.

TimSheard (commented):
I found one thing and I fixed it.  Ready to go!

TimSheard (approved):

PR #3793, opened 2023-10-13
Fix sums with negative RHS.

TimSheard:
When solving for a variable (like 'N') that appears on the right-hand-side of a summation constraint like this
```
SumsTo (Right (Coin 1)) ₳ 15 <= ∑ sum Map{34 -> ₳ 5 | size = 15, sum = ₳ 125} + sum N
```
There is a problem with the current algorithm, if the type of the variable 'N' does not support negative values.
The current algorithm solves this by subtracting the the constants on the right (125) from the constants on the left (15), resulting in (-110) <= N,  which has a solution N is at Least (-109). But since N cannot be negative we must settle for a solution N is at Least 0.

We add special code to consider this case. And also made a few error messages more precise.

# Checklist

- [x] Self-reviewed the diff


lehins (approved):
:+1:

PR #3794, opened 2023-10-16
Added anchor to resign certs

### LINKED ISSUE #3790: Add Anchor to the Committee resignation certificate
It could be useful to let the community know why a committee member has resigned. One way to do this is by adding Anchor to the resignation certificate, which we should do.

No changes to the ledger state are needed. This task only affects the certificate, consequently also the serialization and CDDL spec.

Soupstraw:
# Description

resolves #3790

# Checklist

- [x] Self-reviewed the diff


eras/conway/impl/src/Cardano/Ledger/Conway/TxCert.hs
```diff
@@ -443,7 +444,8 @@ conwayTxCertDelegDecoder = \case
     pure (3, AuthCommitteeHotKeyTxCert cred key)
   15 -> do
     cred <- decCBOR
-    pure (2, ResignCommitteeColdTxCert cred)
+    a <- decCBOR
```
lehins: For consistency sake with other anchors lets use null encoding instead. Don't forget to update cddl too ;) 
```suggestion
    a <- decodeNullStrictMaybe decCBOR
```

lehins (approved):
Perfect. Thank you!

PR #3797, opened 2023-10-17
Cleanup JSON instances for Conway governance

lehins:
# Description

This is a collection of small JSON related issues. 

Also addition of Assertions for EPOCH rule that verify that withdrawals and treasury fields of EnactState are always zero before and after the EPOCH rule.

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/eras/conway/impl/CHANGELOG.md
+++ b/eras/conway/impl/CHANGELOG.md
@@ -2,6 +2,13 @@
 
 ## 1.10.0.0
 
+* Rename `ensPParams` to `ensCurPParams`.
+* Add `ToJSON` instance for `RatifyState`
+* Change `ToJSON` instance for `ConwayGovState`:
+  * Add `"nextRatifyState"` field
+  * Rename `"ratify"` to `"enactState"`
+  * Rename `"gov"` to `"proposals"`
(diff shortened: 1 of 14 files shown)


eras/conway/impl/CHANGELOG.md
```diff
@@ -2,6 +2,17 @@
 
 ## 1.10.0.0
 
+* Add `psDRepDistrG`.
+* Rename `ensPParams` to `ensCurPParams`.
+* Add `ToJSON` instance for `RatifyState`
+* Change `ToJSON` instance for `ConwayGovState`:
+  * Add `"nextRatifyState"` field
```
teodanciu, line 9: The backtick + double quote looks a little weird, is it on purpose?  (it doesn't matter too much anyway, just curious)

lehins: Those are json fields, so they are represented as strings. Backticks will result in formatting as code, which json is, eg:
image

teodanciu: Makes sense, thanks

teodanciu (approved):
Looks good to me

PR #3801, opened 2023-10-21
Fix epoch rule and tests

lehins:
# Description

A whole collection of bug is fixed in this PR:

## In the EPOCH Rule

* `totalObligation` for the `utxosDeposited` field was computed before proposal deposits were returned
* DRepPulser was started with a state that was not fully updated. Creation of the pulser must be the last step in the EPOCH rule
* DRepPulser was started with an epoch number value that was one into the future, which would result in proposals expiring one epoch too soon and theoretically, although very unlikely, could affect DRep expiry as well.

## In the translation

When NewEpochState was translated from Babbage to Conway, only the pparams in the EnactState were updated, but not the state of the Pulser. Which normally would result in PParams to be reset at the very first Epoch in Conway, however, together with a bug in the EPOCH rule this resulted in alternation of `emptyPParams` with actual `PParams`

## In testing

* PParams where not set properly. Similar problem to the one mentioned above. Only the PParams in the EnactState where updated, but not in the DRepPulser. This bug was present both in the `modifyPParams` function and in the `emptyImpNES`
* ConwayFeatures in cardano-ledger-test had wrong assumptions on the state of proposals.

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/eras/conway/impl/src/Cardano/Ledger/Conway/Governance.hs
+++ b/eras/conway/impl/src/Cardano/Ledger/Conway/Governance.hs
@@ -63,6 +63,7 @@ module Cardano.Ledger.Conway.Governance (
   snapshotIds,
   snapshotRemoveIds,
   snapshotLookupId,
+  snapshotGovActionStates,
   fromGovActionStateSeq,
   isConsistent_,
   -- Lenses
@@ -109,6 +110,8 @@ module Cardano.Ledger.Conway.Governance (
   psDRepDistrG,
   dormantEpoch,
   PulsingSnapshot (..),
+  setCompleteDRepPulsingState,
+  setFreshDRepPulsingState,
   psProposalsL,
   psDRepDistrL,
   psDRepStateL,
@@ -177,6 +180,7 @@ import Cardano.Ledger.Conway.Governance.Snapshots (
   isConsistent_,
   snapshotActions,
   snapshotAddVote,
+  snapshotGovActionStates,
   snapshotIds,
   snapshotInsertGovAction,
   snapshotLookupId,
@@ -212,20 +216,31 @@ import Cardano.Ledger.Shelley.Governance
 import Cardano.Ledger.Shelley.LedgerState (
   EpochState (..),
   NewEpochState (..),
+  certDState,
+  certVState,
+  credMap,
+  dsUnified,
+  epochStateGovStateL,
   epochStateIncrStakeDistrL,
   epochStateRegDrepL,
   epochStateStakeDistrL,
+  epochStateTreasuryL,
   epochStateUMapL,
   esLStateL,
+  lsCertState,
+  lsUTxOState,
   lsUTxOStateL,
   nesEsL,
   utxosGovStateL,
+  utxosStakeDistr,
+  vsCommitteeState,
+  vsDReps,
  )
 import Cardano.Ledger.TreeDiff (Expr (App), ToExpr (..))
 import Cardano.Ledger.UMap
 import qualified Cardano.Ledger.UMap as UMap
 import Control.DeepSeq (NFData (..), deepseq)
-import Control.Monad.Trans.Reader (Reader, runReader)
+import Control.Monad.Trans.Reader (Reader, ReaderT, ask, runReader)
 import Control.State.Transition.Extended
 import Data.Aeson (KeyValue, ToJSON (..), object, pairs, (.=))
 import Data.Default.Class (Default (..))
@@ -235,6 +250,7 @@ import Data.Kind (Type)
 import Data.Map.Strict (Map)
 import qualified Data.Map.Strict as Map
 import Data.Pulse (Pulsable (..), pulse)
+import Data.Ratio ((%))
 import Data.Sequence.Strict (StrictSeq (..))
 import qualified Data.Sequence.Strict as SS
 import Data.Set (Set)
@@ -997,7 +1013,7 @@ class
           Right ratifyState' -> ratifyState'
 
 finishDRepPulser :: forall era. DRepPulsingState era -> (PulsingSnapshot era, RatifyState era)
-finishDRepPulser (DRComplete snap ratstate) = (snap, ratstate)
+finishDRepPulser (DRComplete snap ratifyState) = (snap, ratifyState)
 finishDRepPulser (DRPulsing (DRepPulser {..})) =
   (PulsingSnapshot dpProposals finalDRepDistr dpDRepState, ratifyState')
   where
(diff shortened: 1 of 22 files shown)


libs/cardano-ledger-test/src/Test/Cardano/Ledger/Examples/ConwayFeatures.hs
```diff
@@ -371,15 +371,16 @@ vote pf govActionId =
...
+-- state, and 0 proposals in the previous state (inside the DRepPulser) The one current
+-- proposal is proposed in Epoch 0, and expires after Epoch 2. So the number of proposals
+-- in each Epoch should look like this:
+--
 --    Epoch   current  prev
 --    0       1        0
 --    1       1        1
 --    2       1        1
---    3       0        1
+--    3       1        1
```
lehins, line 383: @TimSheard This is the part that looked wrong to me when we discussed the DRep pulser during one of the meetings a little over a week ago. Sure enough, after fixing the implementation the test failed right at this point. Let me know if you have questions why this is the case, I'll be happy to elaborate.

eras/conway/impl/src/Cardano/Ledger/Conway/Governance.hs
```diff
@@ -1126,6 +1108,71 @@ extractDRepPulsingState :: DRepPulsingState era -> RatifyState era
...
+  globals <- ask
+  let ledgerState = epochState ^. esLStateL
+      utxoState = lsUTxOState ledgerState
+      stakeDistr = credMap $ utxosStakeDistr utxoState
+      certState = lsCertState ledgerState
+      dState = certDState certState
+      vState = certVState certState
+      govState = epochState ^. epochStateGovStateL
+      stakeSize = Map.size stakeDistr
+      -- Maximum number of blocks we are allowed to roll back
```
TimSheard, line 1151: I guess it makes some sense to pass the entire epochstate to this function instead of the individual  pieces needed because we want to be sure that we have made all the changes to the epoch state in the EPOCH rule before we run this.
But we lose all knowledge (in the Rule where this is called) of what the snap shots are.
Just a suggestion, bu maybe define a type Snapshots, and a function getSnapshots :: EpochState -> Snapshots, then use a sequence like
let snap = getSpnaps epochstate
setFreshDRepPulsingState snap




























......

eras/conway/impl/src/Cardano/Ledger/Conway/Translation.hs
```diff
@@ -71,13 +76,18 @@ import Lens.Micro
 type instance TranslationContext (ConwayEra c) = ConwayGenesis c
 
 instance Crypto c => TranslateEra (ConwayEra c) NewEpochState where
-  translateEra ctxt nes =
+  translateEra ctxt nes = do
+    let es = translateEra' ctxt $ nesEs nes
+        -- We need to ensure that we have the same initial EnactState in the pulser as
+        -- well as in the current EnactState, otherwise in the very first EPOCH rule call
+        -- the pulsert will reset it.
```
TimSheard: the pulser will reset it.

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Examples/ConwayFeatures.hs
TimSheard: It is a bit surprising that the initial current proposal in epoch0 , i set to expire after epoch 2, so why is is still in effect in epoch 3?

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Examples/ConwayFeatures.hs
```diff
@@ -539,10 +541,15 @@ testGov pf = do
     (utxo0, _) = utxoFromTestCaseData pf (proposal pf)
     committee = Committee @era (Map.fromList [(ccColdCred @era pf, EpochNo 100)]) (unsafeBoundRational (1 % 2))
     committeeState = CommitteeState (Map.fromList [(ccColdCred pf, Just (ccHotCred pf))])
+    initialEnactState =
+      def
+        & ensCurPParamsL .~ pp
+        & ensCommitteeL .~ SJust committee
+    initialDRepPulsingState = DRComplete def (def & rsEnactStateL .~ initialEnactState)
```
TimSheard, line 548: This was a bit surprising to me. I had to think about this for a long time. I guess that one can explain it this way.
In the transiition to the initital epoch, the pre-initial pulser has to complete, and that needs to use the snap shot of the previous EnactState snapshot.  So perhaps the name 'initialEnactState' is not a good name? perhaps the 'initialEnactSnapshot'.  Of course I might just be confused.  Perhaps some comments.

TimSheard (approved):
The hardest thing to follow is what is the state of the pulser at the beginning of time. Do we start in the epoch rule?  or just after the the epoch rule has run. The state of the pulser will have to be different, if we choose one or the other.

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Examples/ConwayFeatures.hs
lehins: initialEnactState is a pretty good name, because it is NOT a snapshot. It is in fact the state that will be set at the next epoch in Conway. In other words it is the future EnactState. Whenever we running tests or whenever we enter Conway from Babbage, we must ensure EnactState is set to the same value in the DRepPulser (in DRComplete) and in the ConwayGovState, otherwise the value in ConwayGovState will get overwritten at the very next epoch

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Examples/ConwayFeatures.hs
lehins: > why is is still in effect in epoch 3

Because that governance action will be removed only after enactment, which is delayed by the whole epoch. Here is a ticket about it in #3804

lehins:
> The hardest thing to follow is what is the state of the pulser at the beginning of time.

The state of the pulser at the beginning of time is in complete state with everything empty and the EnactState matches the EnactState in the GovState

> Do we start in the epoch rule? or just after the the epoch rule has run. The state of the pulser will have to be different, if we choose one or the other.


Putting initialization of pulser at the beginning of the EPOCH rule will not make it any simpler, since no matter how you set it up, the very first thing we must do is actually extract the result of the previous pulser, because that is what we need to enact in the EPOCH rule. So, you still have the same problem of what is your initial state of the pulser

eras/conway/impl/src/Cardano/Ledger/Conway/Governance.hs
lehins: > But we lose all knowledge (in the Rule where this is called) of what the snap shots are.

Do we need that in the EPOCH rule? This is exactly what I am trying to avoid in this PR. That rule is already huge and complex, we don't need to clutter it with any more logic. This is a specialized function that captures all of this information that is being snapshotted, which a reader can always inspect. I really don't get the argument why we need a whole separate type (that would only make it more confusing with another unrelated `Snapshot` type in Shelley). Especially since it can never be used for anything except with `runConwayRatify`.

I want to avoid scattering this information around different modules. This was one of the confusing parts of previous implementation: in one module pulser is constructed and in another all information needed for the pulser is gathered.

eras/conway/impl/src/Cardano/Ledger/Conway/Translation.hs
lehins: Done

PR #3803, opened 2023-10-23
Fix delegation validation

### LINKED ISSUE #3802: Inconsistency on delegation certificates
It seems there is an inconsistency in how delegation certificates are handled. While for normal `ConwayDelegCert` the target pool has to be registered, with `ConwayRegDelegCert` a delegation to an unregistered pool is allowed. This has caused issues to clients, like db-sync https://github.com/input-output-hk/cardano-db-sync/issues/1540 on sanchonet.

 I think the issue comes from using [validateDelegationRegistered](https://github.com/input-output-hk/cardano-ledger/blob/72c1dd752a1761869e8f4b37aafa59858c191bd1/eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Delegs.hs#L240) which should be extended for Conway.

lehins:
# Description

Fixes #3802 

Problem was in the fact that we added new certificate type that allows for delegation and registration at the same time, which did not participate in validation of StakePool existence. 

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Deleg.hs
+++ b/eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Deleg.hs
@@ -68,7 +68,6 @@ data ConwayDelegPredFailure era
   | StakeKeyNotRegisteredDELEG !(Credential 'Staking (EraCrypto era))
   | StakeKeyHasNonZeroRewardAccountBalanceDELEG !Coin
   | DRepAlreadyRegisteredForStakeKeyDELEG !(Credential 'Staking (EraCrypto era))
-  | WrongCertificateTypeDELEG
   deriving (Show, Eq, Generic)
 
 instance ToExpr (ConwayDelegPredFailure era)
@@ -90,8 +89,6 @@ instance Era era => EncCBOR (ConwayDelegPredFailure era) where
         Sum (StakeKeyHasNonZeroRewardAccountBalanceDELEG @era) 4 !> To mCoin
       DRepAlreadyRegisteredForStakeKeyDELEG stakeCred ->
         Sum (DRepAlreadyRegisteredForStakeKeyDELEG @era) 5 !> To stakeCred
-      WrongCertificateTypeDELEG ->
-        Sum (WrongCertificateTypeDELEG @era) 6
 
 instance Era era => DecCBOR (ConwayDelegPredFailure era) where
   decCBOR = decode $ Summands "ConwayDelegPredFailure" $ \case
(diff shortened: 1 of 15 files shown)


eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Deleg.hs
```diff
@@ -68,7 +68,6 @@ data ConwayDelegPredFailure era
   | StakeKeyNotRegisteredDELEG !(Credential 'Staking (EraCrypto era))
   | StakeKeyHasNonZeroRewardAccountBalanceDELEG !Coin
   | DRepAlreadyRegisteredForStakeKeyDELEG !(Credential 'Staking (EraCrypto era))
-  | WrongCertificateTypeDELEG
```
teodanciu, line 71: Just to make sure I understand: is it the case that the way we validate makes it impossible for this Predicate failure to occur? (But it's still possible in Shelley, since we kept it there, right?)

teodanciu (approved):
Looks good to me, just a question for my own education.

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Deleg.hs
lehins: That is correct. We structured the code in Conway in a smarter way that no longer requires this case. In Shelley it is also an impossible case, so we could do the same for shelley, but that'll be some time later in the future.

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Delegs.hs
```diff
@@ -231,27 +232,27 @@ delegsTransition = do
...
+validateStakePoolDelegateeRegistered ::
+  PState era ->
+  KeyHash 'StakePool (EraCrypto era) ->
   Test (KeyHash 'StakePool (EraCrypto era))
-validateDelegationRegistered certState = \case
-  DelegStakeTxCert _ targetPool ->
-    let stPools = psStakePoolParams $ certPState certState
-     in failureUnless (eval (targetPool ∈ dom stPools)) targetPool
-  _ -> pure ()
+validateStakePoolDelegateeRegistered certState targetPool =
```
aniketd:
```suggestion
validateStakePoolDelegateeRegistered poolState targetPool =
```

aniketd (approved):
Awesome! Thanks! :raised_hands:

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Delegs.hs
lehins: Thanks. Done

PR #3808, opened 2023-10-25
Enhance `CommitteeMembersState` query to return quorum and NoConfidence:

lehins:
# Description
When designing the query we forgot about two important pieces of information relevant for the constitutional commitee, namely the quorum value and the overall presence of the committee. There is a difference between an empty committee and no committe, the latter means that the system is in a state of no confidence.

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/libs/cardano-ledger-api/CHANGELOG.md
+++ b/libs/cardano-ledger-api/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## 1.7.0.0
 
+* Deprecate `GetCommitteeState` query
+* Add `GetCommitteeMembersState` query
 * Export `isPlutusScript`
 * Add `NativeScript`, `getNativeScript` and `validateNativeScript`
 * Removed `phaseScript`
--- a/libs/cardano-ledger-api/test/Test/Cardano/Ledger/Api/State/QuerySpec.hs
+++ b/libs/cardano-ledger-api/test/Test/Cardano/Ledger/Api/State/QuerySpec.hs
@@ -2,6 +2,7 @@
 {-# LANGUAGE DataKinds #-}
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE LambdaCase #-}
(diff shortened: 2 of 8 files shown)


libs/cardano-ledger-api/test/Test/Cardano/Ledger/Api/State/QuerySpec.hs
```diff
@@ -47,175 +47,203 @@ spec = do
...
-                propComplete nes
-                propAuthorized nes
-                propActiveAuthorized nes
-                propNotAuthorized nes
-                propResigned nes
-                propUnrecognized nes
-                propFilters ckFilter hkFilter statusFilter nes
+  prop "CommitteeMembersState Query" $ \nes' ckFilter' hkFilter' statusFilter -> do
+    -- replace some arbitrary number of cold keys from the committeeState with the
+    -- ones from the committee so we can have Active members
```
teodanciu: Maybe these comments make more sense a  few lines below, before[ this line ](https://github.com/input-output-hk/cardano-ledger/pull/3808/commits/926a9ec793d3e2b2d80d95b1e29dc23e2fe8689c#diff-fb722545a2358df077b89f77d538341ddfbb9d6a3e68354459abad096b1bfbc7R71) that does the replacing?

lehins: Done

libs/cardano-ledger-api/CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## 1.7.0.0
 
+* Deprecate `GetCommitteeState` query
+* Add `GetCommitteeMembersState` query
```
teodanciu, line 6: i had completely forgotten about updating the changelog for this, uf.  Thanks

teodanciu (approved):
Looks good to me

PR #3809, opened 2023-10-25
CDDL roundtrip testing

lehins:
# Description

* Move cddl files to the era packages
* Switch to using hspec for testing cddl
* Add roundtrip testing to the cddl tests. Meaning that the value is now checked on the Haskell side that reserialization produces the same value. We already do this kind of test for all of the types, but this one uses CDDL generators, rather QuickCheck's arbitrary, which is just a nice litlle extra. 
* Move CDDL tests for TPraos header to `cardano-protocol-tpraos`, rather than run it in shelley.
* Important part is that all of the same functionality is still being tested. No tests were removed. But, no tests have been added for any types that haven't been tested before. We'll need to do that in a separate PR, since current CDDL tests are not exhaustive.

Noticeable speed up in the CDDL testing. This is new:

```shell
$ cabal build cardano-protocol-tpraos cardano-ledger-shelley cardano-ledger-allegra cardano-ledger-mary cardano-ledger-alonzo cardano-ledger-babbage cardano-ledger-conway
...
$ time cabal test cardano-protocol-tpraos cardano-ledger-shelley cardano-ledger-allegra cardano-ledger-mary cardano-ledger-alonzo cardano-ledger-babbage cardano-ledger-conway --test-options='-m CDDL' -j1
...
real	2m31.690s
user	2m22.752s
sys	0m10.389s
```

vs old:
```shell
$ cabal build cardano-ledger-shelley-test cardano-ledger-shelley-ma-test cardano-ledger-alonzo-test cardano-ledger-babbage-test cardano-ledger-conway-test
...
$ time cabal test cardano-ledger-shelley-test cardano-ledger-shelley-ma-test cardano-ledger-alonzo-test cardano-ledger-babbage-test cardano-ledger-conway-test --test-options='-p CDDL' -j1
...
real	3m50.625s
user	3m37.120s
sys	0m15.423s
```

Which means we can add some more CDDL tests.

# Checklist

- [ ] Self-reviewed the diff


eras/alonzo/impl/cardano-ledger-alonzo.cabal
```diff
@@ -102,9 +107,11 @@ library
 library testlib
     exposed-modules:
         Test.Cardano.Ledger.Alonzo.Arbitrary
+        Test.Cardano.Ledger.Alonzo.Binary.Cddl
         Test.Cardano.Ledger.Alonzo.Binary.RoundTrip
         Test.Cardano.Ledger.Alonzo.CostModel
         Test.Cardano.Ledger.Alonzo.ImpTest
+        Paths_cardano_ledger_alonzo
```
nc6: Do we want to expose the Paths module directly? I would have thought that exposing the accessors you've written would be sufficient.

eras/alonzo/impl/test/Main.hs
```diff
@@ -14,4 +15,5 @@ main =
   ledgerTestMain $
```
nc6, line 15: Not directly related to this PR, but I find the sheer number of places where tests can be run/written extremely confusing. We have tests inside the Alonzo impl, and we also have tests inside the `cardano-ledger-test` package. At the least, the CONTRIBUTING.md file would benefit from a guide explaining which tests should go where and why.

flake.nix
```diff
@@ -146,22 +146,26 @@
...
               lib.mkIf pkgs.stdenv.hostPlatform.isWindows {
                 packages.set-algebra.components.tests.tests.buildable = lib.mkForce false;
                 packages.plutus-preprocessor.package.buildable = lib.mkForce false;
                 packages.cardano-ledger-test.package.buildable = lib.mkForce false;
-                packages.cardano-ledger-shelley-ma-test.package.buildable = lib.mkForce false;
-                packages.cardano-ledger-shelley-test.package.buildable = lib.mkForce false;
-                packages.cardano-ledger-alonzo-test.package.buildable = lib.mkForce false;
-                packages.cardano-ledger-babbage-test.package.buildable = lib.mkForce false;
-                packages.cardano-ledger-conway-test.package.buildable = lib.mkForce false;
+                packages.cardano-ledger-shelley.package.buildable = lib.mkForce false;
```
nc6, line 162: This is perhaps a bigger change that should be highlighted, although I can't imagine anyone is actually using the ability to build using nix on Windows

libs/cardano-ledger-binary/testlib/Test/Cardano/Ledger/Binary/Cddl.hs
```diff
@@ -0,0 +1,238 @@
...
+  , cddlVarDiagCbor :: ![DiagCbor]
+  -- ^ Generated CBOR data from the above CDDL spec
+  }
+
+data CddlData = CddlData
+  { cddlData :: !Cddl
+  , cddlNumExamples :: !Int
+  }
+
+beforeAllCddlFile :: HasCallStack => Int -> IO [BSL.ByteString] -> SpecWith CddlData -> Spec
```
nc6: Haddock? It's not obvious from the name or types what this function should do :-)

nc6: Although now I see that the `beforeAll` comes from another function name, so perhaps I just don't know it. But then it seems plausible that others might also not know it.

nc6:
```suggestion
-- | Runs the given specs given CDDL files already loaded
beforeAllCddlFile :: HasCallStack => Int -> IO [BSL.ByteString] -> SpecWith CddlData -> Spec
```
Or something like that

libs/cardano-ledger-binary/testlib/Test/Cardano/Ledger/Binary/Cddl.hs
```diff
@@ -0,0 +1,238 @@
...
+  Version ->
+  T.Text ->
+  SpecWith CddlData
+cddlRoundTripCborSpec version varName =
+  let lbl = label $ Proxy @a
+   in it (T.unpack $ varName <> ": " <> lbl) $ \cddlData ->
+        withCddlVarFile varName cddlData $
+          cddlRoundTripExpectation lbl version version (cborTrip @a)
+
+cddlRoundTripExpectation ::
```
nc6, line 148: I would appreciate a little more documentation on these

libs/cardano-ledger-binary/testlib/Test/Cardano/Ledger/Binary/Cddl.hs
```diff
@@ -0,0 +1,238 @@
...
+  -- | File path to the file with CDDL spec
+  FilePath ->
+  -- | CBOR encoded data that should conform with CDDL spec
+  BSL8.ByteString ->
+  IO (Either String BSL.ByteString)
+validateCddlConformance filePath =
+  readProcessNoFailure "validating cddl conformace" (proc "cddl" [filePath, "validate", "-"])
+
+-- | Run a process with `readProcess` and explode on failure. Accepts lazy bytestring as
+-- input for the spawned process. In case of a successfull process exit, but nonempty
```
nc6:
```suggestion
-- input for the spawned process. In case of a successful process exit, but nonempty
```

libs/cardano-ledger-binary/testlib/Test/Cardano/Ledger/Binary/Cddl.hs
```diff
@@ -0,0 +1,238 @@
...
+-- stderr, it will be printed out.
```
nc6: Printed to where? And the bytestring in the successful return is what?

nc6: In fact, this looks wrong, we are throwing away the stderr in the success case.

nc6 (changes requested):
One issue with the documentation of `readProcessNoFailure`, otherwise everything is minor and may be addresssed/dismissed as desired.

libs/cardano-ledger-binary/testlib/Test/Cardano/Ledger/Binary/Cddl.hs
lehins: You need to be familiar with `hspec` to know it.
It is actually the reason, why this approach is faster, because it prepares the state for all of the tests in its scope only once. As far as I know `tasty` simply does not have capability like that.

I'll add some haddock to `beforeAllCddlFile`

eras/alonzo/impl/cardano-ledger-alonzo.cabal
lehins: Oh yeah, it should be the `other-modules`. Good catch. I'll fix all of them

libs/cardano-ledger-binary/testlib/Test/Cardano/Ledger/Binary/Cddl.hs
lehins: > we are throwing away the stderr in the success case.

This is an unfortunate case, which is described in the comment. Using a cddl file results in a tons of warnings, since we are using only a small portion of the cddl. 

We might wanna find a solution where we only use a portion of the cddl that useful for a particular test case, but that is not possible without parsing the cddl file (or generating one from scratch)

So, for that reason this function does exactly the same as it did before. I just rewrote it using the `typed-process`, rather than inferior and less maintained `process-extras`


That being said, the haddock is indeed wrong, because it reflects my initial intention, rather than current behavior. I'll fix it.

flake.nix
lehins: I am not really sure what this is for, to completely honest with you. But I do know that we don't run any tests on Windows. Moreover it would be wasteful to have Windows CI, since we don't have any OS specific logic that needs testing. As you know, all of  the ledger code is completely pure, so we don't need to worry about this. Ledger and cli code is tested on Windows and Darwin, which is where it really matters.

eras/alonzo/impl/test/Main.hs
lehins: 100% agree. When I have some bandwidth I'll fixup the `CONTRIBUTING.md

For, now we'll have to get by with what's there. Documentation is the least of my worries at the moment

lehins: Turned it into an issue

libs/cardano-ledger-binary/testlib/Test/Cardano/Ledger/Binary/Cddl.hs
lehins: Done

libs/cardano-ledger-binary/testlib/Test/Cardano/Ledger/Binary/Cddl.hs
lehins: Done

libs/cardano-ledger-binary/testlib/Test/Cardano/Ledger/Binary/Cddl.hs
lehins: Done

libs/cardano-ledger-binary/testlib/Test/Cardano/Ledger/Binary/Cddl.hs
lehins: Improved haddock

lehins: Done

libs/cardano-ledger-binary/testlib/Test/Cardano/Ledger/Binary/Cddl.hs
```diff
@@ -0,0 +1,306 @@
...
+  byteStringInput,
+  proc,
+  readProcess,
+  setStdin,
+ )
+import Test.Cardano.Ledger.Binary.RoundTrip
+import Test.Hspec
+import UnliftIO.Temporary (withTempFile)
+
+-- | Contents of CDDL SPec
```
aniketd:
```suggestion
-- | Contents of CDDL Spec
```

libs/cardano-ledger-binary/testlib/Test/Cardano/Ledger/Binary/Cddl.hs
```diff
@@ -0,0 +1,306 @@
...
+  HasCallStack =>
+  -- | Number of random cases to generate
+  Int ->
+  -- | Action that produces a list of valid CDDL specs
+  IO [BSL.ByteString] ->
+  SpecWith CddlData ->
+  Spec
+beforeAllCddlFile numExamples getCddlFiles = beforeAll $ do
+  cddls <- getCddlFiles
+  -- combine all files into one large strict bytestring, while converting in back to the
```
aniketd:
```suggestion
  -- combine all files into one large strict bytestring, while converting it back to the
```

libs/cardano-ledger-binary/testlib/Test/Cardano/Ledger/Binary/Cddl.hs
```diff
@@ -0,0 +1,306 @@
...
+          cddlRoundTripExpectation lbl version version (cborTrip @a)
+
+-- | Verify that random data generated is:
+--
+-- * Decoded successfully into a Haskell type using the decoder in `Trip` and the version
+--   supplied
+--
+-- * When reencoded conforms to the CDDL spec and produces valid `FlatTerm`
+--
+-- * When decoded again from the bytes producedc by the enncoder matches the type exactly
```
aniketd:
```suggestion
-- * When decoded again from the bytes produced by the encoder matches the type exactly
```

libs/cardano-ledger-binary/testlib/Test/Cardano/Ledger/Binary/Cddl.hs
```diff
@@ -0,0 +1,306 @@
...
+                  Right _bsl -> val' `shouldBe` val
+              Left embedErr -> cddlFailure diagCbor embedErr
+
+genCddlDiagCbor :: HasCallStack => Int -> Cddl -> IO [DiagCbor]
+genCddlDiagCbor numCases =
+  fmap (either error (map DiagCbor . BSL8.lines))
+    . readProcessNoFailure "generating examples" (proc "cddl" ["-", "generate", show numCases])
+    . unCddl
+
+-- | Convert randomly genertaed Cbor in special Diag (human readble) format to bianry CBOR
```
aniketd:
```suggestion
-- | Convert randomly generated Cbor in special Diag (human readable) format to binary CBOR
```

aniketd (approved):
Awesome! :raised_hands:

PR #3825, opened 2023-10-27
Prevent voting on expired GovActions

### LINKED ISSUE #3804: Prevent voting on expired governance actions
One of the consequences of delaying enactment by one epoch is that we do not remove expired governance actions until they are either enacted or marked as expired one full epoch after their actual expiry value

in other words such governance actions are retained in the governance state for one full extra epoch past expiry, despite that any votes that are placed during that epoch will never have any affect, because we know that such governance action will be removed at the end of the epoch, either due to being enacted or expired.

Here is a concrete example:

1. An action `g` is proposed at epoch `n`, while `ppGovActionLifetime = 2`
2. Expire for that governance action is set to `n + 2`, which means at epoch `n + 3` it will be considered as expired
3. At epoch `n + 3` we snapshot the whole governance, which still contains gov action `g`, because during the previous epoch it was still being voted on.
4. During the epoch `n + 3` we do all the necessary pulsing perform ratification, at which point we know if governance action `g` is simply removed as expire or is it enacted
5. A the epoch boundary `n + 4` we can remove this governance action `g` from the state

Problem with this whole scenario is that during epoch `n + 3` governance action `g` is known to be expired and we should not allow it to be voted on. There are two ways this could be solved:

1. ~Remove the action `g` from the state after it is snapshotted by the pulser~
2. Add a predicate failure that prevents votes on expired actions.

We decided to go with the latter solution

aniketd:
# Description

Closes #3804 

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/eras/conway/impl/test/Test/Cardano/Ledger/Conway/GovSpec.hs
+++ b/eras/conway/impl/test/Test/Cardano/Ledger/Conway/GovSpec.hs
@@ -6,7 +6,10 @@ module Test.Cardano.Ledger.Conway.GovSpec where
 import Cardano.Ledger.BaseTypes
 import Cardano.Ledger.Conway
 import Cardano.Ledger.Conway.Governance
+import Cardano.Ledger.Conway.PParams (ppGovActionLifetimeL)
+import Cardano.Ledger.Credential (Credential (KeyHashObj))
 import Cardano.Ledger.Shelley.LedgerState
+import Control.Monad (void)
 import Data.Default.Class (Default (..))
 import Data.Maybe
 import Lens.Micro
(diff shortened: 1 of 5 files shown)


eras/conway/impl/test/Test/Cardano/Ledger/Conway/GovSpec.hs
```diff
@@ -6,15 +6,56 @@ module Test.Cardano.Ledger.Conway.GovSpec where
...
+                      (fromJust $ textToUrl "constitution.0")
+                      constitutionHash
+                  )
+                  SNothing
+              )
+      gaidConstitutionProp <- submitProposal constitutionAction
+      passEpoch
+      passEpoch
+      passEpoch
+      void $ voteForProposalFail (DRepVoter $ KeyHashObj khDRep) gaidConstitutionProp
```
lehins: Could you figure out a way to also specify an expected predicate failure?

You'll probably need to improve or create an alternative version o `submitFailingTx`

lehins (approved):
Perfect!

eras/conway/impl/test/Test/Cardano/Ledger/Conway/GovSpec.hs
aniketd: :100: :raised_hands:

PR #3840, opened 2023-11-01
Fix anomalies in Deposits in the Conway Era

### LINKED ISSUE #3834: Fix anomalies in Deposits in the Conway Era.
Code for handling deposits in the Conway Era has at least 3 oversights

1) getTotalDepositsTxBody ignores DRep deposits
2) obligationCertState also ignores DRep deposits
3) The Assertion violations 
     a)  "Deposit pot must equal obligation (PoolReap)"
     b)  "Deposit pot must equal obligation (Ledger)"
     Do not even tell what the pot and obligation is, so its is very hard to diagnose what is wrong
4) obligationCertState, instead of returning a single Coin should return a Triple
        CertObligation { drep :: Coin, pool::Coin, stake:: Coin}
5)  totalObligation, instead of returning a single Coin should return a Quadruple

TimSheard:
Fixes #3834 
# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/libs/cardano-ledger-core/src/Cardano/Ledger/CertState.hs
+++ b/libs/cardano-ledger-core/src/Cardano/Ledger/CertState.hs
@@ -31,6 +31,8 @@ module Cardano.Ledger.CertState (
   payPoolDeposit,
   refundPoolDeposit,
   obligationCertState,
+  Obligations (..),
+  sumObligation,
   -- Lenses
   certDStateL,
   certPStateL,
@@ -435,7 +437,7 @@ ptrsMap :: DState era -> Map Ptr (Credential 'Staking (EraCrypto era))
 ptrsMap (DState {dsUnified = UMap _ ptrmap}) = ptrmap
 
 -- ==========================================================
--- Functions that handle Deposits for pool key hashes.
+-- Functions that handle Deposits
 
 -- | One only pays a deposit on the initial pool registration. So return the
 --   the Deposits unchanged if the keyhash already exists. There are legal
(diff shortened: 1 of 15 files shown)


libs/cardano-ledger-core/src/Cardano/Ledger/CertState.hs
```diff
@@ -461,12 +464,58 @@ refundPoolDeposit keyhash pstate = (coin, pstate {psDeposits = newpool})
       Just c -> (c, Map.delete keyhash pool)
       Nothing -> (mempty, pool)
 
+refundDRepDeposit :: Credential 'DRepRole (EraCrypto era) -> VState era -> (Coin, VState era)
+refundDRepDeposit cred vstate = (coin, newVState)
+  where
+    (coin, newVState) = case Map.lookup cred (vsDReps vstate) of
+      Just drepState -> (drepDeposit drepState, vstate {vsDReps = Map.delete cred (vsDReps vstate)})
+      Nothing -> (mempty, vstate)
```
lehins: This function is not used anywhere

```suggestion
```

libs/cardano-ledger-core/src/Cardano/Ledger/CertState.hs
```diff
@@ -435,7 +438,7 @@ ptrsMap :: DState era -> Map Ptr (Credential 'Staking (EraCrypto era))
 ptrsMap (DState {dsUnified = UMap _ ptrmap}) = ptrmap
 
 -- ==========================================================
--- Functions that handle Deposits for pool key hashes.
+-- Functions that handling Deposits
```
lehins: Is "handling" in a correct form for this sentence?
```suggestion
-- Functions that handle Deposits
```

lehins (approved):
Looks great. Thank you

teodanciu (approved):
Interesting, so the reason the assertion wasn't failing is because the value happened to be missing from both sides of the equation I suppose.  
Looks good to me.

libs/cardano-ledger-core/src/Cardano/Ledger/CertState.hs
TimSheard: done

libs/cardano-ledger-core/src/Cardano/Ledger/CertState.hs
TimSheard: done

PR #3843, opened 2023-11-02
Add ...WithLogs versions of evalScripts and friends

rjmh:
# Description

Add `evalScriptsWithLogs`, `evalPlutusScriptsWithLogs`, `runPlutusScriptWithLogs`

`evalScripts` runs Plutus scripts, which may generate log output, but throws the log output away. This makes it impossible to
measure coverage of the Plutus code, when `evalScripts` is used in tests (e.g. in `sc-tools`), because coverage data is recorded in the log. This PR adds `...WithLogs` versions of the three functions concerned, which behave identically but return the logs as part of the result, so that tests implemented using them can collect coverage information.

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/PlutusScriptApi.hs
+++ b/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/PlutusScriptApi.hs
@@ -17,7 +17,9 @@ module Cardano.Ledger.Alonzo.PlutusScriptApi (
   getSpendingTxIn,
   getDatumAlonzo,
   evalScripts,
+  evalScriptsWithLogs,
   evalPlutusScripts,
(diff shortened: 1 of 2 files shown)


lehins (approved):
This looks great. Thank you!
Could you add these three functions `evalScriptsWithLogs`, `evalPlutusScriptsWithLogs` and `runPlutusScriptWithLogs` to this section of the CHANGELOG: https://github.com/input-output-hk/cardano-ledger/blob/0533bfb1f7b6e76d4a81c7087fea6bb1892f6adb/eras/alonzo/impl/CHANGELOG.md?plain=1#L4

teodanciu (approved):
Looks good to me. 
Thank you for the clear explanation of the context for the change :bow:

Issue #3811, opened 2023-10-25
Implement Conformance tests.


lehins:
The initial attempt has been done in #3545 

Related ticket on the spec side: https://github.com/input-output-hk/formal-ledger-specifications/issues/264

The high level idea behind the conformance tests is:

- [x]  0. Create a package that will contain all of the conformance testing in the `/libs/cardano-ledger-conformance`: 
       #3839
- [x]  1. Pull generated Haskell code from Agda spec from nix:
       #3839
- [x]  2. Implement translation of ledger types `to`/`from` the Haskell version of the spec (only the ledger state needs to have both the `to` and `from` conversions, signal and environment only need the `to` translation:
       #3844
- [x]  3. Use constraint based framework from `cardano-ledger-test` to generate valid data:
       #3852
- [x]  4. Compare the results produced are the same. 
- [x]  5. Set up CI that runs these tests nightly (no need to run it on every PR, unless it will be fast enough)

Each of the above tasks should/could be turned into its own ticket(s) and handled as separate PRs

lehins:
This task is done. All of the higher level objectives were achieved. Further changes to the conformance tests will be improvements to the amount of testing that testing is achieved

PR #3844, opened 2023-11-02
Add mappings to Agda types

Soupstraw:
# Description

Adds the `SpecCastable` type class that maps the implementation types to spec types.

Resolves task 2 of #3811 

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/libs/cardano-ledger-conformance/src/Test/Cardano/Ledger/Conformance.hs
+++ b/libs/cardano-ledger-conformance/src/Test/Cardano/Ledger/Conformance.hs
@@ -1,10 +1,420 @@
-module Test.Cardano.Ledger.Conformance (
+{-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE FlexibleInstances #-}
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE PolyKinds #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE StandaloneDeriving #-}
+{-# LANGUAGE TypeApplications #-}
+{-# LANGUAGE TypeFamilyDependencies #-}
+{-# LANGUAGE TypeOperators #-}
+{-# LANGUAGE UndecidableInstances #-}
 
+module Test.Cardano.Ledger.Conformance (
+  SpecTranslate (..),
 ) where
 
-import Lib (TxBody)
+import Cardano.Crypto.DSIGN (DSIGNAlgorithm (..), SignedDSIGN (..))
+import Cardano.Crypto.Hash (Hash, hashToBytes)
+import Cardano.Ledger.Address (Addr (..), serialiseAddr)
+import Cardano.Ledger.Allegra.Scripts (ValidityInterval (..))
+import Cardano.Ledger.Alonzo (AlonzoScript, AlonzoTxAuxData)
+import Cardano.Ledger.Alonzo.PParams (OrdExUnits (OrdExUnits))
+import Cardano.Ledger.Alonzo.Scripts (CostModels, ExUnits (..), Prices, Tag (..))
+import Cardano.Ledger.Alonzo.Scripts.Data (Data, Datum (..))
+import Cardano.Ledger.Alonzo.Tx (AlonzoTx (..))
+import Cardano.Ledger.Alonzo.TxWits (
+  AlonzoTxWits (..),
+  RdmrPtr (..),
+  Redeemers (..),
+  TxDats (..),
+ )
+import Cardano.Ledger.BaseTypes (
+  EpochNo (..),
+  ProtVer (..),
+  SlotNo (..),
+  StrictMaybe,
+  TxIx (..),
+  getVersion,
+  strictMaybeToMaybe,
+ )
+import Cardano.Ledger.Binary (Sized (..))
+import Cardano.Ledger.Block (txid)
+import Cardano.Ledger.Coin (Coin (..))
+import Cardano.Ledger.Conway.Core (
+  AlonzoEraTxBody (..),
+  CoinPerByte (..),
+  EraScript (..),
+  EraTxWits (..),
+  vldtTxBodyL,
+ )
+import Cardano.Ledger.Conway.PParams (
+  ConwayPParams (..),
+  DRepVotingThresholds,
+  PoolVotingThresholds,
+  THKD (..),
+ )
+import Cardano.Ledger.Conway.TxOut (BabbageTxOut (..))
+import Cardano.Ledger.Core (
+  Era,
+  EraIndependentData,
+  EraIndependentScriptIntegrity,
+  EraIndependentTxBody,
+  EraTx (..),
+  EraTxAuxData (..),
+  EraTxBody (..),
+  EraTxOut (..),
+  PParams,
+ )
+import Cardano.Ledger.Crypto (Crypto (..))
+import Cardano.Ledger.HKD (HKD)
+import Cardano.Ledger.Keys (KeyHash (..), VKey (..))
+import Cardano.Ledger.Keys.WitVKey (WitVKey (..))
+import Cardano.Ledger.SafeHash (SafeHash, extractHash)
+import Cardano.Ledger.Shelley.LedgerState (UTxOState (..))
+import Cardano.Ledger.Shelley.Rules (Identity, UtxoEnv (..))
+import Cardano.Ledger.TxIn (TxId (..), TxIn (..))
+import Cardano.Ledger.UTxO (UTxO (..))
+import Cardano.Ledger.Val (Val (..))
+import Data.Bitraversable (bimapM)
+import Data.ByteString (ByteString)
+import qualified Data.ByteString as BS
+import Data.Data (Typeable)
+import Data.Foldable (Foldable (..))
+import Data.Map.Strict (Map)
+import qualified Data.Map.Strict as Map
+import Data.Sequence.Strict (StrictSeq)
+import Data.Set (Set)
+import qualified Data.Set as Set
+import Data.Text (Text)
+import Data.Word (Word64)
+import Lens.Micro ((^.))
+import qualified Lib as Agda
 
--- Placeholder so that the compiler wouldn't complain about
--- `cardano-ledger-executable-spec` being a redundant dependency
-_placeholder :: TxBody
-_placeholder = undefined
+type SpecTranslationError = Text
+
+class SpecTranslate a where
+  type SpecRep a
+
+  toSpecRep :: a -> Either SpecTranslationError (SpecRep a)
+
+instance SpecTranslate TxIx where
+  type SpecRep TxIx = Integer
+
+  toSpecRep (TxIx x) = pure $ toInteger x
+
+instance SpecTranslate (TxIn era) where
+  type SpecRep (TxIn era) = Agda.TxIn
+
+  toSpecRep (TxIn txId txIx) = toSpecRep (txId, txIx)
+
+byteStringToInteger :: ByteString -> Integer
+byteStringToInteger = BS.foldr' (\x y -> y * 256 + toInteger x) 0
+
+instance SpecTranslate (Addr era) where
+  type SpecRep (Addr era) = Agda.Addr
+
+  toSpecRep = pure . byteStringToInteger . serialiseAddr
+
+instance SpecTranslate (SafeHash c EraIndependentData) where
+  type SpecRep (SafeHash c EraIndependentData) = Agda.DataHash
+
+  toSpecRep _ = pure ()
+
+instance SpecTranslate (SafeHash c EraIndependentScriptIntegrity) where
+  type SpecRep (SafeHash c EraIndependentScriptIntegrity) = Agda.Hash
+
+  toSpecRep = toSpecRep . extractHash
+
+instance SpecTranslate (SafeHash c EraIndependentTxBody) where
+  type SpecRep (SafeHash c EraIndependentTxBody) = Agda.Hash
+
+  toSpecRep = toSpecRep . extractHash
+
+instance EraTxOut era => SpecTranslate (BabbageTxOut era) where
+  type SpecRep (BabbageTxOut era) = Agda.TxOut
+
+  toSpecRep (BabbageTxOut addr val datum _) =
+    let mkTxOut x = do
+          addr' <- toSpecRep addr
+          coin' <- toSpecRep $ coin val
+          pure (addr', (coin', x))
+     in case datum of
+          NoDatum -> mkTxOut Nothing
+          DatumHash h -> mkTxOut . Just =<< toSpecRep h
+          Datum _ -> Left "Inline datums not supported by spec"
+
+instance SpecTranslate Integer where
+  type SpecRep Integer = Integer
+
+  toSpecRep = pure
+
+deriving instance SpecTranslate Coin
+
+deriving instance
+  ( SpecTranslate (TxOut era)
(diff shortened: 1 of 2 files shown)


libs/cardano-ledger-conformance/src/Test/Cardano/Ledger/Conformance.hs
```diff
@@ -1,10 +1,368 @@
...
+import Cardano.Crypto.DSIGN (DSIGNAlgorithm(..), SignedDSIGN (..))
 
-import Lib (TxBody)
+type SpecTranslationError = Text
 
--- Placeholder so that the compiler wouldn't complain about
--- `cardano-ledger-executable-spec` being a redundant dependency
-_placeholder :: TxBody
-_placeholder = undefined
+class SpecCastable a where
```
lehins: We'll also need something similar for converting some of the types back from spec rep.

So, maybe name it `ToSpecRep` and another one will be `FromSpecRep`.

`SpecRep` can be a standalone type family, it doesn't have to be part of the type class

Soupstraw: I don't think we need to, we just have to do all the comparisons on Agda types. The Agda types contain less information than the ones in implementation, so if we wanted to convert the values back to our types then we'd have to fill that in with either default values or provide some context. The prototype of conformance testing did all the comparisons on the Agda types.

lehins: Yeah, that's great. As long as we are not using Show instance for equality :smile:

libs/cardano-ledger-conformance/src/Test/Cardano/Ledger/Conformance.hs
```diff
@@ -1,10 +1,446 @@
...
+
+instance SpecCastable (Addr era) where
+  type SpecRep (Addr era) = Agda.Addr
+
+  toSpecRep = pure . byteStringToInteger . serialiseAddr
+
+instance SpecCastable (SafeHash c EraIndependentData) where
+  type SpecRep (SafeHash c EraIndependentData) = Agda.DataHash
+
+  toSpecRep _ = pure ()
```
lehins, line 122: Is `Agda.DataHash` really a unit?
I guess It makes sense, since Alonzo era is still not implemented

libs/cardano-ledger-conformance/src/Test/Cardano/Ledger/Conformance.hs
```diff
@@ -1,10 +1,446 @@
...
+    let mkTxOut x = do
+          addr' <- toSpecRep addr
+          coin' <- toSpecRep $ coin val
+          pure
+            ( addr'
+            ,
+              ( coin'
+              , x
+              )
+            )
```
lehins: It is really hard for me to believe that you would actually like this sort of style for your code? Why do you choose to write it that way?
```suggestion
          pure (addr', (coin', x))
```

libs/cardano-ledger-conformance/src/Test/Cardano/Ledger/Conformance.hs
```diff
@@ -1,10 +1,455 @@
...
+
+instance SpecCastable (SafeHash c EraIndependentTxBody) where
+  type SpecRep (SafeHash c EraIndependentTxBody) = Agda.Hash
+
+  toSpecRep = toSpecRep . extractHash
+
+instance EraTxOut era => SpecCastable (BabbageTxOut era) where
+  type SpecRep (BabbageTxOut era) = Agda.TxOut
+
+  toSpecRep (BabbageTxOut addr val datum _) =
```
lehins, line 137: We should fail translation when refScript is present, just as with inline Datums
```suggestion
  toSpecRep (BabbageTxOut addr val datum refScript) =
```

libs/cardano-ledger-conformance/src/Test/Cardano/Ledger/Conformance.hs
```diff
@@ -1,10 +1,455 @@
...
+instance SpecCastable VotingThresholds where
+  type SpecRep VotingThresholds = ()
+
+  toSpecRep _ = pure ()
+
+instance SpecCastable CostModels where
+  -- TODO change this to unit once it is fixed in the spec
+  type SpecRep CostModels = Agda.AgdaEmpty
+
+  toSpecRep _ = pure undefined
```
lehins: Leaving undefineds around is a bad idea. Switching to practice like this will help you in the long run. 
```suggestion
  type SpecRep CostModels = Agda.AgdaEmpty

  toSpecRep _ = error "TODO change this to unit once it is fixed in the spec"
```

I understand that you like `undefined` as a place holder. Next time you try using `undefined`, could you first implement this ticket: https://github.com/input-output-hk/cardano-ledger/issues/3376

libs/cardano-ledger-conformance/src/Test/Cardano/Ledger/Conformance.hs
```diff
@@ -1,10 +1,455 @@
...
+    Agda.MkPParams
+      <$> toSpecRep (cppMinFeeA x)
+      <*> toSpecRep (cppMinFeeB x)
+      <*> Right (toInteger $ cppMaxBBSize x)
+      <*> Right (toInteger $ cppMaxTxSize x)
+      <*> Right (toInteger $ cppMaxBHSize x)
+      <*> Right (toInteger $ cppMaxValSize x)
+      -- This field corresponds to minUTxOValue, which we don't have in Conway.
+      -- That is a mistake in the spec types.
+      <*> Right undefined
```
lehins: Ditto

libs/cardano-ledger-conformance/src/Test/Cardano/Ledger/Conformance.hs
```diff
@@ -1,10 +1,455 @@
...
+        )
+      <*> toSpecRep (cppGovActionLifetime x)
+      <*> toSpecRep (cppGovActionDeposit x)
+      <*> toSpecRep (cppDRepDeposit x)
+      <*> toSpecRep (cppDRepActivity x)
+      <*> Right (toInteger $ cppCommitteeMinSize x)
+      <*> Right (toInteger $ cppCommitteeMaxTermLength x)
+      -- This is the minimumAVS field which we ended up not needing. It will be
+      -- removed from the spec.
+      <*> Right undefined
```
lehins: Ditto

libs/cardano-ledger-conformance/src/Test/Cardano/Ledger/Conformance.hs
```diff
@@ -1,10 +1,455 @@
...
+
+instance SpecCastable OrdExUnits where
+  type SpecRep OrdExUnits = Agda.ExUnits
+
+  toSpecRep (OrdExUnits exUnits) = toSpecRep exUnits
+
+instance SpecCastable CoinPerByte where
+  type SpecRep CoinPerByte = Agda.Coin
+
+  toSpecRep (CoinPerByte x) = toSpecRep x
```
lehins: You can use StandaloneDeriving and GND to reduce boiler plate for newtypes
```suggestion
deriving instance SpecCastable OrdExUnits
deriving instance SpecCastable CoinPerByte
```

libs/cardano-ledger-conformance/src/Test/Cardano/Ledger/Conformance.hs
```diff
@@ -1,10 +1,455 @@
...
+
+instance SpecCastable (KeyHash r c) where
+  type SpecRep (KeyHash r c) = Agda.Hash
+
+  toSpecRep (KeyHash h) = toSpecRep h
+
+instance
+  Crypto c =>
+  SpecCastable (VKey k c)
+  where
```
lehins:
```suggestion
instance Crypto c => SpecCastable (VKey k c) where
```

libs/cardano-ledger-conformance/src/Test/Cardano/Ledger/Conformance.hs
```diff
@@ -1,10 +1,455 @@
...
+
+  toSpecRep (WitVKey vk sk) =
+    (,)
+      <$> toSpecRep vk
+      <*> toSpecRep sk
+
+instance Era era => SpecCastable (TxDats era) where
+  type SpecRep (TxDats era) = [(Agda.DataHash, Agda.Datum)]
+
+  toSpecRep (TxDats x) = toSpecRep x
```
lehins: Same here:
```suggestion
deriving instance Era era => SpecCastable (TxDats era)
```

libs/cardano-ledger-conformance/src/Test/Cardano/Ledger/Conformance.hs
```diff
@@ -1,10 +1,455 @@
...
+
+instance (SpecCastable k, SpecCastable v) => SpecCastable (Map k v) where
+  type SpecRep (Map k v) = [(SpecRep k, SpecRep v)]
+
+  toSpecRep m = traverse act $ Map.toList m
+    where
+      act (k, v) = do
+        k' <- toSpecRep k
+        v' <- toSpecRep v
+        pure (k', v')
```
lehins: You already have `bimapM` scope. In fact copied your implementation over :slightly_smiling_face: :
```suggestion
  toSpecRep = traverse (bimapM toSpecRep toSpecRep) . Map.toList
```

Or even better, add this instance:
```
instance (SpecCastable a, SpecCastable b) => SpecCastable (a, b) where
  type SpecRep (a, b) = (a, b)
  toSpecRep (a, b) = (,) <$> toSpecRep a <*> toSpecRep b
```

and you can reuse it here:

```haskell
  toSpecRep = traverse toSpecRep . Map.toList
```
and in a couple of other places

libs/cardano-ledger-conformance/src/Test/Cardano/Ledger/Conformance.hs
```diff
@@ -1,10 +1,455 @@
...
+  , SpecRep (TxOut era) ~ Agda.TxOut
+  ) =>
+  SpecCastable (UTxO era)
+  where
+  type SpecRep (UTxO era) = Agda.UTxO
+
+  toSpecRep utxo =
+    traverse (bimapM toSpecRep toSpecRep)
+      . Map.toList
+      $ unUTxO utxo
```
lehins:
```suggestion
  toSpecRep = toSpecRep . unUTxO
```

libs/cardano-ledger-conformance/src/Test/Cardano/Ledger/Conformance.hs
```diff
@@ -1,10 +1,455 @@
...
+  , Script era ~ AlonzoScript era
+  ) =>
+  SpecCastable (AlonzoTxWits era)
+  where
+  type SpecRep (AlonzoTxWits era) = Agda.TxWitnesses
+
+  toSpecRep x =
+    Agda.MkTxWitnesses
+      <$> toSpecRep (txwitsVKey x)
+      <*> pure []
```
lehins, line 364: I think we should fail the translation if there are any scripts.

libs/cardano-ledger-conformance/src/Test/Cardano/Ledger/Conformance.hs
```diff
@@ -1,10 +1,455 @@
...
+
+instance SpecCastable a => SpecCastable (StrictMaybe a) where
+  type SpecRep (StrictMaybe a) = Maybe (SpecRep a)
+
+  toSpecRep = traverse toSpecRep . strictMaybeToMaybe
+
+instance SpecCastable (AlonzoTxAuxData era) where
+  type SpecRep (AlonzoTxAuxData era) = Agda.AuxiliaryData
+
+  toSpecRep _ = pure ()
```
lehins, line 376: Same here. I might be wrong, but I believe for every feature that executable spec does not support, we should fail the translation. Because we'll need to ensure that our generators do not produce unsupported values, otherwise if they do, the result of running the rule will be different to what Agda spec will expect

libs/cardano-ledger-conformance/src/Test/Cardano/Ledger/Conformance.hs
```diff
@@ -1,10 +1,455 @@
...
+  , EraTx era
+  , AlonzoEraTxBody era
+  , SpecCastable (TxOut era)
+  ) =>
+  Tx era ->
+  Either SpecTranslationError Agda.TxBody
+toAgdaTxBody tx =
+  Agda.MkTxBody
+    <$> toSpecRep (tx ^. bodyTxL . inputsTxBodyL)
+    <*> fmap (zip [0 ..]) (toSpecRep $ tx ^. bodyTxL . outputsTxBodyL)
```
lehins: A slightly shorter way of writing the same thing :slightly_smiling_face: 
```suggestion
    <*> (zip [0 ..] <$> toSpecRep (tx ^. bodyTxL . outputsTxBodyL))
```

lehins (approved):
Bunch of little suggestions, otherwise looks great!

One thing I am not totally sold on is the name `SpecCastable`. This is more of a translation that can fail, rather than casting of one type to another. `SpecTrans` maybe? In any case, it's up to you, this is also only a suggestion

libs/cardano-ledger-conformance/src/Test/Cardano/Ledger/Conformance.hs
Soupstraw: I think I had longer things in the tuples at some point, but then I just forgot to change it

libs/cardano-ledger-conformance/src/Test/Cardano/Ledger/Conformance.hs
lehins: I was thinking more about this while reviewing the PR and I think it will be problematic if we do comparison on the Agda side. This about test failures that we will get, for example instead of hash values that we generated we will get bunch of integers. So it might be hard to map to the original values that we will need to use for debugging failures.

We don't have solve it in this PR, but some food for though. If we are to do the translation from executable spec back to Ledger code we can provide the current Ledger value and just overwrite only the fields that are present in executable spec. Eg.
```haskell
fromSpecRep :: a -> SpecRep a -> a
```
I don't think it would even have to have a failure mode (i.e. `Either`), since I am sure we can represent all of the executable spec types

Soupstraw: I thought of this as well and I think I have a better idea, we can talk about this later.

libs/cardano-ledger-conformance/src/Test/Cardano/Ledger/Conformance.hs
Soupstraw: I think it's safer to succeed at first and then if we can't manage to make the tests pass, then we can fail the translation in these places. Because if we fail here, we discard tests, but if we succeed with the translation, we can expect to get a test failure.

libs/cardano-ledger-conformance/src/Test/Cardano/Ledger/Conformance.hs
Soupstraw: GHC can't figure this one out for some reason :/

lehins: Hmmm. Interesting. I suspect it can't figure out recursively that `SpecRep (DataHash (EraCrypto era)) = Agda.DataHash`

libs/cardano-ledger-conformance/src/Test/Cardano/Ledger/Conformance.hs
lehins: Sure. If you think it makes sense to postpone making this decision, I have nothing against it

PR #3852, opened 2023-11-06
Added constrained generators to conformance tests - Part 1

Soupstraw:
# Description

This PR implements conformance testing with constrained generators. It's now possible to generate test cases with the constrained generator and then run the test against the Agda spec.

I had to disable the following parts of the generator to accommodate for features that the spec does not yet model:
 * failing scripts
 * withdrawals
 * certificates (currently there's no field for certs in the MAlonzo type, but otherwise the certs are implemented in the spec)
 * inline datums

#3811

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/ImpTest.hs
+++ b/eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/ImpTest.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE CPP #-}
 {-# LANGUAGE DataKinds #-}
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE LambdaCase #-}
@@ -67,7 +68,7 @@ import Cardano.Ledger.Conway.Governance (
   ConwayGovState,
   DRepPulsingState (DRComplete),
   EnactState (..),
-  GovAction,
+  GovAction (..),
   GovActionId (..),
   GovActionIx (..),
   GovActionState (..),
@@ -112,7 +113,10 @@ import Cardano.Ledger.Conway.Rules (
   validCommitteeTerm,
   withdrawalCanWithdraw,
  )
-import Cardano.Ledger.Conway.TxBody (ConwayEraTxBody (..))
+import Cardano.Ledger.Conway.TxBody (
+  ConwayEraTxBody (..),
+  proposalProceduresTxBodyL,
+ )
 import Cardano.Ledger.Conway.TxCert (
   ConwayEraTxCert (..),
   Delegatee (..),
@@ -145,9 +149,8 @@ import Cardano.Ledger.Shelley.LedgerState (
   vsCommitteeStateL,
   vsDRepsL,
  )
-import Cardano.Ledger.TxIn (TxId)
+import Cardano.Ledger.TxIn (TxId (..))
 import Cardano.Ledger.Val (Val (..))
-import Control.Monad (void)
 import Control.State.Transition.Extended (STS (..))
 import Data.Default.Class (Default (..))
 import Data.Foldable (Foldable (..))
--- a/eras/shelley/impl/testlib/Test/Cardano/Ledger/Shelley/ImpTest.hs
+++ b/eras/shelley/impl/testlib/Test/Cardano/Ledger/Shelley/ImpTest.hs
@@ -1,5 +1,4 @@
 {-# LANGUAGE AllowAmbiguousTypes #-}
-{-# LANGUAGE BangPatterns #-}
 {-# LANGUAGE DataKinds #-}
 {-# LANGUAGE DeriveFunctor #-}
 {-# LANGUAGE FlexibleContexts #-}
@@ -7,7 +6,6 @@
 {-# LANGUAGE GADTs #-}
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
 {-# LANGUAGE ImplicitParams #-}
-{-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE MultiParamTypeClasses #-}
 {-# LANGUAGE NamedFieldPuns #-}
 {-# LANGUAGE NumericUnderscores #-}
@@ -29,9 +27,9 @@ module Test.Cardano.Ledger.Shelley.ImpTest (
   evalImpTestM,
   execImpTestM,
   ImpTestState,
+  ImpTestEnv (..),
   ImpException (..),
   ShelleyEraImp (..),
-  logStakeDistr,
   emptyShelleyImpNES,
   shelleyImpWitsVKeyNeeded,
   modifyPrevPParams,
@@ -43,17 +41,24 @@ module Test.Cardano.Ledger.Shelley.ImpTest (
   submitTx,
   submitFailingTx,
   trySubmitTx,
-  logEntry,
   modifyNES,
   getsNES,
   impAnn,
   mkTxWits,
-  impNESL,
   runImpRule,
   registerRewardAccount,
   getRewardAccountAmount,
   constitutionShouldBe,
   withImpState,
+  fixupFees,
+  -- Logging
+  logEntry,
+  logStakeDistr,
+  -- Combinators
+  withNoFixup,
+  -- Lenses
+  impNESL,
+  impLastTickL,
 ) where
 
 import Cardano.Crypto.DSIGN (DSIGNAlgorithm (..), seedSizeDSIGN)
@@ -119,11 +124,10 @@ import Cardano.Ledger.UMap as UMap
 import Cardano.Ledger.UTxO (EraUTxO (..), UTxO (..), sumAllCoin)
 import Cardano.Ledger.Val (Val (..))
 import Control.DeepSeq (NFData)
-import Control.Monad (void)
 import Control.Monad.IO.Class
-import Control.Monad.Reader (MonadReader)
+import Control.Monad.Reader (MonadReader (..), asks)
 import Control.Monad.State.Strict (MonadState (..), gets, modify)
-import Control.Monad.Trans.Reader (ReaderT (..), ask)
+import Control.Monad.Trans.Reader (ReaderT (..))
 import Control.State.Transition (STS (..), TRC (..))
 import Control.State.Transition.Trace (applySTSTest)
 import qualified Data.ByteString as BS
@@ -365,21 +369,31 @@ shelleyImpWitsVKeyNeeded nes txb = shelleyWitsVKeyNeeded utxo txb genDelegs
     utxo = nes ^. nesEsL . esLStateL . lsUTxOStateL . utxosUtxoL
     genDelegs = nes ^. nesEsL . esLStateL . lsCertStateL . certDStateL . dsGenDelegsL
 
-newtype ImpTestM era a = ImpTestM (ReaderT (IORef (ImpTestState era)) IO a)
+data ImpTestEnv era = ImpTestEnv
+  { iteState :: !(IORef (ImpTestState era))
+  , iteDoTxFixup :: !Bool
+  -- ^ This flag is used to toggle the fixing up of transactions. If it
+  -- is set to False then any transaction should be submitted as-is.
+  }
+
+iteDoTxFixupL :: Lens' (ImpTestEnv era) Bool
+iteDoTxFixupL = lens iteDoTxFixup (\x y -> x {iteDoTxFixup = y})
+
+newtype ImpTestM era a = ImpTestM (ReaderT (ImpTestEnv era) IO a)
   deriving
     ( Functor
     , Applicative
     , Monad
     , MonadIO
     , MonadUnliftIO
-    , MonadReader (IORef (ImpTestState era))
+    , MonadReader (ImpTestEnv era)
     )
 
 instance MonadState (ImpTestState era) (ImpTestM era) where
   get = ImpTestM $ do
-    liftIO . readIORef =<< ask
+    liftIO . readIORef . iteState =<< ask
   put x = ImpTestM $ do
-    liftIO . flip writeIORef x =<< ask
+    liftIO . flip writeIORef x . iteState =<< ask
 
 instance Example (ImpTestM era ()) where
   type Arg (ImpTestM era ()) = ImpTestState era
@@ -403,8 +417,14 @@ runImpTestM_ impState = void . runImpTestM impState
 runImpTestM :: ImpTestState era -> ImpTestM era b -> IO (b, ImpTestState era)
 runImpTestM impState (ImpTestM m) = do
   ioRef <- newIORef impState
+  let
+    env =
+      ImpTestEnv
+        { iteState = ioRef
+        , iteDoTxFixup = True
+        }
   res <-
-    runReaderT m ioRef `catchAny` \exc -> do
+    runReaderT m env `catchAny` \exc -> do
       logsDoc <- impLog <$> readIORef ioRef
       let logs = renderString (layoutPretty defaultLayoutOptions logsDoc)
           adjustHUnitExc header (HUnitFailure srcLoc failReason) =
@@ -483,7 +503,11 @@ submitTx_ ::
     )
 submitTx_ tx = do
   st <- gets impNES
-  txFixed <- fixupFees tx
+  doFixup <- asks iteDoTxFixup
+  txFixed <-
+    if doFixup
+      then fixupFees tx
+      else pure tx
   lEnv <- impLedgerEnv st
   globals <- use $ to impGlobals
   let
@@ -535,7 +559,7 @@ runImpRule stsEnv stsState stsSignal = do
       ruleName = symbolVal (Proxy @rule)
   globals <- use $ to impGlobals
   case runShelleyBase globals (applySTSTest @(EraRule rule era) trc) of
-    Left [] -> error $ "Impossible: STS rule failed without a predicate failure"
+    Left [] -> error "Impossible: STS rule failed without a predicate failure"
     Left fs ->
       assertFailure $
         unlines $
@@ -735,7 +759,7 @@ registerRewardAccount = do
           .~ SSeq.fromList
             [ mkBasicTxOut
                 (mkAddr (kpSpending, kpDelegator))
-                (inject $ Coin 10000000)
+                (inject $ Coin 10_000_000)
             ]
         & bodyTxL . certsTxBodyL
           .~ SSeq.fromList [RegTxCert @era stakingCredential]
@@ -756,3 +780,7 @@ constitutionShouldBe cUrl = do
           (error "No constitution has been set")
           constitution
   anchorUrl `shouldBe` fromJust (textToUrl $ T.pack cUrl)
+
+-- | Performs the action without running the fix-up function on any transactions
+withNoFixup :: ImpTestM era a -> ImpTestM era a
+withNoFixup = local $ iteDoTxFixupL .~ False
--- a/libs/cardano-ledger-binary/CHANGELOG.md
+++ b/libs/cardano-ledger-binary/CHANGELOG.md
@@ -4,6 +4,10 @@
 
 * Moved `ToExpr` instances out of the main library and into the testlib.
 
+### `testlib`
+
+* Add `diffExprCompact`
+
 ## 1.2.1.0
 
 * Export `decodeListLikeEnforceNoDuplicates` #3791
--- a/libs/cardano-ledger-conformance/cardano-ledger-conformance.cabal
+++ b/libs/cardano-ledger-conformance/cardano-ledger-conformance.cabal
@@ -21,6 +21,7 @@ flag asserts
 library
     exposed-modules:  Test.Cardano.Ledger.Conformance
     hs-source-dirs:   src
+    other-modules:    Test.Cardano.Ledger.Conformance.Orphans
     default-language: Haskell2010
     ghc-options:
         -Wall -Wcompat -Wincomplete-record-updates
@@ -28,20 +29,44 @@ library
         -Wunused-packages
 
     build-depends:
-        base >=4.14 && <4.19,
-        bytestring,
+        base >=4.14 && <5,
         cardano-strict-containers,
         microlens,
+        mtl,
+        bytestring,
+        data-default-class,
         cardano-ledger-binary,
         cardano-ledger-core,
         cardano-ledger-allegra,
+        cardano-ledger-mary,
         cardano-ledger-shelley,
         cardano-ledger-alonzo,
-        cardano-ledger-conway,
+        cardano-ledger-conway:{cardano-ledger-conway, testlib},
         cardano-ledger-executable-spec,
         cardano-crypto-class,
+        cardano-ledger-core:testlib,
+        cardano-ledger-test,
         containers,
+        small-steps,
         text
+
+    if !impl(ghc >=9.2)
+        ghc-options: -Wno-incomplete-patterns
 
     if flag(asserts)
         ghc-options: -fno-ignore-asserts
+
+test-suite tests
+    type:             exitcode-stdio-1.0
+    main-is:          Main.hs
+    hs-source-dirs:   test
+    default-language: Haskell2010
+    ghc-options:
+        -Wall -Wcompat -Wincomplete-record-updates
+        -Wincomplete-uni-patterns -Wredundant-constraints -Wpartial-fields
+        -Wunused-packages -threaded -rtsopts -with-rtsopts=-N
+
+    build-depends:
+        base >=4.14 && <5,
+        cardano-ledger-conformance,
+        cardano-ledger-core:testlib
--- a/libs/cardano-ledger-conformance/src/Test/Cardano/Ledger/Conformance.hs
+++ b/libs/cardano-ledger-conformance/src/Test/Cardano/Ledger/Conformance.hs
@@ -1,8 +1,13 @@
+{-# LANGUAGE AllowAmbiguousTypes #-}
+{-# LANGUAGE DataKinds #-}
+{-# LANGUAGE DefaultSignatures #-}
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE FlexibleInstances #-}
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE NamedFieldPuns #-}
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE PolyKinds #-}
+{-# LANGUAGE RecordWildCards #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE StandaloneDeriving #-}
 {-# LANGUAGE TypeApplications #-}
@@ -12,6 +17,7 @@
 
 module Test.Cardano.Ledger.Conformance (
   SpecTranslate (..),
+  spec,
 ) where
 
 import Cardano.Crypto.DSIGN (DSIGNAlgorithm (..), SignedDSIGN (..))
@@ -20,8 +26,8 @@ import Cardano.Ledger.Address (Addr (..), serialiseAddr)
 import Cardano.Ledger.Allegra.Scripts (ValidityInterval (..))
 import Cardano.Ledger.Alonzo (AlonzoScript, AlonzoTxAuxData)
 import Cardano.Ledger.Alonzo.PParams (OrdExUnits (OrdExUnits))
-import Cardano.Ledger.Alonzo.Scripts (CostModels, ExUnits (..), Prices, Tag (..))
-import Cardano.Ledger.Alonzo.Tx (AlonzoTx (..))
+import Cardano.Ledger.Alonzo.Scripts (AlonzoScript (..), CostModels, ExUnits (..), Prices, Tag (..))
+import Cardano.Ledger.Alonzo.Tx (AlonzoEraTx (..), AlonzoTx (..), IsValid (..))
 import Cardano.Ledger.Alonzo.TxWits (
   AlonzoTxWits (..),
   RdmrPtr (..),
@@ -33,7 +39,7 @@ import Cardano.Ledger.BaseTypes (
   EpochNo (..),
   ProtVer (..),
   SlotNo (..),
-  StrictMaybe,
+  StrictMaybe (..),
   TxIx (..),
   getVersion,
   strictMaybeToMaybe,
@@ -41,22 +47,29 @@ import Cardano.Ledger.BaseTypes (
 import Cardano.Ledger.Binary (Sized (..))
 import Cardano.Ledger.Block (txid)
 import Cardano.Ledger.Coin (Coin (..))
+import Cardano.Ledger.Conway (Conway)
 import Cardano.Ledger.Conway.Core (
   AlonzoEraTxBody (..),
+  BabbageEraTxBody (..),
   CoinPerByte (..),
+  Era (..),
+  EraPParams (..),
+  EraRule,
   EraScript (..),
   EraTxWits (..),
+  PParams (..),
   vldtTxBodyL,
  )
+import Cardano.Ledger.Conway.Governance (GovAction (..), ProposalProcedure (..), VotingProcedures (..))
 import Cardano.Ledger.Conway.PParams (
   ConwayPParams (..),
   DRepVotingThresholds,
   PoolVotingThresholds,
   THKD (..),
  )
+import Cardano.Ledger.Conway.TxBody (ConwayEraTxBody (..))
 import Cardano.Ledger.Conway.TxOut (BabbageTxOut (..))
 import Cardano.Ledger.Core (
-  Era,
   EraIndependentData,
   EraIndependentScriptIntegrity,
   EraIndependentTxBody,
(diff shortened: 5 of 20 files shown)


libs/cardano-ledger-conformance/src/Test/Cardano/Ledger/Conformance.hs
```diff
@@ -55,49 +67,108 @@ import Cardano.Ledger.Conway.PParams (
...
+  impDoFixupL,
+  impLastTickL,
+  logCurPParams,
+  logEntry,
+  modifyNES,
+  trySubmitTx,
+  withImpState,
+ )
+import Test.Cardano.Ledger.Generic.Proof (Reflect (..))
+import Test.Cardano.Ledger.Imp.Common (ToExpr (..), generate, shouldBe)
```
lehins: `generate` should never be used in tests, because it makes the test not reproducible. If you don't wanna solve it now and use it temporarely, that's fine, but make sure to leave a big `FIXME`, since that is not acceptable for property tests. (Imagine, any failure on CI cannot be reproduced locally, which makes the test case useless for practical purposes)

Also, the point of Common is that it brings into scope all of the tooling we need for test, it's like a testing Prelude. Moreover, never is ` Test.Cardano.Ledger.Common` needs to be imported together with `Test.Cardano.Ledger.Imp.Common`
```suggestion
import Test.Cardano.Ledger.Imp.Common
```

libs/cardano-ledger-conformance/src/Test/Cardano/Ledger/Conformance.hs
```diff
@@ -55,49 +67,108 @@ import Cardano.Ledger.Conway.PParams (
...
+import Test.Cardano.Ledger.Common (
+  HasCallStack,
+  Spec,
+  describe,
+  diffExprCompact,
+  it,
+  showExpr,
+  unless,
+  xit,
+ )
```
lehins: `Test.Cardano.Ledger.Imp.Common` already re-exports all of `Test.Cardano.Ledger.Common`.
```suggestion
```

libs/cardano-ledger-conformance/cardano-ledger-conformance.cabal
```diff
@@ -33,16 +33,37 @@ library
...
+    main-is:          Main.hs
+    hs-source-dirs:   test
+    default-language: Haskell2010
+    ghc-options:
+        -Wall -Wcompat -Wincomplete-record-updates
+        -Wincomplete-uni-patterns -Wredundant-constraints -Wpartial-fields
+        -Wunused-packages -threaded -rtsopts -with-rtsopts=-N
+
+    build-depends:
+        base >=4.14 && <4.19,
```
lehins: Upper bound on `base` is dumb. Have no idea why people add it
```suggestion
        base >=4.14 && <5,
```

libs/cardano-ledger-conformance/src/Test/Cardano/Ledger/Conformance.hs
```diff
@@ -418,3 +504,140 @@ instance
...
+        , ueCertState = certState
+        }
+  agdaUtxoEnv <- case toSpecRep utxoEnv of
+    Right x -> pure x
+    Left e -> fail $ T.unpack e
+  let
+    newAgdaUtxoState = Agda.utxoStep agdaUtxoEnv agdaUtxoState agdaTx
+  modify $ impDoFixupL .~ False
+  submitRes <- trySubmitTx tx
+  modify $ impDoFixupL .~ doFixup
```
lehins: Why are using a state variable for what should be a simple argument to `trySubmitTx`?

libs/cardano-ledger-conformance/src/Test/Cardano/Ledger/Conformance.hs
```diff
@@ -418,3 +504,140 @@ instance
...
+    logEntry $ diffExprCompact (toExpr finalAgdaState) (toExpr finalImplState)
+    fail "The final states of spec and implementation do not match"
+  pure submitRes
+
+spec :: HasCallStack => Spec
+spec = describe "Conway conformance tests" . withImpState $ do
+  it "Simple tx" $ do
+    modify $ \x ->
+      x
+        & impLastTickL .~ 100
```
lehins:
```suggestion
    modify (impLastTickL .~ 100)
```

libs/cardano-ledger-conformance/src/Test/Cardano/Ledger/Conformance.hs
```diff
@@ -418,3 +504,140 @@ instance
...
+      Right _ -> case toTestRep newUtxoState of
+        Right x -> Just $ toExpr x
+        Left _ -> Nothing
+      Left e ->
+        error $
+          "Failed to submit transaction on the implementation side:\n"
+            <> show e
+  unless (finalAgdaState == finalImplState) $ do
+    logEntry $ diffExprCompact (toExpr finalAgdaState) (toExpr finalImplState)
+    fail "The final states of spec and implementation do not match"
```
lehins: `fail` does not produce a callstack
```suggestion
    error "The final states of spec and implementation do not match"
```

libs/cardano-ledger-conformance/src/Test/Cardano/Ledger/Conformance.hs
```diff
@@ -418,3 +504,140 @@ instance
...
+    txouts = txb ^. outputsTxBodyL
+    maxValSize = pp ^. ppMaxValSizeL
+    serSize = fromIntegral . BSL.length . serialize (pvMajor (pp ^. ppProtocolVersionL))
+    consumedVal = consumed pp s utxo txb
+    producedVal = produced pp s txb
+  slot <- gets $ view impLastTickL
+  modifyNES $ const nes
+  agdaUtxoState <- case toSpecRep $ nes ^. nesEsL . esLStateL . lsUTxOStateL of
+    Right x -> pure x
+    Left e -> fail $ T.unpack e
```
lehins: Don't use `fail` in tests. Failure messages from IO for MonadFail are terrible
```suggestion
  agdaUtxoState <- 
    expectRight $ toSpecRep $ nes ^. nesEsL . esLStateL . lsUTxOStateL of
```

libs/cardano-ledger-conformance/src/Test/Cardano/Ledger/Conformance.hs
```diff
@@ -418,3 +504,140 @@ instance
...
+  agdaTx <- case toSpecRep tx of
+    Right x -> pure x
+    Left e -> fail $ T.unpack e
```
lehins:
```suggestion
  agdaTx <- expectRight $ toSpecRep tx
```

libs/cardano-ledger-conformance/src/Test/Cardano/Ledger/Conformance.hs
```diff
@@ -418,3 +504,140 @@ instance
...
+      certState <- getsNES $ nesEsL . esLStateL . lsCertStateL
+      govState <- getsNES $ nesEsL . esLStateL . lsUTxOStateL . utxosGovStateL
+      totalObligation certState govState `shouldBe` zero
+    do
+      deposited <- getsNES $ nesEsL . esLStateL . lsUTxOStateL . utxosDepositedL
+      deposited `shouldBe` zero
+    res <- trySubmitTxConform $ mkBasicTx mkBasicTxBody
+    case res of
+      Right _ -> pure ()
+      Left e -> error $ show e
```
lehins:
```suggestion
    _ <- expectRight $ trySubmitTxConform $ mkBasicTx mkBasicTxBody
```

I would recommend adding a helper for this (see `submitGovAcgtion_` for example):
https://github.com/input-output-hk/cardano-ledger/blob/4091022d46bbc6fc338970f1b76e452f3608700a/eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/ImpTest.hs#L440-L450

libs/cardano-ledger-conformance/src/Test/Cardano/Ledger/Conformance.hs
```diff
@@ -418,3 +504,140 @@ instance
...
+      toExpr . specToTestRep @(UTxOState Conway) <$> newAgdaUtxoState
+    finalImplState = case submitRes of
+      Right _ -> case toTestRep newUtxoState of
+        Right x -> Just $ toExpr x
+        Left _ -> Nothing
+      Left e ->
+        error $
+          "Failed to submit transaction on the implementation side:\n"
+            <> show e
+  unless (finalAgdaState == finalImplState) $ do
```
lehins, line 563: I guess you didn't find a better way to compare values without going through Show or ToExpr?

libs/cardano-ledger-conformance/src/Test/Cardano/Ledger/Conformance.hs
```diff
@@ -418,3 +504,140 @@ instance
...
+      Left e -> error $ show e
+  xit "Constrained generator" $ do
+    let
+      us =
+        def
+          { usMinCerts = 0
+          , usMaxCerts = 0
+          , usDatumFreq = 0
+          , usGenerateWithdrawals = False
+          }
```
lehins: Fun fact, I think you can turn Imp test into properties:
```suggestion
  xprop "Constrained generator" $ 
    let
      us =
        def
          { usMinCerts = 0
          , usMaxCerts = 0
          , usDatumFreq = 0
          , usGenerateWithdrawals = False
          }
     in forAll (genTxAndNewEpoch us $ reify @Conway) $ \ (nes, tx, _env) -> do
```

libs/cardano-ledger-conformance/src/Test/Cardano/Ledger/Conformance.hs
```diff
@@ -418,3 +504,140 @@ instance
...
+    modify $ \x ->
+      x
+        & impLastTickL .~ 100
+        & impDoFixupL .~ False
+    (nes, tx, _env) <- liftIO . generate . genTxAndNewEpoch us $ reify @Conway
```
lehins: Don't use `generate` outside debugging things in `ghci` or something
```suggestion
```

I know this is not the end of work on conformance tests, but still, I think we need to solve data generation first as part of this PR

lehins (changes requested):
It feels to me like this PR is a bit unfinished.

Why do you want to merge now?
It doesn't touch very much of the codebase that doesn't deal with conformance testing. Also it is not terribly big, so you can continue working on it.

For part 3. of #3811 I think it is important to generate reproducible valid data.

Let me know if you still want to merge it before it is fully complete for some reason, I am not gonna oppose it

libs/cardano-ledger-conformance/src/Test/Cardano/Ledger/Conformance.hs
Soupstraw: I thought it'd be clearer if it was something named rather than just some `Bool` argument to a function. Also you'll usually want to disable fixups for the whole test, not single transactions.

lehins: This "Imp"erative approach is stating to affect your coding style.

You chose to go with a mutable variable over using a functional approach!

All you need to do is create a separate function that does fixup and another that does not. This way it is not a flag to a function, but two separate functions with a separate name which can be used in all those tests that do or do not need fixing up.

libs/cardano-ledger-conformance/src/Test/Cardano/Ledger/Conformance.hs
```diff
@@ -418,3 +513,142 @@ instance
...
+      us =
+        def
+          { usMinCerts = 0
+          , usMaxCerts = 0
+          , usDatumFreq = 0
+          , usGenerateWithdrawals = False
+          , usNumPreUtxo = 3
+          , usNumColUtxo = 3
+          }
+    modify $ \x ->
```
lehins: I was sure I submitted this comment yesterday, but maybe I forgot to click the "Add" button:
I am pretty sure you can use `prop` with ImpTest if you want to have access to quickcheck generators:
```suggestion
  prop "Constrained generator" $
    let
      us =
        def
          { usMinCerts = 0
          , usMaxCerts = 0
          , usDatumFreq = 0
          , usGenerateWithdrawals = False
          , usNumPreUtxo = 3
          , usNumColUtxo = 3
          }
    in forAll (genTxAndNewEpoch us $ reify @Conway) $ \(nes, tx, _env) -> do          
      modify $ \x ->
```

libs/cardano-ledger-conformance/src/Test/Cardano/Ledger/Conformance.hs
Soupstraw: I think using `Expr` for comparison works just fine, I had to add some functions to `Map` and `Set` to deal with the ordering of elements, but otherwise `Expr` gives me all I need.

Soupstraw: but there's an option to use orphan `Eq` instances if we want to separate pretty printing from comparing, I just haven't felt the need for it so far.

lehins: > there's an option to use orphan Eq instances if we want to separate pretty printing from comparing

That is a way better solution!

Soupstraw: but then we might get a case where diffing and `Eq` instance behave differently so we don't get any useful information from diff. I think it's better to make sure that the equality on `ToExpr` expressions behaves like `Eq`

lehins: > but then we might get a case where diffing and Eq instance behave differently so we don't get any useful information from diff

That would be a bug in ToExpr instance, which we would want to catch.

Also comparison using Eq will be much faster than on Expr

libs/cardano-ledger-conformance/src/Test/Cardano/Ledger/Conformance.hs
Soupstraw: Ok, so I tried the approach you suggested, but I feel that I'd have to introduce so many functions to do the thing that I want to do. For example I'd have to make `trySubmitTxConform` and `trySubmitTxConformNoFixup` instead of just toggling a single boolean in the state.

Soupstraw: Maybe the better solution would be to introduce a `withNoFixup` combinator that would toggle the flag to `True` for a single command. It's a pretty widespread pattern from what I've seen in other libraries.

Soupstraw: In fact, I think we should add a `withExpectingFailure` combinator instead of having all the `submitTxFailing`/`submitTx` flavors of functions.

Soupstraw: We could also do that via a `Reader` monad I think by using the `local` function

libs/cardano-ledger-conformance/src/Test/Cardano/Ledger/Conformance.hs
```diff
@@ -426,3 +503,106 @@ instance
...
+        { ueSlot = lastTick
+        , uePParams = pParams
+        , ueGenDelegs = certState ^. certDStateL . dsGenDelegsL
+        , ueCertState = certState
+        }
+  agdaUtxoEnv <- expectRight $ toSpecRep utxoEnv
+  let
+    newAgdaUtxoState = Agda.utxoStep agdaUtxoEnv agdaUtxoState agdaTx
+  -- We already applied the fixup above, so no point doing it again
+  submitRes <- withNoFixup $ trySubmitTx tx
```
lehins, line 552: I still don't understand, why this would not work? Why do we need the whole mutable variable in the state for the fixup?
```suggestion
  submitRes <- trySubmitTx False tx
```

libs/cardano-ledger-conformance/src/Test/Cardano/Ledger/Conformance.hs
lehins: If you can solve this problem with a `Reader`, you can solve it with a simple function argument (that is what ReaderT is afterall), which is what I am suggesting you do

Soupstraw: I still think it's better to write `withNoFixup . withExpectFailure . submitTx` than having a separate function for every combination of configurations or forcing the user to pass a bunch of `Bool`s every time they try to submit a transaction. Most of the time you'll probably want to use the default configuration anyways.

Soupstraw: This pattern is used in [QuickCheck](https://hackage.haskell.org/package/QuickCheck-2.14.3/docs/src/Test.QuickCheck.Gen.html#resize) and [hspec](https://hackage.haskell.org/package/hspec-2.11.7/docs/Test-Hspec-QuickCheck.html#v:modifyMaxShrinks) also

lehins: It is very different. What you are doing is modifying a mutable variable, where the approaches you mentioned above are using modifying functions of the immutable scope variables, similar to `local` on Reader.

The thing you are missing is the presence of exceptions! Which makes your approach totally invalid.

The easiest way to solve it is with what I suggested: using an argument to a function. Or if you want to add extra immutable environment to the ImpTestM by the means of ReaderT, be my guest. But the solution you have right now is not gonna work

Soupstraw: I went with the `local` approach

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/ImpTest.hs
```diff
@@ -42,16 +49,21 @@ module Test.Cardano.Ledger.Conway.ImpTest (
   logRatificationChecks,
   registerCommitteeHotKey,
   logCurPParams,
+  logTxSummary,
+  -- Utils
+  byteStringToInteger,
```
lehins: Let's move those two functions into `cardano-ledger-conformance`.

Neither the transaction format used for logging nor conversion of bytstring to integer makes sense outside of conformance tests

eras/shelley/impl/testlib/Test/Cardano/Ledger/Shelley/ImpTest.hs
```diff
@@ -365,21 +370,29 @@ shelleyImpWitsVKeyNeeded nes txb = shelleyWitsVKeyNeeded utxo txb genDelegs
     utxo = nes ^. nesEsL . esLStateL . lsUTxOStateL . utxosUtxoL
     genDelegs = nes ^. nesEsL . esLStateL . lsCertStateL . certDStateL . dsGenDelegsL
 
-newtype ImpTestM era a = ImpTestM (ReaderT (IORef (ImpTestState era)) IO a)
+data ImpTestEnv era = ImpTestEnv
+  { iteSt :: !(IORef (ImpTestState era))
```
lehins: You not saving much space, but loosing clarity on what this field actually means
```suggestion
  { iteState :: !(IORef (ImpTestState era))
```

eras/shelley/impl/testlib/Test/Cardano/Ledger/Shelley/ImpTest.hs
```diff
@@ -752,3 +775,7 @@ constitutionShouldBe cUrl = do
           (error "No constitution has been set")
           constitution
   anchorUrl `shouldBe` fromJust (textToUrl $ T.pack cUrl)
+
+-- | Performs the action without running the fix-up function on any transactions
+withNoFixup :: ImpTestM era a -> ImpTestM era a
+withNoFixup = local (iteDoFixupL .~ False)
```
lehins: Now that is the correct way to deal with flags :+1:

libs/cardano-ledger-binary/CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## 1.3.0.0
 
+* Add `diffExprCompact`
 * Moved `ToExpr` instances out of the main library and into the testlib.
```
lehins: Just to clarify which component was affected
```suggestion
* Moved `ToExpr` instances out of the main library and into the testlib.

### `testlib`

* Add `diffExprCompact`
```

eras/conway/impl/CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## 1.12.0.0
 
+* Add `logTxSummary`, `byteStringToInteger`
```
lehins: As my comment below suggests, we need to move those into conformance test suite
```suggestion
```

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/ImpTest.hs
```diff
@@ -12,6 +13,12 @@
 {-# LANGUAGE UndecidableSuperClasses #-}
 {-# OPTIONS_GHC -Wno-orphans #-}
 
+#if __GLASGOW_HASKELL__ == 810
+-- GHC 8.10.7 complains about an incomplete pattern match on AlonzoScript even
+-- if all the cases are covered
+{-# OPTIONS_GHC -Wno-incomplete-patterns #-}
+#endif
+
```
lehins: It'll be easier to enable it for the whole cardano-ledger-conformance package
```suggestion
```

eras/shelley/impl/testlib/Test/Cardano/Ledger/Shelley/ImpTest.hs
```diff
@@ -365,21 +370,29 @@ shelleyImpWitsVKeyNeeded nes txb = shelleyWitsVKeyNeeded utxo txb genDelegs
     utxo = nes ^. nesEsL . esLStateL . lsUTxOStateL . utxosUtxoL
     genDelegs = nes ^. nesEsL . esLStateL . lsCertStateL . certDStateL . dsGenDelegsL
 
-newtype ImpTestM era a = ImpTestM (ReaderT (IORef (ImpTestState era)) IO a)
+data ImpTestEnv era = ImpTestEnv
+  { iteSt :: !(IORef (ImpTestState era))
+  , iteDoFixup :: !Bool
```
lehins: Could you add a little haddock describing the scope of the flag. Is it fixup only for transactions or will it be used for other potential fixups we are going to add (I suspect some might creep in)

Also if it is intended just for transactions (which would be my prefernce) I'd suggest renaming it to `iteDoTxFixup` or just `iteTxFixup`

libs/cardano-ledger-conformance/src/Test/Cardano/Ledger/Conformance.hs
```diff
@@ -426,3 +503,110 @@ instance
...
+            <> show e
+    initialState = Just . toExpr $ specToTestRep @(UTxOState Conway) agdaUtxoState
+  unless (finalAgdaState == finalImplState) $ do
+    logEntry $
+      "Changes in spec Utxo state:\n"
+        ++ diffExprCompact (toExpr initialState) (toExpr finalAgdaState)
+    logEntry $
+      "Changes in impl Utxo state:\n"
+        ++ diffExprCompact (toExpr initialState) (toExpr finalImplState)
+    error "The final states of spec and implementation do not match"
```
lehins:
```suggestion
    expectationFailure "The final states of spec and implementation do not match"
```

libs/cardano-ledger-conformance/src/Test/Cardano/Ledger/Conformance.hs
```diff
@@ -426,3 +503,110 @@ instance
...
+    finalImplState :: Maybe Agda.UTxOState
+    finalImplState = case submitRes of
+      Right _ -> case toTestRep newUtxoState of
+        Right x -> Just x
+        Left _ -> Nothing
+      Left e ->
+        error $
+          "Failed to submit transaction on the implementation side:\n"
+            <> show e
+    initialState = Just . toExpr $ specToTestRep @(UTxOState Conway) agdaUtxoState
```
lehins: Inside tests it is better to use appropriate failure functions (eg. [`assertFailure`](https://github.com/input-output-hk/cardano-ledger/blob/5e91d39b54caa78b3bad809473bb46dfbab373a1/libs/cardano-ledger-core/testlib/Test/Cardano/Ledger/Imp/Common.hs#L116) instead of `error`, because it produces nicer failure messages and failures are raised at deterministic locations:
```suggestion
    initialState = Just . toExpr $ specToTestRep @(UTxOState Conway) agdaUtxoState
  finalImplState :: Maybe Agda.UTxOState <- case submitRes of
      Right _ -> pure $ case toTestRep newUtxoState of
        Right x -> Just x
        Left _ -> Nothing
      Left e ->
        assertFailure $
          "Failed to submit transaction on the implementation side:\n"
            <> show e
```

libs/cardano-ledger-conformance/src/Test/Cardano/Ledger/Conformance.hs
```diff
@@ -426,3 +503,110 @@ instance
...
+  -- We already applied the fixup above, so no point doing it again
+  submitRes <- withNoFixup $ trySubmitTx tx
+  newUtxoState <- getsNES $ nesEsL . esLStateL . lsUTxOStateL
+  let
+    finalAgdaState :: Maybe Agda.UTxOState
+    finalAgdaState =
+      specToTestRep @(UTxOState Conway) <$> newAgdaUtxoState
+    finalImplState :: Maybe Agda.UTxOState
+    finalImplState = case submitRes of
+      Right _ -> case toTestRep newUtxoState of
```
lehins: Why are we translating `newUtxoState` instead of the one produced by `submitRes`?

libs/cardano-ledger-conformance/src/Test/Cardano/Ledger/Conformance.hs
```diff
@@ -426,3 +503,110 @@ instance
...
+      condition nes tx
+        ==> evalImpTestM impState
+        $ do
+          modify (impLastTickL @Conway .~ 100)
+          modifyNES $ const nes
+          logTxSummary tx
+          withNoFixup $ trySubmitTxConform_ tx
+
+txIsValid :: AlonzoEraTx era => p -> Tx era -> Bool
+txIsValid _ tx = (tx ^. isValidTxL) == IsValid True
```
lehins: It is more appropriate to use `const` or use a lambda to modify number of argument for a function, rather than adding bogus arguments that are ignored.
```suggestion
txIsValid :: AlonzoEraTx era => Tx era -> Bool
txIsValid tx = (tx ^. isValidTxL) == IsValid True
```

libs/cardano-ledger-conformance/src/Test/Cardano/Ledger/Conformance.hs
```diff
@@ -426,3 +503,110 @@ instance
...
+      certState <- getsNES $ nesEsL . esLStateL . lsCertStateL
+      govState <- getsNES $ nesEsL . esLStateL . lsUTxOStateL . utxosGovStateL
+      totalObligation certState govState `shouldBe` zero
+    do
+      deposited <- getsNES $ nesEsL . esLStateL . lsUTxOStateL . utxosDepositedL
+      deposited `shouldBe` zero
+    res <- trySubmitTxConform $ mkBasicTx mkBasicTxBody
+    expectRight_ res
+  describe "Constrained generator" $ do
+    it "UTXO conforms when IsValid True" $ conformsWhen txIsValid
```
lehins:
```suggestion
    it "UTXO conforms when IsValid True" $ conformsWhen (const txIsValid)
```

libs/cardano-ledger-core/testlib/Test/Cardano/Ledger/Imp/Common.hs
```diff
@@ -176,11 +179,15 @@ shouldNotReturn f a = withRunInIO $ \run -> H.shouldNotReturn (run f) a
...
 
--- | Return value on the `Right` an fail otherwise. Lifted version of `H.expectRight`.
+-- | Return value on the `Right` and fail otherwise. Lifted version of `H.expectRight`.
 expectRight :: (HasCallStack, Show a, MonadIO m) => Either a b -> m b
 expectRight (Right r) = pure $! r
 expectRight (Left l) = assertFailure $ "Expected Right, got Left:\n" <> show l
 
+-- | Same as `expectRight`, but discards the result
+expectRight_ :: (HasCallStack, Show a, MonadIO m) => Either a b -> m ()
+expectRight_ = void . expectRight
```
lehins: I think it only make sense to add `expectRightDeep_`, since otherwise it is possible to ignore a bottom somewhere in the Right case.

```suggestion
expectRightDeep_ :: (HasCallStack, Show a, NFData b, MonadIO m) => Either a b -> m ()
expectRightDeep_ = void . expectRightDeep
```

libs/cardano-ledger-core/testlib/Test/Cardano/Ledger/Imp/Common.hs
```diff
@@ -202,11 +209,15 @@ shouldBeRight e x = expectRight e >>= (`shouldBe` x)
...
 
--- | Return value on the `Left` an fail otherwise
+-- | Return value on the `Left` and fail otherwise
 expectLeft :: (HasCallStack, Show b, MonadIO m) => Either a b -> m a
 expectLeft (Left l) = pure $! l
 expectLeft (Right r) = assertFailure $ "Expected Left, got Right:\n" <> show r
 
+-- | Same as `expectLeft`, but discards the result
+expectLeft_ :: (HasCallStack, MonadIO m, Show b) => Either a b -> m ()
+expectLeft_ = void . expectLeft
```
lehins: Same comment as above. Change to `expectLeftDeep_`

Could you also add the same functions to `Test.Cardano.Ledger.Common` , except that'd be restricted to `IO`. This way we have two consistent interfaces: one for Imp spec and another for regular hspec

libs/cardano-ledger-conformance/src/Test/Cardano/Ledger/Conformance.hs
```diff
@@ -56,49 +66,96 @@ import Cardano.Ledger.Conway.PParams (
...
+  type TestRep a = SpecRep a
 
   toSpecRep :: a -> Either SpecTranslationError (SpecRep a)
 
+  specToTestRep :: SpecRep a -> TestRep a
+  default specToTestRep :: Coercible (SpecRep a) (TestRep a) => SpecRep a -> TestRep a
+  specToTestRep = coerce
+
+toTestRep :: forall a. SpecTranslate a => a -> Either SpecTranslationError (TestRep a)
+toTestRep x = specToTestRep @a <$> toSpecRep x
```
lehins: This function is wildly inefficient for Maps and Sets. You might wanna make it a type class function with default implementation that uses `coerce` instead of `specToTestRep`

If we don't do that we will end up with converting a ledger Map or a Set to a sorted list and the using a quadratic algorithm for sorting an already sorted list.

libs/cardano-ledger-conformance/src/Test/Cardano/Ledger/Conformance.hs
```diff
@@ -426,3 +503,110 @@ instance
...
+    gen = genTxAndNewEpoch us $ reify @Conway
+   in
+    forAllBlind gen $ \(nes, tx, _env) ->
+      condition nes tx
+        ==> evalImpTestM impState
+        $ do
+          modify (impLastTickL @Conway .~ 100)
+          modifyNES $ const nes
+          logTxSummary tx
+          withNoFixup $ trySubmitTxConform_ tx
```
lehins, line 598: The test is expected to pass without any fixup at all? How does that work?

lehins (changes requested):
This looks great. Few comments, hopefully uncontroversial.

libs/cardano-ledger-conformance/src/Test/Cardano/Ledger/Conformance.hs
Soupstraw: The constrained generator already takes care of balancing the transaction when it generates the NewEpochState and transaction.

libs/cardano-ledger-conformance/src/Test/Cardano/Ledger/Conformance.hs
Soupstraw: Good point

libs/cardano-ledger-conformance/src/Test/Cardano/Ledger/Conformance.hs
lehins: Ah, makes sense. Thanks

libs/cardano-ledger-conformance/src/Test/Cardano/Ledger/Conformance.hs
Soupstraw: `sumbitRes` only returns the `Either _ TxId`, which I have to check to make sure that the transaction was successfully submitted. `newUtxoState` is the actual `NewEpochState` that we want to compare once we're sure that the transaction was successful.

lehins: Oh yeah, sorry, you are grabbing the `newUtxoState` right after submitting the transaction and here you are just ignoring the TxId. Totally makes sense

Soupstraw: I rewrote this part in a more ImpTest-y way to get rid of the nested cases, so hopefully it's a bit more readable now.

lehins (approved):
This is an awesome progress for conformance tests! Great job!

PR #3854, opened 2023-11-07
Integration work for node 8.7.0

jasagredo:
# Description

* Updates `streaming-binary` dependency from `cardano-ledger-byron`
* Fixes the paths to the `cddl` data files.
* Uses `buildable = False` instead of `package.buildable = False` in Nix-Windows

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/eras/allegra/impl/cardano-ledger-allegra.cabal
+++ b/eras/allegra/impl/cardano-ledger-allegra.cabal
@@ -13,8 +13,8 @@ category:           Network
 build-type:         Simple
 data-files:
     cddl-files/allegra.cddl
-    cddl-files/real/crypto.cddl
-    cddl-files/mock/extras.cddl
+    cddl-files/crypto.cddl
+    cddl-files/extras.cddl
 
 extra-source-files: CHANGELOG.md
 
--- a/eras/byron/ledger/impl/cardano-ledger-byron.cabal
+++ b/eras/byron/ledger/impl/cardano-ledger-byron.cabal
@@ -1,6 +1,6 @@
 cabal-version:      3.0
 name:               cardano-ledger-byron
-version:            1.0.0.2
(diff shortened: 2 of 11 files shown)


eras/byron/ledger/impl/cardano-ledger-byron.cabal
```diff
@@ -252,7 +252,7 @@ library
         quiet,
         resourcet,
         streaming,
-        streaming-binary >=0.2 && <0.3,
+        streaming-binary >=0.2 && <0.4,
```
lehins, line 255: When updating bounds in cabal file, make sure to bump up the patch version of the package, otherwise it will not get released, so the new version should be: `1.0.0.3`

eras/allegra/impl/cardano-ledger-allegra.cabal
```diff
@@ -13,8 +13,8 @@ category:           Network
 build-type:         Simple
 data-files:
     cddl-files/allegra.cddl
-    cddl-files/real/crypto.cddl
-    cddl-files/mock/extras.cddl
+    cddl-files/crypto.cddl
+    cddl-files/extras.cddl
```
lehins, line 17: Thank you for catching this! I always forget updating those file paths in cabal file

eras/byron/ledger/impl/cardano-ledger-byron.cabal
jasagredo: Done!

lehins (approved):
Thank you! Looks great! :+1:

eras/allegra/impl/cardano-ledger-allegra.cabal
jasagredo: No worries!

PR #3855, opened 2023-11-07
Remove unreachable proposals

### LINKED ISSUE #3780: Remove invalid proposals after enactment
#3768 implements part of the goal for valid paths of current proposals, namely preventing proposlas with non-existent previous governance action ids of being proposed. However, the second part of this equation is to remove proposals from the governance state that can never be ratified, because a conflicting proposal have been enacted.

For example, this is a current tree of available proposals of PParamUpdatePurpose:

```
[A] ------ B ------ C ------ D
            \
             E ----- F ------ H
                      \
                        J ------ I
```

In this scenario `[A]` is the current latest proposal that have been enacted and if `B` gets enacted we are still left with all of the branches:

```
[B] ------ C ------ D
   \
     E ----- F ------ H
              \
               J ------ I
```

However, as soon as proposal `C` is enacted, none of the proposals: `E`, `F`, `H`, `J` or `I` can ever be enacted. Therefore there is no reason for them to be retained in the state and consume precious StakePool's memory and DRep's efforts of those proposals being considered and voted on.

For this reason, whenever `C` is enacted this should be the new governance state:
```
[C] ------ D
```

On the other hand if `E` is enacted instead, then `C` and `D` should be removed leaving us with this state:
```
[E] ----- F ------ H
           \
            J ------ I
```

aniketd:
# Description

Closes #3780

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/eras/conway/impl/src/Cardano/Ledger/Conway/Governance.hs
+++ b/eras/conway/impl/src/Cardano/Ledger/Conway/Governance.hs
@@ -35,6 +35,7 @@ module Cardano.Ledger.Conway.Governance (
   GovActionId (..),
   GovActionPurpose (..),
   PrevGovActionIds (..),
+  PrevGovActionIdsChildren (..),
   PrevGovActionId (..),
   DRepPulsingState (..),
   DRepPulser (..),
@@ -57,9 +58,9 @@ module Cardano.Ledger.Conway.Governance (
   isDRepVotingAllowed,
   isCommitteeVotingAllowed,
   Proposals,
-  proposalsInsertGovAction,
   proposalsActions,
   proposalsAddVote,
+  proposalsAddProposal,
   proposalsIds,
   proposalsRemoveIds,
   proposalsLookupId,
@@ -116,6 +117,11 @@ module Cardano.Ledger.Conway.Governance (
   psDRepDistrL,
   psDRepStateL,
   RunConwayRatify (..),
+  ensPrevGovActionIdsChildrenL,
+  pgacPParamUpdateL,
+  pgacHardForkL,
+  pgacCommitteeL,
+  pgacConstitutionL,
 
   -- * Exported for testing
   pparamsUpdateThreshold,
@@ -179,14 +185,20 @@ import Cardano.Ledger.Conway.Governance.Procedures (
   indexedGovProps,
  )
 import Cardano.Ledger.Conway.Governance.Proposals (
+  PrevGovActionIds (..),
+  PrevGovActionIdsChildren (..),
   Proposals,
   fromGovActionStateSeq,
   isConsistent_,
+  pgacCommitteeL,
+  pgacConstitutionL,
+  pgacHardForkL,
+  pgacPParamUpdateL,
   proposalsActions,
+  proposalsAddProposal,
   proposalsAddVote,
   proposalsGovActionStates,
   proposalsIds,
-  proposalsInsertGovAction,
   proposalsLookupId,
   proposalsRemoveIds,
  )
@@ -260,8 +272,6 @@ import Data.Typeable
 import GHC.Generics (Generic)
 import Lens.Micro
 import NoThunks.Class (NoThunks (..), allNoThunks)
-
--- ============================
 
 -- | A snapshot of information from the previous epoch stored inside the Pulser.
 --   After the pulser completes, but before the epoch turns, this information
@@ -338,55 +348,6 @@ instance EraPParams era => ToCBOR (PulsingSnapshot era) where
 instance EraPParams era => FromCBOR (PulsingSnapshot era) where
   fromCBOR = fromEraCBOR @era
 
--- ===============================
-
-data PrevGovActionIds era = PrevGovActionIds
-  { pgaPParamUpdate :: !(StrictMaybe (PrevGovActionId 'PParamUpdatePurpose (EraCrypto era)))
-  -- ^ The last enacted GovActionId for a protocol parameter update
-  , pgaHardFork :: !(StrictMaybe (PrevGovActionId 'HardForkPurpose (EraCrypto era)))
-  -- ^ The last enacted GovActionId for a hard fork
-  , pgaCommittee :: !(StrictMaybe (PrevGovActionId 'CommitteePurpose (EraCrypto era)))
-  -- ^ The last enacted GovActionId for a committee change or no confidence vote
-  , pgaConstitution :: !(StrictMaybe (PrevGovActionId 'ConstitutionPurpose (EraCrypto era)))
-  -- ^ The last enacted GovActionId for a new constitution
-  }
-  deriving (Eq, Show, Generic)
-
-instance NoThunks (PrevGovActionIds era)
-instance Era era => NFData (PrevGovActionIds era)
-instance Default (PrevGovActionIds era)
-
-instance Era era => DecCBOR (PrevGovActionIds era) where
-  decCBOR =
-    decode $
-      RecD PrevGovActionIds
-        <! From
-        <! From
-        <! From
-        <! From
-
-instance Era era => EncCBOR (PrevGovActionIds era) where
-  encCBOR PrevGovActionIds {..} =
-    encode $
-      Rec (PrevGovActionIds @era)
-        !> To pgaPParamUpdate
-        !> To pgaHardFork
-        !> To pgaCommittee
-        !> To pgaConstitution
-
-toPrevGovActionIdsPairs :: (KeyValue e a, Era era) => PrevGovActionIds era -> [a]
-toPrevGovActionIdsPairs pga@(PrevGovActionIds _ _ _ _) =
-  let PrevGovActionIds {..} = pga
-   in [ "pgaPParamUpdate" .= pgaPParamUpdate
-      , "pgaHardFork" .= pgaHardFork
-      , "pgaCommittee" .= pgaCommittee
-      , "pgaConstitution" .= pgaConstitution
-      ]
-
-instance Era era => ToJSON (PrevGovActionIds era) where
-  toJSON = object . toPrevGovActionIdsPairs
-  toEncoding = pairs . mconcat . toPrevGovActionIdsPairs
-
 data EnactState era = EnactState
   { ensCommittee :: !(StrictMaybe (Committee era))
   -- ^ Constitutional Committee
@@ -398,6 +359,7 @@ data EnactState era = EnactState
   , ensWithdrawals :: !(Map (Credential 'Staking (EraCrypto era)) Coin)
   , ensPrevGovActionIds :: !(PrevGovActionIds era)
   -- ^ Last enacted GovAction Ids
+  , ensPrevGovActionIdsChildren :: !(PrevGovActionIdsChildren era) -- TODO: @aniketd Move this inside Proposals
   }
   deriving (Generic)
 
@@ -424,6 +386,10 @@ ensWithdrawalsL = lens ensWithdrawals $ \es x -> es {ensWithdrawals = x}
 
 ensPrevGovActionIdsL :: Lens' (EnactState era) (PrevGovActionIds era)
 ensPrevGovActionIdsL = lens ensPrevGovActionIds (\es x -> es {ensPrevGovActionIds = x})
+
+ensPrevGovActionIdsChildrenL :: Lens' (EnactState era) (PrevGovActionIdsChildren era)
+ensPrevGovActionIdsChildrenL =
+  lens ensPrevGovActionIdsChildren (\es x -> es {ensPrevGovActionIdsChildren = x})
 
 ensPrevPParamUpdateL ::
   Lens' (EnactState era) (StrictMaybe (PrevGovActionId 'PParamUpdatePurpose (EraCrypto era)))
@@ -458,13 +424,14 @@ instance EraPParams era => ToJSON (EnactState era) where
   toEncoding = pairs . mconcat . toEnactStatePairs
 
 toEnactStatePairs :: (KeyValue e a, EraPParams era) => EnactState era -> [a]
-toEnactStatePairs cg@(EnactState _ _ _ _ _ _ _) =
+toEnactStatePairs cg@(EnactState _ _ _ _ _ _ _ _) =
   let EnactState {..} = cg
    in [ "committee" .= ensCommittee
       , "constitution" .= ensConstitution
       , "curPParams" .= ensCurPParams
       , "prevPParams" .= ensPrevPParams
       , "prevGovActionIds" .= ensPrevGovActionIds
+      , "prevGovActionIdsChilren" .= ensPrevGovActionIdsChildren
       ]
 
 deriving instance Eq (PParams era) => Eq (EnactState era)
@@ -481,6 +448,7 @@ instance EraPParams era => Default (EnactState era) where
       (Coin 0)
       def
       def
+      def
 
 instance EraPParams era => DecCBOR (EnactState era) where
   decCBOR = decNoShareCBOR
@@ -490,6 +458,7 @@ instance EraPParams era => DecShareCBOR (EnactState era) where
   decShareCBOR _ =
     decode $
       RecD EnactState
+        <! From
         <! From
         <! From
         <! From
@@ -509,6 +478,7 @@ instance EraPParams era => EncCBOR (EnactState era) where
         !> To ensTreasury
         !> To ensWithdrawals
         !> To ensPrevGovActionIds
+        !> To ensPrevGovActionIdsChildren
 
 instance EraPParams era => ToCBOR (EnactState era) where
   toCBOR = toEraCBOR @era
@@ -525,6 +495,7 @@ instance EraPParams era => NoThunks (EnactState era)
 data RatifyState era = RatifyState
   { rsEnactState :: !(EnactState era)
   , rsRemoved :: !(Set (GovActionId (EraCrypto era)))
+  , rsEnacted :: !(Set (GovActionId (EraCrypto era)))
   , rsDelayed :: !Bool
   }
   deriving (Generic)
@@ -547,10 +518,11 @@ instance EraPParams era => ToJSON (RatifyState era) where
   toEncoding = pairs . mconcat . toRatifyStatePairs
 
 toRatifyStatePairs :: (KeyValue e a, EraPParams era) => RatifyState era -> [a]
-toRatifyStatePairs cg@(RatifyState _ _ _) =
+toRatifyStatePairs cg@(RatifyState _ _ _ _) =
   let RatifyState {..} = cg
    in [ "nextEnactState" .= rsEnactState
       , "removedGovActions" .= rsRemoved
+      , "enactedGovActions" .= rsEnacted
       , "ratificationDelayed" .= rsDelayed
       ]
 
@@ -1011,7 +983,7 @@ class
             error (unlines ("Impossible: RATIFY rule never fails, but it did:" : map show ps))
           Right ratifyState' -> ratifyState'
 
-finishDRepPulser :: forall era. DRepPulsingState era -> (PulsingSnapshot era, RatifyState era)
+finishDRepPulser :: DRepPulsingState era -> (PulsingSnapshot era, RatifyState era)
 finishDRepPulser (DRComplete snap ratifyState) = (snap, ratifyState)
 finishDRepPulser (DRPulsing (DRepPulser {..})) =
   (PulsingSnapshot dpProposals finalDRepDistr dpDRepState, ratifyState')
@@ -1030,6 +1002,7 @@ finishDRepPulser (DRPulsing (DRepPulser {..})) =
     !ratifyState =
       RatifyState
         { rsRemoved = mempty
+        , rsEnacted = mempty
         , rsEnactState = dpEnactState
         , rsDelayed = False
         }
@@ -1212,22 +1185,24 @@ instance Era era => NFData (RatifyEnv era) where
               rnf cs
 
 instance EraPParams era => EncCBOR (RatifyState era) where
-  encCBOR (RatifyState es remove del) =
+  encCBOR (RatifyState es removed enacted delayed) =
     encode
       ( Rec (RatifyState @era)
           !> To es
-          !> To remove
-          !> To del
+          !> To removed
+          !> To enacted
+          !> To delayed
       )
 
 instance EraPParams era => DecCBOR (RatifyState era) where
-  decCBOR = decode (RecD RatifyState <! From <! From <! From)
+  decCBOR = decode (RecD RatifyState <! From <! From <! From <! From)
 
 -- TODO: Implement Sharing: https://github.com/input-output-hk/cardano-ledger/issues/3486
 instance EraPParams era => DecShareCBOR (RatifyState era) where
   decShareCBOR _ =
     decode $
       RecD RatifyState
+        <! From
         <! From
         <! From
         <! From
--- a/eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Procedures.hs
+++ b/eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Procedures.hs
@@ -49,6 +49,8 @@ module Cardano.Ledger.Conway.Governance.Procedures (
   govProceduresProposalsL,
   pProcGovActionL,
   gasActionL,
+  gasChildrenL,
+  gasIdL,
 ) where
 
 import Cardano.Crypto.Hash (hashToTextAsHex)
@@ -186,13 +188,18 @@ data GovActionState era = GovActionState
   , gasAction :: !(GovAction era)
   , gasProposedIn :: !EpochNo
   , gasExpiresAfter :: !EpochNo
+  , gasChildren :: Set (GovActionId (EraCrypto era))
   }
   deriving (Generic)
+
+gasIdL :: Lens' (GovActionState era) (GovActionId (EraCrypto era))
+gasIdL = lens gasId $ \x y -> x {gasId = y}
 
 gasDepositL :: Lens' (GovActionState era) Coin
 gasDepositL = lens gasDeposit (\x y -> x {gasDeposit = y})
 
-gasCommitteeVotesL :: Lens' (GovActionState era) (Map (Credential 'HotCommitteeRole (EraCrypto era)) Vote)
+gasCommitteeVotesL ::
+  Lens' (GovActionState era) (Map (Credential 'HotCommitteeRole (EraCrypto era)) Vote)
 gasCommitteeVotesL = lens gasCommitteeVotes (\x y -> x {gasCommitteeVotes = y})
 
 gasDRepVotesL :: Lens' (GovActionState era) (Map (Credential 'DRepRole (EraCrypto era)) Vote)
@@ -207,12 +214,15 @@ gasExpiresAfterL = lens gasExpiresAfter $ \x y -> x {gasExpiresAfter = y}
 gasActionL :: Lens' (GovActionState era) (GovAction era)
 gasActionL = lens gasAction $ \x y -> x {gasAction = y}
 
+gasChildrenL :: Lens' (GovActionState era) (Set (GovActionId (EraCrypto era)))
+gasChildrenL = lens gasChildren $ \x y -> x {gasChildren = y}
+
 instance EraPParams era => ToJSON (GovActionState era) where
   toJSON = object . toGovActionStatePairs
   toEncoding = pairs . mconcat . toGovActionStatePairs
 
 toGovActionStatePairs :: (KeyValue e a, EraPParams era) => GovActionState era -> [a]
-toGovActionStatePairs gas@(GovActionState _ _ _ _ _ _ _ _ _) =
+toGovActionStatePairs gas@(GovActionState _ _ _ _ _ _ _ _ _ _) =
   let GovActionState {..} = gas
    in [ "actionId" .= gasId
       , "committeeVotes" .= gasCommitteeVotes
@@ -223,6 +233,7 @@ toGovActionStatePairs gas@(GovActionState _ _ _ _ _ _ _ _ _) =
       , "action" .= gasAction
       , "proposedIn" .= gasProposedIn
       , "expiresAfter" .= gasExpiresAfter
+      , "children" .= gasChildren
       ]
 
 deriving instance EraPParams era => Eq (GovActionState era)
@@ -238,6 +249,7 @@ instance EraPParams era => DecShareCBOR (GovActionState era) where
   decShareCBOR _ =
     decode $
       RecD GovActionState
+        <! From
         <! From
         <! From
         <! From
@@ -264,6 +276,7 @@ instance EraPParams era => EncCBOR (GovActionState era) where
         !> To gasAction
         !> To gasProposedIn
         !> To gasExpiresAfter
+        !> To gasChildren
 
 -- Ref: https://gitlab.haskell.org/ghc/ghc/-/issues/14046
 instance
@@ -526,8 +539,20 @@ data GovActionPurpose
   | ConstitutionPurpose
   deriving (Eq, Show)
 
-newtype PrevGovActionId (r :: GovActionPurpose) c = PrevGovActionId {unPrevGovActionId :: GovActionId c}
-  deriving (Eq, Show, Generic, EncCBOR, DecCBOR, NoThunks, NFData, ToJSON, Ord)
+newtype PrevGovActionId (r :: GovActionPurpose) c = PrevGovActionId
+  { unPrevGovActionId :: GovActionId c
+  }
+  deriving
+    ( Eq
+    , Show
+    , Generic
+    , EncCBOR
+    , DecCBOR
+    , NoThunks
+    , NFData
+    , ToJSON
+    , Ord
+    )
 
 type role PrevGovActionId nominal nominal
 
--- a/eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Proposals.hs
+++ b/eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Proposals.hs
@@ -1,46 +1,77 @@
+{-# LANGUAGE DataKinds #-}
 {-# LANGUAGE DeriveGeneric #-}
 {-# LANGUAGE DerivingStrategies #-}
 {-# LANGUAGE GADTs #-}
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE MultiParamTypeClasses #-}
+{-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE RankNTypes #-}
+{-# LANGUAGE RecordWildCards #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeApplications #-}
 {-# LANGUAGE UndecidableInstances #-}
 
 module Cardano.Ledger.Conway.Governance.Proposals (
   Proposals,
   proposalsIds,
   proposalsAddVote,
-  proposalsInsertGovAction,
+  proposalsAddProposal,
   proposalsActions,
   proposalsRemoveIds,
   proposalsLookupId,
   fromGovActionStateSeq,
   proposalsGovActionStates,
+  PrevGovActionIds (..),
+  PrevGovActionIdsChildren (..),
+  pgacPParamUpdateL,
+  pgacHardForkL,
+  pgacCommitteeL,
+  pgacConstitutionL,
+  proposalsRemoveDescendentIds,
   -- Testing
   isConsistent_,
 ) where
 
+import Cardano.Ledger.BaseTypes (StrictMaybe (..))
 import Cardano.Ledger.Binary (DecCBOR (..), DecShareCBOR (..), EncCBOR (..))
+import Cardano.Ledger.Binary.Coders (
+  Decode (..),
+  Encode (..),
+  decode,
+  encode,
+  (!>),
+  (<!),
+ )
 import Cardano.Ledger.Conway.Core (Era (..), EraPParams)
 import Cardano.Ledger.Conway.Governance.Procedures (
+  GovAction (..),
   GovActionId,
+  GovActionPurpose (..),
   GovActionState (..),
+  PrevGovActionId (PrevGovActionId),
   Vote,
   Voter (..),
+  gasActionL,
+  gasChildrenL,
   gasCommitteeVotesL,
   gasDRepVotesL,
+  gasIdL,
   gasStakePoolVotesL,
  )
 import Control.DeepSeq (NFData)
-import Data.Aeson (ToJSON)
+import Control.Monad (guard)
+import Data.Aeson (KeyValue, ToJSON (..), object, pairs, (.=))
 import Data.Default.Class (Default (..))
+import Data.Functor (($>))
 import qualified Data.Map as Map
 import Data.Map.Strict (Map)
 import qualified Data.OMap.Strict as OMap
 import Data.Sequence.Strict (StrictSeq (..))
 import Data.Set (Set)
+import qualified Data.Set as Set
 import GHC.Generics (Generic)
-import Lens.Micro (Lens', (%~))
+import Lens.Micro
 import NoThunks.Class (NoThunks)
 
 newtype Proposals era
@@ -67,15 +98,6 @@ instance EraPParams era => DecCBOR (Proposals era) where
 -- TODO: Implement Sharing: https://github.com/input-output-hk/cardano-ledger/issues/3486
 instance EraPParams era => DecShareCBOR (Proposals era) where
   decShareCBOR _ = fromGovActionStateSeq <$> decCBOR
-
--- | Insert a `GovActionState`, overwriting an entry of it if the
--- corresponding `GovActionId` already exists.
-proposalsInsertGovAction ::
-  GovActionState era ->
-  Proposals era ->
-  Proposals era
-proposalsInsertGovAction gas (Proposals omap) =
-  Proposals (omap OMap.||> gas)
 
 -- | Get the sequence of `GovActionState`s
 proposalsActions ::
@@ -127,6 +149,26 @@ proposalsRemoveIds gais (Proposals omap) =
   let (retained, removed) = OMap.extractKeys gais omap
    in (Proposals retained, removed)
 
+-- | Extract `GovActionState`s for all the descendents of a set of
+-- `GovActionId`s from the `Proposals`
+proposalsRemoveDescendentIds ::
+  Set (GovActionId (EraCrypto era)) ->
+  Proposals era ->
+  (Proposals era, Map (GovActionId (EraCrypto era)) (GovActionState era))
+proposalsRemoveDescendentIds gais ps@(Proposals omap) =
+  proposalsRemoveIds (gais <> foldMap (getAllDescendents omap) gais) ps
+
+getAllDescendents ::
+  OMap.OMap (GovActionId (EraCrypto era)) (GovActionState era) ->
+  GovActionId (EraCrypto era) ->
+  Set (GovActionId (EraCrypto era))
+getAllDescendents omap gaid =
+  case OMap.lookup gaid omap of
+    Nothing -> mempty
+    Just gas ->
+      let children = gasChildren gas
+       in children <> foldMap (getAllDescendents omap) children
+
 proposalsLookupId ::
   GovActionId (EraCrypto era) ->
   Proposals era ->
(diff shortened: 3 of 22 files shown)


eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Gov.hs
```diff
@@ -152,6 +160,7 @@ instance EraPParams era => DecCBOR (ConwayGovPredFailure era) where
     7 -> SumD ExpirationEpochTooSmall <! From
     8 -> SumD InvalidPrevGovActionIdsInProposals <! From
```
aniketd: I guess this constructor can be removed and replaced be the newer one, at the same position, instead of adding a new index to the serialization?

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Ledger.hs
```diff
@@ -295,14 +295,23 @@ ledgerTransition = do
...
+                  (utxoState ^. utxosGovStateL . cgEnactStateL . ensPrevGovActionIdsL)
+              , GovRuleState
+                  (utxoState ^. utxosGovStateL . proposalsGovStateL)
+                  (utxoState ^. utxosGovStateL . cgEnactStateL . ensPrevGovActionIdsChildrenL)
               , govProcedures
               )
-        let utxoState' = utxoState & utxosGovStateL . cgProposalsL .~ govActionsState'
+        let utxoState' =
+              utxoState
+                & utxosGovStateL . cgProposalsL .~ govActionsState'
```
aniketd: `cgProposalsL ~ proposalsGovStateL` so may be the latter is a better option to be used here.

eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Procedures.hs
```diff
@@ -541,10 +555,36 @@ data GovActionPurpose
...
+    , NoThunks
+    , NFData
+    , ToJSON
+    , ToExpr
+    , Ord
+    )
 
 type role PrevGovActionId nominal nominal
 
+newtype PrevGovActionChildren (r :: GovActionPurpose) c = PrevGovActionChildren {unPrevGovActionChildren :: Set (GovActionId c)}
```
lehins: Too long
```suggestion
newtype PrevGovActionChildren (r :: GovActionPurpose) c = 
  PrevGovActionChildren {unPrevGovActionChildren :: Set (GovActionId c)}
```

eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Snapshots.hs
```diff
@@ -121,6 +130,50 @@ snapshotAddVote voter vote gai (ProposalsSnapshot omap) =
...
+    ParameterChange prev _ ->
+      go prev gas proposals $
+        \case ParameterChange {} -> True; _ -> False
+    HardForkInitiation prev _ ->
+      go prev gas proposals $
+        \case HardForkInitiation {} -> True; _ -> False
+    TreasuryWithdrawals _ -> Just proposals
+    NoConfidence prev ->
+      go prev gas proposals $
+        \case NoConfidence {} -> True; UpdateCommittee {} -> True; _ -> False
```
lehins: You not saving much by writing them in one line. Try avoid using `;` in Haskell, it is non-idiomatic
```suggestion
      go prev gas proposals $ \case 
        NoConfidence {} -> True
        UpdateCommittee {} -> True
        _ -> False
```

eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Snapshots.hs
```diff
@@ -121,6 +130,50 @@ snapshotAddVote voter vote gai (ProposalsSnapshot omap) =
...
+        SJust (PrevGovActionId prev') ->
+          case OMap.lookup prev' omap' of
+            Nothing -> Nothing
+            Just found ->
+              if not $ cond $ found ^. gasActionL
+                then Nothing
+                else
+                  Just $
+                    ProposalsSnapshot $
+                      OMap.adjust (gasChildrenL %~ Set.insert (action ^. gasIdL)) prev' omap'
```
lehins: Extra negation adds to complexity and makes it just a little bit less readable. Also `Maybe` nesting can be nicely solved with `do` syntax:
```suggestion
    go mprev action (ProposalsSnapshot omap) cond = do
      SJust (PrevGovActionId prev) <- Just mprev
      prevProposal <- OMap.lookup prev omap
      guard (cond $ prevProposal ^. gasActionL)
      Just $ ProposalsSnapshot $
        OMap.adjust (gasChildrenL %~ Set.insert (action ^. gasIdL)) prev omap
```

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Gov.hs
```diff
@@ -110,6 +112,11 @@ data GovEnv era = GovEnv
   , gePrevGovActionIds :: !(PrevGovActionIds era)
   }
 
+data GovRuleState era = GovRuleState
+  { grsProposals :: ProposalsSnapshot era
+  , grsPrevGovActionIdsChildren :: PrevGovActionIdsChildren era
```
lehins: This is used as accumulator in a fold, you definitely want those fields to be strict
```suggestion
  { grsProposals :: !(ProposalsSnapshot era)
  , grsPrevGovActionIdsChildren :: !(PrevGovActionIdsChildren era)
```

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Gov.hs
```diff
@@ -271,65 +257,102 @@ actionWellFormed ga = failureUnless isWellFormed $ MalformedProposal ga
...
+    , gasChildren = mempty
+    }
+
+-- | Add a child to the already enacted actions. Returns a `Just
+-- (PrevGovActionIdsChildren era)` with the child added, or `Nothing` if
+-- the `PrevGovActionId` is not found.
+--
+-- NOTE:
+-- If both the already enacted action as well as the previous Id
+-- referenced by the action being addecd are `SNothing` the checks pass
```
lehins:
```suggestion
-- referenced by the action being added are `SNothing` then checks pass
```

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Gov.hs
```diff
@@ -351,18 +374,47 @@ govTransition = do
...
-                st'
-        applyProps st'' ps
-  stProps <- applyProps st $ indexedGovProps $ SSeq.fromStrict $ OSet.toStrictSeq $ gpProposalProcedures gp
+                expiry
+                currentEpoch
+         in case addToPrevGovActionIdsChildren
+              gai
+              pProcGovAction
+              prevGovActionIds
+              grsPrevGovActionIdsChildren of
```
lehins: Multi-line cases is horrendous, it is always better to add a new binding:
```suggestion
            prevGovActionIdsChildren =
              addToPrevGovActionIdsChildren
                gai
                pProcGovAction
                prevGovActionIds
                grsPrevGovActionIdsChildren
         in case prevGovActionIdsChildren of
```

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Gov.hs
```diff
@@ -351,18 +374,47 @@ govTransition = do
...
+              Just updatedChildren ->
+                pure $
+                  GovRuleState
+                    (snapshotInsertGovAction actionState grsProposals)
+                    updatedChildren
+              -- Didn't find a parent in the already enacted actions.
+              -- Searching in the already accepted proposals.
+              Nothing ->
+                case snapshotAddChild actionState grsProposals of
+                  -- If the parent in not found in the accepted proposals, the predicate fails
```
lehins:
```suggestion
                  -- If the parent is not found in the accepted proposals, the predicate fails
```

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Gov.hs
```diff
@@ -271,65 +257,102 @@ actionWellFormed ga = failureUnless isWellFormed $ MalformedProposal ga
...
-            UpdateCommittee prev _ _ _ ->
-              isValidPrevGovActionId prev $ \case NoConfidence {} -> True; UpdateCommittee {} -> True; _ -> False
-            NewConstitution prev _ ->
-              isValidPrevGovActionId prev $ \case NewConstitution {} -> True; _ -> False
-            InfoAction -> True
-      invalidProposals = Seq.filter (not . isValid) $ SSeq.fromStrict $ OSet.toStrictSeq $ procedures ^. govProceduresProposalsL
-   in failureUnless (Seq.null invalidProposals) $ InvalidPrevGovActionIdsInProposals invalidProposals
+  PrevGovActionIdsChildren era ->
+  Maybe (PrevGovActionIdsChildren era)
+addToPrevGovActionIdsChildren actionId action PrevGovActionIds {..} children =
```
lehins: This function resembles this one: https://github.com/input-output-hk/cardano-ledger/blob/2f29af846d743d1c79e77ab52c2142ec1a127063/eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Ratify.hs#L343

I wonder if we can reduce duplication and extract a common part between the two and reuse it somehow in both places: here and in RATIFY rule

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Gov.hs
```diff
@@ -132,6 +139,7 @@ data ConwayGovPredFailure era
       (Map.Map (Credential 'ColdCommitteeRole (EraCrypto era)) EpochNo)
   | InvalidPrevGovActionIdsInProposals (Seq.Seq (ProposalProcedure era))
   | VotingOnExpiredGovAction (Map.Map (GovActionId (EraCrypto era)) (Voter (EraCrypto era)))
+  | InvalidPrevGovActionId (ProposalProcedure era)
```
lehins: You should be replacing the old predicate failure instead. We haven't deployed Conway yet, so we are still free to change predicate failures.
```suggestion
  | InvalidPrevGovActionId (ProposalProcedure era)
  | VotingOnExpiredGovAction (Map.Map (GovActionId (EraCrypto era)) (Voter (EraCrypto era)))
```

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Arbitrary.hs
```diff
@@ -210,6 +222,7 @@ genGovActionStateFromAction act =
     <*> pure act
     <*> arbitrary
     <*> arbitrary
+    <*> pure Set.empty
```
lehins: Why is set to empty? We want some garbage in there
```suggestion
    <*> arbitrary
```

libs/cardano-ledger-pretty/src/Cardano/Ledger/Pretty/Conway.hs
```diff
@@ -396,7 +400,7 @@ instance PrettyA (Anchor era) where
   prettyA = viaShow
 
 instance PrettyA (PParamsUpdate era) => PrettyA (GovActionState era) where
-  prettyA gas@(GovActionState _ _ _ _ _ _ _ _ _) =
+  prettyA gas@(GovActionState _ _ _ _ _ _ _ _ _ _) =
```
lehins, line 446: The point of this pattern is so you don't forget to update the implementation as well :smile: 

You need to a pretty  field below

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/TypeRep.hs
```diff
@@ -315,6 +315,10 @@ data Rep era t where
   NumDormantEpochsR :: Era era => Rep era EpochNo
   CommitteeStateR :: Era era => Rep era (CommitteeState era)
   VStateR :: Era era => Rep era (VState era)
+  PrevPParamUpdateChildrenR :: Era era => Rep era (GovActionId (EraCrypto era)) -> Rep era (Set (GovActionId (EraCrypto era)))
+  PrevHardForkChildrenR :: Era era => Rep era (GovActionId (EraCrypto era)) -> Rep era (Set (GovActionId (EraCrypto era)))
+  PrevCommitteeChildrenR :: Era era => Rep era (GovActionId (EraCrypto era)) -> Rep era (Set (GovActionId (EraCrypto era)))
+  PrevConstitutionChildrenR :: Era era => Rep era (GovActionId (EraCrypto era)) -> Rep era (Set (GovActionId (EraCrypto era)))
```
lehins: You'll need to ask @TimSheard about correctness of this, I don't have the full grasp on this framework yet

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/PrettyCore.hs
```diff
@@ -1712,7 +1728,7 @@ pcProposalsSnapshot :: ProposalsSnapshot era -> PDoc
 pcProposalsSnapshot x = ppSexp "ProposalsSnapShot" (map pcGovActionState (toList (snapshotActions x)))
 
 pcGovActionState :: GovActionState era -> PDoc
-pcGovActionState gas@(GovActionState _ _ _ _ _ _ _ _ _) =
+pcGovActionState gas@(GovActionState _ _ _ _ _ _ _ _ _ _) =
```
lehins, line 1779: Same here. I wonder why are we duplicating all stuff in `pretty` package and here as well.

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Gov.hs
```diff
@@ -271,65 +257,102 @@ actionWellFormed ga = failureUnless isWellFormed $ MalformedProposal ga
...
-  let applyProps st' Seq.Empty = pure st'
-      applyProps st' ((idx, ProposalProcedure {..}) Seq.:<| ps) = do
-        let expectedDeposit = pp ^. ppGovActionDepositL
-         in pProcDeposit
-              == expectedDeposit
-                ?! ProposalDepositIncorrect pProcDeposit expectedDeposit
+  expectedNetworkId <- liftSTS $ asks networkId
 
+  let processProposal grs@GovRuleState {..} (ix, proposal@ProposalProcedure {..}) = do
+        -- PParam updates well-formedness check
```
lehins:
```suggestion
        -- PParamsUpdate well-formedness check
```

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Gov.hs
```diff
@@ -351,18 +374,47 @@ govTransition = do
...
+                pure $
+                  GovRuleState
+                    (snapshotInsertGovAction actionState grsProposals)
+                    updatedChildren
+              -- Didn't find a parent in the already enacted actions.
+              -- Searching in the already accepted proposals.
+              Nothing ->
+                case snapshotAddChild actionState grsProposals of
+                  -- If the parent in not found in the accepted proposals, the predicate fails
+                  Nothing -> failBecause (InvalidPrevGovActionId proposal) >> pure grs
```
lehins:
```suggestion
                  Nothing -> grs <$ failBecause (InvalidPrevGovActionId proposal)
```

lehins (commented):
This looks great. You are certainly on the right track!

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Gov.hs
lehins: Yes. Just saw your comment. I left the same comment :slightly_smiling_face:

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Ledger.hs
lehins: Sure, if you like. They are one and the same thing fro Conway:

```
proposalsGovStateL = cgProposalsL
```

We don't really need to future proof the rules.

eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Snapshots.hs
aniketd: TIL about using monadic pattern matching trick in this way! Awesome!

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Gov.hs
aniketd: I will surely look into this next. :+1:

eras/conway/impl/src/Cardano/Ledger/Conway/Governance.hs
```diff
@@ -339,7 +349,79 @@ instance EraPParams era => ToCBOR (PulsingSnapshot era) where
 instance EraPParams era => FromCBOR (PulsingSnapshot era) where
   fromCBOR = fromEraCBOR @era
 
--- ===============================
+pgacPParamUpdateL ::
+  Lens' (PrevGovActionIdsChildren era) (Set (GovActionId (EraCrypto era)))
+pgacPParamUpdateL =
+  lens
+    (unPrevGovActionChildren . pgacPParamUpdate)
+    $ \x y -> x {pgacPParamUpdate = PrevGovActionChildren y}
```
aniketd: Is it really okay to make lenses like these that help us make changes easily but bypass the type safety that has been carefully put together for the data-type?

aniketd: Should these lenses be broken into two levels so that they can be used as a composition at their use sites?

eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Procedures.hs
aniketd: Fourmolu just undid the line break, sadly :slightly_frowning_face:

eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Procedures.hs
```diff
@@ -627,3 +668,32 @@ instance EraPParams era => EncCBOR (GovAction era) where
         Sum NewConstitution 5 !> E (encodeNullStrictMaybe encCBOR) gid !> To c
       InfoAction ->
         Sum InfoAction 6
+
+govActionParentL :: Lens' (GovAction era) (Maybe (GovActionId (EraCrypto era)))
```
aniketd: Same question about the lenses here. Should this lens be split into two? One preserving the type-level purpose and the other from any type-level purpose to the result?

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/PrettyCore.hs
```diff
@@ -1665,6 +1671,16 @@ pcPrevGovActionIds PrevGovActionIds {..} =
     , ("LastConstitution", ppStrictMaybe pcPrevGovActionId pgaConstitution)
     ]
 
+pcPrevGovActionIdsChildren :: PrevGovActionIdsChildren era -> PDoc
+pcPrevGovActionIdsChildren PrevGovActionIdsChildren {..} =
+  ppRecord
+    "PrevGovActionIds"
```
aniketd: TODO: Fix the labels.

eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Procedures.hs
lehins: Then do this:

```suggestion
newtype PrevGovActionChildren (r :: GovActionPurpose) c = PrevGovActionChildren
  { unPrevGovActionChildren :: Set (PrevGovActionId r c)
  }
```

eras/conway/impl/src/Cardano/Ledger/Conway/Governance.hs
lehins: It is no different from using the actual record field names, they are exported and allow us to modify this type in very much the same way as the lenses. I don't see any problem here

lehins: Oh, I see what's going on.By doing this we erase the type safety. Then no, this is not good.

eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Procedures.hs
lehins: Also. Make sure to use the `PrevGovActionId` newtype! Otherwise you don't get much type safety

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Gov.hs
```diff
@@ -276,65 +258,102 @@ actionWellFormed ga = failureUnless isWellFormed $ MalformedProposal ga
...
+      StrictMaybe (PrevGovActionId purpose (EraCrypto era)) ->
+      StrictMaybe (PrevGovActionId purpose (EraCrypto era)) ->
+      Lens' (PrevGovActionIdsChildren era) (Set.Set (GovActionId (EraCrypto era))) ->
+      GovActionId (EraCrypto era) ->
+      PrevGovActionIdsChildren era ->
+      Maybe (PrevGovActionIdsChildren era)
+    go pgai prev lens' actionId' children' =
+      if pgai == prev
+        then Just $ children' & lens' %~ Set.insert actionId'
+        else Nothing
```
lehins: The type signature for `go` takes more space then the function actually saves. Also it forces you to use this lens that bypasses the designed type safety: 
```
Lens' (PrevGovActionIdsChildren era) (Set.Set (GovActionId (EraCrypto era)))
```

```suggestion
  case action of
    ParameterChange prev _ -> do
      guard (pgaPParamUpdate == prev)
      let prevGovActionId :: PrevGovActionId 'PParamUpdatePurpose c
          prevGovActionId = PrevGovActionId actionId
      Just $ children & pgacPParamUpdateL %~ insertPrevActionId prevGovActionId
    HardForkInitiation prev _ -> do
      guard (pgaHardFork == prev)
      let prevGovActionId :: PrevGovActionId 'HardForkPurpose c
          prevGovActionId = PrevGovActionId actionId
      Just $ children & pgacHardForkL %~ insertPrevActionId prevGovActionId
    TreasuryWithdrawals _ -> Just children
    NoConfidence prev -> do
      guard (pgaCommittee == prev)
      let prevGovActionId :: PrevGovActionId 'CommitteePurpose c
          prevGovActionId = PrevGovActionId actionId
      Just $ children & pgacCommitteeL %~ insertPrevActionId prevGovActionId
    UpdateCommittee prev _ _ _ -> do
      guard (pgaCommittee == prev)
      let prevGovActionId :: PrevGovActionId 'CommitteePurpose c
          prevGovActionId = PrevGovActionId actionId
      Just $ children & pgacCommitteeL %~ insertPrevActionId prevGovActionId
    NewConstitution prev _ -> do
      guard (pgaConstitution == prev)
      let prevGovActionId :: PrevGovActionId 'ConstitutionPurpose c
          prevGovActionId = PrevGovActionId actionId
      Just $ children & pgacConstitutionL %~ insertPrevActionId prevGovActionId
    InfoAction -> Just children
  where
    insertPrevActionId :: 
      PrevGovActionChildren r c -> PrevGovActionId r c -> PrevGovActionChildren r c 
    insertPrevActionId (PrevGovActionChildren m) prevGovActionId = 
      PrevGovActionChildren (Set.insert prevGovActionId m)
```

eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Procedures.hs
lehins: No, this kind of lenses should not exist. Create a getter function or a `SimpleGetter` that extracts it, that will be safe.

lehins (changes requested):
Some feedback on type safety

eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Procedures.hs
aniketd: Nice!

eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Procedures.hs
aniketd: Your suggestions have significantly simplified everything and we don't need this lens now. It was only needed in the complicated and misunderstood earlier implementation that changed `Ratify`.

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Gov.hs
aniketd: After implementing the newtype wrapper preserving type-safely, this turned out to be needing very little change. We don't need the type signature anymore. And the implementation is shorter with the use of `go`. But I will be happy to switch to your suggested way of implementing this if you still prefer it, let me know :heart:

eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Procedures.hs
```diff
@@ -541,10 +555,36 @@ data GovActionPurpose
...
+    , NoThunks
+    , NFData
+    , ToJSON
+    , ToExpr
+    , Ord
+    )
 
 type role PrevGovActionId nominal nominal
 
+newtype PrevGovActionChildren (r :: GovActionPurpose) c = PrevGovActionChildren {unPrevGovActionChildren :: Set (PrevGovActionId r c)}
```
lehins: Line is too long
```suggestion
newtype PrevGovActionChildren (r :: GovActionPurpose) c = PrevGovActionChildren
  { unPrevGovActionChildren :: Set (PrevGovActionId r c)
  }
```

eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Procedures.hs
aniketd: I'm trying more things to make `fourmolu` comply.

lehins: I made the suggestion that makes fourmolu comply, why doesn't it work for you?

eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Procedures.hs
aniketd: It works! I made a mistake earlier while applying your suggestion. Thanks a ton @lehins !

eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Proposals.hs
```diff
@@ -130,6 +180,27 @@ proposalsRemoveIds gais (Proposals omap) =
...
 
+-- | Extract `GovActionState`s for all the descendents of a set of
+-- `GovActionId`s from the `Proposals`
+proposalsRemoveDescendentIds ::
+  Foldable t =>
+  t (GovActionId (EraCrypto era)) ->
+  Proposals era ->
+  (Proposals era, Map (GovActionId (EraCrypto era)) (GovActionState era))
+proposalsRemoveDescendentIds gais ps@(Proposals omap) =
+  proposalsRemoveIds (foldMap (getAllDescendents omap) gais) ps
```
lehins: We want to remove the supplied ids as well
```suggestion
proposalsRemoveWithDescendentIds ::
  Set (GovActionId (EraCrypto era)) ->
  Proposals era ->
  (Proposals era, Map (GovActionId (EraCrypto era)) (GovActionState era))
proposalsRemoveWithDescendentIds gais ps@(Proposals omap) =
  proposalsRemoveIds (gais <> foldMap (getAllDescendents omap) gais) ps
```

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Epoch.hs
```diff
@@ -279,22 +279,26 @@ epochTransition = do
...
 
     (accountState2, dState2, newEnactState) =
       applyEnactedWithdrawals accountState1 dState1 rsEnactState
-    (newProposals, removedGovActions) =
-      -- It is important that we use current proposals here instead of the ones from the pulser
-      proposalsRemoveIds rsRemoved (govState0 ^. cgProposalsL)
+
+    -- It is important that we use current proposals here instead of the ones from the pulser
+    (newProposals', removedGovActions') = proposalsRemoveDescendentIds rsRemoved (govState0 ^. cgProposalsL)
+    (newProposals, removedGovActions) = proposalsRemoveIds rsRemoved newProposals'
```
lehins: You also need to remove enacted! Also my previous suggestions were done during the meeting, that's why I used ticks. But for consistency with the rest of the rule we should avoid them. Also `newProposals` is a misleading name, as it was mentioned during the meeting too:
```suggestion
    (proposals0, removedInvalidGovActions) = 
      proposalsRemoveWithDescendentIds rsRemoved (govState0 ^. cgProposalsL)
    (proposals1, removedEnactedGovActions) = proposalsRemoveIds rsEnacted proposals0
```

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Epoch.hs
```diff
@@ -279,22 +279,26 @@ epochTransition = do
...
+
     govState1 =
       govState0
         & cgProposalsL .~ newProposals
         & cgEnactStateL .~ (newEnactState & ensPrevPParamsL .~ curPParams)
 
+    allRemoved = removedGovActions' <> removedGovActions
```
lehins:
```suggestion
    allRemovedGovActions = removedInvalidGovActions <> removedEnactedGovActions
```

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Epoch.hs
```diff
@@ -279,22 +279,26 @@ epochTransition = do
...
+
     certState =
       CertState
         { certPState = pState2
-        , certDState = returnProposalDeposits removedGovActions dState2
+        , certDState = returnProposalDeposits (allRemoved <> rsEnacted) dState2
```
lehins:
```suggestion
        , certDState = 
          returnProposalDeposits allRemovedGovActions dState2
```

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/TypeRep.hs
aniketd: @TimSheard, am I doing the right thing here in `cardano-ledger-test`?

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/PrettyCore.hs
aniketd: @TimSheard I noticed that we have functions prefixed `pp` in the pretty package and those prefixed `pc` here under the `Generic/PrettyCore` module. I did spend some time earlier on, on trying to use the `PrettyA` instance from the pretty package here but it led to a rabbit hole I didn't want to go down at that time. I'm going to have a fresh look at it now.

aniketd: I had a discussion with @TimSheard and learned that the functions prefixed `pc` are different in that they are for debugging focussed pretty-printing.

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Gov.hs
```diff
@@ -276,65 +257,95 @@ actionWellFormed ga = failureUnless isWellFormed $ MalformedProposal ga
...
-  let applyProps st' Seq.Empty = pure st'
-      applyProps st' ((idx, ProposalProcedure {..}) Seq.:<| ps) = do
-        let expectedDeposit = pp ^. ppGovActionDepositL
-         in pProcDeposit
-              == expectedDeposit
-                ?! ProposalDepositIncorrect pProcDeposit expectedDeposit
+  expectedNetworkId <- liftSTS $ asks networkId
 
+  let processProposal grs@GovRuleState {..} (ix, proposal@ProposalProcedure {..}) = do
+        -- PParamUpdate well-formedness check
```
lehins:
```suggestion
        -- PParamsUpdate well-formedness check
```

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/TypeRep.hs
aniketd: @TimSheard 's suggestions have been implemented.

lehins (changes requested):
Few more things need to be done. But it is looking good :)

eras/conway/impl/src/Cardano/Ledger/Conway/Governance.hs
```diff
@@ -342,7 +358,83 @@ instance EraPParams era => ToCBOR (PulsingSnapshot era) where
...
+
+pgacConstitutionL ::
+  Lens' (PrevGovActionIdsChildren era) (Set (PrevGovActionId 'ConstitutionPurpose (EraCrypto era)))
+pgacConstitutionL = pgacConstitutionL' . prevGovActionChildrenL
+
+data PrevGovActionIdsChildren era = PrevGovActionIdsChildren
+  { pgacPParamUpdate :: PrevGovActionChildren 'PParamUpdatePurpose (EraCrypto era)
+  , pgacHardFork :: PrevGovActionChildren 'HardForkPurpose (EraCrypto era)
+  , pgacCommittee :: PrevGovActionChildren 'CommitteePurpose (EraCrypto era)
+  , pgacConstitution :: PrevGovActionChildren 'ConstitutionPurpose (EraCrypto era)
```
lehins: Let's get rid of the `newtype PrevGovActionChildren` completely. It does not bring any more type safety or useful functionality, only redundant lenses and wrappers and unwrappers.
Also strict fields in the data type:

```suggestion
  { pgacPParamUpdate :: !(Set (PrevGovActionId 'PParamUpdatePurpose (EraCrypto era)))
  , pgacHardFork :: !(Set (PrevGovActionId 'HardForkPurpose (EraCrypto era)))
  , pgacCommittee :: !(Set (PrevGovActionId 'CommitteePurpose (EraCrypto era)))
  , pgacConstitution :: !(Set (PrevGovActionId 'ConstitutionPurpose (EraCrypto era)))
```

eras/conway/impl/src/Cardano/Ledger/Conway/Governance.hs
```diff
@@ -116,6 +120,15 @@ module Cardano.Ledger.Conway.Governance (
...
   RunConwayRatify (..),
+  ensPrevGovActionIdsChildrenL,
+  pgacPParamUpdateL,
+  pgacHardForkL,
+  pgacCommitteeL,
+  pgacConstitutionL,
+  pgacPParamUpdateL',
+  pgacHardForkL',
+  pgacCommitteeL',
+  pgacConstitutionL',
```
lehins: Removal of `PrevGovActionIdsChildren` will make these redundant
```suggestion
```

eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Procedures.hs
```diff
@@ -544,11 +559,44 @@ data GovActionPurpose
...
+    , NFData
+    , ToJSON
+    , ToExpr
+    , Ord
+    , Default
+    )
+
+prevGovActionChildrenL :: Lens' (PrevGovActionChildren p c) (Set (PrevGovActionId p c))
+prevGovActionChildrenL = lens unPrevGovActionChildren $ \_ y -> PrevGovActionChildren y
+
```
lehins: Redundant
```suggestion
```

eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Proposals.hs
```diff
@@ -121,6 +132,45 @@ proposalsAddVote voter vote gai (Proposals omap) =
...
+        _ -> False
+    InfoAction -> Just proposals
+  where
+    go mprev action (Proposals omap) cond = do
+      SJust (PrevGovActionId prev) <- Just mprev
+      parent <- OMap.lookup prev omap
+      guard $ cond $ parent ^. gasActionL
+      Just $
+        Proposals $
+          OMap.adjust (gasChildrenL %~ Set.insert (action ^. gasIdL)) prev omap
```
lehins: Avoid passing the arguments to local functions, if they already have those arguments in scope anyways:

```suggestion
proposalsAddChild gas proposals@(Proposals omap) =
  case gas ^. gasActionL of
    ParameterChange prev _ ->
      go prev $ \case
        ParameterChange {} -> True
        _ -> False
    HardForkInitiation prev _ ->
      go prev $ \case
        HardForkInitiation {} -> True
        _ -> False
    TreasuryWithdrawals _ -> Just proposals
    NoConfidence prev ->
      go prev $ \case
        NoConfidence {} -> True
        UpdateCommittee {} -> True
        _ -> False
    UpdateCommittee prev _ _ _ ->
      go prev $ \case
        NoConfidence {} -> True
        UpdateCommittee {} -> True
        _ -> False
    NewConstitution prev _ ->
      go prev $ \case
        NewConstitution {} -> True
        _ -> False
    InfoAction -> Just proposals
  where
    go mprev cond = do
      SJust (PrevGovActionId prev) <- Just mprev
      parent <- OMap.lookup prev omap
      guard $ cond $ parent ^. gasActionL
      Just $
        Proposals $
          OMap.adjust (gasChildrenL %~ Set.insert (gas ^. gasIdL)) prev omap
```

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Gov.hs
```diff
@@ -276,65 +257,95 @@ actionWellFormed ga = failureUnless isWellFormed $ MalformedProposal ga
...
+    , gasReturnAddr = returnAddress
+    , gasAction = action
+    , gasProposedIn = curEpoch
+    , gasExpiresAfter = curEpoch + expiry
+    , gasChildren = mempty
+    }
+
+-- | Add a child to the already enacted actions. Returns a `Just
+-- (PrevGovActionIdsChildren era)` with the child added, or `Nothing` if
+-- the `PrevGovActionId` is not found.
```
lehins: Backticks are not treated in haddock in the same way they are in markdown:
```suggestion
-- | Add a child to the already enacted actions. Returns a @`Just`
-- (`PrevGovActionIdsChildren` era)@ with the child added, or `Nothing` if
-- the `PrevGovActionId` is not found.
```

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Gov.hs
```diff
@@ -276,65 +257,95 @@ actionWellFormed ga = failureUnless isWellFormed $ MalformedProposal ga
...
+    UpdateCommittee prev _ _ _ ->
+      go pgaCommittee prev pgacCommitteeL (PrevGovActionId actionId) children
+    NewConstitution prev _ ->
+      go pgaConstitution prev pgacConstitutionL (PrevGovActionId actionId) children
+    InfoAction -> Just children
+  where
+    go pgai prev lens' pgai' children' =
+      if pgai == prev
+        then Just $ children' & lens' %~ Set.insert pgai'
+        else Nothing
```
lehins: Ditto. My previous suggestion included this change, but I guess you missed it. So I'l say it again explicitly. Avoid passing arguments that don't change and are in scope. Another general suggestion: try to avoid ticks in favor of better names, unless no better name comes to mind. 

```suggestion
  case action of
    ParameterChange prev _ ->
      go pgaPParamUpdate prev pgacPParamUpdateL (PrevGovActionId actionId)
    HardForkInitiation prev _ ->
      go pgaHardFork prev pgacHardForkL (PrevGovActionId actionId)
    TreasuryWithdrawals _ -> Just children
    NoConfidence prev ->
      go pgaCommittee prev pgacCommitteeL (PrevGovActionId actionId)
    UpdateCommittee prev _ _ _ ->
      go pgaCommittee prev pgacCommitteeL (PrevGovActionId actionId)
    NewConstitution prev _ ->
      go pgaConstitution prev pgacConstitutionL (PrevGovActionId actionId)
    InfoAction -> Just children
  where
    go pgai prev pgacL pgaiChild =
      if pgai == prev
        then Just $ children & pgacL %~ Set.insert pgaiChild
        else Nothing
```

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Gov.hs
```diff
@@ -356,18 +367,49 @@ govTransition = do
...
                 pProcDeposit
                 pProcReturnAddr
                 pProcGovAction
-                st'
-        applyProps st'' ps
-  stProps <- applyProps st $ indexedGovProps $ SSeq.fromStrict $ OSet.toStrictSeq $ gpProposalProcedures gp
+                expiry
+                currentEpoch
+            prevGovActionIdsChildren =
+              addToPrevGovActionIdsChildren
```
lehins: I think this slight change to the name of the function will make it much clearer on which children we are considering right now.
```suggestion
            prevEnactedGovActionIdsChildren =
              addToEnactedPrevGovActionIdsChildren
```

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Gov.hs
```diff
@@ -356,18 +367,49 @@ govTransition = do
...
+                  GovRuleState
+                    (proposalsInsertGovAction actionState grsProposals)
+                    updatedChildren
+              -- Didn't find a parent in the already enacted actions.
+              -- Searching in the already accepted proposals.
+              Nothing ->
+                case proposalsAddChild actionState grsProposals of
+                  -- If the parent is not found in the accepted proposals, the predicate fails
+                  Nothing -> grs <$ failBecause (InvalidPrevGovActionId proposal)
+                  Just grsProposals' ->
```
lehins: No tick and inline with the `updatedChildren` naming above
```suggestion
                  Just updatedProposals ->
```

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Gov.hs
```diff
@@ -356,18 +367,49 @@ govTransition = do
...
                 pProcGovAction
-                st'
-        applyProps st'' ps
-  stProps <- applyProps st $ indexedGovProps $ SSeq.fromStrict $ OSet.toStrictSeq $ gpProposalProcedures gp
+                expiry
+                currentEpoch
+            prevGovActionIdsChildren =
+              addToPrevGovActionIdsChildren
+                gai
+                pProcGovAction
```
lehins: As with `proposalsAddChild` and `proposalsInsertGovAction` we can pass the whole `actionState`, rather than id and proposal separately.
```suggestion
                actionState
```

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Gov.hs
```diff
@@ -356,18 +367,49 @@ govTransition = do
...
-              addAction
-                currentEpoch
-                (pp ^. ppGovActionLifetimeL)
-                (GovActionId txid idx)
+        -- Ancestry checks and accept proposal
+        let gai = GovActionId txid ix
+            expiry = pp ^. ppGovActionLifetimeL
+            actionState =
+              mkGovActionState
+                gai
```
lehins: With my other suggestion, `gai` binding is no longer needed
```suggestion
        let expiry = pp ^. ppGovActionLifetimeL
            actionState =
              mkGovActionState
                (GovActionId txid ix)
```

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Gov.hs
```diff
@@ -356,18 +367,49 @@ govTransition = do
...
+              -- Found a parent in the already enacted actions
+              Just updatedChildren ->
+                pure $
+                  GovRuleState
+                    (proposalsInsertGovAction actionState grsProposals)
+                    updatedChildren
+              -- Didn't find a parent in the already enacted actions.
+              -- Searching in the already accepted proposals.
+              Nothing ->
+                case proposalsAddChild actionState grsProposals of
```
lehins: Let's make `proposalsAddChild` an internal (not exported) function, which will be called by `proposalsInsertGovAction`. Afterall it should never happen that we insert a proposal without modifying the children
```suggestion
                case proposalsInsertGovAction actionState grsProposals of
```

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Gov.hs
```diff
@@ -356,18 +367,49 @@ govTransition = do
...
+                  -- If the parent is not found in the accepted proposals, the predicate fails
+                  Nothing -> grs <$ failBecause (InvalidPrevGovActionId proposal)
+                  Just grsProposals' ->
+                    pure $
+                      GovRuleState
+                        (proposalsInsertGovAction actionState grsProposals')
+                        grsPrevGovActionIdsChildren
```
lehins: With the suggestion above we can simplify this to:
```suggestion
                    pure $ grs {grsProposals = updatedProposals}
```

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Ratify.hs
```diff
@@ -348,6 +365,40 @@ prevActionAsExpected = \case
...
+  NoConfidence {} -> go (^. pgacCommitteeL)
+  UpdateCommittee {} -> go (^. pgacCommitteeL)
+  NewConstitution {} -> go (^. pgacConstitutionL)
+  InfoAction -> const mempty
+  where
+    go getter =
+      foldl'
+        (\accum pgai -> unPrevGovActionId pgai `Set.insert` accum)
+        Set.empty
+        . getter
```
lehins: If you want to keep the LambdaCase usage just swap the arguments. This way you don't need to use this weird pattern of using a getter. Also `PrevGovActionIdsChildren` already contains Sets, it's just contains a slightly different type from what we need. There is a much more efficient way of solving this without the need to iterate over the whole set:
```suggestion
getSiblings :: PrevGovActionIdsChildren era -> GovAction era -> Set.Set (GovActionId (EraCrypto era))
getSiblings pgac = \case
  ParameterChange {} -> coerce (pgac ^. pgacPParamUpdateL)
  HardForkInitiation {} -> coerce (pgac ^. pgacHardForkL)
  TreasuryWithdrawals _ -> const mempty
  NoConfidence {} -> coerce (pgac ^. pgacCommitteeL)
  UpdateCommittee {} -> coerce (pgac ^. pgacCommitteeL)
  NewConstitution {} -> coerce (pgac ^. pgacConstitutionL)
  InfoAction -> const mempty
```

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Ratify.hs
```diff
@@ -348,6 +365,40 @@ prevActionAsExpected = \case
...
+  UpdateCommittee {} -> go pgacCommitteeL gas
+  NewConstitution {} -> go pgacConstitutionL gas
+  InfoAction -> id
+  where
+    go lenz state =
+      lenz
+        .~ foldl'
+          (\accum gai -> PrevGovActionId gai `Set.insert` accum)
+          Set.empty
+          (state ^. gasChildrenL)
```
lehins: Please, do not go overboard on the function composition. Sometimes few extra characters make code way more readable. Here again, do not pass the same argument available in the scope to local bindings (`gas` in this case). In any case, here too, we can make it more efficient:

```suggestion
updatePrevGovActionIdsChildren gas pgac = case gas ^. gasActionL of
  ParameterChange {} -> pgac & pgacPParamUpdateL .~ coerce children
  HardForkInitiation {} -> pgac & pgacHardForkL .~ coerce children
  TreasuryWithdrawals _ -> pgac
  NoConfidence {} -> pgac & pgacCommitteeL .~ coerce children
  UpdateCommittee {} -> pgac & pgacCommitteeL .~ coerce children
  NewConstitution {} -> pgac & pgacConstitutionL .~ coerce children
  InfoAction -> pgac
  where
    children = gas ^. gasChildrenL
```

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/TypeRep.hs
```diff
@@ -331,14 +338,15 @@ data Rep era t where
   PrevHardForkR :: Era era => Rep era (PrevGovActionId 'HardForkPurpose (EraCrypto era))
   PrevCommitteeR :: Era era => Rep era (PrevGovActionId 'CommitteePurpose (EraCrypto era))
   PrevConstitutionR :: Era era => Rep era (PrevGovActionId 'ConstitutionPurpose (EraCrypto era))
-  RatifyStateR :: Reflect era => Rep era (RatifyState era)
+  RatifyStateR :: (Reflect era, Arbitrary (PParamsHKD StrictMaybe era)) => Rep era (RatifyState era)
```
lehins: Why is this constraint necessary now but it wasn't needed before.

I suspect that @TimSheard will be against change

lehins (changes requested):
I don't see any issues with correctness, however few things can be made a bit simpler.

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Gov.hs
aniketd: > Afterall it should never happen that we insert a proposal without modifying the children

This is expected to happen in the case when the parent of a proposal is in the `PrevGovActionIdsChildren` alone. In that case, we just want to insert the `GovActionState` in the proposals without updating the children and consider it unnecessary. The current semantics do not allow for this since the `Nothing` result is interpreted as "parent not found" and should result in a predicate failure. I will extract this functionality in a separate function under `Governance` so that it encapsulates the action of adding a proposal irrespective of where its parent lies and correctly returns the failing case.

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Gov.hs
aniketd: May I tackle this in a follow up PR with the test coverage?

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/TypeRep.hs
aniketd: This was no longer required, after rebasing on top of master and further introducing the new datatypes correctly into the existing constrained framework. :+1: Thank you! :smile: :raised_hands:

aniketd:
There are two tests in conway failing still. I'm looking into those.

eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Proposals.hs
```diff
@@ -144,3 +186,220 @@ fromGovActionStateSeq = Proposals . OMap.fromFoldable
...
+        Just (updatedChildren, Proposals $ omap OMap.||> gas)
+      (Nothing, Nothing) ->
+        -- Parent not found, cannot accept this proposal
+        Nothing
+      (Just updatedChildren, Nothing) ->
+        -- Parent is an enacted action
+        Just (updatedChildren, Proposals $ omap OMap.||> gas)
+      (Nothing, Just (Proposals updatedOMap)) ->
+        -- Parent is another accepted proposal
+        Just (pgac, Proposals $ updatedOMap OMap.||> gas)
```
lehins: It is not only about readability. By having the case statement on a tuple you are forcing `parentIsAnotherProposal` whenever it is unused, thus doing unnecessary computation.

```suggestion
    case parentIsEnacted of
      Just updatedChildren ->
        -- Parent is an enacted action
        Just (updatedChildren, Proposals $ omap OMap.||> gas)
      Nothing ->
        case parentIsAnotherProposal of
          -- Parent not found, cannot accept this proposal
          Nothing -> Nothing
          -- Parent is another accepted proposal
          Just (Proposals updatedOMap) ->
            Just (pgac, Proposals $ updatedOMap OMap.||> gas)
```

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Gov.hs
```diff
@@ -376,18 +353,38 @@ govTransition = do
...
+                actionState
+                prevGovActionIds
+                grsPrevGovActionIdsChildren
+                grsProposals
+         in case updatedProposalsState of
+              Just (updatedChildren, updatedProposals) ->
+                pure $
+                  GovRuleState
+                    updatedProposals
+                    updatedChildren
```
lehins: Slight compaction
```suggestion
                pure $ GovRuleState updatedProposals updatedChildren
```

lehins (approved):
@aniketd Awesome piece of work! Thank you!

eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Proposals.hs
aniketd: :100: :raised_hands: Thank you!

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Epoch.hs
```diff
@@ -279,22 +279,26 @@ epochTransition = do
...
 
     (accountState2, dState2, newEnactState) =
       applyEnactedWithdrawals accountState1 dState1 rsEnactState
-    (newProposals, removedGovActions) =
-      -- It is important that we use current proposals here instead of the ones from the pulser
-      proposalsRemoveIds rsRemoved (govState0 ^. cgProposalsL)
+
+    -- It is important that we use current proposals here instead of the ones from the pulser
+    (proposals0, removedInvalidGovActions) = proposalsRemoveDescendentIds rsRemoved (govState0 ^. cgProposalsL)
+    (proposals1, removedEnactedGovActions) = proposalsRemoveIds rsEnacted proposals0
```
aniketd, line 289: @lehins Regarding the fix for #4000 I think this is the site at which we can notice that enacted actions weren't being cleaned up at all before this PR was merged. Only `rsRemoved` (expired) actions were being removed before, and without clean up of their descendents.

PR #3858, opened 2023-11-09
Restructure computing Refunds and Deposits in a TxBody across all eras

TimSheard:
Add uniform ways to compute Refunds and Deposits in a TxBody by two methods of ShelleyEraTxBody
  getTotalDepositsTxBody :: PParams era -> CertState era -> TxBody era -> Coin
  getTotalRefundsTxBody :: PParams era -> CertState era -> TxBody era -> Coin

Make instances for all eras
Adjust UTXO and UTXOS rules to use these correctly.

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/eras/conway/impl/src/Cardano/Ledger/Conway/TxBody.hs
+++ b/eras/conway/impl/src/Cardano/Ledger/Conway/TxBody.hs
@@ -21,6 +21,7 @@
 {-# OPTIONS_GHC -Wno-orphans #-}
 
 module Cardano.Ledger.Conway.TxBody (
+  totalTxRefundsConway,
   ConwayEraTxBody (..),
   ConwayTxBody (
     ConwayTxBody,
@@ -53,7 +54,7 @@ import Cardano.Ledger.Babbage.TxBody (
   babbageAllInputsTxBodyF,
   babbageSpendableInputsTxBodyF,
  )
-import Cardano.Ledger.BaseTypes (Network, fromSMaybe, isSJust)
+import Cardano.Ledger.BaseTypes (EpochNo (..), Network, fromSMaybe, isSJust)
 import Cardano.Ledger.Binary (
   Annotator,
   DecCBOR (..),
@@ -76,16 +77,17 @@ import Cardano.Ledger.Binary.Coders (
   ofield,
   (!>),
  )
-import Cardano.Ledger.CertState (CertState)
+import Cardano.Ledger.CertState (CertState (..))
 import Cardano.Ledger.Coin (Coin (..), decodePositiveCoin)
 import Cardano.Ledger.Conway.Core
 import Cardano.Ledger.Conway.Era (ConwayEra)
 import Cardano.Ledger.Conway.Governance.Procedures (ProposalProcedure, VotingProcedures (..))
 import Cardano.Ledger.Conway.PParams (ConwayEraPParams, ppDRepDepositL, ppGovActionDepositL)
 import Cardano.Ledger.Conway.Scripts ()
-import Cardano.Ledger.Conway.TxCert (ConwayEraTxCert, ConwayTxCert (..), ConwayTxCertUpgradeError, pattern RegDRepTxCert)
+import Cardano.Ledger.Conway.TxCert (ConwayEraTxCert, ConwayTxCert (..), ConwayTxCertUpgradeError, pattern RegDRepTxCert, pattern UnRegDRepTxCert)
 import Cardano.Ledger.Conway.TxOut ()
 import Cardano.Ledger.Crypto
+import Cardano.Ledger.DRep (DRepState (..))
 import Cardano.Ledger.Keys (KeyHash (..), KeyRole (..))
 import Cardano.Ledger.Mary.Value (
   MaryValue (..),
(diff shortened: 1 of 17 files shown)


eras/shelley/impl/src/Cardano/Ledger/Shelley/Core.hs
```diff
@@ -43,6 +43,7 @@ class (ShelleyEraTxCert era, EraTxBody era) => ShelleyEraTxBody era where
   updateTxBodyG = updateTxBodyL
 
   getTotalDepositsTxBody :: PParams era -> CertState era -> TxBody era -> Coin
+  getTotalRefundsTxBody :: PParams era -> CertState era -> TxBody era -> Coin
```
lehins: Let's change the types of these functions to this and move them into `EraTxBody` type class:
```suggestion
  getTotalDepositsTxBody :: 
    PParams era ->
    -- | Check whether stake pool is registered or not 
    (KeyHash 'StakePool (EraCrypto era) -> Bool) ->
    TxBody era ->
    Coin
  getTotalRefundsTxBody :: 
    PParams era ->
    -- | Lookup current deposit for Staking credential if one is registered
    (Credential 'Staking (EraCrypto era) -> Maybe Coin) ->
    -- | Lookup current deposit for DRep credential if one is registered
    (Credential 'DRepRole (EraCrypto era) -> Maybe Coin) -> 
    TxBody era ->
    Coin
```

This will allow us to reuse them in `getConsumedValue` and `getProducedValue` as well as in all other places where this information is necessary.

Before you ask why would we change the type signature, it is important for cardano-cli for those functions to have this type.

eras/conway/impl/src/Cardano/Ledger/Conway/TxBody.hs
```diff
@@ -451,16 +456,45 @@ totalDRepDeposits pp txb = nDReps <×> depositPerDRep
...
+
+drepRefundsConway :: ConwayEraTxBody era => CertState era -> TxBody era -> Coin
+drepRefundsConway certState txb = Coin (snd (foldl' accum (vsDReps (certVState certState), 0) (txb ^. certsTxBodyL)))
+  where
+    accum pair@(currMap, currRefund) (UnRegDRepTxCert cred (Coin n)) = case (Map.lookup cred currMap) of
+      (Just (DRepState _ _ (Coin m))) | n == m -> (Map.delete cred currMap, currRefund + m)
+      _ -> pair
+    accum pair@(currMap, currRefund) (RegDRepTxCert cred c _) = case (Map.lookup cred currMap) of
+      Nothing -> (Map.insert cred (DRepState (EpochNo 0) SNothing c) currMap, currRefund)
+      -- we are only ever going to look at 'c', so the bogus epoch (EpochNo 0) and anchor (SNothing) are OK.
```
lehins: This is horrible. If you ever start using bogus values, try to think if it is possible to rewrite a function that doesn't rely on bogus values. In this case it is totally possible. 

In any case, there is a better implementation of this logic right here:

https://github.com/input-output-hk/cardano-ledger/blob/47f7b053553a505e7ab3b6cbf4550c1a199215f7/eras/conway/impl/src/Cardano/Ledger/Conway/UTxO.hs#L113-L122

eras/conway/impl/src/Cardano/Ledger/Conway/TxBody.hs
```diff
@@ -451,16 +456,45 @@ totalDRepDeposits pp txb = nDReps <×> depositPerDRep
...
-    <> totalDRepDeposits pp txb
+  totalTxDepositsShelley pp cs txb -- Stake and StakePool
+    <> totalProposalDeposits pp txb -- only in Conway and later
+    <> totalDRepDeposits pp txb -- only in Conway and later
+
+drepRefundsConway :: ConwayEraTxBody era => CertState era -> TxBody era -> Coin
+drepRefundsConway certState txb = Coin (snd (foldl' accum (vsDReps (certVState certState), 0) (txb ^. certsTxBodyL)))
+  where
+    accum pair@(currMap, currRefund) (UnRegDRepTxCert cred (Coin n)) = case (Map.lookup cred currMap) of
+      (Just (DRepState _ _ (Coin m))) | n == m -> (Map.delete cred currMap, currRefund + m)
```
lehins: This check is unnecessary, validity of deposits is checked in GOVCERT rule
```suggestion
      (Just (DRepState _ _ (Coin m))) -> (Map.delete cred currMap, currRefund + m)
```

lehins (changes requested):
This looks alright. We could merge it as is, but deposits and refunds need more work around deduplication.

We can talk more about this, but I'd rather have it fixed in this PR

eras/conway/impl/src/Cardano/Ledger/Conway/TxBody.hs
TimSheard: Redone as suggested, but without PParams.

eras/conway/impl/src/Cardano/Ledger/Conway/TxBody.hs
TimSheard: no longer applicable.

eras/shelley/impl/src/Cardano/Ledger/Shelley/Core.hs
TimSheard: Did this in a slightly different way. You will like it.

lehins (approved):
:+1:

PR #3860, opened 2023-11-09
Removed `mock/crypto.cddl`, added optional tag to sets

### LINKED ISSUE #3820: Remove `cddl-files/mock/crypto.cddl`
The cddl spec that lists all of the mock crypto is completely useless, since we haven't been using mock crypto for a while now.

Those files need to be removed in all eras.

### LINKED ISSUE #3845: Fix CDDL to indicate Set tag
In Conway we now prefix all sets with an optional 258 tag.

This needs to be clearly indicated in the CDDL file, since some places it still has not been updated.

Soupstraw:
# Description

This PR adds an optional tag to `set` and `nonempty_set` in the CDDL. 
Also removed the `mock` directory and moved the files from `real` directory to `cddl-files`. 

resolves #3845 
resolves #3820 

# Checklist

- [ ] Self-reviewed the diff


eras/conway/impl/cddl-files/extra.cddl
```diff
@@ -1,22 +1,19 @@
...
   ; and can also produce a zero in the denominator
 
-rational =  #6.30(
-   [ numerator   : 1
-   , denominator : 2
-   ])
-  ; real rational is: #6.30([uint, uint])
-  ; but this produces numbers outside the unit interval
-  ; and can also produce a zero in the denominator
+rational =  #6.30([uint, positive_int])
```
lehins, line 14: Should we change this to an `int`? We'd like to test negative values for rationals as well.Could you give it a try and see if it gives you problems?
```suggestion
rational =  #6.30([int, positive_int])
```

lehins (approved):
Great cleanup! Thank you!

eras/conway/impl/cddl-files/extra.cddl
lehins: Eeh, we can do that later at some point. Don't wanna block the PR

PR #3861, opened 2023-11-09
Fail PParamsUpdate deserialization for invalid costmodels in Conway

### LINKED ISSUE #3772: Prevent CostModels for known languages to be decoded in PParamsUpdate
* For language versions that we know, we prevent invalid ones in PParamUpdates in the governance proposals.
* For the unknown ones we do what is currently implemented

It is important that this behavior is only changed for `PParamsUpdate`, not for `PParams`

So, I suggest we do not change `CostModels` `DecCBOR` instance, which would still allow invalid CostModels for languages that are available, but have a special decoder function that is used for `CostModels` in `PParamsUpdate` that prevents invalid `CostModels` for languages that we already support

I am also contemplating to use different types for those two concepts, but I am not sure yet how hard that would be to implement

teodanciu:
# Description

Closes https://github.com/input-output-hk/cardano-ledger/issues/3772

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/eras/alonzo/impl/cardano-ledger-alonzo.cabal
+++ b/eras/alonzo/impl/cardano-ledger-alonzo.cabal
@@ -99,6 +99,7 @@ library testlib
     exposed-modules:
         Test.Cardano.Ledger.Alonzo.Arbitrary
         Test.Cardano.Ledger.Alonzo.Binary.Cddl
+        Test.Cardano.Ledger.Alonzo.Binary.CostModelsSpec
         Test.Cardano.Ledger.Alonzo.Binary.RoundTrip
         Test.Cardano.Ledger.Alonzo.CostModel
         Test.Cardano.Ledger.Alonzo.ImpTest
--- a/eras/alonzo/impl/testlib/Test/Cardano/Ledger/Alonzo/Binary/CostModelsSpec.hs
+++ b/eras/alonzo/impl/testlib/Test/Cardano/Ledger/Alonzo/Binary/CostModelsSpec.hs
@@ -1,0 +1,156 @@
+{-# LANGUAGE DataKinds #-}
+{-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE LambdaCase #-}
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE StandaloneDeriving #-}
+{-# LANGUAGE TypeApplications #-}
+
+module Test.Cardano.Ledger.Alonzo.Binary.CostModelsSpec (spec) where
+
+import Cardano.Ledger.Alonzo.Core
+import Cardano.Ledger.Alonzo.Scripts
+import Cardano.Ledger.BaseTypes
+import Cardano.Ledger.Binary (ToCBOR (..))
+import Cardano.Ledger.Binary.Decoding
+import Cardano.Ledger.Binary.Encoding
+import Cardano.Ledger.Plutus.Language
+import qualified Data.Map as Map
+import Data.Text (Text)
+import Data.Typeable
+import Data.Word (Word8)
+import Lens.Micro
+import Test.Cardano.Ledger.Alonzo.Arbitrary
+import Test.Cardano.Ledger.Common
+
+spec ::
+  forall era.
+  AlonzoEraPParams era =>
+  Proxy era ->
+  Spec
+spec pxy = do
+  describe "CostModels CBOR deserialization" $ do
+    validCostModelProp @era pxy
+    invalidCostModelProp @era Proxy
+    unknownCostModelProp @era Proxy
+
+validCostModelProp ::
+  forall era.
+  AlonzoEraPParams era =>
+  Proxy era ->
+  Spec
+validCostModelProp pxy = do
+  prop "valid CostModels deserialize correctly, both independently and within PParamsUpdate" $
+    \(lang :: Language) -> do
+      forAllShow (genValidCostModelEnc lang) (showEnc pxy) $
+        \validCmEnc -> do
+          encodeAndCheckDecoded @era validCmEnc $
+            \cmDecoded ppuDecoded -> do
+              expectRight cmDecoded >>= (`shouldSatisfy` validCm)
+              ppuRes <- expectRight ppuDecoded
+              ppuRes `shouldSatisfy` \ppu -> (validCm <$> ppu ^. ppuCostModelsL) == SJust True
+  where
+    genValidCostModelEnc lang = genCostModelEncForLanguage lang (costModelParamsCount lang)
+    validCm (CostModels valid errors unknown) =
+      not (null valid) && null errors && null unknown
+
+invalidCostModelProp ::
+  forall era.
+  AlonzoEraPParams era =>
+  Proxy era ->
+  Spec
+invalidCostModelProp pxy = do
+  prop "invalid CostModels fail within PParamsUpdate" $
+    \(lang :: Language) -> do
+      forAllShow (genInvalidCostModelEnc lang) (showEnc pxy) $
+        \invalidCmEnc -> do
+          encodeAndCheckDecoded @era invalidCmEnc $
+            \cmDecoded ppuDecoded -> do
+              -- pre-Conway we are failing when deserializing invalid costmodels
+              if eraProtVerHigh @era < natVersion @9
+                then expectDeserialiseFailure cmDecoded (Just "CostModels")
+                else do
+                  -- post-Conway, we are collecting CostModels deserialization errors
+                  cmRes <- expectRight cmDecoded
+                  cmRes `shouldSatisfy` invalidCm
+
+              -- in no era are we deserializing invalid costmodels within PParamsUpdate
+              expectDeserialiseFailure ppuDecoded Nothing
+  where
+    genInvalidCostModelEnc lang = do
+      let validCount = costModelParamsCount lang
+      count <- choose (0, validCount - 1)
+      genCostModelEncForLanguage lang count
+
+    invalidCm (CostModels valid errors _) =
+      null valid && not (null errors)
+
+unknownCostModelProp ::
+  forall era.
+  AlonzoEraPParams era =>
+  Proxy era ->
+  Spec
+unknownCostModelProp pxy = do
+  prop "unknown CostModels deserialize correctly within PParamsUpdate starting with Conway" $
+    forAllShow genUnknownCostModelEnc (showEnc pxy) $
+      \unknownCmEnc -> do
+        encodeAndCheckDecoded @era unknownCmEnc $
+          \cmDecoded ppuDecoded -> do
+            -- pre-Conway we are failing when deserializing unknown costmodels
+            if eraProtVerHigh @era < natVersion @9
+              then do
+                expectDeserialiseFailure cmDecoded (Just "CostModels")
+                expectDeserialiseFailure ppuDecoded Nothing
+              else do
+                -- post-Conway, we are collecting unknown CostModels
+                cmRes <- expectRight cmDecoded
+                cmRes `shouldSatisfy` unknownCm
+                ppuRes <- expectRight ppuDecoded
+                ppuRes `shouldSatisfy` \ppu -> (unknownCm <$> ppu ^. ppuCostModelsL) == SJust True
+  where
+    genUnknownCostModelEnc = do
+      let firstUnknownLang = fromEnum (maxBound @Language) + 1
+      lang <- choose (fromIntegral firstUnknownLang, maxBound @Word8)
+      NonNegative count <- arbitrary
+      genCostModelsEnc lang count
+    unknownCm (CostModels valid errors unknown) =
+      null valid && null errors && not (null unknown)
+
+encodeAndCheckDecoded ::
+  forall era.
+  AlonzoEraPParams era =>
+  Encoding ->
+  (Either DecoderError CostModels -> Either DecoderError (PParamsUpdate era) -> IO ()) ->
+  IO ()
+encodeAndCheckDecoded cmEnc check = do
+  let ver = eraProtVerHigh @era
+      ppuEnc =
+        fromPlainEncoding
+          . toCBOR
+          . Map.singleton (18 :: Int)
+          . toPlainEncoding ver
+          $ cmEnc
+      cmBytes = serialize ver cmEnc
+      ppuBytes = serialize ver ppuEnc
+  check
+    (decodeFull @CostModels ver cmBytes)
+    (decodeFull @(PParamsUpdate era) ver ppuBytes)
+
+genCostModelsEnc :: Word8 -> Int -> Gen Encoding
+genCostModelsEnc lang count = do
+  values <- vectorOf count (arbitrary :: Gen Int)
+  pure $ fromPlainEncoding . toCBOR $ Map.singleton lang values
+
+genCostModelEncForLanguage :: Language -> Int -> Gen Encoding
+genCostModelEncForLanguage lang =
+  genCostModelsEnc (fromIntegral (fromEnum lang))
+
+expectDeserialiseFailure :: (HasCallStack, Show t) => Either DecoderError t -> Maybe Text -> IO ()
+expectDeserialiseFailure e expectedTxt = do
+  res <- expectLeft e
+  res `shouldSatisfy` \case
+    DecoderErrorDeserialiseFailure txt _ -> maybe True (== txt) expectedTxt
+    _ -> True
+
+showEnc :: forall era. Era era => Proxy era -> Encoding -> String
+showEnc _ enc = show $ toPlainEncoding (eraProtVerHigh @era) enc
--- a/eras/conway/impl/src/Cardano/Ledger/Conway/PParams.hs
+++ b/eras/conway/impl/src/Cardano/Ledger/Conway/PParams.hs
@@ -79,6 +79,7 @@ import Cardano.Ledger.Conway.Core hiding (Value)
 import Cardano.Ledger.Conway.Era (ConwayEra)
 import Cardano.Ledger.Crypto
 import Cardano.Ledger.HKD (HKD, HKDFunctor (..), HKDNoUpdate, NoUpdate (..))
+import Cardano.Ledger.Plutus.CostModels (decodeValidAndUnknownCostModels)
 import Cardano.Ledger.TreeDiff (ToExpr (..))
 import Cardano.Ledger.Val (Val (..))
 import Control.DeepSeq (NFData (..))
@@ -675,7 +676,7 @@ updateField = \case
   11 -> field (\x up -> up {cppTau = THKD (SJust x)}) From
   16 -> field (\x up -> up {cppMinPoolCost = THKD (SJust x)}) From
   17 -> field (\x up -> up {cppCoinsPerUTxOByte = THKD (SJust x)}) From
-  18 -> field (\x up -> up {cppCostModels = THKD (SJust x)}) From
+  18 -> field (\x up -> up {cppCostModels = THKD (SJust x)}) $ D decodeValidAndUnknownCostModels
   19 -> field (\x up -> up {cppPrices = THKD (SJust x)}) From
   20 -> field (\x up -> up {cppMaxTxExUnits = THKD (SJust x)}) From
   21 -> field (\x up -> up {cppMaxBlockExUnits = THKD (SJust x)}) From
--- a/eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Arbitrary.hs
+++ b/eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Arbitrary.hs
@@ -64,7 +64,7 @@ import Data.List (nubBy)
 import qualified Data.Sequence.Strict as Seq
 import Test.Cardano.Data (genNonEmptyMap)
 import Test.Cardano.Data.Arbitrary ()
-import Test.Cardano.Ledger.Alonzo.Arbitrary (genAlonzoScript, unFlexibleCostModels)
+import Test.Cardano.Ledger.Alonzo.Arbitrary (genAlonzoScript, genValidAndUnknownCostModels, unFlexibleCostModels)
 import Test.Cardano.Ledger.Babbage.Arbitrary ()
 import Test.Cardano.Ledger.Common
 
@@ -504,7 +504,7 @@ instance Era era => Arbitrary (ConwayPParams StrictMaybe era) where
       <*> pure NoUpdate
       <*> arbitrary
       <*> arbitrary
-      <*> (THKD . fmap unFlexibleCostModels <$> arbitrary)
+      <*> (THKD <$> oneof [SJust <$> genValidAndUnknownCostModels, pure SNothing])
       <*> arbitrary
       <*> arbitrary
       <*> arbitrary
--- a/eras/alonzo/impl/CHANGELOG.md
+++ b/eras/alonzo/impl/CHANGELOG.md
@@ -12,6 +12,7 @@
 
 * Provide CDDL spec files with `readAlonzoCddlFileNames` and `readAlonzoCddlFiles` from
   `Test.Cardano.Ledger.Alonzo.Binary.Cddl`
+* Expose `genValidAndUnknownCostModels` in `Arbitrary`
 
 ## 1.5.0.0
 
--- a/eras/alonzo/impl/test/Main.hs
+++ b/eras/alonzo/impl/test/Main.hs
@@ -3,7 +3,9 @@
 module Main where
 
 import Cardano.Ledger.Alonzo (Alonzo)
+import Data.Proxy (Proxy (Proxy))
 import qualified Test.Cardano.Ledger.Alonzo.Binary.CddlSpec as CddlSpec
+import qualified Test.Cardano.Ledger.Alonzo.Binary.CostModelsSpec as CostModelsSpec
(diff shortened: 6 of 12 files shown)


eras/conway/impl/src/Cardano/Ledger/Conway/PParams.hs
```diff
@@ -677,7 +678,7 @@ updateField = \case
   11 -> field (\x up -> up {cppTau = THKD (SJust x)}) From
   16 -> field (\x up -> up {cppMinPoolCost = THKD (SJust x)}) From
   17 -> field (\x up -> up {cppCoinsPerUTxOByte = THKD (SJust x)}) From
-  18 -> field (\x up -> up {cppCostModels = THKD (SJust x)}) From
+  18 -> field (\x up -> up {cppCostModels = THKD (SJust x)}) $ D decodeCostModelsFailingOnError
```
lehins: This is an incorrect function to use, because `decodeCostModelsFailingOnError` will fail for unknown languages, which is not what we want.

We want PParamsUpdate to fail whenever there is a CostModel for a known language that is invalid, but it should still succeed whenever the CostModel is for a language that we don't know about. 

In particular this is the line that is wrong for this decoder, but it is ok for PParams: https://github.com/input-output-hk/cardano-ledger/blob/c8874b4ae6ef569d28f492dc6d948bfdc607a3ef/libs/cardano-ledger-core/src/Cardano/Ledger/Plutus/CostModels.hs#L341

eras/alonzo/impl/testlib/Test/Cardano/Ledger/Alonzo/Binary/CostModelsSpec.hs
```diff
@@ -0,0 +1,183 @@
...
+
+checkValidCostModelDecoder ::
+  forall era.
+  AlonzoEraPParams era =>
+  Proxy era ->
+  BSL.ByteString ->
+  BSL.ByteString ->
+  IO ()
+checkValidCostModelDecoder _ cmsBytes ppuBytes = do
+  let ver = (eraProtVerHigh @era)
```
lehins:
```suggestion
  let ver = eraProtVerHigh @era
```

eras/alonzo/impl/testlib/Test/Cardano/Ledger/Alonzo/Binary/CostModelsSpec.hs
```diff
@@ -0,0 +1,183 @@
...
+  let cmDecoded = decodeFull @CostModels ver cmsBytes
+  let ppuDecoded = decodeFull @(PParamsUpdate era) ver ppuBytes
+  -- pre-Conway we are failing when deserializing invalid costmodels
+  if (ver < natVersion @9)
+    then
+      cmDecoded
+        `shouldBeLeftAndSatisfy` ( \case
+                                    DecoderErrorDeserialiseFailure txt _ -> txt == "CostModels"
+                                    _ -> False
+                                 )
```
lehins:
```suggestion
    then do
     res <- expectLeftExpr cmDecoded
     res `shouldSatisfy` \case
       DecoderErrorDeserialiseFailure txt _ -> txt == "CostModels"
       _ -> False
```

eras/alonzo/impl/testlib/Test/Cardano/Ledger/Alonzo/Binary/CostModelsSpec.hs
```diff
@@ -0,0 +1,183 @@
...
+    else -- post-Conway, we are collecting CostModels deserialization errors
+
+      cmDecoded `shouldBeRightAndSatisfy` invalidCm
+
+  -- in no era are we deserializing invalid costmodels within PParamsUpdate
+  ppuDecoded
+    `shouldBeLeftAndSatisfy` ( \case
+                                DecoderErrorDeserialiseFailure _ _ -> True
+                                _ -> False
+                             )
```
lehins:
```suggestion
  res <- expectLeft ppuDecoded
  res `shouldSatisfy` \case
    DecoderErrorDeserialiseFailure _ _ -> True
    _ -> False
```

eras/alonzo/impl/testlib/Test/Cardano/Ledger/Alonzo/Binary/CostModelsSpec.hs
```diff
@@ -0,0 +1,183 @@
...
+  forall era.
+  AlonzoEraPParams era =>
+  Proxy era ->
+  BSL.ByteString ->
+  BSL.ByteString ->
+  IO ()
+checkValidCostModelDecoder _ cmsBytes ppuBytes = do
+  let ver = (eraProtVerHigh @era)
+  let cmDecoded = decodeFull @CostModels ver cmsBytes
+  cmDecoded `shouldBeRightAndSatisfy` validCm
```
lehins:
```suggestion
  expectRightExpr cmDecoded >>= (`shouldSatisfy` validCm)
```

eras/alonzo/impl/testlib/Test/Cardano/Ledger/Alonzo/Binary/CostModelsSpec.hs
```diff
@@ -0,0 +1,183 @@
...
+
+  let ppuDecoded = decodeFull @(PParamsUpdate era) ver ppuBytes
+  ppuDecoded
+    `shouldBeRightAndSatisfy` ( \ppu ->
+                                  (validCm <$> ppu ^. ppuCostModelsL) == SJust True
+                              )
```
lehins:
```suggestion
  res <- expectRight ppuDecoded
  res `shouldSatisfy` \ppu -> (validCm <$> ppu ^. ppuCostModelsL) == SJust True
```

eras/alonzo/impl/testlib/Test/Cardano/Ledger/Alonzo/Binary/CostModelsSpec.hs
```diff
@@ -0,0 +1,183 @@
...
+genInvalidCostModelsEnc lang = do
+  let validCount = costModelParamsCount lang
+  count <- choose (0, validCount - 1)
+  genCostModelsEncoding lang count
+
+genCostModelsEncoding :: Language -> Int -> Gen Encoding
+genCostModelsEncoding lang count = (fromPlainEncoding . toCBOR) <$> enc
+  where
+    enc :: Gen Enc.Enc
+    enc = do
```
lehins: Redundant parenthesis and binding
```suggestion
genCostModelsEncoding lang count = fromPlainEncoding . toCBOR <$> do
```

eras/alonzo/impl/testlib/Test/Cardano/Ledger/Alonzo/Binary/CostModelsSpec.hs
```diff
@@ -0,0 +1,183 @@
...
+      values <- vectorOf count (arbitrary :: Gen Int)
+      pure $
+        mconcat
+          [ Enc.E (TkMapLen 1)
+          , Enc.E (TkInt (fromEnum lang))
+          , Enc.E TkListBegin
+          , Enc.Em (Enc.E . TkInt <$> values)
+          , Enc.E TkBreak
+          ]
```
lehins: This is all you really need 
:smile: 
```suggestion
      pure $ toCBOR $ Map.singleton lang values
```

eras/alonzo/impl/testlib/Test/Cardano/Ledger/Alonzo/Binary/CostModelsSpec.hs
```diff
@@ -0,0 +1,183 @@
...
+
+genInvalidCostModelsEnc ::
+  Language ->
+  Gen Encoding
+genInvalidCostModelsEnc lang = do
+  let validCount = costModelParamsCount lang
+  count <- choose (0, validCount - 1)
+  genCostModelsEncoding lang count
+
+genCostModelsEncoding :: Language -> Int -> Gen Encoding
```
lehins: You want to be able to generate unknown languages
```suggestion
genCostModelsEncoding :: Int -> Int -> Gen Encoding
```

eras/alonzo/impl/testlib/Test/Cardano/Ledger/Alonzo/Binary/CostModelsSpec.hs
```diff
@@ -0,0 +1,183 @@
...
+genValidCostModelsEnc lang =
+  genCostModelsEncoding lang (costModelParamsCount lang)
+
+genInvalidCostModelsEnc ::
+  Language ->
+  Gen Encoding
+genInvalidCostModelsEnc lang = do
+  let validCount = costModelParamsCount lang
+  count <- choose (0, validCount - 1)
+  genCostModelsEncoding lang count
```
lehins:
```suggestion
  genCostModelsEncoding (fromEnum lang) count
```

eras/alonzo/impl/testlib/Test/Cardano/Ledger/Alonzo/Binary/CostModelsSpec.hs
```diff
@@ -0,0 +1,183 @@
...
+                                _ -> False
+                             )
+  where
+    invalidCm (CostModels valid errors _) =
+      null valid && (not . null) errors
+
+genValidCostModelsEnc ::
+  Language ->
+  Gen Encoding
+genValidCostModelsEnc lang =
```
lehins: Will other changes I suggested this should be a valid CostModel.
```suggestion
genValidCostModelsEnc :: Gen Encoding
genValidCostModelsEnc = do
  NonNegative lang <- arbitrary
```

eras/alonzo/impl/testlib/Test/Cardano/Ledger/Alonzo/Binary/CostModelsSpec.hs
```diff
@@ -0,0 +1,183 @@
...
+  Proxy era ->
+  BSL.ByteString ->
+  BSL.ByteString ->
+  IO ()
+checkInvalidCostModelDecoder _ cmsBytes ppuBytes = do
+  let ver = (eraProtVerHigh @era)
+  let cmDecoded = decodeFull @CostModels ver cmsBytes
+  let ppuDecoded = decodeFull @(PParamsUpdate era) ver ppuBytes
+  -- pre-Conway we are failing when deserializing invalid costmodels
+  if (ver < natVersion @9)
```
lehins:
```suggestion
  if ver < natVersion @9
```

eras/alonzo/impl/testlib/Test/Cardano/Ledger/Alonzo/Binary/CostModelsSpec.hs
```diff
@@ -0,0 +1,183 @@
...
+shouldBeRightAndSatisfy (Right r) p =
+  unless (p r) (expectationFailure ("Expectation failed for: " <> show r))
+shouldBeRightAndSatisfy (Left l) _ =
+  expectationFailure $ "Expected Right, got Left: " <> show l
+
+shouldBeLeftAndSatisfy :: (Show a, Show b) => Either a b -> (a -> Bool) -> IO ()
+shouldBeLeftAndSatisfy (Right r) _ =
+  expectationFailure $ "Expected Left, got Right: " <> show r
+shouldBeLeftAndSatisfy (Left l) p =
+  unless (p l) (expectationFailure ("Expectation failed for: " <> show l))
```
lehins: I've recently added similar helpers here: https://github.com/input-output-hk/cardano-ledger/blob/c8874b4ae6ef569d28f492dc6d948bfdc607a3ef/libs/cardano-ledger-core/testlib/Test/Cardano/Ledger/Common.hs#L81-L130
```suggestion
```

eras/alonzo/impl/testlib/Test/Cardano/Ledger/Alonzo/Binary/CostModelsSpec.hs
```diff
@@ -0,0 +1,183 @@
...
+genPParamsUpdateEnc _ cmEnc = (fromPlainEncoding . toCBOR) <$> enc
+  where
+    enc :: Gen Enc.Enc
+    enc = do
+      pure $
+        mconcat
+          [ Enc.E (TkMapLen 1)
+          , Enc.E (TkInt 18)
+          , Enc.E (toPlainEncoding (eraProtVerHigh @era) cmEnc)
+          ]
```
lehins: This is not a `Gen`, it has no randomness. Also usage of `E` is unnecessary:
```suggestion
genPParamsUpdateEnc ::
  forall era.
  Era era =>
  Proxy era ->
  Encoding ->
  Encoding
genPParamsUpdateEnc _ = 
  fromPlainEncoding . toCBOR . Map.singleton (18 :: Int) . toPlainEncoding (eraProtVerHigh @era)
```

lehins (changes requested):
Not quite what needed to be done, but close. Test cases are very nice! Thank you for that

eras/conway/impl/src/Cardano/Ledger/Conway/PParams.hs
teodanciu: You're totally right, I completely missed half of the point of the task, sorry about that. 
Writing another function that is a combination of the two,  `decodeCostModelsFailingOnError` and `decodeCostModelsCollectingErrors` - that fails for invalid but collects for unknown.

teodanciu: "Incidentally", the part I didn't write tests for

eras/alonzo/impl/testlib/Test/Cardano/Ledger/Alonzo/Binary/CostModelsSpec.hs
teodanciu: Code looks much better with these, thanks.

eras/alonzo/impl/testlib/Test/Cardano/Ledger/Alonzo/Binary/CostModelsSpec.hs
teodanciu: :exploding_head:

lehins (changes requested):
You also need to update what sort of CostModel we generate fro PParamUpdates here: https://github.com/input-output-hk/cardano-ledger/blob/3d207b01dca78b231874ab27a2e016cf22753f86/eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Arbitrary.hs#L507

I suspect that is the reason for CI failures

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Arbitrary.hs
```diff
@@ -504,7 +504,7 @@ instance Era era => Arbitrary (ConwayPParams StrictMaybe era) where
       <*> pure NoUpdate
       <*> arbitrary
       <*> arbitrary
-      <*> (THKD . fmap unFlexibleCostModels <$> arbitrary)
+      <*> (THKD <$> oneof [SJust <$> genValidAndUnknownCostModels, pure SNothing])
```
teodanciu, line 507: Is there a better way to do this?

lehins: That's the right way to go about it.

eras/alonzo/impl/cardano-ledger-alonzo.cabal
```diff
@@ -102,6 +102,7 @@ library testlib
         Test.Cardano.Ledger.Alonzo.Binary.RoundTrip
         Test.Cardano.Ledger.Alonzo.CostModel
         Test.Cardano.Ledger.Alonzo.ImpTest
+        Test.Cardano.Ledger.Alonzo.Binary.CostModelsSpec
```
lehins: Nothing important, just alphabetic grouping.
```suggestion
        Test.Cardano.Ledger.Alonzo.Binary.CostModelsSpec
        Test.Cardano.Ledger.Alonzo.Binary.RoundTrip
        Test.Cardano.Ledger.Alonzo.CostModel
        Test.Cardano.Ledger.Alonzo.ImpTest
```

eras/alonzo/impl/testlib/Test/Cardano/Ledger/Alonzo/Binary/CostModelsSpec.hs
```diff
@@ -0,0 +1,156 @@
...
+    \(lang :: Language) -> do
+      forAllShow (genValidCostModelsEnc lang) (showEnc pxy) $
+        \validCmEnc -> do
+          encodeAndCheckDecoded @era validCmEnc $
+            \cmDecoded ppuDecoded -> do
+              expectRight cmDecoded >>= (`shouldSatisfy` validCm)
+              ppuRes <- expectRight ppuDecoded
+              ppuRes `shouldSatisfy` \ppu -> (validCm <$> ppu ^. ppuCostModelsL) == SJust True
+  where
+    genValidCostModelsEnc lang = genCostModelsEncForLanguage lang (costModelParamsCount lang)
```
lehins: Minor nit pick. Each language has one cost model, so it is not plural. The only reason why I am suggesting it, is because it can be confused with generator that produces CostModels for many languages (which even has a separate type `CostModels`) 

```suggestion
    genValidCostModelEnc lang = genCostModelEncForLanguage lang (costModelParamsCount lang)
```

eras/alonzo/impl/testlib/Test/Cardano/Ledger/Alonzo/Binary/CostModelsSpec.hs
```diff
@@ -0,0 +1,156 @@
...
+    validCm (CostModels valid errors unknown) =
+      (not . null) valid && null errors && null unknown
```
lehins: Avoiding redundant function composition. 
```suggestion
      not (null valid) && null errors && null unknown
```
Also this style is just ugly `(foo . bar . baz) blah` IMHO.

eras/alonzo/impl/testlib/Test/Cardano/Ledger/Alonzo/Binary/CostModelsSpec.hs
```diff
@@ -0,0 +1,156 @@
...
+genCostModelsEnc :: Word8 -> Int -> Gen Encoding
+genCostModelsEnc lang count = do
+  values <- vectorOf count (arbitrary :: Gen Int)
+  pure $ fromPlainEncoding . toCBOR $ Map.singleton lang values
+
+genCostModelsEncForLanguage :: Language -> Int -> Gen Encoding
+genCostModelsEncForLanguage lang =
+  genCostModelsEnc (fromIntegral (fromEnum lang))
+
+expectDeserialiseFailure :: Show t => Either DecoderError t -> Maybe Text -> IO ()
```
lehins: Adding a call stack will inform you which usage of the function triggered the expectation failure
```suggestion
expectDeserialiseFailure :: 
  (HasCallStack, Show t) => Either DecoderError t -> Maybe Text -> IO ()
```

eras/alonzo/impl/testlib/Test/Cardano/Ledger/Alonzo/Binary/CostModelsSpec.hs
```diff
@@ -0,0 +1,156 @@
...
+                ppuRes <- expectRight ppuDecoded
+                ppuRes `shouldSatisfy` \ppu -> (unknownCm <$> ppu ^. ppuCostModelsL) == SJust True
+  where
+    genUnknownCostModelsEnc = do
+      let firstUnknownLang = (fromEnum (maxBound @Language)) + 1
+      lang <- choose (fromIntegral firstUnknownLang, maxBound @Word8)
+      NonNegative count <- arbitrary
+      genCostModelsEnc lang count
+    unknownCm (CostModels valid errors unknown) =
+      null valid && null errors && (not . null) unknown
```
lehins: Ditto
```suggestion
      null valid && null errors && not (null unknown)
```

eras/alonzo/impl/testlib/Test/Cardano/Ledger/Alonzo/Binary/CostModelsSpec.hs
```diff
@@ -0,0 +1,156 @@
...
+                expectDeserialiseFailure ppuDecoded Nothing
+              else do
+                -- post-Conway, we are collecting unknown CostModels
+                cmRes <- expectRight cmDecoded
+                cmRes `shouldSatisfy` unknownCm
+                ppuRes <- expectRight ppuDecoded
+                ppuRes `shouldSatisfy` \ppu -> (unknownCm <$> ppu ^. ppuCostModelsL) == SJust True
+  where
+    genUnknownCostModelsEnc = do
+      let firstUnknownLang = (fromEnum (maxBound @Language)) + 1
```
lehins: Redundant
```suggestion
      let firstUnknownLang = fromEnum (maxBound @Language) + 1
```

eras/alonzo/impl/testlib/Test/Cardano/Ledger/Alonzo/Binary/CostModelsSpec.hs
```diff
@@ -0,0 +1,156 @@
...
+              -- in no era are we deserializing invalid costmodels within PParamsUpdate
+              expectDeserialiseFailure ppuDecoded Nothing
+  where
+    genInvalidCostModelsEnc lang = do
+      let validCount = costModelParamsCount lang
+      count <- choose (0, validCount - 1)
+      genCostModelsEncForLanguage lang count
+
+    invalidCm (CostModels valid errors _) =
+      null valid && (not . null) errors
```
lehins: Ditto redundant composition. Also should the unknown be empty here as well?
```suggestion
    invalidCm (CostModels valid errors unknown) =
      null valid && not (null errors) && null unknown
```

eras/alonzo/impl/testlib/Test/Cardano/Ledger/Alonzo/Binary/CostModelsSpec.hs
```diff
@@ -0,0 +1,156 @@
...
+encodeAndCheckDecoded cmEnc check = do
+  let ppuEnc =
+        fromPlainEncoding
+          . toCBOR
+          . Map.singleton (18 :: Int)
+          . toPlainEncoding (eraProtVerHigh @era)
+          $ cmEnc
+  let cmBytes = serialize (eraProtVerHigh @era) cmEnc
+  let ppuBytes = serialize (eraProtVerHigh @era) ppuEnc
+  let ver = eraProtVerHigh @era
```
lehins: A little less code:
```suggestion
  let ver = eraProtVerHigh @era
      cmBytes = serialize ver cmEnc
      ppuBytes = serialize ver ppuEnc
```

lehins (approved):
This is beautiful!

I love the extensive test suite for this functionality!

All of the comments I made are mere suggestions.

eras/alonzo/impl/testlib/Test/Cardano/Ledger/Alonzo/Binary/CostModelsSpec.hs
teodanciu: I was gonna ask you and I forgot, because in the implementation, we are actually updating the invalid cost models when we encounter an error, look: https://github.com/input-output-hk/cardano-ledger/blob/master/libs/cardano-ledger-core/src/Cardano/Ledger/Plutus/CostModels.hs#L340  
Do you think by accident?  Wasn't sure, so I assumed it's on purpose for completion

eras/alonzo/impl/testlib/Test/Cardano/Ledger/Alonzo/Binary/CostModelsSpec.hs
teodanciu: I don't have any sensibility about this, and because of that I always struggle to decide.  I'm happy to follow "no redundant composition", it makes my life easier :)

eras/alonzo/impl/cardano-ledger-alonzo.cabal
teodanciu: It is important to prevent the descent into chaos :D Thank you for noticing

eras/alonzo/impl/testlib/Test/Cardano/Ledger/Alonzo/Binary/CostModelsSpec.hs
lehins: This is actually an interesting case. On one hand there is no possibility for an invalid CM for a known language to all of a sudden become valid, so we could discard it. On the other keeping it in memory not only allows us to debug a problem if there is one, but also prevents us from violating roundtrip property.
So, I think it is best to preserve the current behavior, which means you can ignore my suggestion here.

Thank you for bringing this to my attention, since I haven't actually thought about this case before.

eras/alonzo/impl/testlib/Test/Cardano/Ledger/Alonzo/Binary/CostModelsSpec.hs
teodanciu: Yes completely agree.

eras/alonzo/impl/testlib/Test/Cardano/Ledger/Alonzo/Binary/CostModelsSpec.hs
teodanciu: Totally, it's even used one more time above this line.  Thanks

PR #3863, opened 2023-11-09
Improve deposits refunds re-usability

lehins:
# Description

This PR creates consistent behavior for how we compute deposits and refunds. In particular it adds two functions to `EraTxCert` class that are responsible for getting the refunds and deposits from a list of certificates:

* `getTotalDepositsTxCerts`
* `getTotalRefundsTxCerts`

There are now also matching `EraTxBody` functions:

* `getTotalDepositsTxBody`
* `getTotalRefundsTxBody`

Which will call into `getTotalDepositsTxCerts` and `getTotalRefundsTxBody` respectively for all eras except Conway, because in Conway is the first era where we have one more place where we collect deposits from besides the certificates, namely we also called deposits from proposal procedures.

Most important change in this PR is that it makes deposits and refunds computation and location of implementation logic is consistent. In particular it discourages usage of era specific function for refunds and deposits calculation, except for instance declaration of corresponding functions.

Other improvements include:
* Reduction in duplication in UTXOS/UTXO rules
* Reduction of duplication in `EraUTxO` instances.
* Fixing constraint based generator to account for DRep deposits

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/eras/allegra/impl/src/Cardano/Ledger/Allegra/Rules/Utxo.hs
+++ b/eras/allegra/impl/src/Cardano/Ledger/Allegra/Rules/Utxo.hs
@@ -43,7 +43,6 @@ import Cardano.Ledger.Binary (
   invalidKey,
   serialize,
  )
-import Cardano.Ledger.CertState (certsTotalDepositsTxBody, certsTotalRefundsTxBody)
 import Cardano.Ledger.Coin (Coin)
 import Cardano.Ledger.Core
 import Cardano.Ledger.Crypto (Crypto)
@@ -163,33 +162,33 @@ utxoTransition ::
   ) =>
   TransitionRule (AllegraUTXO era)
 utxoTransition = do
-  TRC (Shelley.UtxoEnv slot pp dpstate genDelegs, u, tx) <- judgmentContext
-  let Shelley.UTxOState utxo _ _ ppup _ _ = u
-  let txb = tx ^. bodyTxL
+  TRC (Shelley.UtxoEnv slot pp certState genDelegs, utxos, tx) <- judgmentContext
+  let Shelley.UTxOState utxo _ _ ppup _ _ = utxos
+  let txBody = tx ^. bodyTxL
 
   {- ininterval slot (txvld tx) -}
-  runTest $ validateOutsideValidityIntervalUTxO slot txb
+  runTest $ validateOutsideValidityIntervalUTxO slot txBody
 
   {- txins txb ≠ ∅ -}
   -- runValidationTransMaybe fromShelleyFailure $ Shelley.validateInputSetEmptyUTxO txb
-  runTest $ Shelley.validateInputSetEmptyUTxO txb
+  runTest $ Shelley.validateInputSetEmptyUTxO txBody
 
   {- minfee pp tx ≤ txfee txb -}
   runTest $ Shelley.validateFeeTooSmallUTxO pp tx
 
   {- txins txb ⊆ dom utxo -}
-  runTest $ Shelley.validateBadInputsUTxO utxo $ txb ^. inputsTxBodyL
+  runTest $ Shelley.validateBadInputsUTxO utxo $ txBody ^. inputsTxBodyL
 
   netId <- liftSTS $ asks networkId
 
   {- ∀(_ → (a, _)) ∈ txouts txb, netId a = NetworkId -}
-  runTest $ Shelley.validateWrongNetwork netId . toList $ txb ^. outputsTxBodyL
+  runTest $ Shelley.validateWrongNetwork netId . toList $ txBody ^. outputsTxBodyL
 
   {- ∀(a → ) ∈ txwdrls txb, netId a = NetworkId -}
-  runTest $ Shelley.validateWrongNetworkWithdrawal netId txb
+  runTest $ Shelley.validateWrongNetworkWithdrawal netId txBody
 
   {- consumed pp utxo txb = produced pp poolParams txb -}
-  runTest $ Shelley.validateValueNotConservedUTxO pp utxo dpstate txb
+  runTest $ Shelley.validateValueNotConservedUTxO pp utxo certState txBody
 
   -- process Protocol Parameter Update Proposals
   ppup' <-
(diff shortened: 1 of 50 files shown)


eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Utxo.hs
```diff
@@ -611,29 +610,36 @@ validateMaxTxSizeUTxO pp tx =
...
   UTxOState era ->
   TxBody era ->
-  Coin ->
+  CertState era ->
   GovState era ->
-  UTxOState era
-updateUTxOState pp UTxOState {utxosUtxo, utxosDeposited, utxosFees, utxosStakeDistr, utxosDonation} txb depositChange govState =
-  let UTxO utxo = utxosUtxo
-      !utxoAdd = txouts txb -- These will be inserted into the UTxO
+  (SafeHash (EraCrypto era) EraIndependentTxBody -> Coin -> m ()) ->
```
Soupstraw:
```suggestion
  (Coin -> m ()) ->
```
The TxBody hash seems redundant, since whatever's using this function already has access to the transaction body.

Soupstraw: Actually, I think it'd be better if `updateUTxOState` was split into two functions: one that computes `depositChange` and then another that updates the UTxOState. I'm just not a big fan of callbacks :sweat_smile:

lehins: Done

lehins: >  I'm just not a big fan of callbacks

I love callbacks

eras/allegra/impl/src/Cardano/Ledger/Allegra/Rules/Utxo.hs
```diff
@@ -212,10 +211,8 @@ utxoTransition = do
   {- txsize tx ≤ maxTxSize pp -}
   runTest $ Shelley.validateMaxTxSizeUTxO pp tx
 
-  let refunded = getTotalRefundsTxBody pp dpstate txb
-  let depositChange = getTotalDepositsTxBody pp dpstate txb Val.<-> refunded
-  tellEvent $ TotalDeposits (hashAnnotated txb) depositChange
-  pure $! Shelley.updateUTxOState pp u txb depositChange ppup'
+  Shelley.updateUTxOState pp utxos txBody certState ppup' $
+    tellEvent . TotalDeposits (hashAnnotated txBody)
```
TimSheard, line 215: So I can't figure out where the computation of refunds and deposits happens in the new code.

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Utxo.hs
```diff
@@ -611,29 +610,36 @@ validateMaxTxSizeUTxO pp tx =
...
-        , utxosDonation = utxosDonation
-        }
+      totalRefunds = certsTotalRefundsTxBody pp certState txBody
+      totalDeposits = certsTotalDepositsTxBody pp certState txBody
+      depositChange = totalDeposits <-> totalRefunds
+  depositChangeEvent depositChange
+  pure $!
+    UTxOState
+      { utxosUtxo = UTxO newUTxO
+      , utxosDeposited = utxosDeposited <> depositChange
```
TimSheard, line 642: Moving the computation of deposits and refunds to the function that makes the changes to the UTxOState make sense, it is a natural extension to 'smartUTxOState, in that it automatically makes the invariants true. Unlike smartUTxOState it is NOT really expensive.  But to someone who is just reading this PR, or someone who is just using unpdateUTxOState a lot of implicit changes happen that user is not aware of. We should add some comments to the defintion of updateUTxOState to make this clear.

libs/cardano-ledger-core/src/Cardano/Ledger/Core/TxCert.hs
```diff
@@ -87,6 +89,26 @@ class
...
+  -- | Compute the total refunds from a list of certificates.
+  getTotalRefundsTxCerts ::
+    Foldable f =>
+    PParams era ->
+    -- | Lookup current deposit for Staking credential if one is registered
+    (Credential 'Staking (EraCrypto era) -> Maybe Coin) ->
+    -- | Lookup current deposit for DRep credential if one is registered
+    (Credential 'DRepRole (EraCrypto era) -> Maybe Coin) ->
+    f (TxCert era) ->
+    Coin
```
TimSheard, line 110: Just a thought (maybe not a good one). But have we considered having a default implementation for these. After all, this is the same for all era's before Conway.

TimSheard (approved):
Looks good

libs/cardano-ledger-core/src/Cardano/Ledger/Core/TxCert.hs
lehins: I certainly did give it a thought on multiple occasions

Default implementation would be good if it was the same for all future eras, rather than the first few. Because we wouldn't need to worry about specifying it for new instances. Providing a default implementation that works only for the first few eras could lead to a dangerous state where we forget to supply the correct one in the future era. 

Although I totally understand the desire to do that, since it would save a few lines of code. I am guilty of doing just that on multiple occasions, due to my laziness.

eras/allegra/impl/src/Cardano/Ledger/Allegra/Rules/Utxo.hs
lehins: In the `updateUTxOState`, which I see that you've discovered it a bit later

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Utxo.hs
lehins: I am certainly on board with adding some haddock explaining this function.

lehins: Done

PR #3867, opened 2023-11-10
`MaryValue` fixes

lehins:
# Description

This PR split into 3 major commits (thus it probably would be easier to review commit by commit)

* improves MaryValue generators
* fixes MaryValue deserialization. It reverts back the enforcement of positive ADA in the MaryValue back to non-negative ADA. This must be the case due to possibility of zero ADA in `ValueNotConservedUTxO`.
* Changes type for ADA in `MaryValue` from `Integer` to `Coin`

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/eras/conway/test-suite/cardano-ledger-conway-test.cabal
+++ b/eras/conway/test-suite/cardano-ledger-conway-test.cabal
@@ -39,7 +39,7 @@ library
         cardano-ledger-conway:{cardano-ledger-conway, testlib} >=1.9 && <1.12,
         cardano-ledger-core:{cardano-ledger-core, testlib} >=1.8.1,
         cardano-ledger-allegra ^>=1.2,
-        cardano-ledger-mary ^>=1.3,
+        cardano-ledger-mary ^>=1.4,
         cardano-ledger-shelley-ma-test >=1.1,
         cardano-ledger-shelley-test >=1.1,
         cardano-ledger-shelley >=1.6,
--- a/eras/mary/impl/src/Cardano/Ledger/Mary/Value.hs
+++ b/eras/mary/impl/src/Cardano/Ledger/Mary/Value.hs
@@ -58,7 +58,7 @@ import Cardano.Ledger.Binary.Coders (
   (<!),
  )
 import Cardano.Ledger.Binary.Version (natVersion)
-import Cardano.Ledger.Coin (Coin (..), CompactForm (..), decodePositiveCoin, integerToWord64, word64ToCoin)
+import Cardano.Ledger.Coin (Coin (..), decodePositiveCoin, integerToWord64, word64ToCoin)
 import Cardano.Ledger.Compactible (Compactible (..))
 import Cardano.Ledger.Crypto (Crypto (ADDRHASH))
 import Cardano.Ledger.Shelley.Scripts (ScriptHash (..))
@@ -183,7 +183,7 @@ instance Crypto c => EncCBOR (MultiAsset c) where
   encCBOR = encodeMultiAssetMaps
 
 -- | The Value representing MultiAssets
-data MaryValue c = MaryValue !Integer !(MultiAsset c)
+data MaryValue c = MaryValue !Coin !(MultiAsset c)
   deriving (Show, Generic)
 
 instance Crypto c => Eq (MaryValue c) where
@@ -196,15 +196,15 @@ instance NoThunks (MaryValue c)
 
 instance Semigroup (MaryValue c) where
   MaryValue c1 m1 <> MaryValue c2 m2 =
-    MaryValue (c1 + c2) (m1 <> m2)
+    MaryValue (c1 <> c2) (m1 <> m2)
 
 instance Monoid (MaryValue c) where
-  mempty = MaryValue 0 mempty
+  mempty = MaryValue mempty mempty
 
 instance Group (MaryValue c) where
   invert (MaryValue c m) =
     MaryValue
-      (-c)
+      (invert c)
       (invert m)
 
 instance Abelian (MaryValue c)
@@ -215,13 +215,14 @@ instance Abelian (MaryValue c)
 instance Crypto c => Val (MaryValue c) where
   s <×> MaryValue c (MultiAsset m) =
     MaryValue
-      (fromIntegral s * c)
+      (s <×> c)
       (MultiAsset (canonicalMap (canonicalMap (fromIntegral s *)) m))
-  isZero (MaryValue c (MultiAsset m)) = c == 0 && Map.null m
-  coin (MaryValue c _) = Coin c
-  inject (Coin c) = MaryValue c (MultiAsset Map.empty)
-  modifyCoin f (MaryValue c m) = MaryValue n m where (Coin n) = f (Coin c)
-  pointwise p (MaryValue c (MultiAsset x)) (MaryValue d (MultiAsset y)) = p c d && pointWise (pointWise p) x y
+  isZero (MaryValue c (MultiAsset m)) = c == zero && Map.null m
+  coin (MaryValue c _) = c
+  inject c = MaryValue c (MultiAsset Map.empty)
+  modifyCoin f (MaryValue c m) = MaryValue (f c) m
+  pointwise p (MaryValue (Coin c) (MultiAsset x)) (MaryValue (Coin d) (MultiAsset y)) =
+    p c d && pointWise (pointWise p) x y
 
   -- returns the size, in Word64's, of the CompactValue representation of MaryValue
   size vv@(MaryValue _ (MultiAsset m))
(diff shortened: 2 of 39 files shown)


eras/mary/impl/testlib/Test/Cardano/Ledger/Mary/Arbitrary.hs
```diff
@@ -205,15 +193,30 @@ genEmptyMultiAsset :: Crypto c => Gen (MultiAsset c)
 genEmptyMultiAsset =
   MultiAsset <$> genNonEmptyMap arbitrary (pure Map.empty)
 
+-- | Better generator for a positive Int that explores more values
+genPositiveInt :: Gen Int
+genPositiveInt =
+  oneof
+    [ choose (1, maxBound)
+    , getPositive <$> arbitrary
```
teodanciu, line 211: `Positive` doesn't generate until `maxBound Int` ? Or in what way are there more values explored?

eras/conway/test-suite/cardano-ledger-conway-test.cabal
```diff
@@ -39,7 +39,7 @@ library
         cardano-ledger-conway:{cardano-ledger-conway, testlib} >=1.9 && <1.12,
         cardano-ledger-core:{cardano-ledger-core, testlib} >=1.8.1,
         cardano-ledger-allegra ^>=1.2,
-        cardano-ledger-mary ^>=1.3,
+        cardano-ledger-mary >=1.3,
```
teodanciu: Not  `>= 1.4`  ?

eras/mary/impl/testlib/Test/Cardano/Ledger/Mary/Arbitrary.hs
lehins: It might come as a surprise, but it is a lot worse than that. It will not even generate values over a hundred with default quickcheck size value of 30:
```haskell
ghci> sample (getPositive <$> (arbitrary :: Gen (Positive Int)))
1
1
1
3
1
10
10
12
8
16
9
```

teodanciu: Wow!  
But why then use it at all in `oneof`  alongside with `choose  (1, maxBound)` ?  Is it to do with the distribution of the generated values?

lehins: >  Is it to do with the distribution of the generated values?

Correct, because small values that are of interest like `1` and `2` are unlikely to be generated with uniform distribution

teodanciu: Interesting, thank you!

eras/conway/test-suite/cardano-ledger-conway-test.cabal
lehins: You are right. Thank you for catching it, as always :wink: 
Fixed

eras/mary/impl/src/Cardano/Ledger/Mary/Value.hs
```diff
@@ -286,64 +282,55 @@ roundupBytesToWords b = quot (b + wordLength - 1) wordLength
...
-decodeMultiAssetMint :: Crypto c => Decoder s (MultiAsset c)
-decodeMultiAssetMint = decodeMultiAssetMaps decodeIntegerBounded64
+        decodeWithEnforcing
+        decodeWithPrunning
+  ma <$ unless (isMultiAssetSmallEnough ma) (fail "MultiAsset is too big to compact")
+  where
+    decodeWithEnforcing =
+      decodeMap decCBOR $ do
+        m <- decodeMap decCBOR $ do
+          amount <- decodeAmount
```
teodanciu, line 329: Something I don't understand here: 
In the previous version, the value-Decoder we passed  to `decodeMap` (second argument) was `(decodeMap decCBOR decodeAmount)` , and after that we would do enforcing.  But in the new version, the decoder seems to be `decodeAmount` (plus the enforcing).  Are they equivalent ? (sorry if it's obvious, I'm probably just missing it)

lehins: I am really sorry, I don't understand the question. It's a bit ambiguous. We are enforcing two things:
* Non-zero amounts
* Non-empty maps.

Which enforcement are you asking about?

lehins: In either way, decoders are equivalent, since deserialization specification of `MultiAsset` didn't change in this PR

eras/mary/impl/src/Cardano/Ledger/Mary/Value.hs
```diff
@@ -177,13 +176,10 @@ instance Group (MultiAsset c) where
     MultiAsset (canonicalMap (canonicalMap ((-1 :: Integer) *)) m)
 
 instance Crypto c => DecCBOR (MultiAsset c) where
-  decCBOR = decodeMultiAssetMint
-
-instance Crypto c => EncCBOR (MultiAsset c) where
-  encCBOR = encodeMultiAssetMaps
+  decCBOR = decodeMultiAsset decodeIntegerBounded64
```
TimSheard, line 179: What happend to the EncCBOR (MutiAsset c))  instance? It seems to have disappeared

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Utxo.hs
```diff
@@ -218,23 +218,23 @@ instance
...
     BadInputsUTxO ins ->
       encodeListLen 2 <> encCBOR (0 :: Word8) <> encCBOR ins
-    (ExpiredUTxO a b) ->
+    ExpiredUTxO a b ->
       encodeListLen 3
         <> encCBOR (1 :: Word8)
         <> encCBOR a
         <> encCBOR b
-    (MaxTxSizeUTxO a b) ->
+    MaxTxSizeUTxO a b ->
```
TimSheard, line 226: Yes we hate those parentheses

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Examples/AlonzoValidTxUTXOW.hs
```diff
@@ -612,7 +620,7 @@ notValidatingWithMintTx pf =
         pf
         [ Inputs' [mkGenesisTxIn 8]
         , Collateral' [mkGenesisTxIn 18]
-        , Outputs' [newTxOut pf [Address (someAddr pf), Amount (MaryValue 0 mint <+> inject (Coin 995))]]
+        , Outputs' [newTxOut pf [Address (someAddr pf), Amount (MaryValue mempty mint <+> inject (Coin 995))]]
```
TimSheard, line 623: In other files you preferred this format
`Outputs' [newTxOut pf [Address (someAddr pf), Amount (MaryValue (Coin 995) mint)]]`
why leave the inject here>

TimSheard (approved):
Looks good to me. The only thing that looks strange is the missing EncCBOR instance.

eras/mary/impl/src/Cardano/Ledger/Mary/Value.hs
teodanciu: I think it has moved in the newtype declaration above  
```
newtype MultiAsset c = MultiAsset (Map (PolicyID c) (Map AssetName Integer))
  deriving (Show, Generic, ToJSON, EncCBOR)
```

since `encodeMultiAssetMaps` was only doing `encCBOR m`

eras/mary/impl/src/Cardano/Ledger/Mary/Value.hs
teodanciu: I'm sorry, I missed that there are two lines containing `decodeMap decCBOR $ do`, at 327 and 328, so I was under the impression that in the case of  `decodeWithEnforcing`, we are passing simply `decodeAmount` as a decoder for the value of the map, rather than `decodeMap decCBOR decodeAmount`.  Sorry!

teodanciu (approved):
Looks good to me

eras/mary/impl/src/Cardano/Ledger/Mary/Value.hs
lehins: Yes. It can be derived with GeneralizedNewtypeDeriving, no need to specify it manually

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Utxo.hs
lehins: :smile: 
Yes we do

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Examples/AlonzoValidTxUTXOW.hs
lehins: I didn't immediately noticed that `inject` and `<+>` are redundant.

In order to avoid running CI once again, I'll leave it as is. I might fix it some other point.

PR #3869, opened 2023-11-13
Indicate that tag 258 is optional for OSet. Fix rational CDDL

lehins:
# Description

* #3860 added optional 258 tags to sets, but not the oset in CDDL. This PR fixes that ommission
* Fix definition of `rational` in CDDL. It is no longer non-negative
* Remove `positive_rational` in favor of new `nonnegative_rational`, since that is the actual type that is used in PParams
* Switch `ppCommitteeMaxTermLength` to `EpochNo`, rather than `Natural`

# Checklist
- [x] Self-reviewed the diff

### DIFF
--- a/eras/conway/impl/CHANGELOG.md
+++ b/eras/conway/impl/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## 1.11.0.0
 
+* Switch `ppCommitteeMaxTermLength` to `EpochNo`, rather than `Natural`
 * Rename data-type `ProposalsSnapshot` to `Proposals`. #3859
   * Rename module `Governance.Snapshots` to `Governance.Proposals`.
   * Rename all the functions related to the data-type.
--- a/eras/conway/impl/cddl-files/extra.cddl
+++ b/eras/conway/impl/cddl-files/extra.cddl
@@ -1,8 +1,12 @@
+; Conway era introduces an optional 258 tag for sets, which will become mandatory in the
+; next era. We recommend all the tooling to account for this future breaking change now,
+; rather than in the next era when it will be enforced.
+
 set<a> = #6.258([* a]) / [* a]
 
 nonempty_set<a> = #6.258([+ a]) / [+ a]
 
-nonempty_oset<a> = #6.258([+ a])
+nonempty_oset<a> = #6.258([+ a]) / [+ a]
 
 positive_int = 1 .. 18446744073709551615
 
--- a/eras/conway/impl/cddl-files/extra.cddl
+++ b/eras/conway/impl/cddl-files/extra.cddl
@@ -15,9 +15,9 @@ unit_interval = #6.30([1, 2])
   ; but this produces numbers outside the unit interval
   ; and can also produce a zero in the denominator
 
-rational =  #6.30([uint, positive_int])
+rational =  #6.30([int, positive_int])
 
-positive_interval = #6.30([uint, positive_int])
+nonnegative_interval = #6.30([uint, positive_int])
 
 
 address =
--- a/eras/conway/impl/src/Cardano/Ledger/Conway/PParams.hs
+++ b/eras/conway/impl/src/Cardano/Ledger/Conway/PParams.hs
@@ -221,7 +221,6 @@ data ConwayPParams f era = ConwayPParams
   , cppCommitteeMinSize :: !(THKD 'GovGroup f Natural)
   -- ^ Minimum size of the Constitutional Committee
   , cppCommitteeMaxTermLength :: !(THKD 'GovGroup f Natural) -- TODO: This too should be EpochNo
-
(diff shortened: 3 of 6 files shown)


eras/conway/impl/CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## 1.11.0.0
 
+* Switch `ppCommitteeMaxTermLength` to `EpochNo`, rather than `Natural`
```
TimSheard, line 5: I think this is not a usefull change.  An EpochNo is a distinct point in time. ppCommitteeMaxTermLength is the length of an interval in this space. So if we use EpochNo we have mixed up the units.  If any thing the interval could be the underlying Num type in EpochNo which is Integer, but then we could have negative intervals, So I think Natural is the right thing.

eras/conway/impl/cddl-files/extra.cddl
```diff
@@ -11,9 +15,9 @@ unit_interval = #6.30([1, 2])
   ; but this produces numbers outside the unit interval
   ; and can also produce a zero in the denominator
 
-rational =  #6.30([uint, positive_int])
+rational =  #6.30([int, positive_int])
 
-positive_interval = #6.30([uint, positive_int])
+nonnegative_interval = #6.30([uint, positive_int])
```
TimSheard, line 20: Are we replacing positive_interval with nonnegative_interval, or is the intent to support both
positive_interval and nonnegative interval?

eras/conway/impl/src/Cardano/Ledger/Conway/PParams.hs
```diff
@@ -260,7 +258,7 @@ data UpgradeConwayPParams f = UpgradeConwayPParams
   { ucppPoolVotingThresholds :: !(HKD f PoolVotingThresholds)
   , ucppDRepVotingThresholds :: !(HKD f DRepVotingThresholds)
   , ucppCommitteeMinSize :: !(HKD f Natural)
-  , ucppCommitteeMaxTermLength :: !(HKD f Natural)
+  , ucppCommitteeMaxTermLength :: !(HKD f EpochNo)
```
TimSheard, line 261: like I said above, it should not be a point in time, but an interval

TimSheard (approved):
Looks good, but I question why we changed from Natural  to EpochNo

eras/conway/impl/CHANGELOG.md
teodanciu: I agree with Tim, having it as Epoch makes it look as it could mean the Epoch representing the end of the term.

teodanciu: Is there any particular reason for the change?

lehins: I agree with the fact that this should not be an `EpochNo`, but I do not agree that it should be a `Natural`. There is a bigger reason why it has to be a different, a narrower type, something like `Word32`, because an improper PParam value will lead to an integer overflow. We need a new type `EpochNoDelta` or `EpochNoInterval` or something, to indicate the intent. I've created a ticket that should take care of this properly: #3870 

That being said, the reason why I switched it to `EpochNo` is for consistency with other pparams that indicate intervals in number of epochs:
  * `EMax`: https://github.com/input-output-hk/cardano-ledger/blob/cf32eab4eacebbec9817000827036822d376f35d/eras/conway/impl/src/Cardano/Ledger/Conway/PParams.hs#L183
  * `govActionLifetime`: https://github.com/input-output-hk/cardano-ledger/blob/cf32eab4eacebbec9817000827036822d376f35d/eras/conway/impl/src/Cardano/Ledger/Conway/PParams.hs#L226
  * `DRepActivity`: https://github.com/input-output-hk/cardano-ledger/blob/cf32eab4eacebbec9817000827036822d376f35d/eras/conway/impl/src/Cardano/Ledger/Conway/PParams.hs#L232
  
I'll merge this change, just so we have at least some sort of consistency for the moment.

eras/conway/impl/cddl-files/extra.cddl
lehins: We have no uses of `positive_interval` anywhere in our types, so that's why I replaced it with `nonnegative_interval`

eras/conway/impl/src/Cardano/Ledger/Conway/PParams.hs
lehins: Agreed. I've captured the proper fix in this ticket: https://github.com/input-output-hk/cardano-ledger/issues/3870

eras/conway/impl/CHANGELOG.md
teodanciu: makes sense

teodanciu (approved):
Looks good to me

PR #3871, opened 2023-11-15
Bump plutus dep to 1.16

michaelpj:
Routine bump, lets us get rid of some instances, and supports newer `nothunks`.

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -18,9 +18,9 @@ source-repository-package
 
 index-state:
   -- Bump this if you need newer packages from Hackage
-  , hackage.haskell.org 2023-11-09T23:50:15Z
+  , hackage.haskell.org 2023-11-15T00:00:00Z
   -- Bump this if you need newer packages from CHaP
-  , cardano-haskell-packages 2023-11-08T10:13:53Z
+  , cardano-haskell-packages 2023-11-15T00:00:00Z
 
 packages:
   eras/allegra/impl
@@ -65,11 +65,6 @@ packages:
   libs/cardano-ledger-test
   libs/plutus-preprocessor
   libs/ledger-state
-
-constraints:
-  -- Plutus is incompatible with 0.1.5 because of the Identity instance, once
-  -- they fix that we shall remove this constraint.
-  nothunks == 0.1.4
 
 program-options
   ghc-options: -Werror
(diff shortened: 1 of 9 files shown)


cabal.project
```diff
@@ -18,9 +18,9 @@ source-repository-package
 
 index-state:
   -- Bump this if you need newer packages from Hackage
-  , hackage.haskell.org 2023-11-09T23:50:15Z
+  , hackage.haskell.org 2023-11-15T00:00:00Z
```
jasagredo, line 21: I think the ledger team prefers to set this to an existing index state, to avoid getting the cabal warning.

michaelpj: It's a time in the past, isn't that sufficient?

jasagredo: Nope
```
Warning: Requested index-state 2023-11-15T00:00:00Z is newer than
'hackage.haskell.org'! Falling back to older state (2023-11-14T11:26:13Z).
Warning: Requested index-state 2023-11-15T00:00:00Z is newer than
'cardano-haskell-packages'! Falling back to older state
(2023-11-14T12:07:29Z).
```

jasagredo: Note this warning will be downgraded when this PR gets into the released Cabal https://github.com/haskell/cabal/pull/8944 but for now a warning is emitted.

lehins: If I understand correctly this whole nix/cabal index thing `2023-11-15T00:00:00Z` is the correct time:
https://github.com/input-output-hk/hackage.nix/blob/f7760ad0bc4014d7fb0786326d14842e9ebd0641/index-state-hashes.nix#L1691

lehins: I guess I was wrong. Then I definitely do not understand it :see_no_evil:

lehins (approved):
Looks good to me. I am not gonna worry about the cabal warning.

cabal.project
jasagredo: https://github.com/input-output-hk/cardano-ledger/pull/3871#pullrequestreview-1731920543

Let's resolve this then

lehins:
@michaelpj We prefer rebase instead of commits that merge master into ledger

This results in a nice linear history

I've fixed this in this PR by merging this PR with a squashing, but just letting you know for the future.

michaelpj:
Ah right, I just hit the "update branch" button because it was there!

PR #3877, opened 2023-11-17
Make it build with ghc-9.8

erikd:
# Description

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/.github/workflows/haskell.yml
+++ b/.github/workflows/haskell.yml
@@ -42,7 +42,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        ghc: ["8.10.7", "9.2.8", "9.6.2"]
+        ghc: ["8.10.7", "9.2.8", "9.6.3", "9.8.1"]
         os: [ubuntu-latest]
 
     steps:
--- a/cabal.project
+++ b/cabal.project
@@ -18,9 +18,9 @@ source-repository-package
 
 index-state:
   -- Bump this if you need newer packages from Hackage
-  , hackage.haskell.org 2023-11-15T00:00:00Z
+  , hackage.haskell.org 2024-01-04T21:59:02Z
   -- Bump this if you need newer packages from CHaP
-  , cardano-haskell-packages 2023-12-24T05:55:00Z
+  , cardano-haskell-packages 2024-01-04T09:36:23Z
 
 packages:
   eras/allegra/impl
@@ -66,8 +66,10 @@ packages:
   libs/plutus-preprocessor
   libs/ledger-state
 
-program-options
-  ghc-options: -Werror
+-- GHC 9.8 introduces new warnings on partial functions.
+if impl(ghc < 9.8)
+  program-options
+    ghc-options: -Werror
 
 package plutus-preprocessor
   haddock-options: "--optghc=-fplugin-opt PlutusTx.Plugin:defer-errors"
--- a/eras/byron/chain/executable-spec/byron-spec-chain.cabal
+++ b/eras/byron/chain/executable-spec/byron-spec-chain.cabal
@@ -1,6 +1,6 @@
 cabal-version:      3.0
 name:               byron-spec-chain
-version:            1.0.0.1
+version:            1.0.0.2
 license:            Apache-2.0
 maintainer:         operations@iohk.io
 author:             IOHK
@@ -27,7 +27,7 @@ library
         -Wincomplete-uni-patterns -Wredundant-constraints -Wunused-packages
 
     build-depends:
-        base >=4.14 && <4.19,
+        base >=4.14 && <5,
         bimap >=0.4 && <0.6,
         bytestring,
         containers,
--- a/eras/byron/crypto/cardano-crypto-wrapper.cabal
+++ b/eras/byron/crypto/cardano-crypto-wrapper.cabal
@@ -1,6 +1,6 @@
 cabal-version:      3.0
 name:               cardano-crypto-wrapper
-version:            1.5.1.0
+version:            1.5.1.1
 license:            Apache-2.0
 maintainer:         operations@iohk.io
 author:             IOHK
@@ -73,7 +73,7 @@ library
         -Wno-safe -Wno-unsafe -Wunused-packages
 
     build-depends:
-        base >=4.14 && <4.19,
+        base >=4.14 && <5,
         aeson,
         base16-bytestring >=1,
         base64-bytestring,
--- a/eras/byron/crypto/test/cardano-crypto-test.cabal
+++ b/eras/byron/crypto/test/cardano-crypto-test.cabal
@@ -1,6 +1,6 @@
 cabal-version: 3.0
 name:          cardano-crypto-test
-version:       1.5.0.0
+version:       1.5.0.1
 license:       Apache-2.0
 maintainer:    operations@iohk.io
 author:        IOHK
@@ -51,7 +51,7 @@ library
         -Wno-safe -Wno-unsafe -Wunused-packages
 
     build-depends:
-        base >=4.14 && <4.19,
+        base >=4.14 && <5,
         bytestring,
         cardano-ledger-binary:{cardano-ledger-binary, testlib} >=1.0,
         cardano-crypto,
--- a/eras/byron/ledger/executable-spec/byron-spec-ledger.cabal
+++ b/eras/byron/ledger/executable-spec/byron-spec-ledger.cabal
@@ -1,6 +1,6 @@
 cabal-version:      3.0
 name:               byron-spec-ledger
-version:            1.0.0.1
+version:            1.0.0.2
 license:            Apache-2.0
 maintainer:         operations@iohk.io
 author:             IOHK
@@ -36,7 +36,7 @@ library
         -Wincomplete-uni-patterns -Wredundant-constraints -Wunused-packages
 
     build-depends:
-        base >=4.14 && <4.19,
+        base >=4.14 && <5,
         bimap >=0.4 && <0.6,
         bytestring,
         cardano-crypto-class,
--- a/eras/conway/impl/cardano-ledger-conway.cabal
+++ b/eras/conway/impl/cardano-ledger-conway.cabal
@@ -1,6 +1,6 @@
 cabal-version:      3.0
 name:               cardano-ledger-conway
-version:            1.12.0.0
+version:            1.12.0.1
 license:            Apache-2.0
 maintainer:         operations@iohk.io
 author:             IOHK
@@ -71,7 +71,7 @@ library
         -Wunused-packages
 
     build-depends:
-        base >=4.14 && <4.19,
+        base >=4.14 && <5,
         aeson >=2.2,
         data-default-class,
         cardano-crypto-class,
--- a/eras/shelley-ma/test-suite/cardano-ledger-shelley-ma-test.cabal
+++ b/eras/shelley-ma/test-suite/cardano-ledger-shelley-ma-test.cabal
@@ -34,12 +34,12 @@ library
     hs-source-dirs:   src
     default-language: Haskell2010
     ghc-options:
-        -Wall -Wcompat -Werror -Wincomplete-record-updates
+        -Wall -Wcompat -Wincomplete-record-updates
         -Wincomplete-uni-patterns -Wpartial-fields -Wredundant-constraints
         -Wunused-packages
 
     build-depends:
-        base >=4.14 && <4.19,
+        base >=4.14 && <5,
         bytestring,
         cardano-ledger-binary:{cardano-ledger-binary, testlib} >=1.2,
         cardano-ledger-core:{cardano-ledger-core, testlib} >=1.3,
@@ -79,7 +79,7 @@ test-suite cardano-ledger-shelley-ma-test
 
     default-language: Haskell2010
     ghc-options:
-        -Wall -Wcompat -Werror -Wincomplete-record-updates
+        -Wall -Wcompat -Wincomplete-record-updates
         -Wincomplete-uni-patterns -Wpartial-fields -Wredundant-constraints
         -Wunused-packages -threaded -rtsopts -with-rtsopts=-N
         "-with-rtsopts=-K4m -M350m"
--- a/libs/cardano-ledger-test/cardano-ledger-test.cabal
+++ b/libs/cardano-ledger-test/cardano-ledger-test.cabal
@@ -91,7 +91,7 @@ library
 
     build-depends:
         array,
-        base >=4.14 && <4.19,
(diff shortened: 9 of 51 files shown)


eras/shelley-ma/test-suite/cardano-ledger-shelley-ma-test.cabal
```diff
@@ -34,12 +34,12 @@ library
...
     default-language: Haskell2010
     ghc-options:
-        -Wall -Wcompat -Werror -Wincomplete-record-updates
+        -Wall -Wcompat -Wincomplete-record-updates
         -Wincomplete-uni-patterns -Wpartial-fields -Wredundant-constraints
         -Wunused-packages
 
     build-depends:
-        base >=4.14 && <4.19,
+        base >=4.14 && <4.21,
```
lehins: I suggest switching all packages to `<5` on `base`.

erikd: Done!

cabal.project
```diff
@@ -90,3 +87,20 @@ benchmarks: true
 
 -- The only sensible test display option
 test-show-details: streaming
+
+if impl(ghc < 9.8)
+  program-options
+    ghc-options: -Werror
```
erikd: This is needed because partial functions like `head` and `tail` now emit a warning.

`ghc-options: -Werror` for all compiler versions will need to be restored once the code is fixed.

cabal.project
```diff
@@ -90,3 +87,20 @@ benchmarks: true
...
+
+if impl(ghc >= 9.8)
+  -- This commit is on master of the plutus repo.
+  source-repository-package
```
michaelpj: s-r-ps in conditionals don't work well, unfortunately.

I think you could just make this unconditional: it would be weird to have the ledger only sometimes depend on master of `plutus` for 9.8, in general that might not compile. Whereas making it always depend on master (for this branch which I assume you're not going to merge soon) is at least consistent.

.github/workflows/haskell.yml
```diff
@@ -42,7 +42,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        ghc: ["8.10.7", "9.2.8", "9.6.2"]
+        ghc: ["8.10.7", "9.2.8", "9.6.3","9.8.1"]
```
lehins:
```suggestion
        ghc: ["8.10.7", "9.2.8", "9.6.3", "9.8.1"]
```

cabal.project
lehins: Why are we using SRP at all? The normal process is to make a releass of Plutus packages that are compatible with ghc-9.8 and then use plutus from CHaPs?

> I think you could just make this unconditional

It would create problems when we would release ledger packages to CHaPs. We can't have SRP on master at all.

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/TxGen.hs
```diff
@@ -79,6 +80,9 @@ import Control.State.Transition.Extended hiding (Assertion)
 import Data.Bifunctor (first)
 import qualified Data.Foldable as F
 import Data.Functor
+#if MIN_VERSION_base(4,19,0)
+  hiding (unzip)
+#endif
```
lehins: Instead of using CPP to hide unzip, let's better make an explicit import of functions that are being used in this module from `Data.Functor`
```suggestion
```

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/TxGen.hs
```diff
@@ -1,4 +1,5 @@
 {-# LANGUAGE BangPatterns #-}
+{-# LANGUAGE CPP #-}
```
lehins: CPP should be the last resort. In this case we can avoid it
```suggestion
```

libs/ledger-state/src/Cardano/Ledger/State/UTxO.hs
```diff
@@ -34,6 +35,9 @@ import Control.SetAlgebra (range)
 import qualified Data.ByteString.Lazy as LBS
 import Data.Foldable as F
 import Data.Functor
+#if MIN_VERSION_base(4,19,0)
+  hiding (unzip)
+#endif
```
lehins: Ditto
```suggestion
```

libs/ledger-state/src/Cardano/Ledger/State/UTxO.hs
```diff
@@ -1,4 +1,5 @@
 {-# LANGUAGE BangPatterns #-}
+{-# LANGUAGE CPP #-}
```
lehins:
```suggestion
```

cabal.project
lehins: Didn't you take care of the `head` and `tail` functions in this PR, or there are still some left over uses of those?

lehins (changes requested):
As always, thank you for doing this compatibility work! We greatly appreciate it!

cabal.project
erikd: I made it unconditional.

If you want to do it without SRPs, you will have to wait until `ghc-9.8` compatible versions if the plutus packages hit CHaP.

erikd:
> Didn't you take care of the head and tail functions in this PR, or there are still some left over uses of those?

@lehins I did not touch these functions because I thought I would leave it to the `ledger` maintainers to sort it out how they see fit.

.github/workflows/haskell.yml
erikd: Fixed.

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/TxGen.hs
erikd: CPP removed.

libs/ledger-state/src/Cardano/Ledger/State/UTxO.hs
erikd: CPP removed.

libs/ledger-state/src/Cardano/Ledger/State/UTxO.hs
erikd: CPP removed.

michaelpj:
Yeah, I doubt this needs to be merged soon, and in a week or so there will be a plutus release with some initial 9.8 support (still needing some allow-newers).

erikd:
This PR no longer uses any SRPs other than `cardano-ledger-executable-spec`, so IMO is ready to be merged.

cabal.project
```diff
@@ -17,10 +17,10 @@ source-repository-package
   --sha256: sha256-dQVyvzVhVUOsLUuX8XQvN/wPGmbIvYeAmX7RpnT1gk8=
 
 index-state:
-  -- Bump this if you need newer packages from Hackage
-  , hackage.haskell.org 2023-11-15T00:00:00Z
+  -- Bump both the following dates if you need newer packages from Hackage
```
lehins: This misleading. It is not requirement to update both of the indices to get newest packages from hackage.
```suggestion
  -- Bump this if you need newer packages from Hackage
```

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Upec.hs
```diff
@@ -144,7 +146,7 @@ votedValue (ProposedPPUpdates pup) pps quorumN =
         --   1) `consensus` is empty, or
         --   2) `consensus` has exactly one element.
         1 ->
-          (Just . applyPPUpdates pps . fst . head . Map.toList)
+          (Just . applyPPUpdates pps . fst . fst . fromJust . List.uncons . Map.toList)
```
lehins: Original implementation is pretty bad.  But the suggested one is not better either. Could you please switch back to `head` so we don't forget to fix it properly.
```suggestion
          (Just . applyPPUpdates pps . fst . head . Map.toList)
```

Proper fix is this:
```haskell
   in case Map.keys consensus of
        -- NOTE that `quorumN` is a global constant, and that we require
        -- it to be strictly greater than half the number of genesis nodes.
        -- The keys in the `pup` correspond to the genesis nodes,
        -- and therefore either:
        --   1) `consensus` is empty, or
        --   2) `consensus` has exactly one element.
        thePPups:_ -> Just $ applyPPUpdates pps thePPups
        -- NOTE that `updatePParams` corresponds to the union override right
        -- operation in the formal spec.
        _ -> Nothing
```

libs/cardano-ledger-test/cardano-ledger-test.cabal
```diff
@@ -126,7 +126,7 @@ library
         mtl,
         nothunks,
         hspec,
-        plutus-ledger-api,
+        plutus-ledger-api ^>=1.18,
```
lehins: Omission of bounds here is intentional. We do not release `cardano-ledger-test` to CHaPs. So there is need to put whole bunch of unnecessary bounds in this package
```suggestion
        plutus-ledger-api,
```

eras/byron/chain/executable-spec/byron-spec-chain.cabal
```diff
@@ -27,7 +27,7 @@ library
         -Wincomplete-uni-patterns -Wredundant-constraints -Wunused-packages
 
     build-depends:
-        base >=4.14 && <4.19,
+        base >=4.14 && <5,
```
lehins, line 30: Change like this requires a patch version update. Could you please bump up the version of `byron-spec-chain` in the cabal file and in the CHANGELOG from `1.0.0.1` to `1.0.0.2`

eras/byron/crypto/cardano-crypto-wrapper.cabal
```diff
@@ -73,7 +73,7 @@ library
         -Wno-safe -Wno-unsafe -Wunused-packages
 
     build-depends:
-        base >=4.14 && <4.19,
+        base >=4.14 && <5,
```
lehins, line 76: Change like this requires a patch version update. Could you please bump up the version of `byron-crypto-wrapper` in the cabal file from `1.5.1.0` to `1.5.1.1`

eras/byron/crypto/test/cardano-crypto-test.cabal
```diff
@@ -51,7 +51,7 @@ library
         -Wno-safe -Wno-unsafe -Wunused-packages
 
     build-depends:
-        base >=4.14 && <4.19,
+        base >=4.14 && <5,
```
lehins, line 54: Change like this requires a patch version update. Could you please bump up the version of `cardano-crypto-test` in the cabal file from `1.5.0.0` to `1.5.0.1`

eras/byron/ledger/executable-spec/byron-spec-ledger.cabal
```diff
@@ -36,7 +36,7 @@ library
         -Wincomplete-uni-patterns -Wredundant-constraints -Wunused-packages
 
     build-depends:
-        base >=4.14 && <4.19,
+        base >=4.14 && <5,
```
lehins, line 39: Change like this requires a patch version update. Could you please bump up the version of `byron-spec-ledger` in the cabal file and in the CHANGELOG from `1.0.0.1` to `1.0.0.2`

eras/conway/impl/cardano-ledger-conway.cabal
```diff
@@ -71,7 +71,7 @@ library
         -Wunused-packages
 
     build-depends:
-        base >=4.14 && <4.19,
+        base >=4.14 && <4.30,
```
lehins: Where did `30` come from?
```suggestion
        base >=4.14 && <5,
```

libs/non-integral/non-integral.cabal
```diff
@@ -23,7 +23,7 @@ library
         -Wall -Wcompat -Wincomplete-record-updates
         -Wincomplete-uni-patterns -Wredundant-constraints -Wunused-packages
 
-    build-depends:    base >=4.14 && <4.19
+    build-depends:    base >=4.14 && <5
```
lehins, line 26: Change like this requires a patch version update. Could you please bump up the version of `non-integral` in the cabal file from `1.0.0.0` to `1.0.0.1`

libs/set-algebra/set-algebra.cabal
```diff
@@ -30,7 +30,7 @@ library
         -Wincomplete-uni-patterns -Wredundant-constraints -Wunused-packages
 
     build-depends:
-        base >=4.14 && <4.19,
+        base >=4.14 && <5,
```
lehins, line 33: Change like this requires a patch version update. Could you please bump up the version of `set-algebra` in the cabal file from `1.1.0.1` to `1.1.0.2`

libs/small-steps-test/small-steps-test.cabal
```diff
@@ -30,7 +30,7 @@ library
         -Wincomplete-uni-patterns -Wredundant-constraints -Wunused-packages
 
     build-depends:
-        base >=4.14 && <4.19,
+        base >=4.14 && <5,
```
lehins, line 33: Change like this requires a patch version update. Could you please bump up the version of `small-steps-test` in the cabal file from `1.0.0.1` to `1.0.0.2`.

Also could I ask you to fix the changelog, it looks totally wrong right now:

```diff
-# Version history for `small-steps-test-test`
+# Version history for `small-steps-test`

-## 1.2.0.3
+## 1.0.0.2

-*
+* GHC-9.8 compatibility

-## 1.2.0.2
+## 1.0.0.1

-* GHC-9.6 ompatibility
+* GHC-9.6 compatibility
```

lehins (changes requested):
Some of the changes are unnecessary and there are bunch of version bumps missing. I know that dealing with compatibility is not the most exciting part of writing programs, so thank you for doing this work!

libs/cardano-ledger-test/cardano-ledger-test.cabal
erikd: Fixed.

libs/small-steps-test/small-steps-test.cabal
erikd: Fixed.

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Upec.hs
erikd: Reverted to the original code.

cabal.project
erikd: Fixed.

lehins (approved):
Beautiful. Thank you!

PR #3878, opened 2023-11-17
Committee query - implement  next epoch change

### LINKED ISSUE #3810: Finish `queryCommitteeMembersState`
Part of the query that predicts what will happen to the constitutional committee member is not done yet, because it needed the work in #3759 and #3801

Figure out the `NextEpochChange` and take care of this TODO:

https://github.com/input-output-hk/cardano-ledger/blob/3ac72fabfef48dd7306b24486687c9f6c1e59c78/libs/cardano-ledger-api/src/Cardano/Ledger/Api/State/Query.hs#L210

teodanciu:
# Description

Closes: https://github.com/input-output-hk/cardano-ledger/issues/3810

I'm not convinced about a few things in this implementation: 
1.  describing members from both current and next epoch committee as `Active`  (with the only difference between them being that the former has `NextEpochChange` either `ToBeRemoved` or `NoChangeExpected`  and the latter `ToBeEnacted`).
In the current acceptance of the Query, `Active`  means not expired, so it is technically compatible with the above, but I think it can be confusing. 
What can we do ? 
* Adding another value to this status, say `ActiveNextEpoch` ? Maybe, but it would kinda mix the categories.  Is it possible for a member from the nextCommittee to be already expired? Then which value would this field take, ActiveNextEpoch or Expired?  
* Another option would be to return only the current committee members, and use the nextEpoch committee members to determine if a member will be removed or no change expected. So `ToBeEnacted` would never appear as a value in this implementation. 
2. members that are Unrecognized (so: they have a hot key in the CommitteeState but no corresponding cold key in either the current committee or the committee in the next epoch), will have the field set to `NoChangeExpected`.  Because what else can it be?  `ToBeRemoved` doesn't really match, since they are not in the committee.  `ToBeEnacted` also no, because they are not in the next committee. 
3. At first I thought that expiration of the committee member should be part of the computation of this field, but I have decided against it, because 
   * the expiration dimension is dealt with in another field , namely MemberStatus
   * expired members are not technically removed from the committee (though their vote doesn't count, in the same way as members without a hot key don't count). 
  So i thought it would be better to keep the two separate. 

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/eras/conway/impl/src/Cardano/Ledger/Conway/Governance.hs
+++ b/eras/conway/impl/src/Cardano/Ledger/Conway/Governance.hs
@@ -617,10 +617,9 @@ instance EraPParams (ConwayEra c) => EraGov (ConwayEra c) where
 
   getConstitution g = Just $ g ^. cgEnactStateL . ensConstitutionL
 
-  getCommitteeMembers g =
-    case g ^. cgEnactStateL . ensCommitteeL of
-      SJust Committee {..} -> Just (committeeMembers, committeeQuorum)
-      SNothing -> Nothing
+  getCommitteeMembers g = ensCommitteeMembers (g ^. cgEnactStateL)
+
+  getNextEpochCommitteeMembers g = ensCommitteeMembers (getRatifyState g ^. rsEnactStateL)
 
   curPParamsGovStateL = curPParamsConwayGovStateL
 
@@ -635,6 +634,13 @@ instance EraPParams (ConwayEra c) => EraGov (ConwayEra c) where
       }
 
   getDRepDistr govst = psDRepDistr . fst $ finishDRepPulser (govst ^. drepPulsingStateGovStateL)
+
+ensCommitteeMembers ::
+  EnactState era ->
+  Maybe (Map (Credential 'ColdCommitteeRole (EraCrypto era)) EpochNo, UnitInterval)
+ensCommitteeMembers ens = case ens ^. ensCommitteeL of
+  SJust Committee {..} -> Just (committeeMembers, committeeQuorum)
+  SNothing -> Nothing
 
 class EraGov era => ConwayEraGov era where
   constitutionGovStateL :: Lens' (GovState era) (Constitution era)
--- a/libs/cardano-ledger-api/cardano-ledger-api.cabal
+++ b/libs/cardano-ledger-api/cardano-ledger-api.cabal
@@ -109,6 +109,7 @@ test-suite cardano-ledger-api-test
         base,
         bytestring,
         cardano-ledger-api,
+        data-default-class,
         testlib,
         cardano-ledger-binary,
         cardano-ledger-babbage:testlib,
--- a/libs/cardano-ledger-api/cardano-ledger-api.cabal
+++ b/libs/cardano-ledger-api/cardano-ledger-api.cabal
@@ -97,6 +97,7 @@ test-suite cardano-ledger-api-test
     other-modules:
         Test.Cardano.Ledger.Api.Tx.Out
         Test.Cardano.Ledger.Api.Tx.Body
+        Test.Cardano.Ledger.Api.State.Imp.QuerySpec
         Test.Cardano.Ledger.Api.State.QuerySpec
 
     default-language: Haskell2010
@@ -118,4 +119,5 @@ test-suite cardano-ledger-api-test
         cardano-ledger-shelley:{cardano-ledger-shelley, testlib},
         cardano-strict-containers,
         containers,
-        microlens
+        microlens,
+        microlens-mtl
--- a/libs/cardano-ledger-api/src/Cardano/Ledger/Api/State/Query.hs
+++ b/libs/cardano-ledger-api/src/Cardano/Ledger/Api/State/Query.hs
@@ -56,7 +56,8 @@ import Cardano.Ledger.Shelley.Governance (
     GovState,
     getCommitteeMembers,
     getConstitution,
-    getDRepDistr
+    getDRepDistr,
+    getNextEpochCommitteeMembers
   ),
  )
 import Cardano.Ledger.Shelley.LedgerState
@@ -165,6 +166,8 @@ queryCommitteeMembersState ::
   Maybe (CommitteeMembersState (EraCrypto era))
 queryCommitteeMembersState coldCredsFilter hotCredsFilter statusFilter nes = do
   (comMembers, comQuorum) <- getCommitteeMembers (queryGovState nes)
+  let nextComMembers =
+        maybe Set.empty (Map.keysSet . fst) (getNextEpochCommitteeMembers (queryGovState nes))
   let comStateMembers =
         csCommitteeCreds $
           nes ^. nesEpochStateL . esLStateL . lsCertStateL . certVStateL . vsCommitteeStateL
@@ -176,8 +179,11 @@ queryCommitteeMembersState coldCredsFilter hotCredsFilter statusFilter nes = do
       relevantColdKeys
         | Set.null statusFilter || Set.member Unrecognized statusFilter =
             withFilteredColdCreds $
-              Map.keysSet comMembers
-                `Set.union` Map.keysSet comStateMembers
+              Set.unions
+                [ Map.keysSet comMembers
+                , Map.keysSet comStateMembers
+                , nextComMembers
+                ]
         | otherwise = withFilteredColdCreds $ Map.keysSet comMembers
 
       relevantHotKeys =
@@ -207,9 +213,17 @@ queryCommitteeMembersState coldCredsFilter hotCredsFilter statusFilter nes = do
                 Nothing -> MemberNotAuthorized
                 Just Nothing -> MemberResigned
                 Just (Just hk) -> MemberAuthorized hk
-        -- TODO: implement after pulsing and snapshots are done
-        let nextEpochChange = NoChangeExpected
-        pure $ CommitteeMemberState hkStatus status mbExpiry nextEpochChange
+        pure $ CommitteeMemberState hkStatus status mbExpiry (nextEpochChange coldCred)
+
+      nextEpochChange :: Credential 'ColdCommitteeRole (EraCrypto era) -> NextEpochChange
+      nextEpochChange ck
+        | not inCurrent && inNext = ToBeEnacted
+        | not inNext = ToBeRemoved
+        | otherwise = NoChangeExpected
+        where
+          inCurrent = Map.member ck comMembers
+          inNext = Set.member ck nextComMembers
+
   pure
     CommitteeMembersState
       { csCommittee = cms
--- a/libs/cardano-ledger-api/src/Cardano/Ledger/Api/State/Query.hs
+++ b/libs/cardano-ledger-api/src/Cardano/Ledger/Api/State/Query.hs
@@ -69,6 +69,7 @@ import Cardano.Ledger.UMap (
 import Control.Monad (guard)
 import Data.Map (Map)
 import qualified Data.Map.Strict as Map
+import Data.Maybe (isJust)
 import Data.Set (Set)
 import qualified Data.Set as Set
 import Lens.Micro ((%~), (&), (<&>), (^.))
@@ -219,10 +220,13 @@ queryCommitteeMembersState coldCredsFilter hotCredsFilter statusFilter nes = do
       nextEpochChange ck
         | not inCurrent && inNext = ToBeEnacted
         | not inNext = ToBeRemoved
+        | expiring = ToBeExpired
         | otherwise = NoChangeExpected
         where
-          inCurrent = Map.member ck comMembers
+          lookupCurrent = Map.lookup ck comMembers
+          inCurrent = isJust lookupCurrent
           inNext = Set.member ck nextComMembers
+          expiring = maybe False (== currentEpoch) lookupCurrent
 
   pure
     CommitteeMembersState
--- a/libs/cardano-ledger-api/src/Cardano/Ledger/Api/State/Query.hs
+++ b/libs/cardano-ledger-api/src/Cardano/Ledger/Api/State/Query.hs
@@ -168,7 +168,7 @@ queryCommitteeMembersState ::
 queryCommitteeMembersState coldCredsFilter hotCredsFilter statusFilter nes = do
   (comMembers, comQuorum) <- getCommitteeMembers (queryGovState nes)
   let nextComMembers =
-        maybe Set.empty (Map.keysSet . fst) (getNextEpochCommitteeMembers (queryGovState nes))
+        maybe Map.empty fst (getNextEpochCommitteeMembers (queryGovState nes))
   let comStateMembers =
         csCommitteeCreds $
           nes ^. nesEpochStateL . esLStateL . lsCertStateL . certVStateL . vsCommitteeStateL
@@ -183,7 +183,7 @@ queryCommitteeMembersState coldCredsFilter hotCredsFilter statusFilter nes = do
               Set.unions
                 [ Map.keysSet comMembers
                 , Map.keysSet comStateMembers
-                , nextComMembers
+                , Map.keysSet nextComMembers
                 ]
         | otherwise = withFilteredColdCreds $ Map.keysSet comMembers
 
@@ -220,13 +220,22 @@ queryCommitteeMembersState coldCredsFilter hotCredsFilter statusFilter nes = do
       nextEpochChange ck
         | not inCurrent && inNext = ToBeEnacted
         | not inNext = ToBeRemoved
-        | expiring = ToBeExpired
+        | Just curTerm <- lookupCurrent
+        , Just nextTerm <- lookupNext
+        , curTerm /= nextTerm
+        , -- if the term is adjusted such that it expires in the next epoch,
+          -- we set it to ToBeExpired instead of TermAdjusted
+          not expiringNext =
+            TermAdjusted nextTerm
+        | expiringCurrent || expiringNext = ToBeExpired
         | otherwise = NoChangeExpected
         where
           lookupCurrent = Map.lookup ck comMembers
+          lookupNext = Map.lookup ck nextComMembers
           inCurrent = isJust lookupCurrent
-          inNext = Set.member ck nextComMembers
-          expiring = maybe False (== currentEpoch) lookupCurrent
+          inNext = isJust lookupNext
+          expiringCurrent = lookupCurrent == Just currentEpoch
+          expiringNext = lookupNext == Just currentEpoch
 
   pure
     CommitteeMembersState
--- a/libs/cardano-ledger-api/test/Test/Cardano/Ledger/Api/State/QuerySpec.hs
+++ b/libs/cardano-ledger-api/test/Test/Cardano/Ledger/Api/State/QuerySpec.hs
@@ -278,8 +278,8 @@ propNextEpoch ::
   Expectation
 propNextEpoch nes = do
   withCommitteeInfo nes $
-    \(Committee cm _) (CommitteeState comStateMembers') nextComMembers' noFilterResult -> do
-      let comMembers = Map.keysSet cm
+    \(Committee comMembers' _) (CommitteeState comStateMembers') nextComMembers' noFilterResult -> do
+      let comMembers = Map.keysSet comMembers'
       let comStateMembers = Map.keysSet comStateMembers'
       let nextComMembers = Map.keysSet nextComMembers'
 
@@ -289,8 +289,25 @@ propNextEpoch nes = do
       filterNext ToBeRemoved noFilterResult
         `shouldSatisfy` (\res -> Map.keysSet res == (comMembers `Set.union` comStateMembers) `Set.difference` nextComMembers)
 
-      filterNext NoChangeExpected noFilterResult
-        `shouldSatisfy` (\res -> Map.keysSet res == (comMembers `Set.intersection` nextComMembers))
+      -- members who are both in current and nextCommittee are either ToBeExpired or NoChangeExpected
+      Map.keysSet (filterNext NoChangeExpected noFilterResult)
+        `Set.union` Map.keysSet (filterNext ToBeExpired noFilterResult)
+        `shouldSatisfy` (== (comMembers `Set.intersection` nextComMembers))
+
+      let currentEpoch = csEpochNo noFilterResult
+      let expiring =
+            Map.keysSet $
+              Map.union
+                (Map.filter (== currentEpoch) comMembers')
+                (Map.filter (== currentEpoch) nextComMembers')
+
+      -- members ToBeExpired have the expiry set to currentEpoch, either in the current committee or in the next one
+      Map.keysSet (filterNext ToBeExpired noFilterResult)
+        `shouldSatisfy` (`Set.isSubsetOf` expiring)
+
+      cmsExpiration
+        <$> filterNext NoChangeExpected noFilterResult
+          `shouldSatisfy` all (all (>= currentEpoch + 1))
   where
     filterNext nextEpochChange cms =
       Map.filter
--- a/libs/cardano-ledger-api/test/Test/Cardano/Ledger/Api/State/QuerySpec.hs
+++ b/libs/cardano-ledger-api/test/Test/Cardano/Ledger/Api/State/QuerySpec.hs
@@ -289,9 +289,12 @@ propNextEpoch nes = do
       filterNext ToBeRemoved noFilterResult
         `shouldSatisfy` (\res -> Map.keysSet res == (comMembers `Set.union` comStateMembers) `Set.difference` nextComMembers)
 
-      -- members who are both in current and nextCommittee are either ToBeExpired or NoChangeExpected
-      Map.keysSet (filterNext NoChangeExpected noFilterResult)
-        `Set.union` Map.keysSet (filterNext ToBeExpired noFilterResult)
+      -- members who are both in current and nextCommittee are either ToBeExpired, TermAdjusted or NoChangeExpected
+      Set.unions
+        [ Map.keysSet (filterNext NoChangeExpected noFilterResult)
+        , Map.keysSet (filterNext ToBeExpired noFilterResult)
+        , Map.keysSet (termAdjusted noFilterResult)
+        ]
         `shouldSatisfy` (== (comMembers `Set.intersection` nextComMembers))
 
       let currentEpoch = csEpochNo noFilterResult
@@ -314,6 +317,14 @@ propNextEpoch nes = do
         ( \case
             CommitteeMemberState _ _ _ nextEpochChange' ->
               nextEpochChange == nextEpochChange'
+        )
+        (csCommittee cms)
+    termAdjusted cms =
+      Map.mapMaybeWithKey
+        ( \k cm ->
+            case cm of
+              CommitteeMemberState _ _ _ (TermAdjusted _) -> Just k
+              _ -> Nothing
         )
         (csCommittee cms)
 
--- a/libs/cardano-ledger-api/test/Test/Cardano/Ledger/Api/State/QuerySpec.hs
+++ b/libs/cardano-ledger-api/test/Test/Cardano/Ledger/Api/State/QuerySpec.hs
@@ -2,9 +2,10 @@
 {-# LANGUAGE DataKinds #-}
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE LambdaCase #-}
-{-# LANGUAGE MonoLocalBinds #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE TypeApplications #-}
+{-# LANGUAGE TypeFamilies #-}
+{-# LANGUAGE TypeOperators #-}
 
 module Test.Cardano.Ledger.Api.State.QuerySpec (spec) where
 
@@ -23,8 +24,10 @@ import Cardano.Ledger.Conway (Conway)
 import Cardano.Ledger.Conway.Governance (
   Committee (..),
   ConwayEraGov,
+  ConwayGovState,
   DRepPulsingState (..),
   RatifyState (..),
+  cgEnactStateL,
   ensCommitteeL,
   newEpochStateDRepPulsingStateL,
   rsEnactStateL,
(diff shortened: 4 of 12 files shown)


eras/shelley/impl/cardano-ledger-shelley.cabal
```diff
@@ -1,6 +1,6 @@
 cabal-version:      3.0
 name:               cardano-ledger-shelley
-version:            1.8.0.0
+version:            1.8.0.1
```
lehins: Patch version is only updated when there are non-visible changes. So, if you ever have to create a lower bound like this `>=x.y.z.1` it means there was an issue with the version. patch version should never appear in the bounds.
```suggestion
version:            1.8.1.0
```

libs/cardano-ledger-api/cardano-ledger-api.cabal
```diff
@@ -58,10 +58,10 @@ library
         cardano-ledger-alonzo >=1.5.1,
         cardano-ledger-babbage >=1.1,
         cardano-ledger-binary >=1.0,
-        cardano-ledger-conway >=1.10,
+        cardano-ledger-conway >=1.11.0.1,
```
lehins: Same here, you see the patch version in the bounds. It means it is wrong

lehins:
> 1. .... What can we do ?

If you look at the haddock for `Unrecognized`, it is exactly that, so no new status is needed. In current state it is still an unknown committee member, but it will be enacted in the future. It might be a confusing name, I am definitely opened to renaming it to something better.
https://github.com/input-output-hk/cardano-ledger/blob/1d167ee0c72c2fdb51774d7998b91f237b9bf0cc/libs/cardano-ledger-api/src/Cardano/Ledger/Api/State/Query/CommitteeMembersState.hs#L44-L46

lehins:
> members that are Unrecognized (so: they have a hot key in the CommitteeState but no corresponding cold key in either the current committee or the committee in the next epoch)

They will be removed at the next epoch, otherwise that Map could grow with garbage CC keys forever

libs/cardano-ledger-api/src/Cardano/Ledger/Api/State/Query.hs
```diff
@@ -165,6 +166,9 @@ queryCommitteeMembersState ::
   Maybe (CommitteeMembersState (EraCrypto era))
 queryCommitteeMembersState coldCredsFilter hotCredsFilter statusFilter nes = do
   (comMembers, comQuorum) <- getCommitteeMembers (queryGovState nes)
+  let nextComMembers =
+        maybe Map.empty fst (getNextEpochCommitteeMembers (queryGovState nes))
+  let allComMembers = comMembers `Map.union` nextComMembers
```
lehins: This is not the right to union with, it should be unioned with `comStateMembers`
```suggestion
```

libs/cardano-ledger-api/src/Cardano/Ledger/Api/State/Query.hs
```diff
@@ -165,6 +166,9 @@ queryCommitteeMembersState ::
   Maybe (CommitteeMembersState (EraCrypto era))
 queryCommitteeMembersState coldCredsFilter hotCredsFilter statusFilter nes = do
   (comMembers, comQuorum) <- getCommitteeMembers (queryGovState nes)
+  let nextComMembers =
+        maybe Map.empty fst (getNextEpochCommitteeMembers (queryGovState nes))
```
lehins, line 171: We don't really care about the Epoch they will expire, since they haven't even gotten enacted yet
```suggestion
  let nextComMembers =
        maybe Set.empty (Map.keysSet . fst) 
        (getNextEpochCommitteeMembers (queryGovState nes))
```

libs/cardano-ledger-api/src/Cardano/Ledger/Api/State/Query.hs
```diff
@@ -176,9 +180,9 @@ queryCommitteeMembersState coldCredsFilter hotCredsFilter statusFilter nes = do
       relevantColdKeys
         | Set.null statusFilter || Set.member Unrecognized statusFilter =
             withFilteredColdCreds $
-              Map.keysSet comMembers
+              Map.keysSet allComMembers
                 `Set.union` Map.keysSet comStateMembers
```
lehins: Next committee members are only relevant, when we care about members that are not in the current committee
```suggestion
              Map.keysSet comMembers
                `Set.union` Map.keysSet comStateMembers
                `Set.union` nextComMembers
```

libs/cardano-ledger-api/src/Cardano/Ledger/Api/State/Query.hs
```diff
@@ -207,9 +211,19 @@ queryCommitteeMembersState coldCredsFilter hotCredsFilter statusFilter nes = do
...
                 Just Nothing -> MemberResigned
                 Just (Just hk) -> MemberAuthorized hk
-        -- TODO: implement after pulsing and snapshots are done
-        let nextEpochChange = NoChangeExpected
-        pure $ CommitteeMemberState hkStatus status mbExpiry nextEpochChange
+        pure $ CommitteeMemberState hkStatus status mbExpiry (nextEpochChange coldCred)
+
+      nextEpochChange :: Credential 'ColdCommitteeRole (EraCrypto era) -> NextEpochChange
+      nextEpochChange ck
+        | inCurrent && not inNext = ToBeRemoved
```
lehins: We need to look at all the relevant keys to report if they will be removed or not
```suggestion
        | Map.member ck relevantColdKeys && not inNext = ToBeRemoved
```

libs/cardano-ledger-api/src/Cardano/Ledger/Api/State/Query.hs
```diff
@@ -207,9 +211,19 @@ queryCommitteeMembersState coldCredsFilter hotCredsFilter statusFilter nes = do
...
+        | not inCurrent && inNext = ToBeEnacted
+        | otherwise = NoChangeExpected
+        where
+          -- This will also include members that are in none of the Committes,
+          -- but only in CommitteeState (Unrecognized members)
```
lehins: I think this comment is very confusing. Below bindings only check membership in current and in future committee, which is clear from the code, extra explanation will only add confusion
```suggestion
```

libs/cardano-ledger-api/src/Cardano/Ledger/Api/State/Query.hs
```diff
@@ -176,9 +180,9 @@ queryCommitteeMembersState coldCredsFilter hotCredsFilter statusFilter nes = do
       relevantColdKeys
         | Set.null statusFilter || Set.member Unrecognized statusFilter =
             withFilteredColdCreds $
-              Map.keysSet comMembers
+              Map.keysSet allComMembers
                 `Set.union` Map.keysSet comStateMembers
-        | otherwise = withFilteredColdCreds $ Map.keysSet comMembers
+        | otherwise = withFilteredColdCreds $ Map.keysSet allComMembers
```
```suggestion
        | otherwise = withFilteredColdCreds $ Map.keysSet comMembers
```

libs/cardano-ledger-api/src/Cardano/Ledger/Api/State/Query.hs
```diff
@@ -195,7 +199,7 @@ queryCommitteeMembersState coldCredsFilter hotCredsFilter statusFilter nes = do
         Credential 'ColdCommitteeRole (EraCrypto era) ->
         Maybe (CommitteeMemberState (EraCrypto era))
       mkMaybeMemberState coldCred = do
-        let mbExpiry = Map.lookup coldCred comMembers
+        let mbExpiry = Map.lookup coldCred allComMembers
```
```suggestion
        let mbExpiry = Map.lookup coldCred comMembers
```

lehins (changes requested):
Not quite right, but it is  really good first attempt. Could you also add a couple of test cases for the new functionality?

lehins:
> At first I thought that expiration of the committee member should be part of the computation of this field, 

That brings me an idea, we could add an extra status: `ToBeExpired`

eras/shelley/impl/cardano-ledger-shelley.cabal
teodanciu: Ok, will update the part before  the patch then.

teodanciu:
>  Could you also add a couple of test cases for the new functionality

There is a test for it: 
https://github.com/input-output-hk/cardano-ledger/pull/3878/commits/c3cdf55e41f8c4c28830836cc9a6f7e12baf8900#diff-fb722545a2358df077b89f77d538341ddfbb9d6a3e68354459abad096b1bfbc7R267


It's still not clear to me what  we want to achieve.  I have assumed this, if we look at an example  ( not saying it's  the right thing ,because I can see quite a few things that are confusing, as i pointed out in the description) : 

```
current:   
c1 -> epoch 
c2 ->  epochExpired
c3 -> epoch
c4 -> epochExpired

committeeState: 
c1 -> h1
c2 -> h2
c80 -> h80
c90 -> h90

next:Committe:
c1 -> epoch
c4 -> epoch
c500 -> epoch
c80 -> epoch

```

Then a query without any filters, would have the domain all the cold keys: so a union of:  current committee, next committee and the keyset of committeeState: 

```
c1 - (Authorized h1, Active, NoChangeExpected)
c2 - (Authorized h2, Expired, ToBeRemoved)
c3 - (NotAuthorized, Active, ToBeRemoved)
c4 - (NotAuthorized, Expired, NoChangeExpected)
c80 - (Authorized h80, Active, ToBeEnacted)
c90 - (Authorized h90, Unrecognized, NoChangeExpected)
c500 - (NotAuthorized, Active, ToBeEnacted)
```

But if I understand from your comment, you would prefer to not have the `nextCommittee` in the domain?  
so basically the query would return results for:
```
c1
c2
c3 
c4 
c80
c90
```
as it does now? In this case, we won't be able to set `ToBeEnacted` status, unless the hot key is in the committeeState (which sound sensible).

teodanciu:
So something like this? 

```
c1 - (Authorized h1, Active, NoChangeExpected)  (it's in current & nextCommittee)
c2 - (Authorized h2, Expired, ToBeRemoved) (it's in current && not in nextCommittee)
c3 -  (NotAuthorized, Active, ToBeRemoved) (it's in current && no the nextCommittee)
c4 - (NotAuthorized, Expired, NoChangeExpected)  (it's in current && the nextCommittee)
c80 - (Authorized h80, ???, ToBeEnacted) (it's in the next committee but not in the current. MemberStatus Active/Expired (depending on the epoch) is a little weird, but what else can it be?  )
c90 - (Authorized h90, Unrecognized,  ToBeRemoved)  (because it's not in nextCommittee)
```

So the difference between this and the implementation in this PR would be:

- if cold keys appear in the nextCommitee, but not in the current committe nor in the committeeState, then they won't be returned in the query
- all results with status Unrecognized will have nextChange to be ToBeRemoved

teodanciu:
As discussed,  the result for this would be: 
```
c1 - (Authorized h1, Active, NoChangeExpected)  (it's in current & nextCommittee)
c2 - (Authorized h2, Expired, ToBeRemoved) (it's in current && not in nextCommittee)
c3 -  (NotAuthorized, Active, ToBeRemoved) (it's in current && no the nextCommittee)
c4 - (NotAuthorized, Expired, NoChangeExpected)  (it's in current && the nextCommittee)
c80 - (Authorized h80, Unrecognized, ToBeEnacted)   (Unrecognized, regardless of the epochNo, because it's not in the current committee)
c90 - (Authorized h90, Unrecognized,  ToBeRemoved)  (Unrecognized, because it's not in currentCommitee,  ToBeRemoved, because it's not in the nextCommittee)
c500 -  (NotAuthorized, Unrecognized, ToBeEnacted)  (Unrecognized, because it's not in currentCommittee, ToBeEnacted, because it's in the nextCommittee)
```

libs/cardano-ledger-api/src/Cardano/Ledger/Api/State/Query.hs
teodanciu: I think in the new implementation, `not inNext` is sufficient  to determine that it has to be removed  (since it means: it's not in current and also not in next)

eras/conway/impl/src/Cardano/Ledger/Conway/Governance.hs
```diff
@@ -657,10 +657,9 @@ instance EraPParams (ConwayEra c) => EraGov (ConwayEra c) where
 
   getConstitution g = Just $ g ^. cgEnactStateL . ensConstitutionL
 
-  getCommitteeMembers g =
-    case g ^. cgEnactStateL . ensCommitteeL of
-      SJust Committee {..} -> Just (committeeMembers, committeeQuorum)
-      SNothing -> Nothing
+  getCommitteeMembers g = ensCommitteeMembers (g ^. cgEnactStateL)
+
+  getNextEpochCommitteeMembers g = ensCommitteeMembers ((getRatifyState g) ^. rsEnactStateL)
```
lehins: Redundant parenthesis
```suggestion
  getNextEpochCommitteeMembers g = ensCommitteeMembers (getRatifyState g ^. rsEnactStateL)
```

libs/cardano-ledger-api/test/Test/Cardano/Ledger/Api/State/QuerySpec.hs
```diff
@@ -51,166 +66,185 @@ spec = do
...
+      let ckHk =
+            Map.foldMapWithKey
+              ( \ck cms ->
+                  case cms of
+                    CommitteeMemberState (MemberAuthorized hk) _ _ _ -> [(ck, Just hk)]
+                    _ -> []
+              )
+              (csCommittee noFilterResult)
+      -- if the member is Authorized, it should appear in the commiteeState
+      Map.filter isJust comStateMembers `shouldBe` Map.fromList ckHk
```
lehins:
```suggestion
      let ckHk =
            Map.mapMapWithKey
              ( \ck cms ->
                  case cms of
                    CommitteeMemberState (MemberAuthorized hk) _ _ _ -> Just (ck, hk)
                    _ -> Nothing
              )
              (csCommittee noFilterResult)
      -- if the member is Authorized, it should appear in the commiteeState
      Map.mapMaybe id comStateMembers `shouldBe` ckHk
```

libs/cardano-ledger-api/test/Test/Cardano/Ledger/Api/State/QuerySpec.hs
```diff
@@ -302,19 +401,18 @@ committeeInfo ::
...
     getCommitteeMembers (nes ^. nesEpochStateL . esLStateL . lsUTxOStateL . utxosGovStateL)
-  let comStateMembers =
-        csCommitteeCreds $
-          nes ^. nesEpochStateL . esLStateL . lsCertStateL . certVStateL . vsCommitteeStateL
-  pure (comMembers, comQurum, comStateMembers)
+  let ledgerState = nes ^. nesEpochStateL . esLStateL
+  let nextCommittee =
+        uncurry (Committee @era) <$> getNextEpochCommitteeMembers (ledgerState ^. lsUTxOStateL . utxosGovStateL)
+  let comState = ledgerState ^. lsCertStateL . certVStateL . vsCommitteeStateL
+  pure (Committee comMembers comQurum, comState, flatColdCreds nextCommittee)
```
lehins:
```suggestion
  let nextCommittee =
        maybe Set.empty (Map.keysSet . fst) $
        getNextEpochCommitteeMembers (ledgerState ^. lsUTxOStateL . utxosGovStateL)
  let comState = ledgerState ^. lsCertStateL . certVStateL . vsCommitteeStateL
  pure (Committee comMembers comQurum, comState, nextCommittee)
```

libs/cardano-ledger-api/test/Test/Cardano/Ledger/Api/State/QuerySpec.hs
```diff
@@ -323,3 +421,7 @@ queryCommitteeMembersStateNoFilters =
     Set.empty
     Set.empty
     Set.empty
+
+flatColdCreds :: Maybe (Committee era) -> Set (Credential 'ColdCommitteeRole (EraCrypto era))
+flatColdCreds maybeCm =
+  Set.fromList $ catMaybes $ mapM Map.keys $ committeeMembers <$> maybeCm
```
lehins: This is way too convoluted :smile: 
No longer needed with my suggestion above: `maybe Set.empty (Map.keysSet . fst)`
```suggestion
```

libs/cardano-ledger-api/test/Test/Cardano/Ledger/Api/State/QuerySpec.hs
```diff
@@ -238,62 +272,127 @@ propFilters ckFilter hkFilter statusFilter nes = do
...
+  Maybe (Committee era) ->
+  Gen (Maybe (Committee era))
+genNextCommittee maybeCm = do
+  let comMembers = foldMap' committeeMembers maybeCm
+  new <- arbitrary @(Map.Map (Credential 'ColdCommitteeRole (EraCrypto era)) EpochNo)
+  q <- arbitrary
+  let half = Map.size new `div` 2
+  oneof $
+    pure
+      <$> [Nothing, Just $ Committee (Map.union (Map.take half comMembers) (Map.take half new)) q]
```
lehins: You are discarding half of newly generated committee. Question is: what is half of a random number? :wink: 

Probably you want to generate a random new committee and remove part of old committee
```suggestion
  retainSize <- choose (0, Map.size comMembers)
  elements 
    [Nothing, Just $ Committee (Map.union new (Map.take retainSize comMembers)) q]
```

libs/cardano-ledger-api/src/Cardano/Ledger/Api/State/Query.hs
```diff
@@ -207,9 +214,20 @@ queryCommitteeMembersState coldCredsFilter hotCredsFilter statusFilter nes = do
...
+      nextEpochChange ck
+        | not inCurrent && inNext = ToBeEnacted
+        | not inNext = ToBeRemoved
+        | expiring = ToBeExpired
+        | otherwise = NoChangeExpected
+        where
+          lookupCurrent = Map.lookup ck comMembers
+          inCurrent = isJust lookupCurrent
+          inNext = Set.member ck nextComMembers
+          expiring = maybe False (< currentEpoch + 1) lookupCurrent
```
lehins: I just realized something about this line. We are reporting that a member will expire by looking into current committee, but it could be possible that the member got his term extended and he will not expire. So we need to be looking into `nextComMembers` for this check.

One more important point is that ToBeExpired is a transition. So if member is already expired then there is no need to mention that he'll expire, it should be `NoChangeExpected`. With regards to the operator that checks expiration. If a member has the expiration set to the current epoch number that means in the next epoch that member will expire.

```suggestion
          expiring = maybe False (== currentEpoch) lookupFuture
```

libs/cardano-ledger-api/src/Cardano/Ledger/Api/State/Query.hs
```diff
@@ -207,9 +214,20 @@ queryCommitteeMembersState coldCredsFilter hotCredsFilter statusFilter nes = do
...
-        -- TODO: implement after pulsing and snapshots are done
-        let nextEpochChange = NoChangeExpected
-        pure $ CommitteeMemberState hkStatus status mbExpiry nextEpochChange
+        pure $ CommitteeMemberState hkStatus status mbExpiry (nextEpochChange coldCred)
+
+      nextEpochChange :: Credential 'ColdCommitteeRole (EraCrypto era) -> NextEpochChange
+      nextEpochChange ck
+        | not inCurrent && inNext = ToBeEnacted
+        | not inNext = ToBeRemoved
+        | expiring = ToBeExpired
```
lehins: You gonna hate me for this, but I think it could be good to add one more prediction `TermAdjusted EpochNo`. Let me know if you are not up this 
```suggestion
        | expiring = ToBeExpired
        | Just curTerm <- lookupCurrent
        , Just newTerm <- lookupFuture
        , curTerm /= newTerm = TermAdjusted newTerm
```

lehins (changes requested):
I've suggested some minor changes.
We can discuss them tomorrow if you have questions about it.

libs/cardano-ledger-api/test/Test/Cardano/Ledger/Api/State/QuerySpec.hs
teodanciu: I was trying to keep the same randomly generated committee size, because I thought the sizes are generated with sense and structure.  But happy to go with your suggestion, since I agree it's less "wtf-inducing".

libs/cardano-ledger-api/src/Cardano/Ledger/Api/State/Query.hs
teodanciu: That's useful information to provide in the query result! 
Are you thinking:   they're in both current and next committee, but with a different expiry epoch?

libs/cardano-ledger-api/test/Test/Cardano/Ledger/Api/State/QuerySpec.hs
lehins: To be honest with you, I am not sure I understand why you set this goal: "I was trying to keep the same randomly generated committee size". Trying to ensure that the size matches the random size does not make much sense to me. 

Moreover, here is an example run of you generator:

```haskell
comMembers = [x,y]
new = [a,b,c,d,e,f,g,h,j]
result = [x,y,a,b,c,d]
```

Why generate `[e,f,g,h,j]` just to discard it.

Also just realized it should be switched to:
```haskell
genNextCommittee maybeCm = 
  oneof [pure Nothing, Just <$> genCom]
  where
    genCom = do
        let comMembers = foldMap' committeeMembers maybeCm
        new <- arbitrary @(Map.Map (Credential 'ColdCommitteeRole (EraCrypto era)) EpochNo)
        ..
```

because there is no point in generating stuff if it will be discarded with Nothing case anyways

libs/cardano-ledger-api/src/Cardano/Ledger/Api/State/Query.hs
lehins: > they're in both current and next committee, but with a different expiry epoch?

Yes. :+1:

libs/cardano-ledger-api/test/Test/Cardano/Ledger/Api/State/QuerySpec.hs
teodanciu: Nice.  Looking at the documentation for this function, I realise that "Map.mapMaybe" could be even better, since I use the key only to create the entry: 

```
      let ckHk =
            Map.mapMaybe
              ( \case
                    CommitteeMemberState (MemberAuthorized hk) _ _ _ -> Just hk
                    _ -> Nothing
              )
              (csCommittee noFilterResult)
```

lehins: Yeah, `mapMaybe` is definitely better. My in-github-pr Haskell coding isn't as good as I would like it to be :smile:

libs/cardano-ledger-api/test/Test/Cardano/Ledger/Api/State/QuerySpec.hs
teodanciu: I have applied this idea here and to the rest of similar functions (generating relevant committeeState and filters): 
https://github.com/input-output-hk/cardano-ledger/pull/3878/commits/e77d5210b0a4819f34646ec341d97e7a32ea7a3a

libs/cardano-ledger-api/src/Cardano/Ledger/Api/State/Query.hs
teodanciu: I think we need to look in both, current and future,  because a member can have ToBeExpired status even when there is no next commitee at all.

There is a case where the two, `ToBeExpired` and `TermAdjusted` are conflicting: 

For an existing committee member, we elect him again, with a Term that is : epoch-of-election + 1.
In this case, after the election, for this member both `TermAdjusted` or `ToBeExpired` kinda apply. 
I have made `ToBeExpired` to have priority - [so I'm only setting `TermAdjusted` in case the term was adjusted AND the new term is not expiring in the next epoch, otherwise it will be `ToBeExpired`](https://github.com/input-output-hk/cardano-ledger/blob/td/committee-query-next-epoch-change/libs/cardano-ledger-api/src/Cardano/Ledger/Api/State/Query.hs#L228)   Seems to be more relevant that hes about to expire, but I can see the argument for `TermAdjusted` too. If we go with `TermAdjusted` in this case, then the member would go from `TermAdjusted` directly to `Expired` + `NoChangeExpected`. 

I have a testcase for this in the other PR: https://github.com/input-output-hk/cardano-ledger/blob/td/committee-query-imp-test/libs/cardano-ledger-api/test/Test/Cardano/Ledger/Api/State/Imp/QuerySpec.hs#L186

teodanciu:
Updated with the suggested changes, and also reflected them in the ImpTests in the other PR: https://github.com/input-output-hk/cardano-ledger/pull/3883

libs/cardano-ledger-api/test/Test/Cardano/Ledger/Api/State/QuerySpec.hs
```diff
@@ -238,62 +269,141 @@ propFilters ckFilter hkFilter statusFilter nes = do
...
+        [ Map.keysSet (filterNext NoChangeExpected noFilterResult)
+        , Map.keysSet (filterNext ToBeExpired noFilterResult)
+        , Map.keysSet (termAdjusted noFilterResult)
+        ]
+        `shouldSatisfy` (== (comMembers `Set.intersection` nextComMembers))
+
+      let currentEpoch = csEpochNo noFilterResult
+      cmsExpiration
+        <$> filterNext ToBeExpired noFilterResult
+          `shouldSatisfy` all (all (== currentEpoch))
```
lehins: There is a test failure on this line, which makes sense, since according to implementation we also need to look into `nextComMembers`.

```
Failures:

  test/Test/Cardano/Ledger/Api/State/QuerySpec.hs:300:11: 
  1) cardano-ledger-api.State.GetCommitteeMembersState CommitteeMembersState Query
       Falsifiable (after 7 tests and 1 shrink):
         fromList []
         Just (Committee {committeeMembers = fromList [(ScriptHashObj (ScriptHash "093103280a4ec6ee07112689e6ffe2af40a7e02f2f70bc4889851183"),EpochNo 8),(ScriptHashObj (ScriptHash "588a6d86d1699c0bd1edec2c92dc13a368255[831](https://github.com/input-output-hk/cardano-ledger/actions/runs/6975772642/job/18983342430?pr=3878#step:15:832)8c48acad2dc56917"),EpochNo 14),(ScriptHashObj (ScriptHash "9c3189b43d0cf6640fd666aa8fd22432d1897d737760b18d86ef57a7"),EpochNo 13),(KeyHashObj (KeyHash "cd1d95b25666eb69a049de36d8b61ff6ba073c6e2ba4109fa9c23bfe"),EpochNo 3)], committeeQuorum = 1 % 1})
         Just (Committee {committeeMembers = fromList [(ScriptHashObj (ScriptHash "093103280a4ec6ee07112689e6ffe2af40a7e02f2f70bc4889851183"),EpochNo 0),(ScriptHashObj (ScriptHash "35f78fe49efa041b892575913026746dc10c1e75ed3595aafe812653"),EpochNo 0),(ScriptHashObj (ScriptHash "588a6d86d1699c0bd1edec2c92dc13a3682558318c48acad2dc56917"),EpochNo 13)], committeeQuorum = 64862647 % 100000000})
         CommitteeState {csCommitteeCreds = fromList [(ScriptHashObj (ScriptHash "093103280a4ec6ee07112689e6ffe2af40a7e02f2f70bc4889851183"),Just (ScriptHashObj (ScriptHash "44b7363c58c9fe78b44a34782ae98cb1639cb95c2c98706ae2786875")))]}
         fromList [ScriptHashObj (ScriptHash "093103280a4ec6ee07112689e6ffe2af40a7e02f2f70bc4889851183"),ScriptHashObj (ScriptHash "20c5d88d990db9dfc6d23af8b27b232d56251edfa7d933b3051c9861"),ScriptHashObj (ScriptHash "588a6d86d1699c0bd1edec2c92dc13a3682558318c48acad2dc56917"),ScriptHashObj (ScriptHash "9c3189b43d0cf6640fd666aa8fd22432d1897d737760b18d86ef57a7"),ScriptHashObj (ScriptHash "b5d7ef53fe92a1eaef21d591bc1c03471070cf55eb4e108b384b59c8"),ScriptHashObj (ScriptHash "e813a73d6b45a99ff5ec56351190ce1b91e7a869e0a3c3e2cd3bcbcf"),KeyHashObj (KeyHash "4575131[835](https://github.com/input-output-hk/cardano-ledger/actions/runs/6975772642/job/18983342430?pr=3878#step:15:836)c8da45e54cbecf03c6fbd3fd6758b491f1c33823f747ff"),KeyHashObj (KeyHash "72dc08f4050aea7e1d26d66b18af06fc7405f639785be8ac16b04151"),KeyHashObj (KeyHash "baa3e30edd36077ae5023610888004789080f28f5bb645e84a739c19"),KeyHashObj (KeyHash "cd1d95b25666eb69a049de36d8b61ff6ba073c6e2ba4109fa9c23bfe")]
         fromList [ScriptHashObj (ScriptHash "138ab805795f19877b0feb86b282c476476d3060155b36d9d60ad499"),ScriptHashObj (ScriptHash "2a9c59edbed880d59976b97b78a93f9800542342b9adff5951e3c738"),ScriptHashObj (ScriptHash "d83d9677f130285887c0c94c503c72217e14938d917937469e21f9b3"),ScriptHashObj (ScriptHash "f2aa8f41c29b98354bc356ee38c57d76e2b8e5f77427a41f7ed51325"),KeyHashObj (KeyHash "278a398ba6e07a7f58db71302031ba69508d3a919331d447b9aa7a64")]
       predicate failed on: fromList [(ScriptHashObj (ScriptHash "093103280a4ec6ee07112689e6ffe2af40a7e02f2f70bc4889851183"),Just (EpochNo 8))]

  To rerun use: --match "/cardano-ledger-api/State/GetCommitteeMembersState/CommitteeMembersState Query/" --seed 581728073
```

lehins (changes requested):
There is a test failure on CI due to a faulty tes, otherwise it looks great.

libs/cardano-ledger-api/test/Test/Cardano/Ledger/Api/State/QuerySpec.hs
teodanciu: Mhm, must have lost the change during a rebase, I can remember changing this.  Will retrieve it

libs/cardano-ledger-api/src/Cardano/Ledger/Api/State/Query.hs
lehins: Your reasoning is 100% solid. I did think about this case as well after I made this suggestion. I was also leaning towards "ToBeExpired" being more relevant than "TermAdjusted". I was curious to see which was gonna be your preference :slightly_smiling_face:

lehins (approved):
Awesome work!

PR #3883, opened 2023-11-21
Committee QuerySpec Imp Test

teodanciu:
# Description

Closes https://github.com/input-output-hk/cardano-ledger/issues/3881

Depends on: https://github.com/input-output-hk/cardano-ledger/pull/3878

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/libs/cardano-ledger-api/test/Test/Cardano/Ledger/Api/State/Imp/QuerySpec.hs
+++ b/libs/cardano-ledger-api/test/Test/Cardano/Ledger/Api/State/Imp/QuerySpec.hs
@@ -76,8 +76,8 @@ spec =
           Map.fromList
             [ (c1, EpochNo 12)
             , (c2, EpochNo 2)
-            , (c3, EpochNo 11)
-            , (c4, EpochNo 6)
+            , (c3, EpochNo 7)
+            , (c4, EpochNo 5)
             ]
     ga1 <-
       electCommittee
@@ -88,17 +88,18 @@ spec =
 
     expectMembers Set.empty
     expectNoFilterQueryResult Nothing
-    passEpoch >> passEpoch
-    -- epoch 2
+    passEpoch >> passEpoch -- epoch 2
     expectMembers $ Map.keysSet newMembers
+    -- members for which the expiration epoch is the current epoch are `ToBeExpired`
     expectNoFilterQueryResult $
       Just $
         Map.fromList
           [ (c1, CommitteeMemberState MemberNotAuthorized Active (Just 12) NoChangeExpected)
           , (c2, CommitteeMemberState MemberNotAuthorized Active (Just 2) ToBeExpired)
-          , (c3, CommitteeMemberState MemberNotAuthorized Active (Just 11) NoChangeExpected)
-          , (c4, CommitteeMemberState MemberNotAuthorized Active (Just 6) NoChangeExpected)
+          , (c3, CommitteeMemberState MemberNotAuthorized Active (Just 7) NoChangeExpected)
+          , (c4, CommitteeMemberState MemberNotAuthorized Active (Just 5) NoChangeExpected)
           ]
+    -- hot cred status of members with registered hot keys becomes `MemberAuthorized`
     hk1 <- registerCommitteeHotKey c1
     hk2 <- registerCommitteeHotKey c2
     expectNoFilterQueryResult $
@@ -106,8 +107,8 @@ spec =
         Map.fromList
           [ (c1, CommitteeMemberState (MemberAuthorized (KeyHashObj hk1)) Active (Just 12) NoChangeExpected)
           , (c2, CommitteeMemberState (MemberAuthorized (KeyHashObj hk2)) Active (Just 2) ToBeExpired)
-          , (c3, CommitteeMemberState MemberNotAuthorized Active (Just 11) NoChangeExpected)
-          , (c4, CommitteeMemberState MemberNotAuthorized Active (Just 6) NoChangeExpected)
+          , (c3, CommitteeMemberState MemberNotAuthorized Active (Just 7) NoChangeExpected)
+          , (c4, CommitteeMemberState MemberNotAuthorized Active (Just 5) NoChangeExpected)
           ]
     expectQueryResult
       (Set.fromList [c2, c3, c5])
@@ -115,7 +116,7 @@ spec =
       (Set.fromList [Active, Unrecognized])
       ( Just $
           Map.fromList
-            [ (c3, CommitteeMemberState MemberNotAuthorized Active (Just 11) NoChangeExpected)
+            [ (c3, CommitteeMemberState MemberNotAuthorized Active (Just 7) NoChangeExpected)
             , (c2, CommitteeMemberState (MemberAuthorized (KeyHashObj hk2)) Active (Just 2) ToBeExpired)
             ]
       )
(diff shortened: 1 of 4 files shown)


libs/cardano-ledger-api/test/Test/Cardano/Ledger/Api/State/Imp/QuerySpec.hs
```diff
@@ -0,0 +1,249 @@
...
+  , GovState era ~ ConwayGovState era
+  ) =>
+  SpecWith (ImpTestState era)
+spec =
+  it "Committee queries" $ do
+    setPParams
+    drep <- setupSingleDRep
+    vals <- replicateM 8 freshKeyHash
+    let (c1, c2, c3, c4, c5, c6, c7, c8) = case vals of
+          [c1', c2', c3', c4', c5', c6', c7', c8'] -> (c1', c2', c3', c4', c5', c6', c7', c8')
```
teodanciu: This is so unpleasant - is there a better way?

lehins: Sure.
```haskell
c1 <- freshKeyHash
c2 <- freshKeyHash
c3 <- freshKeyHash
c4 <- freshKeyHash
c5 <- freshKeyHash
c6 <- freshKeyHash
c7 <- freshKeyHash
c8 <- freshKeyHash
```

teodanciu: Haha! Fair enough

libs/cardano-ledger-api/test/Test/Cardano/Ledger/Api/State/Imp/QuerySpec.hs
```diff
@@ -0,0 +1,252 @@
+{-# LANGUAGE DataKinds #-}
+{-# LANGUAGE OverloadedLists #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeFamilies #-}
+{-# LANGUAGE TypeOperators #-}
+
+module Test.Cardano.Ledger.Api.State.Imp.QuerySpec where
```
lehins, line 7: Make sure to always use explicit exports in public modules

lehins: Nevermind it is not a public module. Nevertheless, explicit exports is very valueable, not only for the reader, but also for compiler optimizations

libs/cardano-ledger-api/test/Tests.hs
```diff
@@ -15,6 +21,8 @@ apiSpec =
       TxBody.spec
     describe "State" $ do
       StateQuery.spec
+    describe "Imp" $ withImpState @Conway $ do
+      ImpQuery.spec @Conway
```
lehins, line 25: I don't think it is necessary, since `withImpState` will resolve the `era`
```suggestion
      ImpQuery.spec
```

Soupstraw (approved):

libs/cardano-ledger-api/test/Test/Cardano/Ledger/Api/State/Imp/QuerySpec.hs
```diff
@@ -0,0 +1,314 @@
...
+            [ (c1, 12)
+            , (c2, 2)
+            , (c3, 7)
+            , (c4, 5)
+            ]
+        )
+    expectMembers Set.empty
+    expectNoFilterQueryResult Nothing
+    passEpoch >> passEpoch -- epoch 2
+    expectMembers $ Set.fromList [c1, c2, c3, c4]
```
lehins: It is better to create common binding for expectations, since it is easier to adjust the test later if necessary. Sort of like `x` is some value and I expect the outcome to be equal to `x` after some operation, rather then `x = 1` and I expect `1` after that operation.

Naturally it doesn't always make sense, but when it does it makes things easier
```suggestion
    let newMembers = Map.fromList [(c1, 12), (c2, 2), (c3, 7), (c4, 5)]
    ga1 <- electCommittee SNothing drep Set.empty newMembers
    expectMembers Set.empty
    expectNoFilterQueryResult Nothing
    passEpoch >> passEpoch -- epoch 2
    expectMembers $ Map.keysSet newMembers
```

libs/cardano-ledger-api/test/Test/Cardano/Ledger/Api/State/Imp/QuerySpec.hs
```diff
@@ -0,0 +1,314 @@
...
+          ]
+  where
+    expectMembers :: Set.Set (KeyHash 'ColdCommitteeRole (EraCrypto era)) -> ImpTestM era ()
+    expectMembers expKhs = do
+      committee <-
+        getsNES $
+          nesEsL . esLStateL . lsUTxOStateL . utxosGovStateL . cgEnactStateL . ensCommitteeL
+      let members = Map.keysSet $ foldMap' committeeMembers committee
+      impAnn "Expecting committee members" $ members `shouldBe` Set.map KeyHashObj expKhs
+    expectQueryResult ckFilter hkFilter statusFilter expResult = do
```
lehins, line 263: Could you add a type signature for this function with `HasCallStack` constraint, this will let the user know which usage of the function caused a test to fail.

libs/cardano-ledger-api/test/Test/Cardano/Ledger/Api/State/Imp/QuerySpec.hs
```diff
@@ -0,0 +1,314 @@
...
+    expectNoFilterQueryResult $
+      Just $
+        Map.fromList
+          [ (c3, CommitteeMemberState MemberResigned Active (Just 9) NoChangeExpected)
+          , (c4, CommitteeMemberState MemberNotAuthorized Active (Just 9) NoChangeExpected)
+          , (c6, CommitteeMemberState (MemberAuthorized (KeyHashObj hk6)) Active (Just 9) NoChangeExpected)
+          , (c7, CommitteeMemberState MemberNotAuthorized Expired (Just 7) NoChangeExpected)
+          ]
+  where
+    expectMembers :: Set.Set (KeyHash 'ColdCommitteeRole (EraCrypto era)) -> ImpTestM era ()
```
lehins: Same as below for `expectQueryResult`
```suggestion
    expectMembers :: 
      HasCallStack => Set.Set (KeyHash 'ColdCommitteeRole (EraCrypto era)) -> ImpTestM era ()
```

libs/cardano-ledger-api/test/Test/Cardano/Ledger/Api/State/Imp/QuerySpec.hs
```diff
@@ -0,0 +1,314 @@
...
+    expectNoFilterQueryResult $
+      Just $
+        Map.fromList
+          [ (c1, CommitteeMemberState (MemberAuthorized (KeyHashObj hk1)) Active (Just 12) NoChangeExpected)
+          , (c2, CommitteeMemberState (MemberAuthorized (KeyHashObj hk2)) Active (Just 2) ToBeExpired)
+          , (c3, CommitteeMemberState MemberNotAuthorized Active (Just 7) NoChangeExpected)
+          , (c4, CommitteeMemberState MemberNotAuthorized Active (Just 5) NoChangeExpected)
+          ]
+    expectQueryResult
+      (Set.fromList [c2, c3, c5])
```
lehins: Have you considered using `OverloadedLists` language extension in this module. It could reduce the noise quite a bit. This is just a suggestion.

libs/cardano-ledger-api/test/Test/Cardano/Ledger/Api/State/Imp/QuerySpec.hs
```diff
@@ -0,0 +1,314 @@
...
+        drep
+        (Set.fromList [c1])
+        ( Map.fromList
+            [ (c3, 9)
+            , (c4, 9)
+            , (c6, 9)
+            ]
+        )
+    passEpoch -- epoch 7
+    -- members whose term changed have next expoch change TermAdjusted
```
lehins:
```suggestion
    -- members whose term changed have next epoch change `TermAdjusted`
```

libs/cardano-ledger-api/test/Test/Cardano/Ledger/Api/State/Imp/QuerySpec.hs
```diff
@@ -0,0 +1,314 @@
...
+      nes <- use impNESL
+      let res =
+            queryCommitteeMembersState
+              (Set.map KeyHashObj ckFilter)
+              (Set.map KeyHashObj hkFilter)
+              statusFilter
+              nes
+      impAnn "Expecting query result" $
+        csCommittee <$> res `shouldBe` Map.mapKeys KeyHashObj <$> expResult
+    expectNoFilterQueryResult expResult =
```
lehins: Ditto type signature and `HasCallStack`

lehins (commented):
Beautiful!!!! I love how clear and readable those test cases are!
:heart:

lehins (approved):

libs/cardano-ledger-api/test/Test/Cardano/Ledger/Api/State/Imp/QuerySpec.hs
teodanciu: Yes  i agree, it makes the intention of what is tested clearer too. Thanks

libs/cardano-ledger-api/test/Test/Cardano/Ledger/Api/State/Imp/QuerySpec.hs
teodanciu: The extension was already declared in the file!  All i had to do is use it, thank you for making me aware of it. 
It removed quite a bit of noise (and tons of parentheses too  :skull: )

PR #3892, opened 2023-11-24
Translate DRep expiry test from `ConwayFeatures` to Imp `GovSpec`

### LINKED ISSUE #3805: Translate ConwayFeatures to ImpTest
ConwayFeatures (like all other [Era]Features) tests are incredibly hard to read and maintain. 
They should be translated to use ImpTest and moved into `cardano-ledger-conway:testlib`

teodanciu:
# Description

Closes https://github.com/input-output-hk/cardano-ledger/issues/3805

# Checklist

- [ ] Self-reviewed the diff


eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/GovSpec.hs
```diff
@@ -217,3 +220,132 @@ spec =
...
                   )
           submitGovAction_ constitutionAction1
+
+    describe "DRep expiry" $ do
+      it "is correlated to number of dormant epochs" $ do
+        modifyPParams $ ppGovActionLifetimeL .~ 2
+        drep <- setupSingleDRep
+
+        -- epoch 0
+        _ <- submitConstitutionGovAction SNothing
```
teodanciu: These are these checks: https://github.com/input-output-hk/cardano-ledger/blob/td/replace-conwayfeatures-with-imp/libs/cardano-ledger-test/src/Test/Cardano/Ledger/Examples/ConwayFeatures.hs#L463

Instead of comparing the number of dormant epochs of current and previous epochStates, it its checking for each epoch that it's the expected number.
Also,  instead of comparing the whole DRepState, we just check that the expiry has been updated as expected, depending on the drepActivity pparam and dormant epochs

lehins (approved):
:tada:

PR #3895, opened 2023-11-29
Constrained v2

### LINKED ISSUE #3821: Constraint generation needs to support sum types
Currently the constraint based generation system doesn't support sum types nicely. We are working on cleaning this up.

### LINKED ISSUE #3822: Constraint generation doesn't support universal quantification over sets
We should support `forAll` over things like sets natively in a way that scales cleanly.

### LINKED ISSUE #3823: Constraint generation needs to be easier to extend
Currently the way we rewrite constraints to "specs" doesn't scale when we add new constructs. On constrained-v2 we are working on a way of decoupling the structure of terms from computing specs that solves this problem.

MaximilianAlgehed:
# Description
This PR introduces version 2 of the `Constrained` approach to generating random data.

It's a significant overhaul of the existing approach and it's far from done. However, it is at the point where
it can be useful for writing properties (e.g. for the `GOV` rule) so getting at least a WIP PR up seems very valuable.

Furthermore, if we can get this merged we might be able to start working on some of the STS property issues in parallel with improving this.

@lehins any guidance on setting up the separate package would be much appreciated - is there a README somewhere I should follow?

# Checklist

- [ ] Self-reviewed the diff


libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/V2/Conway.hs
```diff
@@ -0,0 +1,840 @@
...
+   , Set (KeyHash 'Witness (EraCrypto (ConwayEra c)))
+   , MultiAsset (EraCrypto (ConwayEra c))
+   , StrictMaybe (ScriptIntegrityHash (EraCrypto (ConwayEra c)))
+   , StrictMaybe (AuxiliaryDataHash (EraCrypto (ConwayEra c)))
+   , StrictMaybe Network
+   , VotingProcedures (ConwayEra c)
+   , SOS.OSet (ProposalProcedure (ConwayEra c))
+   , StrictMaybe Coin
+   , Coin
+   ]
```
TimSheard: I am thinking this is some kind of type-level description of TxBody as a record with all these fields. A comment would be nice here. Is it the same for 'ConwayUnivFns'

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/V2/Conway.hs
```diff
@@ -0,0 +1,840 @@
...
+instance (IsConwayUniv univ, Crypto c, Typeable index) => HasSpec univ (SafeHash c index) where
+  type TypeSpec univ (SafeHash c index) = ()
+  emptySpec = ()
+  combineSpec _ _ = TrueSpec
+  genFromTypeSpec _ = pureGen arbitrary
+  conformsTo _ _ = True
+  toPreds _ _ = toPred True
+
+instance HasSimpleRep (TxId c)
+instance (IsConwayUniv univ, Crypto c) => HasSpec univ (TxId c)
```
TimSheard: It seems there is some sort of relationship between the classes HasSimpleRep and HasSpec. Should we comment about this?

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/V2/Conway.hs
```diff
@@ -0,0 +1,840 @@
...
+  getLengthSpec (StringSpec len) = len
+  getLength = BS.length
+
+instance StringLike ShortByteString where
+  lengthSpec = StringSpec
+  getLengthSpec (StringSpec len) = len
+  getLength = SBS.length
+
+data StringFn (univ :: [Type] -> Type -> Type) as b where
+  LengthFn :: StringLike s => StringFn univ '[s] Int
```
TimSheard: I have not seen this type level  'as b' before.  Strange that 'b' is never mentioned elsewhere.

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/V2/Conway/Ledger.hs
```diff
@@ -0,0 +1,51 @@
...
+import Cardano.Ledger.Conway.Governance
+import Data.Map qualified as Map
+import Cardano.Ledger.Conway.Core
+import Cardano.Ledger.Api
+import Cardano.Ledger.Shelley.API.Types
+import Cardano.Ledger.BaseTypes
+import Lens.Micro
+import Data.Set qualified as Set
+import Cardano.Ledger.Conway.PParams
+
```
TimSheard: This just combines and re-exports things from elsewhere? A comment maybe

libs/constrained/src/Constrained/Base.hs
```diff
@@ -0,0 +1,1944 @@
...
+      , Show (TypeSpec univ a)
+      , IsUniverse univ
+      ) => HasSpec univ a where
+  type TypeSpec (univ :: [Type] -> Type -> Type) a
+  type TypeSpec univ a = TypeSpec univ (SimpleRep a)
+  emptySpec       :: TypeSpec univ a
+  combineSpec     :: TypeSpec univ a -> TypeSpec univ a -> Spec univ a
+  genFromTypeSpec :: (HasCallStack, MonadGenError m) => TypeSpec univ a -> GenT m a
+  conformsTo      :: HasCallStack => a -> TypeSpec univ a -> Bool
+  toPreds         :: Term univ a -> TypeSpec univ a -> Pred univ
```
TimSheard: In some sense this is the "meat" of version 2. We should explain more about this.

libs/constrained/src/Constrained/Base.hs
```diff
@@ -0,0 +1,1944 @@
...
+
+  default genFromTypeSpec :: ( HasSimpleRep a
+                             , HasSpec univ (SimpleRep a)
+                             , TypeSpec univ a ~ TypeSpec univ (SimpleRep a)
+                             )
+                          => (HasCallStack, MonadGenError m) => TypeSpec univ a -> GenT m a
+  genFromTypeSpec s = fromSimpleRep <$> genFromTypeSpec @univ s
+
+  default conformsTo :: ( HasSimpleRep a
+                        , HasSpec univ (SimpleRep a)
```
TimSheard: All these default implementations  are the "magic" that make these work. I'd like to know more about these.

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/V2/Conway.hs
TimSheard: After a bit of thought. 'as' is just the second type parameter, right?

libs/constrained/src/Constrained/Core.hs
```diff
@@ -0,0 +1,102 @@
...
+
+import Control.Monad
+import Data.Set (Set)
+import Data.Set qualified as Set
+import Data.Typeable
+import Constrained.List
+
+-- Variables --------------------------------------------------------------
+
+data Var a = Var { nameOf :: Int }
```
lehins:
```suggestion
newtype Var a = Var { nameOf :: Int }
```

libs/constrained/src/Constrained/Core.hs
```diff
@@ -0,0 +1,102 @@
...
+instance (forall a. Rename (f a)) => Rename (List f as) where
+  rename v v' = mapList (rename v v')
+
+-- TODO: make this nice
+freshVar :: Var a -> Set Int -> Var a
+freshVar (Var n) names =
+  let ns = [ v | v <- Set.toList names ]
+  in if n `elem` ns
+     then Var . (1 +) $ maximum ns
+     else Var n
```
lehins: What's the point of converting a `Set` into a list here?
```suggestion
  if n `Set.member` names
     then Var . (1 +) $ maximum ns
     else Var n
```

libs/constrained/src/Constrained/GenT.hs
```diff
@@ -0,0 +1,171 @@
...
+import Control.Monad
+import Test.QuickCheck hiding (Fun, Args)
+import Test.QuickCheck.Gen
+import Test.QuickCheck.Random
+
+data GenMode = Loose
+             | Strict
+             deriving (Ord, Eq, Show)
+
+newtype GenT m a = GenT { runGenT :: GenMode -> Gen (m a) }
```
lehins: Why don't we use an existing implementation `GenT`: https://hackage.haskell.org/package/quickcheck-transformer-0.3.1.2/docs/Test-QuickCheck-GenT.html#t:GenT

This could then be: `GenT (ReaderT GenMode m) a`. It would let you avoid redefining bunch of functions and instances in this module.


We could even define a stack along the lines of:
```
GenT (ReaderT GenMode (WriterT [String] (ExceptT [String] m))) a
```

`WriterT` would be used for failures, `EsceptT` for fatal failures

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/V2/Conway/Ledger.hs
```diff
@@ -0,0 +1,51 @@
...
+{-# LANGUAGE LambdaCase #-}
+{-# LANGUAGE StandaloneDeriving #-}
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE UndecidableInstances #-}
+{-# LANGUAGE FunctionalDependencies #-}
+{-# LANGUAGE RankNTypes #-}
+{-# LANGUAGE QuantifiedConstraints #-}
+{-# LANGUAGE DeriveGeneric #-}
+{-# LANGUAGE ImpredicativeTypes #-}
+{-# LANGUAGE DerivingVia #-}
```
lehins: Do you just copy paste your favorite extensions into each module regardless if they are used or not? :rofl:

lehins (commented):
There is a lot of stuff in here. I'm not going to have time to do a proper review until next week.

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/V2/Conway/Ledger.hs
MaximilianAlgehed: Because it takes time to figure out which are used and not. The right solution is to put them in the cabal file.

lehins: > The right solution is to put them in the cabal file.

That's the worst solution, because it would enable all of them in ghci for all the transitive packages in the project that depend on this package

> Because it takes time to figure out which are used and not. 

Being lazy is not an excuse :wink: 

In this module for example I can clearly see which ones are not used: none of them are used :smile:

MaximilianAlgehed: That's a silly argument when talking about a draft PR... It makes a lot of sense to cut corners and copy paste the preamble when working to get things done.

By all means, you may call it lazy. But then, aren't comments about the preamble in a draft PR an example of a lazy review?

lehins: No worries, I wasn't telling you to change anything. I understand that it is a draft. I just made a comment, cause I found it funny to see an empty module with over 30 extensions in it.

That being said, for a final version of a PR I would appreciate if only extensions that are being used in a module are enabled.

lehins:
> @lehins any guidance on setting up the separate package would be much appreciated 

I think you got it.

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/V2/Conway/Ledger.hs
MaximilianAlgehed: You're right of course. I'm just a bit touchy today. Bad day that's all.

Though I would prefer to put things in the cabal file, if that makes sense? (At least the uncontroversial extensions)

MaximilianAlgehed:
> > @lehins any guidance on setting up the separate package would be much appreciated
> 
> I think you got it.

No nix crap in other words 😅

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/V2/Conway/Ledger.hs
MaximilianAlgehed: > That's the worst solution, because it would enable all of them in ghci for all the transitive packages in the project that depend on this package

That's not the case, @UlfNorell just tried it sitting next to me.

cabal.project
```diff
@@ -65,6 +65,7 @@ packages:
   libs/cardano-ledger-test
   libs/plutus-preprocessor
   libs/ledger-state
+  libs/constrained
```
lehins: We need to rename the library, since there is already a package with that name on Hackage: https://hackage.haskell.org/package/constrained

Maybe a suffix indicating that it integrates with QuickCheck?

* eg. `QuickCheck-constrained`
* or `constrained-generator`?

MaximilianAlgehed: I like `constrained-generators`!

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/V2/Conway/Ledger.hs
lehins: Here is the GHC ticket for reference: https://gitlab.haskell.org/ghc/ghc/-/issues/10827

`stack` will even produce a nice warnigng about this nowadays:

```
> stack repl

Warning: There are Cabal flags for this project which may prevent GHCi from loading your code properly. In some cases it can also load some projects which would otherwise fail to build.
         
         -XOverloadedLists will be used, but it can cause type ambiguity in code not usually compiled with it. It is specified for: foo:lib. But not for: bar:lib.
         
         To resolve, remove the flag(s) from the Cabal file(s) and instead put them at the top of the Haskell files.
```

When I ran into this I spent half a day scratching my head trying to figure this thing out.

I think it might not be a problem for `cabal` since it refuses to load multiple libraries at once, probably due to the same ghci issue:

```
> cabal repl foo bar
Error: cabal: Cannot open a repl for multiple components at once. The targets
'bar' and 'foo' refer to different components..

The reason for this limitation is that current versions of ghci do not support
loading multiple components as source. Load just one component and when you
make changes to a dependent component then quit and reload.
```

In any case, there are other downsides to enabling extensions globally for the whole package especially for a multipackage repository like that. It is not possible to keep track in your head which packages have which extensions enabled globally.

So, please, do not turn on extensions in the cabal file.

MaximilianAlgehed:
@lehins the formatter is broken again...

lehins:
@MaximilianAlgehed, `fourmolu` is not yet on par with your advanced Haskell techniques? :smile: 

Why do you say it is broken. Looks like `fourmolu` on CI is passing.

MaximilianAlgehed:
It's passing now because I changed the code to work around the bug the tool itself told me that it has a few commits ago:
```
Loaded config from: /home/runner/work/cardano-ledger/cardano-ledger/fourmolu.yaml
libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/V2/Conway/Gov.hs
@@ -179,7 +179,8 @@
           cppGovActionLifetime -- :: !(THKD 'GovGroup f EpochInterval)
           cppGovActionDeposit -- :: !(THKD 'GovGroup f Coin)
           cppDRepDeposit -- :: !(THKD 'GovGroup f Coin)
-          _cppDRepActivity -> -- :: !(THKD 'GovGroup f EpochInterval)
+          _cppDRepActivity ->
+             -- :: !(THKD 'GovGroup f EpochInterval)
              [ cppMaxBBSize /=. Lit (THKD $ SJust 0)
              , cppMaxTxSize /=. Lit (THKD $ SJust 0)
              , cppMaxBHSize /=. Lit (THKD $ SJust 0)

  Formatting is not idempotent.
  Please, consider reporting the bug.
libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/V2/Conway/PParams.hs
@@ -4[5](https://github.com/input-output-hk/cardano-ledger/actions/runs/7186214519/job/19571131809#step:4:6),[7](https://github.com/input-output-hk/cardano-ledger/actions/runs/7186214519/job/19571131809#step:4:8) +45,[8](https://github.com/input-output-hk/cardano-ledger/actions/runs/7186214519/job/19571131809#step:4:9) @@
           cppGovActionLifetime -- :: !(THKD 'GovGroup f EpochInterval)
           cppGovActionDeposit -- :: !(THKD 'GovGroup f Coin)
           cppDRepDeposit -- :: !(THKD 'GovGroup f Coin)
-          _cppDRepActivity -> -- :: !(THKD 'GovGroup f EpochInterval)
+          _cppDRepActivity ->
+             -- :: !(THKD 'GovGroup f EpochInterval)
              [ cppMaxBBSize /=. Lit (THKD 0)
              , cppMaxTxSize /=. Lit (THKD 0)
              , cppMaxBHSize /=. Lit (THKD 0)

  Formatting is not idempotent.
  Please, consider reporting the bug.
```

lehins:
Looks like it chokes on some comments. I am not too worried about that, since comments can always be adjusted in a manner that makes fourmolu happy.

An occasional hiccup like that is a very small price to pay for consistently formatted code that is actually readable.

MaximilianAlgehed:
@lehins in this case the comments needed to be where they were to make sense (they were pointing out the type of each argument in a lambda without too much noise. But anyway, the comments are not a huge deal.

Let's agree to disagree on the "actually readable" point.

MaximilianAlgehed:
My main complaints with it are:
1. It doesn't group arguments nicely, ruining code like:
    ```
    caseOn b
      (branch $ \ a -> ...)
      (branch $ \ b -> ...)
   ```
   by turning it into things like:
   ```
    caseOn
      b
      (branch $ \ a -> ...)
      (branch $ \ b -> ...)
   ```
   which is unnecessarily space-consuming and weird-looking.
2. It doesn't align cases, meaning it turns stuff like:
    ```
    foo = case bar of
        Bar a b c d -> True
        Baz a       -> False
        Doodle c    -> True
    ```
    into stuff like
    ```
    foo = case bar of
        Bar a b c d -> True
        Baz a -> False
        Doodle c -> True
    ```
    which can get really cramped in big functions with lots of cases
3. It doesn't align type signatures and has no option to put `->` at the start of the line.

Anyway, I can deal with it but I would never turn it on myself.

lehins:
> Anyway, I can deal with it but I would never turn it on myself.

You would, if you were part of a big team of developers each with their own style of writing Haskell.

> My main complaints with it are:

Every single one of those points slightly affects the aesthetics of the code, which I can certainly agree with. However, fourmolu is an enormous time saver:
* We never have to manually adjust the code to "make it look pretty" or conform to some specific style
* We never have to argue with team mates on what is more appropriate style of coding.

Let me tell you, I will rather deal with slightly quirky looking code than the alternatives of:
* memorizing and following a 1200 line styling guide like other teams do (eg. https://github.com/input-output-hk/ouroboros-consensus/blob/main/docs/website/contents/for-developers/StyleGuide.md)
* or have no guideline at all and have plethora of code style mixes like in [`cardano-base`](https://github.com/input-output-hk/cardano-base/blob/e545ee648cb7ef9d2715286dbc38987ffc3b5e4d/cardano-crypto-class/src/Cardano/Crypto/DSIGN/Class.hs#L433-L444). It is pretty common to even find modules that mix 2 and 4 space indentation in there.

Oh yeah, and one more point, fourmolu style is very git diff friendly! Which is another huge time saver. 

> It doesn't align cases, meaning it turns stuff like:

Example with case alignment you pointed out is very git unfriendly.

> It doesn't align type signatures and has no option to put -> at the start of the line.

It actually does, it is one of the configuration options which would apply throughout the codebase, which we choose not to enable.

MaximilianAlgehed:
> You would, if you were part of a big team of developers each with their own style of writing Haskell.

There was that one thing about assumptions and donkeys...

I absolutely understand the point of using formatters, but the main issue I find is that they make code difficult to read. Pf course, opinions can differ on what makes code readable, but my opinion is that fourmolu is particularly quirky...

But yes, the co-ordination problem is unfortunately particularily difficult to solve in Haskell - I dont know of any reasonable tools...

> Every single one of those points slightly affects the aesthetics of the code, which I can certainly agree with. 

The problem with all these tools is that the inevitable quirks compound to make the code difficult to read. Now, you can argue on weather or not its code or diffs that should be read - my philosophy is that readable code is much more important than readable diffs. But reasonable people can differ on this point.

Soupstraw:
>     1. It doesn't group arguments nicely, ruining code like:
>        ```
>        caseOn b
>          (branch $ \ a -> ...)
>          (branch $ \ b -> ...)
>        ```

I think it should sort of work if you put parentheses around `caseOn b`.

libs/constrained-generators/src/Constrained/Base.hs
```diff
@@ -0,0 +1,2680 @@
...
+    Spec univ a
+  SuspendedSpec ::
+    HasSpec univ a =>
+    Var a ->
+    Pred univ ->
+    Spec univ a
+  TypeSpec ::
+    HasSpec univ a =>
+    TypeSpec univ a ->
+    OrdSet a ->
```
Soupstraw, line 521: What's the purpose of this `OrdSet`?

MaximilianAlgehed: The set of things it _can't_ be

libs/constrained-generators/src/Constrained/Base.hs
```diff
@@ -0,0 +1,2680 @@
...
+noDependencies :: HasVariables univ t => t -> DependGraph univ
+noDependencies (freeVarSet -> xs) = Graph.noDependencies xs
+
+type Hints univ = DependGraph univ
+
+-- TODO: here we can compute both the explicit hints (i.e. constraints that
+-- define the order of two variables) and any whole-program smarts.
+computeHints :: forall m univ. MonadGenError m => [Pred univ] -> m (Hints univ)
+computeHints ps =
+  transitiveClosure <$> pure (fold [x `irreflexiveDependencyOn` y | DependsOn x y <- ps])
```
Soupstraw:
```suggestion
computeHints :: forall univ. [Pred univ] -> Hints univ
computeHints ps =
  transitiveClosure $ fold [x `irreflexiveDependencyOn` y | DependsOn x y <- ps]
```
The monad seems redundant here

libs/constrained-generators/src/Constrained/Base.hs
```diff
@@ -0,0 +1,2751 @@
...
+
+runTerm :: MonadGenError m => Env -> Term univ a -> m a
+runTerm env = \case
+  Lit a -> pure a
+  V v -> findEnv env v
+  App f ts -> do
+    vs <- mapMList (fmap Identity . runTerm env) ts
+    pure $ uncurryList_ runIdentity (sem f) vs
+
+assertPred :: (MonadGenError m, FunctionLike univ) => Env -> Pred univ -> m Property
```
Soupstraw: This looks really similar to `checkPred`, would this work?
```
assertPred e p = property <$> checkPred e p
```

MaximilianAlgehed: You wouldn't get the right counterexample unfortunately.

I've been thinking about how to restructure this to make it nicer. Probably one should just go via `GE` and sort it out that way.

libs/constrained-generators/src/Constrained/Base.hs
```diff
@@ -0,0 +1,2751 @@
...
+    Member
+      | HOLE :? Value s :> Nil <- ctx -> caseBoolSpec spec $ \case
+          True -> MemberSpec $ Set.toList s
+          False -> notMemberSpec s
+      | Value e :! NilCtx HOLE <- ctx -> caseBoolSpec spec $ \case
+          True -> typeSpec $ SetSpec (Set.singleton e) mempty mempty
+          False -> typeSpec $ SetSpec mempty (notEqualSpec e) mempty
+    Elem
+      | HOLE :? Value es :> Nil <- ctx -> caseBoolSpec spec $ \case
+          True -> MemberSpec $ es
```
Soupstraw:
```suggestion
          True -> MemberSpec es
```

MaximilianAlgehed:
@lehins and @Soupstraw I've marked this ready for review - it would be great if you wanted to take a look. There is still quite a lot left to do in terms of documentation and fixing some remaining issues, but I'm at the point where I'm anyway doing a lot of cleanup and writing a bunch of documentation so having some inputs during the process wouldn't be a bad thing.

libs/constrained-generators/src/Constrained/Univ.hs
```diff
@@ -0,0 +1,313 @@
...
+  MapFn :: univ '[a] b -> ListFn univ '[[a]] [b]
+
+-}
+
+newtype Fix fn (as :: [Type]) b = Fix (fn (Fix fn) as b)
+
+instance Member fn (fn' (Fix fn')) => Member fn (Fix fn') where
+  injectFn = Fix . injectFn
+
+deriving via (fn (Fix fn)) instance FunctionLike (fn (Fix fn)) => FunctionLike (Fix fn)
```
Soupstraw, line 65: Just to test my intuition here, would `deriving newtype` here give the same result?

MaximilianAlgehed: Possibly. But I don't know precisely how `deriving newtype` works.

libs/constrained-generators/src/Constrained/Base.hs
```diff
@@ -0,0 +1,3082 @@
...
+    -- | It must be an element of this set
+    OrdSet a ->
+    Spec univ a
+  -- | The empty set
+  ErrorSpec ::
+    [String] ->
+    Spec univ a
+  -- | The set described by some predicates
+  -- over the bound variable
+  SuspendedSpec ::
```
Soupstraw, line 508: I'm not really sure what this does, would be good to have some comments explaining this constructor.

MaximilianAlgehed: C.f. `checkPred` for the semantics, basically the variable ranges over the values denoted by the spec and the variable is subject to the `Pred`

MaximilianAlgehed: Sorry, not `checkPred` obviously... I mean `conformsToSpecM`

libs/constrained-generators/constrained-generators.cabal
```diff
@@ -0,0 +1,65 @@
...
+        Constrained.Instances
+        Constrained.Spec
+        Constrained.List
+        Constrained.GenT
+        Constrained.Univ
+        Constrained.Graph
+        Constrained.Spec.Generics
+        Constrained.Spec.Pairs
+        Constrained.Spec.Maps
+        Constrained.Test
```
Soupstraw, line 34: I think it'd be nice to have some simple API for using the constrained generators, right now everything is exported including a lot of internal functions which require a deep understanding of the generators. 

As a user I'd be interested in writing predicates, calculating a spec from the predicates and then using the spec to either generate values or test some existing value against the predicates. Then the internal stuff could be added as hidden modules. Do you think it'd be viable to have such an API?

MaximilianAlgehed: There is now a module called `Constrained` that aims to be the _minimal_ interface you need to work comfortably with this stuff. However, that still exposes a lot of details that could be better handled with a slightly nicer interface.

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/V2/Conway/DelegCert.hs
```diff
@@ -0,0 +1,51 @@
...
+          ( branch $ \sc mc ->
+              onJust' mc $ \c ->
+                elem_ (pair_ sc c) $ lit [(k, r) | (k, RDPair (fromCompact -> r) _) <- Map.toList rewardMap]
+          )
+          -- ConwayDelegCert !(StakeCredential c) !(Delegatee c)
+          (branch $ \sc _ -> member_ sc $ lit (Map.keysSet delegMap))
+          -- ConwayRegDelegCert !(StakeCredential c) !(Delegatee c) !Coin
+          ( branch $ \_ _ c ->
+              c ==. lit (pp ^. ppKeyDepositL)
+          )
```
Soupstraw: This should be a bit different: `c` should be equal to `ppKeyDeposit` only if the stake credential is not yet in the reward addresses map, and it should be zero otherwise.

We can fix these things in another PR though.

libs/cardano-ledger-test/src/Test/Cardano/Ledger/STS.hs
```diff
@@ -495,154 +59,234 @@ stsProperty ::
...
+          uncurry forAllShrinkBlind (genShrinkFromSpec $ specSig env st) $ \sig ->
+            counterexample (show $ ppString "sig = " <> prettyA sig) $
+              runShelleyBase $ do
+                res <- applySTS @(EraRule r era) $ TRC (env, st, sig)
+                pure $ case res of
+                  Left pfailures -> counterexample (show $ ppList prettyA pfailures) $ property False
+                  Right st' ->
+                    counterexample (show $ ppString "st' = " <> prettyA st') $
+                      conformsToSpec @fn st' (specState env)
+                        .&&. prop env st sig st'
```
Soupstraw, line 87: Continuation passing style makes this look really ugly with all the staircasing. Would it be possible to do this with monadic binds instead?

libs/constrained-generators/src/Constrained/Spec/Maps.hs
```diff
@@ -0,0 +1,196 @@
...
+import Constrained.GenT
+import Constrained.Instances ()
+import Constrained.List
+import Constrained.Univ
+
+------------------------------------------------------------------------
+-- HasSpec
+------------------------------------------------------------------------
+
+data MapSpec fn k v = MapSpec (Set k) [v] (Spec fn Int) (Spec fn (k, v)) (FoldSpec fn v)
```
Soupstraw: It'd be good to have some way of telling what each of the fields does.

libs/cardano-ledger-test/src/Test/Cardano/Ledger/STS.hs
MaximilianAlgehed: Yes. I agree. Fourmolololol made me do it! I don't know if there is a monadic interface for QuickCheck that will let me write this conveniently and have it mean the same thing, but I'll give it a try.

@lehins case in point ;)

MaximilianAlgehed: For example, if one were to switch to `PropertyM` and use `forAllM` we would lose shrinking completely. (For whatever reason, `forAllM` quantified values don't shrink).

lehins: Thank god for fourmolu. I can't stand when people format CPS with the same indentation.  

Why don't you use a single call to `forAllShrinkBlind`? and generate a three tuple, instead of a nested calls to `forAllShrinkBlind`

MaximilianAlgehed: That doesn't solve the problem with `counterexample`, and it means having to introduce a bunch of extra machinery just to make this piece of code look slightly prettier. For example, what you're suggesting isn't even well-typed!

MaximilianAlgehed: Without fourmolu you could write it like this:
```
  -- Generate the environment
  uncurry forAllShrinkBlind (genShrinkFromSpec specEnv)          $ \env ->
  counterexample (show $ ppString "env = " <> prettyA env) $

  -- Generate the state
  uncurry forAllShrinkBlind (genShrinkFromSpec $ specState env)  $ \st ->
  counterexample (show $ ppString "st = " <> prettyA st) $

  -- Generate the signal
  uncurry forAllShrinkBlind (genShrinkFromSpec $ specSig env st) $ \sig ->
  counterexample (show $ ppString "sig = " <> prettyA sig) $

  -- Run the rule and check that the result conforms to the spec
  runShelleyBase $ do
    res <- applySTS @(EraRule r era) $ TRC (env, st, sig)
    pure $ case res of
      Left pfailures -> counterexample (show $ ppList prettyA pfailures) $ property False
      Right st' ->
        counterexample (show $ ppString "st' = " <> prettyA st') $
          conformsToSpec @fn st' (specState env)
            .&&. prop env st sig st'
```

However, here are a few things you _can't_ do:
1. Generate all the things at the same time (because if generation fails for one thing you wouldn't get a QuickCheck `counterexample` for the earlier things)
2. Replace the multiple calls to `counterexample` with one call to `counterexample` (for the same reason, that means that if generation fails for the signal you don't get to see what the environment and the state were).

MaximilianAlgehed: The interface to QuickCheck really does force you to write your code like this, and fourmolu then makes it unreadable...

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/V2/Conway.hs
```diff
@@ -0,0 +1,1254 @@
...
+{-# LANGUAGE RecordWildCards #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE StandaloneDeriving #-}
+{-# LANGUAGE TypeApplications #-}
+{-# LANGUAGE TypeFamilies #-}
+{-# LANGUAGE TypeOperators #-}
+{-# LANGUAGE UndecidableInstances #-}
+{-# OPTIONS_GHC -Wno-orphans #-}
+
+module Test.Cardano.Ledger.Constrained.V2.Conway where
```
TimSheard: -- | In this module we define the HasSpec instances for all the types that appear in the Conway era. We also define new type synonyms (like: ConwayTxBodyTypes) for the  data types that are used in Conway.

Add a comment like this.

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/V2/Conway/CertState.hs
```diff
@@ -0,0 +1,50 @@
+{-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE RecordWildCards #-}
+
+module Test.Cardano.Ledger.Constrained.V2.Conway.CertState where
+
```
TimSheard: -- | In this module we define constraints for the data types used in the "CERTS" STS rules.

Add similar comments to each of the files in this directory.

libs/constrained-generators/src/Constrained.hs
```diff
@@ -0,0 +1,110 @@
+{-# LANGUAGE ImportQualifiedPost #-}
+{-# LANGUAGE PatternSynonyms #-}
+
+module Constrained (
+  Spec (..), -- TODO: it would be nice not to expose this whole thing
```
TimSheard: Is the intention here to re-export everything one might need.
Might want to add a comment

libs/constrained-generators/src/Constrained/Base.hs
```diff
@@ -0,0 +1,3552 @@
...
+
+fst p = x
+snd p = y
+x < y
+
+The dependency graph for these constraints will be the following:
+
+x <- y
+p <- x
+
```
TimSheard: There is a bit of possible confusion here
You have two ideas with similar notation

dependency graph  
  ```
x <- y
  p <- x
```

Ordering
   `z -> x -> y `

Maybe something to make it clear we are talking about two things that are related (so they have similar, but dual, notation)

File: libs/constrained-generators/src/Constrained/Base.hs [hunk 26]
```diff
@@ -0,0 +1,3552 @@
...
+Which is semantically equivalent to:
+
+exists $ \ x ->
+  tm == x
+  preds
+
+-- Reify ----------------------------------------------------------------
+
+Sometimes it's important to be able to perform complex calculations on data
+to obtain values that further constraint later variables. For this purpose
```
TimSheard: constraint -> constrain

libs/constrained-generators/src/Constrained/Base.hs
```diff
@@ -0,0 +1,3552 @@
...
+    Term fn t ->
+    Binder fn a ->
+    Pred fn
+  Case ::
+    HasSpec fn (SumOver as) =>
+    Term fn (SumOver as) ->
+    -- | Each branch of the type is bound with
+    -- only one variable because `as` are types.
+    -- Constructors with multiple arguments are
+    -- encoded with `ProdOver`.
```
TimSheard: Can't find a definition for ProdOver

libs/constrained-generators/src/Constrained/Base.hs
```diff
@@ -0,0 +1,3552 @@
...
+  p <> p' = Block (unpackPred p ++ unpackPred p')
+    where
+      unpackPred (Block ps) = ps
+      unpackPred p = [p]
+
+instance IsUniverse fn => Monoid (Pred fn) where
+  mempty = TruePred
+
+-- | Contexts for Terms, basically a term with a single HOLE
+-- instead of a variable.
```
TimSheard: Add some thing like
This is how we represent a constraint with exactly one free variable. It appears in the hole.

libs/constrained-generators/src/Constrained/Base.hs
```diff
@@ -0,0 +1,3552 @@
...
+  mempty = FreeVars mempty
+
+freeVar :: Name fn -> FreeVars fn
+freeVar n = FreeVars $ Map.singleton n 1
+
+without :: Foldable t => FreeVars fn -> t (Name fn) -> FreeVars fn
+without (FreeVars m) remove = FreeVars $ foldr Map.delete m (toList remove)
+
+class HasVariables fn a | a -> fn where
+  freeVars :: a -> FreeVars fn
```
TimSheard, line 1106: Lots of personal experience makes we want to suggest freeVars has type:: a -> FreeVars fn -> FreeVars fn

libs/constrained-generators/src/Constrained/Env.hs
```diff
@@ -0,0 +1,59 @@
...
+data EnvKey where
+  EnvKey :: Typeable a => !(Var a) -> EnvKey
+
+instance Eq EnvKey where
+  EnvKey v == EnvKey v' = isJust $ eqVar v v'
+
+instance Ord EnvKey where
+  compare (EnvKey v) (EnvKey v')
+    | Just v'' <- cast v = compare v'' v'
+    | otherwise = compare (typeRep v) (typeRep v')
```
TimSheard: Could you explain this use of cast

libs/constrained-generators/src/Constrained/Graph.hs
```diff
@@ -0,0 +1,96 @@
...
+transitiveClosure :: Ord node => Graph node -> Graph node
+transitiveClosure g = foldMap (\x -> dependency x (transitiveDependencies x g)) (nodes g)
+
+noDependencies :: Ord node => Set node -> Graph node
+noDependencies ns = Graph nodeMap nodeMap
+  where
+    nodeMap = Map.fromList ((,mempty) <$> Set.toList ns)
+
+-- TODO: maybe return the cycle if it fails
+topsort :: Ord node => Graph node -> Maybe [node]
```
TimSheard: Yes, the cycle really helps figure what went wrong

libs/constrained-generators/src/Constrained/Instances.hs
```diff
@@ -0,0 +1,67 @@
...
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE StandaloneDeriving #-}
+{-# LANGUAGE TypeApplications #-}
+{-# LANGUAGE TypeFamilies #-}
+{-# LANGUAGE UndecidableInstances #-}
+{-# OPTIONS_GHC -fno-warn-orphans #-}
+
+module Constrained.Instances where
+
```
TimSheard: Maybe a comment about what this module does?

libs/constrained-generators/src/Constrained/List.hs
```diff
@@ -0,0 +1,139 @@
...
+{-# LANGUAGE TypeOperators #-}
+{-# LANGUAGE ViewPatterns #-}
+
+module Constrained.List where
+
+import Data.Functor.Const
+import Data.Kind
+
+-- | A heterogeneous list / an inductive tuple
+data List (f :: k -> Type) (as :: [k]) where
```
TimSheard: Add something like this is the tool we use to define Ctxs
Every Ctx with a hole of type 'a looks something like this
List Value before ++ [Hole a a] ++ List Value after

libs/constrained-generators/src/Constrained/List.hs
```diff
@@ -0,0 +1,139 @@
...
+  curryList_ _ f = f Nil
+  unfoldList _ = Nil
+
+instance TypeList as => TypeList (a : as) where
+  uncurryList f (a :> as) = uncurryList (f a) as
+  uncurryList_ k f (a :> as) = uncurryList_ k (f $ k a) as
+  curryList f a = curryList (\xs -> f (a :> xs))
+  curryList_ p f a = curryList_ p (\xs -> f (p a :> xs))
+  unfoldList f = let xs = unfoldList f in f xs :> xs
+
```
TimSheard, line 150: There is a pattern here that people who have never seen these things often miss. It is worth pointing it out. Every instance for a class (say Foo) on TypeList as has this form:

```
instance Foo (TypeList '[]) where ...

instance Foo (TypeList as) => Foo (TypeList (a : as) where ...
```

Its like a recursive function. There are two instances, one for the Base case, and one for the inductive step.

libs/constrained-generators/src/Constrained/Test.hs
```diff
@@ -0,0 +1,552 @@
...
+-- prop_propagateSpec var tm spec =
+--   let spec' = errorGE $ do
+--         ctx <- toCtx var tm
+--         pure $ propagateSpec spec ctx
+--    in QC.forAll (strictGen $ genFromSpec spec') $ \geval -> fromGE (\err -> counterexample (unlines err) False) $ do
+--         val <- geval
+--         res <- runTerm (singletonEnv var val) tm
+--         pure $ property $ conformsToSpec res spec
+
+-- prop_mapSpec ::
```
TimSheard: If we eventually want to uncomment these. Add some TODO comments. If not we should consider deleting them.

libs/constrained-generators/src/Constrained/Univ.hs
```diff
@@ -0,0 +1,318 @@
...
+that you can plug-in your own functions into the universe.
+
+That might sound like it's no problem, but the issue is with higher-order
+functions. Consider:
+
+data ListFn where
+  MapFn :: _ '[a] b -> ListFn '[[a]] [b]
+
+What should go in place of `_`? It turns out that you need to specify the global
+function universe and make sure the global universe is the same everywhere.
```
TimSheard, line 53: .. and make sure the same global universe applys to every element of the List as well (hence the '[[a]])

libs/constrained-generators/src/Constrained/Univ.hs
```diff
@@ -0,0 +1,318 @@
...
+-- Universes
+------------------------------------------------------------------------
+
+-- TODO: there are faster implementation of this
+data
+  Oneof
+    fn
+    fn'
+    -- NOTE: everything needs to be quantified over the
+    -- fnUerse because otherwise it would be impossible
```
TimSheard: fnUerse to fn Universe?

TimSheard (approved):
This looks great.  Please address the conversations by improving the comment where indicated

libs/cardano-ledger-test/src/Test/Cardano/Ledger/STS.hs
lehins: > fourmolu then makes it unreadable...

On the contrary, this is horrendous and unreadable: 

```
  -- Generate the environment
  uncurry forAllShrinkBlind (genShrinkFromSpec specEnv)          $ \env ->
  counterexample (show $ ppString "env = " <> prettyA env) $

  -- Generate the state
  uncurry forAllShrinkBlind (genShrinkFromSpec $ specState env)  $ \st ->
  counterexample (show $ ppString "st = " <> prettyA st) $

  -- Generate the signal
  uncurry forAllShrinkBlind (genShrinkFromSpec $ specSig env st) $ \sig ->
  counterexample (show $ ppString "sig = " <> prettyA sig) $
```

It is totally unclear where the scope of one functions begins and the other one ends. I would never allow code like this to be merged, even without fourmolu.

So, let's forget about the style and focus on semantics! That is precisely what the automatic formatter is for, so people can put their opinions about the style aside and focus on what is important.

Question was: can we change from the CPS style to a more readable one?
Seems like the answer is: no.

So. I am marking this issue as resolved.

MaximilianAlgehed: Let's agree to disagree. But if line breaks and lambda bother you, do line breaks in do statements also make the code horrendous?

lehins: In my opinion current code looks totally fine. I am not bothered neither by lambdas, nor by stair stepping. I personally was more curious if it was possible to make it a little simpler, but if that is too much work or is not possible then there is nothing need to be done here.

libs/constrained-generators/src/Constrained/Univ.hs
MaximilianAlgehed: The list `'[[a]]` here is just saying that the function takes `[a]` as argument. At this point the arguments have nothing to do with the universe at all.

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/V2/Conway/Ledger.hs
MaximilianAlgehed: It's very odd you ended up looking at a version that doesn't contain anything - it should contain a basic constraint for transactions from the ledger rule state and env?

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/V2/Conway.hs
MaximilianAlgehed: Yes, `as` here is not a keyword.

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/V2/Conway/DelegCert.hs
MaximilianAlgehed: Yes, lets do an overhaul of these things in the next PR. I think there are a bunch of these that are slightly off and where it matters how you express the constraints. So we should maybe sit down together and try to work it out to get the tests running as well as possible.

libs/constrained-generators/src/Constrained/Base.hs
MaximilianAlgehed: That's possibly the right thing to do, but let's wait until we've been able to do some profiling to make this change?

libs/constrained-generators/src/Constrained/Base.hs
MaximilianAlgehed: It's in `Constrained.Univ` - I put a little pointer in the comment.

libs/constrained-generators/src/Constrained/Base.hs
```diff
@@ -0,0 +1,3571 @@
...
+{-# LANGUAGE TypeApplications #-}
+{-# LANGUAGE TypeFamilies #-}
+{-# LANGUAGE TypeOperators #-}
+{-# LANGUAGE UndecidableInstances #-}
+{-# LANGUAGE ViewPatterns #-}
+{-# OPTIONS_GHC -Wno-name-shadowing #-}
+{-# OPTIONS_GHC -Wno-orphans #-}
+{-# OPTIONS_GHC -Wno-redundant-constraints #-}
+
+module Constrained.Base where
```
Soupstraw, line 43: This module should probably be broken into multiple smaller modules at some point

Soupstraw (approved):
LGTM, best get this merged soon so we can start experimenting with it

libs/constrained-generators/src/Constrained/Base.hs
MaximilianAlgehed: That's unfortunately very difficult to do. I've tried, but every time I keep coming back to having to put it back together because there is some circular dependency that you can't break nicely.

The core issue is that `Pred` depends on `Term` which depends on `HasSpec` which depends on `Generics` which depends `Spec` which depends on `Pred`.

However, I'm also willing to entertain the idea of doing something "dependency injectiony" to break this chain. But it would make the code base more complex.

MaximilianAlgehed: (that cycle I gave is not the only example, as you can imagine)

Soupstraw: Yeah, no need to do this now, maybe some time in the future.

MaximilianAlgehed: I'll put in a comment explaining the problem in brief.

libs/constrained-generators/src/Constrained/Base.hs
MaximilianAlgehed: Yes! I'm now using:
```
x `dependsOn` y
p `dependsOn` x
```
instead, it's both clearer and mimics the surface syntax better. Nicely spotted.

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/V2/Conway.hs
MaximilianAlgehed: We now talk about this in a note in `HasSpec`

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/V2/Conway/CertState.hs
MaximilianAlgehed: It would probably be wise to rename these modules to be the name of the rule, so it would be `Test.Cardano.Ledger.Constrained.V2.Conway.CERTS` and `Test....Conway.POOL` etc.

libs/constrained/src/Constrained/GenT.hs
MaximilianAlgehed: I've looked some more at this and I don't think it's worth it, unfortunately.

The problem is that we have this strictness parameter (`GenMode`) that matters for the semantics of all the functions you would normally get "out of the box" with `GenT`. Consequently, what you would end up having to do is to:
- Implement our `GenT` as a newtype wrapper around `ReaderT GenMode (GenT m)`
- Not expose the newtype wrapper
- Not implement the `MonadGen` interface
- Get rid of our instances of `Applicative` and `Monad`
- Possibly simplify `pureGen`

As far as I can tell only the latter two are wins here, but at the cost of making the code more complicated / confusing.

lehins: It is worth it. You just don't have the full picture of ledger in mind. 

By using `quickcheck-transformer` we would be able to write custom generators using the `MonadGen` type class in other test suites in the ledger codebase and be able reuse them here. 

But, that's fine, we can solve this problem later, feel free to keep current definition as it is.

libs/cardano-ledger-test/cardano-ledger-test.cabal
```diff
@@ -152,7 +162,7 @@ test-suite cardano-ledger-test
     ghc-options:
         -Wall -Wcompat -Wincomplete-record-updates
         -Wincomplete-uni-patterns -Wpartial-fields -Wredundant-constraints
-        -Wunused-packages -rtsopts -threaded
+        -rtsopts -threaded
```
lehins: That is an important warning. Pleas don't disable compiler warnings without a really good reason. 
```suggestion
        -Wunused-packages -rtsopts -threaded
```

libs/cardano-ledger-test/cardano-ledger-test.cabal
```diff
@@ -177,7 +187,7 @@ benchmark bench
     ghc-options:
         -Wall -Wcompat -Wincomplete-record-updates
         -Wincomplete-uni-patterns -Wpartial-fields -Wredundant-constraints
-        -Wunused-packages -threaded -rtsopts -O2
```
lehins: Ditto

libs/cardano-ledger-test/cardano-ledger-test.cabal
```diff
@@ -217,7 +227,7 @@ benchmark benchProperty
...
```
lehins: Ditto

libs/cardano-ledger-test/cardano-ledger-test.cabal
```diff
@@ -89,9 +98,9 @@ library
     ghc-options:
         -Wall -Wcompat -Wincomplete-record-updates
         -Wincomplete-uni-patterns -Wpartial-fields -Wredundant-constraints
-        -Wunused-packages
```
lehins, line 92: Ditto

libs/constrained/src/Constrained/GenT.hs
MaximilianAlgehed: Read what I wrote again, we can't provide an instance for `MonadGen` for this type because that would break the intended behaviour. Therefore, it does not give us any reusability.

MaximilianAlgehed:
@lehins @Soupstraw and @TimSheard that's all the comments resolved for now, anything else just let me know. Now we have to try to get everything passing.

libs/constrained/src/Constrained/GenT.hs
lehins: > that would break the intended behaviour.

Could you be a little bit more specific?

libs/cardano-ledger-test/cardano-ledger-test.cabal
UlfNorell: Yes, this was a temporary work-around to make it possible to start a repl. For some reason cabal repl gives a bunch of spurious unused-package warnings.

libs/constrained/src/Constrained/GenT.hs
MaximilianAlgehed: > The problem is that we have this strictness parameter (GenMode) that matters for the semantics of all the functions you would normally get "out of the box" with GenT.

C.f. suchThatT for example.

UlfNorell: For instance, `listOfT` switches the `GenMode` to `Loose` when generating the elements (it's ok to fail occasionally when generating a list, you just get fewer elements). Using `listOf` from `MonadGen` would not have this behaviour. Most of the interesting combinators interact with the `GenMode` in some way.

libs/constrained-generators/src/Constrained/GenT.hs
```diff
@@ -0,0 +1,225 @@
...
+fromGEProp (Result es p) = foldr (counterexample . unlines) (property p) es
+
+fromGEDiscard :: Testable p => GE p -> Property
+fromGEDiscard (Result es p) = foldr (counterexample . unlines) (property p) es
+fromGEDiscard _ = discard
+
+headGE :: Foldable t => t a -> GE a
+headGE t
+  | null t = fatalError ["head of empty structure"]
+  | otherwise = pure $ head $ toList t
```
lehins: Could you please rewrite this without using `head`, because ghc-9.8 will complain with a warning

libs/cardano-ledger-test/cardano-ledger-test.cabal
lehins: @UlfNorell You can add this to you `cabal.project.local`, which will turn the errors into warnings only for you locally
```
if impl(ghc < 9.8)
  program-options
    ghc-options: -Wwarn
```

Note, that this is necessary `if impl(ghc < 9.8)` due to some cabal peculiarity, since that is how errors are turned on in `cabal.project`

UlfNorell:
This passes Haskell CI now. I had to do some acrobatics to make it build with 8.10.7, but nothing too horrendous.

I have no idea what's up with the Hydra failures. It fails on nix things that I'm pretty sure we haven't messed with, but I could be wrong. If someone could take a look that would be much appreciated.

UlfNorell:
@lehins Is the hydra failures anything to worry about or can we merge this?

lehins:
> Is the hydra failures anything to worry about or can we merge this?

Nope, nothing to worry about

PR #3896, opened 2023-11-29
Fix Brute force failure

TimSheard:
Fixed broken test, test comparing the brute force DRepDistr computation on the first MockChainState of an epoch to the Pulsed DrepDistr on the final MockhainState of the Epoch. 

This failed because we need the intermediate NES in the first state, since the LEDGERS rule changes the NES after the TICK rule runs. To do this we changed the internals of MockChainState from this
```
data MockChainState era = MockChainState
  { mcsNes :: !(NewEpochState era)
  , mcsLastBlock :: !SlotNo
  , mcsCount :: !Int -- Counts the blocks made
  }
```
to this
```
data MockChainState era = MockChainState
  { mcsNes :: !(NewEpochState era)
  , mcsTickNes ::  !(NewEpochState era)
  , mcsLastBlock :: !SlotNo
  , mcsCount :: !Int -- Counts the blocks made
  }
```

This stores the intermediate NewEpochState, after the TICK rule, but before the LEDGERS rule

# Description

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/Trace/DrepCertTx.hs
+++ b/libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/Trace/DrepCertTx.hs
@@ -194,7 +194,7 @@ pulserWorks mcsfirst mcslast =
         ++ "\nLast "
         ++ showPulserState mcslast
     )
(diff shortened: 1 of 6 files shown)


lehins (approved):
Beautiful. Simple and elegant

Issue #3406, opened 2023-05-08
Refactor Plutus interface


Soupstraw:
Currently some functions have constraints for both `PlutusV1` and `PlutusV2`. I think these constraints can be removed when we make `transScriptPurpose` a class method.

Soupstraw:
I tried to do some refactoring, but it looks like this is a lot of work. My idea was to create closed type families from `Language` to Plutus types. That way we can get a closed world on all Plutus types and that should make it easier to write and reuse functions that deal with Plutus. We'll also want to have the `EraPlutusContext` typeclass that takes both a Plutus version and an era as its parameters. This should allow us to easily map types from that era to the Plutus types each supported Plutus version (types such as DCert and ScriptPurpose).

The function that caused the most problems with this approach was `collectTwoPhaseScriptInputs` in `Cardano.Ledger.Alonzo.PlutusScriptApi`. The core of the problem was that the function outputs a list that contains a monomorphic `Language` type. We want this function to work on both Plutus V1 and V2 scripts. In that function we have to call the `txInfo` method, which is supposed to return a `TxInfo` corresponding to the language version we're interested in. However, since the language value that we have to match on is monomorphic, we don't learn the actual type of `TxInfo`. 

I think we'll have to revisit the return type of `collectTwoPhaseScriptInputs` and consider using existential types to address these issues. This would probably mean a lot of changes in the codebase, but the result will be more type safe.

lehins:
Fixed in #3899

PR #3893, opened 2023-11-27
Move `tree-diff` dependency to tests together with all instances

jasagredo:
# Description

`tree-diff` uses GPL license so it should not be in the non-testing code in our case. This PR moves all the instances to the test packages so that only when testing this dependency is pulled in. See how when requesting to build just the libraries, no `tree-diff` is pulled from hackage:

```
cardano-ledger [ master][✘»!+?][λ 9.6.3]
❯ cabal clean

cardano-ledger [ master][✘»!+?][λ 9.6.3]
❯ rm -rf ~/.cabal/store/ghc-9.6.3/tree-diff-*

cardano-ledger [ master][✘»!+?][λ 9.6.3]
❯ cabal build cardano-ledger-shelley:lib:cardano-ledger-shelley cardano-ledger-byron:lib:cardano-ledger-byron cardano-ledger-allegra:lib:cardano-ledger-allegra cardano-ledger-mary:lib:cardano-ledger-mary cardano-ledger-alonzo:lib:cardano-ledger-alonzo cardano-ledger-conway:lib:cardano-ledger-conway cardano-ledger-babbage:lib:cardano-ledger-babbage cardano-ledger-api:lib:cardano-ledger-api set-algebra:lib:set-algebra cardano-crypto-wrapper:lib:cardano-crypto-wrapper cardano-data:lib:cardano-data cardano-ledger-binary:lib:cardano-ledger-binary vector-map:lib:vector-map small-steps:lib:small-steps non-integral:lib:non-integral --dry-run
Build profile: -w ghc-9.6.3 -O1
In order, the following would be built (use -v for more details):
 - non-integral-1.0.0.0 (lib) (first run)
 - small-steps-1.0.0.0 (lib) (first run)
 - vector-map-1.0.1.0 (lib) (first run)
 - cardano-ledger-binary-1.2.1.1 (lib) (first run)
 - cardano-data-1.1.2.0 (lib) (first run)
 - cardano-crypto-wrapper-1.5.1.0 (lib) (first run)
 - set-algebra-1.1.0.1 (lib) (first run)
 - cardano-ledger-byron-1.0.0.4 (lib) (first run)
 - cardano-ledger-core-1.10.0.0 (lib:internal) (first run)
 - cardano-ledger-core-1.10.0.0 (lib) (first run)
 - cardano-ledger-shelley-1.9.0.0 (lib:internal) (first run)
 - cardano-ledger-shelley-1.9.0.0 (lib) (first run)
[...]

PR #3899, opened 2023-11-30
Bring Plutus language version to the type level

### LINKED ISSUE #3849: Bring Plutus language version to the type level
We are introducing a whole lot of new features in Conway that will be available in PlutusV3. In order to have sane implementation of Plutus context translation we need to have Plutus language at the type level. This will require restructure of how we deal with Plutus in previous Alonzo and Babbage eras. It will be a nice cleanup and safety improvement for those eras as well and for eras going forward.

This is currently blocking:
* #3788 
* #3633

lehins:
# Description

Fixes #3849 and #3406

Other improvements:

* Instead of accepting `UTxO` the validation function `ppViewHashesMatch` now accepts  `ScriptsProvided` and remove the no longer needed `languages` function. This approach avoids duplicate computation of `ScritpsProvided`, which was done in the `languages` function.

# Checklist

- [x] Self-reviewed the diff


libs/cardano-ledger-test/src/Test/Cardano/Ledger/Examples/AlonzoBBODY.hs
teodanciu, line 1: These `Example` tests from this folder must have been a nightmare to adjust. :grimacing:

eras/babbage/test-suite/test/Tests.hs
```diff
@@ -25,7 +25,6 @@ defaultTests =
     [ Golden.tests
     , Tripping.tests
     , txInfoTests (Proxy @Babbage)
-    , txInfoTestsBabbageOnly
```
Soupstraw, line 28: Why was this test removed?

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/TreeDiff.hs
```diff
@@ -29,6 +32,12 @@ instance ToExpr PoolVotingThresholds
 
 instance ToExpr DRepVotingThresholds
 
+-- Scritps
```
Soupstraw:
```suggestion
-- Scripts
```

eras/conway/test-suite/test/Tests.hs
```diff
@@ -4,6 +4,7 @@ module Main where
 
 import Cardano.Ledger.Conway (Conway)
 import Data.Proxy (Proxy (..))
+
```
Soupstraw: Is there a reason for this newline?

libs/cardano-ledger-core/src/Cardano/Ledger/Plutus/CostModels.hs
```diff
@@ -205,6 +206,9 @@ getCostModelLanguage (CostModel lang _ _) = lang
 getCostModelParams :: CostModel -> [Integer]
 getCostModelParams (CostModel _ cm _) = cm
 
+getCostModelEvaluationContext :: CostModel -> PV1.EvaluationContext
+getCostModelEvaluationContext = cmEvalCtx
```
Soupstraw, line 210: Could we use the field accessor directly?

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Examples/AlonzoInvalidTxUTXOW.hs
```diff
@@ -925,7 +930,7 @@ noCostModelTx pf =
         , Collateral' [mkGenesisTxIn 11]
         , Outputs' [newTxOut pf [Address (someAddr pf), Amount (inject $ Coin 4995)]]
         , Txfee (Coin 5)
-        , WppHash (newScriptIntegrityHash pf (pp pf) [PlutusV2] redeemers (mkTxDats (Data (PV1.I 123))))
+        , WppHash (newScriptIntegrityHash pf pp' [PlutusV1] redeemers (mkTxDats (Data (PV1.I 123))))
```
Soupstraw, line 933: Why was this changed to `PlutusV1`?

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Examples/AlonzoInvalidTxUTXOW.hs
```diff
@@ -343,11 +344,14 @@ alonzoUTXOWTests pf =
...
-              (trustMeP pf True $ noCostModelTx pf)
-              ( Left [fromUtxos @era (CollectErrors [NoCostModel PlutusV2])]
-              )
+            let pp' = updatePParams pf (pp pf) (Costmdls emptyCostModels)
+             in testUTXOW
+                  (UTXOW pf)
+                  (initUTxO pf)
+                  pp'
+                  (trustMeP pf True $ noCostModelTx pf pp')
+                  ( Left [fromUtxos @era (CollectErrors [NoCostModel PlutusV1])]
```
Soupstraw, line 353: Why was this changed to `PlutusV1`?

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Examples/AlonzoCollectInputs.hs
```diff
@@ -164,25 +165,26 @@ collectInputs ::
...
-  Either (TranslationError (EraCrypto era)) VersionedTxInfo
-getTxInfo (Alonzo _) = txInfo
-getTxInfo (Babbage _) = txInfo
-getTxInfo (Conway _) = txInfo
-getTxInfo era = error ("getTxInfo Not defined in era " ++ show era)
+  Either (ContextError era) (Data era)
+mkPlutusScriptContext' (Alonzo _) = mkPlutusScriptContext
+mkPlutusScriptContext' (Babbage _) = mkPlutusScriptContext
+mkPlutusScriptContext' (Conway _) = mkPlutusScriptContext
+mkPlutusScriptContext' era = error ("mkPlutusScriptContext is not defined in era " ++ show era)
```
Soupstraw, line 187:
```suggestion
mkPlutusScriptContext' era = error ("mkPlutusScriptContext' is not defined in era " ++ show era)
```

libs/cardano-ledger-core/src/Cardano/Ledger/Plutus/TxInfo.hs
```diff
@@ -155,17 +134,12 @@ transCred (KeyHashObj (KeyHash kh)) =
...
-
-transTxOutAddr :: EraTxOut era => TxOut era -> Maybe PV1.Address
-transTxOutAddr txOut = do
-  -- filter out Byron addresses without uncompacting them
-  case txOut ^. bootAddrTxOutF of
-    Just _ -> Nothing
-    -- The presence of a Byron address is caught above in the Just case
-    Nothing -> transAddr (txOut ^. addrTxOutL)
+transAddr addr = do
+  Addr _netId object stake <- Just addr
```
Soupstraw:
```suggestion
  let Addr _netId object stake = addr
```

Soupstraw: Actually, can this function even return a `Nothing`? It should probably just return a `PV1.Address`

Soupstraw: Ok, I see what it's doing now, but I'm not sure I prefer it over the old implementation

libs/cardano-ledger-core/src/Cardano/Ledger/Plutus/TxInfo.hs
```diff
@@ -179,46 +153,17 @@ slotToPOSIXTime ei sysS s = do
...
-      Map.insert (PV1.StakingHash (transCred cred)) n ans
-
-getWitVKeyHash :: (Crypto c, Typeable kr) => WitVKey kr c -> PV1.PubKeyHash
-getWitVKeyHash =
-  PV1.PubKeyHash
-    . PV1.toBuiltin
-    . hashToBytes
-    . (\(KeyHash x) -> x)
-    . hashKey
-    . (\(WitVKey x _) -> x)
```
Soupstraw, line 218: Were these unused all this time?

libs/cardano-ledger-core/src/Cardano/Ledger/Plutus/Language.hs
```diff
@@ -63,46 +76,105 @@ import Data.Aeson (
...
+  serialisedScript,
+ )
+import qualified PlutusLedgerApi.V1 as PV1
+import qualified PlutusLedgerApi.V2 as PV2
+import qualified PlutusLedgerApi.V3 as PV3
 
-data Plutus = Plutus
-  { plutusLanguage :: !Language
-  , plutusScript :: !BinaryPlutus
+newtype PlutusRunnable (l :: Language) = PlutusRunnable
```
Soupstraw, line 106: I think it'd help a lot to have docs for all of the new Plutus types, because it's a bit hard to keep track what they all do.

Soupstraw (commented):
Looks great so far. I'll finish the rest of the review tomorrow. Some of the comments I wrote down for myself as notes, so feel free to ignore them for now.

eras/babbage/test-suite/test/Tests.hs
lehins: That is because the only thing it was testing was presence of incompatible Plutus versions within a translation, which is no longer possible thanks to the type checker. 

https://github.com/input-output-hk/cardano-ledger/blob/bcf6051f4a0b69c82e4de7327d7d8801fa1bad87/eras/babbage/test-suite/src/Test/Cardano/Ledger/Babbage/TxInfo.hs#L346C1-L351

eras/conway/test-suite/test/Tests.hs
lehins: Nope. Just an accident

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Examples/AlonzoInvalidTxUTXOW.hs
lehins: Because `PlutusV2` is not available in Alonzo.

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Examples/AlonzoInvalidTxUTXOW.hs
lehins: Because PlutusV2 is not available in Alonzo.

libs/cardano-ledger-core/src/Cardano/Ledger/Plutus/TxInfo.hs
lehins: `getWitVKeyHash` was really dead code, unused anywhere

`transWithdrawals` have moved to `Alonzo.TxInfo`, since it  can only be used in Alonzo and Babbage and there is a new implementation of this function in Conway

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Examples/AlonzoCollectInputs.hs
lehins: I've changed the error message to actually mean that the function `mkPlutusScriptContext` is not defined for that era. The fact that `mkPlutusScriptContext'` function is throwing an error will be apparent from the call stack. Which is why it usually makes no sense to add a name of the function into the error message (it was useful in the old days when `error` did not have `HasCallStack`).

libs/cardano-ledger-core/src/Cardano/Ledger/Plutus/CostModels.hs
lehins: Oh crap, thanks to your question I just found a bigger issue. `CostModel` constructor should not have been exported and the smart constructor `mkCostModel` must be used instead. That was why those accessor functions (`getCostModelLanguage` and `getCostModelParams`) where added in the first place.

So, the answer to your question is no. We should not be able to use the `cmEvalCtx` accessor directly.

It was accidentally exported in #3893. I've fixed it by hiding back the constructor.

eras/conway/test-suite/test/Tests.hs
lehins: Fixed

libs/cardano-ledger-core/src/Cardano/Ledger/Plutus/TxInfo.hs
lehins: Sure. Switched to an explicit case statement

libs/cardano-ledger-core/src/Cardano/Ledger/Plutus/Language.hs
lehins: Thank you for pointing out lack of haddock. I've added some. Please let me know if it can be improved.

lehins:
@Soupstraw I've addressed all of your comments.

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Plutus/TxInfo.hs
```diff
@@ -103,147 +73,188 @@ import Cardano.Ledger.Binary.Coders (
...
-instance Crypto c => DecCBOR (TranslationError c) where
-  decCBOR = decode (Summands "TranslationError" dec)
-    where
-      dec 0 = SumD ByronTxOutInContext <! From
-      dec 1 = SumD TranslationLogicMissingInput <! From
-      dec 2 = SumD RdmrPtrPointsToNothing <! From
-      dec 3 = SumD LanguageNotSupported <! From
-      dec 4 = SumD InlineDatumsNotSupported <! From
-      dec 5 = SumD ReferenceScriptsNotSupported <! From
-      dec 6 = SumD ReferenceInputsNotSupported <! From
```
Soupstraw, line 184: We got rid of all of this with type level magic? :astonished:

lehins: For Alonzo they are gone indeed. 

Those failures are all still present in Babbage and Conway, except for `LanguageNotSupported`, which was removed due to type level magic

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Scripts.hs
```diff
@@ -103,58 +154,108 @@ instance NoThunks Tag
...
+
+-- | Get value level plutus language of the plutus script
+plutusScriptLanguage :: AlonzoEraScript era => PlutusScript era -> Language
+plutusScriptLanguage ps = withPlutusScript ps plutusLanguage
+
+-- | Extract binary representation of the script.
+plutusScriptBinary :: AlonzoEraScript era => PlutusScript era -> PlutusBinary
+plutusScriptBinary ps = withPlutusScript ps plutusBinary
+
+-- | Verifies wether Plutus script is well formed or not, which simply means wether it
```
Soupstraw:
```suggestion
-- | Verifies whether Plutus script is well formed or not, which simply means whether it
```

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Scripts.hs
```diff
@@ -180,54 +281,63 @@ instance DecCBOR Tag where
...
-              PlutusV3 -> PV3.deserialiseScript
-          eWellFormed :: Either PV1.ScriptDecodeError PV1.ScriptForEvaluation
-          eWellFormed = deserialiseScript (transProtocolVersion pv) bytes
-       in isRight eWellFormed
+  case toPlutusScript script of
+    Just plutusScript -> isValidPlutusScript (pvMajor pv) plutusScript
+    Nothing ->
+      case getNativeScript script of
+        Just timelockScript -> deepseq timelockScript True
+        Nothing -> False -- Impossible case
```
Soupstraw: Maybe we should throw an error here instead? That's how we've usually done it for impossible cases.

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxAuxData.hs
```diff
@@ -136,37 +142,38 @@ instance Era era => EncCBOR (AlonzoTxAuxDataRaw era) where
...
 mkAlonzoTxAuxData atadrMetadata allScripts =
   mkMemoized $ AlonzoTxAuxDataRaw {atadrMetadata, atadrTimelock, atadrPlutus}
   where
-    partitionScripts (tss, pss1, pss2, pss3) =
+    partitionScripts (tss, pss) =
       \case
-        TimelockScript ts -> (ts :<| tss, pss1, pss2, pss3)
-        PlutusScript (Plutus PlutusV1 ps1) -> (tss, ps1 : pss1, pss2, pss3)
-        PlutusScript (Plutus PlutusV2 ps2) -> (tss, pss1, ps2 : pss2, pss3)
-        PlutusScript (Plutus PlutusV3 ps3) -> (tss, pss1, pss2, ps3 : pss3)
```
Soupstraw, line 151: What the hell is this, Haskell for cats? :cat:

lehins: :rofl:

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Scripts.hs
lehins: Sure, I can switch it to `error`.

lehins: Done.

eras/alonzo/test-suite/src/Test/Cardano/Ledger/Alonzo/Translation/TranslatableGen.hs
```diff
@@ -36,49 +49,65 @@ import Test.QuickCheck (
 import Test.QuickCheck.Gen (Gen (MkGen))
 import Test.QuickCheck.Random (mkQCGen)
 
+data TxInfoLanguage era where
```
Soupstraw, line 52: I think this could also do with some haddock.

eras/conway/test-suite/test/Test/Cardano/Ledger/Conway/TxInfo.hs
```diff
@@ -4,18 +4,12 @@
...
-import qualified Test.Cardano.Ledger.Babbage.TxInfo as B
+import Test.Tasty (TestTree, testGroup)
 
 txInfoTests ::
   forall era.
-  (ExtendedUTxO era, EraTx era, B.BabbageEraTxBody era, B.BabbageTxInfoTests era) =>
   Proxy era ->
   TestTree
-txInfoTests p = B.txInfoTestsV2 p PlutusV3 -- The V2 tests in Babbage should all hold for V3
+txInfoTests _p = testGroup "Conway Plutus Tests" [] -- Disabled for now
```
Soupstraw, line 14: Make sure to have an issue to re-enable this at some point.

libs/cardano-ledger-core/src/Cardano/Ledger/Plutus/Evaluate.hs
```diff
@@ -10,296 +10,219 @@
...
 import qualified PlutusLedgerApi.V2 as PV2
 import qualified PlutusLedgerApi.V3 as PV3
 import Prettyprinter (Pretty (..))
 
-data PlutusWithContext era = PlutusWithContext
-  { pwcScript :: !Plutus
-  , pwcDatums :: ![Data era]
-  , pwcExUnits :: !ExUnits
-  , pwcCostModel :: !CostModel
+data PlutusWithContext where
```
Soupstraw, line 80: This needs some haddock

libs/cardano-ledger-api/src/Cardano/Ledger/Api/Scripts/ExUnits.hs
```diff
@@ -51,44 +38,55 @@ import Cardano.Ledger.Alonzo.TxWits (
...
 data TransactionScriptFailure era
-  = -- | A redeemer was supplied that does not point to a
-    --  valid plutus evaluation site in the given transaction.
-    RedeemerNotNeeded !RdmrPtr !(ScriptHash (EraCrypto era))
-  | -- | A redeemer was supplied which points to a script hash which
+  = -- | A redeemer was supplied which points to a script hash which
     -- we cannot connect to a Plutus script.
     RedeemerPointsToUnknownScriptHash !RdmrPtr
-  | -- | Missing redeemer. The first parameter is the redeemer pointer which cannot be resolved,
+  | -- | Missing redeemer. The first parameter is the ,
```
Soupstraw: the what?

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Plutus/Evaluate.hs
```diff
@@ -75,43 +57,35 @@ import NoThunks.Class (NoThunks)
...
-
 -- ========================================================================
 
 -- | When collecting inputs for two phase scripts, 3 things can go wrong.
 data CollectError era
   = NoRedeemer !(ScriptPurpose era)
   | NoWitness !(ScriptHash (EraCrypto era))
   | NoCostModel !Language
-  | BadTranslation !(TranslationError (EraCrypto era))
+  | BadTranslation !(ContextError era)
```
teodanciu, line 67: I see none of the changes (including this, exporting lookupPlutusScript, removing collectTwoPhaseScriptInputs ) are mentioned in the CHANGELOG, but perhaps they are not intended to be used by anyone else.
`CollectError` I see is used the Utxos rule predicate failure, so maybe it would help to mention this change?

libs/cardano-ledger-api/src/Cardano/Ledger/Api/Scripts/ExUnits.hs
```diff
@@ -51,44 +38,55 @@ import Cardano.Ledger.Alonzo.TxWits (
...
+      !( Map
+          RdmrPtr
+          (ScriptPurpose era, Maybe (PlutusScript era), ScriptHash (EraCrypto era))
+       )
   | -- | Missing datum.
     MissingDatum !(DataHash (EraCrypto era))
   | -- | Plutus evaluation error, for any version
-    ValidationFailure ValidationFailed
+    ValidationFailure
+      -- | Supplied execution units in the transactoin, which were ignored for calculating
```
Soupstraw:
```suggestion
      -- | Supplied execution units in the transaction, which were ignored for calculating
```

libs/cardano-ledger-api/src/Cardano/Ledger/Api/Scripts/ExUnits.hs
```diff
@@ -183,185 +172,81 @@ evalTxExUnitsWithLogs ::
...
+      let ptrToPlutusScriptNoContext =
+            Map.map (\(sp, mps, sh) -> (sp, fst <$> mps, sh)) ptrToPlutusScript
+      (plutusScript, scriptContext) <-
+        note (MissingScript pointer ptrToPlutusScriptNoContext) mPlutusScript
+      let lang = plutusScriptLanguage plutusScript
+      costModel <-
+        note (NoCostModelInLedgerState lang) $ Map.lookup lang costModels
+      -- Similar to getSpendingDatum, but with more informative errors. It is OK to use
+      -- inline datums, when they are present, since for PlutusV1 presence of inline
+      -- datums would short circuit earlier on PlutuContext translation.
```
Soupstraw:
```suggestion
      -- datums would short circuit earlier on PlutusContext translation.
```

eras/conway/impl/src/Cardano/Ledger/Conway/TxInfo.hs
```diff
@@ -10,183 +11,301 @@
...
+
+instance NoThunks (ContextError (ConwayEra c))
+
+instance Crypto c => NFData (ContextError (ConwayEra c))
+
+instance Crypto c => EncCBOR (ContextError (ConwayEra c)) where
+  encCBOR = \case
+    -- We start at tag 8, just in case to avoid clashes with previous eras.
+    BabbageContextError babbageContextError ->
+      encode $ Sum BabbageContextError 8 !> To babbageContextError
```
Soupstraw, line 104: Why are we encoding an additional tag here? Couldn't we just use the same tag as the final error that we're actually encoding?

libs/cardano-ledger-core/src/Cardano/Ledger/Plutus/Language.hs
Soupstraw: Looks good to me

Soupstraw (approved):
This is a huge improvement. LGTM overall, but I'm sure I didn't catch everything because it's such a massive PR with lots of moving parts.

eras/alonzo/test-suite/src/Test/Cardano/Ledger/Alonzo/Translation/TranslatableGen.hs
lehins: I have a separate plan for PlutusContext golden tests, which will lead to this type disappearing, so I won't bother documenting it now.

eras/conway/test-suite/test/Test/Cardano/Ledger/Conway/TxInfo.hs
lehins: Done: https://github.com/input-output-hk/cardano-ledger/issues/3923

eras/alonzo/test-suite/src/Test/Cardano/Ledger/Alonzo/Translation/TranslatableGen.hs
lehins: This will be done as part of this ticket: https://github.com/input-output-hk/cardano-ledger/issues/3923

eras/conway/impl/src/Cardano/Ledger/Conway/TxInfo.hs
lehins: I don't understand your question. Could you make a code suggestion, so it is clear what you mean?

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Plutus/Evaluate.hs
lehins: They absolutely have to be in the CHANGELOG. Thank you for reminding me to finish adjusting changelogs. I stopped half way through the PR.

eras/conway/impl/src/Cardano/Ledger/Conway/TxInfo.hs
Soupstraw, line 104: Why not this?
```suggestion
      encCBOR babbageContextError
```

lehins: Oh, I see what you mean now. We'll use `Sum` coders because there will be a bunch of errors added here when we finish integration of the Conway features (eg. VotingNotSupported for PlutusV1 and PlutusV2)

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Plutus/Context.hs
```diff
@@ -0,0 +1,125 @@
...
+import Data.Kind (Type)
+import Data.Text (Text)
+import NoThunks.Class (NoThunks)
+import qualified PlutusLedgerApi.V1 as P (ToData, toData)
+import qualified PlutusLedgerApi.V1 as PV1
+import qualified PlutusLedgerApi.V2 as PV2
+import qualified PlutusLedgerApi.V3 as PV3
+
+class (PlutusLanguage l, AlonzoEraScript era) => EraPlutusTxInfo (l :: Language) era where
+  toPlutusTxCert :: proxy l -> TxCert era -> Either (ContextError era) (PlutusTxCert l)
```
teodanciu, line 44: I'm confused about the `ContextError` in these functions. Is it the `data  ContextError era :: Type` defined below in `EraPlutusContext era` class?  I suppose yes, but there needs to be nothing else specified to link the two in that case? Or is that happening when defining the instances? 
If the question doesn't make sense, please ignore it, I will ask you during the meeting tomorrow!

eras/conway/impl/src/Cardano/Ledger/Conway/TxInfo.hs
Soupstraw: Babbage also uses `Sum` coders, so we should invoke those from Conway as well. Otherwise the encoding will have redundant tags that won't add any useful information (e.g. `InlineDatumsNotSupported` would be encoded as `\x08\x04...` instead of just `\x04...`). And we can still add new Conway errors from tag 8 and upwards.

Soupstraw: I think that's how we do it for PredicateFailures also

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Plutus/Context.hs
lehins: I think I might understand the question. I can add `EraPlutusContext` as a superclass to the `EraPlutusTxInfo` if that would make it clearer. However, it is not necessary for a very interesting reason:

Associated data and type families (ones that are defined in a type class) are no different from the standalone data and type families and can in fact be treated the same at the use site. I think the only difference between the two is the fact that you can not create a type family instance without adding a type class instance for the associated type families.

eras/conway/impl/src/Cardano/Ledger/Conway/TxInfo.hs
lehins: Those are not the types that are being stored on chain, so we do not need to worry about representing them in the most efficient manner, if that is your concern.

If I am missing what you are trying to say please elaborate. Or maybe produce a full DecCBOR instance of what you mean, because I don't understand the issue that you are trying to address here.

Soupstraw: Yeah, I think we can leave it as is if we're not worried about efficiency.

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Plutus/Context.hs
teodanciu: Interesting, thank you for the explanation! :+1:

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Plutus/Evaluate.hs
lehins: I've added a bunch of missing changelog entries for `cardano-ledger-alonzo`

libs/cardano-ledger-core/src/Cardano/Ledger/Plutus/Evaluate.hs
lehins: Added some haddock

libs/cardano-ledger-api/src/Cardano/Ledger/Api/Scripts/ExUnits.hs
lehins: Thank you for catching this. I meant to remove the whole sentence, since I moved the documentation to the actual parameters (haddock nowadays supports this)

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Plutus/Context.hs
lehins: I did add `EraPlutusContext` as a superclass, which made a few constraints in one of the testing modules redundant :slightly_smiling_face:

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Plutus/Context.hs
```diff
@@ -0,0 +1,125 @@
...
+  , NFData (ContextError era)
+  , NoThunks (ContextError era)
+  , EncCBOR (ContextError era)
+  , DecCBOR (ContextError era)
+  ) =>
+  EraPlutusContext era
+  where
+  data ContextError era :: Type
+
+  mkPlutusScriptContext ::
```
teodanciu, line 79: So this function, ` mkPlutusScriptContext `, seems to be the "boss" . 
All the functions from `EraPlutusTxInfo` (  `toPlutusTxCert`, `toPlutusScriptPurpose`, `toPlutusTxInfo`, `toPlutusScriptContext`)   are used to build  `mkPlutusLanguageContext`, which in turn is used in each  `mkPlutusScriptContext ` implementation.  It is the one ultimately used in  in `Evaluate.collectPlutusScriptsWithContext` to build `PlutusWithContext`, which then gets passed to `evalPlutusScripts` in  Utxos rule. 
Not really a question or useful comment, sorry - just validating my understanding.  Anyway, such a clean model compared to `ExtendedUtxo` , `txInfo` and `valContext` - and their unclear (to me) interactions. 
Love how it all fits together so neatly.

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Plutus/Evaluate.hs
```diff
@@ -0,0 +1,214 @@
...
+import qualified Data.Set as Set
+import Data.Text (Text)
+import Debug.Trace (traceEvent)
+import GHC.Generics
+import Lens.Micro
+import NoThunks.Class (NoThunks)
+
+-- ===============================================================
+-- From the specification, Figure 8 "Scripts and their Arguments"
+-- ===============================================================
```
teodanciu, line 58: Out of curiosity, I checked the alonzo spec, and Figure 8 is now something else: 



Figure 7 is "Scripts and their arguments", but don't know, this comment seems pretty unnecessary to me , since the implementation is quite different.
Here is Figure 7:

teodanciu (approved):
Epic! 
The relationship between everything involved needed to execute the scripts is now so clear, as is, supporting a new Plutus Version, and/or Era. 
I'm in an awe of this work!  :bow:

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Plutus/Context.hs
lehins: You are right on point!

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Plutus/Evaluate.hs
lehins: Looks like the comment has off-by-one error :slightly_smiling_face: 

It should say:
```
 From the specification, Figure 7 "Scripts and their Arguments"
```

I'll fix it my subsequent PR.

lehins:
@Soupstraw and @teodanciu Thank you for reviewing this PR! I know it was not a small task.

PR #3902, opened 2023-11-30
Fix pvCanFollow usage in Conway and improve clarity in Shelley

### LINKED ISSUE #3873: Protect against invalid versions in HardForkInitiation
In the GOV rule we should have a predicate failure for invalid protocol versions.

Proposal should be rejected if it does not satisfy this function:
https://github.com/input-output-hk/cardano-ledger/blob/77853aee069ff7bdd7c0d7254c4c59d43dafc807/eras/shelley/impl/src/Cardano/Ledger/Shelley/PParams.hs#L527-L530

The important part of this validation is that comparison happens with either:
* current protocol version if it contains prevGovActionId for the last enacted GovAction
* or protocol version in the previous proposal otherwise

### LINKED ISSUE #3901: `pvCanFollow` bug
QuickCheck gives me the following counterexample when running `prop_GOV` on `constrained-v2`:
```
env = GovEnv
       { TxId =
           (TxId
              #"792ef2223a28f0b983a80496554bdba75cf59f81f9b6df4fbfbb10fe5e0aab34")
       , Epoch = EpochNo 0
       , PParams =
           PParams
            { Tau = 7127508027 % 3125000000000
            , Rho = 13 % 100
            , Prices =
                Prices
                 { prMem = 7415734805198999827 % 100000000000000
                 , prSteps = 8720716715657812863 % 100000000 }
            , PoolVotingThresholds =
                PoolVotingThresholds
                 { MotionNoConfidence = 89819943734781 % 125000000000000
                 , HardForkInitiation = 7977325399271 % 15625000000000
                 , CommitteeNormal = 36281717 % 100000000
                 , CommitteeNoConfidence = 1503 % 2500 }
            , PoolDeposit = (Coin 2)
            , NOpt = 2
            , MinPoolCost = (Coin 2)
            , MinFeeB = (Coin 1)
            , MinFeeA = (Coin 1)
            , CommitteeMinSize = 2
            , MaxValSize = 0
            , MaxTxSize = 0
            , MaxTxExUnits = ExUnits{ memory = 2, steps = 0 }
            , MaxCollateralInputs = 2
            , MaxBlockExUnits = ExUnits{ memory = 2, steps = 0 }
            , MaxBHSize = 1
            , MaxBBSize = 1
            , KeyDeposit = (Coin 2)
            , GovActionLifetime = EpochInterval 0
            , GovActionDeposit = (Coin 0)
            , EMax = EpochInterval 0
            , DRepVotingThresholds =
                DRepVotingThresholds
                 { UpdateToConstitution = 18457225352107 % 100000000000000
                 , TreasuryWithdrawal = 68786360827 % 100000000000
                 , PPTechnicalGroup = 1231013 % 10000000
                 , PPNetworkGroup = 5696218757238569771 % 10000000000000000000
                 , PPGovGroup = 8415522393753699 % 12500000000000000
                 , PPEconomicGroup = 7 % 100
                 , MotionNoConfidence = 3 % 10
                 , HardForkInitiation = 325243479 % 1000000000
                 , CommitteeNormal = 4252300940397 % 6250000000000
                 , CommitteeNoConfidence =
                     1598411698412947 % 10000000000000000 }
            , DRepDeposit = (Coin 2)
            , DRepActivity = EpochInterval 2
            , CostModels = Map{  }
            , CommitteeMaxTermLength = EpochNo 2
            , CollateralPercentage = 2
            , CoinsPerUTxOByte = (Coin 0)
            , A0 = 1374691538511693889 % 25 }
       , PrevGovActionId =
           PrevGovActionIds
            { LastPParamUpdate = ?-
            , LastHardFork = ?-
            , LastCommittee =
                ?PrevGovActionId {unPrevGovActionId = GovActionId {gaidTxId = TxId {unTxId = SafeHash "c477d1e2047c84c102e7d3ca44576aa18
d85377960fd2ed0b5c86a18ed243ac6"}, gaidGovActionIx = GovActionIx 1}}
            , LastConstitution =
                ?PrevGovActionId {unPrevGovActionId = GovActionId {gaidTxId = TxId {unTxId = SafeHash "3066937bcc940f95dc6249ea7cbdc55a3
b14468440233978217432546883d964"}, gaidGovActionIx = GovActionIx 0}} } }
st = [GovActionState
       { Id =
           GovActionId
            { Transaction ID =
                (TxId
                   #"e6ac71356c26c34cfb08093db61e333de74d155666bd53b34a401feabcd4b4e2")
            , Gov Action Index = 2 }
       , CommitteVotes = Map{  }
       , DRepVotes =
           Map
            { (ScriptCred
                 #"94c72545f9d795ec4b883bd543c3583a3b5776c37fe19611907c6679") ->
                VoteNo }
       , StakePoolVotes =
           Map
            { (KeyHash
                 #"32b42698bce528da9535e1b1d21bac65831b5ee5997fc321532039c8") ->
                VoteNo
            , (KeyHash
                 #"a025b7ee731ec5304e82c1ac41db1d7980b4db9c1d630c0440ab8f3d") ->
                Abstain }
       , Deposit = (Coin 1)
       , Return Address =
           RewardAcnt
            { network = Testnet
            , credential =
                (KeyCred
                   #"7fb37ce4801d7766f965794f675da01c9748b0fb0436f90fd80dfbdc") }
       , Action =
           HardForkInitiation
            { previous governance action id = ?-
            , protocol version = ProtVer{ major = 0, minor = 2 } }
       , Proposed In = EpochNo 1
       , Expires After = EpochNo 2 }]
sig = GovProcedures
       { VotingProcedures = Map{  }
       , ProposalProcedures =
           [ProposalProcedure
             { Deposit = (Coin 0)
             , ReturnAddr =
                 RewardAcnt
                  { network = Testnet
                  , credential =
                      (ScriptCred
                         #"48be9b8edd1615207e01587ac989d172ba0a7fbdf8b07eeeca8bfd0f") }
             , GovAction =
                 UpdateCommittee
                  { previous governance action id =
                      ?PrevGovActionId {unPrevGovActionId = GovActionId {gaidTxId = TxId {unTxId = SafeHash "c477d1e2047c84c102e7d3ca445
76aa18d85377960fd2ed0b5c86a18ed243ac6"}, gaidGovActionIx = GovActionIx 1}}
                  , membersToRemove = {}
                  , membersToAdd = Map{  }
                  , quorum = 4 % 5 }
             , Anchor =
                 Anchor
                  { url = Url {urlToText = "/"}
                  , datahash =
                      #"52477d6b1af2349b2aa56134f2055106d8936e61842fa8ef63e9b133475c9e21" } }] }
[ProposalsCantFollow (fromList [(GovActionId {gaidTxId = TxId {unTxId = SafeHash "e6ac71356c26c34cfb08093db61e333de74d155666bd53b34a401f
eabcd4b4e2"}, gaidGovActionIx = GovActionIx 2},(ProtVer {pvMajor = Version 0, pvMinor = 2},ProtVer {pvMajor = Version 0, pvMinor = 1}))]
)]
```

Now, this is a verbose counterexample but it boils down to `ProposalsCantFollow` with version `0.2` not being able to follow `0.1`.

Looking at this code:
```
-- | If the GovAction part of the GovActionState is a HardFork, test that
--   it's ProtVer is legal. If it is not then return the data to make one
--   entry in the Map inside PredicateFailure 'ProposalsCantFollow'
legalProtVer ::
  EraPParams era =>
  PParams era ->
  PrevGovActionIds era ->
  Proposals era ->
  GovActionState era ->
  Maybe (GovActionId (EraCrypto era), (ProtVer, ProtVer))
legalProtVer pp pgaids ps gas = do
  (newProtVer, prevProtVer) <- preceedingHardFork (gasAction gas) pp pgaids ps
  if pvCanFollow newProtVer (SJust prevProtVer)
    then Nothing
    else Just (gasId gas, (newProtVer, prevProtVer))
```
We see that `pvCanFollow` is called with `pvCanFollow new (Just old)` but if we look at `pvCanFollow` we see that the definition is:
```
pvCanFollow :: ProtVer -> StrictMaybe ProtVer -> Bool
pvCanFollow _ SNothing = True
pvCanFollow (ProtVer m n) (SJust (ProtVer m' n')) =
  (succVersion m, 0) == (Just m', n') || (m, n + 1) == (m', n')
```
Now, this boils down to the check `(succVersion 0, 0) == (Just 0, 1) || (0, 2 + 1) == (0, 1)`, which clearly isn't true.

lehins:
# Description

Fix wrong usage of `pvCanFollow` in Conway: fix #3901 and fix #3873
This problem happened because there was no documentation on `pvCanFollow` and names for bindings where totally useless. This proves my belief even further that ticks are not very useful for bindings.

Besides the bug fix this PR also includes:

* Make `pvCanFollow` match the spec and remove the `SNothing` wrapper
  around the newly proposed `ProtVer`
* Reduce duplication by adding `hasLegalProtVerUpdate`
* Add helper functions `failOnJust`, `failOnNonEmpty`, `failureOnJust`,
  `failureOnNonEmpty` to `small-steps`, which allow us to lazily get the
  offending value, without forcing the validation, when validation is
  turned off.
`cardano-ledger-alonzo-test-1.1.2.8` has not been released yet, so the
version bump in https://github.com/input-output-hk/cardano-ledger/pull/3876 was incorrect

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Gov.hs
+++ b/eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Gov.hs
@@ -440,7 +440,7 @@ legalProtVer ::
   Maybe (GovActionId (EraCrypto era), (ProtVer, ProtVer))
 legalProtVer pp pgaids ps gas = do
   (newProtVer, prevProtVer) <- preceedingHardFork (gasAction gas) pp pgaids ps
-  if pvCanFollow newProtVer (SJust prevProtVer)
+  if pvCanFollow prevProtVer newProtVer
     then Nothing
     else Just (gasId gas, (newProtVer, prevProtVer))
 
--- a/eras/shelley/impl/src/Cardano/Ledger/Shelley/PParams.hs
+++ b/eras/shelley/impl/src/Cardano/Ledger/Shelley/PParams.hs
@@ -27,6 +27,7 @@ module Cardano.Ledger.Shelley.PParams (
   emptyPPPUpdates,
   Update (..),
   pvCanFollow,
+  hasLegalProtVerUpdate,
 
   -- * JSON helpers
   shelleyCommonPParamsHKDPairs,
@@ -524,10 +525,26 @@ updatePParams :: EraPParams era => PParams era -> PParamsUpdate era -> PParams e
 updatePParams = applyPPUpdates
 {-# DEPRECATED updatePParams "Use applyPPUpdates instead" #-}
 
-pvCanFollow :: ProtVer -> StrictMaybe ProtVer -> Bool
-pvCanFollow _ SNothing = True
-pvCanFollow (ProtVer m n) (SJust (ProtVer m' n')) =
-  (succVersion m, 0) == (Just m', n') || (m, n + 1) == (m', n')
+-- | Check whether the new protocol version is a legitimate version bump with respect to the
+-- previous one.
+pvCanFollow ::
+  -- | Previous protocol version
+  ProtVer ->
+  -- | New protocol version
+  ProtVer ->
+  Bool
+pvCanFollow (ProtVer curMajor curMinor) (ProtVer newMajor newMinor) =
+  (succVersion curMajor, 0) == (Just newMajor, newMinor)
+    || (curMajor, curMinor + 1) == (newMajor, newMinor)
+
+-- | Check whether `PParamsUpdate` contains a valid `ProtVer` update. When a protocol version
+-- update is not included in `PParamsUpdate` it is considered a legal update.
+hasLegalProtVerUpdate ::
+  (ProtVerAtMost era 8, EraPParams era) => PParams era -> PParamsUpdate era -> Bool
+hasLegalProtVerUpdate pp ppu =
+  case ppu ^. ppuProtocolVersionL of
+    SNothing -> True
+    SJust newProtVer -> pvCanFollow (pp ^. ppProtocolVersionL) newProtVer
 
 upgradeUpdate ::
   forall era.
--- a/eras/alonzo/test-suite/cardano-ledger-alonzo-test.cabal
+++ b/eras/alonzo/test-suite/cardano-ledger-alonzo-test.cabal
@@ -1,6 +1,6 @@
 cabal-version: 3.0
 name:          cardano-ledger-alonzo-test
-version:       1.1.2.9
+version:       1.1.2.8
 license:       Apache-2.0
 maintainer:    operations@iohk.io
 author:        IOHK
--- a/eras/shelley/impl/CHANGELOG.md
+++ b/eras/shelley/impl/CHANGELOG.md
@@ -3,6 +3,8 @@
 ## 1.9.0.0
 
 * Change the type of `ppEMaxL`
+* Change the type of `pvCanFollow` to make sure new protocol version is not optional
+* Add `hasLegalProtVerUpdate`
 
 ### `testlib`
 
--- a/eras/shelley/impl/cardano-ledger-shelley.cabal
+++ b/eras/shelley/impl/cardano-ledger-shelley.cabal
@@ -124,7 +124,7 @@ library
         nothunks,
         quiet,
         set-algebra >=1.0,
-        small-steps >=1.0,
+        small-steps >=1.0.1,
         cardano-strict-containers,
         text,
         time,
--- a/eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Newpp.hs
+++ b/eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Newpp.hs
@@ -32,7 +32,7 @@ import Cardano.Ledger.Shelley.LedgerState (
 import Cardano.Ledger.Shelley.PParams (
   ProposedPPUpdates (ProposedPPUpdates),
   emptyPPPUpdates,
-  pvCanFollow,
+  hasLegalProtVerUpdate,
  )
 import Control.DeepSeq (NFData)
 import Control.State.Transition (
@@ -130,9 +130,7 @@ updatePpup ppupSt pp =
     & futureProposalsL .~ emptyPPPUpdates
   where
     ProposedPPUpdates newProposals = futureProposals ppupSt
-    goodPV ppu =
-      pvCanFollow (pp ^. ppProtocolVersionL) $ ppu ^. ppuProtocolVersionL
     ps =
-      if all goodPV newProposals
+      if all (hasLegalProtVerUpdate pp) newProposals
         then ProposedPPUpdates newProposals
         else emptyPPPUpdates
--- a/eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Ppup.hs
+++ b/eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Ppup.hs
@@ -26,7 +26,7 @@ import Cardano.Ledger.BaseTypes (
   Globals (stabilityWindow),
   ProtVer,
   ShelleyBase,
-  StrictMaybe (SJust),
+  StrictMaybe (..),
   epochInfoPure,
   invalidKey,
  )
@@ -44,7 +44,7 @@ import Cardano.Ledger.Shelley.Governance
 import Cardano.Ledger.Shelley.PParams (
   ProposedPPUpdates (ProposedPPUpdates),
   Update (..),
-  pvCanFollow,
+  hasLegalProtVerUpdate,
  )
 import Cardano.Ledger.Slot (
   Duration (Duration),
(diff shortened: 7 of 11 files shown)


eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Gov.hs
```diff
@@ -440,7 +440,7 @@ legalProtVer ::
   Maybe (GovActionId (EraCrypto era), (ProtVer, ProtVer))
 legalProtVer pp pgaids ps gas = do
   (newProtVer, prevProtVer) <- preceedingHardFork (gasAction gas) pp pgaids ps
-  if pvCanFollow newProtVer (SJust prevProtVer)
+  if pvCanFollow prevProtVer newProtVer
```
MaximilianAlgehed, line 443: Can I suggest we maybe flip the order of the arguments in `pvCanFollow`? Right now it reads really weirdly, "prevProtVer can follow newProtVer". That's a recipe for disaster..?

lehins: I do agree with you that it reads weirdly, but unfortunately we should not change the order now because that would be dangerous. Not only because this function has been around for a long time and we don't know if there are other users of it out there, but also because this is function is part of the Shelley spec. So we would have to amend the specification as well. 

All in all it is not worth it. Good documentation and clearer implementation should be enough for users to not create a disaster.

eras/shelley/impl/src/Cardano/Ledger/Shelley/PParams.hs
```diff
@@ -524,10 +525,26 @@ updatePParams :: EraPParams era => PParams era -> PParamsUpdate era -> PParams e
...
+  -- | Previous protocol version
+  ProtVer ->
+  -- | New protocol version
+  ProtVer ->
+  Bool
+pvCanFollow (ProtVer curMajor curMinor) (ProtVer newMajor newMinor) =
+  (succVersion curMajor, 0) == (Just newMajor, newMinor)
+    || (curMajor, curMinor + 1) == (newMajor, newMinor)
+
+-- | Check whether PParamsUpdate contains a valid `ProtVer` update. No protocol version
```
teodanciu: Just an amusing thing (not worth a rebase though) - maybe "Empty protocol version update" would be a little better than "No protocol version" -  At first pass I read it as in: there is no protocol version update that is considered a legal update :laughing: 
But of course it's very clear from the code

teodanciu (approved):
This is great, I love that the dangerous parts (which hold the risk of getting the arguments reversed) are contained in  the higher level `hasLegalProtVerUpdate` and `legalProtVer`

Looks good to me.

eras/shelley/impl/src/Cardano/Ledger/Shelley/PParams.hs
lehins: I had to rebase on master any ways, so I decided to fixup the haddock. I changed it to:

>  When a protocol version update is not included in `PParamsUpdate` it is considered a legal update.

Thank you for pointing out the ambiguity.

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Gov.hs
Soupstraw: How about we deprecate `pvCanFollow` and introduce `canBumpTo` which has the arguments the right way around? But it's also fine if we keep it as it is because of the spec.

Soupstraw: Or we can keep the function around for our implementation but not export it and instead export only `canBumpTo`.

lehins: Order of arguments is not important! As long as documentation clearly states what function does that is the most important part.

It's like arguing on the order of buffers in `memcpy`, it is all a matter of perspective.

MaximilianAlgehed: Argument order clearly matters, as evidenced by the bug..?

teodanciu: It matters that the arguments are passed in correctly, but the interpretation of their order is subjective. I think changing the order would cause problems than solve.

lehins: > Argument order clearly matters, as evidenced by the bug..?

@MaximilianAlgehed Lack of documentation is the cause of the bug, not the order of the arguments!

Which everyone should learn from: writing documentation for functions is important!

I am guilty of this as well, so if you ever see me forgetting haddock in my PRs, especially for exported functions please call me out on it.

Soupstraw: Most end-users go by the mantra that an hour of debugging can save minutes of reading the documentation :upside_down_face: 

I guess the thing is that `pvCanFollow` sounds like it can be written as an infix function, but then it does the opposite of how it reads. Of course it's not a good practice to rely on that behavior, but so many libraries have conditioned people to think like that so it feels dangerous to betray their expectations like that.

PR #3903, opened 2023-12-01
Add lenient decoder for `Addr`

### LINKED ISSUE #3898: cardano-ledger-core's `decodeAddr` / `deserialiseAddr` fail to deserialize mainnet pointer address
| package | version |
| --- | --- |
| cardano-ledger-core | 1.8.0.0 | 

> [!IMPORTANT]
>  That there's no changes on the impacted module in 1.9.0.0, so while I haven't tried with 1.9.0.0 I hardly believe it fixes the problem. 

It seems that since the internals of the [Cardano.Ledger.Address](https://github.com/input-output-hk/cardano-ledger/blob/master/libs/cardano-ledger-core/src/Cardano/Ledger/Address.hs) module was reworked to utilize state transformers, something broke regarding the parsing of pointer addresses. 

More specifically, trying to decode (the bech32-decoded representation of): `addr1gy5p8wv6sr8mgqjrwj7s75pft9y94ftwqey9vnlcqhew2xaumxqe2cdam3npgv60hqa` (or `412813b99a80cfb4024374bd0f502959485aa56e0648564ff805f2e51bbcd9819561bddc6614` in base16) using `decodeAddr` yields an error.

https://github.com/input-output-hk/cardano-ledger/blob/45c67892540d52aa6af9972c58bb20a125ceea09/libs/cardano-ledger-core/src/Cardano/Ledger/Address.hs#L142-L143

https://github.com/input-output-hk/cardano-ledger/blob/45c67892540d52aa6af9972c58bb20a125ceea09/libs/cardano-ledger-core/src/Cardano/Ledger/Address.hs#L565-L571

However, that address is an actual mainnet address found in one of the blocks of the Mary era. It can also be properly decoded using old code as well as alternative implementations like the one from [cardano-address](https://github.com/IntersectMBO/cardano-addresses/blob/0b66b8ad3703ab349414de58f3503843dfb22a1e/core/lib/Data/Word7.hs#L7) for de-serializing pointers. 

From a quick read through the ledger's code, it seems that the problem could come from how the new decoders _assume_ a certain length for the pointers in `decodeVariableLengthWord16` and `decodeVariableLengthWord32`. This leads to always decoding the same number of bytes whereas the encoding is meant to work with variable-length naturals and thus, may end up shorter than expected here. 

https://github.com/input-output-hk/cardano-ledger/blob/45c67892540d52aa6af9972c58bb20a125ceea09/libs/cardano-ledger-core/src/Cardano/Ledger/Address.hs#L815

lehins:
# Description

Fixes #3898

Besides addition of extra functions that do lenient address decoding this PR also deprecates `deserialiseAddr`, because:

* semantics of that function has silently changed back in Alonzo, which has a potential to affect some clinets
* it is used in cardano-api, hydra and db-sync
* it is redundant, since `decodeAddr` is a more general version of that function

Other changes in this PR:

* Improve `Addr` decoding benchmark
* Switch to strict `StateT`: tiny barely noticeable improvement in performance
* Remove modules that have been deprecated for a long time
* Deprecate `mkRwdAcnt` as useless

# Checklist

- [ ] Self-reviewed the diff


libs/cardano-ledger-core/bench/Addr.hs
```diff
@@ -7,41 +7,39 @@
...
+generateAddrAsBytestring qcGen count =
+  pure $ unGen (replicateM count (serialiseAddr <$> genAddr)) qcGen 30
   where
-    genAddr :: IO (Addr StandardCrypto)
-    genAddr = generate arbitrary
+    genAddr :: Gen (Addr StandardCrypto)
+    genAddr = arbitrary
 
 tryDecodeAddr :: [ByteString] -> ()
 tryDecodeAddr xs =
```
teodanciu, line 39: Would this benefit from `HasCallStack` in the signature ?

libs/cardano-ledger-core/bench/Addr.hs
```diff
@@ -19,7 +19,7 @@ import Test.QuickCheck.Random (QCGen, mkQCGen)
 main :: IO ()
 main = do
   defaultMain $
-    map (\c -> env (generateAddrAsBytestring qcGen c) decodeAddrBench) (map (* 500) [1 .. 2])
```
teodanciu: I can't figure out how this change impacts the result -  I will ask you on Wednesday!

libs/cardano-ledger-api/src/Cardano/Ledger/Api/Tx/Address.hs
```diff
@@ -1,13 +1,27 @@
...
+  -- out for `decodeAddrLenient`.
   decodeAddr,
   decodeAddrEither,
   decodeAddrShort,
   decodeAddrShortEither,
-  getNetwork,
+
+  -- ** Lenient decoders
+
+  -- | These a lenient decoders that do not fail for addresses with known bugs
```
teodanciu: Maybe: `These *are* lenient decoders .. `   or `These lenient decoders do not fail ..`

teodanciu (approved):
Looks good to me

libs/cardano-ledger-core/bench/Addr.hs
lehins: No, it is only used in one place. So error location is unambiguous. Also it should be impossible.

`HasCallStack` is very useful in testing and in the main logic of the program wherever failure is possible. Benchmarks aren't so much, since you don't really expect those to fail.

libs/cardano-ledger-core/bench/Addr.hs
lehins: It does not affect it in any way. :smile: It is just clearer on what is going on.

libs/cardano-ledger-api/src/Cardano/Ledger/Api/Tx/Address.hs
lehins: Done. Thank you.

eras/shelley/test-suite/test/Test/Cardano/Ledger/Shelley/Serialisation/Golden/Address.hs
```diff
@@ -193,10 +193,6 @@ goldenTests_ShelleyCrypto =
               }
         )
         "82d818582183581c4bf3c2ee56bfef278d65f7388c46efa12a1069698e474f77adf0cf6aa0001ab4aad9a5"
-    , T.testCase "fail on extraneous bytes" $
```
Soupstraw, line 196: Why is this test removed?

Soupstraw: Ok, I see the test is moved to core tests :+1:

Soupstraw (approved):
LGTM

PR #3907, opened 2023-12-04
cardano-ledger-conformance: Remove CHANGELOG.md from cabal file

erikd:
Currently empty.

# Description

Without this file, this repository cannot be used as a `source-package-repository` because:
```
> cabal build all
HEAD is now at a62edfae6 CI: Add ghc-9.8.1 to the build matrix
Error: cabal: sdist of cardano-ledger-conformance-0.1.0.0: filepath wildcard
'CHANGELOG.md' does not match any files.
```

# Checklist

- [ ] Self-reviewed the diff


lehins (changes requested):
We should instead remove `CHANGELOG.md` from the cabal file, because this package will never be released, thus changelog makes little sense with absence of releasing and versioning

lehins (approved):
Thank you!

PR #3909, opened 2023-12-04
Make `impAnn` a bit more useful, by making the logs scoped by `impAnn`

lehins:
# Description

Reviewing #3852 made me realize that sometimes we want to display lengthy Imp logs only when there is some failure in the current scope, but not otherwise.

So, imagine we have some very lengthy log entry (in this example: `"Found a problem with x and y"`), but we want to show this only when `x` and `y` are not equal, but for all ther failures in the test this log entry would not be relevant and should not be displayed. Then instead of writing this:

```haskell
when (x /= y) $ do
  logEntry "Found a problem with x and y"
  expectationFailure $ showTreeDiff x y
```

we can now write this:
```haskell
impAnn "Ensuring x and y validity" $ do
  logEntry "Found a problem with x and y"
  x `shouldBeExpr` y
```

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/eras/shelley/impl/testlib/Test/Cardano/Ledger/Shelley/ImpTest.hs
+++ b/eras/shelley/impl/testlib/Test/Cardano/Ledger/Shelley/ImpTest.hs
@@ -599,14 +599,18 @@ instance Show ImpException where
       <> displayException e
 instance Exception ImpException
 
--- | Annotation for when failure happens
+-- | Annotation for when failure happens. All the logging done within annotation will be
+-- discarded if there no failures within the annotation.
 impAnn :: NFData a => String -> ImpTestM era a -> ImpTestM era a
 impAnn msg m = do
-  catchAnyDeep m $ \exc ->
+  logs <- use impLogL
+  res <- catchAnyDeep m $ \exc ->
     throwIO $
       case fromException exc of
         Just (ImpException ann origExc) -> ImpException (msg : ann) origExc
         Nothing -> ImpException [msg] exc
+  impLogL .= logs
+  pure res
 
 -- | Adds a string to the log, which is only shown if the test fails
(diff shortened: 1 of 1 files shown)


eras/shelley/impl/testlib/Test/Cardano/Ledger/Shelley/ImpTest.hs
```diff
@@ -599,14 +599,18 @@ instance Show ImpException where
...
 impAnn :: NFData a => String -> ImpTestM era a -> ImpTestM era a
 impAnn msg m = do
-  catchAnyDeep m $ \exc ->
+  logs <- use impLogL
+  res <- catchAnyDeep m $ \exc ->
     throwIO $
       case fromException exc of
         Just (ImpException ann origExc) -> ImpException (msg : ann) origExc
         Nothing -> ImpException [msg] exc
+  impLogL .= logs
```
teodanciu, line 612: But these scopes couldn't be related to each other?  So if we have a failure, there is no chance that the logs in this other scope could help track down the failure?

teodanciu (approved):
Just a question, approving (since I'm almost certain of the answer, but just to check that I understand). 
Looks good to me

eras/shelley/impl/testlib/Test/Cardano/Ledger/Shelley/ImpTest.hs
Soupstraw: I think the idea is that each scope is hand-crafted to help debug a specific problem, so if there is some information you would like to have in multiple scopes, you'll just turn those loggers into a function and run it again in each scope where that information is relevant.

Soupstraw (approved):
Nice!

PR #3912, opened 2023-12-05
Removed call to tail, and the 'watchPulser' test

TimSheard:
The following test causes a failure on Master

TASTY_PATTERN='/Watch pulser on traces of length 120./' cabal test --test-options "--quickcheck-replay=433821"

to make a step test, where we test a property on every (state{n}, state{n+1}) pair in trace we used the following pattern
pairs = zip states (tail states)
But in very rare occasions, when the number of states is 0 or 1,  tail raises the empty List error.
We can easily fix this by pattern matching, and return true when that is the case.

We also removed the '/Watch pulser on traces of length 120./' test, which uses trace and prints too much stuff.

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/Trace/TraceMonad.hs
+++ b/libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/Trace/TraceMonad.hs
@@ -444,7 +444,9 @@ mockChainProp proof n genTxAndUpdateEnv propf = do
 --   conjoin [p pstate0 sig1 state1, p state0 sig2 state2, p state0 sig3 state3]
 -- test that 'p' holds between state0 and stateN for all N
 stepProp :: (MockChainState era -> MockBlock era -> MockChainState era -> Property) -> (Trace (MOCKCHAIN era) -> Property)
-stepProp p tr = conjoin (map (\(sig, st) -> p state0 sig st) (tail (zip sigs states)))
+stepProp p tr = case (zip sigs states) of
+  [] -> property True
+  (_ : more) -> conjoin (map (\(sig, st) -> p state0 sig st) more)
   where
     state0 = tr ^. traceInitState
     states = traceStates OldestFirst tr
@@ -454,7 +456,9 @@ stepProp p tr = conjoin (map (\(sig, st) -> p state0 sig st) (tail (zip sigs sta
 --   conjoin [p pstate0 sig1 state1, p state1 sig2 state2, p state2 sig3 state3]
 --   test that 'p' holds bteween (state0 stateN sig(N+1) state(N+1) for all N
 deltaProp :: (MockChainState era -> MockBlock era -> MockChainState era -> Property) -> (Trace (MOCKCHAIN era) -> Property)
-deltaProp p tr = conjoin (map (\(statei, sig, statej) -> p statei sig statej) (zip3 states sigs (tail states)))
+deltaProp p tr = case states of
+  [] -> property True
+  (_ : more) -> conjoin (map (\(statei, sig, statej) -> p statei sig statej) (zip3 states sigs more))
   where
     states = traceStates OldestFirst tr
     sigs = traceSignals OldestFirst tr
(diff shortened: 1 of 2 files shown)


libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/Trace/TraceMonad.hs
```diff
@@ -454,7 +456,9 @@ stepProp p tr = conjoin (map (\(sig, st) -> p state0 sig st) (tail (zip sigs sta
 --   conjoin [p pstate0 sig1 state1, p state1 sig2 state2, p state2 sig3 state3]
 --   test that 'p' holds bteween (state0 stateN sig(N+1) state(N+1) for all N
 deltaProp :: (MockChainState era -> MockBlock era -> MockChainState era -> Property) -> (Trace (MOCKCHAIN era) -> Property)
-deltaProp p tr = conjoin (map (\(statei, sig, statej) -> p statei sig statej) (zip3 states sigs (tail states)))
+deltaProp p tr = case states of
+  [] -> property True
+  (_ : more) -> conjoin (map (\(statei, sig, statej) -> p statei sig statej) (zip3 states sigs more))
```
lehins: Less noise and simpler
```suggestion
  _ : more -> conjoin (zipWith3 p states sigs more)
```

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/Trace/TraceMonad.hs
```diff
@@ -454,7 +456,9 @@ stepProp p tr = conjoin (map (\(sig, st) -> p state0 sig st) (tail (zip sigs sta
 --   conjoin [p pstate0 sig1 state1, p state1 sig2 state2, p state2 sig3 state3]
 --   test that 'p' holds bteween (state0 stateN sig(N+1) state(N+1) for all N
 deltaProp :: (MockChainState era -> MockBlock era -> MockChainState era -> Property) -> (Trace (MOCKCHAIN era) -> Property)
```
lehins:
```suggestion
deltaProp :: (MockChainState era -> MockBlock era -> MockChainState era -> Property) -> Trace (MOCKCHAIN era) -> Property
```

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/Trace/TraceMonad.hs
```diff
@@ -444,7 +444,9 @@ mockChainProp proof n genTxAndUpdateEnv propf = do
 --   conjoin [p pstate0 sig1 state1, p state0 sig2 state2, p state0 sig3 state3]
 -- test that 'p' holds between state0 and stateN for all N
 stepProp :: (MockChainState era -> MockBlock era -> MockChainState era -> Property) -> (Trace (MOCKCHAIN era) -> Property)
-stepProp p tr = conjoin (map (\(sig, st) -> p state0 sig st) (tail (zip sigs states)))
+stepProp p tr = case (zip sigs states) of
+  [] -> property True
+  (_ : more) -> conjoin (map (\(sig, st) -> p state0 sig st) more)
```
lehins: Besides the stylistic suggestion I have a question: Is there a guarantee that there will be as many signals as there are states? Otherwise `zip`/`zipWith` will discard the extras, which is ok if it expected.


```suggestion
stepProp p tr = case zipWith (p state0) sigs states of
  [] -> property True
  (_ : more) -> conjoin more
```

lehins (approved):
Just a little linting. Otherwise its great.

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/Trace/TraceMonad.hs
TimSheard: There is always one more state than signals. But the number of these depends on lots of factors.

PR #3913, opened 2023-12-06
Add certs to conformance testing

Soupstraw:
# Description

This adds support for certificates in conformance testing. Currently we can't do `RegKey` and `UnRegKey`, since these are not in the spec yet.

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/libs/cardano-ledger-conformance/src/Test/Cardano/Ledger/Conformance.hs
+++ b/libs/cardano-ledger-conformance/src/Test/Cardano/Ledger/Conformance.hs
@@ -22,7 +22,7 @@ module Test.Cardano.Ledger.Conformance (
 
 import Cardano.Crypto.DSIGN (DSIGNAlgorithm (..), SignedDSIGN (..))
 import Cardano.Crypto.Hash (Hash, hashToBytes)
-import Cardano.Ledger.Address (Addr (..), serialiseAddr)
+import Cardano.Ledger.Address (Addr (..), RewardAcnt (..), serialiseAddr)
 import Cardano.Ledger.Allegra.Scripts (ValidityInterval (..))
 import Cardano.Ledger.Alonzo (AlonzoScript, AlonzoTxAuxData)
 import Cardano.Ledger.Alonzo.PParams (OrdExUnits (OrdExUnits))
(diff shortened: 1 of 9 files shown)


libs/cardano-ledger-conformance/src/Test/Cardano/Ledger/Conformance.hs
```diff
@@ -463,20 +478,103 @@ instance
...
+      <*> toSpecRep p
+  toSpecRep (ConwayTxCertPool (RetirePool kh e)) =
+    Agda.RetirePool
+      <$> toSpecRep (KeyHashObj kh)
+      <*> toSpecRep e
+  toSpecRep (ConwayTxCertGov (ConwayRegDRep c d _)) =
+    Agda.RegDRep
+      <$> toSpecRep c
+      <*> toSpecRep d
+      <*> pure () -- TODO Are the anchors supposed to be optional?
```
lehins, line 551: Yes, they must be optional [according to the spec](https://github.com/cardano-foundation/CIPs/tree/master/CIP-1694#drep-registration-certificates), but `()` isn't really an Anchor, so that's why it is probably not important for the spec

> an optional anchor

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/Preds/Certs.hs
```diff
@@ -241,7 +244,7 @@ certsPreds UnivSize {usMinCerts, usMaxCerts} p = case whichTxCert p of
           ( 1
           , sRegPool ^$ poolParams
           ,
-            [ -- Pick a random PoolParams, except constrain the fields poolId, poolRewAcnt, poolOwners, and pooMetadaa
+            [ -- Pick a random PoolParams, except constrain the fields poolId, poolRewAcnt, poolOwners, and poolMetadata
```
lehins, line 247: Previous name was funnier :smile:

lehins (approved):
Looks great. Thanks!

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/Preds/Certs.hs
Soupstraw: I could rename the field instead :thinking:

lehins: :rofl:

PR #3915, opened 2023-12-06
 Utilize `Inject` type class for `Val`'s `inject`

lehins:
# Description

Utilize `Inject` type class for `Val`'s `inject`:

There were two separate functions with the same name `inject` in the codebase, which is never good. That beeing said they were indeed very similar. The one from `Inject` type class was just more general the then one from the `Val` type class. It only makes sense to utilize the more general one in both places. Because of this `Inject` type class was moved to `BaseTypes` module instead, since it is a more general purpose type class now.

This PR also removes some of the unused and redundant stuff

It also utilizes `inject` for GOV test

# Checklist

- [x] Self-reviewed the diff


libs/cardano-ledger-core/src/Cardano/Ledger/BaseTypes.hs
```diff
@@ -830,3 +836,8 @@ newtype EpochInterval = EpochInterval
 -- | Add a EpochInterval (a positive change) to an EpochNo to get a new EpochNo
 addEpochInterval :: EpochNo -> EpochInterval -> EpochNo
 addEpochInterval (EpochNo n) (EpochInterval m) = EpochNo (n + fromIntegral m)
+
+class Inject t s where
+  inject :: t -> s
+  default inject :: t ~ s => t -> s
+  inject = id
```
TimSheard, line 843: I have never seen this kind of default instance. It is quite clever. 
I take it, one just says
`instance Inject T T where`
and one gets the default instance, but if one says
`instance Inject S T where`
one gets an error.

libs/cardano-ledger-core/src/Cardano/Ledger/Rules/ValidationMode.hs
```diff
@@ -106,32 +100,12 @@ applySTSNonStatic ::
...
-    . fromFailure []
-    . first (mapMaybe toPredicateFailureMaybe . NE.toList)
-
 -- ===========================================================
 
-class Inject t s where
-  inject :: t -> s
-
-class InjectMaybe t s where
-  injectMaybe :: t -> Maybe s
```
TimSheard, line 126: Are we missing some important capability by removing this. If I remember, this point of this was to handle Predicate failures that became no longer possible when we transitioned from one Era to another,  the new Era no-longer had  that kind of Predicate failure?

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/ApplyTx.hs
```diff
@@ -30,7 +30,7 @@ import Cardano.Ledger.Shelley.TxBody (
 import Cardano.Ledger.Shelley.TxCert (ShelleyDelegCert (..), ShelleyTxCert (..))
 import Cardano.Ledger.TxIn (TxId (..), TxIn (..))
 import Cardano.Ledger.UTxO (UTxO (..))
-import Cardano.Ledger.Val (Val (inject), (<+>), (<->))
+import Cardano.Ledger.Val (Val ((<+>), (<->)), inject)
```
TimSheard, line 33: Does Val re-export inject? Is that a good idea?

TimSheard (approved):
Looks good to me.

libs/cardano-ledger-core/src/Cardano/Ledger/BaseTypes.hs
lehins: That is correct

libs/cardano-ledger-core/src/Cardano/Ledger/Rules/ValidationMode.hs
lehins: Nah, it was really dead code that wasn't used for anything. I think it was used as initial attempt to achieve something, but it didn't go anywhere.

Important part is that I wouldn't be able to remove it if we needed it.

In any case, if we ever need, we can always use 
```haskell
instance Inject MyType (Maybe MyOtherType) where
  inject ...
```

But, it doesn't look like we'll ever need it.

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/ApplyTx.hs
lehins: Yes, `Val` module does re-export `inject`. It is a good idea to reduce breakage downstream. Also, `Inject` is now a required superclass for `Val`, so it is useful to re-export it for convenience as well.

PR #3921, opened 2023-12-08
Fix committee validation in Ratify

### LINKED ISSUE #3889: Current validation of UpdateCommitte action is incorrect
On this line we use current CommitteMembers for validation:

https://github.com/input-output-hk/cardano-ledger/blob/afedb7d519761ccdd9c013444aa4b3e0bf0e68ef/eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Ratify.hs#L312

Which is incorrect. We must be using the new committee members that are being proposed.

Here is the corresponding spec part of the code:

https://github.com/input-output-hk/formal-ledger-specifications/blob/eafd83a7931cd6af07212253a183115e740f6c1e/src/Ledger/GovernanceActions.lagda#L308-L311

teodanciu:
# Description

Closes https://github.com/input-output-hk/cardano-ledger/issues/3889

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Ratify.hs
+++ b/eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Ratify.hs
@@ -63,7 +63,7 @@ import Cardano.Ledger.Conway.Governance (
   votingDRepThreshold,
   votingStakePoolThreshold,
  )
-import Cardano.Ledger.Conway.Governance.Procedures (committeeMembersL, gasActionL, gasChildrenL)
+import Cardano.Ledger.Conway.Governance.Procedures (gasActionL, gasChildrenL)
 import Cardano.Ledger.Conway.PParams (
   ConwayEraPParams,
   ppCommitteeMaxTermLengthL,
@@ -306,8 +306,7 @@ ratifyTransition = do
     ( env@RatifyEnv {reCurrentEpoch}
       , st@( RatifyState
               rsEnactState@EnactState
-                { ensCommittee
-                , ensCurPParams
+                { ensCurPParams
                 , ensTreasury
                 , ensPrevGovActionIds
                 , ensPrevGovActionIdsChildren
@@ -323,7 +322,7 @@ ratifyTransition = do
   case reorderedActions of
     gas@GovActionState {gasId, gasAction, gasExpiresAfter} :<| sigs -> do
       if prevActionAsExpected gasAction ensPrevGovActionIds
-        && validCommitteeTerm ensCommittee ensCurPParams reCurrentEpoch
+        && validCommitteeTerm gasAction ensCurPParams reCurrentEpoch
         && not rsDelayed
         && withdrawalCanWithdraw gasAction ensTreasury
         && committeeAccepted env st gas
@@ -395,14 +394,17 @@ updatePrevGovActionIdsChildren gas pgac =
 
 validCommitteeTerm ::
   ConwayEraPParams era =>
-  StrictMaybe (Committee era) ->
+  GovAction era ->
   PParams era ->
   EpochNo ->
   Bool
-validCommitteeTerm committee pp currentEpoch =
-  let maxCommitteeTerm = pp ^. ppCommitteeMaxTermLengthL
-      members = foldMap' (^. committeeMembersL) committee
-   in all (<= currentEpoch + maxCommitteeTerm) members
+validCommitteeTerm govAction pp currentEpoch =
+  case govAction of
+    UpdateCommittee _ _ newMembers _ -> withinMaxTermLength newMembers
+    _ -> True
+  where
+    committeeMaxTermLength = pp ^. ppCommitteeMaxTermLengthL
+    withinMaxTermLength = all (<= currentEpoch + committeeMaxTermLength)
 
 instance EraGov era => Embed (ConwayENACT era) (ConwayRATIFY era) where
   wrapFailed = absurd
--- a/eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/GovSpec.hs
+++ b/eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/GovSpec.hs
@@ -11,8 +11,9 @@ module Test.Cardano.Ledger.Conway.Imp.GovSpec where
 
 import Cardano.Ledger.BaseTypes
 import Cardano.Ledger.CertState (vsNumDormantEpochsL)
+import Cardano.Ledger.Coin (Coin (Coin))
 import Cardano.Ledger.Conway.Governance
-import Cardano.Ledger.Conway.PParams (ppDRepActivityL, ppGovActionLifetimeL)
+import Cardano.Ledger.Conway.PParams
 import Cardano.Ledger.Conway.Rules (ConwayGovPredFailure (..))
 import Cardano.Ledger.Core
 import Cardano.Ledger.Credential (Credential (KeyHashObj))
@@ -22,14 +23,17 @@ import Cardano.Ledger.Keys (
   KeyRole (..),
  )
 import Cardano.Ledger.Shelley.LedgerState
+import Control.Monad (replicateM_)
 import Control.State.Transition.Extended (PredicateFailure)
 import Data.Default.Class (Default (..))
+import Data.Foldable (Foldable (..), traverse_)
 import qualified Data.Map.Strict as Map
 import Data.Maybe
 import qualified Data.Sequence.Strict as SSeq
 import qualified Data.Set as Set
 import Lens.Micro
 import Test.Cardano.Ledger.Conway.ImpTest
+import Test.Cardano.Ledger.Core.Rational (IsRatio (..))
 import Test.Cardano.Ledger.Imp.Common
 
 spec ::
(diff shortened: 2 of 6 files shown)


eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Ratify.hs
```diff
@@ -395,14 +394,17 @@ updatePrevGovActionIdsChildren gas pgac =
 
 validCommitteeTerm ::
   ConwayEraPParams era =>
-  StrictMaybe (Committee era) ->
   PParams era ->
   EpochNo ->
+  GovAction era ->
```
teodanciu: Not sure whether to mention the change of the signature in CHANGELOG, since it doesn't appear there to begin with (it's exported for testing). I suppose people could still use it, though.
If that's the case, maybe it's not worth changing the order of these parameters? I only did it so I can use the lambda case, but can't decide what's worth doing or not!

lehins: `prevActionAsExpected`, `withdrawalCanWithdraw` and `validCommitteTerm` all having `gasAction` as the first argument looks nice and consistent. 

I don't think lambda case is worth it :slightly_smiling_face: 

That being said, it will still change the type of the function, so mentioning it in the changelog is a good practice, but it will not be the end of the world if you don't do it.

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Ratify.hs
```diff
@@ -395,14 +394,17 @@ updatePrevGovActionIdsChildren gas pgac =
...
-  let maxCommitteeTerm = pp ^. ppCommitteeMaxTermLengthL
-      members = foldMap' (^. committeeMembersL) committee
-   in all (<= currentEpoch + maxCommitteeTerm) members
+validCommitteeTerm govAction pp currentEpoch =
+  case govAction of
+    UpdateCommittee _ _ newMembers _ -> valid newMembers
+    _ -> True
+  where
+    maxCommitteeTerm = pp ^. ppCommitteeMaxTermLengthL
+    valid = all (<= currentEpoch + maxCommitteeTerm)
```
lehins: `valid` is a little bit too general, maybe it could be a good idea to be more specific with the name?
 
```suggestion
    UpdateCommittee _ _ newMembers _ -> withinMaxTermLength newMembers
    _ -> True
  where
    committeeMaxTermLength = pp ^. ppCommitteeMaxTermLengthL
    withinMaxTermLength = all (<= currentEpoch + committeeMaxTermLength)
```

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/ImpTest.hs
```diff
@@ -622,10 +625,14 @@ logRatificationChecks gaId = do
     unlines
       [ "----- RATIFICATION CHECKS -----"
       , "prevActionAsExpected:\t" <> show (prevActionAsExpected gasAction ensPrevGovActionIds)
-      , "validCommitteeTerm:\t" <> show (validCommitteeTerm ensCommittee ensCurPParams currentEpoch)
+      , "validCommitteeTerm:\t" <> show (validCommitteeTerm gasAction ensCurPParams currentEpoch)
       , "notDelayed:\t\t??"
       , "withdrawalCanWithdraw:\t" <> show (withdrawalCanWithdraw gasAction curTreasury)
       , "committeeAccepted:\t" <> show (committeeAccepted ratEnv ratSt gas)
+      , "committeeAccepted:\t\t"
+          <> show (committeeAccepted ratEnv ratSt gas)
```
lehins: Duplicate information
```suggestion
      , "committeeAccepted:\t\t"
          <> show (committeeAccepted ratEnv ratSt gas)
```

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/ImpTest.hs
```diff
@@ -622,10 +625,14 @@ logRatificationChecks gaId = do
...
+          <> " [To Pass, at least: "
+          <> show (votingCommitteeThreshold ratSt gasAction)
```
lehins, line 638: Just like with SPOs and DReps it will be useful to add the actual accepted ratio, not just the threshold:
```suggestion
          <> show (committeeAcceptedRatio ratEnv ratSt gas)
          <> " >= "
          <> show (votingCommitteeThreshold ratSt gasAction)
          <> "]"
```

Note, I am suggesting to change the type signatures (thus implementation) of all three functions to be consistent:

* `committeeAcceptedRatio :: RatifyEnv era -> RatifyState era -> GovActionState era -> Rational`
* `dRepAcceptedRatio :: RatifyEnv era -> GovActionState era -> Rational`
* and `spoAcceptedRatio :: RatifyEnv era -> GovActionState era -> Rational`

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/ImpTest.hs
```diff
@@ -59,15 +62,7 @@ import Cardano.Ledger.BaseTypes (
...
-import Cardano.Ledger.Conway.Core (
-  Era (..),
-  EraIndependentTxBody,
-  EraTx (..),
-  EraTxBody (..),
-  EraTxOut (..),
-  PParams,
-  PParamsHKD,
- )
+import Cardano.Ledger.Conway.Core (DRepVotingThresholds (..), Era (..), EraIndependentTxBody, EraTx (..), EraTxBody (..), EraTxOut (..), PParams, PParamsHKD)
```
lehins: Line is too long. Also it is a Core module, no need to have explicit imports:
```suggestion
import Cardano.Ledger.Conway.Core
```

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/GovSpec.hs
```diff
@@ -272,7 +275,94 @@ spec =
...
+          curConstitution <- getsNES $ newEpochStateGovStateL . constitutionGovStateL
+          curConstitution `shouldBe` constitution
+          pure govIdConst1
+
+        -- after enough epochs pass, the expiration of the new members becomes acceptable
+        -- and the new committee is enacted
+        impAnn "New committee is enacted" $ do
+          currentEpoch <- getsNES nesELL
+          let (EpochNo delta) = exceedingExpiry - maxTermLength - currentEpoch
+          replicateM_ (fromIntegral delta) passEpoch
```
lehins: EpochNo is a Word64, so this calculation can easily overflow resulting in an "infinite" call in `replicateM_`. Not something you want to run into when debugging tests.
```suggestion
          let delta = fromIntegral exceedingExpiry - fromIntegral (maxTermLength + currentEpoch)
          replicateM_ delta passEpoch
```

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/GovSpec.hs
```diff
@@ -272,7 +275,94 @@ spec =
...
+            let initialMembers = [(c1, maxExpiry), (c2, maxExpiry)]
+            govIdCom1 <-
+              electCommittee
+                SNothing
+                drep
+                Set.empty
+                initialMembers
+            passEpoch >> passEpoch
+            expectMembers [c1, c2]
+            pure ([c1, c2], govIdCom1)
```
lehins: Improve resilience to changes of `initialMembers`
```suggestion
            expectMembers $ Map.keysSet initialMembers
            pure (Map.keysSet initialMembers, govIdCom1)
```

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/GovSpec.hs
```diff
@@ -272,7 +275,94 @@ spec =
...
+            _ <-
+              electCommittee
+                (SJust govIdCom1)
+                drep
+                Set.empty
+                membersExceedingExpiry
+            passEpoch >> passEpoch
+            -- the new committee has not been enacted
+            expectMembers initialMembers
+            pure ([c3, c4], exceedingExpiry)
```
lehins: Ditto
```suggestion
            pure (Map.keysSet initialMembers, exceedingExpiry)
```

lehins (approved):
Looks great. Although I'd like to suggest few minor improvements.

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/ImpTest.hs
teodanciu: Eek! Thanks for catching

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/ImpTest.hs
teodanciu: Funnily, I'm pretty sure this is why I duplicated the line above, because I intended to achieve exactly this. But then something went wrong... Let me try again. Thanks!

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Ratify.hs
teodanciu: Totally!  Initially it was even worse, I had called it "ok"  :laughing:  But definitely worth investing in good names, thanks for the suggestions.

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/GovSpec.hs
teodanciu: Had to apply `unEpochNo` before `fromIntegral`, otherwise I would get  `Could not deduce (Integral EpochNo)
        arising from a use of ‘fromIntegral’`, which make sense i suppose.

PR #3924, opened 2023-12-12
Split type classes out of eras Core module

lehins:
# Description

The idea of defining all core classes in a single module have proved to be a suboptimal and not always possible. Quite often it resulted in cyclic module dependency issues, which usually could be worked around, but not always.

Major goal of this PR is to move class definitions to their respective modules (eg, `BabbageEraTxOut` into `Cardano.Ledger.Babbage.TxOut`), while still re-exporting all of the core classes from the core modules. In other words dependency direction is inverted when it comes to the Core modules.

Along with this restructure, some of the re-exportations have been cleaned up and a bunch of deprecated functions have been removed. All of which have been noted in the CHANGELOGs

* `cardano-ledger-core`:
  * Move `Cardano.Ledger.Keys` to hidden module `Cardano.Ledger.Keys.Internal` and re-export all three modules from `Cardano.Ledger.Keys`
    * `Cardano.Ledger.Keys.Internal`
    * `Cardano.Ledger.Keys.Bootstrap`
    * and `Cardano.Ledger.Keys.WitsVKey`
* `cardano-ledger-shelley/marry/allegra/alonzo/babbage/conway`:
  * Move `[Era]Era[Type]`: from `Cardano.Ledger.[Era].Core` to `Cardano.Ledger.[Era].[Type]`. For example move `ShelleyEraTxBody`: from `Cardano.Ledger.Shelley.Core` to `Cardano.Ledger.Shelley.TxBody`
  * Re-export all core type classes from `Cardano.Ledger.[Era].Core`
  * Stop exporting from `Cardano.Ledger.[Era].Core` internal PParams hkd* lenses and type families. Only export dedicated lenses for dealing with `PParams` and `PParamsUpdate`, not the `PParamsHKD`
  * Stop exporting types that are not relevant for the module (eg. exporting `PoolParams` or `ShelleyEraTxOut` from `Cardano.Ledger.Shelley.TxBody`)

In other words, there are no semantic changes in this PR, just reorganization. One small exception is the `ToJSON` instances for `PoolVotingThresholds` and `DRepVotingThreshold` where implemented manually instead of deriving using `DeriveAnyClass`. 

# Checklist

- [x] Self-reviewed the diff


libs/cardano-ledger-core/src/Cardano/Ledger/Keys/Internal.hs
```diff
@@ -261,6 +263,9 @@ deriving newtype instance
   Crypto c =>
   FromJSON (KeyHash disc c)
 
+instance Crypto b => Default (KeyHash a b) where
+  def = KeyHash def
```
teodanciu, line 267: Worth adding the new Default instance to Changelog?

teodanciu: Nermind, I see now it was orphan,.

eras/shelley/impl/src/Cardano/Ledger/Shelley/API/Types.hs
```diff
@@ -26,6 +26,11 @@ import Cardano.Ledger.BaseTypes as X (
 
 -- TODO deprecate these?
 
+-- deprecated
+
+-- deprecated
```
teodanciu: These look like leftovers? As well as the TODO [a few lines above ](eras/shelley/impl/src/Cardano/Ledger/Shelley/API/Types.hs)

lehins: Somehow fourmolu moved them here from lines below where `KeyPair` and `KeyPairs` are imported.

Looks like a bug in fourmolu. I'll remove them if there will be other changes I need to make to this PR.

teodanciu (approved):
Looks good to me.  And nice to have all these deprecated stuff removed.

eras/shelley/impl/src/Cardano/Ledger/Shelley/API/Types.hs
lehins: Removed

PR #3929, opened 2023-12-14
Improve comment on `unit_interval` in cddl

lehins:
# Description

* Remove unused `rational`
* Add a better comment about `unit_interval`

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/eras/conway/impl/cddl-files/extra.cddl
+++ b/eras/conway/impl/cddl-files/extra.cddl
@@ -11,9 +11,20 @@ nonempty_oset<a> = #6.258([+ a]) / [+ a]
 positive_int = 1 .. 18446744073709551615
 
 unit_interval = #6.30([1, 2])
-  ; real unit_interval is: #6.30([uint, uint])
-  ; but this produces numbers outside the unit interval
-  ; and can also produce a zero in the denominator
+  ; unit_interval = #6.30([uint, uint])
+  ;
+  ; Comment above depicts the actual definition for `unit_interval`.
+  ;
+  ; Unit interval is a number in the range between 0 and 1, which
+  ; means there are two extra constraints:
+  ; * numerator <= denominator
+  ; * denominator > 0
+  ;
+  ; Relation between numerator and denominator cannot be expressed in CDDL, which
+  ; poses a problem for testing. We need to be able to generate random valid data
+  ; for testing implementation of our encoders/decoders. Which means we cannot use
+  ; the actual definition here and we hard code the value to 1/2
+
 
 nonnegative_interval = #6.30([uint, positive_int])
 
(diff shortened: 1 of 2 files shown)


eras/conway/impl/cddl-files/extra.cddl
```diff
@@ -11,9 +11,20 @@ nonempty_oset<a> = #6.258([+ a]) / [+ a]
 positive_int = 1 .. 18446744073709551615
 
 unit_interval = #6.30([1, 2])
-  ; real unit_interval is: #6.30([uint, uint])
-  ; but this produces numbers outside the unit interval
-  ; and can also produce a zero in the denominator
+  ; unit_interval = #6.30([uint, uint])
+  ;
+  ; Comment above depicts the acual definition for `unit_interval`.
```
teodanciu: `actual`

eras/conway/impl/cddl-files/extra.cddl
```diff
@@ -11,11 +11,20 @@ nonempty_oset<a> = #6.258([+ a]) / [+ a]
 positive_int = 1 .. 18446744073709551615
 
 unit_interval = #6.30([1, 2])
```
Soupstraw, line 13: I'm thinking that it might be a good idea to have `#6.30([1, 1])` and  `#6.30([0, 1])` also as possibilities, that way we can test more edge cases.

Soupstraw (approved):
LGTM

teodanciu (approved):
Looks good to me

eras/conway/impl/cddl-files/extra.cddl
lehins: It will be better to create a PR that splits out the concerns of testing and specifying.

One that will be used for testing will contain a whole bunch of random values, while the one o that used for specifying will have the correct types.

We can do it at some later point. I only created this PR because we were asked to clarify the comment on the `unit_interval`

eras/conway/impl/cddl-files/extra.cddl
lehins: Done

eras/conway/impl/cddl-files/extra.cddl
Soupstraw: I think it would be a good idea to extend CDDL so that it's possible to express these kinds of constraints. Maybe once `Cuddle` is in a more developed state we can think about that.

Soupstraw: I think [Cue](https://cuelang.org/) has some interesting ideas that we could borrow.

lehins: I created a ticket that captures this as a future task https://github.com/input-output-hk/cardano-ledger/issues/3931

PR #3932, opened 2023-12-14
Add ledger event TxUTxODiff

Jimbo4350:
# Description

This emits consumed and created UTxOs due to a single tx

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/eras/allegra/impl/src/Cardano/Ledger/Allegra/Rules/Utxo.hs
+++ b/eras/allegra/impl/src/Cardano/Ledger/Allegra/Rules/Utxo.hs
@@ -57,12 +57,20 @@ import Cardano.Ledger.Shelley.Governance
 import Cardano.Ledger.Shelley.LedgerState (PPUPPredFailure)
 import qualified Cardano.Ledger.Shelley.LedgerState as Shelley
 import Cardano.Ledger.Shelley.PParams (Update)
-import Cardano.Ledger.Shelley.Rules (PpupEnv (..), ShelleyPPUP, ShelleyPpupPredFailure)
+import Cardano.Ledger.Shelley.Rules (
+  PpupEnv (..),
+  ShelleyPPUP,
+  ShelleyPpupPredFailure,
+ )
 import qualified Cardano.Ledger.Shelley.Rules as Shelley
 import Cardano.Ledger.Shelley.Tx (ShelleyTx (..))
 import Cardano.Ledger.Shelley.UTxO (txup)
 import Cardano.Ledger.TxIn (TxIn)
-import Cardano.Ledger.UTxO (EraUTxO (..), UTxO (..), txouts)
+import Cardano.Ledger.UTxO (
+  EraUTxO (..),
+  UTxO (..),
+  txouts,
+ )
 import qualified Cardano.Ledger.Val as Val
 import Cardano.Slotting.Slot (SlotNo)
 import Control.Monad.Trans.Reader (asks)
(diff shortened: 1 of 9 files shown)


eras/allegra/impl/src/Cardano/Ledger/Allegra/Rules/Utxo.hs
```diff
@@ -138,6 +138,14 @@ instance
...
   = UpdateEvent (Event (EraRule "PPUP" era))
   | TotalDeposits (SafeHash (EraCrypto era) EraIndependentTxBody) Coin
+  | -- | The UTxOs consumed and created by a signal tx
+    TxUTxODiff
+      ( (UTxO era)
+      , -- \^ UTxOs consumed
+        (UTxO era)
+      )
+
+-- \^ UTxOs created
```
smelc: The comment at the bottom seems not attached to the pair's member. Is that intended?

lehins: Using a tuple is totally unnecessary. No only haddock can't be used to document the fields, but also tuples come with unnecessary overhead with respect to runtime and extra code that works operates on it
```suggestion
      (UTxO era)
      -- ^ UTxOs consumed
        (UTxO era)
      -- ^ UTxOs created
```

Alternatively, if you are worried (I would be) that the user of this event can mixup which ones are additions and which ones are deletions, we can create a separate type and use it here instead:
```haskell
data UTxODiff era = UTxODiff
  { utxoDiffConsumed :: !(UTxO era)
  , utxoDiffProduced :: !(UTXO era)
  }
```

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Rules/Utxos.hs
```diff
@@ -142,6 +142,14 @@ data AlonzoUtxosEvent era
...
   | SuccessfulPlutusScriptsEvent (NonEmpty PlutusWithContext)
   | FailedPlutusScriptsEvent (NonEmpty PlutusWithContext)
+  | -- | The UTxOs consumed and created by a signal tx
+    TxUTxODiff
+      ( (UTxO era)
+      , -- \^ UTxOs consumed
+        (UTxO era)
+      )
+
+-- \^ UTxOs created
```
lehins: Ditto
```suggestion
      (UTxO era)
      -- ^ UTxOs consumed
      (UTxO era)
      -- ^ UTxOs created
```

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Utxo.hs
```diff
@@ -615,20 +629,23 @@ updateUTxOState ::
   CertState era ->
   GovState era ->
   (Coin -> m ()) ->
+  ((UTxO era, UTxO era) -> m ()) ->
```
lehins: Using tuples in function arguments is an anti-pattern
```suggestion
  (UTxO era -> UTxO era -> m ()) ->
```

eras/allegra/impl/src/Cardano/Ledger/Allegra/Rules/Utxo.hs
lehins: Problem is that it was intended as a haddock, but it is impossible to document elements of a tuple, so the haddock was turned into comment by fourmolu and then floated out in this funny way, since fourmolu isn't that great when dealing with comments

lehins (changes requested):
Just one issue: tuples should not be used in data types. Otherwise this is great.

lehins (approved):
Beautiful. Thank you!

Issue #3922, opened 2023-12-09
Fix CDDL for transaction witnesses


lehins:
Current binary format for witnesses has some issues:

* It allows duplicates, which are just discarded
* It uses list syntax, where all of them are either sets or maps
* Fields are optional, but they are also allowed to be empty 

We need to solve all of those issues in Conway mostly in a backwards compatible manner.

We should split this task into two separate smaller ones, because deserializer

1. Backwards incompatible changes:

* [x] - Disallow empty fields
* [x] #4698 (postponed till Dijkstra era)

2. Future compatibility. Add alternative and more appropriate usage of:

* [x] - Maps. Done in: #4040
* [x] - Sets. Done in: #4003 

Witnesses CDDL in the end for Conway should read:

```
transaction_witness_set =
  { ? 0: nonempty_set<vkeywitness>
  , ? 1: nonempty_set<native_script>
  , ? 2: nonempty_set<bootstrap_witness>
  , ? 3: nonempty_set<plutus_v1_script>
  , ? 4: nonempty_set<plutus_data>
  , ? 5: [+ redeemer ] / { + [ tag: redeemer_tag, index: uint ] => [data: plutus_data, ex_units: ex_units ] }
  , ? 6: nonempty_set<plutus_v2_script>
  , ? 7: nonempty_set<plutus_v3_script>
  }
```

lehins:
#3933 took care of some parts of this ticket that were possible with the current setup.

* empty fields are no longer allowed
* duplicate plutus scripts are not allowed either
* plutus scripts have an optional set tag.


This is the current CDDL:
```
transaction_witness_set =
  { ? 0: [+ vkeywitness ]
  , ? 1: [+ native_script ]
  , ? 2: [+ bootstrap_witness ]
  , ? 3: nonempty_set<plutus_v1_script>
  , ? 4: [+ plutus_data ]
  , ? 5: [+ redeemer ]
  , ? 6: nonempty_set<plutus_v2_script>
  , ? 7: nonempty_set<plutus_v3_script>
  }
```

This means that we still need to fix theses fields: `0`, `1`, `2`, `4`, `5`.

lehins:
Closing as Done. One part of it was converted into an issue and postponed till Dijkstra #4698

PR #3933, opened 2023-12-14
Improve cbor serialization of TxWits in Conway

teodanciu:
# Description

Solves part of:  https://github.com/IntersectMBO/cardano-ledger/issues/3922

* Tightens the cddl for `transaction_witness_set`  : 
   - not allowing empty collections for any field 
   - for plutus scripts, also not allowing duplicates
 * Updates the DecCBOR instance of TxWitsRaw to match the cddls changes

Note: still remaining: checking for duplicates in Annotator fields, not possible at the moment with the current Annotator mechanism

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/eras/alonzo/impl/CHANGELOG.md
+++ b/eras/alonzo/impl/CHANGELOG.md
@@ -60,6 +60,7 @@
 * Add `alwaysSucceedsLang` and `alwaysFailsLang`
 * Change `alwaysSucceeds` and `alwaysFails` to accept the language version at the type level.
 * Remove `genAlonzoScript`
+* Export `genAlonzoScript`, `genNativeScript`, `genPlutusScript` from Arbitrary
 
 ## 1.5.1.0
 
--- a/eras/alonzo/impl/CHANGELOG.md
+++ b/eras/alonzo/impl/CHANGELOG.md
@@ -59,8 +59,7 @@
 * Add `mkPlutusScript'`
 * Add `alwaysSucceedsLang` and `alwaysFailsLang`
 * Change `alwaysSucceeds` and `alwaysFails` to accept the language version at the type level.
-* Remove `genAlonzoScript`
-* Export `genAlonzoScript`, `genNativeScript`, `genPlutusScript` from Arbitrary
+* Export `genNativeScript` and `genPlutusScript` from Arbitrary
 
 ## 1.5.1.0
 
--- a/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxWits.hs
+++ b/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxWits.hs
@@ -82,6 +82,7 @@ import Cardano.Ledger.Binary (
   decodeList,
   decodeListLenOrIndef,
   decodeMapLikeEnforceNoDuplicates,
+  decodeNonEmptyList,
   encodeFoldableEncoder,
   encodeListLen,
   ifDecoderVersionAtLeast,
@@ -118,6 +119,7 @@ import Cardano.Ledger.Shelley.TxWits (ShelleyTxWits (..), keyBy, shelleyEqTxWits
 import Control.DeepSeq (NFData)
 import Control.Monad (when, (>=>))
 import Data.Bifunctor (Bifunctor (first))
+import Data.List.NonEmpty as NE (toList)
 import Data.Map.Strict (Map)
 import qualified Data.Map.Strict as Map
 import Data.Maybe (mapMaybe)
@@ -603,16 +605,7 @@ instance
       txWitnessField 1 =
         fieldAA
           addScripts
-          ( fmap Map.fromList
-              <$> listDecodeA
-                ( fmap
-                    ( \ns ->
-                        let script = fromNativeScript ns
-                         in (hashScript @era script, script)
-                    )
-                    <$> From
-                )
-          )
+          (D nativeScriptsDecoder)
       txWitnessField 2 =
         fieldAA
           (\x wits -> wits {atwrBootAddrTxWits = x})
@@ -627,6 +620,16 @@ instance
       txWitnessField 7 = fieldA addScripts (decodePlutus SPlutusV3)
       txWitnessField n = field (\_ t -> t) (Invalid n)
       {-# INLINE txWitnessField #-}
+
+      nativeScriptsDecoder :: Decoder s (Annotator (Map (ScriptHash (EraCrypto era)) (Script era)))
+      nativeScriptsDecoder =
+        ifDecoderVersionAtLeast
+          (natVersion @9)
+          (mapTraverseableDecoderA (decodeNonEmptyList pairDecoder) (Map.fromList . NE.toList))
+          (mapTraverseableDecoderA (decodeList pairDecoder) Map.fromList)
+        where
+          pairDecoder :: Decoder s (Annotator (ScriptHash (EraCrypto era), Script era))
+          pairDecoder = fmap (asHashedPair . fromNativeScript) <$> decCBOR
 
       addScripts ::
         Map (ScriptHash (EraCrypto era)) (Script era) ->
@@ -686,3 +689,10 @@ alonzoEqTxWitsRaw txWits1 txWits2 =
   shelleyEqTxWitsRaw txWits1 txWits2
     && eqRawType (txWits1 ^. datsTxWitsL) (txWits2 ^. datsTxWitsL)
     && eqRawType (txWits1 ^. rdmrsTxWitsL) (txWits2 ^. rdmrsTxWitsL)
+
+mapTraverseableDecoderA ::
+  Traversable f =>
+  Decoder s (f (Annotator a)) ->
+  (f a -> m b) ->
+  Decoder s (Annotator (m b))
+mapTraverseableDecoderA decList transformList = fmap transformList . sequence <$> decList
--- a/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxWits.hs
+++ b/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxWits.hs
@@ -614,7 +614,12 @@ instance
       txWitnessField 2 =
         fieldAA
           (\x wits -> wits {atwrBootAddrTxWits = x})
-          (setDecodeA From)
+          ( D $
+              ifDecoderVersionAtLeast
+                (natVersion @9)
+                (mapTraverseableDecoderA (decodeNonEmptyList decCBOR) (Set.fromList . NE.toList))
+                (mapTraverseableDecoderA (decodeList decCBOR) Set.fromList)
+          )
       txWitnessField 3 = fieldA addScripts (decodePlutus SPlutusV1)
       txWitnessField 4 =
         fieldAA
--- a/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxWits.hs
+++ b/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxWits.hs
@@ -84,7 +84,9 @@ import Cardano.Ledger.Binary (
   decodeNonEmptyList,
   encodeFoldableEncoder,
   encodeListLen,
+  encodeTag,
   ifDecoderVersionAtLeast,
+  ifEncodingVersionAtLeast,
   natVersion,
   setTag,
  )
@@ -543,7 +545,13 @@ instance AlonzoEraScript era => EncCBOR (AlonzoTxWitsRaw era) where
         Encode ('Closed 'Dense) (Map.Map (ScriptHash (EraCrypto era)) (Plutus l))
       encodePlutus slang =
         E
-          (encCBOR . map plutusBinary . Map.elems)
+          ( \m ->
+              let enc = encCBOR . map plutusBinary $ Map.elems m
+               in ifEncodingVersionAtLeast
+                    (natVersion @9)
+                    (encodeTag setTag <> enc)
+                    enc
+          )
           (Map.mapMaybe (toPlutusScript >=> toPlutusSLanguage slang) scripts)
       toScript ::
         forall l h. PlutusLanguage l => Map.Map h (Plutus l) -> Map.Map h (Script era)
--- a/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxWits.hs
+++ b/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxWits.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE BangPatterns #-}
 {-# LANGUAGE DataKinds #-}
 {-# LANGUAGE DeriveGeneric #-}
 {-# LANGUAGE DerivingVia #-}
@@ -77,9 +78,15 @@ import Cardano.Ledger.Binary (
   EncCBOR (..),
   EncCBORGroup (..),
   ToCBOR (..),
+  allowTag,
   decodeList,
+  decodeListLenOrIndef,
+  decodeMapLikeEnforceNoDuplicates,
   encodeFoldableEncoder,
   encodeListLen,
+  ifDecoderVersionAtLeast,
+  natVersion,
+  setTag,
  )
 import Cardano.Ledger.Binary.Coders
 import Cardano.Ledger.Core
@@ -109,7 +116,7 @@ import Cardano.Ledger.Plutus.Language (
 import Cardano.Ledger.SafeHash (SafeToHash (..))
 import Cardano.Ledger.Shelley.TxWits (ShelleyTxWits (..), keyBy, shelleyEqTxWitsRaw)
 import Control.DeepSeq (NFData)
-import Control.Monad ((>=>))
+import Control.Monad (when, (>=>))
 import Data.Bifunctor (Bifunctor (first))
 import Data.Map.Strict (Map)
 import qualified Data.Map.Strict as Map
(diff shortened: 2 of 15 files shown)


eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxWits.hs
```diff
@@ -682,3 +690,9 @@ alonzoEqTxWitsRaw txWits1 txWits2 =
   shelleyEqTxWitsRaw txWits1 txWits2
     && eqRawType (txWits1 ^. datsTxWitsL) (txWits2 ^. datsTxWitsL)
     && eqRawType (txWits1 ^. rdmrsTxWitsL) (txWits2 ^. rdmrsTxWitsL)
+
+nonEmptyListDecodeA :: Decode ('Closed 'Dense) (Annotator x) -> Decode ('Closed 'Dense) (Annotator [x])
+nonEmptyListDecodeA dx = D $ do
```
teodanciu: @lehins -  this is the way I wrote it to begin with before you changed it. It feels wrong, but I don't know where else I could check that the collection is not empty. Happy to change it to make it better

lehins: Yeah, that's fine. I'd suggest moving it next to the definition of `listDecodeA`

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxWits.hs
```diff
@@ -612,13 +618,64 @@ instance
...
+        ifDecoderVersionAtLeast
+          (natVersion @9)
+          (decode nonEmptyList)
+          (decode list)
+        where
+          asPair ns =
+            let script = fromNativeScript ns
+             in (hashScript @era script, script)
+          pairDecoder :: Decoder s (Annotator (ScriptHash (EraCrypto era), Script era))
+          pairDecoder = fmap asPair <$> decCBOR
```
lehins: This way it can be reused for Plutus scripts too
```suggestion
          asHashedPair script =
            let !scriptHash = hashScript @era script
             in (scriptHash, script)
          pairDecoder :: Decoder s (Annotator (ScriptHash (EraCrypto era), Script era))
          pairDecoder = fmap (asHashedPair . fromNativeScript) <$> decCBOR
```

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxWits.hs
```diff
@@ -612,13 +618,64 @@ instance
...
+            fmap Map.fromList <$> listDecodeA (D pairDecoder)
+
+      scriptDecoder ::
+        Decoder s (Script era) ->
+        Decoder s (Map (ScriptHash (EraCrypto era)) (Script era))
+      scriptDecoder decodeScript =
+        fmap Map.fromList $ decodeCollection decodeListLenOrIndef $ do
+          script <- decodeScript
+          let !scriptHash = hashScript @era script
+          pure (scriptHash, script)
```
lehins:
```suggestion
          asHashedPair <$> decodeScript
```

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxWits.hs
```diff
@@ -612,13 +618,64 @@ instance
...
+      {-# INLINE scriptDecoder #-}
+
+      scriptDecoderV9 ::
+        Decoder s (Script era) ->
+        Decoder s (Map (ScriptHash (EraCrypto era)) (Script era))
+      scriptDecoderV9 decodeScript = do
+        scriptMap <- decodeMapLikeEnforceNoDuplicates decodeListLenOrIndef $ do
+          script <- decodeScript
+          let !scriptHash = hashScript @era script
+          pure (scriptHash, script)
```
lehins:
```suggestion
          asHashedPair <$> decodeScript
```

eras/alonzo/impl/testlib/Test/Cardano/Ledger/Alonzo/Binary/TxWitsSpec.hs
```diff
@@ -0,0 +1,238 @@
...
+import Test.Cardano.Ledger.Common
+
+spec ::
+  forall era.
+  ( AlonzoEraScript era
+  , Script era ~ AlonzoScript era
+  ) =>
+  Proxy era ->
+  Spec
+spec pxy = do
```
lehins: Let's use consistently `AllowAmbiguousTypes` for `spec`.
```suggestion
spec ::
  forall era.
  ( AlonzoEraScript era
  , Script era ~ AlonzoScript era
  ) =>
  Spec
spec = do
```

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxWits.hs
teodanciu: Yes great, was thinking of removing this duplication and then forgot,thanks.

eras/alonzo/impl/testlib/Test/Cardano/Ledger/Alonzo/Binary/TxWitsSpec.hs
teodanciu: Nice. Also did it for `CostModelsSpec`, removed tons of noise, thanks.

libs/cardano-ledger-binary/src/Cardano/Ledger/Binary/Decoding/Decoder.hs
```diff
@@ -758,14 +758,15 @@ decodeMapLikeEnforceNoDuplicates ::
...
   Decoder s (Map.Map k v)
-decodeMapLikeEnforceNoDuplicates decodeLenOrIndef =
+decodeMapLikeEnforceNoDuplicates decodeLenOrIndef decodeKv = do
+  void decodeTagMaybe
   -- We first decode into a list because most of the time the encoded Map will be in sorted
   -- order and there is a nice optimization on the `Map.fromList` that can take advantage of
   -- that fact. In case when encoded data is not sorted the penalty of going through a list
   -- is insignificant.
-  decodeListLikeEnforceNoDuplicates decodeLenOrIndef (:) $ \xs ->
+  decodeListLikeEnforceNoDuplicates decodeLenOrIndef (:) (\xs ->
```
teodanciu: EDIT: this comment refers to the line above, `void recodeTagMaybe`

I have doubts about whether to add this: here, or in the decoder that uses this function, or the opposite, even lower. 

Since it's enforcing no duplicates, it's  conceivable that other types deserialized with this will be specified to have the tag. 
On the other hand, we also have `decodeListLikeEnforceNoDuplicates`, so maybe the tag could be there. However, the latter is used for `decodeSetLikeEnforceNoDuplicates`, which has a mandatory tag.

I could add it in `decodeListLikeEnforceNoDuplicates` and parameterize whether to have it or not. 

What do you think is best @lehins ?

lehins: Nope. Please revert any changes to this function.

This function is named `decodeListLikeEnforceNoDuplicates`, which has nothing to do with sets and in fact is used with an `OMap`, which cannot have a tag for a prefix

Another note is: do not use `decodeTagMaybe`, because it will allow any tag. We need to allow 258 tag, which means you can use `allowTag setTag` for doing that

lehins: `decodeSetLikeEnforceNoDuplicates` does not have a mandatory tag, it allows a tag, which means it will consume 258 tag if it is present and it will fail if the tag differs from 258

teodanciu: Indeed, I misread `allowTag` thinking it was mandatory, but I see my mistake, thanks!

eras/alonzo/impl/CHANGELOG.md
```diff
@@ -60,6 +60,7 @@
 * Add `alwaysSucceedsLang` and `alwaysFailsLang`
 * Change `alwaysSucceeds` and `alwaysFails` to accept the language version at the type level.
 * Remove `genAlonzoScript`
+* Export `genAlonzoScript`, `genNativeScript`, `genPlutusScript` from Arbitrary
```
lehins:
```suggestion
* Export `genNativeScript`, `genPlutusScript` from Arbitrary
```

eras/conway/impl/cddl-files/conway.cddl
```diff
@@ -427,18 +427,31 @@ drep_voting_thresholds =
...
+plutus_v1_scripts = distinct_nonempty_set_or_list<plutus_v1_script>
 plutus_v1_script = bytes
+
+; plutus_v2_scripts = nonempty_set<plutus_v2_script>
+plutus_v2_scripts = distinct_nonempty_set_or_list<plutus_v2_script>
 plutus_v2_script = bytes
+
+; plutus_v3_scripts = nonempty_set<plutus_v3_script>
+plutus_v3_scripts = distinct_nonempty_set_or_list<plutus_v3_script>
 plutus_v3_script = bytes
```
lehins: If I understand it correctly, I don't think definition of `distinct_nonempty_set_or_list` is needed. We can simplify it a bit (which is a good thing, cause I'll get tons of questions otherwise)
```suggestion
plutus_v1_scripts = nonempty_set<plutus_script>

plutus_v2_scripts = nonempty_set<plutus_script>

plutus_v3_scripts = nonempty_set<plutus_script>

; The real type of plutus script is `plutus_script = bytes`, however we enforce 
; uniqueness when many scripts are supplied, thus we need to hack around for tests
; in order to avoid generating duplicates, because the cddl tool we use for 
; roundtrip testing doesn't generate distinct collections
plutus_script = distinct_bytes<bytes>
```

eras/conway/impl/cddl-files/extra.cddl
```diff
@@ -53,3 +53,11 @@ bounded_bytes = bytes .size (0..64)
   ;  ( reminder: in CBOR, the indefinite-length encoding of bytestrings
   ;    consists of a token #2.31 followed by a sequence of definite-length
   ;    encoded bytestrings and a stop code )
+
+; a non-empty list or set of at most 6 different elements of different sizes, hence distinct.
+; the type must support .size, for example: bytes or uint
+distinct_nonempty_set_or_list<a> =
+ #6.258(distinct_bytes<a>) /
+ distinct_bytes<a>
```
lehins:
```suggestion
```

eras/conway/impl/cddl-files/extra.cddl
```diff
@@ -53,3 +53,11 @@ bounded_bytes = bytes .size (0..64)
...
+
+distinct_bytes<a> = [ a .size 8, ? a .size 16, ? a .size 20, ? a .size 24, ? a .size 30, ? a .size 32]
```
lehins: Love it! :smile:

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxWits.hs
```diff
@@ -532,8 +545,14 @@ instance AlonzoEraScript era => EncCBOR (AlonzoTxWitsRaw era) where
...
         E
-          (encCBOR . map plutusBinary . Map.elems)
+          ( \m ->
+              ifEncodingVersionAtLeast
+                (natVersion @9)
+                (encodeTag setTag <> encElems m)
+                (encElems m)
+          )
           (Map.mapMaybe (toPlutusScript >=> toPlutusSLanguage slang) scripts)
+      encElems = encCBOR . map plutusBinary . Map.elems
```
lehins: I think this way is slightly simpler
```suggestion
          ( \m ->
              let enc = encCBOR . map plutusBinary $ Map.elems m in
              ifEncodingVersionAtLeast
                (natVersion @9)
                (encodeTag setTag <> enc)
                enc
          )
          (Map.mapMaybe (toPlutusScript >=> toPlutusSLanguage slang) scripts)
```

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxWits.hs
```diff
@@ -612,15 +643,62 @@ instance
...
+        pure scriptMap
+      {-# INLINE scriptDecoderV9 #-}
+
+      scriptDecoder ::
+        Decoder s (Script era) ->
+        Decoder s (Map (ScriptHash (EraCrypto era)) (Script era))
+      scriptDecoder decodeScript =
+        fmap Map.fromList $
+          decodeCollection decodeListLenOrIndef $
+            asHashedPair <$> decodeScript
```
lehins:
```suggestion
        fmap Map.froList $ decodeList $ asHashedPair <$> decodeScript
```

libs/cardano-ledger-binary/CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## 1.3.0.0
 
+* Add `decodeNonEmptyList`
+* Expose `decodeCollection` and `decodeCollectionWithLen`
```
lehins: These guys should stay hidden, they are a little bit too specific. It is safer to use `decodeList`
```suggestion
```

libs/cardano-ledger-binary/src/Cardano/Ledger/Binary/Decoding/Decoder.hs
```diff
@@ -74,6 +75,8 @@ module Cardano.Ledger.Binary.Decoding.Decoder (
   decodeSetLikeEnforceNoDuplicates,
   decodeListLikeEnforceNoDuplicates,
   decodeMapContents,
+  decodeCollection,
+  decodeCollectionWithLen,
```
lehins: As per comment on changelog
```suggestion
```

libs/cardano-ledger-binary/src/Cardano/Ledger/Binary/Decoding/Decoder.hs
```diff
@@ -482,6 +485,13 @@ decodeList decodeValue =
     (decodeListWith decodeValue)
 {-# INLINE decodeList #-}
 
+decodeNonEmptyList :: Decoder s a -> Decoder s [a]
```
lehins: I think it will be cooler if we also reflect this invariant in the type:
```suggestion
decodeNonEmptyList :: Decoder s a -> Decoder s (NonEmpty a)
```

You can also replace the `DecCBOR` instance for `NonEmpty` with this function.

libs/cardano-ledger-binary/src/Cardano/Ledger/Binary/Decoding/Decoder.hs
```diff
@@ -482,6 +485,13 @@ decodeList decodeValue =
...
+decodeNonEmptyList decodeValue = do
+  xs <- decodeList decodeValue
+  when (null xs) $ fail "Empty list found, expected non-empty"
+  pure xs
```
lehins: Will require `import Data.List.NonEmpty as NE`
```suggestion
  case NE.nonEmpty xs of
    Nothing -> fail "Empty list found, expected non-empty"
    Just ne -> pure ne
```

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxWits.hs
```diff
@@ -633,3 +711,6 @@ alonzoEqTxWitsRaw txWits1 txWits2 =
   shelleyEqTxWitsRaw txWits1 txWits2
     && eqRawType (txWits1 ^. datsTxWitsL) (txWits2 ^. datsTxWitsL)
     && eqRawType (txWits1 ^. rdmrsTxWitsL) (txWits2 ^. rdmrsTxWitsL)
+
+mapListDecoderA :: Decoder s [Annotator a] -> ([a] -> m b) -> Decoder s (Annotator (m b))
+mapListDecoderA decList transformList = fmap transformList . sequence <$> decList
```
lehins: My suggestion about `NonEmpty` will require this slight change (also `m` was unnecessary):
```suggestion
mapListDecoderA :: Foldable f => Decoder s (f (Annotator a)) -> (f a -> b) -> Decoder s (Annotator b)
mapListDecoderA decList transformList = fmap transformList . sequence . toList <$> decList
```

lehins (approved):
This looks great. Just a few suggestions.

eras/alonzo/impl/CHANGELOG.md
teodanciu: Flatmapped  :heavy_check_mark:

eras/conway/impl/cddl-files/conway.cddl
teodanciu: That would be awesome, but I think in order for this to work,  the definition of `nonempty_set` would have to be this:
`nonempty_set<a> = #6.258(a) / a`
instead of  what it is, namely:
`nonempty_set<a> = #6.258([+ a]) / [+ a]` . 

With your suggestion, there is an extra list in `plutus_v1_scripts` , from the need to define `distinct_bytes` as a  list:
`nonempty_set<plutus_script>` = ` #6.258([+ plutus_script]) / [+ plutus_script]` =  ` #6.258([+ distinct_bytes<bytes>]) / [+ distinct_bytes<bytes>]`  = ` #6.258([+ [bytes, bytes... ]) / [+ [bytes, bytes ...]]`

teodanciu: As discussed, this problem goes away if `distinct` is replaced with `a .size 8 / a.size 16 ...`  (instead of the list). 
Thanks!

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxWits.hs
teodanciu: Love it, thanks!

PR #3934, opened 2023-12-15
Type safety bugfixes

lehins:
# Description

#3893 exported a bunch of constructors that it shouldn't have

This PR fixes that.

Besides that there are two more fixes:

* Enum instance for Version was dangerous, since it was derived from the underlying `Word64`
* `decodeVMap` did not enforce no duplicate. Which is safe to fix without a hard fork, because that type is not used for the on-chain data

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/libs/cardano-ledger-binary/src/Cardano/Ledger/Binary/Decoding/Decoder.hs
+++ b/libs/cardano-ledger-binary/src/Cardano/Ledger/Binary/Decoding/Decoder.hs
@@ -65,10 +65,20 @@ module Cardano.Ledger.Binary.Decoding.Decoder (
   setTag,
   decodeMap,
   decodeMapByKey,
+  decodeMapLikeEnforceNoDuplicates,
   decodeVMap,
   decodeSeq,
   decodeStrictSeq,
   decodeSetTag,
+  decodeListLikeWithCount,
+  decodeSetLikeEnforceNoDuplicates,
+  decodeListLikeEnforceNoDuplicates,
+  decodeMapContents,
+  decodeMapNoDuplicates,
+
+  -- **** Applicaitve
+  decodeMapTraverse,
+  decodeMapContentsTraverse,
 
   -- *** Time
   decodeUTCTime,
@@ -76,10 +86,6 @@ module Cardano.Ledger.Binary.Decoding.Decoder (
   -- *** Network
   decodeIPv4,
   decodeIPv6,
-  decodeMapContents,
-  decodeMapNoDuplicates,
-  decodeMapTraverse,
-  decodeMapContentsTraverse,
 
   -- ** Lifted @cborg@ decoders
   decodeBool,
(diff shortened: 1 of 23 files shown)


libs/cardano-ledger-binary/src/Cardano/Ledger/Binary/Decoding/Decoder.hs
```diff
@@ -861,49 +859,84 @@ decodeSetEnforceNoDuplicates ::
...
-  Monoid (f a) =>
-  -- | Check for membership. Decoder will fail if this predicate returns True
-  (a -> f a -> Bool) ->
-  -- | Add an aelement into the decoded Set like data structure
-  (a -> f a -> f a) ->
+  forall s a b c.
+  Monoid b =>
+  -- | Add an element into the decoded Set like data structure
+  (a -> b -> b) ->
+  -- | Get the number of elements in the data structure
```
teodanciu: Could we specify in the comment the meaning of both elements of the tuple that this function should return? I'm  a little confused about what the `c` value should be

teodanciu: Looking at how it's used, I understand now, it's a way to transform the intermediate structure (`b`, in our case list) to the final one. I wonder if `final` is a good way to express this? Maybe `getSetWithLen` ? It's tough to name these things!

teodanciu (approved):
Looks good to me

libs/cardano-ledger-binary/src/Cardano/Ledger/Binary/Decoding/Decoder.hs
lehins: Unfortunately it is not necessarily going to be a `Set`, it could for example be an `OSet` or even a `Map`, like in the example of scripts in the TxWits type. That's why I called it a "SetLike", cause on the wire it looks like a Set, but in Haskell it can be something else entirely.

> It's tough to name these things!

:100: 


I'll try to document the argument a little better. That should be a better solution to this problem when compared to the name of the argument anyways.

libs/cardano-ledger-binary/src/Cardano/Ledger/Binary/Decoding/Decoder.hs
```diff
@@ -861,49 +857,91 @@ decodeSetEnforceNoDuplicates ::
...
+  -- | Add an element into the decoded Set like data structure
+  (a -> b -> b) ->
+  -- | Get the final data structure from the decoded sequence of values and the number of
+  -- elements it contains. This is useful when a sequence on the wire is represented by a
+  -- @set@, namely no duplicates are allowed, while the Haskell representation uses some
+  -- other data structure that enforces no duplicates by some other means. For example a
+  -- `Map`, where keys are hashes of the values encoded on the wire. The final size of the
+  -- data strucutre taht isreturned will be used to enforce the invariant of a Set, that
+  -- the number of elements decoded matches the final size of the Set, thus ensuring no
+  -- duplicates were encountered.
```
teodanciu, line 938: That's an awesome explanation, thank you!

Issue #3942, opened 2023-12-18
Change `hkdCommitteeMaxTermLengthL` to use `EpochInterval`


lehins:
In #3876 we adjusted a few protocol parameters to use `EpochInterval`. We forgot about one protocol parameter that also should use that type, namely `committeeMaxTermLength`

lehins:
Fixed in #3947

Issue #3906, opened 2023-12-02
Change width of `ppMaxTxSizeL`


lehins:
[Network relies for transaction size to be `Word32`:](https://github.com/input-output-hk/ouroboros-network/blob/684f8e84fe5208a7f0ad6e3ee7814f008cd9e8a0/ouroboros-network-protocols/src/Ouroboros/Network/Protocol/TxSubmission2/Type.hs#L26)

We should ensure it is impossible for protocol parameter to be set to a higher value.

Change `ppMaxTxSizeL` to `Word32`

lehins:
Fixed in #3947

Issue #3905, opened 2023-12-02
Restrict width of `ppMaxBHSizeL` and `ppMaxBBSizeL`


lehins:
* Change `ppMaxBHSizeL` to `Word16` [(in consensus)](https://github.com/input-output-hk/ouroboros-consensus/blob/3a1fde044c31a254a2642079459cb947fb49c25a/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/VolatileDB/API.hs#L155)
* and `ppMaxBBSizeL` to `Word32` [(in network)](https://github.com/input-output-hk/ouroboros-network/blob/684f8e84fe5208a7f0ad6e3ee7814f008cd9e8a0/ouroboros-network-api/src/Ouroboros/Network/SizeInBytes.hs#L12) and [(in consensus)](https://github.com/input-output-hk/ouroboros-consensus/blob/3a1fde044c31a254a2642079459cb947fb49c25a/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/VolatileDB/Impl/Types.hs#L113-L114)

Consensus and network rely on those values to not exceed this width, thus to avoid dangers of overflow, we should change those PParams.

These two types need to change as well:

https://github.com/input-output-hk/cardano-ledger/blob/2ac3c2061578122c4a99cad83456855bdd9f3bba/eras/shelley/impl/src/Cardano/Ledger/Chain.hs#L32-L33

lehins:
Fixed in #3947

PR #3947, opened 2023-12-19
Change the type of some pparam fields to align with how they are represented in Consensus

TimSheard:
Addresses issue  #3942, issue #3905, and issue #3906
Changes a few PParam fields to  Word32 and Word16 to align with how they are represents in Consensus.

Also changed 2 fields in the type BHeaderView, to align with the PParams changes

# Checklist

- [ ] Self-reviewed the diff


CHANGELOG.md
```diff
@@ -8,7 +8,8 @@ If you are looking for the Ledger Releasing and Versioning Process then you can
 [RELEASING.md](https://github.com/intersectmbo/cardano-ledger/blob/master/RELEASING.md#changelogmd).
 
 ## 8.7
-
+-- Changed the underlying types of the 3 PParam fields MaxTxBodySize (to Word32) MaxBBodySize (to Word32) MaxBHeaderSize (to Word16)
+-- Changed the fields of BHeaderView to align with the changes in PParams
```
lehins: This is a top level changelog that only lists information relevant to `cardano-node` users. So, neither of those changes are relevant for people who don't use ledger as a library. However, if they were relevant, they would have to be listed in a new section, since 8.7 has already been released and the next one will be 8.8
```suggestion

```

eras/shelley/test-suite/src/Test/Cardano/Ledger/Shelley/Generator/Update.hs
```diff
@@ -261,9 +261,9 @@ genShelleyPParamsUpdate c@Constants {maxMinFeeA, maxMinFeeB, maxMajorPV} pp = do
     emptyPParamsUpdate
       & ppuMinFeeAL .~ fmap Coin minFeeA
       & ppuMinFeeBL .~ fmap Coin minFeeB
-      & ppuMaxBBSizeL .~ maxBBSize
-      & ppuMaxTxSizeL .~ maxTxSize
-      & ppuMaxBHSizeL .~ maxBHSize
+      & ppuMaxBBSizeL .~ (fromIntegral <$> maxBBSize)
+      & ppuMaxTxSizeL .~ (fromIntegral <$> maxTxSize)
+      & ppuMaxBHSizeL .~ (fromIntegral <$> maxBHSize)
```
lehins: This guy will overflow. Since `200000` is much bigger that `65535`

I'd suggest creating a separate `genWord16` and `genWord32` instead of using `genNatural`

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/Preds/PParams.hs
```diff
@@ -79,9 +79,9 @@ genPParams proof tx bb bh = do
           , Prices (Script.Prices (nonNegativeInterval 1.0) (nonNegativeInterval 1.0))
           , defaultCostModels proof
           , MaxValSize 1000
-          , MaxTxSize tx
-          , MaxBBSize bb
-          , MaxBHSize bh
+          , MaxTxSize (fromIntegral tx) -- In the Model these are Natural
+          , MaxBBSize (fromIntegral bb) -- But in the PParams, they are
+          , MaxBHSize (fromIntegral bh) -- Word32, Word32, and Word16
```
lehins: Are there any guarantees they are not going to overflow when converted from a Natural?

It might be worth using something like that instead of `fromIntegral`:

```haskell
fromIntegralBounded :: (HasCallStack, Integral a, Bounded b, Num b) => a -> b
fromIntegralBounded x =
  | toInteger (minBound :: b) <= xi && toInteger (maxBound :: b) = fromIntegral x
  | otherwise = 
    error $ show x <> " is out of bounds: " <> show (minBound :: b, maxBound ::b)
  where
    xi = toInteger x
```

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/Vars.hs
```diff
@@ -1099,9 +1099,15 @@ maxTxSize p =
...
         NaturalR
-        (Yes (PParamsR p) (withEraPParams p (pparamsWrapperL . ppMaxTxSizeL)))
+        (Yes (PParamsR p) (withEraPParams p (pparamsWrapperL . ppMaxTxSizeL . word32NaturalL)))
     )
 
+word32NaturalL :: Lens' Word32 Natural
+word32NaturalL = lens fromIntegral (\_ y -> fromIntegral (toInteger y))
+
+word16NaturalL :: Lens' Word16 Natural
+word16NaturalL = lens fromIntegral (\_ y -> fromIntegral (toInteger y))
```
lehins: Same here. These lenses could use some protection against overflow

libs/cardano-protocol-tpraos/src/Cardano/Protocol/TPraos/BHeader.hs
```diff
@@ -354,16 +354,16 @@ prevHashToNonce = \case
 issuerIDfromBHBody :: Crypto c => BHBody c -> KeyHash 'BlockIssuer c
 issuerIDfromBHBody = hashKey . bheaderVk
 
-bHeaderSize :: forall c. BHeader c -> Int
-bHeaderSize = BS.length . bHeaderBytes
+bHeaderSize :: forall c. BHeader c -> Word16
+bHeaderSize = fromIntegral . BS.length . bHeaderBytes
```
lehins: This can easily overflow and that would be really bad! We need to switch it back to `Int`
```suggestion
bHeaderSize :: forall c. BHeader c -> Int
bHeaderSize = BS.length . bHeaderBytes
```

libs/cardano-ledger-core/src/Cardano/Ledger/BHeaderView.hs
```diff
@@ -22,9 +22,9 @@ data BHeaderView c = BHeaderView
   { bhviewID :: KeyHash 'BlockIssuer c
   -- ^ The block issuer. In the TPraos protocol, this can be a
   --  Genesis delegate, everywhere else it is the stake pool ID.
-  , bhviewBSize :: Natural
+  , bhviewBSize :: Word32
   -- ^ The purported size (in bytes) of the block body.
-  , bhviewHSize :: Int
+  , bhviewHSize :: Word16
```
lehins: Switch it back to `Int`, since that is the field that used for reporting the actual size of the header. In this PR we want to change the type only for parameters that limit the size.
```suggestion
  , bhviewHSize :: Int
```

eras/shelley/test-suite/src/Test/Cardano/Ledger/Shelley/Generator/Update.hs
TimSheard: So now I use explicit  choose(n,m) where n and m are word16, word32, or word64, that way I don't need to use genNatural.

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/Preds/PParams.hs
TimSheard: So I did this
```
fromIntegralBounded :: forall a b.
  (HasCallStack, Integral a, Show a, Integral b, Bounded b, Show b) =>
  String -> a -> b
fromIntegralBounded name x
  | toInteger (minBound :: b) <= xi && xi <= toInteger (maxBound :: b) = fromIntegral x
  | otherwise = 
    error $ "While converting "++name++", "++show x <> " is out of bounds: " <> show (minBound :: b, maxBound ::b)
  where
    xi = toInteger x
```

This way the error tells what you were converting. Had to fix a few typos

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/Vars.hs
TimSheard: Done

CHANGELOG.md
```diff
@@ -8,7 +8,6 @@ If you are looking for the Ledger Releasing and Versioning Process then you can
 [RELEASING.md](https://github.com/intersectmbo/cardano-ledger/blob/master/RELEASING.md#changelogmd).
 
 ## 8.7
```
lehins, line 10: Bring back the newline
```suggestion
## 8.7

```

eras/conway/impl/src/Cardano/Ledger/Conway/PParams.hs
```diff
@@ -133,7 +133,7 @@ ppDRepVotingThresholdsL = ppLens . hkdDRepVotingThresholdsL @era @Identity
 ppCommitteeMinSizeL :: forall era. ConwayEraPParams era => Lens' (PParams era) Natural
 ppCommitteeMinSizeL = ppLens . hkdCommitteeMinSizeL @era @Identity
 
-ppCommitteeMaxTermLengthL :: forall era. ConwayEraPParams era => Lens' (PParams era) EpochNo
+ppCommitteeMaxTermLengthL :: forall era. ConwayEraPParams era => Lens' (PParams era) EpochInterval
```
lehins, line 136: Change of type for `ppCommitteeMaxTermLengthL` and `ppuCommitteeMaxTermLengthL` warrants a changelog entry

libs/cardano-ledger-core/src/Cardano/Ledger/Core/PParams.hs
```diff
@@ -375,15 +376,15 @@ ppMinFeeBL :: forall era. EraPParams era => Lens' (PParams era) Coin
 ppMinFeeBL = ppLens . hkdMinFeeBL @era @Identity
 
 -- | Maximal block body size
-ppMaxBBSizeL :: forall era. EraPParams era => Lens' (PParams era) Natural
+ppMaxBBSizeL :: forall era. EraPParams era => Lens' (PParams era) Word32
```
lehins, line 379: We need a changelog entry at least for the 6 lenses `ppMaxBBSizeL`, `ppMaxTxSizeL`, `ppMaxBHSizeL`, `ppuMaxBBSizeL`, `ppuMaxTxSizeL` and `ppuMaxBHSizeL`

Because they have a direct impact on the downstream users

eras/shelley/impl/src/Cardano/Ledger/Chain.hs
```diff
@@ -70,7 +70,7 @@ chainChecks maxpv ccd bhv = do
   unless (m <= maxpv) $ throwError (ObsoleteNodeCHAIN m maxpv)
   unless (fromIntegral (bhviewHSize bhv) <= ccMaxBHSize ccd) $
```
lehins: Here is another overflow, which we must guard against properly:
```suggestion
  unless (bhviewHSize bhv <= (fromIntegral :: Word16 -> Int) (ccMaxBHSize ccd)) $
```

Using `fromIntegral` without a type signature is really dangerous. We should start adding a type signature at every use site.

lehins (changes requested):
One more dangerous bug needs to be squashed and I think we are good to go.

libs/cardano-ledger-core/src/Cardano/Ledger/Core/PParams.hs
TimSheard: done

eras/conway/impl/src/Cardano/Ledger/Conway/PParams.hs
TimSheard: done

CHANGELOG.md
TimSheard: done

CHANGELOG.md
```diff
@@ -8,7 +8,8 @@ If you are looking for the Ledger Releasing and Versioning Process then you can
...
-
+	
+- Changed the type of lenses ppCommitteeMaxTermLengthL, ppuCommitteeMaxTermLengthL, ppMaxBBSizeL, ppMaxTxSizeL, ppMaxBHSizeL, ppuMaxBBSizeL, ppuMaxTxSizeL and ppuMaxBHSizeL
```
lehins: That is not the changelog that we need to change. Changelog for the package need to be adjusted.

```suggestion
```

eras/conway/impl/src/Cardano/Ledger/Conway/PParams.hs
lehins: Wrong changelog

libs/cardano-ledger-core/src/Cardano/Ledger/Core/PParams.hs
lehins: Wring changelog

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Newpp.hs
```diff
@@ -108,7 +108,7 @@ newPpTransition = do
         == utxosDeposited utxoSt
           ?! UnexpectedDepositPot (Coin oblgCurr) (utxosDeposited utxoSt)
 
-      if (ppNew' ^. ppMaxTxSizeL + ppNew' ^. ppMaxBHSizeL) < (ppNew' ^. ppMaxBBSizeL)
+      if (ppNew' ^. ppMaxTxSizeL + fromIntegral (ppNew' ^. ppMaxBHSizeL) < ppNew' ^. ppMaxBBSizeL)
```
lehins: Although this is not really possible, since we would never submit such high values in Babbage (last era that runs `NEWPP` rule), there is a chance of overflow here. It is safer to convert to Integer. Here is another example of dangerous usage of `fromIntegral`
```suggestion
      if (toInteger (ppNew' ^. ppMaxTxSizeL) + toInteger (ppNew' ^. ppMaxBHSizeL) < toInteger (ppNew' ^. ppMaxBBSizeL))
```

eras/shelley/test-suite/src/Test/Cardano/Ledger/Shelley/Generator/Update.hs
```diff
@@ -242,9 +240,9 @@ genShelleyPParamsUpdate c@Constants {maxMinFeeA, maxMinFeeB, maxMajorPV} pp = do
   -- TODO generate Maybe types so not all updates are full
   minFeeA <- genM $ genInteger 0 (unCoin maxMinFeeA)
   minFeeB <- genM $ genInteger 0 (unCoin maxMinFeeB)
-  maxBBSize <- genM $ genNatural low hi
-  maxTxSize <- genM $ genNatural low hi
-  maxBHSize <- genM $ genNatural low hi
+  maxBBSize <- genM $ choose (low, hi)
+  maxTxSize <- genM $ choose (low, hi)
+  maxBHSize <- genM $ choose (25000, 65535 :: Word16) -- Must stay in the range of Word16, but can't be too small
```
lehins: This way the intention is clearer
```suggestion
  -- Must stay in the range of Word16, but can't be too small
  maxBHSize <- genM $ choose (25000, maxBound :: Word16)
```

eras/shelley/test-suite/src/Test/Cardano/Ledger/Shelley/Generator/Update.hs
```diff
@@ -261,9 +259,9 @@ genShelleyPParamsUpdate c@Constants {maxMinFeeA, maxMinFeeB, maxMajorPV} pp = do
     emptyPParamsUpdate
       & ppuMinFeeAL .~ fmap Coin minFeeA
       & ppuMinFeeBL .~ fmap Coin minFeeB
-      & ppuMaxBBSizeL .~ maxBBSize
-      & ppuMaxTxSizeL .~ maxTxSize
-      & ppuMaxBHSizeL .~ maxBHSize
+      & ppuMaxBBSizeL .~ (fromIntegral <$> maxBBSize)
+      & ppuMaxTxSizeL .~ (fromIntegral <$> maxTxSize)
+      & ppuMaxBHSizeL .~ (fromIntegral <$> maxBHSize)
```
lehins: Redundant conversion
```suggestion
      & ppuMaxBBSizeL .~ maxBBSize
      & ppuMaxTxSizeL .~ maxTxSize
      & ppuMaxBHSizeL .~ maxBHSize
```

eras/shelley/test-suite/src/Test/Cardano/Ledger/Shelley/Generator/Update.hs
```diff
@@ -138,14 +138,18 @@ genPParams c@Constants {maxMinFeeA, maxMinFeeB, minMajorPV, maxMajorPV} = do
...
     szGen = do
-      blockBodySize <- genNatural low hi
+      blockBodySize <- choose (low, hi)
       (blockBodySize,,)
-        <$> rangeUpTo (blockBodySize `div` 2)
-        <*> rangeUpTo (blockBodySize `div` 2)
-    rangeUpTo :: Natural -> Gen Natural
-    rangeUpTo upper = genNatural low upper
+        <$> choose (blockBodySize, blockBodySize `div` 2)
+        <*> choose (25000, 65535 :: Word16) -- Must stay in the range of Word16, but can't be too small
```
lehins: Ditto
```suggestion
        <*> choose (25000, maxBound :: Word16) -- Must stay in the range of Word16, but can't be too small
```

eras/conway/impl/src/Cardano/Ledger/Conway/PParams.hs
TimSheard: OK it now where it beongs

libs/cardano-ledger-core/src/Cardano/Ledger/Core/PParams.hs
TimSheard: Where they belong

CHANGELOG.md

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Newpp.hs
TimSheard: done

eras/shelley/test-suite/src/Test/Cardano/Ledger/Shelley/Generator/Update.hs
TimSheard: done

eras/shelley/test-suite/src/Test/Cardano/Ledger/Shelley/Generator/Update.hs
TimSheard: done

eras/shelley/test-suite/src/Test/Cardano/Ledger/Shelley/Generator/Update.hs
TimSheard: done

lehins (approved):
Awesome. Thank you!

PR #3954, opened 2023-12-22
Fix how we update `CostModels`

lehins:
# Description

This is a follow up to #3953 

There are a couple of problems with CostModels in Conway:

* first one is that we retain invalid and unknown CostModels indefinitely. This PR clears them out as soon as a valid replacement has been added
* ToJSON/FromJSON did not account for unknown and invalid CostModels, thus information was lost during roundtrip and there was no way for a user to see this very useful information
* Constructor for `CostMododels` was exported thus it was very easy to invalidate invariant that CBOR and JSON instances rely on. Creation and modification of `CostModels` is done with smart constructors and through `Monoid` instance

Besides fixing CostModels this PR also adds more tests and organizes a little bit by removing duplicate functionality

# Checklist

- [x] Self-reviewed the diff


libs/cardano-ledger-core/src/Cardano/Ledger/Plutus/CostModels.hs
```diff
@@ -286,18 +297,47 @@ decodeCostModelFailHard lang = do
...
+        P.CMUnknownParamError {} -> 0
+        P.CMInternalReadError {} -> 1
+        P.CMInternalWriteError {} -> 2
+        P.CMTooFewParamsError {} -> 3
+
+instance EncCBOR CostModelError where
+  encCBOR =
+    encode . \case
+      CostModelError (P.CMUnknownParamError err) ->
+        Sum (CostModelError . P.CMUnknownParamError) 1 !> To err
```
teodanciu: Why start at 1 and not at 0 ?

libs/cardano-ledger-core/testlib/Test/Cardano/Ledger/Plutus.hs
```diff
@@ -30,6 +31,14 @@ import qualified PlutusLedgerApi.Test.V2.EvaluationContext as PV2
 import qualified PlutusLedgerApi.Test.V3.EvaluationContext as PV3
 import PlutusLedgerApi.V1 as PV1
 
+-- | Construct a test cost model where all paramters are set to the same value
```
teodanciu: Small typo `parameters`

libs/cardano-ledger-core/src/Cardano/Ledger/Plutus/CostModels.hs
```diff
@@ -37,7 +37,12 @@ module Cardano.Ledger.Plutus.CostModels (
   costModelToMap,
   costModelFromMap,
   decodeCostModelFailHard,
-  CostModels (..),
+  CostModels,
```
teodanciu, line 40: Maybe this will be easier to find if we move it up after `CostModel` export ?

eras/alonzo/impl/testlib/Test/Cardano/Ledger/Alonzo/Arbitrary.hs
```diff
@@ -253,23 +253,22 @@ genValidCostModel lang = do
   either (\err -> error $ "Corrupt cost model: " ++ show err) pure $
     mkCostModel lang newParamValues
 
-genValidCostModelPair :: Language -> Gen (Language, CostModel)
-genValidCostModelPair lang = (,) lang <$> genValidCostModel lang
+genValidCostModels :: Set.Set Language -> Gen CostModels
+genValidCostModels = fmap mkCostModels . sequence . Map.fromSet genValidCostModel
```
teodanciu: This is not strictly related to this change, but since we removed one useless function (genValidCostModelPair), maybe we could also remove its exact copy a few lines above, `genCostModel`, or at least reduce it's scope, since it's used only in [one place](https://github.com/IntersectMBO/cardano-ledger/pull/3954/commits/9d82ccdab94d04e52aa2bc7de35bc398b437624e#diff-334ba93db40dbbd431eb2831c9fe3226881e21c81945dc0dd3551ca9f215b62dL248), in the Arbitrary instance of Costmodel. 
Just because it's polluting a little the outline of the module that has become much clearer with this PR. And also it's kinda confusing to have `genValidCostModel` and `genCostModel`, whereas both generate valid ones, but in a different form.  But just an idea, it doesn't make much difference

eras/alonzo/impl/testlib/Test/Cardano/Ledger/Alonzo/Binary/CostModelsSpec.hs
```diff
@@ -9,30 +10,43 @@
...
-spec pxy = do
-  describe "CostModels CBOR deserialization" $ do
-    validCostModelProp @era pxy
+spec :: forall era. AlonzoEraPParams era => Spec
+spec = do
+  describe "CBOR deserialization" $ do
+    validCostModelProp @era Proxy
     invalidCostModelProp @era Proxy
     unknownCostModelProp @era Proxy
+  prop "applyPPUpdates" $ \valid validUpdate unknown unknownUpdate -> do
```
teodanciu, line 33: The formatting looks a bit strange here, don't know why fourmolu doesn't pick it up, but it looks like it should be aligned with the line above `unKnownCostModelProp @era Proxy`, no?

eras/alonzo/impl/CHANGELOG.md
```diff
@@ -53,6 +53,7 @@
 
 ### `testlib`
 
+* Remove `Test.Cardano.Ledger.Alonzo.CostModel` in favor of `Test.Cardano.Ledger.Plutus`
```
teodanciu, line 56: Maybe we should also mention the removal of `costModelParamsCount` from Arbitrary

libs/cardano-ledger-core/src/Cardano/Ledger/Plutus/CostModels.hs
```diff
@@ -355,63 +360,77 @@ instance NFData CostModelError
...
-    Just lang ->
-      case mkCostModel lang cmIds of
-        Right cm -> CostModels (Map.insert lang cm validCMs) errs invalidCMs
-        Left e -> CostModels validCMs (addError lang e errs) updatedInvalidCMs
-    Nothing -> CostModels validCMs errs updatedInvalidCMs
+updateCostModels (CostModels oldValid oldErrors oldUnk) (CostModels modValid modErrors modUnk) =
+  CostModels
+    newValid
+    (newErrors Map.\\ newValid)
+    (Map.union modUnk oldUnk Map.\\ Map.mapKeys languageToWord8 newValid)
```
teodanciu, line 422: Nice, so this keeps the property that a Language is either in the valid or errors map that we have when using `mkCostModelsLenient`.   The resulting unknown costmodels  could still have keys that are in `errors`  (which is not possible in `mkCostModelsLenient`), but  don't think it matters.

eras/alonzo/impl/testlib/Test/Cardano/Ledger/Alonzo/Binary/CostModelsSpec.hs
lehins: No, this was intentional, since this property has nothing to do with serialization. It is just part of the `CostModels` `Spec`

libs/cardano-ledger-core/src/Cardano/Ledger/Plutus/CostModels.hs
```diff
@@ -366,6 +367,12 @@ data CostModels = CostModels
   }
   deriving stock (Eq, Ord, Show, Generic)
 
+instance Semigroup CostModels where
+  (<>) = updateCostModels
```
teodanciu, line 391: Why not use it in [PParams.ppUpdateCostModels](https://github.com/IntersectMBO/cardano-ledger/blob/lehins/cost-models-fixes/eras/conway/impl/src/Cardano/Ledger/Conway/PParams.hs#L1192)   ?
`SJust costModelUpdate -> THKD $ curCostModel <> costModelUpdate`   

Any particular reason?

eras/alonzo/impl/testlib/Test/Cardano/Ledger/Alonzo/Binary/CostModelsSpec.hs
teodanciu: Oh yes, makes sense!  Missed the content of "describe"

eras/alonzo/impl/testlib/Test/Cardano/Ledger/Alonzo/Binary/CostModelsSpec.hs
```diff
@@ -9,30 +10,43 @@
...
+    let
+      validExpected = mkCostModels (costModelsValid validUpdate <> costModelsValid valid)
+      unknownExpected = unknownUpdate <> unknown
```
teodanciu, line 36: I suppose the chance of the generated `unknown` to be actually a valid one can be neglected

eras/alonzo/test-suite/test/Test/Cardano/Ledger/Alonzo/Golden.hs
```diff
@@ -17,21 +17,24 @@ import Cardano.Ledger.Alonzo.PParams (
...
-  mkCostModel,
- )
 import Cardano.Ledger.Alonzo.TxBody (AlonzoTxOut (..), utxoEntrySize)
 import Cardano.Ledger.BaseTypes (StrictMaybe (..), boundRational)
 import Cardano.Ledger.Binary (decCBOR, decodeFullAnnotator)
 import Cardano.Ledger.Binary.Plain as Plain (serialize)
 import Cardano.Ledger.Block (Block (..))
 import Cardano.Ledger.Coin (Coin (..))
 import Cardano.Ledger.Mary.Value (valueFromList)
+import Cardano.Ledger.Plutus.CostModels (
```
teodanciu, line 27: Shouldn't we update the dependency on cardano-ledger-core  in [cardano-ledger-alonzo-test.cabal ](https://github.com/IntersectMBO/cardano-ledger/blob/lehins/cost-models-fixes/eras/alonzo/test-suite/cardano-ledger-alonzo-test.cabal#L49) to 

` cardano-ledger-core:{cardano-ledger-core, testlib} >=1.10 && <1.11`  ?

eras/alonzo/impl/testlib/Test/Cardano/Ledger/Alonzo/Binary/CostModelsSpec.hs
lehins: On the contrary, it is desired. I actually found a bug in my implementation earlier thanks to the fact that we generate unknown updates that can be valid

libs/cardano-ledger-core/src/Cardano/Ledger/Plutus/CostModels.hs
lehins: It'll actually be better if I'll organize functions and type relative to `CostModel` vs `CostModels`. So'll create a section for `CostModels` and will move this export to the top of that section.

eras/alonzo/impl/testlib/Test/Cardano/Ledger/Alonzo/Arbitrary.hs
lehins: Great catch. Done

eras/alonzo/impl/CHANGELOG.md
lehins: Done.

libs/cardano-ledger-core/src/Cardano/Ledger/Plutus/CostModels.hs
lehins: No reason. It's not really worth a change though.

eras/alonzo/test-suite/test/Test/Cardano/Ledger/Alonzo/Golden.hs
lehins: Yes. Nice catch. Done.

libs/cardano-ledger-core/src/Cardano/Ledger/Plutus/CostModels.hs
lehins: > The resulting unknown costmodels could still have keys that are in errors (which is not possible in mkCostModelsLenient), but don't think it matters.

That is actually necessary for rountripping. If we were to remove costmodels that caused errors from unknown then the errors field would be lost, since we do not serialize errors.

libs/cardano-ledger-core/testlib/Test/Cardano/Ledger/Plutus.hs
lehins: Fixed

teodanciu (approved):
Looks good to me

eras/alonzo/impl/testlib/Test/Cardano/Ledger/Alonzo/Binary/CostModelsSpec.hs
teodanciu: I see, because we're flattening them after. Makes sense.

libs/cardano-ledger-core/src/Cardano/Ledger/Plutus/CostModels.hs
lehins: Funny you say that. I wrote it initially starting with `0`, naturally. Then I looked at serialization of some predicate failure, which started at `1` and I thought to myself: "Oh, I guess we start at `1` for those decoders", so I switched it to start at `1` for "consistency". But now I can't even find that predicate failure :smile: 

I've switched to start at 0, thank you for bringing this up!

PR #3956, opened 2023-12-27
Allow 128 byte Url and DnsNames for decoder version atleast 9.

### LINKED ISSUE #3941: Increase the byte length limit for Url
It has been decided to increase the length limit for `Url` type from 64 bytes to 128 bytes. We can't really go much higher than 128 bytes because that would be very disruptive for hardware wallet engineers.

This limit should be increased at the deserialization level starting with protocol version 9

A new QuickCheck generator also needs to be implemented for the roundtrip tests and CDDL needs to be adjusted for Conway for Anchors and StakePool params.

TimSheard:
Make it possible to have Urls and DnsNames contain at most 128 bytes with dcoder versions at least 9
Add max byte size as extra parameters to functions  textToDns, and textToUrl,
Changed calls to "textToUrl xxx"  to   "textToUrl 64 xxx".  Same for textToDns.

Fixes #3941 

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/libs/cardano-ledger-core/src/Cardano/Ledger/BaseTypes.hs
+++ b/libs/cardano-ledger-core/src/Cardano/Ledger/BaseTypes.hs
@@ -534,7 +534,7 @@ textSizeN :: MonadFail m => Int -> Text -> m Text
 textSizeN n t =
   if BS.length (encodeUtf8 t) <= n
     then pure t
-    else fail $ ("Text exceeds " ++ show n ++ " bytes:" ++ show t)
+    else fail $ "Text exceeds " ++ show n ++ " bytes:" ++ show t
 
 textDecCBOR :: Int -> Decoder s Text
 textDecCBOR n = decCBOR >>= textSizeN n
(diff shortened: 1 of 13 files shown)


libs/cardano-ledger-core/src/Cardano/Ledger/BaseTypes.hs
```diff
@@ -526,17 +527,25 @@ a ==> b = not a || b
...
 
-text64 :: MonadFail m => Text -> m Text
-text64 t =
-  if BS.length (encodeUtf8 t) <= 64
+textSizeN :: MonadFail m => Int -> Text -> m Text
+textSizeN n t =
+  if BS.length (encodeUtf8 t) <= n
     then pure t
-    else fail $ "Text exceeds 64 bytes:" ++ show t
+    else fail $ ("Text exceeds " ++ show n ++ " bytes:" ++ show t)
```
lehins:
```suggestion
    else fail $ "Text exceeds " ++ show n ++ " bytes:" ++ show t
```

lehins (changes requested):
We also need to adjust the new limit in CDDL:

* here https://github.com/IntersectMBO/cardano-ledger/blob/8e281ac3372f996eff4a19e241526eebbf96a0eb/eras/conway/impl/cddl-files/conway.cddl#L373

* and here: https://github.com/IntersectMBO/cardano-ledger/blob/8e281ac3372f996eff4a19e241526eebbf96a0eb/eras/conway/impl/cddl-files/conway.cddl#L352

lehins (approved):
LGTM. Thank you.

PR #3960, opened 2023-12-30
Plutus script purpose

### LINKED ISSUE #3533: Add ScriptPurpose to Conway
Conway requires two new ScriptPurposes:

* VotingPurpose (depends on #3542)
* ProposalPurpose (needed for runnable constitution, i.e. GovernancePolicy)

Related PR #3480

### LINKED ISSUE #3663: Implement redeemer pointers for new scripts
`Tag` in `RdmrPtr` need to add two new tags: one for voting another for proposing.

~The easiest way is to add two new constructors to the existing type and prevent new tags when decoding before conway with versioned decoder.~

A more type safe solution would be to convert `RdmrPtr` into a type family and add it to new `AlonzoEraScript` type class.

lehins:
# Description

Fixes #3663 
Fixes #3533 

Introduce PlutusPurpose:

ScriptPurpuse and RdmrPtr are very much related concepts. This commit
replaces them with a single type family that is parametric on the era
and on the type that does the selection whether it is a plutus script
purpose or it is a redeemer pointer. Those types are called `AsItem` and
`AsIndex` respectively.

Besides making plutus script purpose parametric this commit also
introduces new purposes for Conway: `Voting` and `Proposing`

# Checklist

- [x] Self-reviewed the diff


eras/conway/impl/cddl-files/conway.cddl
```diff
@@ -469,13 +469,19 @@ constr<a> =
   ; similarly for tag range: 6.1280 .. 6.1400 inclusive
   / #6.102([uint, [* a]])
 
-redeemer = [ tag: redeemer_tag, index: uint, data: plutus_data, ex_units: ex_units ]
+redeemers =
+  [ + [ tag: redeemer_tag, index: uint, data: plutus_data, ex_units: ex_units ] ]
+; TODO: Add alternative implementation that reflects the reality more accuratly:
```
Soupstraw, line 474: Make sure there's an issue for this

lehins: Thank you for keeping me honest :wink: 
There is already one #3922

eras/alonzo/impl/CHANGELOG.md
```diff
@@ -2,6 +2,18 @@
 
 ## 1.6.0.0
 
+* Deprecated `getAlonzoSpendingTxIn` in favor of `plutusPurposeSpendingTxIn`
+* Remove `rdptr` in favor of new `redeemerPointer`
```
aniketd: Reads like a duplicate of line 14.

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxBody.hs
```diff
@@ -604,3 +640,71 @@ instance
...
+  forall era.
+  MaryEraTxBody era =>
+  TxBody era ->
+  AlonzoPlutusPurpose AsItem era ->
+  StrictMaybe (AlonzoPlutusPurpose AsIndex era)
+alonzoRedeemerPointer txBody = \case
+  AlonzoSpending txIn ->
+    AlonzoSpending <$> indexOf txIn (txBody ^. inputsTxBodyL)
+  AlonzoMinting policyID ->
+    AlonzoMinting <$> indexOf policyID (txBody ^. mintedTxBodyF :: Set (PolicyID (EraCrypto era)))
```
Soupstraw, line 654: Should we switch the type of `mintedTxBody` to be an `OSet`?

lehins: That would be great, but we can't do that right now, unfortunately
We would have to change `mintTxBodyF` to an `OMap` first.

eras/alonzo/impl/CHANGELOG.md
lehins: It is indeed a duplicate. Thanks.

Soupstraw (approved):
LGTM

PR #3961, opened 2024-01-02
Switch ContextError to an injective type family from a data family

lehins:
# Description

This change is necessary because we can not use the initial approach of downgrading type families to previous eras. If we don't switch to type family from data family we can not use type families in the errors (eg. `TxCert`)

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/eras/babbage/test-suite/src/Test/Cardano/Ledger/Babbage/TxInfo.hs
+++ b/eras/babbage/test-suite/src/Test/Cardano/Ledger/Babbage/TxInfo.hs
@@ -9,16 +9,14 @@
 module Test.Cardano.Ledger.Babbage.TxInfo where
 
 import Cardano.Ledger.Address (Addr (..), BootstrapAddress (..))
-import Cardano.Ledger.Alonzo (AlonzoEra)
 import Cardano.Ledger.Alonzo.Plutus.Context (
   ContextError,
   EraPlutusTxInfo (toPlutusTxInfo),
   PlutusTxInfo,
  )
-import Cardano.Ledger.Alonzo.Plutus.TxInfo (TxOutSource (..))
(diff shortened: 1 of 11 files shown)


eras/babbage/test-suite/src/Test/Cardano/Ledger/Babbage/TxInfo.hs
```diff
@@ -271,8 +267,8 @@ txInfoTestsV1 ::
   , BabbageEraTxBody era
   , Value era ~ MaryValue (EraCrypto era)
   , EraPlutusTxInfo 'PlutusV1 era
-  , Inject (ContextError (AlonzoEra (EraCrypto era))) (ContextError era)
-  , Inject (ContextError (BabbageEra (EraCrypto era))) (ContextError era)
+  , Inject (AlonzoContextError era) (ContextError era)
+  , Inject (BabbageContextError era) (ContextError era)
```
Soupstraw: Instead of having two constraints, we could use the `AlonzoContextError` constructor below to get rid of the first constraint. This would make the types a bit nicer at the cost of making the implementation a tiny bit noisier, but I'm also fine with keeping the two constraints.

Soupstraw (approved):
LGTM

eras/babbage/test-suite/src/Test/Cardano/Ledger/Babbage/TxInfo.hs
lehins: I did exactly that in the Babbage era. I was too lazy and didn't want to do it here :grin: 

I'll change, since you brought it up.

lehins: Done

PR #3962, opened 2024-01-02
Add `estimateMinFeeTx`

lehins:
# Description

Current version of the function `evaluateTransactionFee` does not handle Byron witnesses and it is not recursive, so it cannot account for the fact that occasionally the change to the fee will affects the size of the transaction and thus the minimum fee. New `estomateMinFeeTx` improves this logic

Besides adding this new functionality to ledger API there are a few minor additions:

* Addition of `forceDRepPulsingState`, which can be used to memoize the result of a premature completion of DRep pulser
* Addition of `Arbitrary` instances for bootstrap addresses and `ByronKeyPair` type that can be used to test functionality related to Byron addresses and witnesses
* Adds Enc/DecCBOR instances for `SignedDSIGN`

# Checklist

- [x] Self-reviewed the diff


eras/conway/impl/src/Cardano/Ledger/Conway/Governance.hs
```diff
@@ -1022,6 +1023,13 @@ finishDRepPulser (DRPulsing (DRepPulser {..})) =
         }
     !ratifyState' = runConwayRatify dpGlobals ratifyEnv ratifyState ratifySig
 
+-- | Force computation of DRep stake distribution and figure out the next enact
+-- state. This operation is useful in cases when access to new EnactState or DRep stake
+-- distribution is needed moer than once. It is safe to call this function at any
+-- point. Whenever pulser is already in computed state this will be a noop.
+forceDRepPulsingState :: ConwayEraGov era => NewEpochState era -> NewEpochState era
+forceDRepPulsingState nes = nes & newEpochStateDRepPulsingStateL %~ completeDRepPulsingState
+
```
Soupstraw, line 1032: I think we should add a property test that checks that a pulser is always completed after applying `completeDRepPulsingState`

Soupstraw (approved):
LGTM

libs/cardano-ledger-api/src/Cardano/Ledger/Api/Tx.hs
```diff
@@ -71,4 +78,76 @@ import Cardano.Ledger.Api.Tx.AuxData
...
+-- @since 1.8.0
+estimateMinFeeTx ::
+  forall era.
+  EraTx era =>
+  -- | The current protocol parameters.
+  PParams era ->
+  -- | The transaction.
+  Tx era ->
+  -- | The number of key witnesses still to be added to the transaction.
+  Int ->
```
teodanciu, line 107: How come this is changed to `Int`? Wouldn't `Word` (as it is in the deprecated function) be a better fit for this type, since it's unsigned?

eras/conway/impl/src/Cardano/Ledger/Conway/Governance.hs
lehins: I think it is obvious from implementation that this property will hold. Moreover this is a helper function for downstream users and ledger has no use for it, so I am not too worried about it

Soupstraw:
We should also use this in the fixup function of `ImpTest`

libs/cardano-ledger-api/src/Cardano/Ledger/Api/Tx.hs
lehins: What do you think will happen if I supply `-1` to such a function? :smile: Nothing, it will work the same.

It this example a `-1` would overflow  to `18446744073709551615`, but then due to conversion back to Int it would overflow again to `-1`, however that is the consequence of the implementation. If there was a manual loop or a enumeration involved it would easily overflow.

We should always use `Int` instead of `Word` in such scenarios. This is the same reason why for example `replicate` accepts an `Int`, instead of a `Word`. Because accidental overflow could result in non-termination.

It could be beneficial to use `Natural`. Problem with it is the error reporting is terrible with it.

teodanciu: Oh interesting. Thanks for the thorough explanation

teodanciu (approved):
Looks good to me

PR #3963, opened 2024-01-03
Abstract `getWitsVKeyNeeded`

### LINKED ISSUE #3485: Need for `ConwayUtxoEnv`
`UtxoEnv` is defined as:

```haskell
data UtxoEnv era
  = UtxoEnv
      SlotNo
      (PParams era)
      (CertState era)
      (GenDelegs (EraCrypto era))
```

However, `GenDelegs` is no longer a thing in Conway. Moreover, all three utxo rules require different arguments, but the same environment is used.

So, we need three different environments types, with one new one for Conway's UTXOW rule that can be used in all of the UTXO[W|S] rules:

```haskell
data ShelleyUtxoEnv era = ShelleyUtxoEnv
  { sUtxoEnvPParams :: !(PParams era)
  , sUtxoEnvCertState :: !(CertState era)
  }
data ShelleyUtxosEnv era = ShelleyUtxosEnv
  { sUtxosEnvSlotNo :: !SlotNo
  , sUtxosEnvPParams :: !(PParams era)
  }
data ShelleyUtxowEnv era = ShelleyUtxowEnv
  { sUtxowEnvSlotNo :: !SlotNo
  , sUtxowEnvPParams :: !(PParams era)
  , sUtxowEnvCertState :: !(CertState era)
  , sUtxowEnvGenDelegs :: !(GenDelegs (EraCrypto era))
  }
data ConwayUtxowEnv era = ConwayUtxowEnv
  { cUtxowEnvSlotNo :: !SlotNo
  , cUtxowEnvPParams :: !(PParams era)
  , cUtxowEnvCertState :: !(CertState era)
  }
```

lehins:
# Description

Turns out `GenDelegs` is totally unnecessary in the `UtxoEnv`. This PR removes it.

Also it adds an era agnostic `getWitsVKeyNeeded` that we can now use in UTXOW rules for all eras.

This is a nice organization and simplification, but the real reason for this change is that we need to provide a proper min fee computation for CLI, which will be implemented in a subsequent PR

Fixes #3485 

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Rules/Ledger.hs
+++ b/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Rules/Ledger.hs
@@ -24,7 +24,6 @@ import Cardano.Ledger.Keys (DSignable, Hash)
 import Cardano.Ledger.Shelley.Core
 import Cardano.Ledger.Shelley.LedgerState (
   CertState (..),
-  DState (..),
   LedgerState (..),
   UTxOState (..),
  )
@@ -79,31 +78,28 @@ ledgerTransition ::
   ) =>
   TransitionRule (someLEDGER era)
 ledgerTransition = do
-  TRC (LedgerEnv slot txIx pp account, LedgerState utxoSt dpstate, tx) <- judgmentContext
+  TRC (LedgerEnv slot txIx pp account, LedgerState utxoSt certState, tx) <- judgmentContext
   let txBody = tx ^. bodyTxL
 
-  dpstate' <-
+  certState' <-
     if tx ^. isValidTxL == IsValid True
       then
         trans @(EraRule "DELEGS" era) $
           TRC
             ( DelegsEnv slot txIx pp tx account
-            , dpstate
+            , certState
             , StrictSeq.fromStrict $ txBody ^. certsTxBodyL
             )
-      else pure dpstate
-
-  let CertState _ _pstate dstate = dpstate
-      genDelegs = dsGenDelegs dstate
+      else pure certState
 
   utxoSt' <-
     trans @(EraRule "UTXOW" era) $
       TRC
-        ( UtxoEnv @era slot pp dpstate genDelegs
+        ( UtxoEnv @era slot pp certState
         , utxoSt
         , tx
         )
-  pure $ LedgerState utxoSt' dpstate'
+  pure $ LedgerState utxoSt' certState'
 
 instance
   ( DSignable (EraCrypto era) (Hash (EraCrypto era) EraIndependentTxBody)
--- a/eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Utxow.hs
+++ b/eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Utxow.hs
@@ -49,6 +49,7 @@ import Cardano.Ledger.BaseTypes (
   (==>),
  )
 import Cardano.Ledger.Binary (DecCBOR (..), EncCBOR (..), decodeRecordSum, encodeListLen)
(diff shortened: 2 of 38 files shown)


eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Utxow.hs
```diff
@@ -598,8 +551,7 @@ instance
   where
   inject = UtxoFailure
 
--- | Calculate the set of hash keys of the required witnesses for update
--- proposals.
+-- | Deprecated.
```
teodanciu, line 556: Any reason we don't deprecate it with {- DEPRECATED #-}  ? 
I see it's only called from the two now deprecated functions, `propWits` and `witsVKeyNeededGov`

lehins: There is no point. It is not an exported function. DEPRECATED pragma is for letting downstream users know that a function will disappear soon and they should act accordingly. Once deprecated functions in this module that use this `proposedUpdatesWitnesses` are removed then this function will result in an unused function warning and can be removed as well. So, it is enough to leave a comment that this is truly an unused function.

eras/shelley/impl/testlib/Test/Cardano/Ledger/Shelley/ImpTest.hs
```diff
@@ -356,20 +349,15 @@ instance
...
-  , EraTx era
-  , ShelleyEraTxBody era
-  ) =>
+impWitsVKeyNeeded ::
+  EraUTxO era =>
   NewEpochState era ->
   TxBody era ->
   Set.Set (KeyHash 'Witness (EraCrypto era))
-shelleyImpWitsVKeyNeeded nes txb = shelleyWitsVKeyNeeded utxo txb genDelegs
+impWitsVKeyNeeded nes txBody = getWitsVKeyNeeded certState utxo txBody
```
teodanciu, line 357: I wonder if this function is even necessary, it seems like it served the same purpose as the era-agnostic `getWitsVKeyNeeded` and doesn't do much besides lens-ing. 
But it's not important for this PR, just a thought.

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Rules/Ledger.hs
```diff
@@ -79,31 +78,28 @@ ledgerTransition ::
   ) =>
   TransitionRule (someLEDGER era)
 ledgerTransition = do
-  TRC (LedgerEnv slot txIx pp account, LedgerState utxoSt dpstate, tx) <- judgmentContext
+  TRC (LedgerEnv slot txIx pp account, LedgerState utxoSt certState, tx) <- judgmentContext
```
teodanciu, line 81: Poor certState, it seems to be misnamed everywhere 😆

eras/shelley/impl/testlib/Test/Cardano/Ledger/Shelley/ImpTest.hs
lehins: Yeah, I didn't really check how it is used if at all. From a quick grep it indeed looks like we can get rid of it

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Utxow.hs
teodanciu: Aha, makes sense.

teodanciu (approved):
Looks good to me.
Love how the second commit cleans up so much noise, awesome.

PR #3964, opened 2024-01-03
Ignore thunks in `utxosDeposited`

Soupstraw:
# Description

This fixes the "thunks present" failures in the nightly tests. We allow thunks in the `utxosDeposited` field because we only use that field in tests. 

# Checklist

- [x] Self-reviewed the diff


teodanciu (approved):
Awesome that it fixes the NoThunks test, I tried without success in the past, probably I missed some bangs :laughing: 
Looks good to me

PR #3971, opened 2024-01-04
ToJson instances for context error

lehins:
# Description

This PR transfers `ToJSON` instance for `CollectError` failure type and adds `ToJSON` instances for other types that are needed for above instance

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Plutus/Evaluate.hs
+++ b/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Plutus/Evaluate.hs
@@ -5,6 +5,7 @@
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE PatternSynonyms #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE StandaloneDeriving #-}
 {-# LANGUAGE TypeApplications #-}
@@ -26,7 +27,7 @@ import Cardano.Ledger.Alonzo.Plutus.Context (EraPlutusContext (..))
 import Cardano.Ledger.Alonzo.Scripts (plutusScriptLanguage)
 import Cardano.Ledger.Alonzo.Tx (ScriptPurpose (..), indexedRdmrs)
 import Cardano.Ledger.Alonzo.UTxO (AlonzoEraUTxO (getSpendingDatum), AlonzoScriptsNeeded (..))
-import Cardano.Ledger.BaseTypes (ProtVer (pvMajor), natVersion)
+import Cardano.Ledger.BaseTypes (ProtVer (pvMajor), kindObject, natVersion)
 import Cardano.Ledger.Binary (DecCBOR (..), EncCBOR (..))
 import Cardano.Ledger.Binary.Coders
 import Cardano.Ledger.Plutus.CostModels (costModelsValid)
@@ -42,6 +43,7 @@ import Cardano.Ledger.UTxO (EraUTxO (..), ScriptsProvided (..), UTxO (..))
 import Cardano.Slotting.EpochInfo (EpochInfo)
 import Cardano.Slotting.Time (SystemStart)
 import Control.Monad (guard)
+import Data.Aeson (ToJSON (..), (.=), pattern String)
 import Data.List (intercalate)
 import qualified Data.Map.Strict as Map
 import Data.Maybe (mapMaybe)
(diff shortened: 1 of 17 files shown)


libs/cardano-ledger-core/src/Cardano/Ledger/BaseTypes.hs
```diff
@@ -853,3 +855,8 @@ class Inject t s where
   inject :: t -> s
   default inject :: t ~ s => t -> s
   inject = id
+
+-- | Helper function for a common pattern of creating objects
+kindObject :: Text -> [Pair] -> Value
+kindObject name obj =
+  object $ ["kind" .= name] ++ obj
```
teodanciu: Could you use `("kind" .= name) : obj`  here  instead of making a list and concatenating it?

libs/cardano-ledger-core/src/Cardano/Ledger/Core/TxCert.hs
```diff
@@ -152,6 +156,19 @@ instance NoThunks (PoolCert c)
 instance NFData (PoolCert c) where
   rnf = rwhnf
 
+instance Crypto c => ToJSON (PoolCert c) where
+  toJSON = \case
+    RegPool poolParams ->
+      object
```
teodanciu: We could use 
`kindObject "RegPool" [ "poolParams" .= toJSON poolParams]`  here  I think (not that it makes a difference)

libs/cardano-ledger-core/src/Cardano/Ledger/Core/TxCert.hs
```diff
@@ -152,6 +156,19 @@ instance NoThunks (PoolCert c)
...
+        [ "kind" .= String "RegPool"
+        , "poolParams" .= toJSON poolParams
+        ]
+    RetirePool poolId epochNo ->
+      object
+        [ "kind" .= toJSON poolId
+        , "poolParams" .= toJSON epochNo
```
teodanciu: Is it "epochNo" here instead of "poolParams"  ?

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Plutus/Evaluate.hs
```diff
@@ -91,6 +93,38 @@ instance (EraTxCert era, DecCBOR (ContextError era)) => DecCBOR (CollectError er
...
+      object
+        [ "kind" .= String "CollectError"
+        , "error" .= String "NoCostModel"
+        , "language" .= toJSON lang
+        ]
+    BadTranslation err ->
+      object
+        [ "kind" .= String "BadTranslation"
+        , "error" .= toJSON err
+        ]
```
teodanciu: We could use `kindObject` in these expressions, though probably it's the same number of characters in the end.

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxWits.hs
```diff
@@ -153,6 +155,29 @@ instance EncCBORGroup RdmrPtr where
...
+        [ "kind" .= String "RedeemerPointerMinting"
+        , "value" .= n
+        ]
+    RdmrPtr Cert n ->
+      object
+        [ "kind" .= String "RedeemerPointerCertifying"
+        , "value" .= n
+        ]
+    RdmrPtr Rewrd n ->
+      object
```
teodanciu: Same observation about `kindObject` as before

teodanciu (approved):
Looks good to me,  added a few small remarks.

lehins:
@teodanciu thank you for catching all those deficiencies in the PR. All fixed.

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxWits.hs
lehins: That was the reason why I added the function `kindObject`. That and also the fact that I like the name :smile:

libs/cardano-ledger-core/src/Cardano/Ledger/BaseTypes.hs
lehins: Yes. Nice catch

libs/cardano-ledger-core/src/Cardano/Ledger/Core/TxCert.hs
lehins: Yes. Done

libs/cardano-ledger-core/src/Cardano/Ledger/Core/TxCert.hs
lehins: Yes, also  `"kind"` should be set to `"RetirePool"`, not `poolId`.

Done

PR #3973, opened 2024-01-05
Move functionality of Cardano.Ledger.Pretty  to Test.Cardano.Ledger.Generic.PrettyCore

### LINKED ISSUE #3926: Get rid of  `cardano-ledger-pretty`
After a bit of investigation it looks like no one is using `cardano-ledger-pretty`. Originally the goal was to use it in testing, however `tree-diff` package proved to be a lot more useful for tests.

This means that `cardano-ledger-pretty` should be ripped out, because currently it is just an unnecessary overhead.

There are likely some uses of the package in testing, all of which should be replaced with `ToExpr`

### LINKED ISSUE #3974: Fix Pretty instance for TxBody
I'm running the `UTXO` rule with the following transaction:
```
sig = Tx
       { body =
           TxBody (Conway)
            { spending inputs =
                {(TxIn
                    (TxId
                       #"7c28e0aa2650644cb45205d4d96e304f59087e4aed421d587afd3dc704c10369")
                    3)
                ,(TxIn
                    (TxId
                       #"d08a4a34021c0dc43a92f9719fbf0caefdfdf2cd4b34490f323843decc3bd2d6")
                    2)}
            , collateral inputs =
                {(TxIn
                    (TxId
                       #"63616843c217fa17c3e37f93c5fb37de703e2e35ce010471663c170b13dcc789")
                    1)}
            , reference inputs =
                {(TxIn
                    (TxId
                       #"67fa2779c284d840793d822d07911659507c28e1749373417d8d349f9097118a")
                    2)}
            , outputs =
                [TxOut
                  { address =
                      (BootstrapAddress
                        { root =
                            00010100010101010001010001010100000001000100000000000101
                        , derivationpath = ?-
                        , networkmagic = Testnet 0
                        , remain = {  }
                        , type = ATVerKey })
                  , value = (Value (Coin 6) [])
                  , datum = NoDatum
                  , reference script = ?- }]
            , collateral return =
                ?TxOut
                  { address =
                      (BootstrapAddress
                        { root =
                            00010001010101010000000000000000000000010001000001000000
                        , derivationpath = ?-
                        , networkmagic = MainOrStage
                        , remain = {  }
                        , type = ATRedeem })
                  , value = (Value (Coin 0) [])
                  , datum =
                      (DatumHash
                         #"88c4284a47a355c219a07a2101b3bae19ddad3e35e4d780856a8bb7e55dae14f")
                  , reference script =
                      (Expires SlotNo 1) (ScriptHash
                                            #"a0d442ce189bc932515c85be74ac939df19ea8450c03030234790921") }
            , total collateral = ?(Coin 0)
            , withdrawals =
                (Wdr
                  { RewardAcnt
                     { network = Mainnet
                     , credential =
                         (ScriptCred
                            #"9f8a6975b82da2142fed134f53033dae8ed8e8fa39c4a78ddc92470c") } ->
                      (Coin 0) })
            , transaction fee = (Coin 1)
            , validity interval =
                ValidityInterval
                 { invalidBefore = ?- , invalidHereafter = ?SlotNo 4 }
            , required signer hashes = {}
            , mint = (MultiAsset [])
            , script integrity hash = ?-
            , auxiliary data hash = ?-
            , network id = ?-
            , voting procedures =
                Map
                 { CommitteeVoter (KeyHashObj (KeyHash "ab2070615b033705336abb07db4d9a4655631fbc0aaadbfee6388607")) ->
                     Map{  } }
            , proposal procedures = [] }
       , wits =
           AlonzoTxWits
            { keys = {}
            , bootstrap witnesses =
                {BootstrapWitness
                  { key =
                      VerKeyEd25519DSIGN "0d7fa3a9f7e3d76a478ec595906cc5c5b013ac696fb3eea779379b986674e62b"
                      hash =  (KeyHash
                                 #"a6188767df7b905758c8a131b8c603573d83b8b70f09097b77872853")
                  , signature =
                      SignedDSIGN (SigEd25519DSIGN "7572cfc1597c5fe74c527fcabd69a6b12e7c14dbe7e7dda3fb530d24adc0f7f729e8e91eb814e81dd0eb
25ed7a56efb9db742b8be4514e26fda21cbd6075b51c")
                  , chaincode = *17l6zly
                  , attributes = *17l6zly }}
            , scripts map = Map{  }
            , Data map = Map{  }
            , Redeemer map = Map{  } }
       , isValid = True
       , auxiliaryData = ?- }
```
In the following utxo set:
```
utxo =
          UTxO
           { (TxIn
                (TxId
                   #"7c28e0aa2650644cb45205d4d96e304f59087e4aed421d587afd3dc704c10369")
                3) ->
               TxOut
                { address =
                    (Addr
                       Testnet
                       (KeyCred
                          #"ee9905b1cc8cf873ab88b01aa2693dff73a300a7e722da403bff6603")
                       (BaseRef
                          (KeyCred
                             #"60c2b8115e33eea7c8d536a3678bea1012cdfc960678a48fd8dbedad")))
                , value = (Value (Coin 3) [])
                , datum = NoDatum
                , reference script =
                    (Expires SlotNo 3) (ScriptHash
                                          #"eefa211d43eb53ecc9c82024f9600bc36fe7203ea76dd9c0996fe8e5") }
           , (TxIn
                (TxId
                   #"d08a4a34021c0dc43a92f9719fbf0caefdfdf2cd4b34490f323843decc3bd2d6")
                2) ->
               TxOut
                { address =
                    (Addr
                       Mainnet
                       (ScriptCred
                          #"20d21f29634727dd5cea8f233305af3f6481a9e45502d6998704af7e")
                       (PtrRef (Ptr SlotNo 0 8 3)))
                , value = (Value (Coin 3) [])
                , datum = NoDatum
                , reference script = ?- }
           , (TxIn
                (TxId
                   #"f6e36a222b6519b9929ba85450749b2fd8d6663f7ed17c1b251cdecd7d73302f")
                1) ->
               TxOut
                { address =
                    (BootstrapAddress
                      { root =
                          04040302030102020003030203040201040400010104040302040202
                      , derivationpath = ?-
                      , networkmagic = Testnet 3
                      , remain = {  }
                      , type = ATVerKey })
                , value = (Value (Coin 6) [])
                , datum = NoDatum
                , reference script = ?- } }
```
Now, there are many things wrong with this transaction so it failing to go through is no strange thing. However, one of the errors I get appears to be a bit misleading?
```
ValueNotConservedUTxO
 { coin consumed = (Value (Coin 6) []) , coin produced = (Value (Coin 9) []) },
```
If we count up the value consumed by the transaction on the input side we do indeed get 6, and we have one collateral input with value 3 (which I guess is where the 9 is coming from?) but this doesn't seem to me like we are _producing_ 9 coins here. If anything, I take it we are consuming 9 coins and only returning 6!

TimSheard:
First step in the process of removing cardano-ledger-pretty
1) removed all reference to Cardano.Ledger.Pretty in Test.Cardano.Ledger
2) did this by moving functionality of Cardano.Ledger.Pretty  to Test.Cardano.Ledger.Generic.PrettyCore
3) Added new pretty printers for Predicate failures

Fixes #3974
Fixes #3926

# Description

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/libs/cardano-ledger-test/cardano-ledger-test.cabal
+++ b/libs/cardano-ledger-test/cardano-ledger-test.cabal
@@ -71,6 +71,7 @@ library
         Test.Cardano.Ledger.Generic.MockChain
         Test.Cardano.Ledger.Generic.ModelState
         Test.Cardano.Ledger.Generic.PrettyCore
+        Test.Cardano.Ledger.Generic.PrettyTest
         Test.Cardano.Ledger.Generic.Properties
         Test.Cardano.Ledger.Generic.Same
         Test.Cardano.Ledger.Generic.Scriptic
@@ -92,6 +93,7 @@ library
     build-depends:
         array,
         base >=4.14 && <5,
+        bech32,
         bytestring,
         cardano-data,
         cardano-crypto-class,
@@ -107,7 +109,6 @@ library
         cardano-ledger-binary:{cardano-ledger-binary, testlib},
         cardano-ledger-conway:{cardano-ledger-conway, testlib},
         cardano-ledger-core:{cardano-ledger-core, testlib},
-        cardano-ledger-pretty,
         cardano-ledger-mary,
         cardano-ledger-shelley:{cardano-ledger-shelley, testlib},
         cardano-ledger-shelley-test,
--- a/eras/alonzo/test-suite/cardano-ledger-alonzo-test.cabal
+++ b/eras/alonzo/test-suite/cardano-ledger-alonzo-test.cabal
@@ -47,7 +47,6 @@ library
         cardano-ledger-alonzo:{cardano-ledger-alonzo, testlib} >=1.5.1,
         cardano-ledger-binary:{cardano-ledger-binary, testlib} >=1.0,
         cardano-ledger-core:{cardano-ledger-core, testlib} >=1.10 && <1.11,
-        cardano-ledger-pretty,
         cardano-ledger-allegra >=1.2,
         cardano-ledger-shelley:{cardano-ledger-shelley, testlib} >=1.6 && <1.10,
         cardano-ledger-shelley-test >=1.2,
--- a/eras/alonzo/test-suite/src/Test/Cardano/Ledger/Alonzo/AlonzoEraGen.hs
+++ b/eras/alonzo/test-suite/src/Test/Cardano/Ledger/Alonzo/AlonzoEraGen.hs
@@ -67,7 +67,6 @@ import Cardano.Ledger.Mary.Value (
  )
 import Cardano.Ledger.Plutus.Data (Data (..))
 import Cardano.Ledger.Plutus.Language (Language (..))
-import Cardano.Ledger.Pretty.Alonzo ()
 import Cardano.Ledger.Shelley.PParams (Update)
 import Cardano.Ledger.TxIn (TxIn)
 import Cardano.Ledger.UTxO (
--- a/eras/shelley-ma/test-suite/cardano-ledger-shelley-ma-test.cabal
+++ b/eras/shelley-ma/test-suite/cardano-ledger-shelley-ma-test.cabal
@@ -43,7 +43,6 @@ library
         bytestring,
         cardano-ledger-binary:{cardano-ledger-binary, testlib} >=1.2,
         cardano-ledger-core:{cardano-ledger-core, testlib} >=1.3,
-        cardano-ledger-pretty,
         cardano-ledger-allegra:{cardano-ledger-allegra, testlib} ^>=1.3,
         cardano-ledger-mary:{cardano-ledger-mary, testlib} ^>=1.5,
         cardano-slotting,
--- a/eras/shelley-ma/test-suite/src/Test/Cardano/Ledger/AllegraEraGen.hs
+++ b/eras/shelley-ma/test-suite/src/Test/Cardano/Ledger/AllegraEraGen.hs
@@ -29,7 +29,6 @@ import Cardano.Ledger.Binary (encCBOR, serialize')
 import Cardano.Ledger.Coin (Coin)
 import qualified Cardano.Ledger.Crypto as CryptoClass
 import Cardano.Ledger.Keys (KeyHash)
-import Cardano.Ledger.Pretty.Mary ()
 import Cardano.Ledger.Shelley.API (KeyRole (Witness))
 import Cardano.Ledger.Shelley.PParams (Update)
 import Cardano.Ledger.Shelley.Tx (pattern ShelleyTx)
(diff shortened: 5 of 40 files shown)


lehins (approved):
Nice. Next step is to remove `cardano-ledger-pretty`

libs/cardano-ledger-test/cardano-ledger-test.cabal
euonymos, line 1: I think you forgot to bump the version of cardano-ledger-test here. This is probably why the latest version on CHaP is dated back to the summer 2023?

euonymos: I see, it's set to `9.9.9.9` and "not supposed to be released". Why is that the case @lehins? Though it's a testing library, prettyprinting stuff is something others may want to use.

lehins: We do not want to provide custom pretty printing functionality for public use, because it comes with a lot of overhead for the ledger teqm. Also, "others may want to use" is not good enough reason for us to provide this functionality in a public library. Quick search through github revealed that noone uses that functionality, so we decided to move it into the test suite.

euonymos: Thank you for the reply. We are using it in the new version of [the Cardano test emulator](https://milestones.projectcatalyst.io/projects/1000121) we are working on. It sounds like you recommend not using it all, but I don't see why we might want to roll out our own version from scratch. BTW the older versions of that `cardano-ledger-test` are in CHaP, which might be misleading. Technically it's public either way, it's just a matter of convenience. Not only prettyprinting stuff from the test library is useful for our project. And it's not a big deal to use it directly, though I would prefer having it in CHaP with proper version bumps when they happen. Anyway, thanks for your hard work on the ledger.

lehins: > BTW the older versions of that cardano-ledger-test are in CHaP, which might be misleading.

I know. CHaP is immutable, so we can't remove it. Initial version of `cardano-ledger-test` was released early on by mistake.

> Technically it's public either way, it's just a matter of convenience.

Not exactly. Providing official support for any functionality puts extra overhead on the ledger team, including testing, releasing, etc of that functionality. For example bunch of current pretty printers that are used for testing are very much incomplete, which is not a problem for us, but it would be a big problem for us if we maintained a library that provided this incomplete functionality. Like we did in the past.

> And it's not a big deal to use it directly

I would advise against it, since we do not guarantee its correctness, use it at your own risk.

I'd suggest downstream users of ledger roll out their own pretty printers. If you are willing to maintain a library with pretty printers, which would become popular in the community, then we'd be happy to add such a library to CHaPs.

Issue #3970, opened 2024-01-04
Use `estimateMinFeeTx` to fixup transactions in `ImpTest`


Soupstraw:
We should also use this in the fixup function of `ImpTest`

_Originally posted by @Soupstraw in https://github.com/IntersectMBO/cardano-ledger/issues/3962#issuecomment-1875541599_

lehins:
Fixed in #3976

Issue #3972, opened 2024-01-04
Remove `MissingRequiredSigners` once we are in `Conway`


lehins:
Once we enter `Conway` era we can remove `MissingRequiredSigners` predicate failure and the associated `requiredSignersAreWitnessed` predicate check. This is true because `getWitsVKeyNeeded` includes the `reqSignerHashes` for all eras starting from Alonzo. Another reason why we can remove `MissingRequiredSigners` in Conway is because it is cannot be produced once we passed Babbage.

The fix was implemented #3976

PR #3976, opened 2024-01-06
Introduce accurate min fee calculation function: `calcMinFeeTx`

### LINKED ISSUE #3767: Improve `fixupTx` fee calculation
Ok, so the only reason why this works right now is because the `minFeeA` is set to 0.

Setting witnesses changes the size of the transaction, which in turn changes the fee.

So this whole fixupTx needs to be to reach a fixed point.

That being said, we don't have to fix this in this PR, because it is not as trivial as it sounds

_Originally posted by @lehins in https://github.com/input-output-hk/cardano-ledger/pull/3733#discussion_r1340016648_

lehins:
# Description

Addition of `calcMinFeeTx` function is a better alternative to `estimateMinFeeTx` introduced in #3962 

In order to implement this function it was necessary to:

* add `genesisKeyHashCount`, that counts up how many genesis key transaction will add
* Properly fix `getWitsVKeyNeeded` to also include `reqSignerHashes` as it was designed in the spec.

This last change is important, because that is the only change in this PR that actually affects the rules. Currently for Babbage and Alonzo it introduces a redundant check. It is not a problem since checking presence of a few extra hashes from `reqSignerHashes` in a map of provided hashes is not expensive at all. However this complete functionality of `getWitsVKeyNeeded` according to the spec and allows us to use it for fee calculation. Also when we enter Conway, we'l be able to get rid of the redundant check, see #3972

Moreover, this new function `calcMinFeeTx` was utilized in the ImpTest in order to take care fixing up the fee during creation of test transactions. Fixes #3767 

Some extra cleanup was also added:
* Deprecation of poorly named `txid` in favor of better named `txIdTxBody` and `txIdTx`
* Deprecation of poorly named `txup` function and switching of `PPUP` rule signal from `Maybe Update` to `StrinctMaybe Update` in order to reduce unnecessary conversions
* Add `ProtVerAtMost era 8` as a superclass constraint to `ShelleyEraTxBody` effectively disabling the type class from Conway onwards. This resulted in a cleanup of redundant constraints and fixup of over-constrained functions
* Also, `estimateMinFeeTx` functionality that was introduced in #3962 was moved to a new module `Cardano.Ledger.Tools` in order to make it accessible to the test suite

# Checklist

- [ ] Self-reviewed the diff


libs/cardano-ledger-core/src/Cardano/Ledger/Tools.hs
```diff
@@ -0,0 +1,124 @@
+{-# LANGUAGE DataKinds #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeApplications #-}
+{-# LANGUAGE TypeOperators #-}
+
+-- | This is a module that contains functionality that is not necessary for ledger
+-- operation, but is usefuk for testing as well as for downstream users of ledger
```
teodanciu: I assume this should say `useful` ? :laughing:

libs/cardano-ledger-core/CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## 1.10.0.0
 
+* Add `estimateMinFeeTx`, `addDummyWitsTx` and move `setMinFeeTx` to `Cardano.Ledger.Tools`
```
teodanciu, line 8: I wonder if we should mention the removal of `estimateMinFeeTx` and `addDummyWitsTx` functions in the cardano-ledger-api changelog as well, since maybe someone using the api might not know to look into core to figure out what happened?

eras/shelley/impl/src/Cardano/Ledger/Shelley/AdaPots.hs
```diff
@@ -140,7 +140,7 @@ instance Show Produced where
 
 -- | Compute the Coin part of what is consumed by a TxBody, itemized as a 'Consume'
 consumedTxBody ::
-  ShelleyEraTxBody era =>
+  EraTxBody era =>
```
teodanciu, line 142: Just curious, I can't see the link between the main change  (adding `ProtVerAtMost era 8` to  `ShelleyEraTxBody` ) and this - was this maybe overconstrained as it was?

eras/allegra/impl/src/Cardano/Ledger/Allegra/TxBody.hs
```diff
@@ -84,7 +84,7 @@ import GHC.Generics (Generic)
 import Lens.Micro
 import NoThunks.Class (NoThunks (..))
 
-class ShelleyEraTxBody era => AllegraEraTxBody era where
+class EraTxBody era => AllegraEraTxBody era where
```
teodanciu, line 88: I suppose removing `ShelleyEraTxBody`  constraint was necessary so that Conway can still have an instance for this, correct?

libs/cardano-ledger-api/test/Test/Cardano/Ledger/Api/Tx/Body.hs
```diff
@@ -89,10 +89,28 @@ evaluateTransactionBalance pp dpstate (UTxO u) txBody = consumed <-> produced
...
     withdrawals = fold . unWithdrawals $ txBody ^. withdrawalsTxBodyL
 
+evaluateTransactionBalanceShelley ::
+  (EraTxBody era, ShelleyEraTxCert era) =>
+  PParams era ->
+  CertState era ->
+  UTxO era ->
+  TxBody era ->
+  Value era
+evaluateTransactionBalanceShelley pp dpstate (UTxO u) txBody = consumed <-> produced
```
teodanciu: This duplication is a bit sad, couldn't we define the function  `evaluateTransactionBalance` above using this one: 
```
evaluateTransactionBalance pp dpstate utxo txBody =
    evaluateTransactionBalanceShelley pp dpstate utxo txBody<> (txBody ^. mintValueTxBodyF)
```
?

eras/babbage/impl/src/Cardano/Ledger/Babbage/Rules/Utxow.hs
```diff
@@ -312,10 +312,15 @@ babbageUtxowTransition = do
...
   -- TODO can we add the required signers to witsVKeyNeeded so we dont need the check below?
+  --
+  --      ^ Answer: yes we can. It is now implememnted. We just can't remove
+  --      MissingRequiredSigners predicate failure until we are in Conway
 
   {-  THIS DOES NOT APPPEAR IN THE SPEC as a separate check, but
       witsVKeyNeeded must include the reqSignerHashes in the union   -}
   {- reqSignerHashes txbody ⊆ witsKeyHashes -}
+  -- TODO: Remove once we are in Conway. This check is now handled by
+  -- validateNeededWitnesses, as per spec
```
teodanciu: Won't it fail with a different Predicate failure in  `validateNeededWitnesses`  (`MissingVKeyWitnessesUTXOW`)  ?  Is that not something we need to be careful about changing?

libs/cardano-ledger-core/src/Cardano/Ledger/Tools.hs
```diff
@@ -47,6 +55,41 @@ setMinFeeTx pp tx =
...
+  -- | The current protocol parameters.
+  PParams era ->
+  -- | The transaction.
+  Tx era ->
+  -- | Number of extra KeyHash witnesses that will be supplied for satisfying native
+  -- scripts. It is impossible to know how many of these is required without knowing the
+  -- actual witnesses supplied and the time when the transaction will be
+  -- submitted. Therefore we put this burden on the user.
+  --
+  -- This number can also be used to specify all of the redundand extra key witnesses that
```
teodanciu: typo: `redundant`

libs/cardano-ledger-core/src/Cardano/Ledger/Core.hs
```diff
@@ -235,6 +235,11 @@ class
   getTotalRefundsTxBody pp lookupStakingDeposit lookupDRepDeposit txBody =
     getTotalRefundsTxCerts pp lookupStakingDeposit lookupDRepDeposit (txBody ^. certsTxBodyL)
 
+  -- | This function is not used in the ledger rules. It is only used by the downstream
+  -- tooling to figure out how many witnesses should be supplied for Genesis keys.
+  genesisKeyHashCount :: TxBody era -> Int
```
teodanciu: Do we have a convention for when to use functions prefixed by `get` in these typeclasses? I have seen a few cases where the version that didn't start with `get` was deprecate din favour of the `get`-prefixed one  (for example `minFee` and `getMinFree`). 
Is it perhaps to do with the comment that you added that it's not used in the rules?

libs/cardano-ledger-core/src/Cardano/Ledger/Core.hs
```diff
@@ -235,6 +235,11 @@ class
...
+  genesisKeyHashCount _ = 0
```
teodanciu: Why have a default implementation - is it because they don't exist starting with Conway?

eras/shelley/impl/testlib/Test/Cardano/Ledger/Shelley/Imp/EpochSpec.hs
```diff
@@ -38,7 +38,7 @@ spec = describe "EPOCH" $ do
     do
       deposited <- getsNES $ nesEsL . esLStateL . lsUTxOStateL . utxosDepositedL
       deposited `shouldBe` zero
-    _ <- submitTx "simple transaction" $ mkBasicTx mkBasicTxBody
+    _ <- submitTxAnn "simple transaction" $ mkBasicTx mkBasicTxBody
```
teodanciu: You could use `submitTxAnn_` here

eras/shelley/impl/testlib/Test/Cardano/Ledger/Shelley/ImpTest.hs
```diff
@@ -470,56 +479,61 @@ fixupFees tx = do
...
         & bodyTxL . outputsTxBodyL %~ (remainingTxOut :<|)
-    txId = TxId . hashAnnotated $ balancedTx ^. bodyTxL
-  impRootTxIdL .= txId
-  wits <- mkTxWits (balancedTx ^. bodyTxL)
-  pure $ balancedTx & witsTxL .~ wits
+    nativeScriptKeyWitsCount = 0 -- TODO figure out wits for native scripts
+    outsWithoutRemaining = tx ^. bodyTxL . outputsTxBodyL
+    fee = calcMinFeeTx utxo pp txNoWits nativeScriptKeyWitsCount
+    remainderAvailable = remainingTxOut ^. coinTxOutL <-> fee
+    remainingTxOut' = remainingTxOut & coinTxOutL .~ remainderAvailable
```
teodanciu, line 498: We could define this as 
`remainingTxOut & coinTxOutL %~ (<-> fee)`, but maybe you wanted to emphasize the relationship to `remainderAvailable` ?

eras/shelley/impl/testlib/Test/Cardano/Ledger/Shelley/ImpTest.hs
```diff
@@ -470,56 +479,61 @@ fixupFees tx = do
...
+    txNoWits =
+      tx
         & bodyTxL . inputsTxBodyL %~ Set.insert (TxIn impRootTxId $ TxIx 0)
         & bodyTxL . outputsTxBodyL %~ (remainingTxOut :<|)
-    txId = TxId . hashAnnotated $ balancedTx ^. bodyTxL
-  impRootTxIdL .= txId
-  wits <- mkTxWits (balancedTx ^. bodyTxL)
-  pure $ balancedTx & witsTxL .~ wits
+    nativeScriptKeyWitsCount = 0 -- TODO figure out wits for native scripts
+    outsWithoutRemaining = tx ^. bodyTxL . outputsTxBodyL
```
teodanciu, line 495: I don't understand why this is called `outsWithoutRemaining`,  it seems to be set to all the outs via `outputsTxBodyL` ?

eras/shelley/impl/CHANGELOG.md
```diff
@@ -5,7 +5,7 @@
 * Add `shelleyGenesisKeyHashCount`
 * Change `PPUP` signal to `StrictMaybe` from `Maybe`
 * Deprecate `txup` function.
-* Prevent instances for `ShelleyEraTxBody` starting with protocol version 9.
+* Prevent instances for `SahelleyEraTxBody` starting with protocol version 9.
```
teodanciu, line 8: Introduced a spurious `a` here in `ShelleyEraTxBody` :)

eras/shelley/impl/testlib/Test/Cardano/Ledger/Shelley/ImpTest.hs
```diff
@@ -726,10 +716,15 @@ addKeyPair kp@(KeyPair vk _) = do
...
-    Map.lookup (coerceKeyRole kh) impKeyPairs
+lookupKeyPair keyHash = do
+  keyPairs <- gets impKeyPairs
+  case Map.lookup keyHash keyPairs of
+    Just keyPair -> pure keyPair
+    Nothing ->
+      error $
+        "Could not find a keypair corresponding to keyHash:"
+          ++ show keyHash
+          ++ "\nAlways use `freshKeyHash` to create key hashes."
```
teodanciu, line 731: Nice that it gives a hint about what could fix the problem :pray:

teodanciu (approved):
Looks good to me, tons of improvements packed in this PR :+1:

Soupstraw:
I assume it also closes #3970 (which is basically a duplicate of #3767)

eras/shelley/impl/testlib/Test/Cardano/Ledger/Shelley/ImpTest.hs
Soupstraw: I think the idea is that the leftover coin gets added to this later so it's not the final sequence of outputs.

eras/shelley/impl/CHANGELOG.md
lehins: Thank you. Fixed

eras/shelley/impl/src/Cardano/Ledger/Shelley/AdaPots.hs
lehins: Yes. It was just overconstraint.

eras/shelley/impl/testlib/Test/Cardano/Ledger/Shelley/ImpTest.hs
lehins: Yes, "remaining" is in the sense of "remainder of the money" or "change given back to you"

That was the naming Joosep adopted, so I just went with it as well

eras/babbage/impl/CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## 1.6.0.0
 
+* Use `getAlonzoWitsVKeyNeeded` for `getWitsVKeyNeeded`
```
Soupstraw: Is it necessary to add this to the changelog? I thought we only have to add things that have changed at the type level.

eras/shelley/impl/testlib/Test/Cardano/Ledger/Shelley/ImpTest.hs
lehins: `remainderAvailable` is used below. So that was the reason it is not collapses into the single modification using `%~`

eras/shelley/impl/testlib/Test/Cardano/Ledger/Shelley/ImpTest.hs
teodanciu: Makes sense, thank you both :+1:

eras/babbage/impl/CHANGELOG.md
lehins: Also it is good to record any semantic changes to the behavior of a function.

In this case however, you are right that we don't need to add this line because `getWitsVKeyNeeded` was not even available in the previous version. 

So, thank you for pointing it out. I'll remove it.

eras/shelley/impl/testlib/Test/Cardano/Ledger/Shelley/Imp/EpochSpec.hs
lehins: Nice. Thank you. Done

libs/cardano-ledger-core/src/Cardano/Ledger/Core.hs
lehins: Precisely. All future eras after Babbage will have it set to 0, therefore it is a safe default.

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/UTxO.hs
```diff
@@ -233,3 +235,15 @@ getAlonzoScriptsNeeded (UTxO u) txBody =
         toList (txBody ^. certsTxBodyL)
 
     !minted = map (\pId@(PolicyID hash) -> (Minting pId, hash)) $ Set.toList $ txBody ^. mintedTxBodyF
+
+-- | Just like `getShelleyWitsVKeyNeeded`, but also requires `reqSignerHashesTxBodyL`.
+getAlonzoWitsVKeyNeeded ::
```
Soupstraw, line 240: I wonder if it'd be a good idea to add a `ProtVerAtMost` constraint to `get*WitsVKeyNeeded` functions. That would make it harder to accidentally use e.g. `getShelleyWitsVKeyNeeded` in Babbage when we have a newer version of this function available in Alzono.

eras/shelley/impl/testlib/Test/Cardano/Ledger/Shelley/ImpTest.hs
```diff
@@ -450,76 +456,90 @@ getRewardAccountAmount rewardAcount = do
...
   ShelleyEraImp era =>
   Tx era ->
-  ImpTestM
-    era
-    ( Either
-        [PredicateFailure (EraRule "LEDGER" era)]
-        (State (EraRule "LEDGER" era), Tx era)
-    )
-submitTx_ tx = do
+  ImpTestM era (Either [PredicateFailure (EraRule "LEDGER" era)] (Tx era))
```
Soupstraw, line 528: I think I'd prefer returning the `TxId` rather than the `Tx` in all of the ImpTest functions that deal with submitting a transaction. Returning a `Tx` would also work since you can access the `TxId` from it as well, but it's a good idea to agree on which one we're going to use and then use that across all `ImpTest` modules. 

The way I'd like to think about it is that `Tx` is used for transactions that are not yet submitted to the chain and `TxId` represents a transaction on the chain. That would have the added bonus that whenever you submit a transaction, you'll actually have to look it up from the UTxO and we might catch some additional bugs with that.

libs/cardano-ledger-core/src/Cardano/Ledger/Core.hs
lehins: `minfee` used to be a standalone function. I was adding a type family suffix and prefixing with `get` every getter function that became a type class function, i.e. became era agnostic, like `getMinFeeTx` for example.

I think for consistency sake this one should be also named `getGenesisKeyHashCountTxBody`. 

That being said, there was no official convention enforced for naming. Which I think we should get together at some point and develop as a team.

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/ImpTest.hs
```diff
@@ -329,22 +331,23 @@ trySubmitVote ::
...
-                  gaId
-                  ( VotingProcedure
-                      { vProcVote = vote
-                      , vProcAnchor = SNothing
-                      }
-                  )
-              )
-          )
+trySubmitVote vote voter gaId =
+  fmap (fmap txIdTx) $
```
Soupstraw, line 335: We wouldn't have to do this if `trySubmitTx` returned a `TxId`, see my comment above

Soupstraw (commented):
Looks good overall. 

I have bit of mixed feelings about losing the `ShelleyEraTxBody` superclass, since it makes the system of typeclasses less uniform. Sure, we can get rid of some `ProtVerAtMost` constraints, but we also have to add an explicit `ShelleyEraTxBody` constraint to some functions that didn't need it before. Then again it's nice to not have to implement `ShelleyEraTxBody` with a bunch of dummy methods from Conway onwards.

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/UTxO.hs
lehins: But this function actually uses `getShelleyWitsVKeyNeeded`.

eras/babbage/impl/src/Cardano/Ledger/Babbage/Rules/Utxow.hs
lehins: That's a very good question.

There are two important points in ledger that we need to consider with respect to predicate failures:

* If a transaction was invalid for any reason, it should continue being invalid in the future. Regardless of the type of predicate failure produced. That is necessary for consensus. In other words consensus does not care why transaction is invalid all it cares about is that it is invalid, because it doesn't propagate invalid transactions.
* We cannot start producing new types of predicate failures for current era that were not available before without a protocol version bump, because that would break NodeToClient communication protocol on the serialization level.

Neither one of those restrictions are preventing us from producing an extra predicate failure, which is what happening right now in this PR. In the matter of fact, writing this explanation up makes me realize that we can even produce a different predicate failure for the same predicate check, as long as that new type of predicate check existed before. So, technically we can take care of this TODO even now.

What we cannot do until we are in Conway, is to remove the actual `MissingRequiredSigners` type constructor, because that would break NodeToClient communication on the client side. This is because a newer client can be talking to an older node, that is capable producing this predicate failure.

libs/cardano-ledger-core/src/Cardano/Ledger/Tools.hs
lehins: I think so too :laughing:

libs/cardano-ledger-core/src/Cardano/Ledger/Tools.hs
lehins: Done

libs/cardano-ledger-core/CHANGELOG.md
lehins: `estimateMinFeeTx` is still re-exported in `cardano-ledger-api`.

Also, neither of those two functions are in any release on chaps yet.

eras/allegra/impl/src/Cardano/Ledger/Allegra/TxBody.hs
lehins: Correct.

libs/cardano-ledger-api/test/Test/Cardano/Ledger/Api/Tx/Body.hs
lehins: Yes, we can. Thank you!!!

eras/shelley/impl/testlib/Test/Cardano/Ledger/Shelley/ImpTest.hs
lehins: I totally hear you. However, the problem is that transaction that was supplied to the function is not the transaction that gets submitted. We modify it by fixing up the fee and by adding witnesses to it. Without returning the modified transaction there is no way to get access to what has been submitted, which can be useful for some tests.

That was the main reason why I switched from the `TxId` to `Tx`.

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/ImpTest.hs
lehins: See my other comment.

If it proves to be a common enough pattern then we can create extra helper functions that return TxId instead of Tx, but calling `txIdTx` on the result is quite easy.

lehins:
> I have bit of mixed feelings about losing the ShelleyEraTxBody superclass, since it makes the system of typeclasses less uniform. 

It allows us to deprecate a lot of functionality in bulk. Starting with the next era that follows  Conway we'll do the same thing for `ShelleyEraTxCert`. So, it will become a uniform patter over time: if all functionality from the era is deprecated we can deprecate the whole class.

>  we can get rid of some ProtVerAtMost constraints, we also have to add an explicit ShelleyEraTxBody constraint to some functions that didn't need it before

We can replace one for another. That's all that has happend in this PR.

> Then again it's nice to not have to implement ShelleyEraTxBody with a bunch of dummy methods from Conway onwards.

Yeah for all of the eras that will follow.

I think it makes the deprecation process much cleaner when we can deprecate a class as a whole.

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/UTxO.hs
Soupstraw: oh, right, never mind, that would be too much hassle

eras/shelley/impl/testlib/Test/Cardano/Ledger/Shelley/ImpTest.hs
Soupstraw: Oh right, we can't get much besides the UTxO entry with the `TxId`, yeah that makes sense.

PR #3977, opened 2024-01-08
Added GetAccountStateQuery

### LINKED ISSUE #3968: Add a  `GetAccountState` query
Creating a `TreasuryWithdrawal` proposal requires specifying the current treasury amount for proposal policy to be able to validate the proposal.

In order for cardano-cli to be able to submit the current amount in the treasury we should provide a query for it.

It would make sense to provide a query for the `AccountState`, rather than just for the treasury amount, since knowing reserves could be useful in some other setting.

Soupstraw:
# Description

Adds `GetAccountStateQuery` to the API.

resolves #3968 

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/libs/cardano-ledger-api/src/Cardano/Ledger/Api/State/Query.hs
+++ b/libs/cardano-ledger-api/src/Cardano/Ledger/Api/State/Query.hs
@@ -1,6 +1,5 @@
 {-# LANGUAGE DataKinds #-}
 {-# LANGUAGE GADTs #-}
-{-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 
 module Cardano.Ledger.Api.State.Query (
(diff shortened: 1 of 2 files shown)


libs/cardano-ledger-api/src/Cardano/Ledger/Api/State/Query.hs
```diff
@@ -243,3 +247,8 @@ queryCommitteeMembersState coldCredsFilter hotCredsFilter statusFilter nes = do
       , csQuorum = comQuorum
       , csEpochNo = currentEpoch
       }
+
+queryAccountState ::
+  NewEpochState era ->
+  AccountState
+queryAccountState = view $ nesEsL . esAccountStateL
```
teodanciu, line 253: Is this the same as `n ^. nesEsL . esAccountStateL` ?

teodanciu:
Could you add an entry in the Changelog file?

libs/cardano-ledger-api/src/Cardano/Ledger/Api/State/Query.hs
Soupstraw: Yeah, `^.` is `flip view` iirc

lehins (approved):
Perfect. Could you also please create a ticket on consensus repo to integrate this query?

lehins:
@Soupstraw needs a fourmolu and a rebase

PR #3978, opened 2024-01-10
Proposals pruning rewrite

### LINKED ISSUE #3894: Test unreachable proposals clean up functionality
Following #3855 write sanity unit tests to cover the functionality.

### LINKED ISSUE #3979: Fix pruning of unreachable and expired proposals
Currently there are issues with retiring expired proposals and cleaning up proposals that re unreachable due to removal of expired and enacted proposals.

We also need to organize and simplify how we do tracking of the whole proposal hierarchy

aniketd:
# Description

Closes #3979  

## Checklist

- [x] Reexport `ConwayEraTxBody` from `Conway.Core` without import cycles
- [x] Update `CHANGELOG`
- [x] Bump versions
- [x] Write documentation

# Checklist

- [x] Self-reviewed the diff


eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Proposals.hs
```diff
@@ -1,132 +1,424 @@
...
+pfHardForkL = lens pfHardFork $ \x y -> x {pfHardFork = y}
+
+pfCommitteeL :: Lens' (PForest f era) (f (GovPurposeId 'CommitteePurpose era))
+pfCommitteeL = lens pfCommittee $ \x y -> x {pfCommittee = y}
+
+pfConstitutionL :: Lens' (PForest f era) (f (GovPurposeId 'ConstitutionPurpose era))
+pfConstitutionL = lens pfConstitution $ \x y -> x {pfConstitution = y}
+
+pfrPParamUpdateL :: Lens' (PForest PRoot era) (PRoot (GovPurposeId 'PParamUpdatePurpose era))
+pfrPParamUpdateL = lens pfPParamUpdate $ \x y -> x {pfPParamUpdate = y}
```
teodanciu: Couldn't we define these ones in terms of the `pf` ones,  in order to reduce duplication? 
```
pfrPParamUpdateL = pfPParamUpdateL
..
pfhPParamUpdateL = pfPParamUpdateL
..
```

aniketd: You mean like this: `pfrPParamUpdateL = pfPParamUpdateL`?

teodanciu: Yeah, and same for the hierarchy , `pfhPParamUpdateL = pfPParamUpdateL`

teodanciu: It doesn't make much difference anyway..

eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Proposals.hs
```diff
@@ -1,132 +1,424 @@
...
+pnChildrenL :: Lens' (PNode a) (Set a)
+pnChildrenL = lens pnChildren $ \x y -> x {pnChildren = y}
+
+newtype PHierarchy a = PHierarchy
+  { unPHierarchy :: Map a (PNode a)
+  }
+  deriving stock (Show, Eq, Ord, Generic)
+  deriving newtype (NoThunks, NFData, Default)
+
+pHierarchyNTL :: Lens' (PHierarchy a) (Map a (PNode a)) -- FIXME: what a terrible name!
```
teodanciu: `pHierarchyNodesL` could work I think, a little less `wtf` than `NTL` perhaps :grinning:

aniketd: Agree :100: , I will get to it in the next push :+1:

eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Proposals.hs
```diff
@@ -1,132 +1,424 @@
...
+
+pHierarchyL :: Lens' (Proposals era) (PForest PHierarchy era)
+pHierarchyL = lens pHierarchy $ \x y -> x {pHierarchy = y}
+
+instance EraPParams era => ToJSON (Proposals era) where
+  toJSON = toJSON . pProps
+  toEncoding = toEncoding . pProps
+
+proposalsAddAction :: forall era. GovActionState era -> Proposals era -> Maybe (Proposals era)
+proposalsAddAction gas ps =
```
teodanciu, line 276: I don't understand all the details of the algorithm - I will try more, but just one question:  are we still checking that the `previous` stored in the same is the same as the `previous` assumed by the proposal?  i assume so, but I can't find where this check is ... any hints?

eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Proposals.hs
aniketd: It definitely helps reduce visual noise :+1: I will try this. Thanks!

eras/conway/impl/CHANGELOG.md
```diff
@@ -23,22 +23,44 @@
...
+      * `PForest`
+      * `PRoot`
+      * `PNode`
+      * `PHierarchy`
+      * `pRootsL`
+      * `prRootL`
+      * `prChildrenL`
+      * `pnChildrenL`
+      * `pHierarchyL`
+      * `pHierarchyNTL`
```
aniketd: This has become `pHierarchyNodesL`

eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Proposals.hs
aniketd: This is done :+1: Super thanks!

eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Proposals.hs
aniketd: Yes, this is correct. We return a `Nothing` here when we don't find the previous action id where we expect it to be.

eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Procedures.hs
```diff
@@ -546,38 +539,37 @@ data GovActionPurpose
...
-    , Show
-    , Generic
-    , EncCBOR
-    , DecCBOR
-    , NoThunks
-    , NFData
-    , ToJSON
-    , Ord
-    )
+  deriving (Eq, Ord)
```
lehins: I don't think Generic needs to be standalone
```suggestion
  deriving (Eq, Ord, Generic)
```

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Ratify.hs
```diff
@@ -309,18 +307,19 @@ ratifyTransition = do
...
+              _rsEnacted
+              _rsExpired
               rsDelayed
             )
       , RatifySignal rsig
       ) <-
     judgmentContext
+  -- FIXME: `reorderActions` is unnecessary after it is done the first
+  -- time. The solution may not be trivial.
   let reorderedActions = reorderActions rsig
```
lehins: That is pretty bad. We have O(n^2) operation inside of a loop. 

Nice catch. It is easy to fix. This call to `reorderActions` should be done here:

https://github.com/IntersectMBO/cardano-ledger/blob/d837d75820e3f7475c68abadfb08065073478485/eras/conway/impl/src/Cardano/Ledger/Conway/Governance.hs#L993-L995

```haskell
 runConwayRatify globals ratifyEnv ratifyState (RatifySignal ratifySig) = 
   let ratifyResult = 
         runReader
            (applySTS @(ConwayRATIFY era) 
            (TRC (ratifyEnv, ratifyState, RatifySignal (reorderActions ratifySig))))
```

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Arbitrary.hs
```diff
@@ -126,12 +123,11 @@ instance Crypto c => Arbitrary (ConwayGovCert c) where
...
-------------------------------------------------------------------------------------------
--- Cardano.Ledger.Conway.Goverance  ------------------------------------------------------
-------------------------------------------------------------------------------------------
-
 instance
-  (Era era, Arbitrary (PParams era), Arbitrary (PParamsHKD StrictMaybe era)) =>
+  ( Era era
+  , Arbitrary (PParams era)
+  , Arbitrary (PParamsHKD StrictMaybe era)
+  ) =>
```
lehins: Why this change?
```suggestion
  (Era era, Arbitrary (PParams era), Arbitrary (PParamsHKD StrictMaybe era)) =>
```

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Arbitrary.hs
```diff
@@ -141,7 +137,10 @@ instance
       <*> arbitrary
 
 instance
-  (Era era, Arbitrary (PParams era), Arbitrary (PParamsUpdate era)) =>
+  ( Era era
+  , Arbitrary (PParams era)
+  , Arbitrary (PParamsUpdate era)
+  ) =>
```
lehins: Ditto
```suggestion
  (Era era, Arbitrary (PParams era), Arbitrary (PParamsUpdate era)) =>
```

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Arbitrary.hs
```diff
@@ -164,24 +163,11 @@ instance
...
-      <*> arbitrary
-      <*> arbitrary
-      <*> arbitrary
-
 instance
-  (Era era, Arbitrary (PParams era), Arbitrary (PParamsUpdate era)) =>
+  ( Era era
+  , Arbitrary (PParams era)
+  , Arbitrary (PParamsUpdate era)
+  ) =>
```
lehins:
```suggestion
  (Era era, Arbitrary (PParams era), Arbitrary (PParamsUpdate era)) =>
```

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Arbitrary.hs
```diff
@@ -193,39 +179,208 @@ instance
...
+
+data ProposalsForEnactment era
+  = ProposalsForEnactment
+      (Proposals era)
+      (Seq.Seq (GovActionId (EraCrypto era)))
+      (Set.Set (GovActionId (EraCrypto era)))
+  deriving (Show, Eq)
+
+instance
+  forall era.
```
lehins:
```suggestion
```

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Binary/RoundTrip.hs
```diff
@@ -62,7 +62,7 @@ roundTripConwayEraTypesSpec = do
   describe "Conway State Types" $ do
     roundTripShareEraTypeSpec @era @EnactState
     roundTripShareEraTypeSpec @era @GovActionState
-    roundTripShareEraTypeSpec @era @Proposals
+    -- roundTripShareEraTypeSpec @era @Proposals -- TODO: @aniketd
```
Soupstraw:
```suggestion
    pendingWith "TODO: @aniketd" -- roundTripShareEraTypeSpec @era @Proposals
```

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/SpecClass.hs
```diff
@@ -148,7 +150,7 @@ instance Specification Size Int where
   genFromS _ _ _ = genFromSize
 
 instance
-  (Era era, Ord dom {- HasRep dom, HasRep rng, HasRep c, -}, Ord rng, Adds rng) =>
+  (EraPParams era, Arbitrary (PParamsHKD StrictMaybe era), Ord dom {- HasRep dom, HasRep rng, HasRep c, -}, Ord rng, Adds rng) =>
```
Soupstraw:
```suggestion
  (EraPParams era, Arbitrary (PParamsHKD StrictMaybe era), Ord dom, Ord rng, Adds rng) =>
```
Somebody might have just accidentally left that comment in while debugging

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/TypeRep.hs
```diff
@@ -930,6 +932,9 @@ genSizedRep _ VoteR = arbitrary
 
 genRep ::
   forall era b.
+  ( Arbitrary (PParamsHKD StrictMaybe era)
+  , EraPParams era
+  ) =>
```
Soupstraw: This function shouldn't have any constraints, you probably just have to case match on a proof somewhere

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/TypeRep.hs
```diff
@@ -864,6 +868,7 @@ genSizedRep _ DStateR =
     )
 genSizedRep _ GovActionIdR = arbitrary
 genSizedRep _ GovActionIxR = GovActionIx <$> choose (0, 100)
+genSizedRep _ ProposalsR = arbitrary
```
Soupstraw: I think this is the place where all of the `EraPParams` and `Arbitrary` constraints for `PParams` come from. You probably have to do a case analysis on the proof here and then remove the constraints from the function.

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Arbitrary.hs
```diff
@@ -193,39 +179,208 @@ instance
...
   ) =>
-  Gen (Seq.StrictSeq (GovActionState era))
-uniqueIdGovActions = Seq.fromList . nubBy (\x y -> gasId x == gasId y) <$> arbitrary
+  Gen (SSeq.StrictSeq (GovActionState era))
+_uniqueIdGovActions = SSeq.fromList . nubBy (\x y -> gasId x == gasId y) <$> arbitrary
+
+instance
+  (Era era, Crypto (EraCrypto era)) =>
+  Arbitrary (PrevGovActionIds era)
+  where
```
lehins: Redundant constraint 
```suggestion
instance Era era => Arbitrary (PrevGovActionIds era) where
```

eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Proposals.hs
```diff
@@ -1,132 +1,498 @@
...
+pfhPParamUpdateL = pfPParamUpdateL
+
+pfhHardForkL :: Lens' (PForest PHierarchy era) (PHierarchy (GovPurposeId 'HardForkPurpose era))
+pfhHardForkL = pfHardForkL
+
+pfhCommitteeL :: Lens' (PForest PHierarchy era) (PHierarchy (GovPurposeId 'CommitteePurpose era))
+pfhCommitteeL = pfCommitteeL
+
+pfhConstitutionL :: Lens' (PForest PHierarchy era) (PHierarchy (GovPurposeId 'ConstitutionPurpose era))
+pfhConstitutionL = pfConstitutionL
```
Soupstraw: Why are these lenses necessary when we already have a more general lens above?

eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Proposals.hs
```diff
@@ -1,132 +1,498 @@
...
+deriving instance Era era => NoThunks (PForest PRoot era)
+deriving instance Era era => NFData (PForest PRoot era)
+deriving instance Era era => Default (PForest PRoot era)
+
+deriving instance Era era => Eq (PForest PHierarchy era)
+deriving instance Era era => Ord (PForest PHierarchy era)
+deriving instance Era era => Generic (PForest PHierarchy era)
+deriving instance Era era => NoThunks (PForest PHierarchy era)
+deriving instance Era era => NFData (PForest PHierarchy era)
+deriving instance Era era => Default (PForest PHierarchy era)
```
Soupstraw: I think this should work
```suggestion
deriving instance Eq (f (GovPurposeId r era)) => Eq (PForest f era)
deriving instance Ord (f (GovPurposeId r era)) => Ord (PForest f era)
deriving instance Generic (f (GovPurposeId r era)) => Generic (PForest f era)
deriving instance NoThunks (f (GovPurposeId r era)) => NoThunks (PForest f era)
deriving instance NFData (f (GovPurposeId r era)) => NFData (PForest f era)
deriving instance Default (f (GovPurposeId r era)) => Default (PForest f era)
```

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Arbitrary.hs
```diff
@@ -193,39 +179,208 @@ instance
...
+  Gen (Proposals era)
+genProposals range = do
+  pgais <- arbitrary
+  i <- chooseInt range
+  go (def & pRootsL .~ fromPrevGovActionIds pgais) i
+  where
+    go :: Proposals era -> Int -> Gen (Proposals era)
+    go ps = \case
+      0 -> pure ps
+      n -> do
```
lehins: If you supply a negative number here by mistake you'll get an infinite loop
```suggestion
    go ps n
      | n <= 0 = pure ps
      | otherwise = do
```

eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Proposals.hs
lehins: I don't think this is gonna work, because `r` isn't present on the right hand side.
What you'll probably need is `QuantifiedConstraints`:
```haskell
deriving instance (forall r. Eq (f (GovPurposeId (r :: GovActionPurpose) era)) => 
  Eq (PForest f era)
```

eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Proposals.hs
```diff
@@ -1,132 +1,498 @@
...
-newtype Proposals era
-  = Proposals
-      (OMap.OMap (GovActionId (EraCrypto era)) (GovActionState era))
-  deriving newtype (Show, Eq)
-  deriving stock (Generic)
+-- | The root of a single `Proposals` tree. `prRoot` is always expected
+-- to be equal to the respective `PrevGovActionId` at the end of every
+-- epoch boundary
+data PRoot a = PRoot
+  { prRoot :: !(Maybe a)
```
lehins: I think we should be using strict data structures in this module
```suggestion
  { prRoot :: !(StrictMaybe a)
```

eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Proposals.hs
```diff
@@ -1,132 +1,498 @@
...
+                      & pHierarchyL . lenzH . pHierarchyNodesL %~ Map.insert newId (newNode $ Just parentId)
+                else
+                  if Map.member parentId $ ps ^. pHierarchyL . lenzH . pHierarchyNodesL
+                    then
+                      Just $
+                        ps
+                          & pPropsL %~ (OMap.||> gas)
+                          & pHierarchyL . lenzH . pHierarchyNodesL %~ Map.adjust (pnChildrenL %~ Set.insert newId) parentId
+                          & pHierarchyL . lenzH . pHierarchyNodesL %~ Map.insert newId (newNode $ Just parentId)
+                    else Nothing
```
Soupstraw:
```suggestion
            SJust parentId
              | Just parentId == ps ^. pRootsL . lenzR . prRootL ->
                  Just $
                    ps
                      & pPropsL %~ (OMap.||> gas)
                      & pRootsL . lenzR . prChildrenL %~ Set.insert newId
                      & pHierarchyL . lenzH . pHierarchyNodesL %~ Map.insert newId (newNode $ Just parentId)
              | Map.member parentId $ ps ^. pHierarchyL . lenzH . pHierarchyNodesL ->
                      Just $
                        ps
                          & pPropsL %~ (OMap.||> gas)
                          & pHierarchyL . lenzH . pHierarchyNodesL %~ Map.adjust (pnChildrenL %~ Set.insert newId) parentId
                          & pHierarchyL . lenzH . pHierarchyNodesL %~ Map.insert newId (newNode $ Just parentId)
              | _ -> Nothing
```

eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Proposals.hs
```diff
@@ -1,132 +1,498 @@
...
+prRootL = lens prRoot $ \x y -> x {prRoot = y}
+
+prChildrenL :: Lens' (PRoot a) (Set a)
+prChildrenL = lens prChildren $ \x y -> x {prChildren = y}
+
+-- | A non-root node in a `Proposals` tree. `prParent` is expected to be
+-- a `Nothing` only at the begining when no governance actions has been
+-- enacted yet.
+data PNode a = PNode
+  { pnParent :: !(Maybe a)
```
lehins: Ditto
```suggestion
  { pnParent :: !(StrictMaybe a)
```

eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Proposals.hs
```diff
@@ -1,132 +1,498 @@
...
+  -- down the tree and not up.
+  , pnChildren :: !(Set a)
+  }
+  deriving (Show, Eq, Ord, Generic, NoThunks, NFData, Default)
+
+pnChildrenL :: Lens' (PNode a) (Set a)
+pnChildrenL = lens pnChildren $ \x y -> x {pnChildren = y}
+
+-- | A single proposal-tree. This map represents all the action-ids that
+-- form the tree.
```
lehins: English is not my string suite, but am I wrong in thinking that it should be `a`, not `the`?
```suggestion
-- form a tree.
```

eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Proposals.hs
```diff
@@ -1,132 +1,498 @@
...
+              -- there are no proposals enacted yet, which means that
+              -- the root has to be `Nothing`
+              if isNothing $ ps ^. pRootsL . lenzR . prRootL
+                then
+                  Just $
+                    ps
+                      & pPropsL %~ (OMap.||> gas)
+                      & pRootsL . lenzR . prChildrenL %~ Set.insert newId
+                      & pHierarchyL . lenzH . pHierarchyNodesL %~ Map.insert newId (newNode Nothing)
+                else Nothing
```
Soupstraw:
```suggestion
            SNothing
              -- We only accept actions with parent `SNothing` when
              -- there are no proposals enacted yet, which means that
              -- the root has to be `Nothing`
              | isNothing $ ps ^. pRootsL . lenzR . prRootL ->
                  Just $
                    ps
                      & pPropsL %~ (OMap.||> gas)
                      & pRootsL . lenzR . prChildrenL %~ Set.insert newId
                      & pHierarchyL . lenzH . pHierarchyNodesL %~ Map.insert newId (newNode Nothing)
              | Nothing
```

eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Proposals.hs
```diff
@@ -1,132 +1,498 @@
...
+prChildrenL = lens prChildren $ \x y -> x {prChildren = y}
+
+-- | A non-root node in a `Proposals` tree. `prParent` is expected to be
+-- a `Nothing` only at the begining when no governance actions has been
+-- enacted yet.
+data PNode a = PNode
+  { pnParent :: !(Maybe a)
+  -- ^ This field lacks its lens only because we didn't need one yet
+  -- because, in all operations implemented so far, we only tranverse
+  -- down the tree and not up.
```
lehins: This can be a comment instead of haddock, but honestly this sort of comments aren't particularly useful IMHO
```suggestion
  -- This field lacks its lens only because we didn't need one yet
  -- because, in all operations implemented so far, we only traverse
  -- down the tree and not up.
```

eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Proposals.hs
```diff
@@ -1,132 +1,498 @@
...
+  , pnChildren :: !(Set a)
+  }
+  deriving (Show, Eq, Ord, Generic, NoThunks, NFData, Default)
+
+pnChildrenL :: Lens' (PNode a) (Set a)
+pnChildrenL = lens pnChildren $ \x y -> x {pnChildren = y}
+
+-- | A single proposal-tree. This map represents all the action-ids that
+-- form the tree.
+newtype PHierarchy a = PHierarchy
```
lehins: This really is a (potentially disconnected) graph. Which, in our case, when combine with the root, will always form a tree. But, as far as the definition goes this is really is a graph.
```suggestion
newtype PGraph a = PGraph
```

eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Proposals.hs
```diff
@@ -140,264 +506,302 @@ proposalsAddVote voter vote gai (Proposals omap) =
...
+    , "> HardFork"
+    , show $ ps ^. pRootsL . pfrHardForkL
+    , "> Committee"
+    , show $ ps ^. pRootsL . pfrCommitteeL
+    , "> Constitution"
+    , show $ ps ^. pRootsL . pfrConstitutionL
+    ]
+      <> ( if showRoots
+            then
+              [ "Heirarchy"
```
Soupstraw:
```suggestion
              [ "Hierarchy"
```

eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Proposals.hs
```diff
@@ -1,132 +1,498 @@
...
+    NewConstitution parent _ ->
+      update pfrConstitutionL pfhConstitutionL parent
+    InfoAction -> Just $ ps & pPropsL %~ (OMap.||> gas)
+  where
+    update ::
+      Lens' (PForest PRoot era) (PRoot (GovPurposeId p era)) ->
+      Lens' (PForest PHierarchy era) (PHierarchy (GovPurposeId p era)) ->
+      StrictMaybe (GovPurposeId p era) ->
+      Maybe (Proposals era)
+    update lenzR lenzH prev =
```
lehins: What about this naming convention? We usually use `L` suffix for indicating lens and `lenz` seems weird. But I won't care much if you keep it as is.
```suggestion
    update rootL graphL prev =
```

eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Proposals.hs
```diff
@@ -140,264 +506,302 @@ proposalsAddVote voter vote gai (Proposals omap) =
...
-  , pgacConstitution :: !(Set (PrevGovActionId 'ConstitutionPurpose (EraCrypto era)))
+proposalsLookupId gai (Proposals omap _ _) = OMap.lookup gai omap
+
+deriving instance Era era => Eq (PForest StrictMaybe era)
+deriving instance Era era => Ord (PForest StrictMaybe era)
+deriving instance Era era => Generic (PForest StrictMaybe era)
+deriving instance Era era => NoThunks (PForest StrictMaybe era)
+deriving instance Era era => NFData (PForest StrictMaybe era)
+deriving instance Era era => ToJSON (PForest StrictMaybe era)
+deriving instance Era era => Default (PForest StrictMaybe era)
```
Soupstraw: This becomes unnecessary with the changes I suggested above
```suggestion
deriving instance Era era => Eq (PForest StrictMaybe era)
deriving instance Era era => Ord (PForest StrictMaybe era)
deriving instance Era era => Generic (PForest StrictMaybe era)
deriving instance Era era => NoThunks (PForest StrictMaybe era)
deriving instance Era era => NFData (PForest StrictMaybe era)
deriving instance Era era => ToJSON (PForest StrictMaybe era)
deriving instance Era era => Default (PForest StrictMaybe era)
```

eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Proposals.hs
```diff
@@ -140,264 +506,302 @@ proposalsAddVote voter vote gai (Proposals omap) =
...
+                ( withoutSiblings
+                    & pHierarchyL . lenzH . pHierarchyNodesL .~ newHierarchy
+                    & pRootsL . lenzR . prRootL ?~ gpi -- Set the new root
+                    & pRootsL . lenzR . prChildrenL .~ pnChildren newRootNode -- Set the new root children
+                    & pPropsL .~ newOMap
+                , removed `Map.union` removedActions `Map.union` enactedAction
+                )
+
+proposalsShowDebug :: Era era => Proposals era -> Bool -> String
+proposalsShowDebug ps showRoots =
```
lehins: Debugging related functionality should live in the `testlib`

eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Proposals.hs
```diff
@@ -140,264 +506,302 @@ proposalsAddVote voter vote gai (Proposals omap) =
...
+        remove rs hy lenzR lenzH parent =
+          let gpi = GovPurposeId gai
+           in ( rs & lenzR . prChildrenL %~ Set.delete gpi
+              , hy
+                  & lenzH . pHierarchyNodesL %~ Map.delete gpi
+                  & case parent of
+                    SNothing -> id
+                    SJust parentGpi ->
+                      lenzH . pHierarchyNodesL %~ Map.adjust (pnChildrenL %~ Set.delete gpi) parentGpi
+              )
```
lehins: `roots` and `hierarchy` are already in scope, so they don't need to be passed as arguments to a local binding again
```suggestion
        remove ::
          Lens' (PForest PRoot era) (PRoot (GovPurposeId p era)) ->
          Lens' (PForest PHierarchy era) (PHierarchy (GovPurposeId p era)) ->
          StrictMaybe (GovPurposeId p era) ->
          (PForest PRoot era, PForest PHierarchy era)
        remove lenzR lenzH parent =
          let gpi = GovPurposeId gai
           in ( roots & lenzR . prChildrenL %~ Set.delete gpi
              , hierarchy
                  & lenzH . pHierarchyNodesL %~ Map.delete gpi
                  & case parent of
                    SNothing -> id
                    SJust parentGpi ->
                      lenzH . pHierarchyNodesL %~ Map.adjust (pnChildrenL %~ Set.delete gpi) parentGpi
              )
```

eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Proposals.hs
```diff
@@ -140,264 +506,302 @@ proposalsAddVote voter vote gai (Proposals omap) =
...
-   in (Proposals retained, removed)
-
--- | Extract `GovActionState`s for all the descendents of a set of
--- `GovActionId`s from the `Proposals`
+proposalsRemoveIds gais ps =
+  let (retainedOMap, removedFromOMap) = OMap.extractKeys gais $ ps ^. pPropsL
+      (roots, hierarchy) = foldl' removeEach (ps ^. pRootsL, ps ^. pHierarchyL) gais
+   in (Proposals retainedOMap roots hierarchy, removedFromOMap)
+  where
+    removeEach accum@(roots, hierarchy) gai =
```
lehins: Avoid space leaks. Tuples are lazy in their elements
```suggestion
    removeEach accum@(!roots, !hierarchy) gai =
```

eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Proposals.hs
```diff
@@ -140,264 +506,302 @@ proposalsAddVote voter vote gai (Proposals omap) =
...
+  Set (GovActionId (EraCrypto era))
+getAllDescendents (Proposals omap _roots hierarchy) gai = case OMap.lookup gai omap of
+  Nothing -> Set.empty
+  Just gas -> case gas ^. gasActionL of
+    ParameterChange _parent _ _ -> collectDescendents pfhPParamUpdateL
+    HardForkInitiation _parent _ -> collectDescendents pfhHardForkL
+    TreasuryWithdrawals _ _ -> Set.empty
+    NoConfidence _parent -> collectDescendents pfhCommitteeL
+    UpdateCommittee _parent _ _ _ -> collectDescendents pfhCommitteeL
+    NewConstitution _parent _ -> collectDescendents pfhConstitutionL
```
lehins: If fields in the type isn't important there is no need to pattern match on it
```suggestion
    ParameterChange {} -> collectDescendents pfhPParamUpdateL
    HardForkInitiation {} -> collectDescendents pfhHardForkL
    TreasuryWithdrawals {} -> Set.empty
    NoConfidence {} -> collectDescendents pfhCommitteeL
    UpdateCommittee {} -> collectDescendents pfhCommitteeL
    NewConstitution {} -> collectDescendents pfhConstitutionL
```

eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Proposals.hs
```diff
@@ -140,264 +506,302 @@ proposalsAddVote voter vote gai (Proposals omap) =
...
+      Set (GovActionId (EraCrypto era))
+    collectDescendents lenz = Set.map unGovPurposeId $ go lenz $ GovPurposeId gai
+    go ::
+      forall p.
+      Lens' (PForest PHierarchy era) (PHierarchy (GovPurposeId p era)) ->
+      GovPurposeId p era ->
+      Set (GovPurposeId p era)
+    go lenz gpi =
+      case Map.lookup gpi $ hierarchy ^. lenz . pHierarchyNodesL of
+        Nothing -> error "Impossible! getAllDescendents: GovPurposeId not found"
```
lehins: We don't want to crash the network if there is a bug.
```suggestion
        -- Impossible! getAllDescendents: GovPurposeId not found
        Nothing -> assert False mempty
```

eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Proposals.hs
```diff
@@ -1,132 +1,498 @@
...
+      update pfrCommitteeL pfhCommitteeL parent
+    UpdateCommittee parent _ _ _ ->
+      update pfrCommitteeL pfhCommitteeL parent
+    NewConstitution parent _ ->
+      update pfrConstitutionL pfhConstitutionL parent
+    InfoAction -> Just $ ps & pPropsL %~ (OMap.||> gas)
+  where
+    update ::
+      Lens' (PForest PRoot era) (PRoot (GovPurposeId p era)) ->
+      Lens' (PForest PHierarchy era) (PHierarchy (GovPurposeId p era)) ->
```
lehins:
```suggestion
      (forall f . Lens' (PForest f era) (f (GovPurposeId p era))) ->
```

eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Proposals.hs
```diff
@@ -140,264 +506,302 @@ proposalsAddVote voter vote gai (Proposals omap) =
...
+  Seq (GovActionId (EraCrypto era)) ->
+  Set (GovActionId (EraCrypto era)) ->
+  Proposals era ->
+  ( Proposals era
+  , Map (GovActionId (EraCrypto era)) (GovActionState era) -- Removed due to enactment
+  , Map (GovActionId (EraCrypto era)) (GovActionState era) -- Removed due to expiry
+  )
+proposalsApplyEnactment enactedGais expiredGais props =
+  let (unexpiredProposals, expiredRemoved) = proposalsRemoveDescendentIds expiredGais props
+      (enactedProposalsState, enactedRemoved) = foldl' enact (unexpiredProposals, Map.empty) enactedGais
```
lehins: Let's err on the side of caution and return the original. This way nothing will be removed. This is what we should do:
```suggestion
      (enactedProposalsState, enactedRemoved) =
        case foldlM' enact (unexpiredProposals, Map.empty) enactedGais of
          Nothing -> assert False (props, mempty, mempty)
          Just res -> res
```

eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Proposals.hs
```diff
@@ -140,264 +506,302 @@ proposalsAddVote voter vote gai (Proposals omap) =
...
+  where
+    enact (ps, removed) gai = case proposalsLookupId gai ps of
+      Nothing ->
+        error $ -- QUESTION: Is using `error` the best in this function? (Used atleast twice)
+          unlines
+            [ "Impossible! Enacted GovActionId:"
+            , "\t\t" <> show gai
+            , "\t" <> "not found in Proposals OMap"
+            ]
+      Just gas -> case gas ^. gasActionL of
```
lehins:
```suggestion
    enact (!ps, !removed) gai = do
      gas <- proposalsLookupId gai ps
      case gas ^. gasActionL of
```

eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Proposals.hs
```diff
@@ -140,264 +506,302 @@ proposalsAddVote voter vote gai (Proposals omap) =
...
+              (newOMap, enactedAction) = OMap.extractKeys (Set.singleton gai) $ withoutSiblings ^. pPropsL
+           in assert
+                (ps ^. pRootsL . lenzR . prRootL == strictMaybeToMaybe parent)
+                ( withoutSiblings
+                    & pHierarchyL . lenzH . pHierarchyNodesL .~ newHierarchy
+                    & pRootsL . lenzR . prRootL ?~ gpi -- Set the new root
+                    & pRootsL . lenzR . prChildrenL .~ pnChildren newRootNode -- Set the new root children
+                    & pPropsL .~ newOMap
+                , removed `Map.union` removedActions `Map.union` enactedAction
+                )
```
lehins: Switch the whole thing to Maybe
```suggestion
          Maybe
          ( Proposals era
          , Map (GovActionId (EraCrypto era)) (GovActionState era)
          )
        enactFromRoot lenzR lenzH parent = do
          let gpi = GovPurposeId gai
              siblings = Set.delete gai $ Set.map unGovPurposeId (ps ^. pRootsL . lenzR . prChildrenL)
              (withoutSiblings, removedActions) = proposalsRemoveDescendentIds siblings ps
              newHierarchy = Map.delete gpi $ withoutSiblings ^. pHierarchyL . lenzH . pHierarchyNodesL
              (newOMap, enactedAction) = OMap.extractKeys (Set.singleton gai) $ withoutSiblings ^. pPropsL
          newRootNode <- Map.lookup gpi $ ps ^. pHierarchyL . lenzH . pHierarchyNodesL
          assert (ps ^. pRootsL . lenzR . prRootL == strictMaybeToMaybe parent) $
            Just ( withoutSiblings
                    & pHierarchyL . lenzH . pHierarchyNodesL .~ newHierarchy
                    & pRootsL . lenzR . prRootL ?~ gpi -- Set the new root
                    & pRootsL . lenzR . prChildrenL .~ pnChildren newRootNode -- Set the new root children
                    & pPropsL .~ newOMap
                  , removed `Map.union` removedActions `Map.union` enactedAction
                  )
```

eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Proposals.hs
```diff
@@ -140,264 +506,302 @@ proposalsAddVote voter vote gai (Proposals omap) =
...
+            , "\t" <> "not found in Proposals OMap"
+            ]
+      Just gas -> case gas ^. gasActionL of
+        ParameterChange parent _ _ -> enactFromRoot pfrPParamUpdateL pfhPParamUpdateL parent
+        HardForkInitiation parent _ -> enactFromRoot pfrHardForkL pfhHardForkL parent
+        TreasuryWithdrawals _ _ -> enactWithoutRoot
+        NoConfidence parent -> enactFromRoot pfrCommitteeL pfhCommitteeL parent
+        UpdateCommittee parent _ _ _ -> enactFromRoot pfrCommitteeL pfhCommitteeL parent
+        NewConstitution parent _ -> enactFromRoot pfrConstitutionL pfhConstitutionL parent
+        InfoAction -> enactWithoutRoot
```
lehins, line 560:
```suggestion
        InfoAction -> Just enactWithoutRoot
```

eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Proposals.hs
```diff
@@ -140,264 +506,302 @@ proposalsAddVote voter vote gai (Proposals omap) =
...
+        error $ -- QUESTION: Is using `error` the best in this function? (Used atleast twice)
+          unlines
+            [ "Impossible! Enacted GovActionId:"
+            , "\t\t" <> show gai
+            , "\t" <> "not found in Proposals OMap"
+            ]
+      Just gas -> case gas ^. gasActionL of
+        ParameterChange parent _ _ -> enactFromRoot pfrPParamUpdateL pfhPParamUpdateL parent
+        HardForkInitiation parent _ -> enactFromRoot pfrHardForkL pfhHardForkL parent
+        TreasuryWithdrawals _ _ -> enactWithoutRoot
```
lehins, line 556:
```suggestion
        TreasuryWithdrawals _ _ -> Just enactWithoutRoot
```

lehins (commented):
This is looking great. I'll do a second pass next week once you get a chance to address all the current feedback

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Binary/RoundTrip.hs
aniketd: This has been retored in the subsequent PR :+1:

eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Proposals.hs
aniketd: Removed! :raised_hands:

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/TypeRep.hs
aniketd: I did this, and it got rid of some of them. Thanks! :raised_hands:

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/TypeRep.hs
aniketd: Yeps! :+1:

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/Env.hs
```diff
@@ -89,13 +89,13 @@ instance Show (Name era) where
   show (Name (VRaw n _ _ _)) = n
 
 -- | Does not satisfy extensionality
-instance Eq (Name era) where
+instance EraPParams era => Eq (Name era) where
```
lehins: All of the constraints added in this PR in `cardano-ledger-test` need to be reverted (the `EraPParams` and `Arbitrary (PParamsHKD StrictMaybe era))`). Once you remove all of them and you get to the function that complains about missing constraint, let me know and I can help you in solving that compilation error.

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/TreeDiff.hs
```diff
@@ -71,12 +71,40 @@ instance ToExpr (PParamsHKD StrictMaybe era) => ToExpr (ProposalProcedure era)
...
+      ]
+
+instance Era era => ToExpr (PForest PGraph era) where
+  toExpr (PForest p h com con) =
+    toExpr
+      [ toExpr p
+      , toExpr h
+      , toExpr com
+      , toExpr con
+      ]
```
lehins: Why not let Generic do its thing?
 ```suggestion
instance Era era => ToExpr (PForest PRoot era)

instance Era era => ToExpr (PForest PGraph era)
```

File: eras/conway/impl/cardano-ledger-conway.cabal [hunk 42]
```diff
@@ -135,6 +142,9 @@ library testlib
         cardano-strict-containers,
         small-steps
 
+    if flag(asserts)
+        ghc-options: -fno-ignore-asserts
+
```
lehins: Unless you are using `assert`s in the testlib (which would be silly, since we should be using hspec instead in tests), I don't think it is necessary.
```suggestion
```

eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Proposals.hs
```diff
@@ -1,132 +1,461 @@
...
+                      & pPropsL %~ (OMap.||> gas)
+                      & pRootsL . forestL . prChildrenL %~ Set.insert newId
+                      & pGraphL . forestL . pGraphNodesL %~ Map.insert newId (newNode $ SJust parentId)
+              | Map.member parentId $ ps ^. pGraphL . forestL . pGraphNodesL ->
+                  Just $
+                    ps
+                      & pPropsL %~ (OMap.||> gas)
+                      & pGraphL . forestL . pGraphNodesL %~ Map.adjust (pnChildrenL %~ Set.insert newId) parentId
+                      & pGraphL . forestL . pGraphNodesL %~ Map.insert newId (newNode $ SJust parentId)
+              | otherwise -> Nothing
```
lehins: Let's reduce duplication and simplify it a bit:
```suggestion
    update forestL prev
      | prev == ps ^. pRootsL . forestL . prRootL =
        Just $ ps
          & pPropsL %~ (OMap.||> gas)
          & pRootsL . forestL . prChildrenL %~ Set.insert newId
          & pGraphL . forestL . pGraphNodesL %~ Map.insert newId (newNode prev)
      | SJust parentId <- prev
      , Map.member parentId $ ps ^. pGraphL . forestL . pGraphNodesL =
         Just $ ps
             & pPropsL %~ (OMap.||> gas)
             & pGraphL . forestL . pGraphNodesL %~ 
             ( Map.insert newId (newNode $ SJust parentId) . 
               Map.adjust (pnChildrenL %~ Set.insert newId) parentId)
      | otherwise = Nothing
      where
        newId = GovPurposeId $ gas ^. gasIdL
        newNode parent = PNode parent Set.empty
```

eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Proposals.hs
```diff
@@ -1,132 +1,461 @@
...
+  forall era.
+  EraPParams era =>
+  ProposalsSerializable era ->
+  Maybe (Proposals era)
+mkProposals (ProposalsSerializable pgais omap) = do
+  ps@(Proposals omap' _roots _hierarchy) <-
+    foldl'
+      (\mprops gas -> proposalsAddAction gas =<< mprops)
+      initialProposals
+      (OMap.toStrictSeq omap)
```
lehins: `OMap` has `Foldable` instance, this conversion is unnecessary
```suggestion
      omap
```

eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Proposals.hs
```diff
@@ -1,132 +1,461 @@
...
+  = ProposalsSerializable
+      (PrevGovActionIds era)
+      (OMap.OMap (GovActionId (EraCrypto era)) (GovActionState era))
+
+instance EraPParams era => EncCBOR (ProposalsSerializable era) where
+  encCBOR (ProposalsSerializable roots props) =
+    encode $ Rec (ProposalsSerializable @era) !> To roots !> To props
+
+instance EraPParams era => DecCBOR (ProposalsSerializable era) where
+  decCBOR = decode $ RecD ProposalsSerializable <! From <! From
```
lehins: This is readundant. We can just use a regular tuple
```suggestion
```

eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Proposals.hs
```diff
@@ -1,132 +1,461 @@
...
 
 instance EraPParams era => EncCBOR (Proposals era) where
-  encCBOR = encCBOR . proposalsActions
+  encCBOR ps =
+    let roots = toPrevGovActionIds $ ps ^. pRootsL
+        props = ps ^. pPropsL
+     in encCBOR $ ProposalsSerializable roots props
```
lehins:
```suggestion
     in encCBOR (roots, props)
```

eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Proposals.hs
```diff
@@ -1,132 +1,461 @@
...
 
 instance EraPParams era => DecCBOR (Proposals era) where
-  decCBOR = fromGovActionStateSeq <$> decCBOR
+  decCBOR =
+    decCBOR >>= \ps ->
+      case mkProposals ps of
+        Nothing -> fail "Could not decode Proposals"
+        Just ps' -> pure ps'
```
lehins: It is better to turn `mkProposal` into `MonadFail`. This will allow for better error reporting
```suggestion
  decCBOR = decCBOR >>= uncurry mkProposals
```

eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Proposals.hs
```diff
@@ -1,132 +1,461 @@
...
+    constitutionRC = roots ^. pfConstitutionL . prChildrenL
+    isSubset h = and $ flip Set.isSubsetOf (Map.keysSet h) . pnChildren <$> Map.elems h
+
+-- | Reconstruct the @`Proposals`@ forest from an @`OMap`@ of
+-- @`GovActionState`@s and the 4 roots (@`PrevGovActionIds`@)
+mkProposals ::
+  forall era.
+  EraPParams era =>
+  ProposalsSerializable era ->
+  Maybe (Proposals era)
```
lehins:
```suggestion
  (EraPParams era, MonadFail m) =>
  PrevGovActionIds era ->
  OMap.OMap (GovActionId (EraCrypto era)) (GovActionState era) ->
  m (Proposals era)
```

eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Proposals.hs
```diff
@@ -1,132 +1,461 @@
...
+mkProposals (ProposalsSerializable pgais omap) = do
+  ps@(Proposals omap' _roots _hierarchy) <-
+    foldl'
+      (\mprops gas -> proposalsAddAction gas =<< mprops)
+      initialProposals
+      (OMap.toStrictSeq omap)
+  guard (omap == omap' && proposalsAreConsistent ps) $> ps
```
lehins:
```suggestion
  unless (omap == omap') $ fail "OMap is malformed"
  unless (proposalsAreConsistent ps) $ fail "Proposals structure is inconsistent"
  pure ps
```

eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Proposals.hs
```diff
@@ -1,132 +1,461 @@
...
+    isSubset h = and $ flip Set.isSubsetOf (Map.keysSet h) . pnChildren <$> Map.elems h
+
+-- | Reconstruct the @`Proposals`@ forest from an @`OMap`@ of
+-- @`GovActionState`@s and the 4 roots (@`PrevGovActionIds`@)
+mkProposals ::
+  forall era.
+  EraPParams era =>
+  ProposalsSerializable era ->
+  Maybe (Proposals era)
+mkProposals (ProposalsSerializable pgais omap) = do
```
lehins:
```suggestion
mkProposals pgais omap = do
```

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Epoch.hs
```diff
@@ -279,21 +279,41 @@ epochTransition = do
...
 
     govState1 =
       govState0
-        & cgProposalsL .~ proposals1
-        & cgEnactStateL .~ (newEnactState & ensPrevPParamsL .~ curPParams)
+        & cgProposalsL .~ newProposals
+        & cgEnactStateL
+          .~ ( newEnactState
+                & ensPrevPParamsL .~ curPParams
+             )
```
lehins: Let's reduce the diff a little
```suggestion
        & cgEnactStateL .~ (newEnactState & ensPrevPParamsL .~ curPParams)
```

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Ratify.hs
```diff
@@ -310,18 +293,18 @@ ratifyTransition = do
...
               rsDelayed
             )
       , RatifySignal rsig
       ) <-
     judgmentContext
-  let reorderedActions = reorderActions rsig
-  case reorderedActions of
-    gas@GovActionState {gasId, gasAction, gasExpiresAfter} :<| sigs -> do
+  -- FIXME: `reorderActions` is unnecessary after it is done the first
+  -- time. The solution may not be trivial.
```
lehins: Fixed
```suggestion
```

eras/conway/impl/src/Cardano/Ledger/Conway/UTxO.hs
```diff
@@ -115,7 +116,7 @@ conwayProducedValue pp isStakePool txBody =
   shelleyProducedValue pp isStakePool txBody
     <+> inject (txBody ^. treasuryDonationTxBodyL)
 
-instance Crypto c => EraUTxO (ConwayEra c) where
+instance (Crypto c, EraTx (ConwayEra c)) => EraUTxO (ConwayEra c) where
```
lehins: If you ever find yourself requiring constraints, that you know exist somewhere in the codebase, it means you need to bring them into scope. In this case, they are already in scope, so those should be redundant.
```suggestion
instance Crypto c => EraUTxO (ConwayEra c) where
```

eras/conway/impl/src/Cardano/Ledger/Conway/UTxO.hs
```diff
@@ -130,7 +131,7 @@ instance Crypto c => EraUTxO (ConwayEra c) where
 
   getWitsVKeyNeeded _ = getConwayWitsVKeyNeeded
 
-instance Crypto c => AlonzoEraUTxO (ConwayEra c) where
+instance (Crypto c, EraTx (ConwayEra c), AlonzoEraTx (ConwayEra c)) => AlonzoEraUTxO (ConwayEra c) where
```
lehins: Ditto
```suggestion
instance Crypto c => AlonzoEraUTxO (ConwayEra c) where
```

eras/conway/impl/test/Test/Cardano/Ledger/Conway/GovActionReorderSpec.hs
```diff
@@ -1,41 +1,39 @@
 {-# LANGUAGE TypeApplications #-}
+{-# OPTIONS_GHC -Wno-unused-imports #-}
```
lehins: Seems like you forgotten a warning disabler. It's ok to remove all imports and tests, just to add them in another PR.
```suggestion
```

eras/conway/impl/test/Test/Cardano/Ledger/Conway/GovActionReorderSpec.hs
```diff
@@ -1,41 +1,39 @@
...
+--   prop "orders actions correctly" $ do
+--     actionsList <-
+--       traverse
+--         (>>= genGovActionStateFromAction)
+--         (govActionGenerators @Conway)
+--     let sortedActions = sortOn (actionPriority . gasAction) actionsList
+--     shuffledActions <- shuffle actionsList
+--     pure $
+--       Seq.fromList sortedActions
+--         == reorderActions (Seq.fromList shuffledActions)
```
lehins: Another PR will bring it back and it will result in a smaller diff if the stuff in this module is completely removed. Keeping just the `spec` as a placeholder is enough.
Just leave a comment.
```suggestion
-- Bringing back tests in: https://github.com/IntersectMBO/cardano-ledger/pull/3981
```

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Arbitrary.hs
```diff
@@ -199,39 +180,191 @@ instance
...
+_uniqueIdGovActions = SSeq.fromList . nubBy (\x y -> gasId x == gasId y) <$> arbitrary
+
+instance Era era => Arbitrary (PrevGovActionIds era) where
+  arbitrary =
+    fmap PrevGovActionIds $
+      PForest
+        <$> (fmap GovPurposeId <$> arbitrary)
+        <*> (fmap GovPurposeId <$> arbitrary)
+        <*> (fmap GovPurposeId <$> arbitrary)
+        <*> (fmap GovPurposeId <$> arbitrary)
```
lehins: I don't think these are necessary
```suggestion
        <$> arbitrary <*> arbitrary <*> arbitrary <*> arbitrary
```

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/GovSpec.hs
```diff
@@ -154,9 +135,7 @@ spec =
         ConwayGovState expectedProposals expectedEnactState expectedPulser <- getsNES newEpochStateGovStateL
 
         impAnn "EnactState reflects the submitted governance action" $ do
-          let enactStateWithChildren =
-                initialEnactState
-                  & ensPrevGovActionIdsChildrenL . pgacConstitutionL %~ Set.insert (PrevGovActionId govActionId)
+          let enactStateWithChildren = initialEnactState
           expectedEnactState `shouldBe` enactStateWithChildren
```
lehins:
```suggestion
          expectedEnactState `shouldBe` initialEnactState
```

lehins (commented):
Once these comments are addressed, I think we can merge the PR.
This is an awesome piece of work. I love it!

eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Proposals.hs
aniketd: Awesome! :raised_hands:

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Epoch.hs
```diff
@@ -279,21 +279,38 @@ epochTransition = do
...
+    -- newly submitted proposals. We only need to apply the enactment
+    -- operations to this superset to get a new set of proposals with:
+    -- enacted actions and their sibling subtrees, as well as expired
+    -- actions and their subtrees, removed, and with all the votes
+    -- intact for the rest of them.
+    --
+    -- QUESTION: SHould we check here that the @`PrevGovActionIds`@ in
+    -- the new enact state match up with the resultant roots of the
+    -- @`Proposals`@. How do we handle the case when they dont? For now,
+    -- we leave this to the constrained property-tests.
```
lehins: EPOCH rule can't fail. So, we just need to write tests that verify that this case does not happen.
However, this is how we will enforce it #3998 later on
```suggestion
```

eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Proposals.hs
```diff
@@ -1,132 +1,430 @@
...
+  }
+  deriving (Generic)
+
+deriving instance
+  ( Show (f (GovPurposeId 'PParamUpdatePurpose era))
+  , Show (f (GovPurposeId 'HardForkPurpose era))
+  , Show (f (GovPurposeId 'CommitteePurpose era))
+  , Show (f (GovPurposeId 'ConstitutionPurpose era))
+  ) =>
+  Show (PForest f era)
```
lehins: I think we can do the same trick here:
```suggestion
deriving instance 
  (forall p. Show (f (GovPurposeId (p :: GovActionPurpose) era)) => 
  Show (PForest f era)
```

eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Proposals.hs
```diff
@@ -1,132 +1,430 @@
...
+          Just $
+            ps
+              & pPropsL %~ (OMap.||> gas)
+              & pRootsL . forestL . prChildrenL %~ Set.insert newId
+              & pGraphL . forestL . pGraphNodesL %~ Map.insert newId (newNode prev)
+      | SJust parentId <- prev
+      , Map.member parentId $ ps ^. pGraphL . forestL . pGraphNodesL =
+          Just $
+            ps
+              & pPropsL %~ (OMap.||> gas)
```
lehins: One more slight reduction in duplication:
```suggestion
    TreasuryWithdrawals _ _ -> Just psWithGas
    NoConfidence parent ->
      update pfCommitteeL parent
    UpdateCommittee parent _ _ _ ->
      update pfCommitteeL parent
    NewConstitution parent _ ->
      update pfConstitutionL parent
    InfoAction -> Just 
  where
    psWithGas = ps & pPropsL %~ (OMap.||> gas)
    update ::
      (forall f. Lens' (PForest f era) (f (GovPurposeId p era))) ->
      StrictMaybe (GovPurposeId p era) ->
      Maybe (Proposals era)
    update forestL prev
      | prev == ps ^. pRootsL . forestL . prRootL =
          Just $
            psWithGas
              & pRootsL . forestL . prChildrenL %~ Set.insert newId
              & pGraphL . forestL . pGraphNodesL %~ Map.insert newId (newNode prev)
      | SJust parentId <- prev
      , Map.member parentId $ ps ^. pGraphL . forestL . pGraphNodesL = Just psWithGas
```

eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Proposals.hs
```diff
@@ -140,264 +438,241 @@ proposalsAddVote voter vote gai (Proposals omap) =
...
+  Proposals era ->
+  ( Proposals era
+  , Map (GovActionId (EraCrypto era)) (GovActionState era) -- Removed due to enactment
+  , Map (GovActionId (EraCrypto era)) (GovActionState era) -- Removed due to expiry
+  )
+proposalsApplyEnactment enactedGais expiredGais props =
+  let (unexpiredProposals, expiredRemoved) = proposalsRemoveDescendentIds expiredGais props
+      (enactedProposalsState, enactedRemoved) =
+        case foldlM' enact (unexpiredProposals, Map.empty) enactedGais of
+          Nothing -> assert False (props, mempty)
```
lehins: Actually the correct decision in this impossible case is to definitely remove the enacted proposals. It is not as dangerous if we keep around the expired ones. But the enacted ones must be removed, otherwise they could be enacted again at the next epoch. Imagine the same TreasuryWithdrawal being enacted every epoch, that would be pretty bad :smile: 
```suggestion
          Nothing -> assert False (proposalsRemoveIds props enactedGais)
```

eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Proposals.hs
```diff
@@ -1,132 +1,430 @@
...
+            . proposalsAddAction gas
+            =<< mprops
+      )
+      initialProposals
+      omap
+  unless (omap == omap') $ fail "mkProposals: OMap is malformed"
+  unless (proposalsAreConsistent ps) $ fail "mkProposals: Proposals structure is inconsistent"
+  pure ps
+  where
+    initialProposals = pure $ def & pRootsL .~ fromPrevGovActionIds pgais
```
lehins: Current implementation is too lazy.
With this change the case statement forces the Maybe computation and `foldlM'` will ensure thunks are not built up in on the monadic computation. Also it would be nice to add a little bit more error reporting on problematic GovActionState if one ever occurs.
```suggestion
    foldlM'
      ( \props gas ->
        case proposalsAddAction gas props of
          Nothing -> fail $ "mkProposals: Could not add a proposal " ++ show (gas .^ gasId)
          Just props' -> pure props'
      )
      initialProposals
      omap
  unless (omap == omap') $ fail "mkProposals: OMap is malformed"
  unless (proposalsAreConsistent ps) $ fail "mkProposals: Proposals structure is inconsistent"
  pure ps
  where
    initialProposals = def & pRootsL .~ fromPrevGovActionIds pgais
```

lehins (approved):
Looks great! 
:+1:

eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Proposals.hs
aniketd: @lehins do you think we should `proposalsRemoveDescendentIds` here so that we don't ever collect dangling actions in the ledger state?

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Ratify.hs
```diff
@@ -330,69 +311,37 @@ ratifyTransition = do
...
-                  , rsEnacted = Set.insert gasId rsEnacted
-                  , rsRemoved = Set.union (Set.delete gasId siblings) rsRemoved
-                  , rsDelayed = delayingAction gasAction
-                  }
+          let
+            st' =
+              st
+                & rsEnactStateL .~ newEnactState
+                & rsDelayedL .~ delayingAction gasAction
+                & rsEnactedL %~ (Seq.:|> gasId)
```
lehins: Switch this to the actual `GovernaceActionState`. Many things will become simpler.
```suggestion
                & rsEnactedL %~ (Seq.:|> gas)
```

eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Proposals.hs
lehins: No, `proposalsRemoveIds` is enough, but we do need to make sure that the roots are updated.

eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Proposals.hs
```diff
@@ -140,264 +433,245 @@ proposalsAddVote voter vote gai (Proposals omap) =
...
--- | Extract `GovActionState`s for all the descendents of a set of
--- `GovActionId`s from the `Proposals`
+proposalsRemoveIds gais ps =
+  let (retainedOMap, removedFromOMap) = OMap.extractKeys gais $ ps ^. pPropsL
+      (roots, hierarchy) = foldl' removeEach (ps ^. pRootsL, ps ^. pGraphL) gais
+   in (Proposals retainedOMap roots hierarchy, removedFromOMap)
+  where
+    removeEach accum@(!roots, !hierarchy) gai =
+      case OMap.lookup gai $ ps ^. pPropsL of
+        Nothing -> accum
```
lehins: We should add an `assert` here, cause that should never happen, right?
```suggestion
        Nothing -> assert False accum
```

eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Proposals.hs
```diff
@@ -140,264 +433,245 @@ proposalsAddVote voter vote gai (Proposals omap) =
...
   (Proposals era, Map.Map (GovActionId (EraCrypto era)) (GovActionState era))
-proposalsRemoveIds gais (Proposals omap) =
-  let (retained, removed) = OMap.extractKeys gais omap
-   in (Proposals retained, removed)
-
--- | Extract `GovActionState`s for all the descendents of a set of
--- `GovActionId`s from the `Proposals`
+proposalsRemoveIds gais ps =
+  let (retainedOMap, removedFromOMap) = OMap.extractKeys gais $ ps ^. pPropsL
+      (roots, hierarchy) = foldl' removeEach (ps ^. pRootsL, ps ^. pGraphL) gais
```
lehins: This should let us avoid the lookup couple of lines below, because we'll be iterating over the actual states, not the ids.
```suggestion
      (roots, hierarchy) = foldl' removeEach (ps ^. pRootsL, ps ^. pGraphL) removedFromOMap
```

eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Proposals.hs
aniketd: Just leaving a note: This removal is not needed now since the change to using `GovActionState` for the enacted sequence allowed us to use an `assert` in a single place.

PR #3981, opened 2024-01-10
Test proposals rewrite

### LINKED ISSUE #3894: Test unreachable proposals clean up functionality

### LINKED ISSUE #3979: Fix pruning of unreachable and expired proposals

aniketd:
# Description

Closes #3894 

## Pending
- [x] Get rid of the pattern synonyms to make sure `Conway` constraint is done away with and all tests are sufficiently polymorphic over `era`
- [x] Reduces duplication by comparing entire tree structures rather than individual proposals
- [x] Address any other review comments

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/eras/conway/impl/test/Test/Cardano/Ledger/Conway/GovActionReorderSpec.hs
+++ b/eras/conway/impl/test/Test/Cardano/Ledger/Conway/GovActionReorderSpec.hs
@@ -1,8 +1,32 @@
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeApplications #-}
+
 module Test.Cardano.Ledger.Conway.GovActionReorderSpec (spec) where
 
+import Cardano.Ledger.Conway (Conway)
+import Cardano.Ledger.Conway.Governance (GovActionState (..), actionPriority, reorderActions)
+import Data.Foldable (Foldable (..))
+import Data.List (sort)
+import qualified Data.Sequence.Strict as Seq
+import Test.Cardano.Ledger.Binary.Arbitrary ()
 import Test.Cardano.Ledger.Common
+import Test.Cardano.Ledger.Conway.Arbitrary (ShuffledGovActionStates (..))
 
 spec :: Spec
-spec = pure ()
-
--- Bringing back tests in: https://github.com/IntersectMBO/cardano-ledger/pull/3981
+spec =
+  describe "Conway governance actions reordering" $ do
+    prop "preserves length when reordered" $
+      \(actions :: Seq.StrictSeq (GovActionState Conway)) ->
+        Seq.length actions `shouldBe` Seq.length (reorderActions @Conway actions)
+    prop "sorts by priority" $
+      \(actions :: Seq.StrictSeq (GovActionState Conway)) ->
+        sort (toList (actionPriority . gasAction @Conway <$> actions))
+          `shouldBe` toList (actionPriority . gasAction <$> reorderActions actions)
+    prop "same priority actions are not rearranged" $
+      \(a :: GovActionState Conway) (as :: Seq.StrictSeq (GovActionState Conway)) ->
+        let filterPrio b = actionPriority (gasAction a) == actionPriority (gasAction b)
+         in filter filterPrio (toList $ reorderActions @Conway (a Seq.:<| as))
+              `shouldBe` filter filterPrio (toList $ reorderActions (a Seq.:<| as))
+    prop "orders actions correctly with shuffles" $
+      \(ShuffledGovActionStates gass shuffledGass :: ShuffledGovActionStates Conway) -> do
+        reorderActions (Seq.fromList gass) `shouldBe` reorderActions (Seq.fromList shuffledGass)
--- a/eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp.hs
+++ b/eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp.hs
@@ -6,13 +6,10 @@
 {-# LANGUAGE TypeFamilies #-}
 {-# LANGUAGE TypeOperators #-}
 
-module Test.Cardano.Ledger.Conway.Imp (
-  spec,
-) where
+module Test.Cardano.Ledger.Conway.Imp (spec) where
 
-import Cardano.Ledger.BaseTypes (Inject)
 import Cardano.Ledger.Conway.Governance (ConwayGovState, GovState)
-import Cardano.Ledger.Conway.Rules (ConwayGovPredFailure (..))
+import Cardano.Ledger.Conway.Rules (ConwayGovPredFailure, ConwayLedgerPredFailure)
 import Cardano.Ledger.Core (EraRule)
 import Control.State.Transition.Extended (PredicateFailure)
 import Test.Cardano.Ledger.Common
@@ -26,7 +23,8 @@ spec ::
   forall era.
   ( ConwayEraImp era
   , GovState era ~ ConwayGovState era
-  , Inject (ConwayGovPredFailure era) (PredicateFailure (EraRule "LEDGER" era))
+  , PredicateFailure (EraRule "LEDGER" era) ~ ConwayLedgerPredFailure era
+  , PredicateFailure (EraRule "GOV" era) ~ ConwayGovPredFailure era
   ) =>
   Spec
 spec = do
--- a/eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/GovSpec.hs
+++ b/eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/GovSpec.hs
@@ -11,6 +11,7 @@
 
 module Test.Cardano.Ledger.Conway.Imp.GovSpec where
 
+import Cardano.Ledger.Address (RewardAccount (..))
 import Cardano.Ledger.Allegra.Scripts (
   pattern RequireAllOf,
   pattern RequireAnyOf,
@@ -22,8 +23,15 @@ import Cardano.Ledger.CertState (vsNumDormantEpochsL)
 import Cardano.Ledger.Coin (Coin (Coin))
 import Cardano.Ledger.Conway.Core
 import Cardano.Ledger.Conway.Governance
-import Cardano.Ledger.Conway.PParams (dvtCommitteeNoConfidenceL, pvtCommitteeNoConfidenceL, pvtPPSecurityGroupL)
-import Cardano.Ledger.Conway.Rules (ConwayGovPredFailure (..))
+import Cardano.Ledger.Conway.PParams (
+  dvtCommitteeNoConfidenceL,
+  pvtCommitteeNoConfidenceL,
+  pvtPPSecurityGroupL,
+ )
+import Cardano.Ledger.Conway.Rules (
+  ConwayGovPredFailure (..),
+  ConwayLedgerPredFailure (ConwayGovFailure),
+ )
 import Cardano.Ledger.Credential (Credential (KeyHashObj))
 import Cardano.Ledger.DRep (drepExpiryL)
 import Cardano.Ledger.Keys (KeyHash, KeyRole (..))
@@ -32,11 +40,13 @@ import Control.Monad (replicateM_)
 import Control.State.Transition.Extended (PredicateFailure)
 import Data.Default.Class (Default (..))
 import Data.Foldable (Foldable (..), traverse_)
+import qualified Data.List.NonEmpty as NE
 import qualified Data.Map.Strict as Map
 import Data.Maybe
 import qualified Data.Sequence as Seq
 import qualified Data.Sequence.Strict as SSeq
 import qualified Data.Set as Set
+import Data.Tree
 import Lens.Micro
 import Test.Cardano.Ledger.Conway.ImpTest
 import Test.Cardano.Ledger.Core.Rational (IsRatio (..))
@@ -46,6 +56,8 @@ spec ::
   forall era.
   ( ConwayEraImp era
   , GovState era ~ ConwayGovState era
+  , PredicateFailure (EraRule "LEDGER" era) ~ ConwayLedgerPredFailure era
+  , PredicateFailure (EraRule "GOV" era) ~ ConwayGovPredFailure era
   , Inject (ConwayGovPredFailure era) (PredicateFailure (EraRule "LEDGER" era))
   ) =>
   SpecWith (ImpTestState era)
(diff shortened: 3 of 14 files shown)


eras/conway/impl/test/Test/Cardano/Ledger/Conway/GovActionReorderSpec.hs
```diff
@@ -10,31 +11,25 @@ import Data.Foldable (Foldable (..))
...
+          `shouldBe` toList (actionPriority . gasAction <$> reorderActions actions)
+    prop "same priority actions are not rearranged" $
+      \(a :: GovActionState Conway) (as :: Seq.StrictSeq (GovActionState Conway)) ->
+        let filterPrio b = actionPriority (gasAction a) == actionPriority (gasAction b)
+         in filter filterPrio (toList $ reorderActions @Conway (a Seq.:<| as))
+              `shouldBe` filter filterPrio (toList $ reorderActions (a Seq.:<| as))
+    prop "orders actions correctly" $
+      \(actionsList :: [GovActionState Conway]) -> do
+        let sortedActions = sortOn (actionPriority . gasAction) actionsList
+        Seq.fromList sortedActions `shouldBe` reorderActions (Seq.fromList actionsList)
```
Soupstraw:
```suggestion
        let sortedActions = sortOn (actionPriority . gasAction) actionsList
        shuffledActions <- shuffle actionsList
        Seq.fromList sortedActions `shouldBe` reorderActions (Seq.fromList shuffledActions)
```

Soupstraw: I think the purpose of this test is to show that no matter how the actions are permuted, we still end up with the same output

Soupstraw: But I've implemented this in a weird way. I think that `sortedActions` should  just be replaced with `reorderActions`, since it's defined exactly as the actual implementation

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/GovSpec.hs
```diff
@@ -41,8 +46,8 @@ import Test.Cardano.Ledger.Imp.Common
 spec ::
   forall era.
   ( ConwayEraImp era
-  , GovState era ~ ConwayGovState era
   , Inject (ConwayGovPredFailure era) (PredicateFailure (EraRule "LEDGER" era))
+  , era ~ Conway
```
Soupstraw: I think it'd be better to just replace `era` with Conway

lehins: No, we should remove this restriction! Otherwise these tests will not be available in the next era

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/ImpTest.hs
```diff
@@ -752,4 +776,156 @@ electBasicCommittee = do
...
+  era ~ Conway =>
+  Forest (StrictMaybe (GovActionId (EraCrypto era))) ->
+  Tree (StrictMaybe (GovActionId (EraCrypto era)))
+pattern I a <- Node SNothing a
+  where
+    I a = Node SNothing a
+
+pattern J ::
+  forall era.
+  era ~ Conway =>
```
Soupstraw: Why is this constraint necessary?

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/GovSpec.hs
Soupstraw: But it's better if you can get rid of it of course

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/GovSpec.hs
```diff
@@ -55,6 +60,809 @@ spec =
...
+                        p11'
+                        [ J p111' []
+                          , J p112' []
+                          ]
+                      ]
+                  , J p2' [J p21' []]
+                  ]
+            ] <-
+            getProposalsForest
+          SJust p1 `shouldBe` p1'
```
Soupstraw: I think you could also compare the whole subtree at once instead of comparing each node manually, should save some lines.

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Proposals.hs
```diff
@@ -0,0 +1,102 @@
...
+      prop "Removing leaf nodes keeps Proposals consistent" $
+        \(ps :: Proposals Conway) ->
+          let test gais =
+                let ps' = fst $ proposalsRemoveDescendentIds gais ps
+                 in do
+                      proposalsSize ps' `shouldBe` proposalsSize ps - Set.size gais
+                      ps' `shouldSatisfy` proposalsAreConsistent
+           in case proposalsIds ps of
+                Empty -> True `shouldBe` True
+                _pids :|> pid :|> pid1 :|> pid2 :|> pid3 -> test $ Set.fromList [pid, pid1, pid2, pid3]
```
Soupstraw: I think it'd help with readability if there were functions like `takeInit` and `takeTail` that take the first n elements from either the beginning or the end of the seq.

eras/conway/impl/test/Test/Cardano/Ledger/Conway/GovActionReorderSpec.hs
aniketd: `shuffle` is a `Gen` monad here, while the test expects just an `IO`, after having moved the generator to the `Arbitrary` module. The generated list of actions is arbitrary and unsorted or in other words, shuffled already. I have checked this by eye-balling the generated list. I think this  should be fine as it is, since we already test that the sorting should result in the same operation as that of using `reorderActions`. What do you think?

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/GovSpec.hs
aniketd: I tried, but there seem to be multiple rabbit holes to go through. :smile:

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/ImpTest.hs
aniketd: I have changed this to directly use `Conway` instead of a type-variable. For  some reason, and I tried fixing it with the double constraints for pattern synonyms as well, these result in GHC attempting to unify two different variables.

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/GovSpec.hs
aniketd: I'm not sure how to do this, since the trees are being constructed in parts and we would need to reconstruct the tree any way to be able to compare the whole with another whole. Can you give me an example of how you mean we could do this?

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/GovSpec.hs
lehins: Well, we need to jump through them. Because the tests that we are writing now, must be usable for future eras.

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/GovSpec.hs
Soupstraw: I see now that the problem is that one of the trees is polymorphic over `StrictMaybe` of `GovActionId` and the other is just `GovActionId`. You'd probably have to map over the second tree with `SJust` and then you could do a direct comparison on the whole tree.

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/GovSpec.hs
```diff
@@ -44,20 +54,823 @@ import Test.Cardano.Ledger.Imp.Common
...
+          SJust p2 `shouldBe` p2'
+          SJust p11 `shouldBe` p11'
+          SJust p21 `shouldBe` p21'
+          SJust p111 `shouldBe` p111'
+          SJust p112 `shouldBe` p112'
```
Soupstraw:
```suggestion
          a <-
            submitConstitutionGovActionForest
              SNothing
              [ E
                  [ E
                      [ E []
                      , E []
                      ]
                  ]
              , E
                  [ E []
                  ]
              ]
          [  I []
          , I []
          , J _ [] -- Due to the committee enactment
          , I b
          ] <-
            getProposalsForest
          fmap SJust a `shouldBe` b
```

Would this work?

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/GovSpec.hs
lehins: > I think you could also compare the whole subtree at once instead of comparing each node manually, should save some lines.

> .. You'd probably have to map over the second tree ...

@aniketd That is exactly what I was suggesting you during our 1on1 as well

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/GovSpec.hs
aniketd: Awesome! Thank you! Yes, this would work I will do this in as many places as is possible. We also discuss your point during our 1 on 1, Alexey and I :+1:

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/ImpTest.hs
aniketd: I got rid of the pattern synonyms entirely. :+1:

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/GovSpec.hs
aniketd: Type application at a single place eradicated the need for these constraints. :bow: :raised_hands: . Thank you very much!

eras/conway/impl/test/Test/Cardano/Ledger/Conway/GovActionReorderSpec.hs
aniketd: @Soupstraw do you still think we need a `shuffle` here in the test, or is it okay to get the shuffled stuff from the generator itself?

Soupstraw: We need to check that two different shuffles still give the same thing when reordered, so the shuffle is still necessary.

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/GovSpec.hs
aniketd: Done as much as I could do :+1:

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/GovSpec.hs
```diff
@@ -59,6 +71,680 @@ spec =
...
+            , Node
+                SNothing
+                [ Node p1' [c]
+                  , d
+                  ]
+            ] <-
+            getProposalsForest
+          SJust p1 `shouldBe` p1'
+          SJust <$> a `shouldBe` c
+          SJust <$> b `shouldBe` d
```
lehins: Why not using this approach?
```suggestion
          getProposalsForest `shouldReturn`
            [ Node SNothing []
            , Node SNothing []
            , Node SNothing []
            , Node
                SNothing
                [ Node (SJust p1) [SJust <$> a]
                , SJust <$> b
                ]
            ]
```

eras/conway/impl/test/Test/Cardano/Ledger/Conway/GovActionReorderSpec.hs
aniketd: @Soupstraw I would like to discuss this property with you when we can. 

I doubt if this property should hold, because the sorting is on a surjection (namely `actionPriority`): meaning that there are multiple elements in the list that will have the same `actionPriority` and so if we shuffle arbitrarily, we can still get a different order after sorting.

I don't know how this was working earlier though, I will check. And I'm not sure if I'm correct either. So let's discuss on Slack.

```haskell
data ShuffledGovActionStates era = 
  ShuffledGovActionStates [GovActionState era] [GovActionState era]
  deriving Show

instance 
  (Era era, Arbitrary (PParamsUpdate era)) => 
  Arbitrary (ShuffledGovActionStates era) 
  where
  arbitrary = do
    gass <- arbitrary
    shuffledGass <- shuffle gass
    pure $ ShuffledGovActionStates gass shuffledGass

-- [ ... ]

prop "orders actions correctly" $
  \(ShuffledGovActionStates @Conway gass shuffledGass) -> do
    let sortedGass = sortOn (actionPriority . gasAction) gass
    -- Set.fromList gass `shouldBe` Set.fromList shuffledGass
    -- Set.fromList sortedGass `shouldBe` Set.fromList shuffledGass
    Seq.fromList sortedGass `shouldBe` reorderActions (Seq.fromList shuffledGass)
```
This test fails.

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/GovSpec.hs
aniketd: :exploding_head: :100: Will do this!!!

eras/conway/impl/test/Test/Cardano/Ledger/Conway/GovActionReorderSpec.hs
Soupstraw: Ok, I went over the original test again and it looks like the idea behind this test is that each type of action in the list appears only once. You're right that this property wouldn't hold when we have more than one of some type of proposal.

aniketd: Sorry I missed that earlier @Soupstraw. Thank you! :raised_hands:

eras/conway/impl/test/Test/Cardano/Ledger/Conway/GovActionReorderSpec.hs
```diff
@@ -1,8 +1,37 @@
...
+         in filter filterPrio (toList $ reorderActions @Conway (a Seq.:<| as))
+              `shouldBe` filter filterPrio (toList $ reorderActions (a Seq.:<| as))
+    prop "orders actions correctly with shuffles" $
+      \(ShuffledGovActionStates @Conway gass shuffledGass) -> do
+        let sortedGass = sortOn (actionPriority . gasAction) gass
+        Seq.fromList sortedGass `shouldBe` reorderActions (Seq.fromList shuffledGass)
+    prop "orders actions correctly" $
+      \(gass :: [GovActionState Conway]) -> do
+        let sortedGass = sortOn (actionPriority . gasAction) gass
+        Seq.fromList sortedGass `shouldBe` reorderActions (Seq.fromList gass)
```
Soupstraw:
```suggestion
    prop "orders actions correctly with shuffles" $
      \(ShuffledGovActionStates @Conway gass shuffledGass) -> do
        reorderActions (Seq.fromList gass) `shouldBe` reorderActions (Seq.fromList shuffledGass)
```
I'm not sure why `reorderActions` was inlined here in the first place

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Proposals.hs
```diff
@@ -0,0 +1,110 @@
...
+  _pids :|> pid -> [pid]
+  Empty -> []
+
+takeInit :: StrictSeq (GovActionId c) -> [GovActionId c]
+takeInit = \case
+  pid :<| pid1 :<| pid2 :<| pid3 :<| _pids -> [pid, pid1, pid2, pid3]
+  pid :<| pid1 :<| pid2 :<| _pids -> [pid, pid1, pid2]
+  pid :<| pid1 :<| _pids -> [pid, pid1]
+  pid :<| _pids -> [pid]
+  Empty -> []
```
Soupstraw: Why not generalize those to take an arbitrary number of elements?

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp.hs
```diff
@@ -25,7 +23,9 @@ spec ::
   forall era.
   ( ConwayEraImp era
   , GovState era ~ ConwayGovState era
-  , Inject (ConwayGovPredFailure era) (PredicateFailure (EraRule "LEDGER" era))
+  , EraCrypto era ~ StandardCrypto
```
lehins: Why is this constraint necessary? It should never appear in polymorphic ledger code
```suggestion
```

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/GovSpec.hs
```diff
@@ -45,8 +56,10 @@ import Test.Cardano.Ledger.Imp.Common hiding (Success)
 spec ::
   forall era.
   ( ConwayEraImp era
+  , EraCrypto era ~ StandardCrypto
```
lehins: Same here
```suggestion
```

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/GovSpec.hs
```diff
@@ -45,8 +56,10 @@ import Test.Cardano.Ledger.Imp.Common hiding (Success)
...
   , GovState era ~ ConwayGovState era
-  , Inject (ConwayGovPredFailure era) (PredicateFailure (EraRule "LEDGER" era))
+  , PredicateFailure (EraRule "LEDGER" era) ~ ConwayLedgerPredFailure era
+  , PredicateFailure (EraRule "GOV" era) ~ ConwayGovPredFailure era
```
lehins, line 60: You need to use `inject` on the predicate failures in order to satisfy constraints
```suggestion
  , Inject (ConwayGovPredFailure era) (PredicateFailure (EraRule "LEDGER" era))
```

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/GovSpec.hs
```diff
@@ -59,6 +72,565 @@ spec =
...
+                ProposalProcedure
+                  { pProcDeposit = pp ^. ppGovActionDepositL
+                  , pProcReturnAddr = RewardAccount Testnet (KeyHashObj khPropRwd)
+                  , pProcGovAction = constitutionAction
+                  , pProcAnchor = def
+                  }
+          -- submitGovAction_ constitutionAction
+          submitFailingProposal
+            constitutionProposal
+            [ ConwayGovFailure $ InvalidPrevGovActionId constitutionProposal
```
lehins, line 110: Like so
```suggestion
            [ inject $ ConwayGovFailure $ InvalidPrevGovActionId constitutionProposal
```

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/GovSpec.hs
```diff
@@ -59,6 +72,565 @@ spec =
       it "Hardfork minorFollow" (secondHardForkFollows minorFollow)
       it "Hardfork majorFollow" (secondHardForkFollows majorFollow)
       it "Hardfork cantFollow" secondHardForkCantFollow
+    let _submitInvalidChildConstitutionGovAction govActionId = do
```
lehins: Never use underscores for bindings, unless they are really needed for some reason, despite being unused. So why hasn't it been removed, or used?
```suggestion
    let submitInvalidChildConstitutionGovAction govActionId = do
```

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/GovSpec.hs
```diff
@@ -59,6 +72,565 @@ spec =
...
+                  }
+          -- submitGovAction_ constitutionAction
+          submitFailingProposal
+            constitutionProposal
+            [ ConwayGovFailure $ InvalidPrevGovActionId constitutionProposal
+            ]
+    context "Proposal-forests" $ do
+      context "GOV, EPOCH, RATIFY and ENACT" $ do
+        it "Proposals survive multiple epochs without any activity" $ do
+          modifyPParams $ ppGovActionLifetimeL .~ EpochInterval 4 -- + 2 epochs to pass to get the desired effect
```
lehins: Line is too long
```suggestion
          -- + 2 epochs to pass to get the desired effect:
          modifyPParams $ ppGovActionLifetimeL .~ EpochInterval 4
```

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/ImpTest.hs
```diff
@@ -915,3 +923,103 @@ proposalsShowDebug ps showRoots =
...
+              ( Anchor
+                  (fromJust $ textToUrl 64 "constitution.dummy.0")
+                  constitutionHash
+              )
+              SNothing
+          )
+  submitGovAction constitutionAction
+
+getProposalsForest ::
+  (ConwayEraGov era, EraCrypto era ~ StandardCrypto) =>
```
lehins:
```suggestion
  ConwayEraGov era=>
```

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/ImpTest.hs
```diff
@@ -915,3 +923,103 @@ proposalsShowDebug ps showRoots =
...
+      Forest (StrictMaybe (GovActionId (EraCrypto era)))
+    mkForest forestL ps =
+      let h = ps ^. pGraphL . forestL . pGraphNodesL
+          s = SSeq.fromStrict $ proposalsIds ps
+          getOrderedChildren cs = toList $ Seq.filter (`Set.member` Set.map unGovPurposeId cs) s
+          go c = (SJust c, getOrderedChildren $ h Map.! GovPurposeId c ^. peChildrenL)
+       in unfoldForest go (getOrderedChildren $ ps ^. pRootsL . forestL . prChildrenL)
+
+submitConstitutionGovActionTree ::
+  (ShelleyEraImp era, ConwayEraTxBody era, ToExpr (PParamsHKD Identity era)) =>
```
lehins: Add `ToExpr (PParamsHKD Identity era)` to `ShelleyEraImp` instead
```suggestion
  (ShelleyEraImp era, ConwayEraTxBody era) =>
```

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/ImpTest.hs
```diff
@@ -915,3 +923,103 @@ proposalsShowDebug ps showRoots =
...
+  (ShelleyEraImp era, ConwayEraTxBody era) =>
+  ImpTestM era (RewardAccount (EraCrypto era), GovActionId (EraCrypto era))
+submitTreasuryWithdrawalsGovAction = do
+  rewardAccount <- registerRewardAccount @era
+  let withdrawalAction =
+        TreasuryWithdrawals
+          (Map.singleton rewardAccount $ Coin 1_000_000)
+          SNothing
+  gaid <- submitGovAction withdrawalAction
+  pure (rewardAccount, gaid)
```
lehins: This function is unused and we already have `submitTreasuryWithdrawals`
```suggestion
```

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/GovSpec.hs
```diff
@@ -59,6 +72,565 @@ spec =
...
+                           ]
+          passNEpochs 3
+          getProposalsForest
+            `shouldReturn` [ Node SNothing []
+                           , Node SNothing []
+                           , Node SNothing []
+                           , Node SNothing [SJust <$> b]
+                           ]
+        it "Subtrees are pruned when competing proposals are enacted" $ do
+          (dRep, committeeMember, _) <- electBasicCommittee
```
lehins:
```suggestion
          (dRep, committeeMember, committeeGovActionId) <- electBasicCommittee
```

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/GovSpec.hs
```diff
@@ -59,6 +72,565 @@ spec =
...
+                  ]
+              ]
+
+          [ Node SNothing []
+            , Node SNothing []
+            , Node _ [] -- Due to the committee enactment
+            , Node SNothing c
+            ] <-
+            getProposalsForest
+          fmap SJust <$> a `shouldBe` c
```
lehins: Can you do this instead?
```suggestion
          getProposalsForest `shouldReturn`
            [ Node SNothing []
            , Node SNothing []
            , Node (SJust committeeGovActionId) [] -- Due to the committee enactment
            , Node SNothing (SJust <$> a)
            ] 
```

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/GovSpec.hs
```diff
@@ -59,6 +72,565 @@ spec =
...
+          submitYesVote_ (DRepVoter dRep) p2
+          submitYesVote_ (CommitteeVoter committeeMember) p2
+          passNEpochs 2
+          [ Node SNothing []
+            , Node SNothing []
+            , Node _ []
+            , d
+            ] <-
+            getProposalsForest
+          SJust <$> b `shouldBe` d
```
lehins: Same here, just reuse the `committeeGovActionId`

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/ImpTest.hs
```diff
@@ -915,3 +923,103 @@ proposalsShowDebug ps showRoots =
...
+      StrictMaybe (GovActionId (EraCrypto era))
+    mkRoot lenz ps = fmap unGovPurposeId $ ps ^. pRootsL . lenz . prRootL
+    mkForest ::
+      (forall f. Lens' (PForest f era) (f (GovPurposeId p era))) ->
+      Proposals era ->
+      Forest (StrictMaybe (GovActionId (EraCrypto era)))
+    mkForest forestL ps =
+      let h = ps ^. pGraphL . forestL . pGraphNodesL
+          s = SSeq.fromStrict $ proposalsIds ps
+          getOrderedChildren cs = toList $ Seq.filter (`Set.member` Set.map unGovPurposeId cs) s
```
lehins: Not that it is terribly important for these tests, just making a tip for future.

Avoid redundant conversions.
```suggestion
          s = toList $ proposalsIds ps
          getOrderedChildren cs = filter (`Set.member` Set.map unGovPurposeId cs) s
```

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/ImpTest.hs
```diff
@@ -915,3 +923,103 @@ proposalsShowDebug ps showRoots =
...
+submitConstitutionGovActionForest p forest =
+  unfoldForestM go $ fmap (fmap $ const p) forest
+  where
+    go (Node parent children) = do
+      n <- submitConstitutionGovAction $ GovPurposeId <$> parent
+      pure (n, fmap (\(Node _child subtree) -> Node (SJust n) subtree) children)
+
+mkCorruptGovActionId :: GovActionId c -> GovActionId c
+mkCorruptGovActionId (GovActionId txi (GovActionIx gaix)) =
+  GovActionId txi $ GovActionIx $ gaix + 999
```
lehins: This should be a local binding. We don't need a global helper like this

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/ImpTest.hs
```diff
@@ -915,3 +923,103 @@ proposalsShowDebug ps showRoots =
...
+    , Node (mkRoot pfHardForkL ps) $ mkForest pfHardForkL ps
+    , Node (mkRoot pfCommitteeL ps) $ mkForest pfCommitteeL ps
+    , Node (mkRoot pfConstitutionL ps) $ mkForest pfConstitutionL ps
+    ]
+  where
+    mkRoot ::
+      Lens' (PForest PRoot era) (PRoot (GovPurposeId p era)) ->
+      Proposals era ->
+      StrictMaybe (GovActionId (EraCrypto era))
+    mkRoot lenz ps = fmap unGovPurposeId $ ps ^. pRootsL . lenz . prRootL
```
lehins:
```suggestion
    mkRoot rootL ps = fmap unGovPurposeId $ ps ^. pRootsL . rootL . prRootL
```

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Proposals.hs
lehins: Why not just use `toList` and do those operations on list?

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Proposals.hs
```diff
@@ -0,0 +1,110 @@
...
+           in ps' `shouldSatisfy` isRight . toPForestEither
+    context "Removal" $ do
+      prop "Removing leaf nodes keeps Proposals consistent" $
+        \(ps :: Proposals Conway) ->
+          let test gais =
+                let ps' = fst $ proposalsRemoveDescendentIds gais ps
+                 in do
+                      proposalsSize ps' `shouldBe` proposalsSize ps - Set.size gais
+                      ps' `shouldSatisfy` isRight . toPForestEither
+           in test $ Set.fromList $ takeTail $ proposalsIds ps
```
lehins: What are those nested lets for?
```suggestion
        \(ps :: Proposals Conway) -> do
          let gais = Set.fromList $ takeTail $ proposalsIds ps
              ps' = fst $ proposalsRemoveDescendentIds gais ps
          proposalsSize ps' `shouldBe` proposalsSize ps - Set.size gais
```

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Proposals.hs
```diff
@@ -0,0 +1,110 @@
...
+      prop "Removing root nodes keeps Proposals consistent" $
+        \(ps :: Proposals Conway) ->
+          let test gais =
```
lehins: Same here, simplify in the same manner as suggested above

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Proposals.hs
```diff
@@ -0,0 +1,110 @@
...
+  pid :<| _pids -> [pid]
+  Empty -> []
+
+spec :: Spec
+spec = do
+  describe "Proposals" $ do
+    context "Construction" $ do
+      prop "Generator generates consistent Proposals" $
+        \(ps :: Proposals Conway) ->
+          ps `shouldSatisfy` isRight . toPForestEither
```
lehins: This test is useless. `toPForest` is invoked on every addition and removal from proposals. So whenever `ps` is generated, it is already called.
```suggestion
```

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Proposals.hs
```diff
@@ -0,0 +1,110 @@
...
+      prop "Adding new nodes keeps Proposals consistent" $
+        \(ProposalsNewActions ps actions :: ProposalsNewActions Conway) ->
+          let ps' =
+                foldl'
+                  (\p action -> fromMaybe (error "Unable to add action") $ proposalsAddAction action p)
+                  ps
+                  actions
+           in ps' `shouldSatisfy` isRight . toPForestEither
```
lehins: Again, this is useless, since `proposalsAddAction` calls `toPForest` 

How about checking that addition actually happened correctly, eg:
```haskell
actions `shouldBe` (actions `Map.intersecton` OMap.toMap (pProps ps))
```

lehins (commented):
I only got a chance to look at parts of the PR.
I'll do a follow up review tomorrow

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Proposals.hs
Soupstraw: we actually have `take` and `takeLast` for `StrictSeq`

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/GovSpec.hs
aniketd: This was an important one. I added a simple test use this.

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Proposals.hs
```diff
@@ -0,0 +1,88 @@
...
+import Data.Sequence.Strict (StrictSeq)
+import qualified Data.Set as Set
+import Test.Cardano.Ledger.Common
+import Test.Cardano.Ledger.Conway.Arbitrary (
+  ProposalsForEnactment (..),
+  ProposalsNewActions (..),
+ )
+
+takeTail :: StrictSeq (GovActionId c) -> [GovActionId c]
+takeTail = reverse . take 4 . reverse . toList
```
aniketd: Can there be a better way? :sweat_smile:

lehins: Sure:
```suggestion
takeTail n = toList . StrictSeq.takeLast n
```

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Proposals.hs
```diff
@@ -0,0 +1,88 @@
...
+          proposalsSize ps' `shouldBe` proposalsSize ps - Set.size gais
+      prop "Removing root nodes keeps Proposals consistent" $
+        \(ps :: Proposals Conway) -> do
+          let gais = Set.fromList $ takeInit $ proposalsIds ps
+              ps' = fst $ proposalsRemoveWithDescendants gais ps
+          proposalsSize ps' `shouldSatisfy` (<= proposalsSize ps)
+      prop "Removing non-member nodes throws an AssertionFailure" $
+        \(ProposalsNewActions ps actions :: ProposalsNewActions Conway) ->
+          (evaluate . force) (proposalsRemoveWithDescendants (Set.fromList $ gasId <$> actions) ps)
+            `shouldThrow` anyException
```
lehins: You should almost never expect any exception, cause that could mask some critical bug that throws some other exception that you did not expect.

Make sure to apply this suggestion in all places where `anyException` is used
```suggestion
          evaluate (force (proposalsRemoveWithDescendants (Set.fromList $ gasId <$> actions) ps))
            `shouldThrow` \AssertionFailed {} -> True
```

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/ImpTest.hs
```diff
@@ -13,6 +18,11 @@
 {-# LANGUAGE UndecidableInstances #-}
 {-# LANGUAGE UndecidableSuperClasses #-}
 {-# OPTIONS_GHC -Wno-orphans #-}
+#if __GLASGOW_HASKELL__ < 963
+-- `ToExpr (PParamsHKD Identity era)` is reported as redundant otherwise
+{-# OPTIONS_GHC -Wno-redundant-constraints #-}
+#else
+#endif
```
lehins: CPP should be the last resort in solving a problem
```suggestion
```

Add it to `ShelleyEraImp` type class as superclass constraint that will solve your problem.

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/ImpTest.hs
```diff
@@ -594,13 +596,14 @@ expectMissingGovActionId ::
   GovActionId (EraCrypto era) ->
   ImpTestM era ()
 expectMissingGovActionId govActionId =
+  -- FIXME: @aniketd also check that the id isn't present in enacted or children
```
lehins: No need. That would be a violation of an invariant which would result in an `assert` failure
```suggestion
```

lehins (approved):
Looks much better than before. Nice.

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Proposals.hs
aniketd: Thank you! :sweat_smile:

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Proposals.hs
aniketd: Awesome! Thank you!

PR #3982, opened 2024-01-10
Added SPO voting thresholds for security relevant parameters

### LINKED ISSUE #3925: StakePool PParams security group
CIP-1694 is getting ammended with addition of a security group for PParams that Stake Pool Operatos will be able to vote on: https://github.com/cardano-foundation/CIPs/pull/622

Once the above PR is merged and spec is adjusted we can implement it in Ledger.

I envision it being implemented in a similar fashion that DRep thresholds and groups are implemented. THKD would just get an extra argument of a new kind `StakePoolGroup` with two values ('SecurityGroup` and `NoGroup`)

Soupstraw:
# Description

This PR adds a way to mark protocol parameter fields as `SecurityRelevant`. The security relevant fields require sufficient votes from SPOs, where the required votes are determined by the `pvtSecurityRelevant` parameter.

resolves #3925 

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/eras/conway/impl/src/Cardano/Ledger/Conway/Governance.hs
+++ b/eras/conway/impl/src/Cardano/Ledger/Conway/Governance.hs
@@ -211,9 +211,11 @@ import Cardano.Ledger.Conway.Governance.Proposals (
  )
 import Cardano.Ledger.Conway.PParams (
   ConwayEraPParams (..),
+  DRepGroup (..),
   DRepVotingThresholds (..),
-  PPGroup (..),
+  PPGroups (..),
   PoolVotingThresholds (..),
+  StakePoolGroup (..),
   dvtPPEconomicGroupL,
   dvtPPGovGroupL,
   dvtPPNetworkGroupL,
@@ -677,7 +679,7 @@ pparamsUpdateThreshold pp ppu =
         GovGroup -> dvtPPGovGroupL
         TechnicalGroup -> dvtPPTechnicalGroupL
         EconomicGroup -> dvtPPEconomicGroupL
-      lookupGroupThreshold grp =
+      lookupGroupThreshold (PPGroups grp _) =
         pp ^. ppDRepVotingThresholdsL . thresholdLens grp
    in Set.foldr' max minBound $
         Set.map lookupGroupThreshold $
@@ -738,7 +740,16 @@ votingStakePoolThresholdInternal pp isElectedCommittee action =
         { pvtCommitteeNoConfidence
         , pvtCommitteeNormal
         , pvtHardForkInitiation
+        , pvtPPSecurityGroup
         } = pp ^. ppPoolVotingThresholdsL
+      isSecurityRelevant (PPGroups _ s) =
+        case s of
+          SecurityGroup -> True
+          NoStakePoolGroup -> False
+      paramChangeThreshold ppu
+        | any isSecurityRelevant (modifiedPPGroups ppu) =
+            VotingThreshold pvtPPSecurityGroup
+        | otherwise = NoVotingAllowed
    in case action of
         NoConfidence {} -> VotingThreshold pvtCommitteeNoConfidence
         UpdateCommittee {} ->
@@ -748,7 +759,7 @@ votingStakePoolThresholdInternal pp isElectedCommittee action =
               else pvtCommitteeNoConfidence
         NewConstitution {} -> NoVotingAllowed
         HardForkInitiation {} -> VotingThreshold pvtHardForkInitiation
-        ParameterChange {} -> NoVotingAllowed
+        ParameterChange _ ppu _ -> paramChangeThreshold ppu
         TreasuryWithdrawals {} -> NoVotingAllowed
         InfoAction {} -> NoVotingThreshold
 
--- a/eras/conway/impl/src/Cardano/Ledger/Conway/PParams.hs
+++ b/eras/conway/impl/src/Cardano/Ledger/Conway/PParams.hs
@@ -1,6 +1,5 @@
 {-# LANGUAGE AllowAmbiguousTypes #-}
 {-# LANGUAGE DataKinds #-}
-{-# LANGUAGE DeriveFunctor #-}
 {-# LANGUAGE DeriveGeneric #-}
 {-# LANGUAGE DerivingVia #-}
 {-# LANGUAGE FlexibleContexts #-}
@@ -41,7 +40,11 @@ module Cardano.Ledger.Conway.PParams (
   ppuDRepDepositL,
   ppuDRepActivityL,
   PoolVotingThresholds (..),
+  pvtCommitteeNoConfidenceL,
+  pvtCommitteeNormalL,
+  pvtPPSecurityGroupL,
   DRepVotingThresholds (..),
+  dvtCommitteeNoConfidenceL,
   dvtPPNetworkGroupL,
   dvtPPGovGroupL,
   dvtPPTechnicalGroupL,
@@ -55,7 +58,9 @@ module Cardano.Ledger.Conway.PParams (
   UpgradeConwayPParams (..),
   toUpgradeConwayPParamsUpdatePairs,
   THKD (..),
-  PPGroup (..),
+  DRepGroup (..),
+  PPGroups (..),
+  StakePoolGroup (..),
   conwayModifiedPPGroups,
 )
 where
@@ -110,7 +115,7 @@ import NoThunks.Class (NoThunks (..))
 import Numeric.Natural (Natural)
 
 class BabbageEraPParams era => ConwayEraPParams era where
-  modifiedPPGroups :: PParamsUpdate era -> Set PPGroup
+  modifiedPPGroups :: PParamsUpdate era -> Set PPGroups
   ppuWellFormed :: PParamsUpdate era -> Bool
 
   hkdPoolVotingThresholdsL :: HKDFunctor f => Lens' (PParamsHKD f era) (HKD f PoolVotingThresholds)
@@ -185,8 +190,18 @@ data PoolVotingThresholds = PoolVotingThresholds
   , pvtCommitteeNormal :: !UnitInterval
   , pvtCommitteeNoConfidence :: !UnitInterval
   , pvtHardForkInitiation :: !UnitInterval
+  , pvtPPSecurityGroup :: !UnitInterval
   }
   deriving (Eq, Ord, Show, Generic)
+
+pvtCommitteeNormalL :: Lens' PoolVotingThresholds UnitInterval
+pvtCommitteeNormalL = lens pvtCommitteeNormal (\x y -> x {pvtCommitteeNormal = y})
+
+pvtCommitteeNoConfidenceL :: Lens' PoolVotingThresholds UnitInterval
+pvtCommitteeNoConfidenceL = lens pvtCommitteeNoConfidence (\x y -> x {pvtCommitteeNoConfidence = y})
+
+pvtPPSecurityGroupL :: Lens' PoolVotingThresholds UnitInterval
+pvtPPSecurityGroupL = lens pvtPPSecurityGroup (\x y -> x {pvtPPSecurityGroup = y})
 
 instance NoThunks PoolVotingThresholds
 
@@ -194,16 +209,17 @@ instance NFData PoolVotingThresholds where
   rnf = rwhnf
 
 instance Default PoolVotingThresholds where
-  def = PoolVotingThresholds def def def def
+  def = PoolVotingThresholds def def def def def
 
 instance ToJSON PoolVotingThresholds where
-  toJSON pvt@(PoolVotingThresholds _ _ _ _) =
+  toJSON pvt@(PoolVotingThresholds _ _ _ _ _) =
     let PoolVotingThresholds {..} = pvt
      in object
           [ ("motionNoConfidence", toJSON pvtMotionNoConfidence)
           , ("committeeNormal", toJSON pvtCommitteeNormal)
           , ("committeeNoConfidence", toJSON pvtCommitteeNoConfidence)
           , ("hardForkInitiation", toJSON pvtHardForkInitiation)
+          , ("ppSecurityGroup", toJSON pvtPPSecurityGroup)
           ]
 
 instance FromJSON PoolVotingThresholds where
@@ -214,22 +230,25 @@ instance FromJSON PoolVotingThresholds where
         <*> o .: "committeeNormal"
         <*> o .: "committeeNoConfidence"
         <*> o .: "hardForkInitiation"
+        <*> o .: "ppSecurityGroup"
 
 instance EncCBOR PoolVotingThresholds where
   encCBOR PoolVotingThresholds {..} =
-    encodeListLen 4
+    encodeListLen 5
       <> encCBOR pvtMotionNoConfidence
       <> encCBOR pvtCommitteeNormal
       <> encCBOR pvtCommitteeNoConfidence
       <> encCBOR pvtHardForkInitiation
+      <> encCBOR pvtPPSecurityGroup
 
 instance DecCBOR PoolVotingThresholds where
   decCBOR =
-    decodeRecordNamed "PoolVotingThresholds" (const 4) $ do
+    decodeRecordNamed "PoolVotingThresholds" (const 5) $ do
       pvtMotionNoConfidence <- decCBOR
       pvtCommitteeNormal <- decCBOR
       pvtCommitteeNoConfidence <- decCBOR
       pvtHardForkInitiation <- decCBOR
+      pvtPPSecurityGroup <- decCBOR
       pure $ PoolVotingThresholds {..}
 
 data DRepVotingThresholds = DRepVotingThresholds
@@ -300,6 +319,9 @@ dvtPPGovGroupL = lens dvtPPGovGroup (\x y -> x {dvtPPGovGroup = y})
 dvtUpdateToConstitutionL :: Lens' DRepVotingThresholds UnitInterval
 dvtUpdateToConstitutionL = lens dvtUpdateToConstitution (\x y -> x {dvtUpdateToConstitution = y})
 
+dvtCommitteeNoConfidenceL :: Lens' DRepVotingThresholds UnitInterval
+dvtCommitteeNoConfidenceL = lens dvtCommitteeNoConfidence (\x y -> x {dvtCommitteeNoConfidence = y})
+
 instance EncCBOR DRepVotingThresholds where
   encCBOR DRepVotingThresholds {..} =
     encodeListLen 10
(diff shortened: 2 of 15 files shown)


eras/conway/impl/src/Cardano/Ledger/Conway/Governance.hs
```diff
@@ -736,7 +738,16 @@ votingStakePoolThresholdInternal pp isElectedCommittee action =
         { pvtCommitteeNoConfidence
         , pvtCommitteeNormal
         , pvtHardForkInitiation
+        , pvtSecurityRelevant
         } = pp ^. ppPoolVotingThresholdsL
+      isSecurityRelevant (PInfo _ s) =
+        case s of
```
teodanciu, line 746: You could simplify as `s == SecurityRelevant` (unless you explicitly wanted to enumerate all cases )

Soupstraw: I kind of prefer case matching, because that forces anyone that changes the type to revisit this piece of code. But it probably doesn't matter here since I doubt we're going to add any constructors to `Security`.

eras/conway/impl/src/Cardano/Ledger/Conway/PParams.hs
```diff
@@ -329,6 +351,10 @@ instance DecCBOR DRepVotingThresholds where
       dvtTreasuryWithdrawal <- decCBOR
       pure $ DRepVotingThresholds {..}
 
+data ParameterInfo
+  = PInfo PPGroup Security
```
lehins: I'd like to see a bit more consistency in naming. Also "info" is a pretty bad name for the concept. "Groups" is much better and it reflects the spec and the CIP. `Security` name by itself is too general and it does not explain in any way that it is only specific to SPOs
```suggestion
data PPGroups
  = PPGroups DRepGroup StakePoolGroup
```

eras/conway/impl/src/Cardano/Ledger/Conway/PParams.hs
```diff
@@ -337,6 +363,11 @@ data PPGroup
   | GovGroup
   deriving (Eq, Ord, Show)
 
+data Security
+  = SecurityRelevant
+  | Regular
+  deriving (Eq, Ord, Show)
```
lehins: Let's call it a group, rather than "relevant"
```suggestion
data StakePoolGroup
  = SecurityGroup
  | NoStakePoolGroup
  deriving (Eq, Ord, Show)
```

eras/conway/impl/src/Cardano/Ledger/Conway/PParams.hs
```diff
@@ -349,8 +380,16 @@ instance ToPPGroup 'TechnicalGroup where
 instance ToPPGroup 'GovGroup where
   toPPGroup = GovGroup
 
+class ToSecurity (t :: Security) where
```
lehins:
```suggestion
class ToStakePoolGroup (t :: StakePoolGroup) where
```

eras/conway/impl/src/Cardano/Ledger/Conway/PParams.hs
```diff
@@ -1191,8 +1230,8 @@ conwayApplyPPUpdates pp ppu =
         SNothing -> THKD curCostModel
         SJust costModelUpdate -> THKD $ updateCostModels curCostModel costModelUpdate
 
-conwayModifiedPPGroups :: ConwayPParams StrictMaybe era -> Set PPGroup
-conwayModifiedPPGroups
+conwayModifiedParameterInfos :: ConwayPParams StrictMaybe era -> Set ParameterInfo
+conwayModifiedParameterInfos
```
lehins: I believe infos is a bad name for this concept. Let's stick with the group, as it is in the spec and the CIP.
```suggestion
conwayModifiedPPGroups :: ConwayPParams StrictMaybe era -> Set ParameterInfo
conwayModifiedPPGroups
```

eras/conway/impl/test/data/conway-genesis.json
```diff
@@ -3,7 +3,8 @@
     "committeeNormal": 0,
     "committeeNoConfidence": 0,
     "hardForkInitiation": 0,
-    "motionNoConfidence": 0
+    "motionNoConfidence": 0,
+    "securityRelevant": 0
```
lehins: Consistency with DRep threshold
```suggestion
    "ppSecurityGroup": 0
```

eras/conway/impl/src/Cardano/Ledger/Conway/PParams.hs
```diff
@@ -110,7 +115,7 @@ import NoThunks.Class (NoThunks (..))
 import Numeric.Natural (Natural)
 
 class BabbageEraPParams era => ConwayEraPParams era where
-  modifiedPPGroups :: PParamsUpdate era -> Set PPGroup
+  modifiedParameterInfos :: PParamsUpdate era -> Set ParameterInfo
```
lehins:
```suggestion
  modifiedPPGroups :: PParamsUpdate era -> Set ParameterInfo
```

eras/conway/impl/src/Cardano/Ledger/Conway/PParams.hs
```diff
@@ -399,10 +438,10 @@ instance (Typeable t, ToJSON a) => ToJSON (THKD t StrictMaybe a) where
 instance (Typeable t, FromJSON a) => FromJSON (THKD t StrictMaybe a) where
   parseJSON = fmap THKD . parseJSON
 
-ppGroup :: forall t a. ToPPGroup t => THKD t StrictMaybe a -> Set PPGroup
-ppGroup = \case
+paramInfos :: forall t s a. (ToPPGroup t, ToSecurity s) => THKD ('PInfo t s) StrictMaybe a -> Set ParameterInfo
+paramInfos = \case
```
lehins:
```suggestion
ppGroups :: forall t s a. (ToPPGroup t, ToSecurity s) => THKD ('PInfo t s) StrictMaybe a -> Set ParameterInfo
ppGroups = \case
```

eras/conway/impl/src/Cardano/Ledger/Conway/PParams.hs
```diff
@@ -644,7 +683,7 @@ instance Crypto c => BabbageEraPParams (ConwayEra c) where
     lens (unTHKD . cppCoinsPerUTxOByte) $ \pp x -> pp {cppCoinsPerUTxOByte = THKD x}
 
 instance Crypto c => ConwayEraPParams (ConwayEra c) where
-  modifiedPPGroups (PParamsUpdate ppu) = conwayModifiedPPGroups ppu
+  modifiedParameterInfos (PParamsUpdate ppu) = conwayModifiedParameterInfos ppu
```
lehins: Bad rename. Please bring back
```suggestion
  modifiedPPGroups (PParamsUpdate ppu) = conwayModifiedPPGroups ppu
```

eras/conway/impl/src/Cardano/Ledger/Conway/PParams.hs
```diff
@@ -214,22 +230,25 @@ instance FromJSON PoolVotingThresholds where
         <*> o .: "committeeNormal"
         <*> o .: "committeeNoConfidence"
         <*> o .: "hardForkInitiation"
+        <*> o .: "securityRelevant"
```
lehins: Inconsistent naming. We need to follow the DRep thresholds example.
```suggestion
        <*> o .: "ppSecurityGroup"
```

eras/conway/impl/src/Cardano/Ledger/Conway/PParams.hs
```diff
@@ -185,25 +190,36 @@ data PoolVotingThresholds = PoolVotingThresholds
   , pvtCommitteeNormal :: !UnitInterval
   , pvtCommitteeNoConfidence :: !UnitInterval
   , pvtHardForkInitiation :: !UnitInterval
+  , pvtSecurityRelevant :: !UnitInterval
```
lehins: Follow the DRep naming
```suggestion
  , pvtPPSecurityGroup :: !UnitInterval
```

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/GovSpec.hs
```diff
@@ -81,6 +83,123 @@ spec =
...
+        modifyPParams $ \pp ->
+          pp
+            & ppDRepVotingThresholdsL . dvtCommitteeNoConfidenceL .~ 1 %! 2
+            & ppPoolVotingThresholdsL . pvtCommitteeNoConfidenceL .~ 1 %! 2
+            & ppCommitteeMaxTermLengthL .~ EpochInterval 200
+        let
+          getCommittee =
+            getsNES $
+              nesEsL . esLStateL . lsUTxOStateL . utxosGovStateL . cgEnactStateL . ensCommitteeL
+          assertNoCommittee =
```
lehins, line 99: Without explicit `HasCallStack constraint, you won't know which occurrence triggered the failure
```suggestion
          assertNoCommittee :: HasCallStack => ImpTestM era ()
          assertNoCommittee =
```

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/GovSpec.hs
```diff
@@ -81,6 +83,123 @@ spec =
...
+          electCommittee
+            SNothing
+            drep
+            mempty
+            committeeMap
+        khSPO <- setupPoolWithStake $ Coin 42_000_000
+        logStakeDistr
+        submitYesVote_ (StakePoolVoter khSPO) gaidCommittee
+        replicateM_ 4 passEpoch
+        void . impAnn "Committee should be elected" $ do
```
lehins: What's the `void` for?
```suggestion
        impAnn "Committee should be elected" $ do
```

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/GovSpec.hs
```diff
@@ -81,6 +83,123 @@ spec =
...
+          assertNoCommittee =
+            do
+              committee <- getCommittee
+              impAnn "There should not be a committee" $ committee `shouldBe` SNothing
+        assertNoCommittee
+        khCC <- freshKeyHash
+        drep <- setupSingleDRep
+        let committeeMap =
+              Map.fromList
+                [ (khCC, EpochNo 50)
```
lehins: I think we need to make `electCommittee` accept `Credential`, instead of `KeyHash`, since we will soon have script support. This will also avoid this `mapKeys` below `(Map.mapKeys KeyHashObj committeeMap)`
```suggestion
                [ (KeyHashObj khCC, EpochNo 50)
```

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/GovSpec.hs
```diff
@@ -81,6 +83,123 @@ spec =
...
+              , pProcAnchor = def
+              }
+        submitYesVote_ (StakePoolVoter khSPO) gaidNoConf
+        submitYesVote_ (DRepVoter $ KeyHashObj drep) gaidNoConf
+        replicateM_ 4 passEpoch
+        assertNoCommittee
+      it "SPO needs to vote on security-relevant parameter changes" $ do
+        (drep, ccCred) <- electBasicCommittee
+        khPool <- setupPoolWithStake $ Coin 42_000_000
+        gaidThreshold <- do
```
lehins: A little annotation could be useful, I think.
```suggestion
        gaidThreshold <- impAnn "Update StakePool thresholds" $ do
```

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/GovSpec.hs
```diff
@@ -81,6 +83,123 @@ spec =
...
+                }
+          submitYesVote_ (DRepVoter drep) gaidMinFee
+          submitYesVote_ (CommitteeVoter ccCred) gaidMinFee
+          pure gaidMinFee
+        passEpoch
+        logAcceptedRatio gaidMinFee
+        passEpoch
+        do
+          pp <- getsNES $ nesEsL . curPParamsEpochStateL
+          (pp ^. ppMinFeeAL) `shouldBe` Coin 44
```
lehins: This locks us in into the fact that initial value is `44`. We should avoid hard coding values like this. It is better to record the original value for the `initMinFeeA` at the beginning of the test.

```suggestion
          (pp ^. ppMinFeeAL) `shouldBe` initMinFeeA
```

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/GovSpec.hs
```diff
@@ -81,6 +83,123 @@ spec =
...
+          submitYesVote_ (StakePoolVoter khPool) gaidMinFee
+        passEpoch
+        logStakeDistr
+        logAcceptedRatio gaidMinFee
+        logRatificationChecks gaidMinFee
+        passEpoch
+        pp <- getsNES $ nesEsL . curPParamsEpochStateL
+        (pp ^. ppMinFeeAL) `shouldBe` Coin 12_345
```
lehins: Same here. Use a binding for `newMinFeeA` instead at the beginning of the test for both update and verification
```suggestion
        (pp ^. ppMinFeeAL) `shouldBe` newMinFeeA
```

libs/cardano-ledger-core/src/Cardano/Ledger/Tools.hs
```diff
@@ -137,6 +139,9 @@ estimateMinFeeTx pp tx numKeyWits numByronKeyWits = setMinFeeTx pp tx' ^. bodyTx
           , Byron.aaNetworkMagic = Byron.NetworkTestnet maxBound
           }
 
+integerToByteStringN :: Int -> Integer -> (ByteString, Maybe Integer)
+integerToByteStringN len = BS.unfoldrN len (\n -> Just (fromInteger n, n `shiftR` 8))
```
lehins: Good idea on extracting this into a utility function. Although, I doubt that left over `Integer` will be useful. Also we can make it more general, it doesn't have to be `Integer`.
```suggestion
integralToByteStringN :: (Integral i, Bits i) => Int -> i -> ByteString
integralToByteStringN len = fst . BS.unfoldrN len (\n -> Just (fromIntegral n, n `shiftR` 8))
```

eras/shelley/impl/testlib/Test/Cardano/Ledger/Shelley/ImpTest.hs
```diff
@@ -178,6 +182,7 @@ data ImpTestState era = ImpTestState
   , impLastTick :: !SlotNo
   , impGlobals :: !Globals
   , impLog :: !(Doc ())
+  , impHashVerKeyVRFIdx :: !Integer
```
lehins: Why do we need a separate index? We could just have one index for all hashes.

So, renaming `impSafeHashIdx` to `impHashIdx` and just use that.

Less state we have to keep around the better.

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/ImpTest.hs
```diff
@@ -553,15 +590,29 @@ calculateCommitteeAcceptedRatio gaId = do
...
       eNo
 
+calculatePoolAcceptedRatio :: ConwayEraGov era => GovActionId (EraCrypto era) -> ImpTestM era Rational
+calculatePoolAcceptedRatio gaId = do
+  ratEnv <- getRatifyEnv
+  gas <- getGovActionState gaId
+  pure $
+    spoAcceptedRatio
+      ratEnv
+      gas
```
lehins: It looks like Chinese, words are written from top to bottom :rofl:  
```suggestion
  pure $ spoAcceptedRatio ratEnv gas
```

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/ImpTest.hs
```diff
@@ -362,6 +391,14 @@ submitProposal ::
   ImpTestM era (GovActionId (EraCrypto era))
 submitProposal proposal = trySubmitProposal proposal >>= expectRightExpr
 
+submitProposal_ ::
+  ( ShelleyEraImp era
+  , ConwayEraTxBody era
+  ) =>
+  ProposalProcedure era ->
+  ImpTestM era ()
+submitProposal_ = void . submitProposal
```
lehins: Great minds think alike :wink:  I also added this function in one of my PRs :slightly_smiling_face:

eras/shelley/impl/testlib/Test/Cardano/Ledger/Shelley/ImpTest.hs
```diff
@@ -740,6 +775,17 @@ freshKeyHash = do
   addKeyPair kp
   pure $ hashKey kh
 
+giveCoin ::
+  ShelleyEraImp era =>
+  Addr (EraCrypto era) ->
+  Coin ->
```
lehins: Let's make this function a little bit more general and better named. We have plans on transitioning previous eras to `ImpTest` too. It is possible to create more specific functions out of more general ones, but not the other way around.
```suggestion
sendCoinTo ::
  ShelleyEraImp era =>
  Addr (EraCrypto era) ->
  Coin ->
  ImpTestM era ()
sendCoinTo addr = sendValueTo addr . inject

sendValueTo ::
  ShelleyEraImp era =>
  Addr (EraCrypto era) ->
  Value era ->
```

lehins (commented):
Logic looks great. I am not too fond of some naming choices, but that should be easy to fix.

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/ImpTest.hs
Soupstraw: Yeah, I noticed that as well :smile:

eras/shelley/impl/testlib/Test/Cardano/Ledger/Shelley/ImpTest.hs
```diff
@@ -303,11 +311,33 @@ logStakeDistr = do
...
+    seedSize = fromIntegral . seedSizeDSIGN $ Proxy @(DSIGN (EraCrypto era))
+
+freshKeyHashVRF ::
+  forall era.
+  ShelleyEraImp era =>
+  ImpTestM era (Hash (HASH (EraCrypto era)) (VerKeyVRF (EraCrypto era)))
+freshKeyHashVRF = do
+  idx <- dispenseIndex
+  let res = mkHashVerKeyVRF @era $ fromIntegral idx
+  pure res
```
```suggestion
  pure $ mkHashVerKeyVRF @era $ fromIntegral idx
```

lehins (approved):
I've discussed this with @WhatisRT and current behavior implemented in this PR is indeed the desired behavior. 

Great work!

PR #3983, opened 2024-01-10
Include reference scripts size in min fee calculation

### LINKED ISSUE #3952: Addition of new `minFeeRefScriptCoinsPerByte` protocol parameter
Fee calculation need to be adjusted in Conway. It has been decided that size of reference scripts should contribute to the fee amount charged in a transaction.

* Addition of new protocol parameter: `minFeeRefScripts :: NonNegativeInterval`
* Adding `UTxO` argument to `getMinFeeTx`:
```haskell
getMinFeeTx :: UTxO era -> PParams era -> Tx era -> Coin
```
* Adding another factor to the calculation in Babbage, along the lines of:
```haskell
  ...
  <+> Coin (round ((refScriptsTotalSize % 1) * unboundRational (pp ^. ppMinFeeRefScriptsL)))
  where
    refSciptsTotalSize =
      maybe 0 (toInteger . getSum) $ foldMap (fmap Sum . getPlutusScriptSize) refScripts
    txBody = tx ^. bodyTxL
    ins = (txBody ^. referenceInputsTxBodyL) `Set.union` (txBody ^. inputsTxBodyL)
    refScripts = getReferenceScripts utxo ins
```
Where `getPlutusScriptSize :: Script era -> Maybe Int` is a function that calculates the size of a plutus script for any `Script`

teodanciu:
# Description

Resolves https://github.com/IntersectMBO/cardano-ledger/issues/3952

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/eras/babbage/impl/src/Cardano/Ledger/Babbage/UTxO.hs
+++ b/eras/babbage/impl/src/Cardano/Ledger/Babbage/UTxO.hs
@@ -8,6 +8,7 @@ module Cardano.Ledger.Babbage.UTxO (
   getBabbageSpendingDatum,
   getBabbageScriptsProvided,
   getReferenceScripts,
+  getReferenceScriptsNonDistinct,
 ) where
 
 import Cardano.Ledger.Alonzo.TxWits (unTxDats)
@@ -25,7 +26,7 @@ import Cardano.Ledger.Binary (sizedValue)
 import Cardano.Ledger.Crypto
 import Cardano.Ledger.Mary.UTxO (getConsumedMaryValue)
 import Cardano.Ledger.Plutus.Data (Data)
-import Cardano.Ledger.Shelley.UTxO (shelleyProducedValue)
+import Cardano.Ledger.Shelley.UTxO (getShelleyMinFeeTxUtxo, shelleyProducedValue)
 import Cardano.Ledger.TxIn (TxIn)
 import Cardano.Ledger.UTxO (EraUTxO (..), ScriptsProvided (..), UTxO (..))
 import Control.Applicative
@@ -52,6 +53,8 @@ instance Crypto c => EraUTxO (BabbageEra c) where
   getScriptsHashesNeeded = getAlonzoScriptsHashesNeeded
 
   getWitsVKeyNeeded = getAlonzoWitsVKeyNeeded
+
+  getMinFeeTxUtxo pp tx _ = getShelleyMinFeeTxUtxo pp tx
 
 instance Crypto c => AlonzoEraUTxO (BabbageEra c) where
   getSupplementalDataHashes = getBabbageSupplementalDataHashes
@@ -139,9 +142,15 @@ getReferenceScripts ::
   UTxO era ->
   Set (TxIn (EraCrypto era)) ->
   Map.Map (ScriptHash (EraCrypto era)) (Script era)
-getReferenceScripts (UTxO mp) inputs = Map.foldl' accum Map.empty (eval (inputs ◁ mp))
-  where
-    accum ans txOut =
-      case txOut ^. referenceScriptTxOutL of
-        SNothing -> ans
-        SJust script -> Map.insert (hashScript script) script ans
+getReferenceScripts utxo ins = Map.fromList (getReferenceScriptsNonDistinct utxo ins)
+
+getReferenceScriptsNonDistinct ::
+  BabbageEraTxOut era =>
+  UTxO era ->
+  Set (TxIn (EraCrypto era)) ->
+  [(ScriptHash (EraCrypto era), Script era)]
+getReferenceScriptsNonDistinct (UTxO mp) inputs =
+  [ (hashScript script, script)
+  | txOut <- Map.elems (eval (inputs ◁ mp))
+  , SJust script <- [txOut ^. referenceScriptTxOutL]
+  ]
--- a/eras/conway/impl/CHANGELOG.md
+++ b/eras/conway/impl/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## 1.13.0.0
 
+* Add implementation for `getMinFeeTxUtxo`
 * Add `cppMinFeeRefScriptCoinsPerByte` to `ConwayPParams` and `ppMinFeeRefScriptCoinsPerByteL`
 * Add `ucppMinFeeRefScriptCoinsPerByte` to `UpgradeConwayPParams` and `ppuMinFeeRefScriptCoinsPerByteL`
 * Fix `ConwayTxBody` pattern synonym, by changing its certificates arguments to `OSet`
--- a/eras/conway/impl/CHANGELOG.md
+++ b/eras/conway/impl/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## 1.13.0.0
 
+* Add `cppMinFeeRefScriptCoinsPerByte` to `ConwayPParams` and `ppMinFeeRefScriptCoinsPerByteL`
+* Add `ucppMinFeeRefScriptCoinsPerByte` to `UpgradeConwayPParams` and `ppuMinFeeRefScriptCoinsPerByteL`
 * Fix `ConwayTxBody` pattern synonym, by changing its certificates arguments to `OSet`
   from a `StrictSeq`.
 * Add `VotingPurpose` and `ProposingPurpose` pattern synonyms
--- a/eras/conway/impl/src/Cardano/Ledger/Conway/PParams.hs
+++ b/eras/conway/impl/src/Cardano/Ledger/Conway/PParams.hs
@@ -31,6 +31,7 @@ module Cardano.Ledger.Conway.PParams (
   ppGovActionDepositL,
   ppDRepDepositL,
   ppDRepActivityL,
+  ppMinFeeRefScriptCoinsPerByteL,
   ppuPoolVotingThresholdsL,
   ppuDRepVotingThresholdsL,
   ppuCommitteeMinSizeL,
@@ -39,6 +40,7 @@ module Cardano.Ledger.Conway.PParams (
   ppuGovActionDepositL,
   ppuDRepDepositL,
   ppuDRepActivityL,
+  ppuMinFeeRefScriptCoinsPerByteL,
   PoolVotingThresholds (..),
   pvtCommitteeNoConfidenceL,
   pvtCommitteeNormalL,
@@ -126,6 +128,7 @@ class BabbageEraPParams era => ConwayEraPParams era where
   hkdGovActionDepositL :: HKDFunctor f => Lens' (PParamsHKD f era) (HKD f Coin)
   hkdDRepDepositL :: HKDFunctor f => Lens' (PParamsHKD f era) (HKD f Coin)
   hkdDRepActivityL :: HKDFunctor f => Lens' (PParamsHKD f era) (HKD f EpochInterval)
+  hkdMinFeeRefScriptCoinsPerByteL :: HKDFunctor f => Lens' (PParamsHKD f era) (HKD f CoinPerByte)
 
 ppPoolVotingThresholdsL ::
   forall era. ConwayEraPParams era => Lens' (PParams era) PoolVotingThresholds
@@ -152,6 +155,9 @@ ppDRepDepositL = ppLens . hkdDRepDepositL @era @Identity
 
 ppDRepActivityL :: forall era. ConwayEraPParams era => Lens' (PParams era) EpochInterval
 ppDRepActivityL = ppLens . hkdDRepActivityL @era @Identity
+
+ppMinFeeRefScriptCoinsPerByteL :: forall era. ConwayEraPParams era => Lens' (PParams era) CoinPerByte
+ppMinFeeRefScriptCoinsPerByteL = ppLens . hkdMinFeeRefScriptCoinsPerByteL @era @Identity
 
 ppuPoolVotingThresholdsL ::
   forall era. ConwayEraPParams era => Lens' (PParamsUpdate era) (StrictMaybe PoolVotingThresholds)
@@ -184,6 +190,10 @@ ppuDRepDepositL = ppuLens . hkdDRepDepositL @era @StrictMaybe
 ppuDRepActivityL ::
   forall era. ConwayEraPParams era => Lens' (PParamsUpdate era) (StrictMaybe EpochInterval)
 ppuDRepActivityL = ppuLens . hkdDRepActivityL @era @StrictMaybe
+
+ppuMinFeeRefScriptCoinsPerByteL ::
+  forall era. ConwayEraPParams era => Lens' (PParamsUpdate era) (StrictMaybe CoinPerByte)
+ppuMinFeeRefScriptCoinsPerByteL = ppuLens . hkdMinFeeRefScriptCoinsPerByteL @era @StrictMaybe
 
 data PoolVotingThresholds = PoolVotingThresholds
   { pvtMotionNoConfidence :: !UnitInterval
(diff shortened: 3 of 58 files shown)


eras/conway/impl/src/Cardano/Ledger/Conway/PParams.hs
```diff
@@ -516,6 +526,8 @@ data ConwayPParams f era = ConwayPParams
   -- ^ The amount of a DRep registration deposit
   , cppDRepActivity :: !(THKD ('PPGroups 'GovGroup 'NoStakePoolGroup) f EpochInterval)
   -- ^ The number of Epochs that a DRep can perform no activity without losing their @Active@ status.
+  , cppMinFeeRefScriptCoinsPerByte :: !(THKD ('PPGroups 'EconomicGroup 'SecurityGroup) f CoinPerByte)
+  -- ^ Reference scripts fee for the minimum fee calculation
```
lehins, line 530: You need to add this field in the translation to plutus here https://github.com/IntersectMBO/cardano-ledger/blob/996231f0bc288aa5fdf7cdb501b03ca782e79aea/eras/conway/impl/src/Cardano/Ledger/Conway/Plutus/Context.hs#L197

libs/cardano-ledger-core/src/Cardano/Ledger/SafeHash.hs
```diff
@@ -142,6 +143,9 @@ class SafeToHash t where
   -- | Extract the original bytes from 't'
   originalBytes :: t -> ByteString
 
+  originalBytesSize :: t -> Int
+  originalBytesSize = BS.length . originalBytes
```
lehins, line 148: Overwrite this default implementation in the instance for `ShortByteString` few lines below.
This will avoid a memcopy whenever we just want to get the size

Same needs to be done with `SafeToHash` instance for `MemoBytes`

libs/cardano-ledger-core/src/Cardano/Ledger/Tools.hs
```diff
@@ -47,14 +48,32 @@ import Lens.Micro
...
+setMinFeeTxUtxo :: EraUTxO era => PParams era -> Tx era -> UTxO era -> Tx era
+setMinFeeTxUtxo pp tx utxo =
+  _setMinFeeTx (\t -> getMinFeeTxUtxo pp t utxo) tx
+
+_setMinFeeTx ::
+  EraTx era =>
+  (Tx era -> Coin) ->
+  Tx era ->
+  Tx era
+_setMinFeeTx f tx =
```
lehins: Never use `_` prefix, unless you have a binding that you really need to keep around for some reason, despite that it being unused.
Underscore prefix is the instruction for the compiler to not warn you about something not being used, so it is dangerous to prefix with underscore things that are being used, because a simple spelling mistake can hide a bug. Small example that depicts a problem:
```haskell
foo _a _b _c _d _o _p = bar _a _b _c _d _c _p
```
Compiler will not warn you about the problem above

It is better to add a suffix `Internal`, if there is a need to emphasize that a function should not be for public consumption.

eras/babbage/impl/src/Cardano/Ledger/Babbage/UTxO.hs
```diff
@@ -137,9 +141,17 @@ getReferenceScripts ::
...
-getReferenceScripts (UTxO mp) inputs = Map.foldl' accum Map.empty (eval (inputs ◁ mp))
+getReferenceScripts utxo ins = Map.fromList (getReferenceScriptsNonDistinct utxo ins)
+
+getReferenceScriptsNonDistinct ::
+  BabbageEraTxOut era =>
+  UTxO era ->
+  Set (TxIn (EraCrypto era)) ->
+  [(ScriptHash (EraCrypto era), Script era)]
+getReferenceScriptsNonDistinct (UTxO mp) inputs =
+  Foldable.foldl' accum [] (Map.elems (eval (inputs ◁ mp)))
```
lehins: This will avoid the reversing artifact of foldl. Also it is a bit simpler, IMHO:
```suggestion
  [(hashScript script, script)
   | txOut <- Map.elems (eval (inputs ◁ mp))
   , SJust script <- [txOut ^. referenceScriptTxOutL]
   ]
```

eras/shelley/impl/src/Cardano/Ledger/Shelley/API/Wallet.hs
```diff
@@ -446,7 +446,7 @@ getRewardProvenance globals newepochstate =
 
 -- | The minimum fee calculation.
 -- Used for the default implentation of 'evaluateTransactionFee'.
-evaluateMinFee :: EraTx era => PParams era -> Tx era -> Coin
+evaluateMinFee :: EraTx era => PParams era -> Tx era -> Int -> Coin
 evaluateMinFee = getMinFeeTx
 {-# DEPRECATED evaluateMinFee "In favor of 'getMinFeeTx'" #-}
```
lehins: We should remove this function.

Whenever type signature for a deprecated has to change, there is no longer a point to keep it around. Because the only reason deprecated functions are useful is for backwards compatibility.

Alternative would to: `evaluateMinFee pp tx = getMinFeeTx pp tx 0`, but it has been deprecated for so long that there is no longer a reason for keeping it around

eras/shelley/impl/src/Cardano/Ledger/Shelley/API/Wallet.hs
```diff
@@ -468,6 +468,8 @@ evaluateTransactionFee ::
   Tx era ->
   -- | The number of key witnesses still to be added to the transaction.
   Word ->
+  -- | The total size of reference scripts
+  Int ->
```
lehins: Ditto. This function could be removed too, however, this is an alternative implementation that we could reuse for testing somehow. So, maybe you can put somewhere in a core testlib. I believe this function is used in one of the tests already, so it would be sad if we had to remove tests because of this.

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/UtxoSpec.hs
```diff
@@ -0,0 +1,140 @@
...
+      checkMinFee spendingScript [spendingScript]
+
+    it "reference scripts not required for spending the input count towards the minFee calculation" $ do
+      spendingScript <- mkScript
+      extraScripts <- sequence [mkScript, mkScript, mkScript]
+      checkMinFee spendingScript $ spendingScript : extraScripts
+
+    it "a scripts referenced several times counts for each reference towards the minFee calculation" $ do
+      spendingScript <- mkScript
+      extraScripts <- sequence [mkScript, mkScript]
```
lehins:
```suggestion
      extraScripts <- replicateM 2 mkScript
```

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/UtxoSpec.hs
```diff
@@ -0,0 +1,140 @@
...
+  SpecWith (ImpTestState era)
+spec = describe "UTxO" $ do
+  describe "Reference scripts" $ do
+    it "required reference script counts towards the minFee calculation" $ do
+      spendingScript <- mkScript
+      checkMinFee spendingScript [spendingScript]
+
+    it "reference scripts not required for spending the input count towards the minFee calculation" $ do
+      spendingScript <- mkScript
+      extraScripts <- sequence [mkScript, mkScript, mkScript]
```
lehins:
```suggestion
      extraScripts <- replicateM 3 mkScript
```

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/UtxoSpec.hs
```diff
@@ -0,0 +1,140 @@
...
+              <&> ( \s ->
+                      mkBasicTxOut @era (rootOut ^. addrTxOutL) (inject (Coin 100))
+                        & referenceScriptTxOutL @era .~ SJust (fromNativeScript s)
+                  )
+      tx <-
+        submitTx . mkBasicTx $
+          mkBasicTxBody
+            & outputsTxBodyL @era
+              .~ SSeq.fromList outs
+      let refIns = (\i -> txInAt (fromIntegral i) tx) <$> [1 .. length scripts]
```
lehins: TxIns are zero based indexed, so I think we should change how we fixup the transaction. First of all this code confused me for a moment, secondly theoretically it is possible for txOut with change to not be added.
I believe the change txOut should go at the end of the outputs, rather than a te the beginning. This will make tests more reliable IMHO.

This would cause this code to be changed to proper 0-base index:
(note that it could also be `[0..]`, since you are using it with `zip`)
```suggestion
      let refIns = (\i -> txInAt (fromIntegral i) tx) <$> [0 .. length scripts - 1]
```

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/UtxoSpec.hs
```diff
@@ -0,0 +1,140 @@
...
+  where
+    checkMinFee :: NativeScript era -> [NativeScript era] -> ImpTestM era ()
+    checkMinFee scriptToSpend refScripts = do
+      refScriptFee <- setRefScriptFee
+      -- lock an input with a script
+      scriptSpendIn <- createScriptUtxo scriptToSpend
+      -- create outputs with reference scripts and the return them mapped to their corresponding inputs
+      refScriptInToScripts <- createRefScriptsUtxos refScripts
+      -- spend the initial input by passing the reference scripts
+      tx <- spendScriptUsingRefScripts scriptSpendIn $ Map.keys refScriptInToScripts
```
lehins: You actually need a Set over in `spendScriptUsingRefScripts` 
```suggestion
      tx <- spendScriptUsingRefScripts scriptSpendIn $ Map.keysSet refScriptInToScripts
```

eras/conway/impl/src/Cardano/Ledger/Conway/Tx.hs
```diff
@@ -66,6 +72,20 @@ instance Crypto c => EraTx (ConwayEra c) where
...
+  ( EraTx era
+  , AlonzoEraTxWits era
+  , ConwayEraPParams era
+  ) =>
+  PParams era ->
+  Tx era ->
+  Int ->
+  Coin
+getConwayMinFeeTx pp tx refScriptsSize =
+  alonzoMinFeeTx pp tx <+> refScriptsSize <×> refScriptFee
```
lehins: That is one of the very few times when I will suggest redundant parenthesis :grin: 
```suggestion
  alonzoMinFeeTx pp tx <+> (refScriptsSize <×> refScriptFee)
```

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/UtxoSpec.hs
```diff
@@ -0,0 +1,140 @@
...
+    setRefScriptFee :: ImpTestM era Coin
+    setRefScriptFee = do
+      let refScriptFee = Coin 10
+      modifyPParams $ ppMinFeeRefScriptCoinsPerByteL .~ CoinPerByte refScriptFee
+      pure refScriptFee
+
+    txInAt :: Integer -> Tx era -> TxIn (EraCrypto era)
+    txInAt index tx =
+      let txid = txIdTx tx
+       in mkTxInPartial txid index
```
lehins: It could be a useful function for testing in general, so maybe we should put it into core testlib. Also type signature could be more general and with `HasCallStack`
```suggestion
    txInAt :: (HasCallStack, Integral i) => i -> Tx era -> TxIn (EraCrypto era)
    txInAt index tx =
      let txid = txIdTx tx
       in mkTxInPartial txid (toInteger index)
```

lehins (approved):
This is great. Few suggestions, but nothing critical.

eras/conway/impl/src/Cardano/Ledger/Conway/PParams.hs
teodanciu: Didn't know about this, thanks

lehins: Yeah, it is a very recent addition

eras/conway/impl/src/Cardano/Ledger/Conway/Tx.hs
teodanciu: Good point, and not even hlint complains about it,

lehins: > not even hlint complains about it

I think that is because those are non-standard operators that are defined in another package (instead of this module) and hlint is not aware of their infix settings. So it can't know if those parenthesis are redundant or not. 

Which is sort of the same reason why I suggested those parenthesis. It was also because those operators are non-standard and the reader might not know the precedence.

eras/shelley/impl/src/Cardano/Ledger/Shelley/API/Wallet.hs
teodanciu: Indeed, it's used in two tests: `Fees` in shelley-test and `AlonzoAPI` in cardano-ledger-test.

What do you think about replacing there the usage of `evaluateTransactionFee` with `estimateMinFeeTx`  and passing zero as number of byron witnesses ? (since the former doesn't set `bootAddrTxWitsL`) 
In this way we'll have more tests for the function that we're exposing, `estimateMinFeeTx` and fewer variations of this function.

lehins: > What do you think about replacing there the usage of evaluateTransactionFee with estimateMinFeeTx

Yeah, let's do that. `evaluateTransactionFee` is just dumber version of `estimateMinFeeTx`

eras/babbage/impl/src/Cardano/Ledger/Babbage/UTxO.hs
teodanciu: Yes, much nicer, thanks

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/UtxoSpec.hs
teodanciu: Added a commit that changes the fixup. I agree, it's more intuitive  and likely more robust.

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/UtxoSpec.hs
teodanciu: I have extracted it in `Test.Cardano.Ledger.Core.Utils`, but that place seems like the "miscellaneous drawer" everyone has in the house that is rarely useful. 
Couldn't find a better place in testlib. I was tempted to add it in `Core` just below the definitions of `txIdTx` and `txIdTxBody`,  to make it easier to find, but because it's using `mkTxInPartial` which has a warning to be used only in tests, I  decided against it. 
Happy to move it to better place than `Utils`

lehins: Sounds like as good of a place as any, at least for the time being. Maybe we'll find a better home for it in the future.
Thank you.

eras/conway/impl/CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## 1.13.0.0
 
+* Add `cppMinFeeRefScriptCoinsPerByte` to `ConwayPParams` and `cppMinFeeRefScriptCoinsPerByteL`
```
```suggestion
* Add `cppMinFeeRefScriptCoinsPerByte` to `ConwayPParams` and `ppMinFeeRefScriptCoinsPerByteL`
```

PR #3986, opened 2024-01-11
Remove libs/cardano-ledger-pretty

TimSheard:
removed libs/cardano-ledger-pretty
Follow on to PR #3973
This library is no longer supported. Instead a slightly improved version is made part of llibs/cardano-ledger-test

A tiny fix that disallows Ptrs in the stake credential of Conway Addresses.
# Checklist

- [ ] Self-reviewed the diff


CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ If you are looking for the Ledger Releasing and Versioning Process then you can
 
 ## 8.7
 
+- Remove the library libs/cardano-ledger-pretty
```
lehins: This time it is the right CHANGELOG :slightly_smiling_face: , but not the right version.

Next version that will be releaed that is affected is 8.8

```suggestion
## 8.8

- Remove the library libs/cardano-ledger-pretty

## 8.7
```

lehins (approved):
When removing or adding a package or a section in a package we need to run `./scripts/gen-hie.sh`, which will make the necessary adjustments.

PR #3989, opened 2024-01-11
Fix too many discards.

TimSheard:
Figure out how to get non-empty poolDistr from "random" SnapShots.

The test  "Bruteforce = Pulsed, in every epoch, on traces of length 150" 
Some times fails because at the epoch boundary because "There are no stakepools to choose an issuer from"
We set this up to discard, but occasionally we reach 150 discards and the test fails.
A lot of analysis discovers that in the random initial NewEpochState, the mark snapshot is not well formed, in that when ones computes the PoolDistr from such a snap shot, PoolDistr is empty.

The fix is to arrange random snapshots such that they never result in an empty PoolDistr. So we had to change the constraints for 'regPools' , 'incrementalStake', and 'delegations' so this never happens. In order to do that
we had to change the order we solve constraints, so that 'utxo' and  'incementalStake' are solved before 'delegations' . To do that we created a new Stage  utxoStage, in the module Preds/UTxO.hs, 
so we can see the UTxO when specifying the DState and PState.

# Description

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/libs/cardano-ledger-test/cardano-ledger-test.cabal
+++ b/libs/cardano-ledger-test/cardano-ledger-test.cabal
@@ -37,6 +37,7 @@ library
         Test.Cardano.Ledger.Constrained.Preds.Certs
         Test.Cardano.Ledger.Constrained.Preds.LedgerState
         Test.Cardano.Ledger.Constrained.Preds.NewEpochState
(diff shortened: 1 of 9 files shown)


lehins (approved):
Thank you for fixing it!

PR #3991, opened 2024-01-15
Imp native script support

lehins:
# Description

This PR adds native script support for ImpTests framework. It works by automatically adding script witnesses for all things in the transaction that need them and satisfying necessary key witnesses for scripts to validate.

It works like this:
```haskell
  keyHash <- freshKeyHash
  scriptHash <- impAddNativeScript $ RequireAllOf (SSeq.singleton (RequireSignature keyHash))
```
At this point whenever any transaction tries to unlock anything that is locked by a `scriptHash` transaction, the fixup mechanism will automatically add the necessary witnesses.

# Checklist

- [x] Self-reviewed the diff


eras/allegra/impl/testlib/Test/Cardano/Ledger/Allegra/ImpTest.hs
```diff
@@ -24,4 +32,39 @@ instance
...
+        Nothing -> satisfyMOf m xs
+        Just kps -> do
+          kps' <- satisfyMOf (m - 1) xs
+          Just $ kps <> kps'
+    satisfyScript = \case
+      RequireSignature keyHash -> do
+        keyPair <- Map.lookup keyHash keyPairs
+        Just $ Map.singleton keyHash keyPair
+      RequireAllOf ss -> satisfyMOf (length ss) ss
+      RequireAnyOf ss -> satisfyMOf 1 ss
```
Soupstraw, line 63: It'd be good to shuffle `ss` once we add `MonadGen` instance to `ImpTestM`

eras/shelley/impl/testlib/Test/Cardano/Ledger/Shelley/ImpTest.hs
```diff
@@ -357,7 +393,28 @@ instance
...
+            kps' <- satisfyMOf (m - 1) xs
+            Just $ kps <> kps'
+      satisfyScript = \case
+        RequireSignature keyHash -> do
+          keyPair <- Map.lookup keyHash keyPairs
+          Just $ Map.singleton keyHash keyPair
+        RequireAllOf ss -> satisfyMOf (length ss) ss
+        RequireAnyOf ss -> satisfyMOf 1 ss
+        RequireMOf m ss -> satisfyMOf m ss
+    pure $ satisfyScript script
```
Soupstraw, line 417: Could we reuse some of this in `impAllegraSatisfyNativeScript`?

lehins: Not at this point. I have plans for improving how we deal with native scripts which would allow us to reuse this logic. That will come in the future, when we have a bit more time. However, at this point, `MultiSigs` and `Timelocks` are two totally separate types that just happen to have pattern synonyms with same names.

eras/allegra/impl/testlib/Test/Cardano/Ledger/Allegra/ImpTest.hs
lehins: Yes, I thought about this as well. Adding deterministic randomness to imp tests would be hella useful. Something we'll definitely add over time. Although, not sure what is this `MonadGen` thing you are talking about.

Soupstraw: I meant something like this, but there are probably more ways to approach this problem:
https://hackage.haskell.org/package/quickcheck-transformer-0.3.1.2/docs/Test-QuickCheck-GenT.html#t:MonadGen

Soupstraw (approved):
LGTM

eras/allegra/impl/testlib/Test/Cardano/Ledger/Allegra/ImpTest.hs
lehins: Ah, yeah, that`MonadGen`. Something like that would definitely work.
First we need to figure out how to get our hands on QuickCheck's QGen in the `ImpTestM`, the rest will be easy. I know it is possible, I just need to experiment with a little.

Soupstraw: Yeah, that's work for another PR.

PR #3993, opened 2024-01-16
Change Test.Cardano.Ledger.Generic.Proof so that there is no more Crypto Evidence.

TimSheard:
Proof no longer takes an Evidence parameter (choosing which crypto, either Standard or Mock). Everything now runs in Standard Crypto. This simplifies many things, because we no longer need to pattern match against (Alonzo _) but just Alonzo (for example).

# Description

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/libs/cardano-ledger-test/src/Test/Cardano/Ledger/Alonzo/Tools.hs
+++ b/libs/cardano-ledger-test/src/Test/Cardano/Ledger/Alonzo/Tools.hs
@@ -47,7 +47,7 @@ import Test.Cardano.Ledger.Examples.STSTestUtils (
   someAddr,
   someKeys,
  )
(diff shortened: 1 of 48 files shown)


lehins (changes requested):
Could I ask you please go through the changes in the PR and remove all of the redundant parenthesis.

It'll take me a whole day to make suggestions on the PR

TimSheard:
So I ran a bunch of commands like
`sd "(Shelley)" "Shelley" $(git ls-files . -- *.hs)`
which turned out to be no-ops since "(" and ")" have special meaning in regular expressions.
I think it s fixed now.

lehins (approved):
Beautiful! Thank you!

PR #3996, opened 2024-01-17
Index aware PlutusPurpose

lehins:
# Description

* Add `AsIxItem`, which allows us having a `PlutusPurpose` be both item and index aware, thus avoiding the reverse lookup of an item in order to find its index.
* Fix `ConwayTxBody` pattern synonym, by changing its certificates arguments to `OSet`
  from a `StrictSeq`
* Add `SpendingPurpose`, `MintingPurpose`, `CertifyingPurpose`, `RewardingPurpose`, 
`VotingPurpose` and `ProposingPurpose` pattern synonyms to allow era agnostic handling of `PlutusPurpose`

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Rules/Utxow.hs
+++ b/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Rules/Utxow.hs
@@ -36,7 +36,7 @@ import Cardano.Ledger.Alonzo.Rules.Utxo (
   AlonzoUtxoEvent,
   AlonzoUtxoPredFailure,
  )
-import Cardano.Ledger.Alonzo.Scripts (plutusScriptLanguage)
+import Cardano.Ledger.Alonzo.Scripts (plutusScriptLanguage, toAsIndex, toAsItem)
 import Cardano.Ledger.Alonzo.Tx (hashScriptIntegrity)
 import Cardano.Ledger.Alonzo.TxWits (
   unRedeemers,
@@ -227,11 +227,9 @@ hasExactSetOfRedeemers ::
   AlonzoScriptsNeeded era ->
   Test (AlonzoUtxowPredFailure era)
 hasExactSetOfRedeemers tx (ScriptsProvided scriptsProvided) (AlonzoScriptsNeeded scriptsNeeded) = do
-  let txBody = tx ^. bodyTxL
-      redeemersNeeded =
-        [ (rp, (sp, sh))
+  let redeemersNeeded =
+        [ (hoistPlutusPurpose toAsIndex sp, (hoistPlutusPurpose toAsItem sp, sh))
         | (sp, sh) <- scriptsNeeded
-        , SJust rp <- [redeemerPointer @era txBody sp]
         , Just script <- [Map.lookup sh scriptsProvided]
         , not (isNativeScript script)
         ]
--- a/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Scripts.hs
+++ b/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Scripts.hs
@@ -20,6 +20,7 @@
 {-# LANGUAGE TypeOperators #-}
 {-# LANGUAGE UndecidableInstances #-}
 {-# LANGUAGE UndecidableSuperClasses #-}
+{-# LANGUAGE ViewPatterns #-}
 {-# OPTIONS_GHC -Wno-orphans #-}
 
 module Cardano.Ledger.Alonzo.Scripts (
@@ -40,6 +41,10 @@ module Cardano.Ledger.Alonzo.Scripts (
   toPlutusSLanguage,
 
   -- ** Plutus Purpose
+  pattern SpendingPurpose,
+  pattern MintingPurpose,
+  pattern CertifyingPurpose,
+  pattern RewardingPurpose,
   AlonzoPlutusPurpose (..),
   AsItem (..),
   AsIndex (..),
@@ -167,7 +172,21 @@ class
     (forall l. PlutusLanguage l => Plutus l -> a) ->
     a
 
-  plutusPurposeSpendingTxIn :: PlutusPurpose AsItem era -> Maybe (TxIn (EraCrypto era))
+  mkSpendingPurpose :: f Word32 (TxIn (EraCrypto era)) -> PlutusPurpose f era
+
+  toSpendingPurpose :: PlutusPurpose f era -> Maybe (f Word32 (TxIn (EraCrypto era)))
+
+  mkMintingPurpose :: f Word32 (PolicyID (EraCrypto era)) -> PlutusPurpose f era
+
+  toMintingPurpose :: PlutusPurpose f era -> Maybe (f Word32 (PolicyID (EraCrypto era)))
+
+  mkCertifyingPurpose :: f Word32 (TxCert era) -> PlutusPurpose f era
+
+  toCertifyingPurpose :: PlutusPurpose f era -> Maybe (f Word32 (TxCert era))
+
+  mkRewardingPurpose :: f Word32 (RewardAcnt (EraCrypto era)) -> PlutusPurpose f era
+
+  toRewardingPurpose :: PlutusPurpose f era -> Maybe (f Word32 (RewardAcnt (EraCrypto era)))
 
   upgradePlutusPurposeAsIndex ::
     AlonzoEraScript (PreviousEra era) =>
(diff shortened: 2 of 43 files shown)


libs/cardano-data/test/Test/Cardano/Data/OSet/StrictSpec.hs
```diff
@@ -71,15 +72,15 @@ spec =
         fromSet set `shouldSatisfy` invariantHolds'
     prop "fromStrictSeq preserves order" $
```
teodanciu: The name still has `fromStrictSeq`, should it say `fromList preserves order` ?

eras/conway/impl/src/Cardano/Ledger/Conway/TxBody.hs
```diff
@@ -329,22 +329,24 @@ basicConwayTxBodyRaw =
...
     SNothing
     mempty
 
-data ConwayTxBodyUpgradeError
+data ConwayTxBodyUpgradeError c
   = CTBUETxCert ConwayTxCertUpgradeError
   | -- | The TxBody contains an update proposal from a pre-Conway era. Since
     --   this can only have come from the genesis delegates, we just discard it.
     CTBUEContainsUpdate
+  | -- | In eras prior to Conway dupicate certificates where allowed
```
teodanciu: Small typo,  `duplicate`

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Scripts.hs
```diff
@@ -215,17 +216,35 @@ isValidPlutusScript pv ps = withPlutusScript ps (isValidPlutus pv)
...
+  deriving stock (Show)
+  deriving newtype (Eq, Ord, NFData, NoThunks, EncCBOR, DecCBOR, Generic)
 
 newtype AsItem ix it = AsItem {unAsItem :: it}
-  deriving newtype (Eq, Ord, Show, NFData, NoThunks, EncCBOR, DecCBOR, Generic)
+  deriving stock (Show)
+  deriving newtype (Eq, Ord, NFData, NoThunks, EncCBOR, DecCBOR, Generic)
 
-instance ToJSON a => ToJSON (AsIndex a b) where
+data AsIxItem ix it = AsIxItem
```
teodanciu, line 256: A little inconsistent with the `AsIndex`, which I presume refers to the same context? 
Is `AsIndexItem` too cumbersome?

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/UTxO.hs
```diff
@@ -208,38 +220,81 @@ getInputDataHashesTxBody (UTxO mp) txBody (ScriptsProvided scriptsProvided) =
...
-      hash <- getScriptHash addr
-      return (AlonzoSpending (AsItem txIn), hash)
+    certifyingScriptsNeeded =
+      AlonzoScriptsNeeded $
+        case foldl' addUniqueTxCertPurpose (Map.empty, 0, []) (txBody ^. certsTxBodyL) of
+          (_, _, certPurposes) -> reverse certPurposes
+      where
+        -- We need to do this funny index manipulation here because we've allowed
+        -- duplicate certificates all the way until Conway. This prevented second
+        -- occurance of a duplicate certificate in the sequence to be used. In order to
```
teodanciu, line 240: small typo, `occurrence`

teodanciu, line 240: small typo: `occurrence` and also in the next line

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Rules/Utxow.hs
```diff
@@ -227,11 +227,9 @@ hasExactSetOfRedeemers ::
   AlonzoScriptsNeeded era ->
   Test (AlonzoUtxowPredFailure era)
 hasExactSetOfRedeemers tx (ScriptsProvided scriptsProvided) (AlonzoScriptsNeeded scriptsNeeded) = do
-  let txBody = tx ^. bodyTxL
-      redeemersNeeded =
-        [ (rp, (sp, sh))
+  let redeemersNeeded =
+        [ (hoistPlutusPurpose toAsIndex sp, (hoistPlutusPurpose toAsItem sp, sh))
```
teodanciu, line 231: Aha so, 
Before:  in order to check that the redeemers we were provided (using  indices) are exactly what we need, we would use AlonzoScriptsNeeded to get the items from ScriptPurpose for each tag an then `indexOf`  on it to find the index  (via the now defunct `rdptr` function).
But now: when constructing AlonzoScriptsNeeded, we pre-emptively index and associate it with the item, so we just need to get them from `getAlonzoScriptsNeeded` and compare them with keys of the redeemers provided in TxWits.
Is this about right?

lehins: That's the perfect explanation of the change of how it worked before and how it'll work after this PR.

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/UTxO.hs
```diff
@@ -208,38 +220,81 @@ getInputDataHashesTxBody (UTxO mp) txBody (ScriptsProvided scriptsProvided) =
...
+        -- preserve this behavior we need to use the index of the first occurance of a
+        -- duplicate certificate.
+        addUniqueTxCertPurpose (!certScriptHashes, !ix, !certPurposes) txCert =
```
teodanciu, line 260: I struggle to get what's happening here. 
I understand that we're trying to keep the behaviour caused by the possibility of duplicate certificates - but what should actually happen if  that's the case? 
"prevented second occurrence of a duplicate certificate to be used" , so this means that it would always return the first one, as per `indexOf` , no? 
So, in  this implementation,  i can tell we're checking for hashes that we have already seen (identifying duplicates), but I don't understand the "ix + 1" part and what actually is the indexing result  in case of duplicates.  The code is simple enough, so it's just me not getting it, sorry!

teodanciu (approved):
Looks good to me :exploding_head:

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/UTxO.hs
lehins: The `ix + 1` part is to count the actual index of each element. It is only when we see a duplicate we use the index of the first occurrence of the element, but that should not affect indices of other elements. For example:

```haskell
-- these are our certificates
cert = [c0, c1, c2, c3, c4, c5]
```

Let's say `c3` is locked by a native script or a key witness, so it does not participate in the script purpose

Also, let's say `c1 == c4`. Here is what we should get for the plutus purpose:
```haskell
-- these are our certificates
plutusPurpose = [(0, c0), (1, c1), (2, c2), (1, c4), (5, c5)]
```

You see, the count must continue no matter what, thus the counter `ix + 1`, but whenever we find a duplicate we use the stored `ix'`.

Does that help?

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Scripts.hs
lehins: It has bothered me too, I was actually gonna rename `AsIndex` to `AsIx`, but that was much work. But, I just don't like `AsIndexItem`, because it just doesn't sound right. It would have to be `AsIndexedItem`, but that just too long, especially since it is always used as part of another type, eg: :grin: 
```haskell
AlonzoPlutusPurpose AsIndexedItem (AlonzoEra StandardCrypto)
```

lehins: What do you think, should we rename `AsIndex` to `AsIx` for consistency?

I could create a tech dept ticket that we could address at some point later.

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/UTxO.hs
teodanciu: Yes!  Thank you for taking the time to explain!

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Scripts.hs
teodanciu: I think `Ix` is more common in our codebase for types than `Index`  (except for MemoHashIndex..), but we have `TxIx` and `Data.Ix`.  
So I think what you propose, a ticket to change to `AsIx`  is better than what I said `AsIndexItem`, which indeed, is long, ungrammatical and inconsistent!  :D

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/UTxO.hs
lehins: Done

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Scripts.hs
lehins: Agree. Created a ticket: https://github.com/IntersectMBO/cardano-ledger/issues/4036

eras/conway/impl/src/Cardano/Ledger/Conway/TxBody.hs
lehins: Fixed

libs/cardano-data/test/Test/Cardano/Data/OSet/StrictSpec.hs
lehins: Fixed.

PR #3997, opened 2024-01-18
Added DRep delegation injections to Conway

### LINKED ISSUE #3995: DRep delegation injection
Turns out that it is necessary to be able to generate and consume injected delegations in a streaming fashion.

In order to accommodate this we need to be able to do delegations of stake pools and DReps at the same time. For this to work we need to add this ability to `ConwayGenesis`. In other words this field need to be changed from `DRep` to `Delegatee`:

https://github.com/IntersectMBO/cardano-ledger/blob/53a64c898408966940c66fa4b29d42f2656240f4/eras/conway/impl/src/Cardano/Ledger/Conway/Genesis.hs#L50

Also injection of StakePool delegations needs to be implemented, similar to how it is done in `Cardano.Ledger.Shelley.Transition`.

Note that this ability of injecting delegations in Conway should not prevent Shelley injection functionality from being used at the same time. In other words, they should be complementary instead of exclusive.

Soupstraw:
# Description

resolves #3995

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/eras/conway/impl/CHANGELOG.md
+++ b/eras/conway/impl/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## 1.12.0.0
 
+* Add `cgDelegsL`
+* Add `FromJSON`, `EncCBOR` and `DecCBOR` instances for `Delegatee`
 * Add `pvtPPSecurityGroup`
 * Add lenses:
   * `pvtCommitteeNormalL`
@@ -21,8 +23,8 @@
 * Add `ToJSON` instance for `ConwayContextError`, `ConwayTxCert`, `ConwayDelegCert`,
   `Delegatee` and `ConwayGovCert`
 * Add `forceDRepPulsingState`
-* Add `registerInitialDReps` and `registerDRepDelegs`
-* Add `cgDRepDelegs`, `cgInitialDReps` to `ConwayGenesis`
+* Add `registerInitialDReps` and `registerDelegs`
+* Add `cgDelegs`, `cgInitialDReps` to `ConwayGenesis`
 * Changed the type of lenses ppCommitteeMaxTermLengthL, ppuCommitteeMaxTermLengthL
 * Change 'getScriptWitnessConwayTxCert' so that DRepRegistration certificate requires a witness
 * Implement `ToJSON` and `FromJSON` instances for `PoolVotingThresholds` and
--- a/eras/conway/impl/src/Cardano/Ledger/Conway/Transition.hs
+++ b/eras/conway/impl/src/Cardano/Ledger/Conway/Transition.hs
@@ -3,6 +3,7 @@
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE FlexibleInstances #-}
 {-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE TypeFamilies #-}
 {-# LANGUAGE UndecidableSuperClasses #-}
 {-# OPTIONS_GHC -Wno-orphans #-}
@@ -11,7 +12,7 @@ module Cardano.Ledger.Conway.Transition (
   ConwayEraTransition (..),
   TransitionConfig (..),
   toConwayTransitionConfigPairs,
-  registerDRepDelegs,
+  registerDelegs,
   registerInitialDReps,
 ) where
 
@@ -22,9 +23,10 @@ import Cardano.Ledger.Conway.Core (Era (..))
 import Cardano.Ledger.Conway.Era
 import Cardano.Ledger.Conway.Genesis (ConwayGenesis (..), toConwayGenesisPairs)
 import Cardano.Ledger.Conway.Translation ()
+import Cardano.Ledger.Conway.TxCert (Delegatee, getStakePoolDelegatee, getVoteDelegatee)
 import Cardano.Ledger.Credential (Credential)
 import Cardano.Ledger.Crypto
-import Cardano.Ledger.DRep (DRep, DRepState)
+import Cardano.Ledger.DRep (DRepState)
 import Cardano.Ledger.Keys (KeyRole (..))
 import Cardano.Ledger.Shelley.LedgerState (
   NewEpochState,
@@ -37,7 +39,8 @@ import Cardano.Ledger.Shelley.LedgerState (
   vsDRepsL,
  )
 import Cardano.Ledger.Shelley.Transition
-import qualified Cardano.Ledger.UMap as UMap
+import Cardano.Ledger.UMap (UMElem (..), umElemsL)
+import Control.Applicative (Alternative (..))
 import Data.Aeson (
   FromJSON (..),
   KeyValue (..),
@@ -51,15 +54,16 @@ import Data.Aeson (
 import Data.ListMap (ListMap)
 import qualified Data.ListMap as ListMap
 import qualified Data.Map.Strict as Map
+import Data.Maybe.Strict (StrictMaybe (..), maybeToStrictMaybe)
 import GHC.Generics
 import Lens.Micro
 import NoThunks.Class (NoThunks (..))
 
 class EraTransition era => ConwayEraTransition era where
-  tcDRepDelegsL ::
+  tcDelegsL ::
     Lens'
       (TransitionConfig era)
-      (ListMap (Credential 'Staking (EraCrypto era)) (DRep (EraCrypto era)))
+      (ListMap (Credential 'Staking (EraCrypto era)) (Delegatee (EraCrypto era)))
 
   tcInitialDRepsL ::
     Lens'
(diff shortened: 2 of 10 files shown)


eras/conway/impl/src/Cardano/Ledger/Conway/Transition.hs
```diff
@@ -124,17 +130,40 @@ registerInitialDReps cfg =
...
-registerDRepDelegs cfg =
-  nesEsL . esLStateL . lsCertStateL . certDStateL . dsUnifiedL %~ drepDelegs
-  where
-    drepDelegs umap =
-      UMap.unUView $
-        Map.foldrWithKey'
-          UMap.insert'
-          (UMap.DRepUView umap)
-          (ListMap.toMap $ cfg ^. tcDRepDelegsL)
+registerDelegs cfg = registerDRepDelegs cfg . registerStakeDelegs cfg
```
lehins: Unfortunately this is not gonna work, because this update has to happen in a single pass, not in two passes like it is implemented in this PR.

I suggest not using the views and instead modifying the Map in `umElems :: !(Map (Credential 'Staking c) (UMElem c))` directly. We don't need to worry about the invariant of keeping the `Ptr` map correct for this, since we are not going to have any pointer resolution in Conway, so it will be safe to manipulate the Map, instead of a the views.

eras/conway/impl/src/Cardano/Ledger/Conway/TxCert.hs
```diff
@@ -358,6 +359,39 @@ instance Crypto c => ToJSON (Delegatee c) where
...
+        !> To kh
+  encCBOR (DelegVote dRep) =
+    encode $
+      Sum DelegVote 1
+        !> To dRep
+  encCBOR (DelegStakeVote kh dRep) =
+    encode $
+      Sum DelegStakeVote 2
+        !> To kh
+        !> To dRep
```
lehins: There is a little too much noise when it is written like this. How about:
```suggestion
  encCBOR = encode . \case 
    DelegStake kh -> Sum DelegStake 0 !> To kh
    DelegVote dRep -> Sum DelegVote 1 !> To dRep
    DelegStakeVote kh dRep -> Sum DelegStakeVote 2 !> To kh !> To dRep
```

lehins (changes requested):
Unfortunately implementation for injection still has the same problem of requiring to load the whole input into memory, because it has to pass over the source ListMap twice.

I made a suggestion that should help you fix that.

eras/conway/impl/src/Cardano/Ledger/Conway/Transition.hs
Soupstraw: I got rid of the functions that let you register DRep and stake pool delegations separately. We haven't released those yet so it should be fine to do so.

lehins: Yeah, that's fine. They'll have to reintegrate those in consensus, but hat should not be a big deal.

eras/conway/impl/src/Cardano/Ledger/Conway/Transition.hs
```diff
@@ -124,17 +128,30 @@ registerInitialDReps cfg =
...
+        (maybeToStrictMaybe $ getStakePoolDelegatee d)
+        (maybeToStrictMaybe $ getVoteDelegatee d)
+    overrideSJust _ (SJust x) = SJust x
+    overrideSJust x _ = x
+    joinUMElems (UMElem _ _ nsp ndrep) (UMElem rdp ptrs osp odrep) =
+      UMElem
+        rdp
+        ptrs
+        (overrideSJust osp nsp)
+        (overrideSJust odrep ndrep)
```
lehins: That's just an alternative. Also don't be too stingy on descriptive names :wink: 
```suggestion
        (newStakePool <|> oldStakePool)
        (newDRep <|> oldDRrep)
```

eras/conway/impl/src/Cardano/Ledger/Conway/Transition.hs
```diff
@@ -124,17 +128,30 @@ registerInitialDReps cfg =
   where
     drepsMap = ListMap.toMap $ cfg ^. tcInitialDRepsL
 
-registerDRepDelegs ::
-  ConwayEraTransition era =>
+registerDelegs ::
+  forall era.
+  ( ConwayEraTransition era
+  , TranslationContext era ~ ConwayGenesis (EraCrypto era)
```
lehins: Why do you need this constraint?
```suggestion
```

eras/conway/impl/src/Cardano/Ledger/Conway/Transition.hs
```diff
@@ -124,17 +128,30 @@ registerInitialDReps cfg =
...
+  nesEsL . esLStateL . lsCertStateL . certDStateL . dsUnifiedL . umElemsL
+    %~ \m -> ListMap.foldrWithKey (\(k, v) -> Map.insertWith joinUMElems k $ delegateeToUMElem v) m delegs
   where
-    drepDelegs umap =
-      UMap.unUView $
-        Map.foldrWithKey'
-          UMap.insert'
-          (UMap.DRepUView umap)
-          (ListMap.toMap $ cfg ^. tcDRepDelegsL)
+    delegs = cfg ^. tcTranslationContextL . cgDelegsL
```
lehins: Aha, this would bypass your `protectMainnetLens` protection that you wrote:
```suggestion
    delegs = cfg ^. tcDelegsL
```

eras/conway/impl/src/Cardano/Ledger/Conway/Transition.hs
```diff
@@ -86,9 +90,9 @@ instance Crypto c => EraTransition (ConwayEra c) where
 instance Crypto c => ConwayEraTransition (ConwayEra c) where
   tcConwayGenesisL = lens ctcConwayGenesis (\g x -> g {ctcConwayGenesis = x})
 
-  tcDRepDelegsL =
+  tcDelegsL =
     protectMainnetLens "DRepDelegs" null $
```
lehins:
```suggestion
    protectMainnetLens "ConwayDelegs" null $
```

libs/cardano-ledger-api/CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## 1.8.0.0
 
+* Add `registerDelegs`
+* Remove `registerDRepDelegs`
```
lehins: Just modify the original entry below:
```diff
-* Add `registerInitialDReps`, `registerDRepDelegs`
+* Add `registerInitialDReps`, `registerDelegs`
```
```suggestion
```

eras/conway/impl/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## 1.12.0.0
 
+* Add `cgDelegsL`
+* Add `registerDelegs`
+* Remove `registerDRepDelegs`
```
lehins: Same here. It is the same version and previous addition hasn't been released yet
```suggestion
```
```diff
-* Add `registerInitialDReps` and `registerDRepDelegs`
+* Add `registerInitialDReps` and `registerDelegs`
```

Also change this line:
```diff
-* Add `cgDRepDelegs`, `cgInitialDReps` to `ConwayGenesis`
+* Add `cgDelegs`, `cgInitialDReps` to `ConwayGenesis`
```

lehins (changes requested):
One more important issue and I think it is good to go.

eras/conway/impl/src/Cardano/Ledger/Conway/Transition.hs
Soupstraw: I thought so as well at first, but it's not. Alternative takes the first `SJust` it sees and sticks with that, while `overrideSJust` takes the second argument when both are `SJust`.

eras/conway/impl/src/Cardano/Ledger/Conway/Transition.hs
Soupstraw: Because of `cgDelegsL`

eras/conway/impl/src/Cardano/Ledger/Conway/Transition.hs
Soupstraw: oh, right I see now that you flipped the arguments.

eras/conway/impl/src/Cardano/Ledger/Conway/Transition.hs
Soupstraw: This is fixed by your other suggestion

eras/conway/impl/src/Cardano/Ledger/Conway/Transition.hs
Soupstraw: Good catch!

lehins (approved):
Beautiful!

PR #3999, opened 2024-01-18
Add PParamUpdates to the plutus context, by transforming them to Plutus Data

### LINKED ISSUE #3994: PParamsUpdate translation to PlutusV3 context
#3987 took care of almost all of the Conway era translation of transaction to PlutusV3 context, with one exception being the PParamsUpdate translation in the ParameterUpdate governance action.

Here is the placeholder for this translation (actual comment in TODO can be ignored, it is only relevant for future eras after Conway):

https://github.com/IntersectMBO/cardano-ledger/blob/50f98b1a53e72103294c86fc4b947e2b48e3678a/eras/conway/impl/src/Cardano/Ledger/Conway/TxInfo.hs#L426-L429

Unlike other PParamsUpdate will be a translation of a ledger type into `BuiltinData`, instead of a specialized type in the plutus context:

* `ChangedParameters` type: https://github.com/IntersectMBO/plutus/blob/9aa445f45d35033a272f37a59e5e2223d3db7ecb/plutus-ledger-api/src/PlutusLedgerApi/V3/Contexts.hs#L252

The correct way of doing this is by creating `ToData` instance for `PParamUpdate (ConwayEra c)` type. We should use lenses, instead of manually looking into the ConwayPParamsHKD type.

* `ToData` type class: https://github.com/IntersectMBO/plutus/blob/9aa445f45d35033a272f37a59e5e2223d3db7ecb/plutus-tx/src/PlutusTx/IsData/Class.hs#L38-L40

* `BuiltinData` definition (which just a wrapper around `Data`): https://github.com/IntersectMBO/plutus/blob/9aa445f45d35033a272f37a59e5e2223d3db7ecb/plutus-tx/src/PlutusTx/Builtins/Internal.hs#L434-L435
```haskell
data BuiltinData = BuiltinData ~PLC.Data
```
* While `Data` is defined here (available through `PlutusLedgerApi.Common`): https://github.com/IntersectMBO/plutus/blob/9aa445f45d35033a272f37a59e5e2223d3db7ecb/plutus-core/plutus-core/src/PlutusCore/Data.hs#L41-L46
```haskell
data Data =
      Constr Integer [Data]
    | Map [(Data, Data)]
    | List [Data]
    | I Integer
    | B BS.ByteString
```

Instead of textual keys for parameter names it has been decided to use same tags that are used in CBOR serialization of parameter updates, this would guarantee that they will be stable. 

Relevant tools for the task: https://github.com/IntersectMBO/plutus/blob/9aa445f45d35033a272f37a59e5e2223d3db7ecb/plutus-ledger-api/src/PlutusLedgerApi/V3.hs#L132-L141

TimSheard:
Adds a new class ToPlutusData 
Uses table driven approach to make instance for PParamsUpdate

resolves #3994 

# Checklist

- [x] Self-reviewed the diff


TimSheard:
Still a work in progress as it doesn't do anything yet with the plutus context

libs/cardano-ledger-core/src/Cardano/Ledger/Plutus/ToPlutusData.hs
```diff
@@ -0,0 +1,259 @@
...
+  toPlutusData n = I n
+  fromPlutusData (I n) = Just n
+  fromPlutusData _ = Nothing
+
+-- =====================================================================================
+-- Tools for building ToPlutusData instances for (PParamUpdates era) for many eras.
+
+-- | Pair the tag, and exisitenially hide the type of the lens for the field with that Lens'
+data PParam era where
+  PParam :: (DecCBOR t, ToPlutusData t) => Integer -> Lens' (PParamsUpdate era) (StrictMaybe t) -> PParam era
```
lehins: Let's change this to `Word`, since that is what used for CBOR tags. Conversion to `Integer` is cheap for the ToPlutusData conversion, but doing other operations on Integer and conversion back to Word, are more expensive, which is unnecessary.

Also we are not doing any arithmetic on those tags, so there are no dangers of overflow
```suggestion
  PParam :: (DecCBOR t, ToPlutusData t) => Word -> Lens' (PParamsUpdate era) (StrictMaybe t) -> PParam era
```

libs/cardano-ledger-core/src/Cardano/Ledger/Plutus/ToPlutusData.hs
```diff
@@ -0,0 +1,259 @@
...
+-- ====================================================================
+-- Generic, Table (Map.Map Integer (PParam era)) driven translators for
+-- PParamUpdate, A table for some era, will make instances for (PParamUpdate era) for that Era.
+
+-- | Translate a PParamUpdate to Data, using the PParamMap.
+pparamUpdateToData :: Map.Map Integer (PParam era) -> PParamsUpdate era -> Data
+pparamUpdateToData pparamMap ppu = Map (concat (Map.mapWithKey toMapPair pparamMap))
+  where
+    toMapPair tag (PParam _ lense) =
+      case ppu ^. lense of
```
lehins: Convention is to use `L` suffix for lenses
```suggestion
    toMapPair tag (PParam _ ppuL) =
      case ppu ^. ppuL of
```

libs/cardano-ledger-core/src/Cardano/Ledger/Plutus/ToPlutusData.hs
```diff
@@ -0,0 +1,259 @@
...
+  where
+    accum m p@(PParam n _) = Map.insert n p m
+
+-- ====================================================================
+-- Generic, Table (Map.Map Integer (PParam era)) driven translators for
+-- PParamUpdate, A table for some era, will make instances for (PParamUpdate era) for that Era.
+
+-- | Translate a PParamUpdate to Data, using the PParamMap.
+pparamUpdateToData :: Map.Map Integer (PParam era) -> PParamsUpdate era -> Data
+pparamUpdateToData pparamMap ppu = Map (concat (Map.mapWithKey toMapPair pparamMap))
```
lehins: This operation is better done with a fold, which will also guarantee the correct order. (`toMapPair` will need to be adjusted to handle the acc)
```suggestion
pparamUpdateToData pparamMap ppu = Map (Map.foldrWithKey' toMapPair [] pparamMap)
```

libs/cardano-ledger-core/src/Cardano/Ledger/Plutus/ToPlutusData.hs
```diff
@@ -0,0 +1,259 @@
...
+instance ToPlutusData Integer where
+  toPlutusData n = I n
+  fromPlutusData (I n) = Just n
+  fromPlutusData _ = Nothing
+
+-- =====================================================================================
+-- Tools for building ToPlutusData instances for (PParamUpdates era) for many eras.
+
+-- | Pair the tag, and exisitenially hide the type of the lens for the field with that Lens'
+data PParam era where
```
lehins: This type needs to go into `Cardano.Ledger.Core.PParams`, since it is no longer related to just the PlutusData. You'll need to invert the dependency of modules, which will work when my other suggestions are applied

libs/cardano-ledger-core/src/Cardano/Ledger/Plutus/ToPlutusData.hs
```diff
@@ -0,0 +1,259 @@
...
+      where
+        accum ppu n (PParam _ lense) =
+          if tag == n
+            then case fromPlutusData dat of
+              Nothing -> ppu
+              Just t -> ppu & lense .~ t
+            else ppu
+pparamUpdateFromData _ _ = Nothing
+
+-- ========================================================
```
lehins: This experiment with using this for pre-conway pparams for decoding shows that we can use it for things other than ToPlutusData. For this reason everything that is below should travel to `Cardano.Ledger.Shelley.PParams`

Although, please don't expand on it any further in this PR. We have a very high priority on the original issue of translating PParams to the Plutus context.

eras/conway/impl/src/Cardano/Ledger/Conway/PParamsUpdate.hs
```diff
@@ -0,0 +1,127 @@
+{-# LANGUAGE FlexibleInstances #-}
+{-# LANGUAGE TypeApplications #-}
+{-# OPTIONS_GHC -Wno-orphans #-}
+
+module Cardano.Ledger.Conway.PParamsUpdate where
```
lehins: This is plutus specific and is not related in any way to updating plutus parameters. Let's create a module with this name instead. (And yes, please create a new subfolder)
```suggestion
module Cardano.Ledger.Conway.Plutus.ToPlutusData where
```

eras/conway/impl/src/Cardano/Ledger/Conway/PParamsUpdate.hs
```diff
@@ -0,0 +1,127 @@
...
+      <$> fromPlutusData motion
+      <*> fromPlutusData comm
+      <*> fromPlutusData noconf
+      <*> fromPlutusData update
+      <*> fromPlutusData hard
+      <*> fromPlutusData network
+      <*> fromPlutusData econ
+      <*> fromPlutusData tech
+      <*> fromPlutusData gov
+      <*> fromPlutusData treasury
```
lehins: Stick with non semantic names as you've done for `PoolVotingThresholds`, since the order is very important and it is easier to see the order is correct with something like letters of alphabet or with numerics (eg. `x00`, `x01`, ..) 
```suggestion
  fromPlutusData (List [a, b, c, d, e, f, g, h, i, j]) =
    DRepVotingThresholds
      <$> fromPlutusData a
      <*> fromPlutusData b
      <*> fromPlutusData c
      <*> fromPlutusData d
      <*> fromPlutusData e
      <*> fromPlutusData f
      <*> fromPlutusData g
      <*> fromPlutusData h
      <*> fromPlutusData i
      <*> fromPlutusData j
```

eras/conway/impl/src/Cardano/Ledger/Conway/PParamsUpdate.hs
```diff
@@ -0,0 +1,127 @@
...
+
+-- | Instance for ConwayEra, by changing 'conwayPParamMap' we can get instances for other Eras.
+instance Crypto c => ToPlutusData (PParamsUpdate (ConwayEra c)) where
+  toPlutusData = pparamUpdateToData conwayPParamMap
+  fromPlutusData = pparamUpdateFromData conwayPParamMap
+
+-- | Instance for ShelleyEra
+instance Crypto c => ToPlutusData (PParamsUpdate (ShelleyEra c)) where
+  toPlutusData = pparamUpdateToData shelleyPParamMap
+  fromPlutusData = pparamUpdateFromData shelleyPParamMap
```
lehins: This must be removed. We do not want provide functionality for passing shelley pparams to plutus
```suggestion
```

libs/cardano-ledger-core/src/Cardano/Ledger/Plutus/ToPlutusData.hs
```diff
@@ -0,0 +1,259 @@
...
+-- | A Map for the Shelley, Allegra, and Mary eras
+shelleyPParamMap ::
+  (ProtVerAtMost era 8, ProtVerAtMost era 6, ProtVerAtMost era 4, EraPParams era) =>
+  Map.Map Integer (PParam era)
+shelleyPParamMap = makePParamMap shelleyPParam
+
+-- | Decode any era, given the right (Map.Map Integer (PParam era))
+decodePParamsUpdate :: EraPParams era => Map.Map Integer (PParam era) -> Decoder s (PParamsUpdate era)
+decodePParamsUpdate ppMap = do
+  mapParts <- decodeMapContents $ do
```
lehins: This is incorrect, because `decodeMapContents` does not guard against duplicates.

Can you please remove anything related to decoding form this PR. That is not the goal of this work and when we do get to using this approach for decoding, I would like to have it all contained in a separate PR.

libs/cardano-ledger-core/src/Cardano/Ledger/Plutus/ToPlutusData.hs
```diff
@@ -0,0 +1,259 @@
...
+  fromPlutusData (Constr 1 [x]) = SJust <$> fromPlutusData x
+  fromPlutusData _ = Nothing
+
+instance ToPlutusData a => ToPlutusData [a] where
+  toPlutusData xs = List (map toPlutusData xs)
+  fromPlutusData (List xs)
+    | all isJust mlist = Just $ catMaybes mlist
+    | otherwise = Nothing
+    where
+      mlist = map fromPlutusData xs
```
lehins:
```suggestion
  fromPlutusData (List xs) = mapM fromPlutusData xs
```

libs/cardano-ledger-core/src/Cardano/Ledger/Plutus/ToPlutusData.hs
```diff
@@ -0,0 +1,259 @@
...
+    | all isJust ps = Just $ Map.fromList (catMaybes ps)
+    | otherwise = Nothing
+    where
+      ps = map unpair pairs
+      unpair (d1, d2) =
+        do
+          x <- fromPlutusData d1
+          y <- fromPlutusData d2
+          Just (x, y)
+  fromPlutusData _ = Nothing
```
lehins:
```suggestion
  fromPlutusData (Map pairs) =
    Map.fromList <$> 
    mapM (\(k, v) -> (,) <$> fromPlutusData k <*> fromPlutusData v)) pairs
```

libs/cardano-ledger-core/src/Cardano/Ledger/Plutus/ToPlutusData.hs
```diff
@@ -0,0 +1,259 @@
...
+  fromPlutusData (List [major, minor]) = ProtVer <$> fromPlutusData major <*> fromPlutusData minor
+  fromPlutusData _ = Nothing
+
+instance ToPlutusData Nonce where
+  toPlutusData NeutralNonce = Constr 0 []
+  toPlutusData (Nonce hash) = Constr 1 [B (hashToBytes hash)]
+  fromPlutusData (Constr 0 []) = Just NeutralNonce
+  fromPlutusData (Constr 1 [B bytes]) = Nonce <$> (hashFromBytes bytes)
+  fromPlutusData _ = Nothing
+
```
lehins: We are never translating this to plutus. So, this instance should be removed
```suggestion
```

libs/cardano-ledger-core/src/Cardano/Ledger/Plutus/ToPlutusData.hs
```diff
@@ -0,0 +1,259 @@
...
+instance ToPlutusData Prices where
+  toPlutusData p = List [toPlutusData (prMem p), toPlutusData (prSteps p)]
+  fromPlutusData (List [x, y]) = Prices <$> fromPlutusData x <*> fromPlutusData y
+  fromPlutusData _ = Nothing
+
+instance ToPlutusData Coin where
+  toPlutusData (Coin n) = I n
+  fromPlutusData (I n) | n > 0 = Just (Coin n) -- DO WE NEED TO TEST n > 0 ?
```
lehins: We don't. Unfortunately there is no invariant on the Coin type that it can never be negative. In fact there are a couple of places in the code where it can be negative, albeit for invalid transactions.
```suggestion
  fromPlutusData (I n) = Just (Coin n)
```

libs/cardano-ledger-core/src/Cardano/Ledger/Plutus/ToPlutusData.hs
```diff
@@ -0,0 +1,259 @@
...
+  fromPlutusData _ = Nothing
+
+instance ToPlutusData Word32 where
+  toPlutusData w32 = I (toInteger @Word32 w32)
+  fromPlutusData (I n) | n >= 0 = Just $ fromInteger @Word32 n
```
lehins: Same should apply to other `Word` like conversions
```suggestion
  fromPlutusData (I n) | n >= 0 && n <= toInteger (maxBound @Word32) =
    Just $ fromInteger @Word32 n
```

libs/cardano-ledger-core/src/Cardano/Ledger/Plutus/ToPlutusData.hs
```diff
@@ -0,0 +1,259 @@
...
+  toPlutusData :: x -> Data
+  fromPlutusData :: Data -> Maybe x
+  fromPlutusData _ = Nothing
+
+instance ToPlutusData a => ToPlutusData (StrictMaybe a) where
+  toPlutusData SNothing = Constr 0 []
+  toPlutusData (SJust x) = Constr 1 [toPlutusData x]
+  fromPlutusData (Constr 0 []) = Just SNothing
+  fromPlutusData (Constr 1 [x]) = SJust <$> fromPlutusData x
+  fromPlutusData _ = Nothing
```
lehins: We don't need to translate that either. In the matter of fact, because this instance is available the implementation of `pparamUpdateFromData` is incorrect, because it assumes pparam values are encoded as StrictMaybe
```suggestion
```

libs/cardano-ledger-core/src/Cardano/Ledger/Plutus/ToPlutusData.hs
```diff
@@ -0,0 +1,259 @@
...
+    updatePPU :: Integer -> Data -> Map.Map Integer (PParam era) -> PParamsUpdate era -> PParamsUpdate era
+    updatePPU tag dat mp ppu0 = Map.foldlWithKey' accum ppu0 mp
+      where
+        accum ppu n (PParam _ lense) =
+          if tag == n
+            then case fromPlutusData dat of
+              Nothing -> ppu
+              Just t -> ppu & lense .~ t
+            else ppu
+pparamUpdateFromData _ _ = Nothing
```
lehins: That's just a monadic fold, also we need to guard against duplicates and make sure we are not expecting StrictMaybe for values
```suggestion
pparamUpdateFromData pparamMap (Map pairs) = 
  foldlM accum (emptyPParamsUpdate, pparamMap) pairs
  where
    accum (!ppu, !leftOverMap) (key, value) = do
      tag <- fromPlutusData key
      val <- fromPlutusData value
      PParam _ ppuL <- Map.lookup tag leftOverMap
      pure (ppu $ ppuL .~ SJust val, Map.delete tag leftOverMap)
```

libs/cardano-ledger-core/src/Cardano/Ledger/Plutus/ToPlutusData.hs
```diff
@@ -0,0 +1,259 @@
...
+
+-- | Pair the tag, and exisitenially hide the type of the lens for the field with that Lens'
+data PParam era where
+  PParam :: (DecCBOR t, ToPlutusData t) => Integer -> Lens' (PParamsUpdate era) (StrictMaybe t) -> PParam era
+
+-- | Turn a list into a Map, this assures we have no duplicates.
+makePParamMap :: [PParam era] -> Map.Map Integer (PParam era)
+makePParamMap xs = foldl' accum Map.empty xs
+  where
+    accum m p@(PParam n _) = Map.insert n p m
```
lehins: This will be quite a bit faster
```suggestion
makePParamMap xs = Map.fromList [(n, p) | p@(PParam n _) <- xs]
```

eras/conway/impl/src/Cardano/Ledger/Conway/PParamsUpdate.hs
```diff
@@ -0,0 +1,127 @@
...
+
+-- | A Decoder for the Conway era
+decodeConway :: ConwayEraPParams era => Decoder s (PParamsUpdate era)
+decodeConway = decodePParamsUpdate conwayPParamMap
```
lehins: Please remove that as well, we'll do this at a later time.
```suggestion
```

lehins (changes requested):
There are some issues in the PR, but the overall direction is great!

The only thing that is missing now is the actual translation to PlutusV3 context and roundtrip tests that should have the reusable functionality placed in `cardano-ledger-core:testlib:Test.Cardano.Ledger.Plutus.ToPlutusData`
and the actual Conway tests should go into `cardano-ledger-conway:tests:Test.Cardano.Ledger.Conway.Plutus.ToPlutusData`

libs/cardano-ledger-core/src/Cardano/Ledger/Plutus/ToPlutusData.hs
lehins: I don't have a type checker in github :grin: 

But is better if I make a half working suggestion than try to explain in words what I mean.

TimSheard: Took a while to debug this.
The biggest problem was the order of the pattern matching in the do

libs/cardano-ledger-core/src/Cardano/Ledger/Plutus/ToPlutusData.hs
TimSheard: I am not sure I understand this.

libs/cardano-ledger-core/src/Cardano/Ledger/Plutus/ToPlutusData.hs
TimSheard: done

libs/cardano-ledger-core/src/Cardano/Ledger/Plutus/ToPlutusData.hs
TimSheard: So I removed all this.

libs/cardano-ledger-core/src/Cardano/Ledger/Plutus/ToPlutusData.hs
TimSheard: done

libs/cardano-ledger-core/src/Cardano/Ledger/Plutus/ToPlutusData.hs
lehins: You already fixed the problem that I was talking about. So you are no longer even using this instance, so it can be safely removed.

eras/conway/impl/src/Cardano/Ledger/Conway/PParams.hs
```diff
@@ -1295,3 +1301,105 @@ conwayModifiedPPGroups
...
+      <*> fromPlutusData f
+      <*> fromPlutusData g
+      <*> fromPlutusData h
+      <*> fromPlutusData i
+      <*> fromPlutusData j
+  fromPlutusData _ = Nothing
+
+instance ToPlutusData CoinPerByte where
+  toPlutusData (CoinPerByte c) = toPlutusData @Coin c
+  fromPlutusData x = CoinPerByte <$> fromPlutusData @Coin x
```
lehins: This can be derived using GeneralizedNewtypeDeriving

eras/conway/impl/src/Cardano/Ledger/Conway/PParams.hs
```diff
@@ -1295,3 +1301,105 @@ conwayModifiedPPGroups
...
+
+instance Crypto c => ToPlutusData (PParamsUpdate (ConwayEra c)) where
+  toPlutusData = pparamUpdateToData conwayPParamMap
+  fromPlutusData = pparamUpdateFromData conwayPParamMap
+
+class ConwayEraPlutusTxInfo era where
+  ppuToData :: PParamsUpdate era -> Data
```
lehins: This class should go into a new module `Cardano.Ledger.Conway.Plutus.Context`

And the name of the function should be `toPlutusChangedParameters`

It also needs to be parameterized on language

Also note the change in the return type.

```suggestion
class EraPlutusTxInfo l era => ConwayEraPlutusTxInfo (l :: Language) era where
  toPlutusChangedParameters :: proxy l -> PParamsUpdate era -> PV3.ChangedParameters
```

libs/cardano-ledger-core/testlib/Test/Cardano/Ledger/Plutus/ToPlutusData.hs
```diff
@@ -0,0 +1,72 @@
...
+              ("\nIntermediate Data\n" ++ show (toPlutusData y) ++ "\n")
+              (Just (y :: x) === fromPlutusData (toPlutusData y))
+        )
+    )
+
+plutusDataRoundtripSpec :: Spec
+plutusDataRoundtripSpec =
+  do
+    describe "roundtrip ToPlutusData" $
+      do
```
lehins: Slightly unusually formatting. Also this is not a reusable test case, so it should just go into the test suite in to `Test.Cardano.Ledger.PlutusSpec` module, instead of be exported from `testlib`
```suggestion
plutusDataRoundtripSpec = do
  describe "roundtrip ToPlutusData" $ do
```

libs/cardano-ledger-core/testlib/Test/Cardano/Ledger/Plutus/ToPlutusData.hs
```diff
@@ -0,0 +1,72 @@
...
+roundtrip =
+  prop
+    (show (typeRep @x))
+    ( property
+        ( \y ->
+            counterexample
+              ("\nIntermediate Data\n" ++ show (toPlutusData y) ++ "\n")
+              (Just (y :: x) === fromPlutusData (toPlutusData y))
+        )
+    )
```
lehins: hspec will already print the generated value during test failure, so there is no need to print it again. Also using `shouldBe` will give all of the information necessary during the test failure (it'll even diff it). That is why `hspec` is great: 
```suggestion
  prop (show (typeRep @x)) $ \ y ->
    fromPlutusData (toPlutusData y) `shouldBe` Just (y :: x)
```

libs/cardano-ledger-core/testlib/Test/Cardano/Ledger/Plutus/ToPlutusData.hs
```diff
@@ -0,0 +1,72 @@
+{-# LANGUAGE AllowAmbiguousTypes #-}
+{-# LANGUAGE DataKinds #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeApplications #-}
+
+module Test.Cardano.Ledger.Plutus.ToPlutusData (
+  roundtrip,
```
lehins: We have bunch of round tripping properties in this package. It'll be good to distinguish with a suffix. Also adding suffix `Spec` will let the user know what it is. It could have been either a `Property` or `Expectation`, but it is a `Spec`, which is totally fine, just a nice convention to distinguish between the three concepts.
```suggestion
  roundTripPlutusDataSpec,
```

libs/cardano-ledger-core/src/Cardano/Ledger/Plutus/ToPlutusData.hs
```diff
@@ -0,0 +1,139 @@
...
+
+instance ToPlutusData Word8 where
+  toPlutusData w8 = I (toInteger @Word8 w8)
+  fromPlutusData (I n) | n >= 0 && n <= toInteger (maxBound @Word8) = Just $ fromInteger @Word8 n
+  fromPlutusData _ = Nothing
+
+instance ToPlutusData EpochInterval where
+  toPlutusData (EpochInterval w32) = I (toInteger @Word32 w32)
+  fromPlutusData (I n) | n >= 0 = Just . EpochInterval $ fromInteger @Word32 n
+  fromPlutusData _ = Nothing
```
lehins: It is better to derive this. This way we guarantee that the upper bound check is also done:
```haskell
    | n >= 0 && n <= toInteger (maxBound @Word32) =
```
```suggestion
deriving instance ToPlutusData EpochInterval
```

libs/cardano-ledger-core/src/Cardano/Ledger/Plutus/ToPlutusData.hs
```diff
@@ -0,0 +1,139 @@
...
+
+instance ToPlutusData Prices where
+  toPlutusData p = List [toPlutusData (prMem p), toPlutusData (prSteps p)]
+  fromPlutusData (List [x, y]) = Prices <$> fromPlutusData x <*> fromPlutusData y
+  fromPlutusData _ = Nothing
+
+instance ToPlutusData Coin where
+  toPlutusData (Coin n) = I n
+  fromPlutusData (I n) = Just (Coin n)
+  fromPlutusData _ = Nothing
```
lehins:
```suggestion
deriving instance ToPlutusData Coin
```

libs/cardano-ledger-core/src/Cardano/Ledger/Plutus/ToPlutusData.hs
lehins: This instance is not needed and should be removed

libs/cardano-ledger-core/testlib/Test/Cardano/Ledger/Plutus/ToPlutusData.hs
```diff
@@ -0,0 +1,72 @@
...
+import Test.Hspec
+import Test.Hspec.QuickCheck
+import Test.QuickCheck
+import Type.Reflection (Typeable, typeRep)
+
+-- ==================================================================================
+
+-- | Test the roundtrip property via 'fromPlutusData' and 'toPlutusData' for any type
+--   meeting all the constraints below.
+roundtrip :: forall x. (Typeable x, Show x, Eq x, Arbitrary x, ToPlutusData x) => Spec
```
lehins: It is important to add `HasCallStack` to this kind of helpers:
```suggestion
roundtrip :: forall x. (HasCallStack, Typeable x, Show x, Eq x, Arbitrary x, ToPlutusData x) => Spec
```

eras/conway/impl/src/Cardano/Ledger/Conway/PParams.hs
```diff
@@ -113,6 +114,12 @@ import GHC.Generics (Generic)
 import Lens.Micro
 import NoThunks.Class (NoThunks (..))
 import Numeric.Natural (Natural)
+import PlutusLedgerApi.Common (Data (..))
+
+-- import Cardano.Ledger.Conway.Plutus.ToPlutusData(conwayPParamMap)
```
lehins: dead code
```suggestion
```

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Plutus/ToPlutusData.hs
```diff
@@ -0,0 +1,26 @@
...
+import Cardano.Ledger.Babbage.PParams (CoinPerByte)
+import Cardano.Ledger.Conway (ConwayEra)
+import Cardano.Ledger.Conway.PParams (DRepVotingThresholds, PoolVotingThresholds)
+import Cardano.Ledger.Core (PParamsUpdate)
+import Cardano.Ledger.Crypto (StandardCrypto)
+import Test.Cardano.Ledger.Conway.Arbitrary ()
+import Test.Cardano.Ledger.Core.Arbitrary ()
+import Test.Cardano.Ledger.Plutus.ToPlutusData (roundtrip)
+
+import Test.Hspec
```
lehins: Common re-exports all of the stuff necessary for testing.
```suggestion
import Test.Cardano.Ledger.Common
```

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Plutus/ToPlutusData.hs
```diff
@@ -0,0 +1,26 @@
+{-# LANGUAGE TypeApplications #-}
+
+module Test.Cardano.Ledger.Conway.Plutus.ToPlutusData where
```
lehins: There are no reusable test here either, so they should got into the test suite under this module name, rather than in the `testlib`.

For this also the convention with hspec tests is to export only a single function `spec` that has all of the tests in it. 
```suggestion
module Test.Cardano.Ledger.Conway.PlutusSpec (spec) where
```

eras/conway/impl/test/Main.hs
```diff
@@ -34,3 +35,4 @@ main =
         CostModelsSpec.spec @Conway
       describe "TxWits" $ do
         TxWitsSpec.spec @Conway
+      conwayPlutusDataRoundtripSpec
```
lehins: So, this will look like so:
```suggestion
      PlutusSpec.spec
```

eras/conway/impl/test/Main.hs
```diff
@@ -13,6 +13,7 @@ import qualified Test.Cardano.Ledger.Conway.DRepRatifySpec as DRepRatify
 import qualified Test.Cardano.Ledger.Conway.GenesisSpec as Genesis
 import qualified Test.Cardano.Ledger.Conway.GovActionReorderSpec as GovActionReorder
 import qualified Test.Cardano.Ledger.Conway.Imp as ConwayImp
+import Test.Cardano.Ledger.Conway.Plutus.ToPlutusData (conwayPlutusDataRoundtripSpec)
```
lehins: See my comments below about this.
```suggestion
import Test.Cardano.Ledger.Conway.PlutusSpec as PlutsuSpec
```

eras/conway/impl/src/Cardano/Ledger/Conway.hs
```diff
@@ -26,7 +26,8 @@ import Cardano.Ledger.Conway.Rules ()
 import Cardano.Ledger.Conway.Transition ()
 import Cardano.Ledger.Conway.Translation ()
 import Cardano.Ledger.Conway.Tx ()
-import Cardano.Ledger.Conway.TxInfo ()
+
+-- import Cardano.Ledger.Conway.TxInfo ()
```
lehins: This import is necessary. It brings orphan instances into scope toegther with the ConwayEra type in order to un-orphan them.
```suggestion
import Cardano.Ledger.Conway.TxInfo ()
```

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Plutus/ToPlutusData.hs
```diff
@@ -0,0 +1,26 @@
...
+
+import Test.Hspec
+
+-- ================================
+
+conwayPlutusDataRoundtripSpec :: Spec
+conwayPlutusDataRoundtripSpec =
+  do
+    describe "roundtrip ToPlutusData Conway instances" $
+      do
```
lehins: Inconsistent placement of `do` and rename to `spec`
```suggestion
spec :: Spec
spec = do
  describe "roundtrip ToPlutusData Conway instances" $ do
```

libs/cardano-ledger-core/src/Cardano/Ledger/Core/PParams.hs
```diff
@@ -546,3 +559,46 @@ downgradePParamsUpdate ::
...
+pparamUpdateFromData pparamMap (Map pairs) =
+  fst
+    <$> foldlM accum (emptyPParamsUpdate, pparamMap) pairs
+  where
+    accum (!ppu, !leftOverMap) (key, value) = do
+      tag <- fromPlutusData key
+      PParam _ ppuL <- Map.lookup tag leftOverMap
+      val <- fromPlutusData value
+      pure (ppu & ppuL .~ SJust val, Map.delete tag leftOverMap)
+pparamUpdateFromData _ _ = Nothing
```
lehins: It would be better to put these two functions into a new module `Cardano.Ledger.Conway.Plutus.Context` together with `ConwayEraPlutusTxInfo`, because those functions will be used for creating instances for that class. This way it will be easier for people to find.

This functionality is very much plutus specific and can't be used prior to Conway, so we don't need to to keep it in the `Cardano.Ledger.Core.PParams`

The `PParam` GADT, on the other hand, we'll be able to use other places later on, so this is a perfect home for it.

lehins (commented):
This looks great. Semantically that is exactly what we need.
Just a few comments on shuffling things around for consistency with the rest of the codebase

eras/conway/impl/src/Cardano/Ledger/Conway/PParams.hs
TimSheard: done

libs/cardano-ledger-core/src/Cardano/Ledger/Core/PParams.hs
TimSheard: done

lehins (approved):
@TimSheard There was a few things still needed to be rearranged in this PR, but because we will try to make a release today and this functionality is needed for the next release I took upon myself to finish this up. Check out the last few commits to see what I've done.

PR #4002, opened 2024-01-23
Rename `Acnt` -> `Account`

### LINKED ISSUE #3928: Rename `Acnt` -> `Account`
This applies to all fields and types that use that name, eg. 
* `ppRewardAcnt` -> `ppRewardAccount`, 
* `prAcnt` -> `prAccountState`
* `RewardAcnt` -> `RewardAccount`
* ...

Important part of this rename is that it should be done with deprecation helpers, eg:

```haskell
type RewardAcnt = RewardAccount
pattern RewardAcnt = RewardAccount
```

in order to make the transition smoother.

Lucsanszky:
# Description

This PR renames `Acnt` to `Account` by deprecating the following:
* `prAcnt`
* `ppRewardAcnt`
* `fromCborRewardAcnt`
* `decodeRewardAcnt`
* `putRewardAcnt`
* `deserialiseRewardAcnt`
* `serialiseRewardAcnt`
* `RewardAcnt`
* `deserialiseRewardAcntOld`
* `delPlAcnt`
* `mkVKeyRwdAcnt`

Furthermore, the PR also renames some occurrences of `acnt` for consistency, without any deprecation helpers (as these are just internal functions or helpers, etc.).

Resolves #3928

# Checklist

- [x] Self-reviewed the diff


libs/cardano-ledger-core/src/Cardano/Ledger/Address.hs
```diff
@@ -193,12 +211,21 @@ addrPtrNormalize = \case
   addr -> addr
 
 -- | An account based address for rewards
-data RewardAcnt c = RewardAcnt
+data RewardAccount c = RewardAccount
   { getRwdNetwork :: !Network
   , getRwdCred :: !(Credential 'Staking c)
```
lehins: While you are at it, let's rename these two as well. The convention adopted couple years ago (after this type creation) is to prefix fields with abbreviation of the type, thus the `ra` prefix:
```suggestion
  { raNetwork :: !Network
  , raCredential :: !(Credential 'Staking c)
```

lehins (commented):
This looks great!

We just need changelog entries for the renames and I also suggested two extra renames that can be made in the vicinity of a reward account

Lucsanszky:
Sorry for the rebase after (pre-)review but I wanted to keep the commits organised so we can easily revert if need be. Also, I found some more instances of `a/Acnt`. I will rename those as well.

Lucsanszky:
This is now ready for review. I tried to keep the commits as organised as possible. I also attempted to keep changes separate, except for the final commit (f51ed297d66828fec2445f44c6ff45e60e7cbd6b): that was used to rename some internal stuff for consistency and since those are internal, not exported things, I didn't add deprecation helpers for them. I am also quite certain that I missed a few things here and there.

libs/cardano-ledger-core/testlib/Test/Cardano/Ledger/Core/KeyPair.hs
```diff
@@ -142,12 +145,20 @@ makeWitnessesFromScriptKeys txbodyHash hashKeyMap scriptHashes =
   let witKeys = Map.restrictKeys hashKeyMap scriptHashes
    in mkWitnessesVKey txbodyHash (Map.elems witKeys)
 
+mkVKeyRwdAccount ::
+  Crypto c =>
+  Network ->
+  KeyPair 'Staking c ->
+  RewardAccount c
+mkVKeyRwdAccount network keys = RewardAccount network $ KeyHashObj (hashKey $ vKey keys)
```
lehins: Let's avoid shortening in new name.
```suggestion
mkVKeyRewardAccount ::
  Crypto c =>
  Network ->
  KeyPair 'Staking c ->
  RewardAccount c
mkVKeyRewardAccount network keys = RewardAccount network $ KeyHashObj (hashKey $ vKey keys)
```

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Deleg.hs
```diff
@@ -79,14 +82,18 @@ import NoThunks.Class (NoThunks (..))
 data DelegEnv era = DelegEnv
   { slotNo :: SlotNo
   , ptr_ :: Ptr
-  , acnt_ :: AccountState
+  , account_ :: AccountState
```
lehins: Can you revert this rename?

We should address this type separately, cause all fields in this type are named terribly and `account_` is not much better of a name :cry:

libs/cardano-ledger-core/src/Cardano/Ledger/Address.hs
```diff
@@ -193,20 +211,29 @@ addrPtrNormalize = \case
...
-  { getRwdNetwork :: !Network
-  , getRwdCred :: !(Credential 'Staking c)
+data RewardAccount c = RewardAccount
+  { raNetwork :: !Network
+  , raCredential :: !(Credential 'Staking c)
   }
   deriving (Show, Eq, Generic, Ord, NFData, ToJSONKey, FromJSONKey)
 
+pattern RewardAcnt :: Network -> Credential 'Staking c -> RewardAccount c
+pattern RewardAcnt a b = RewardAccount a b
```
lehins: I believe you can do something like this:
```suggestion
pattern RewardAcnt {getRwdNetwork,getRwdCred} = RewardAccount getRwdNetwork getRwdCred

{-# DEPRECATE getRwdNetwork "In favor of `raNetwork`" #-}
{-# DEPRECATE getRwdCred "In favor of `raCredential`" #-}
```

This will allow even for record updates to still work, eg. `ra { getRwdNetwork = foo }`

libs/cardano-ledger-core/testlib/Test/Cardano/Ledger/Core/KeyPair.hs
```diff
@@ -142,12 +145,20 @@ makeWitnessesFromScriptKeys txbodyHash hashKeyMap scriptHashes =
...
+
 mkVKeyRwdAcnt ::
   Crypto c =>
   Network ->
   KeyPair 'Staking c ->
-  RewardAcnt c
-mkVKeyRwdAcnt network keys = RewardAcnt network $ KeyHashObj (hashKey $ vKey keys)
+  RewardAccount c
+mkVKeyRwdAcnt = mkVKeyRwdAccount
+{-# DEPRECATED mkVKeyRwdAcnt "Use `mkVKeyRwdAccount` instead" #-}
```
lehins:
```suggestion
mkVKeyRwdAcnt = mkVKeyRewardAccount
{-# DEPRECATED mkVKeyRwdAcnt "Use `mkVKeyRewardAccount` instead" #-}
```

libs/cardano-ledger-core/CHANGELOG.md
```diff
@@ -67,6 +75,8 @@
 
 ### `testlib`
 
+* Deprecate `deserialiseRewardAcntOld` in favor of `deserialiseRewardAccountOld`
+* Deprecate `mkVKeyRwdAcnt` in favor of `mkVKeyRwdAccount`
```
```suggestion
* Deprecate `mkVKeyRwdAcnt` in favor of `mkVKeyRewardAccount`
```

libs/ledger-state/src/Cardano/Ledger/State/UTxO.hs
```diff
@@ -236,10 +236,14 @@ countSnapShotStat SnapShot {..} =
...
   { ppsPoolId :: !(Stat (KeyHash 'StakePool C))
-  , ppsRewardAcnt :: !(Stat (Credential 'Staking C))
+  , ppsRewardAccount :: !(Stat (Credential 'Staking C))
   , ppsOwners :: !(Stat (KeyHash 'Staking C))
   }
 
+ppsRewardAcnt :: PoolParamsStats -> Stat (Credential 'Staking C)
+ppsRewardAcnt = ppsRewardAccount
+{-# DEPRECATED ppsRewardAcnt "Use `ppsRewardAccount` instead" #-}
+
```
lehins: This deprecation  isn't needed, we don't release this package. It's more of a benchmarking tool that we used in ledger internally
```suggestion
```

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Generic/PrettyCore.hs
```diff
@@ -2526,10 +2526,14 @@ instance PrettyA (VotingProcedure era) where
...
 
-pcRewardAcnt :: RewardAcnt c -> PDoc
-pcRewardAcnt (RewardAcnt net cred) = ppSexp "RewAccnt" [pcNetwork net, pcCredential cred]
+pcRewardAccount :: RewardAccount c -> PDoc
+pcRewardAccount (RewardAccount net cred) = ppSexp "RewAccnt" [pcNetwork net, pcCredential cred]
 
-instance PrettyA (RewardAcnt c) where prettyA = pcRewardAcnt
+pcRewardAcnt :: RewardAccount c -> PDoc
+pcRewardAcnt = pcRewardAccount
+{-# DEPRECATED pcRewardAcnt "Use `pcRewardAccount` instead" #-}
```
lehins: cardano-ledegr-test isn't one of the packages that we release, so no need to go through deprecation cycle
```suggestion
```

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/Vars.hs
```diff
@@ -2124,8 +2124,12 @@ gaPParamsUpdate = Var (V "gsPParamsUpdate" (PParamsUpdateR reify) No)
...
 gaProtVer = Var (V "gaProtVer" (ProtVerR reify) No)
 
-gaRewardAcnt :: Era era => Term era (Map (RewardAcnt (EraCrypto era)) Coin)
-gaRewardAcnt = Var (V "gaRewardAcnt" (MapR RewardAcntR CoinR) No)
+gaRewardAccount :: Era era => Term era (Map (RewardAccount (EraCrypto era)) Coin)
+gaRewardAccount = Var (V "gaRewardAccount" (MapR RewardAccountR CoinR) No)
+
+gaRewardAcnt :: Era era => Term era (Map (RewardAccount (EraCrypto era)) Coin)
+gaRewardAcnt = gaRewardAccount
+{-# DEPRECATED gaRewardAcnt "Use `gaRewardAccount` instead" #-}
```
lehins: Ditto (See my comment below)
```suggestion
```

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/TypeRep.hs
```diff
@@ -351,6 +352,10 @@ data Rep era t where
   DelegateeR :: Era era => Rep era (Delegatee (EraCrypto era))
   VoteR :: Rep era Vote
 
+pattern RewardAcntR :: () => (Era era, t ~ RewardAccount (EraCrypto era)) => Rep era t
+pattern RewardAcntR = RewardAccountR
+{-# DEPRECATED RewardAcntR "Use `RewardAccountR` instead" #-}
+
```
lehins: Ditto, see my comment below

```suggestion
```

lehins (changes requested):
Very nice work! Thank you.

libs/cardano-ledger-core/src/Cardano/Ledger/Address.hs
Lucsanszky: Ah yes. Done, cheers! :+1:

libs/ledger-state/src/Cardano/Ledger/State/UTxO.hs
Lucsanszky: In that case, I guess I can just drop the final commit (30032d6de91eda5924565ef7b8f501010a1dc058) as well, right? Since that only renames some internal stuff.

lehins: Nah, don't drop it. Those renames are useful! Just the deprecation is unnecessary.

Lucsanszky: Ok, thanks for the clarification. It should be all good now (dropped some renames earlier but resurrected them from the reflog without the deprecation helpers).

libs/cardano-ledger-core/CHANGELOG.md
```diff
@@ -71,6 +78,8 @@
 
 ### `testlib`
 
+* Deprecate `deserialiseRewardAcntOld` in favor of `deserialiseRewardAccountOld`
+* Deprecate `mkVKeyRwdAcnt` in favor of `mkVKeyRewardAccount`
```
lehins: These two need to move into the `testlib` section of the `1.11.0.0` version

lehins (approved):
Perfect!
Thank you!

lehins:
With respect to failing CI, it is unrelated to your changes. Both are flaky tests, one of which I already have a fix for.

PR #4003, opened 2024-01-23
Prefix Set encoding with tag 258

lehins:
# Description

Starting with Conway we need to prefix all places with a tag where a Set is used when encoding

Original problem was reported in https://github.com/IntersectMBO/cardano-cli/issues/567

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/eras/conway/impl/cddl-files/conway.cddl
+++ b/eras/conway/impl/cddl-files/conway.cddl
@@ -430,9 +430,9 @@ drep_voting_thresholds =
   ]
 
 transaction_witness_set =
-  { ? 0: [+ vkeywitness ]
+  { ? 0: nonempty_set<vkeywitness>
   , ? 1: [+ native_script ]
-  , ? 2: [+ bootstrap_witness ]
+  , ? 2: nonempty_set<bootstrap_witness>
   , ? 3: nonempty_set<plutus_v1_script>
   , ? 4: [+ plutus_data ]
   , ? 5: redeemers
--- a/eras/conway/impl/cddl-files/conway.cddl
+++ b/eras/conway/impl/cddl-files/conway.cddl
@@ -434,7 +434,7 @@ transaction_witness_set =
   , ? 1: nonempty_set<native_script>
   , ? 2: nonempty_set<bootstrap_witness>
   , ? 3: nonempty_set<plutus_v1_script>
-  , ? 4: [+ plutus_data ]
+  , ? 4: nonempty_set<plutus_data>
   , ? 5: redeemers
   , ? 6: nonempty_set<plutus_v2_script>
   , ? 7: nonempty_set<plutus_v3_script>
--- a/eras/conway/impl/cddl-files/conway.cddl
+++ b/eras/conway/impl/cddl-files/conway.cddl
@@ -431,7 +431,7 @@ drep_voting_thresholds =
 
 transaction_witness_set =
   { ? 0: nonempty_set<vkeywitness>
-  , ? 1: [+ native_script ]
+  , ? 1: nonempty_set<native_script>
   , ? 2: nonempty_set<bootstrap_witness>
   , ? 3: nonempty_set<plutus_v1_script>
   , ? 4: [+ plutus_data ]
--- a/libs/cardano-data/testlib/Test/Cardano/Data/Arbitrary.hs
+++ b/libs/cardano-data/testlib/Test/Cardano/Data/Arbitrary.hs
@@ -13,7 +13,7 @@ import Test.Cardano.Ledger.Binary.Arbitrary ()
 import Test.QuickCheck
 
 instance (Arbitrary a, Ord a) => Arbitrary (OSet.OSet a) where
-  arbitrary = fmap (OSet.fromSet . Set.fromList) . shuffle . Set.toList =<< arbitrary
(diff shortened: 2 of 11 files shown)


libs/cardano-data/testlib/Test/Cardano/Data/Arbitrary.hs
```diff
@@ -13,7 +13,7 @@ import Test.Cardano.Ledger.Binary.Arbitrary ()
 import Test.QuickCheck
 
 instance (Arbitrary a, Ord a) => Arbitrary (OSet.OSet a) where
-  arbitrary = fmap (OSet.fromSet . Set.fromList) . shuffle . Set.toList =<< arbitrary
```
teodanciu, line 16: Curious if this is broken or just inefficient - can't figure out if the result is different with this change :thinking:

teodanciu (approved):
Looks good to me, but I wonder whether we need to change the cddl and if that could be related to the RoundTrip tests failures on CI

libs/cardano-data/testlib/Test/Cardano/Data/Arbitrary.hs
lehins: It is broken. The logic goes like this:
* Generate a random set, so there are no duplicates
* convert it to a list
* shuffle the list, so the order is no sorted
* convert it back to a set - effectively sorting it again and thus negating the whole shuffling operation
* convert it to an ordered set

So, we need to ensure the order is random, which is the fix in this PR

lehins:
>  I wonder whether we need to change the cddl and if that could be related to the RoundTrip tests failures on CI

Yes we do. It was a bit late for me last night to wait for the end of CI run.
@teodanciu Could you take a look at the last three commits?

libs/cardano-data/testlib/Test/Cardano/Data/Arbitrary.hs
teodanciu: Ah, forgot that creating a Set is sorting them. Makes sense, thanks.

eras/conway/impl/cddl-files/conway.cddl
```diff
@@ -430,9 +430,9 @@ drep_voting_thresholds =
   ]
 
 transaction_witness_set =
-  { ? 0: [+ vkeywitness ]
+  { ? 0: nonempty_set<vkeywitness>
   , ? 1: [+ native_script ]
-  , ? 2: [+ bootstrap_witness ]
+  , ? 2: nonempty_set<bootstrap_witness>
```
teodanciu, line 435: It wasn't per the plan in https://github.com/IntersectMBO/cardano-ledger/issues/3922
to have these two as nonempty_set, like this, no? 
Rather: 
```
? 0: [+ vkeywitness ] / { + $vkey => $signature }
 ? 2: [+ bootstrap_witness ]
```

Any reason for this discrepancy? To be fair, I didn't question why the plan was like that to begin with, but curious if something caused this change.

aniketd (approved):
Awesome! :raised_hands:

eras/conway/impl/cddl-files/conway.cddl
```diff
@@ -434,7 +434,7 @@ transaction_witness_set =
   , ? 1: nonempty_set<native_script>
   , ? 2: nonempty_set<bootstrap_witness>
   , ? 3: nonempty_set<plutus_v1_script>
-  , ? 4: [+ plutus_data ]
+  , ? 4: nonempty_set<plutus_data>
```
teodanciu, line 437: The deserialisation still doesn't fail if there are duplicates,  but that's more an implementation detail - this type reflects the set tag,which we need anyway. 
I will update the ticket https://github.com/IntersectMBO/cardano-ledger/issues/3922
to mention uncommenting the test we have for this: https://github.com/IntersectMBO/cardano-ledger/blob/master/eras/alonzo/impl/testlib/Test/Cardano/Ledger/Alonzo/Binary/TxWitsSpec.hs#L114

eras/conway/impl/cddl-files/conway.cddl
lehins: Very good observation. We'll change the plan to use a `nonemty_set` instead. It will be much more complicated to switch it to a Map, for very little benefit. 

The idea behind a Map, is that a unique can only produce a unique signature, so one of the `s1` and `s2` will have to be an invalid signature `[(k1, s1), (k1, s2)]`.

We check those signatures in the rules, so it maybe even better to produce a predicate failure saying that the signature is invalid, rather that a deserialization failure in this case.

eras/conway/impl/cddl-files/conway.cddl
lehins: Yes. Thank you.

teodanciu:
> > I wonder whether we need to change the cddl and if that could be related to the RoundTrip tests failures on CI
> 
> Yes we do. It was a bit late for me last night to wait for the end of CI run. @teodanciu Could you take a look at the last three commits?

I did, still looks good to me :+1:

eras/conway/impl/cddl-files/conway.cddl
teodanciu: Makes sense, thanks.

Issue #4000, opened 2024-01-22
Treasury Recuring Deposits


Hornan7:
## Context:
- We made a governance to withdraw 3 millions from the treasury 
- The action was ratified on Epoch 212 and enacted on Epoch 213 which is when I received the 3 millions, (With the help of our DRep that has over 51% of total voting stake + ourselves as the Constitutional Committee)

TxId: `d9fe066acdb60da93f0268baf337807122428eece43d898f8452e9d566d96ba9`
Index: `0`

## Issue:
As much as I like the treasury's devotion to my cause, it seems that its sending me recurring deposits of 3 millions every epochs
.....which is now at a total of 24 Millions 8 Epochs later. (Epoch 221)
```
[
    {
        "address": "stake_test1uqvj4r4h0fh5qhvpjmwedkcnj5xysvq2plutdapw0yq69mccrcnd7",
        "delegationDeposit": 2000000,
        "rewardAccountBalance": 24000368729198,
        "stakeDelegation": "pool1knw0wt62r2yv7v455j2p94lx98g9ltjv3a3aq8m4urlncwftf6e",
        "voteDelegation": "keyHash-c1ba49d52822bc4ef30cbf77060251668f1a6ef15ca46d18f76cc758"
    }
]
```
To be sure my pool wasn't involved, and the rewards i get from it, I did the same thing from an un-delegated registered stake address.

TxId: `7277b7cd4df7674ecc437c0c9fabb33bdbf9c60a13898a3b31729f27c450032a`
Index: `0`

This time with 64 millions and the second epoch after enactment it went up to 128 millions.
I spreaded it across 5 unregistered wallet for now to make sure to preserve the chain from its vote/stake influence. 
```
                           TxHash                                 TxIx        Amount
--------------------------------------------------------------------------------------
5d2e8ab96595145f09efc9d03a787fd1aeaeca94decaad6fa0167031c7753266     4        72000997370689 lovelace + TxOutDatumNone
                           TxHash                                 TxIx        Amount
--------------------------------------------------------------------------------------
5d2e8ab96595145f09efc9d03a787fd1aeaeca94decaad6fa0167031c7753266     2        16000000000000 lovelace + TxOutDatumNone
                           TxHash                                 TxIx        Amount
--------------------------------------------------------------------------------------
5d2e8ab96595145f09efc9d03a787fd1aeaeca94decaad6fa0167031c7753266     3        16000000000000 lovelace + TxOutDatumNone
[...]

PR #4004, opened 2024-01-24
Verify that enacted gov action is removed

lehins:
# Description

Changes to the TreasuryWithdrawal tests might seem to be redundant, but they were necessary for reproducing the bug in #4000 in `cardano-ledger-conway-1.11`

Improvements that are actually useful in this PR is the ability to submit multiple proposals in a single transaction. 

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/EpochSpec.hs
+++ b/eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/EpochSpec.hs
@@ -9,6 +9,7 @@
 
 module Test.Cardano.Ledger.Conway.Imp.EpochSpec (spec) where
 
+import Cardano.Ledger.Address (RewardAccount (..))
 import Cardano.Ledger.BaseTypes (EpochInterval (..), textToUrl)
 import Cardano.Ledger.Coin
 import Cardano.Ledger.Conway.Core
@@ -30,6 +31,8 @@ import Cardano.Ledger.Shelley.LedgerState (
  )
 import Cardano.Ledger.Val
 import Data.Default.Class (Default (..))
+import qualified Data.List.NonEmpty as NE
+import qualified Data.Map.Strict as Map
 import Data.Maybe (fromJust)
 import Data.Maybe.Strict (StrictMaybe (..))
 import Lens.Micro ((&), (.~))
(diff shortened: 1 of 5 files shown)


eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/EpochSpec.hs
```diff
@@ -100,20 +103,32 @@ spec =
       treasuryStart <- getsNES $ nesEsL . esAccountStateL . asTreasuryL
 
       rewardAcount <- registerRewardAccount
+      rewardAcountOther <- registerRewardAccount
+      govPolicy <- getGovPolicy
       let withdrawalAmount = Coin 666
-      govActionId <- submitTreasuryWithdrawals [(rewardAcount, withdrawalAmount)]
```
Soupstraw, line 104: It might be a good idea to keep around a test where we only try to submit a single treasury withdrawal as well.

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/ImpTest.hs
```diff
@@ -411,6 +419,15 @@ submitProposal ::
   ImpTestM era (GovActionId (EraCrypto era))
 submitProposal proposal = trySubmitProposal proposal >>= expectRightExpr
 
+submitProposals ::
+  (ShelleyEraImp era, ConwayEraTxBody era, HasCallStack) =>
+  NE.NonEmpty (ProposalProcedure era) ->
```
Soupstraw, line 424: It wouldn't be wrong to submit a transaction with zero proposals, so the non-empty list here seems to be a bit too restrictive.

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/ImpTest.hs
```diff
@@ -411,6 +419,15 @@ submitProposal ::
...
+  ImpTestM era (NE.NonEmpty (GovActionId (EraCrypto era)))
```
Soupstraw, line 425:
```suggestion
  [ProposalProcedure era] ->
  ImpTestM era [GovActionId (EraCrypto era)]
```

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/ImpTest.hs
lehins: It would be wrong to submit a transaction that is intended to have proposals, that does not include proposals. See the name of the function `submitProposals`.

If you want to submit a transaction without proposals, then `submitTx` should be used.

Also, it is not excessive, it is actually very useful, since it allows to extract at least that one returned governance action id without call to `head` or `error`.

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/ImpTest.hs
lehins: Same here. NonEmpty was intentional.

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/EpochSpec.hs
lehins: If we try to create a unit test for every possible permutation of all situations that can happen in our system we'll never get to the end of testing. It is ok to combine multiple scenarios in one unit test.

Soupstraw (approved):

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/EpochSpec.hs
lehins: Done

PR #4013, opened 2024-01-25
Remove `EnactState` from `ConwayGovState`

### LINKED ISSUE #3998: Stop storing EnactState in the ledger state
We are wrongfully storing temporary variables from `EnactState` in the ledger state: `ensTreasury`, `ensWithdrawals`, `ensPrevGovActionIds`.

They have confused people on multiple occasions and I am afraid can lead to bugs in the future. We need to retain only the relevant values from the enact state. This will require proper initialization of the EnactState during DRep pulser initialization and restoring the fields from the Enactstate once the pulser is done

This is the change that is needed to the `ConwayGovState`:
```diff
-  , cgEnactState :: !(EnactState era)
+  , cgCommittee :: !(StrictMaybe (Committee era))
+  , cgConstitution :: !(Constitution era)
+  , cgCurPParams :: !(PParams era)
+  , cgPrevPParams :: !(PParams era)
```

Soupstraw:
# Description

This PR removes the `EnactState` field from `ConwayGovState`. The fields that have to be stored in the state are copied over to the `ConwayGovState`.

Since we're not storing the whole enact state in the ledger state any more, we don't have to check the invariants in the `EPOCH` rule.

resolves #3998

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/eras/conway/impl/src/Cardano/Ledger/Conway/Governance.hs
+++ b/eras/conway/impl/src/Cardano/Ledger/Conway/Governance.hs
@@ -87,9 +87,12 @@ module Cardano.Ledger.Conway.Governance (
   proposalsSize,
   proposalsLookupId,
   proposalsActionsMap,
-  cgProposalsL,
-  cgEnactStateL,
-  cgDRepPulsingStateL,
+  cgsProposalsL,
+  cgsDRepPulsingStateL,
+  cgsCurPParamsL,
+  cgsPrevPParamsL,
+  cgsCommitteeL,
+  cgsConstitutionL,
   ensCommitteeL,
   ensConstitutionL,
   ensCurPParamsL,
@@ -144,6 +147,8 @@ module Cardano.Ledger.Conway.Governance (
   psDRepDistrL,
   psDRepStateL,
   RunConwayRatify (..),
+  govStatePrevGovActionIds,
+  mkEnactState,
 
   -- * Exported for testing
   pparamsUpdateThreshold,
@@ -159,6 +164,7 @@ import Cardano.Ledger.BaseTypes (
   StrictMaybe (..),
   UnitInterval,
   isSJust,
+  strictMaybeToMaybe,
  )
 import Cardano.Ledger.Binary (
   DecCBOR (..),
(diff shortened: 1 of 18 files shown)


eras/conway/impl/src/Cardano/Ledger/Conway/Governance.hs
```diff
@@ -501,39 +509,68 @@ toRatifyStatePairs cg@(RatifyState _ _ _ _) =
...
+  , cgsDRepPulsingState :: !(DRepPulsingState era)
+  -- ^ The 'cgsDRepPulsingState' field is a pulser that incrementally computes the stake distribution of the DReps
   --   over the Epoch following the close of voting at end of the previous Epoch. It assembles this with some of
   --   its other internal components into a (RatifyEnv era) when it completes, and then calls the RATIFY rule
   --   and eventually returns the updated RatifyState. The pulser is created at the Epoch boundary, but does
   --   no work until it is pulsed in the 'NEWEPOCH' rule, whenever the system is NOT at the epoch boundary.
+  , cgsCommittee :: !(StrictMaybe (Committee era))
+  , cgsConstitution :: !(Constitution era)
+  , cgsCurPParams :: !(PParams era)
+  , cgsPrevPParams :: !(PParams era)
```
lehins: Could you please put these four fields before the `cgsDRepPulsingState`? Make sure to adjust the `EncCBOR` instance too.

This will help later with implementing sharing of committee credentials when deserializing.

I know it is a bit pedantic, but, it is a very unique field, I'd like to keep it separated at least a little bit from others, so keeping it last will help with that.

I can make this change at some point later if you'd rather not deal with it now.

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Epoch.hs
```diff
@@ -308,10 +289,14 @@ epochTransition = do
...
+    -- Apply the values from the computed EnactState to the GovState
     govState1 =
       govState0
-        & cgProposalsL .~ newProposals
-        & cgEnactStateL .~ (newEnactState & ensPrevPParamsL .~ curPParams)
+        & cgsProposalsL .~ newProposals
+        & cgsPrevPParamsL .~ curPParams
+        & cgsCommitteeL .~ ensCommittee
+        & cgsConstitutionL .~ ensConstitution
+        & cgsCurPParamsL .~ ensCurPParams
```
lehins: Just so it is easier to see how pparams rotation happens
```suggestion
        & cgsCommitteeL .~ ensCommittee
        & cgsConstitutionL .~ ensConstitution
        & cgsCurPParamsL .~ ensCurPParams
        & cgsPrevPParamsL .~ curPParams
```

lehins (approved):
This is great. Exactly what we needed

Thank you.

aniketd (approved):
Grand! :raised_hands:

PR #4015, opened 2024-01-26
Shelley ledger spec: amend erratum on nonce stabilization window

amesgen:
Companion to https://github.com/IntersectMBO/ouroboros-consensus/pull/927


WhatisRT (approved):
OK, so this is effectively just documenting that the fix proposed in the erratum is now implemented in Conway. LGTM

PR #4016, opened 2024-01-26
Fix Arbitrary instance and invariant checking for `Proposals`

lehins:
# Description

* Fix flaky cddl test [triggered on CI](https://github.com/IntersectMBO/cardano-ledger/actions/runs/7690989169/job/20955610055?pr=4002#step:15:3664)
* Fix invalid fallback implementation of Mirror downloading, which made into a flake setop on CI resulting in [this failure](https://github.com/IntersectMBO/cardano-ledger/actions/runs/7695877143/job/20969683863)
* Fix flaky proposal tests that happened [here](https://github.com/IntersectMBO/cardano-ledger/actions/runs/7666414661/job/20894171243#step:15:4068) and [here](https://github.com/IntersectMBO/cardano-ledger/actions/runs/7654638435/job/20859039711?pr=4011#step:15:4054). More on that below

Fix test data generation and improve invariant checking:

* Fix `Arbitrary` instance for `Proposals`:

   * Allow empty to be generated
   * Generate `InfoAction` and `TreasuryWithdrawal`

* Remove `proposalsAreConsistent` in favor of `toPForest`. Besides
  checking the invariant it also constructs a forest of `Tree`s that
  can be used for debugging.

* Invariant checking improved:

  * Remove invariant that `pProps` domain should equal to all children
    and that domain of `pGraph` equals to the domain of `pProps`
  * Enforce that all children are uniqe and that parents are correctly
    specified and matches the inverse relation from a parent to the
    child

`toPForest` not only checks the invariant, but serves as a nice debugging tool, since it allows us to print the hierarchy in a readable fashion, eg.:

```haskell
{pfPParamUpdate =
x
, pfHardFork =
x
|
+- be4eaa8cc3a8e95fd0d02daaa55705a467f1327392dd7ea3749d128b8e0815c5#671359595
|
`- c0a4ef9bfeb257488525de250468e24b3c6340a6872ef366877c46a28a842c37#1165802016
   |
   `- 6cad5468f5320b34d411913292c2a40b1258fa11c7d5eecff6ad4b7325285422#1412147132
, pfCommittee =
a3e6fc1f9eec9a66c5ee56819f4a93658f6d018ef74d991d6d25242b8915d163#104556382
|
+- e6d969f5d7f10a8681aa8f9cf42328e98aaf653ec00353f340f37397c8181378#402911737
|  |
|  `- bca38b2ec2b04afb4d6fa61f3f1b8d5f3fbc77646626e4e7998b141cb41421fc#1225600682
|
`- e718c251f603af13b459fb22c4c0c0fdf03452ac06a3dbf45653ee87ce1c6ca7#1350459191
, pfConstitution =
f829acfb70d5a73396db0b1de29e34053cf80475b69382c9f390aef21e82d58a#767672592
|
`- 51f119927ec29d32df70c4bd90ba47b9542e59394ec9d8d5f8ffa1faebc56424#3360066490
   |
   +- 66a72530e4e79424a584e6f92f5a040f3b7122192622d06969e9318f4e5b49b8#45552015
   |
   +- 68b88d61580ee03573e0162a7546d1b5928b5798dad719ee40d2a801c77f279a#2339408668
   |
   `- dd04849e45e40f23775c0db04652641d8257593a7f5186625e3c836da57ecf66#179964133
}
```
# Checklist

- [ ] Self-reviewed the diff


eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Proposals.hs
```diff
@@ -407,16 +352,14 @@ mkProposals pgais omap = do
       initialProposals
       omap
   unless (omap == omap') $ fail "mkProposals: OMap is malformed"
-  unless (proposalsAreConsistent ps) $ fail "mkProposals: Proposals structure is inconsistent"
```
aniketd, line 410: Do we want to check the invariant here by using `toPForestEither`? Similar to the checks after addition or deletion of a proposal?

eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Proposals.hs
```diff
@@ -679,3 +631,145 @@ fromPrevGovActionIds (PrevGovActionIds pforest@(PForest _ _ _ _)) =
         & pfHardForkL . prRootL .~ pfHardFork
         & pfCommitteeL . prRootL .~ pfCommittee
         & pfConstitutionL . prRootL .~ pfConstitution
+
+---------------------
+-- Debugging tools --
+---------------------
+
+-- | Wraper type, which serves as a composition of @`Tree` . `StrictMaybe`@
```
aniketd, line 639:
```suggestion
-- | Wrapper type, which serves as a composition of @`Tree` . `StrictMaybe`@
```

eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Proposals.hs
```diff
@@ -679,3 +631,145 @@ fromPrevGovActionIds (PrevGovActionIds pforest@(PForest _ _ _ _)) =
...
+  let allNodes =
+        Set.unions
+          [ Set.map unGovPurposeId nodesPParamUpdate
+          , Set.map unGovPurposeId nodesHardFork
+          , Set.map unGovPurposeId nodesCommittee
+          , Set.map unGovPurposeId nodesConstitution
+          ]
+      propsMap = OMap.toMap pProps
+      guardUnknown = do
+        let unknown = allNodes Set.\\ Map.keysSet propsMap
```
aniketd, line 679: `Set.difference` Return elements of the first set not existing in the second set, in that, it is not a commutative operation. Should we instead check that the union and intersection of the two sets is the same? We could even test directly for equality?

eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Proposals.hs
lehins: Those checks are done on the addition of each GovActionState, which happens few lines above

eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Proposals.hs
aniketd: Maybe ```(A `union` B) \\ (A `intersection` B)```

lehins: No we should not. :smile:  

That would be the same wrong invariant that was in the previous implementation: 
https://github.com/IntersectMBO/cardano-ledger/blob/6fafb7269679cd32125e833ca70ca31f8d1bd4fc/eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Proposals.hs#L353-L354

lehins: There is, however, one invariant that I just thought of that would be useful to implement that is related to this:

Namely, making sure that every govAction that exists in the `propsMap` that can have a parent, specifies the same parent in the `pRoots` V `pGraph`

I'll send you a commit once I implement it.

lehins: > Maybe (A `union` B) \\ (A `intersection` B)

This won't work by itself.

```
(A `union` B) \\ (A `intersection` B) == (A \\ B) `union` (B \\ A)
```

In our case `(B \\ A)` is all of the governance actions that don't have parents (InfoAction and TreasuryWithdrawals), so we should not report them as unknown. It only is only `(A \\ B)` that are guaranteed to be invalid.

There is however part of `(B \\ A)` that can be invalid, we only need to check such B that has a parent. In which case, the check I suggest here will do the trick: https://github.com/IntersectMBO/cardano-ledger/pull/4016#discussion_r1469930329

eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Proposals.hs
aniketd: Oh yes, got it!

eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Proposals.hs
aniketd: Awesome :raised_hands:

eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Proposals.hs
```diff
@@ -679,3 +631,145 @@ fromPrevGovActionIds (PrevGovActionIds pforest@(PForest _ _ _ _)) =
...
+                ]
+        unless (Set.size allNodes == sumSizes) $ do
+          Left $
+            "Duplicate govActionIds found between different purposes: "
+              ++ show (sumSizes - Set.size allNodes)
+  guardUnknown
+  guardUnique
+  pure PForest {pfPParamUpdate, pfHardFork, pfCommittee, pfConstitution}
+
+toPTree :: (Ord a, Show a) => PRoot a -> Map a (PEdges a) -> Either String (TreeMaybe a, Set a)
```
aniketd, line 698: Do you think it is a good idea to have a property test for every successful `toPTree` to be checked against something similar to
```haskell
unfoldTree
  ( \(node, children, rest) -> 
      ( node
      , [(child, peChildren $ rest Map.! child, rest Map.delete child) | child <- children]
      )
  ) 
  (root, rootChildren, restOfGraph) -- untested code
```

aniketd (approved):
Super awesome! :100: :raised_hands:

eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Proposals.hs
lehins: Sorry, I am not sure what you mean. What would that property test be for?
`toPTree` is the thing that checks the invariant. What would this unfolder be checking?

aniketd: Oh nevermind, I was thinking from the perspective of constructing the tree starting from the root, but by using the library function just to test its equivalence with our implementation. It was just an idea, and I wasn't too sure.

lehins: Current implementation is essentially `unfoldTreeM`, except the loop is written manually.

We don't need to test that this conversion happens correctly. It would just be duplicating the logic.

PR #4021, opened 2024-01-29
Move unclaimed rewards from proposals to treasury

### LINKED ISSUE #4017: Conway BUG - Its possible to retire a stakeaddress while an active proposal is running
Tested on current SanchoNet

```console
$ cardano-node --version
cardano-node 8.7.3 - linux-x86_64 - ghc-8.10
git rev a4a8119b59b1fbb9a69c79e1e6900e91292161e7
```

Its possible to retire a stakeaddress while an active governance proposal is open and running!

What happens to the deposit-amount after the proposal reached its expire epoch? 

It would be good to have an information about such deposits via cardano-cli and the `query stake-address-info` request with a new output entry like `proposalDeposits` or similar.

Similar to the checks about remaining rewards on a stakeaddress, it should be impossible to retire a stakeadress while this keyhash is involved in an open proposal which required a deposit amount.

Other than with the rewards-address for a stakepool-registration (which would receive the pool deposit amount), the governance action proposal needs to be signed with the stake secret key that is set to be the receiving address for the deposit-return.

As an alternative, maybe a good idea to offer the retirement but refunding all deposits at the same time. Like with the stakeaddress deposit amount. Which than would also remove/end all governance action proposals for that keyhash?

@lehins

Soupstraw:
# Description

Unclaimed rewards from proposals are now moved to the treasury instead of being burned.

The `passTick` function now always checks that ADA is preserved after the TICK rule is run.

resolves #4017 

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/GovSpec.hs
+++ b/eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/GovSpec.hs
@@ -688,9 +688,10 @@ proposalWithRewardAccount ::
   ImpTestM era (ProposalProcedure era)
 proposalWithRewardAccount action = do
   rewardAccount <- registerRewardAccount
+  govActionDeposit <- getsNES $ nesEsL . curPParamsEpochStateL . ppGovActionDepositL
   pure
     ProposalProcedure
-      { pProcDeposit = Coin 123
+      { pProcDeposit = govActionDeposit
       , pProcReturnAddr = rewardAccount
       , pProcGovAction = action
       , pProcAnchor = def
--- a/eras/shelley/impl/src/Cardano/Ledger/Shelley/AdaPots.hs
+++ b/eras/shelley/impl/src/Cardano/Ledger/Shelley/AdaPots.hs
@@ -1,4 +1,5 @@
 {-# LANGUAGE DataKinds #-}
+{-# LANGUAGE DeriveGeneric #-}
 {-# LANGUAGE DerivingStrategies #-}
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE NamedFieldPuns #-}
@@ -15,29 +16,35 @@ module Cardano.Ledger.Shelley.AdaPots (
   Consumed (..),
   consumedTxBody,
   producedTxBody,
+  sumAdaPots,
 ) where
 
 import Cardano.Ledger.CertState (
   CertState (..),
-  DState (..),
-  PState (..),
+  Obligations (..),
   certsTotalDepositsTxBody,
   certsTotalRefundsTxBody,
+  obligationCertState,
   rewards,
+  sumObligation,
  )
 import Cardano.Ledger.Coin (Coin (..))
 import Cardano.Ledger.Compactible (fromCompact)
 import Cardano.Ledger.Core
+import Cardano.Ledger.Shelley.Governance (EraGov (..))
 import Cardano.Ledger.Shelley.LedgerState.Types (
   AccountState (..),
   EpochState (..),
   LedgerState (..),
   UTxOState (..),
+  lsUTxOStateL,
+  utxosGovStateL,
  )
 import Cardano.Ledger.Shelley.TxBody (unWithdrawals)
-import Cardano.Ledger.UMap (UView (RewDepUView), sumDepositUView, sumRewardsUView)
+import Cardano.Ledger.UMap (sumRewardsUView)
 import Cardano.Ledger.UTxO (UTxO (..), coinBalance, txInsFilter, txouts)
 import Data.Foldable (fold)
+import GHC.Generics (Generic)
 import Lens.Micro ((^.))
 
 data AdaPots = AdaPots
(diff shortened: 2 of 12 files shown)


eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/GovSpec.hs
```diff
@@ -892,3 +897,40 @@ ccVoteOnConstitutionFailsWithMultipleVotes = do
       [ inject $
           DisallowedVoters @era [(CommitteeVoter ccCred, committeeProposal)]
       ]
+
+depositMovesToTreasuryWhenStakingAddressUnregisters :: ConwayEraImp era => ImpTestM era ()
```
lehins: I would suggest moving this into `EpochSpec` module, cause the functionality it is testing has very little to do with the GOV rule

eras/shelley/impl/src/Cardano/Ledger/Shelley/AdaPots.hs
```diff
@@ -65,38 +79,49 @@ totalAdaPotsES (EpochState (AccountState treasury_ reserves_) ls _ _) =
     , utxoAdaPot = coins
     , keyDepositAdaPot = keyDeposits_
     , poolDepositAdaPot = poolDeposits_
```
lehins: These two should be removed, since they are no part of `obligationsPot`
```suggestion
```

eras/shelley/impl/src/Cardano/Ledger/Shelley/AdaPots.hs
```diff
@@ -65,38 +79,49 @@ totalAdaPotsES (EpochState (AccountState treasury_ reserves_) ls _ _) =
...
-    , depositsAdaPot = deposits
     , feesAdaPot = fees_
+    , obligationsPot =
+        Obligations
+          { oblStake = keyDeposits_
+          , oblProposal = zero
+          , oblPool = poolDeposits_
+          , oblDRep = drepDeposits
+          }
+          <> govStateObligations
```
lehins: Why not? ([`obligationCertState`](https://github.com/IntersectMBO/cardano-ledger/blob/70b969118ba0aa73b4cffca9fceccd1d44ec1218/libs/cardano-ledger-core/src/Cardano/Ledger/CertState.hs#L34)
```suggestion
    , obligationsPot = obligationCertState certState <> govStateObligations
```

eras/shelley/impl/src/Cardano/Ledger/Shelley/AdaPots.hs
```diff
@@ -65,38 +79,49 @@ totalAdaPotsES (EpochState (AccountState treasury_ reserves_) ls _ _) =
...
+sumAdaPots :: AdaPots -> Coin
+sumAdaPots
+  AdaPots
+    { treasuryAdaPot
+    , reservesAdaPot
+    , rewardsAdaPot
+    , utxoAdaPot
+    , feesAdaPot
+    , -- , keyDepositAdaPot  -- We don't count these two, as their
+    -- , poolDepositAdaPot -- sum is always depositsAdaPot
```
lehins: One more reason for those two fields to be removed
```suggestion
```

eras/shelley/impl/testlib/Test/Cardano/Ledger/Shelley/ImpTest.hs
```diff
@@ -740,53 +750,46 @@ runImpRule stsEnv stsState stsSignal = do
...
   ) =>
   ImpTestM era ()
 passTick = do
   impLastTick <- gets impLastTick
   curNES <- getsNES id
+  preAdaPots <- getAdaPots
   nes <- runImpRule @"TICK" () curNES impLastTick
   impLastTickL += 1
   impNESL .= nes
+  impAnn "Checking ADA preservation at the epoch boundary" $ do
```
lehins: It might be an overkill to do this check for every TICK, since nothing changes except the pulser state on every tick. It is a valuable check on every `EPOCH` rule though, do you think it would make sense to move it to `passEpoch` instead?

eras/shelley/impl/testlib/Test/Cardano/Ledger/Shelley/ImpTest.hs
```diff
@@ -1014,3 +1017,20 @@ constitutionShouldBe cUrl = do
...
+expectNotRegisteredRewardAddress (RewardAccount _ cred) = do
+  umap <- getsNES $ nesEsL . esLStateL . lsCertStateL . certDStateL . dsUnifiedL
+  Map.member cred (rdPairMap umap) `shouldBe` False
+
+expectTreasury :: HasCallStack => Coin -> ImpTestM era ()
+expectTreasury c =
+  impAnn "Checking treasury amount" $ do
+    treasuryAmt <- getsNES $ nesEsL . esAccountStateL . asTreasuryL
+    unless (treasuryAmt == c) . expectationFailure $
+      "Treasury amount mismatch\n\texpected: " <> show c <> "\n\tgot: " <> show treasuryAmt
```
lehins: hspec will do a better job at showing the difference
```suggestion
    c `shouldBe` treasuryAmt
```

lehins (approved):
Looks great. Just a few minor suggestions.

PR #4023, opened 2024-01-29
Add a function `registerInState` to `EraTransition`

dnadales:
... which allows to inject transition configuration data in the given `NewEpochState`.

This change allows the Ledger to define this logic in this component without needing to do it in the Consensus layer. In this way, we improve the separation of concerns.

Part of https://github.com/IntersectMBO/ouroboros-consensus/issues/899

# Checklist

- [x] Self-reviewed the diff


eras/conway/impl/src/Cardano/Ledger/Conway/Transition.hs
amesgen, line 1: This PR could also remove the `ConwayEraTransition` class as it won't be used by Consensus anymore after this PR.

eras/shelley/impl/src/Cardano/Ledger/Shelley/Transition.hs
```diff
@@ -85,6 +86,16 @@ class
...
 
+  registerInState ::
+    -- | Extract data from the given transition configuration and store it in the given state.
+    --
+    -- FIXME: Are there any other contextual consideration we need to comment on here?
+    --
+    -- FIXME: Perhaps 'injectInState' is a more fitting name?
+    TransitionConfig era ->
+    NewEpochState era ->
+    NewEpochState era
```
amesgen: I think the name could mention a keyword like "testing" or "nonMainnet" to emphasize that this functionality is not relevant (or rather, will even throw an imprecise exception) for mainnet.

eras/shelley/impl/src/Cardano/Ledger/Shelley/Transition.hs
amesgen, line 1: In Consensus, we don't directly use this modules, but rather the re-exports from `Cardano.Ledger.Api.Transition`. I think it makes to remove all the "fine-grained" registration functions (like `registerInitialFunds`) there and instead only export `registerInState` (or whatever name we end up choosing).

eras/shelley/impl/src/Cardano/Ledger/Shelley/Transition.hs
lehins: How about `injectTestState`?

> will even throw an imprecise exception) for mainnet.

Also describing this in the haddock would be good.

eras/shelley/impl/src/Cardano/Ledger/Shelley/Transition.hs
lehins: >  I think it makes to remove all the "fine-grained" registration

Yes, we can stop exporting them from this module:

@amesgen What about `createInitialState` and `registerInitialStaking` in here: https://github.com/IntersectMBO/ouroboros-consensus/blob/63fa2385a360eff10208c09f8ec055a13dae1984/ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Node/TPraos.hs#L345-L346

Do you still need them?

lehins (commented):
Looks good so far. These changes totally make sense and I also agree with @amesgen comments.

eras/shelley/impl/src/Cardano/Ledger/Shelley/Transition.hs
amesgen: > What about `createInitialState` and `registerInitialStaking` in here: https://github.com/IntersectMBO/ouroboros-consensus/blob/63fa2385a360eff10208c09f8ec055a13dae1984/ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Node/TPraos.hs#L345-L346
> 
> Do you still need them?

We hope to find this out later today :smile:

eras/shelley/impl/src/Cardano/Ledger/Shelley/Transition.hs
dnadales: I'll add some haddock :fish: As for the name, mentioning "test" in the name makes sense. I don't know if we want to modify it slightly to reflect that it is injecting into the state, so, for instance, `injectIntoTestState`. What do you think?

dnadales: I could copy the comments of the other functions. We might remove them or make them shorter if said functions are not exposed anymore.

```
    -- /Warning/ - Should only be used in testing and benchmarking. Will result in an error
    -- when 'NetworkId' is set to 'Mainnet'.
```

lehins: > reflect that it is injecting into the state, so, for instance, injectIntoTestState.

Sounds good with me, if you prefer that name instead

eras/conway/impl/src/Cardano/Ledger/Conway/Transition.hs
```diff
@@ -132,18 +105,23 @@ instance Crypto c => FromJSON (TransitionConfig (ConwayEra c)) where
...
+      where
+        tcInitialDRepsL =
+          protectMainnetLens "InitialDReps" null $
+            tcConwayGenesisL . lens cgInitialDReps (\g x -> g {cgInitialDReps = x})
 
 registerDelegs ::
-  forall era.
-  ConwayEraTransition era =>
+  forall era c.
+  (era ~ ConwayEra c, Crypto c) =>
```
dnadales: TODO: I should remove the type equality as well. I missed this :/

eras/shelley/impl/src/Cardano/Ledger/Shelley/Transition.hs
amesgen: FTR: We looked at that code in `TPraos`, and figured that we only need `createInitialState` and `injectIntoTestState`, but not `registerInitialStaking` or other fine-grained functions.

However, we need to change `createInitialState` to not register the initial funds; otherwise, `injectIntoTestState` fails to register the initial funds as they are already there :smile: See 77f8f22127e7ad5e2bb9e3a98bc53406dbe76fea.

eras/conway/impl/src/Cardano/Ledger/Conway/Transition.hs
```diff
@@ -100,16 +80,9 @@ instance Crypto c => EraTransition (ConwayEra c) where
...
-
-  tcDelegsL =
-    protectMainnetLens "ConwayDelegs" null $
-      tcConwayGenesisL . lens cgDelegs (\g x -> g {cgDelegs = x})
-
-  tcInitialDRepsL =
-    protectMainnetLens "InitialDReps" null $
-      tcConwayGenesisL . lens cgInitialDReps (\g x -> g {cgInitialDReps = x})
+tcConwayGenesisL ::
+  Lens' (TransitionConfig (ConwayEra c)) (ConwayGenesis (EraCrypto (ConwayEra c)))
```
amesgen:
```suggestion
  Lens' (TransitionConfig (ConwayEra c)) (ConwayGenesis c)
```

dnadales: Good catch! Thanks!

eras/conway/impl/src/Cardano/Ledger/Conway/Transition.hs
dnadales: Done. Thanks!

amesgen: FTR: The `ConwayEraTransition` class is actually necessary such that post-Conway eras can register the stuff that is new since Conway. I didn't think of that while writing the comment above.

eras/conway/impl/src/Cardano/Ledger/Conway/Transition.hs
```diff
@@ -12,8 +12,6 @@ module Cardano.Ledger.Conway.Transition (
   ConwayEraTransition (..),
   TransitionConfig (..),
   toConwayTransitionConfigPairs,
-  registerDelegs,
-  registerInitialDReps,
```
lehins, line 16: Could you please add an entry to the changelog (for the latest version) that these functions have been removed

eras/shelley/impl/src/Cardano/Ledger/Shelley/Transition.hs
```diff
@@ -24,8 +24,7 @@ module Cardano.Ledger.Shelley.Transition (
   tcInitialStakingL,
   mkShelleyTransitionConfig,
   createInitialState,
-  registerInitialFunds,
-  registerInitialStaking,
+  registerInitialFundsThenStaking,
```
lehins, line 27: These changes and addition of `injectIntoTestState` should definitely be listed in the changelog. Version needs to be bumped as well to `1.10.0.0` in the cabal (`1.9.0.0` -> `1.10.0.0`) file and the changelog (`1.9.0.1` -> `1.10.0.0`).

This will require some bounds to be adjusted in other packages in the ledger repo:

* All era packages and the api will need a lower bound updated (due to dependency on the new `injectIntoTestState`)
* while any other package in the ledger repo will just need an upper bound relaxed.

libs/cardano-ledger-api/src/Cardano/Ledger/Api/Transition.hs
```diff
@@ -21,28 +21,19 @@ module Cardano.Ledger.Api.Transition (
   tcInitialFundsL,
   tcInitialStakingL,
   createInitialState,
-  registerInitialFunds,
-  registerInitialStaking,
-  registerInitialDReps,
-  registerDelegs,
+  injectIntoTestState,
```
lehins, line 24: These changes should be listed in the changelog as well. No need to update the version since that has already been done in a recent PR.

lehins (changes requested):
PR looks great. Just some changelog entries and version+bounds adjustments.

libs/cardano-ledger-api/src/Cardano/Ledger/Api/Transition.hs
dnadales: Also mention that we updated the semantics of `createInitialState`.

lehins (approved):
Thank you!! This looks great.
Makes me happy to see improvements to this interface

PR #4030, opened 2024-01-31
Report Plutus `ScriptHash` upon execution failure

### LINKED ISSUE #3731: Report Plutus ScriptHash upon execution failure
Currently we report base64 encoded binary version of the script, but the script hash, which would be useful for debugging.

Add the output of `hashScript` here:
https://github.com/input-output-hk/cardano-ledger/blob/03bad4b6a9522e43257f436d3e0f20b5017328de/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxInfo.hs#L900

Also improving failure message here
https://github.com/input-output-hk/cardano-ledger/blob/03bad4b6a9522e43257f436d3e0f20b5017328de/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxInfo.hs#L865
to:
```haskell
firstLines = 
  [ "The " ++ show lang ++ " script failed:"
  , "Base64-encoded script bytes:"
  ]
```

Lucsanszky:
# Description
Resolves #3731 

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Plutus/Evaluate.hs
+++ b/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Plutus/Evaluate.hs
@@ -143,7 +143,7 @@ collectPlutusScriptsWithContext ::
   PParams era ->
   Tx era ->
   UTxO era ->
-  Either [CollectError era] [PlutusWithContext]
+  Either [CollectError era] [PlutusWithContext (EraCrypto era)]
 collectPlutusScriptsWithContext epochInfo sysStart pp tx utxo =
   -- TODO: remove this whole complicated check when we get into Conway. It is much simpler
   -- to fail on a CostModel lookup in the `apply` function (already implemented).
@@ -168,15 +168,15 @@ collectPlutusScriptsWithContext epochInfo sysStart pp tx utxo =
     ScriptsProvided scriptsProvided = getScriptsProvided utxo tx
     AlonzoScriptsNeeded scriptsNeeded = getScriptsNeeded utxo (tx ^. bodyTxL)
     neededPlutusScripts =
-      mapMaybe (\(sp, sh) -> (,) sp <$> lookupPlutusScript scriptsProvided sh) scriptsNeeded
+      mapMaybe (\(sp, sh) -> (,) (sh, sp) <$> lookupPlutusScript scriptsProvided sh) scriptsNeeded
     usedLanguages = Set.fromList $ map (plutusScriptLanguage . snd) neededPlutusScripts
 
-    getScriptWithRedeemer (plutusPurpose, plutusScript) =
+    getScriptWithRedeemer ((plutusScriptHash, plutusPurpose), plutusScript) =
       let redeemerIndex = hoistPlutusPurpose toAsIndex plutusPurpose
        in case lookupRedeemer redeemerIndex $ tx ^. witsTxL . rdmrsTxWitsL of
-            Just (d, exUnits) -> Right (plutusScript, plutusPurpose, d, exUnits)
+            Just (d, exUnits) -> Right (plutusScript, plutusPurpose, d, exUnits, plutusScriptHash)
             Nothing -> Left (NoRedeemer (hoistPlutusPurpose toAsItem plutusPurpose))
-    apply (plutusScript, plutusPurpose, d, exUnits) = do
+    apply (plutusScript, plutusPurpose, d, exUnits, plutusScriptHash) = do
       let lang = plutusScriptLanguage plutusScript
       costModel <- maybe (Left (NoCostModel lang)) Right $ Map.lookup lang costModels
       case mkPlutusScriptContext plutusScript plutusPurpose pp epochInfo sysStart utxo tx of
@@ -188,6 +188,7 @@ collectPlutusScriptsWithContext epochInfo sysStart pp tx utxo =
                   PlutusWithContext
                     { pwcProtocolVersion = protVerMajor
                     , pwcScript = Left plutus
+                    , pwcScriptHash = plutusScriptHash
                     , pwcDatums = PlutusDatums (getPlutusData <$> datums)
                     , pwcExUnits = exUnits
                     , pwcCostModel = costModel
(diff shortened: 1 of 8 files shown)


eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Plutus/Evaluate.hs
```diff
@@ -209,17 +211,17 @@ merge f (x : xs) zs = merge f xs (gg x zs)
 
 -- | Evaluate a list of Plutus scripts. All scripts in the list must evaluate to `True`.
 evalPlutusScripts ::
-  EraTx era =>
+  (EraTx era, Crypto c) =>
   Tx era ->
-  [PlutusWithContext] ->
-  ScriptResult
+  [PlutusWithContext c] ->
+  ScriptResult c
```
lehins: `c` and `era` should never appear together, since `EraCrypto era` is always the crypto for the era 
```suggestion
  [PlutusWithContext (EraCrypto era)] ->
  ScriptResult (EraCrypto era)
```

Lucsanszky: Ah yes, I was planning to change those since at one point I realised that there is no point in introducing an extra type variable where we already have the `era`, it just seemed redundant. Hence why at certain parts I only used `EraCrypto era` and was about to clean up the rest. But I didn't know that having both is actually discouraged for other reasons than redundancy. Thanks!

By the way, what necessitated the introduction of `EraCrypto` in the first place (given the fact that `Crypto c` already seems extraneous at this point since it is always `Crypto StandardCrypto`)?

teodanciu: Not sure if this is what you're asking, but: if you have an `Era era` constraint (or a super class, eg: `EraTx` in this case),  then you have already the `Crypto` type for it, and you reference it with `EraCrypto era`   (https://github.com/IntersectMBO/cardano-ledger/blob/master/eras/conway/impl/src/Cardano/Ledger/Conway/Era.hs#L42)

teodanciu: Oh or maybe you're asking why was it necessary to be introduced at all?  Nevermind my comment if so :)

Lucsanszky: The latter. :) I was wondering why it was introduced at all. I mean sure, it is more descriptive than the `Crypto c` constraint but still, I was wondering if there were some other reasons.
Regardless, thanks for the clarification regarding the `Era` constraints as well. :)

eras/alonzo/test-suite/test/Test/Cardano/Ledger/Alonzo/PlutusScriptExamples.hs
```diff
@@ -178,6 +179,7 @@ explainTest plutus mode ds =
         PlutusWithContext
           { pwcProtocolVersion = eraProtVerLow @era
           , pwcScript = Left plutus
+          , pwcScriptHash = mkScriptHash @(EraCrypto era) 1
```
Lucsanszky: It seems that it might be okay to use a "dummy hash" here. Regardless, I thought that I will point this out and ask if this could cause some issues down the line?

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Examples/AlonzoCollectInputs.hs
```diff
@@ -105,6 +106,7 @@ collectTwoPhaseScriptInputsOutputOrdering = do
           testSystemStart
           (initUTxO apf)
           (validatingTx apf)
+    scriptHash = ScriptHash "58503a1d89a21fc9fc53d6a7cccef47341175a8f47636f57ccbdca2d"
```
Lucsanszky: ~Not the most elegant solution and a PITA to maintain if the script changes but I wanted to avoid the `getScriptsNeeded` and `getScriptsHashesNeeded` dance. By the way, is there an easier way to get the hash of the script here? Also, would it make sense to introduce some helpers for hashing the script (if there aren't any already)?~ #4037 :rocket:

Lucsanszky:
I will mark this ready for review once #4037 is merged.

libs/cardano-ledger-core/CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## 1.11.0.0
 
+* Add `pwcScriptHash` field to `PlutusWithContext`
```
lehins, line 5:
```suggestion
* Add `pwcScriptHash` field to `PlutusWithContext`
* Parameterize `PlutusWithContext c` on crypto.
```

libs/cardano-ledger-core/src/Cardano/Ledger/Plutus/Evaluate.hs
```diff
@@ -163,34 +162,36 @@ instance EncCBOR PlutusDatums where
...
 
-instance FromCBOR PlutusWithContext where
-  fromCBOR = Plain.decodeRecordNamed "PlutusWithContext" (const 5) $ do
+instance Crypto c => FromCBOR (PlutusWithContext c) where
+  fromCBOR = Plain.decodeRecordNamed "PlutusWithContext" (const 6) $ do
     pwcProtocolVersion <- fromCBOR
     toPlainDecoder pwcProtocolVersion $ decodeWithPlutus $ \plutus -> do
       let lang = plutusLanguage plutus
           pwcScript = Left plutus
+      pwcScriptHash <- decCBOR
```
lehins, line 184: There are two approaches we can take here:
* Either not having script hash in the decoding at all
* or maybe use the encoded one to validate that the hash of the computed one is correct.


This type is only ever encoded for debugging, so we don't need to worry about efficiency here, therefore I think validation is the way to go:
```suggestion
      pwcScriptHash <- decCBOR
      unless (pwcScriptHash == hasPlutusScript plutus) $ 
        fail $ "ScriptHash mismatch. Encoded: " <> show pwcScriptHash <> 
          " doesn't match the actual: " <> show (hasPlutusScript plutus)
```

lehins (approved):
Perfect!
Thank you!

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Plutus/Evaluate.hs
lehins: > I was wondering why it was introduced at all.

We were discussing just that on Wednesday meeting. :smile: 

I suspect there was two reasons:
* Ability to use different crypto algortihms  during testing. Which is sort of helpful for VRF, but not at all for hashes and signing algorithms
* Assumption that it would be more versatile, which proved to be incorrect.

In other words,  this is the decision that was made early on and now we are stuck with it. But, maybe we'll be able to get rid of it in the future. I have a descent plan on how it can be made possible.

Lucsanszky: > I suspect there was two reasons:
> 
>     * Ability to use different crypto algortihms  during testing. Which is sort of helpful for VRF, but not at all for hashes and signing algorithms
> 
>     * Assumption that it would be more versatile, which proved to be incorrect.

~But the above is about the `Crypto` constraint, right? I remember we were discussing that. I was wondering about the history behind `EraCrypto`. Was it introduced for convenience?~ Never mind! I did some more reading and see that `EraCrypto` was just a rename from `Crypto`. All is clear now, thanks!

PR #4031, opened 2024-01-31
Further proposals improvements. Take 2

lehins:
# Description

This is a follow up to: #4025

One important place where invariant was not verified was when enacted proposals are being removed. With adding the check there we no longer need to bother adding any other checks, as long as other internal functions are not being used for modifying the Proposals

Other changes are minor cleanups and performance improvements

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Proposals.hs
+++ b/eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Proposals.hs
@@ -249,19 +249,19 @@ proposalsAddAction gas ps = withGovActionParent gas (Just psWithGas) update
       StrictMaybe (GovPurposeId p era) ->
       GovPurposeId p era ->
       Maybe (Proposals era)
-    update forestL parent newId
-      | parent == ps ^. pRootsL . forestL . prRootL =
+    update govRelationL parent newId
+      | parent == ps ^. pRootsL . govRelationL . prRootL =
           Just $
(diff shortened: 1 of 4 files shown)


eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Proposals.hs
```diff
@@ -339,55 +339,51 @@ proposalsRemoveIds gais ps =
...
+          then
+            ( roots & govRelationL . prChildrenL %~ Set.delete gpi
+            , graph & govRelationL . pGraphNodesL %~ Map.delete gpi
+            )
+          else
+            ( roots
+            , graph
+                & govRelationL . pGraphNodesL %~ Map.delete gpi
+                & case parent of
+                  SNothing -> id
```
aniketd: Correct me if I'm wrong, but we _could_ have an `assert` here because this should not happen, but since this invariant is being checked in the check-function, we can ignore it is this branch?

aniketd (approved):
Awesome! :raised_hands: :100:

eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Proposals.hs
lehins: Yes, you are totally right. I've added the `assert` here, despite it being redundant, it serves as a good indicator to the reader that this is an impossible case.

Nice observation.

PR #4032, opened 2024-02-01
Make Haddock happy

zliu41:
Haddock doesn't seem to like `-- |` on constructor arguments.


lehins:
Those two variants are both legal on arguments: `-- |` and `-- ^`, except older haddocks versions do not support either of them. Also fourmolu will change `-- ^` to this `-- |` for consistency. So, the way to make haddock happy is to use newer ghc version with newer haddock

PR #4040, opened 2024-02-02
Redeemers serialization fixes

lehins:
# Description

This PR addresses part of the #3922 ticket, namely switched redeemers to use Map encoding by default

It also fixes CDDL for Conway PParams 

Extra nice cleanup is addition of Map conversion functions and deprecation of `keyBy`

# Checklist

- [x] Self-reviewed the diff


eras/conway/impl/cddl-files/conway.cddl
```diff
@@ -408,6 +408,7 @@ protocol_param_update =
   , ? 30: coin                   ; governance action deposit
   , ? 31: coin                   ; DRep deposit
   , ? 32: epoch                  ; DRep inactivity period
+  , ? 33: coin                   ; MinFee RefScriptCoinsPerByte
```
teodanciu, line 411: Oops :grimacing: 
I wonder why round trip tests didn't fail, is it normal?

lehins: Yes, cause we currently don't test that data generated in Haskell conforms with the CDDL spec. In the matter of fact we can't test that right now, because CDDL tests are only as good as ruby generators. However, cuddle will solve that problem for us. :wink:

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxWits.hs
```diff
@@ -544,15 +552,30 @@ instance AlonzoEraScript era => DecCBOR (Annotator (RedeemersRaw era)) where
...
+          nonEmptyDecoder
+          (RedeemersRaw . Map.fromList . NE.toList)
+      decodeMapRedeemers = decodeRedeemersWith $ do
+        (_, xs) <- decodeListLikeWithCount decodeMapLenOrIndef (:) $ \_ -> do
+          ptr <- decCBOR
+          (annData, exUnits) <- decCBOR
+          pure $ (\d -> (ptr, (d, exUnits))) <$> annData
+        case NE.nonEmpty xs of
+          Nothing -> fail "Expected redeemers map to be non-empty"
+          Just neList -> pure $ NE.reverse neList
```
teodanciu, line 576: Is the `reverse` because passing `:` to  `decodeListLikeWithCount` is reversing the list ?

lehins: Correct.

eras/conway/impl/cddl-files/conway.cddl
teodanciu: But something I don't understand is:  the ruby tool generates data based on the cddl, so it wouldn't have generated a value for this field.  But the field is not optional in PParams no? (for @Identity).  In the  deserialization it doesn't seem optional either. But it must be, since all these fields in the cddl are optional.. So am I misreading and it's actually optional in the deserialization? 
But don't worry, I will ask you when we speak next, this is not related to the change or PR!

teodanciu (approved):
Looks good to me

eras/conway/impl/cddl-files/conway.cddl
lehins: It is optional, because this CDDL is for PParamsUpdate, not PParams :wink:

teodanciu: Oh yeah! Thanks, I can sleep better now  :laughing:

PR #4042, opened 2024-02-02
Fix the prose being inconsistent with the figure

WhatisRT:
# Description

Found by @HeinrichApfelmus. @TimSheard  can you check that the second field in `RewardUpdate` does go into the reserves in the code?

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/eras/shelley/formal-spec/epoch.tex
+++ b/eras/shelley/formal-spec/epoch.tex
@@ -1392,7 +1392,7 @@ The $\fun{createRUpd}$ function does the following:
     labeled ``go'').
     As given by $\fun{maxPool}$, each pool can receive a maximal amount, determined by its
     performance.  The difference between the maximal amount and the actual amount received is
-    added to the amount moved to the treasury.
+    added to the amount moved to the reserves.
   \item The fee pot will be reduced by $\var{feeSS}$.
 \end{itemize}
 


HeinrichApfelmus:
Credit for prompting the cross-check between prose and figure goes to @KtorZ .

KtorZ:
Was it me 😅? Maybe, I just remember complaining about stuff, which sounds a lot like me.

eras/shelley/formal-spec/epoch.tex
```diff
@@ -1392,7 +1392,7 @@ \subsection{Reward Update Calculation}
     labeled ``go'').
     As given by $\fun{maxPool}$, each pool can receive a maximal amount, determined by its
     performance.  The difference between the maximal amount and the actual amount received is
-    added to the amount moved to the treasury.
+    added to the amount moved to the reserves.
```
lehins, line 1395: I've tried to understand and verify that this functionality is indeed the same in the implementation, but I failed to do so. I'd need to spend quite a bit of time to understand how pool rewards work in order to be able to say with confidence that this is exactly what happens. Maybe it will be easier for @TimSheard to confirm this, since he was very much involved with reward computation

HeinrichApfelmus: This particular piece of prose concerns to the fate of the quantity Δr2 in Figure 51.

The quantity Δr2 is the part of the reward pot that could have been paid out in this epoch, but was not, because the pools weren't saturated or had suboptimal owner stake.

The relation to the prose is: In the result of `createRUpd`, is Δr2 added to the first component ("moved to the treasury") or the second component ("moved to the rewards") of the quadruple?

Figure 51 adds it to the reserves. The question is whether the `cardano-ledger` source code does the same thing. (I believe so, but it's been a while since I last looked at it.)

createRUpd

TimSheard: The situation does not look good. See the file  "cardano-ledger/eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState/IncrementalStake.hs"


f```
ilteredRewards@FilteredRewards
        { frRegistered
        , frTotalUnregistered
        } = filterAllRewards' (rs ru) prevProVer dState
      -- Note: domain filteredRewards is a subset of domain (rewards dstate)
      registeredAggregated = aggregateCompactRewards prevProVer frRegistered
      -- Note: domain registeredAggregated is a subset of domain (rewards dstate)
      as' =
        as
          { asTreasury = addDeltaCoin (asTreasury as) (deltaT ru) <> frTotalUnregistered
          , asReserves = addDeltaCoin (asReserves as) (deltaR ru)
          }
```

Comment (TimSheard) [reply eras/shelley/formal-spec/epoch.tex:1395]: This will be easy to fix, but for a long-time, money has been going to the wrong place.

Comment (TimSheard) [reply eras/shelley/formal-spec/epoch.tex:1395]: I guess people pay attention to the "prose" not the diagram

Comment (TimSheard) [reply eras/shelley/formal-spec/epoch.tex:1395]: Maybe this is right.  (deltaR ru) sees to be the thing we are discussing. and that is added to the reserves.

Comment (TimSheard) [reply eras/shelley/formal-spec/epoch.tex:1395]: The key to figuring this out is found in the file
"cardano-ledger/eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState/IncrementalStake.hs"
where the pulser returns a RewardUpdate. I actually wrote this code, but it was based on
older code written by Jared, and I carefully used the same varaible names used in that code.
It appears that those names are based on the names in the figure above.

In that file we see the following RewardUpdate is returned.

```
pure
      ( RewardUpdate
          { deltaT = DeltaCoin deltaT1
          , deltaR = invert (toDeltaCoin deltaR1) <> toDeltaCoin deltaR2
          , rs = rs''
          , deltaF = invert (toDeltaCoin feesSS)
          }
```

Both the variable names and the actions seem to accurately track the figure.	  
Then in the function 'applyRUpdFiltered' The account state is updated as follows
```
as'= as { asTreasury = addDeltaCoin (asTreasury as) (deltaT ru) <> frTotalUnregistered
         , asReserves = addDeltaCoin (asReserves as) (deltaR ru)
         }
```
Where `ru` is the the RewardUpdate returned above. So it seems the reserves are updated as expected.

TimSheard (approved):
I carefully looked at the code, and explained what I found. So I believe the reserves are updated  as they should be.

PR #4049, opened 2024-02-05
Support for Plutus scripts in ImpTest

### LINKED ISSUE #4022: Support for Plutus scripts in ImpTest
Native script support has been added in the imp test framework. We now need support for Plutus scripts in a similar fashion

We also need to diversify what sort of scripts we use for testing, besides the alwaysSuceeds and alwaysFails

Soupstraw:
# Description

This PR adds the necessary fixup phases to make it possible to submit transactions which need to run scripts. I also converted the fixup pipeline to a point-free representation in order to improve readability and make it easier to modify the pipeline.

resolves #4022

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/eras/alonzo/impl/test/Main.hs
+++ b/eras/alonzo/impl/test/Main.hs
@@ -7,6 +7,7 @@ import qualified Test.Cardano.Ledger.Alonzo.Binary.CddlSpec as CddlSpec
 import qualified Test.Cardano.Ledger.Alonzo.Binary.CostModelsSpec as CostModelsSpec
 import qualified Test.Cardano.Ledger.Alonzo.Binary.TxWitsSpec as TxWitsSpec
 import qualified Test.Cardano.Ledger.Alonzo.BinarySpec as BinarySpec
+import qualified Test.Cardano.Ledger.Alonzo.Imp as AlonzoImp
 import Test.Cardano.Ledger.Alonzo.ImpTest ()
 import Test.Cardano.Ledger.Common
 import Test.Cardano.Ledger.Core.JSON (roundTripJsonEraSpec)
@@ -21,6 +22,7 @@ main =
       roundTripJsonEraSpec @Alonzo
       describe "Imp" $ do
         ShelleyImp.spec @Alonzo
+        AlonzoImp.spec @Alonzo
       describe "CostModels" $ do
         CostModelsSpec.spec @Alonzo
       describe "TxWits" $ do
--- a/eras/alonzo/impl/testlib/Test/Cardano/Ledger/Alonzo/Imp/UtxosSpec.hs
+++ b/eras/alonzo/impl/testlib/Test/Cardano/Ledger/Alonzo/Imp/UtxosSpec.hs
@@ -1,0 +1,66 @@
+{-# LANGUAGE AllowAmbiguousTypes #-}
+{-# LANGUAGE GADTs #-}
+{-# LANGUAGE RankNTypes #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeApplications #-}
+
+module Test.Cardano.Ledger.Alonzo.Imp.UtxosSpec (spec) where
+
+import Cardano.Ledger.Address (Addr (..))
+import Cardano.Ledger.Alonzo.Core (AlonzoEraTxOut (..), AlonzoEraTxWits (..))
+import Cardano.Ledger.BaseTypes (Inject (..), Network (..), StrictMaybe (..))
+import Cardano.Ledger.Coin (Coin (..))
+import Cardano.Ledger.Core (EraTx (..), EraTxBody (..), EraTxOut (..))
+import Cardano.Ledger.Credential (Credential (..), StakeReference (..))
+import Cardano.Ledger.Plutus.Data (Data (..), hashData)
+import qualified Data.Sequence.Strict as SSeq
+import qualified Data.Set as Set
+import Lens.Micro ((&), (.~))
+import qualified PlutusLedgerApi.V1 as P
+import Test.Cardano.Ledger.Alonzo.ImpTest (
+  ImpTestState,
+  PlutusArgs (..),
+  ShelleyEraImp,
+  impAddPlutusScript,
+  submitTxAnn,
+  submitTxAnn_,
+ )
+import Test.Cardano.Ledger.Common
+import Test.Cardano.Ledger.Core.Utils (txInAt)
+import Test.Cardano.Ledger.Plutus (ScriptTestContext (..))
+import Test.Cardano.Ledger.Plutus.Examples (guessTheNumber3)
+
+spec ::
+  forall era.
+  ( ShelleyEraImp era
+  , AlonzoEraTxWits era
+  , AlonzoEraTxOut era
+  ) =>
+  SpecWith (ImpTestState era)
+spec = describe "UTXOS" $ do
+  it "Plutus script transactions are fixed up" $ do
+    let spendDatum = P.I 3
+    shSpending <-
+      impAddPlutusScript
+        ScriptTestContext
+          { stcScript = guessTheNumber3
+          , stcArgs =
+              PlutusArgs
+                { paSpendDatum = Just spendDatum
+                , paData = P.I 3
+                }
+          }
+    tx0 <-
+      submitTxAnn "Sumbit a transaction with a script output" $
+        mkBasicTx mkBasicTxBody
+          & bodyTxL . outputsTxBodyL
+            .~ SSeq.singleton
+              ( mkBasicTxOut
+                  (Addr Testnet (ScriptHashObj shSpending) StakeRefNull)
+                  (inject (Coin 100))
+                  & dataHashTxOutL .~ SJust (hashData @era $ Data spendDatum)
+              )
+    submitTxAnn_ "Submit a transaction that consumes the script output" $
+      mkBasicTx mkBasicTxBody
+        & bodyTxL . inputsTxBodyL
+          .~ Set.singleton (txInAt (0 :: Int) tx0)
--- a/eras/alonzo/impl/testlib/Test/Cardano/Ledger/Alonzo/ImpTest.hs
+++ b/eras/alonzo/impl/testlib/Test/Cardano/Ledger/Alonzo/ImpTest.hs
@@ -1,43 +1,219 @@
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE NumericUnderscores #-}
+{-# LANGUAGE RankNTypes #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TupleSections #-}
+{-# LANGUAGE TypeApplications #-}
+{-# LANGUAGE TypeFamilies #-}
+{-# LANGUAGE TypeOperators #-}
 {-# LANGUAGE UndecidableInstances #-}
 {-# OPTIONS_GHC -Wno-orphans #-}
 
 module Test.Cardano.Ledger.Alonzo.ImpTest (
   module ImpTest,
   initAlonzoImpNES,
+  PlutusArgs (..),
+  impAddPlutusScript,
+  alonzoFixupTx,
 ) where
 
 import Cardano.Crypto.DSIGN.Class (DSIGNAlgorithm (..))
 import Cardano.Crypto.Hash (Hash)
 import Cardano.Ledger.Alonzo (AlonzoEra)
-import Cardano.Ledger.Alonzo.PParams (AlonzoEraPParams, ppMaxValSizeL)
-import Cardano.Ledger.Coin (Coin)
-import Cardano.Ledger.Core (EraIndependentTxBody)
+import Cardano.Ledger.Alonzo.Core (AlonzoEraScript (..), AlonzoEraTxOut (..), Era (..), EraGov)
+import Cardano.Ledger.Alonzo.PParams (AlonzoEraPParams, getLanguageView, ppCostModelsL, ppMaxTxExUnitsL, ppMaxValSizeL)
+import Cardano.Ledger.Alonzo.Scripts (ExUnits (..), plutusScriptLanguage, toAsIndex)
+import Cardano.Ledger.Alonzo.Tx (hashScriptIntegrity)
+import Cardano.Ledger.Alonzo.TxBody (AlonzoEraTxBody (..))
+import Cardano.Ledger.Alonzo.TxWits (AlonzoEraTxWits (..), Redeemers (..), TxDats (..))
+import Cardano.Ledger.Alonzo.UTxO (AlonzoScriptsNeeded (..))
+import Cardano.Ledger.Coin (Coin (..))
+import Cardano.Ledger.Core (
+  EraIndependentTxBody,
+  EraScript (..),
+  EraTx (..),
+  EraTxBody (..),
+  EraTxWits (..),
+  ScriptHash,
+ )
 import Cardano.Ledger.Crypto (Crypto (..))
+import Cardano.Ledger.Plutus.Data (Data (..), Datum (..), binaryDataToData, hashData)
+import Cardano.Ledger.Plutus.Language (Language (..), Plutus, PlutusLanguage, hashPlutusScript)
 import Cardano.Ledger.Shelley.LedgerState (
   NewEpochState,
   StashedAVVMAddresses,
   curPParamsEpochStateL,
+  esLStateL,
+  lsUTxOStateL,
   nesEsL,
+  utxosUtxoL,
  )
+import Cardano.Ledger.Shelley.UTxO (EraUTxO (..), txinLookup)
+import Cardano.Ledger.TxIn (TxIn)
+import Control.Monad (forM)
 import Data.Default.Class (Default)
-import Lens.Micro ((&), (.~))
+import Data.Foldable (Foldable (..))
+import qualified Data.Map.Strict as Map
+import Data.Maybe (catMaybes)
+import Data.Set (Set)
+import qualified Data.Set as Set
+import Lens.Micro ((%~), (&), (.~), (<>~), (^.))
+import Lens.Micro.Mtl (use, (%=), (.=))
 import Test.Cardano.Ledger.Allegra.ImpTest (impAllegraSatisfyNativeScript)
 import Test.Cardano.Ledger.Alonzo.TreeDiff ()
 import Test.Cardano.Ledger.Common
+import Test.Cardano.Ledger.Plutus (PlutusArgs (..), ScriptTestContext (..), testingCostModels)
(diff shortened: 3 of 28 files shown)


eras/shelley/impl/testlib/Test/Cardano/Ledger/Shelley/ImpTest.hs
```diff
@@ -201,13 +212,31 @@ import UnliftIO.Exception (
...
+  , paSpendDatum :: Maybe Data
+  }
+
+data ScriptTestContext where
+  ScriptTestContext ::
+    PlutusLanguage l =>
+    { astcScript :: Plutus l
+    , astcArgs :: PlutusArgs
+    } ->
+    ScriptTestContext
```
lehins: Let's put these two types into `Test.Cardano.Ledger.Plutus` in the `cardano-ledegr-core:testlib`

eras/alonzo/impl/testlib/Test/Cardano/Ledger/Alonzo/Imp/ScriptSpec.hs
```diff
@@ -0,0 +1,66 @@
+{-# LANGUAGE AllowAmbiguousTypes #-}
+{-# LANGUAGE GADTs #-}
+{-# LANGUAGE RankNTypes #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeApplications #-}
+
+module Test.Cardano.Ledger.Alonzo.Imp.ScriptSpec (spec) where
```
lehins: I would like to have all Imp tests to be placed into the modules that match the rules that they are testing.
We can add more **hidden** modules if it will be necessary with more specific names (e.g `Test.Cardano.Ledger.Alonzo.Imp.Utxo.ScriptSpec`), but I think we need to have this hierarchy, in order for developers to easily infer which ledger rule we are testing.
```suggestion
module Test.Cardano.Ledger.Alonzo.Imp.UtxosSpec (spec) where
```

eras/alonzo/impl/testlib/Test/Cardano/Ledger/Alonzo/ImpTest.hs
```diff
@@ -1,43 +1,238 @@
...
+  forall era l.
+  ( PlutusLanguage l
+  , AlonzoEraScript era
+  ) =>
+  Plutus l ->
+  ImpTestM era (Script era)
+plutusToScript p =
+  case mkPlutusScript @era p of
+    Just x -> pure $ fromPlutusScript x
+    Nothing -> fail "Plutus version not supported by era"
```
lehins: It is better to use `error` instead of `fail`, because that give us the callstack. Also make sure to add `HasCallStack` 
```suggestion
    Nothing -> error "Plutus version not supported by era"
```

eras/shelley/impl/testlib/Test/Cardano/Ledger/Shelley/ImpTest.hs
```diff
@@ -582,101 +644,190 @@ impAddNativeScript ::
...
+  let
+    rootTxOut = last . toList $ tx ^. bodyTxL . outputsTxBodyL
+    modifyRootTxOut _ Empty =
+      error $
+        "Expected there to be at least the root output, make sure that "
+          <> "addEmptyRootTxOut comes before updateBalance in the fixup pipeline"
+    modifyRootTxOut v xs =
+      SSeq.dropLast 1 xs
+        :|> (fromJust (SSeq.lookup (length xs - 1) xs) & coinTxOutL .~ v)
+    leftOver = modifyCoin (\(Coin c) -> Coin $ -c) $ coin balance
```
lehins: Coin should never be negative!

eras/shelley/impl/testlib/Test/Cardano/Ledger/Shelley/ImpTest.hs
```diff
@@ -417,20 +466,31 @@ initShelleyImpNES rootCoin =
...
-        Testnet
-        (KeyHashObj $ hashKey vk)
-        (StakeRefBase (KeyHashObj testKeyHash))
     utxo =
-      UTxO $
-        Map.fromList
-          [
+      UTxO
+        . Map.fromList
+        $ [
```
lehins: Sorry, but what's the point of this change like this?

`foo . bar $ []` has more overhead than `foo $ bar []`

eras/shelley/impl/testlib/Test/Cardano/Ledger/Shelley/ImpTest.hs
```diff
@@ -202,12 +217,27 @@ import UnliftIO.Exception (
...
+    PlutusLanguage l =>
+    { astcScript :: Plutus l
+    , astcArgs :: PlutusArgs
+    } ->
+    ScriptTestContext
+
 data ImpTestState era = ImpTestState
   { impNES :: !(NewEpochState era)
   , impRootTxIn :: !(TxIn (EraCrypto era))
+  , impCollateralTxIds :: ![TxIn (EraCrypto era)]
```
lehins, line 223: I don't think it is a good idea to inject special collateral as a separate set into the imp state. It would be better to have those pre-created on per transaction basis. That being said, we can implement this functionality later, so feel free to continue going the route you've established for this PR.

eras/shelley/impl/testlib/Test/Cardano/Ledger/Shelley/ImpTest.hs
```diff
@@ -352,6 +391,17 @@ freshKeyHashVRF = do
   idx <- dispenseIndex
   pure $ mkHashVerKeyVRF @era $ fromIntegral idx
 
+testKeyHash :: Crypto c => KeyHash kd c
+testKeyHash = mkKeyHash (-1)
+
+impAddr :: Crypto c => Int -> Addr c
+impAddr idx =
+  let KeyPair vk _ = mkKeyPair idx
```
lehins, line 390: There is `mkAddr` that can help with this

eras/alonzo/impl/testlib/Test/Cardano/Ledger/Alonzo/ImpTest.hs
```diff
@@ -1,43 +1,240 @@
...
+  pure $
+    tx
+      & witsTxL . scriptTxWitsL %~ Map.union (Map.fromList scriptWits)
+      & witsTxL . rdmrsTxWitsL
+        %~ ( \(Redeemers rdmrs) ->
+              Redeemers $
+                Map.union (Map.fromList newRedeemers) rdmrs
+           )
+      & witsTxL . datsTxWitsL <>~ TxDats (Map.fromList $ catMaybes txDats)
+      & witsTxL . datsTxWitsL <>~ TxDats (Map.fromList $ catMaybes inputDats)
```
lehins:
```suggestion
      & witsTxL . datsTxWitsL <>~ TxDats (Map.fromList $ catMaybes (txDats ++ inputDats))
```

eras/alonzo/impl/testlib/Test/Cardano/Ledger/Alonzo/ImpTest.hs
```diff
@@ -1,43 +1,240 @@
...
+        case txOut ^. datumTxOutF of
+          Datum bin -> do
+            logEntry $ "Datum in inputs: " <> show bin
+            let dat = binaryDataToData bin
+            pure $ Just (hashData dat, dat)
+          _ -> pure Nothing
+      Nothing -> fail "TxIn not found in the UTxO"
+  pure $
+    tx
+      & witsTxL . scriptTxWitsL %~ Map.union (Map.fromList scriptWits)
```
lehins:
```suggestion
      & witsTxL . scriptTxWitsL <>~ Map.fromList scriptWits
```

eras/alonzo/impl/testlib/Test/Cardano/Ledger/Alonzo/ImpTest.hs
```diff
@@ -1,43 +1,240 @@
...
+      & witsTxL . rdmrsTxWitsL
+        %~ ( \(Redeemers rdmrs) ->
+              Redeemers $
+                Map.union (Map.fromList newRedeemers) rdmrs
+           )
```
lehins: I think it'll be cleaner if we add a Monoid instance for `Redeemers`
```suggestion
      & witsTxL . rdmrsTxWitsL <>~  Redeemers (Map.fromList newRedeemers)
```

lehins (commented):
I know that the PR is still a draft, that's why I haven't read through it in detail, just skimmed through it real quick out of curiosity of where it is heading. Here is some of my feedback 
:rocket:

eras/shelley/impl/testlib/Test/Cardano/Ledger/Shelley/ImpTest.hs
Soupstraw: I probably put something there and forgot to change it back, haven't gotten to cleaning up this PR yet, need to get the tests passing first.

lehins: Got it. I was really hoping you gonna say that :wink:

Soupstraw: Oh, but it's actually necessary to do this here because of the `<>` now that I'm also including the collateral UTxOs in the initial state

lehins: Ah, yes. The diff was confusing.

In that case, I'd suggest switching it to `:`, since the first element is just a singleton list

eras/shelley/impl/testlib/Test/Cardano/Ledger/Shelley/ImpTest.hs
Soupstraw: `balance` usually has a negative value here, so I'm making it positive.

Soupstraw: If I'm doing everything right then no negative coin should be escaping outside of this mess of a function.

lehins: It doesn't matter. If you believe that `consumedValue <-> producedValue` can become negative, you should be using `<->` there

In other words you should never rely on `Coin` to be negative, even for internal computation.

Soupstraw: I restored `fixupFees` so this is no longer relevant

eras/shelley/impl/testlib/Test/Cardano/Ledger/Shelley/ImpTest.hs
Soupstraw: `mkAddr` requires a staking keypair, but I don't need to stake with the addresses I'm generating

eras/alonzo/impl/test/Main.hs
```diff
@@ -25,3 +26,4 @@ main =
         CostModelsSpec.spec @Alonzo
       describe "TxWits" $ do
         TxWitsSpec.spec @Alonzo
+      UtxosSpec.spec @Alonzo
```
lehins: See Shelley and Conway for structure, we need to be consistent

This should be:
```suggestion
      Imp.spec @Alonzo
```

Where `Imp.spec` aggregates spec of all rules from Alonzo

eras/alonzo/impl/testlib/Test/Cardano/Ledger/Alonzo/Imp/UtxosSpec.hs
```diff
@@ -0,0 +1,66 @@
...
+import Test.Cardano.Ledger.Plutus.Examples (guessTheNumber3)
+
+spec ::
+  forall era.
+  ( ShelleyEraImp era
+  , AlonzoEraTxWits era
+  , AlonzoEraTxOut era
+  ) =>
+  SpecWith ()
+spec = describe "ScriptSpec" . withImpState @era $ do
```
lehins: Here too. See how it is done in Conway and Shelley. The goal is to call `withImpState` once per era, not per rule.

This will require addition of a new module `Test.Cardano.Ledger.Alonzo.Imp`

eras/alonzo/impl/testlib/Test/Cardano/Ledger/Alonzo/Imp/UtxosSpec.hs
```diff
@@ -0,0 +1,66 @@
...
+            .~ SSeq.singleton
+              ( mkBasicTxOut
+                  (Addr Testnet (ScriptHashObj shSpending) StakeRefNull)
+                  (inject (Coin 100))
+                  & dataHashTxOutL .~ SJust (hashData @era $ Data spendDatum)
+              )
+    submitTxAnn_ "Submit a transaction that consumes the script output" $
+      mkBasicTx mkBasicTxBody
+        & bodyTxL . inputsTxBodyL
+          .~ Set.singleton (mkTxInPartial (txIdTx tx0) 1)
```
lehins: Why is the index `1`? I thought we made the root input be appended, not prepended

Teodora added a nice helper function that can be useful here
```suggestion
          .~ Set.singleton (txInAt (1 :: Int) tx0)
```

Lucsanszky (commented):
I had a look. LGTM but I will let others to approve as I am not that familiar with the codebase yet.

eras/shelley/impl/testlib/Test/Cardano/Ledger/Shelley/ImpTest.hs
```diff
@@ -628,57 +672,104 @@ impAddNativeScript ::
...
-addAddrTxWits tx nativeScriptsKeyPairs = do
+updateAddrTxWits tx = do
   let txBody = tx ^. bodyTxL
       txBodyHash = hashAnnotated txBody
-  witsVKeyNeeded <- impWitsVKeyNeeded txBody
-  keyPairs <- mapM lookupKeyPair $ Set.toList witsVKeyNeeded
+  scriptsRequired <- impScriptsRequired tx
+  let nativeScripts = Map.elems scriptsRequired
+  nativeScriptsKeyPairs <- mapM impSatisfyNativeScript nativeScripts
+  witsVKeyNeeded <- impWitsVKeyNeeded (tx ^. bodyTxL)
```
```suggestion
  witsVKeyNeeded <- impWitsVKeyNeeded txBody
```

eras/shelley/impl/testlib/Test/Cardano/Ledger/Shelley/ImpTest.hs
```diff
@@ -628,57 +672,104 @@ impAddNativeScript ::
...
+
+impScriptsRequired ::
+  EraUTxO era =>
   Tx era ->
-  ImpTestM era (Tx era, Map (KeyHash 'Witness (EraCrypto era)) (KeyPair 'Witness (EraCrypto era)))
-addScriptTxWits tx = do
+  ImpTestM
+    era
+    (Map (ScriptHash (EraCrypto era)) (NativeScript era))
+impScriptsRequired tx = do
```
lehins: Maybe just for added clarity
```suggestion
impNativeScriptsRequired tx = do
```

eras/shelley/impl/testlib/Test/Cardano/Ledger/Shelley/ImpTest.hs
```diff
@@ -628,57 +672,104 @@ impAddNativeScript ::
...
   utxo <- getUTxO
-  ImpTestState {impScripts} <- get
+  ImpTestState {impNativeScripts} <- get
   -- TODO: Add proper support for Plutus Scripts
```
lehins: Done
```suggestion
```

eras/shelley/impl/testlib/Test/Cardano/Ledger/Shelley/ImpTest.hs
```diff
@@ -628,57 +672,104 @@ impAddNativeScript ::
...
+    era
+    (Map (KeyHash 'Witness (EraCrypto era)) (KeyPair 'Witness (EraCrypto era)))
+impNativeScriptKeyPairs tx = do
+  utxo <- getUTxO
+  ImpTestState {impNativeScripts} <- get
+  -- TODO: Add proper support for Plutus Scripts
+  let needed = getScriptsNeeded utxo (tx ^. bodyTxL)
+      hashesNeeded = getScriptsHashesNeeded needed
+      scriptsRequired = impNativeScripts `Map.restrictKeys` hashesNeeded
+      nativeScripts = Map.elems scriptsRequired
```
lehins: Reduce duplication:
```suggestion
  scriptsRequired <- impScriptsRequired
  let nativeScripts = Map.elems scriptsRequired
```

eras/shelley/impl/testlib/Test/Cardano/Ledger/Shelley/ImpTest.hs
```diff
@@ -707,22 +798,30 @@ fixupFees tx nativeScriptKeyWits = do
...
   ImpTestM era (Tx era)
-fixupTx txRaw = do
-  (txWithScriptWits, nativeScriptKeyPairs) <- addScriptTxWits txRaw
-  txWithFee <- fixupFees txWithScriptWits (Map.keysSet nativeScriptKeyPairs)
-  txFixed <- addAddrTxWits txWithFee (Map.elems nativeScriptKeyPairs)
+shelleyFixupTx =
+  addNativeScriptTxWits
+    >=> addRootTxIn
+    -- Add the TxWits so we can calculate the fee correctly
+    >=> updateAddrTxWits
```
lehins: This is step is redundant, fee calculation function `calcMinFeeTxNativeScriptWits` will already do that.
```suggestion
```

eras/alonzo/impl/testlib/Test/Cardano/Ledger/Alonzo/ImpTest.hs
```diff
@@ -1,43 +1,178 @@
...
+  , ScriptsNeeded era ~ AlonzoScriptsNeeded era
+  ) =>
+  Tx era ->
+  ImpTestM era (Tx era)
+alonzoFixupTx =
+  addNativeScriptTxWits
+    >=> addCollateralInput
+    >=> addRootTxIn
+    >=> fixupPlutusScripts
+    >=> updateAddrTxWits
```
lehins: Same, redundant
```suggestion
```

lehins (approved):
Beautiful! 
I only had some minor comments.

eras/alonzo/impl/testlib/Test/Cardano/Ledger/Alonzo/Imp/UtxosSpec.hs
Soupstraw: Good catch, I think it was consuming the rootTxOut from previous transaction :sweat_smile:.

PR #4050, opened 2024-02-05
Fixed the 'All Tx are valid on traces of length 150'  intermittent bug

### LINKED ISSUE #4057: Test failure on CI in the new `constraint-generator` package
https://github.com/IntersectMBO/cardano-ledger/actions/runs/7805271233/job/21289095718?pr=4050#step:15:4624

```
      Int8
        prop_complete: FAIL (0.02s)
          *** Failed! Falsified (after 52 tests):
          
          genFromSpec
              [Int8]
              TypeSpec (ListSpec [] TrueSpec (TypeSpec [..4] []) (FoldSpec Id (TypeSpec [11..19] []))) []
          
          Can't generate list of ints with fold constraint
            elemSpec = TypeSpec [..4] []
            normElemSpec = TypeSpec [-128..4] []
            foldSpec = TypeSpec [11..19] []
          
          Generating an element:
            elemS = TypeSpec [4..4] []
            foldS = TypeSpec [11..11] []
            fuel  = 12
            lst   = [4,4,4,4,4,4,4,4,4,4,4,4,4,4,4,4,4,4,4,4,4,4,4,4,4,4,4,4,4,-44,-5,-29,4,-8,0,-25,-10,1]
          
          Fatal error: 
          Ran out of tries on suchThatT
          
          Use --quickcheck-replay=145764 to reproduce.
          Use -p '/listSumRangeUpper.Int8.prop_complete/' to rerun this test only.
```

TimSheard:
Fixed the 'All Tx are valid on traces of length 150' bug.
This was occasionally triggered at the Epoch boundary when a ParameterChange Proposal was passed, 
and the PParamUpdate made random (stupid) changes to fields that made the Tx in the trace no longer valid.
Prime candidates were PParam fields that affected the computation of the fee, legal Tx sizes, and min txout values, etc. 
Added function 'reasonable' to ensure bad PParamUpdates are not used. Allows users to adjust PParamUpdates in Proposals
Added EraClass -- Everything you need to build Era polymorphic Tx, all in one place

I ran this test 1000's of times without a failure while I ate dinner. I hope we have fixed it for good.
Fixes #4057 addressing the  '/listSumRangeUpper.Int8.prop_sound/' bug, because it made the tests fail.

# Description

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/libs/cardano-ledger-test/src/Test/Cardano/Ledger/EraClass.hs
+++ b/libs/cardano-ledger-test/src/Test/Cardano/Ledger/EraClass.hs
@@ -1,0 +1,156 @@
+-- | Everything you need to build Era polymorphic transactions, all in one place
+module Test.Cardano.Ledger.EraClass (
+  Era (EraCrypto),
+  -- Tx
+  EraTx (Tx, mkBasicTx, bodyTxL, witsTxL),
+  AlonzoEraTx (isValidTxL),
+  -- TxBody
+  EraTxBody (TxBody, mkBasicTxBody, inputsTxBodyL, outputsTxBodyL, feeTxBodyL, withdrawalsTxBodyL, auxDataHashTxBodyL, certsTxBodyL),
+  AllegraEraTxBody (vldtTxBodyL),
+  MaryEraTxBody (mintTxBodyL, mintValueTxBodyF, mintedTxBodyF),
+  AlonzoEraTxBody (
+    collateralInputsTxBodyL,
+    reqSignerHashesTxBodyL,
+    scriptIntegrityHashTxBodyL,
+    networkIdTxBodyL,
+    redeemerPointer,
+    redeemerPointerInverse
+  ),
+  ConwayEraTxBody (currentTreasuryValueTxBodyL, votingProceduresTxBodyL, proposalProceduresTxBodyL, treasuryDonationTxBodyL),
+  -- TxOut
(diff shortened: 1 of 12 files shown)


libs/cardano-ledger-test/src/Test/Cardano/Ledger/EraClass.hs
```diff
@@ -0,0 +1,155 @@
+-- | Everything you need to build Era polymorphic transactions, all in one place
+module Test.Cardano.Ledger.EraClass (
```
lehins, line 2: Is this module really necessary?

All you need to do is import `Cardano.Ledger.Conway.Core` in order to get almost all of it.

Seems like unnecessary work that will get out of date pretty quickly

lehins (approved):
Looks good. Thank you for fixing the test. 

FYI, there is a conflict

libs/cardano-ledger-test/src/Test/Cardano/Ledger/EraClass.hs
TimSheard: Actually, It can't get out of date, because then it won't compile. Perhaps it might be best in the API. It is hard to remember all the names, and where they come from. Only about 60% is in Cardano.Ledger.Core . It is all the exceptions (AonzoEraTx, BabbageEraPParams ...) that are now nicely documented in one place.

lehins: It's up to you. If it makes it easy for you, feel free to keep it. I am definitely not updating that module.

PR #4053, opened 2024-02-06
Rename committeeQuorum to committeeThreshold

### LINKED ISSUE #4018: Conway mislabeling of Constitutional Committee Parameter
Tested on current Sancho Testnet 

$ cardano-node --version
cardano-node 8.7.3 - linux-x86_64 - ghc-8.10
git rev a4a8119b59b1fbb9a69c79e1e6900e91292161e7

Following Mike Hornan, Nicolas Cerny, and I's recent running of a series of Test Suites to determine what Ratification outcomes would occur across various combinations of Yes/No Votes from the Constitutional Committee's voting on a Proposal. It was discovered that the, so named, 'Quorum' Parameter of the Constitutional Committee is in fact acting with the behaviour of a Threshold Parameter.

It is therefore **highly** suggested that the official On-Chain naming of the CC Quorum Parameter be changed to reflect its core function. Which is that of a Threshold. 

More information about the testing patterns that were run on Sancho Testnet are available at Mike Hornan's Governance GitHub Repository: https://github.com/Hornan7/Testnet_Scenarios 

NB: The definition of Quorum is as follows -

"The minimum number of members of an assembly or society that must be present at any of its meetings to make the proceedings of that meeting valid."

It is most certainly **not** supposed to be the actual **target** of the required amount of Yes Votes for something to Pass.

It is **only** supposed to be the amount of people required to actually show up to vote on something before that vote can be considered valid. 

As it stands,  the Quorum Parameter is currently set up to be an actual **target** for the **amount of Yes Votes** required to deem an Action to be Constitutional.

i.e. It may not be working as intended. 

If it is...

Then that Parameter needs to be renamed to be **Threshold**. To avoid any user confusion being created such as the confusion Mike and I had during our predictions of the Voting Outcomes in the Testing Repository linked to above. 

It also needs to be made clear in any documentation that the now renamed CC Member 'Threshold' Parameter setting takes precedence over **all** other Governance Action Threshold Parameter Settings, in the DRep sphere, when it comes to the Constitutional Committee's Voting Process as to the Constitutionality of a Proposal. As was shown during our testing. 

I hope this request is acceptable. 

Thank you kindly 🙏🏻

aniketd:
# Description

Resolves #4018 

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/eras/conway/impl/CHANGELOG.md
+++ b/eras/conway/impl/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## 1.13.0.0
 
+* Rename `committeeQuorum` to `committeeThreshold` #4053
 * Changed `GovActionState` to have 1 field (`gasProposalProcedure`) rather than 3 (`gasDeposit`, `gasAction`, `gasReturnAddr`)
   * the old field names (`gasDeposit`, `gasAction`, `gasReturnAddr`) become functions, and the lenses
   * (`gasDepositL`, `gasActionL`, `gasReturnAddrL`) have the same type, but behave differently.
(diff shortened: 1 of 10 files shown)


lehins (approved):
Perfect!

PR #4055, opened 2024-02-06
Update MinFeeRefScriptCoinsPerByte.

### LINKED ISSUE #4052: Change the type of `minFeeRefScriptCoinsPerByte`
In a recent PR #3983 a new protocol parameter has been added. One oversigt was that the type that was used does not allow for a more fine grain tuning of the fee, because it is backed by an `CoinPerByte`, which is essentially a `Coin`. In other words it only allows one lovelace per byte granularity, which might not be enough in the future. 

In order to allow us to be as precise as necessary, we need to:

* [x] - We need to change the type to `NonNegativeInterval`
* [x] - Change the formula to use rounding
* [x] - Rename the parameter to `minFeeRefScriptCostPerByte`
* [x] - Adjust CDDL

Here is the implementation of the fee calculation:

https://github.com/IntersectMBO/cardano-ledger/blob/e73acc68d1ee98e59849288a9825b8fe1fb83324/eras/conway/impl/src/Cardano/Ledger/Conway/Tx.hs#L85-L87

Which should be changed to this new calculation:

```haskell
 alonzoMinFeeTx pp tx <+> refScriptsFee
 where 
   refScriptFee = unboundRational (pp ^. ppMinFeeRefScriptCoinsPerByteL)
   refScriptsFee = Coin (floor (fromIntegral @Int @Rational refScriptsSize * refScriptFee)) 
```

aniketd:
* Rename it to MinFeeRefScriptCostPerByte.
* Update its type from Coin to NonNegativeInterval.

# Description

Fixes #4052 

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/eras/conway/impl/src/Cardano/Ledger/Conway/Tx.hs
+++ b/eras/conway/impl/src/Cardano/Ledger/Conway/Tx.hs
@@ -24,14 +24,14 @@ import Cardano.Ledger.Alonzo.TxSeq (
   AlonzoTxSeq (AlonzoTxSeq, txSeqTxns),
   hashAlonzoTxSeq,
  )
-import Cardano.Ledger.Babbage.PParams (CoinPerByte (..))
 import Cardano.Ledger.Babbage.Tx as BabbageTxReExport (
   AlonzoEraTx (..),
   AlonzoTx (..),
  )
-import Cardano.Ledger.Coin (Coin)
+import Cardano.Ledger.BaseTypes (unboundRational)
+import Cardano.Ledger.Coin (Coin (Coin))
 import Cardano.Ledger.Conway.Era (ConwayEra)
-import Cardano.Ledger.Conway.PParams (ConwayEraPParams, ppMinFeeRefScriptCoinsPerByteL)
+import Cardano.Ledger.Conway.PParams (ConwayEraPParams, ppMinFeeRefScriptCostPerByteL)
 import Cardano.Ledger.Conway.TxAuxData ()
 import Cardano.Ledger.Conway.TxBody ()
 import Cardano.Ledger.Conway.TxWits ()
@@ -82,9 +82,10 @@ getConwayMinFeeTx ::
   Int ->
   Coin
 getConwayMinFeeTx pp tx refScriptsSize =
-  alonzoMinFeeTx pp tx <+> (refScriptsSize <×> refScriptFee)
+  alonzoMinFeeTx pp tx <+> refScriptsFee
   where
-    CoinPerByte refScriptFee = pp ^. ppMinFeeRefScriptCoinsPerByteL
+    refScriptCostPerByte = unboundRational (pp ^. ppMinFeeRefScriptCostPerByteL)
+    refScriptsFee = Coin (floor (fromIntegral @Int @Rational refScriptsSize * refScriptCostPerByte))
 
 instance Crypto c => AlonzoEraTx (ConwayEra c) where
   {-# SPECIALIZE instance AlonzoEraTx (ConwayEra StandardCrypto) #-}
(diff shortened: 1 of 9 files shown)


eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/UtxoSpec.hs
```diff
@@ -148,8 +149,13 @@ spec = describe "UTxO" $ do
       TimelockScript tl -> SBS.length $ getMemoRawBytes tl
       PlutusScript ps -> withPlutusScript ps (SBS.length . unPlutusBinary . plutusBinary)
 
-    setRefScriptFee :: ImpTestM era Coin
+    setRefScriptFee :: ImpTestM era NonNegativeInterval
     setRefScriptFee = do
-      let refScriptFee = Coin 10
-      modifyPParams $ ppMinFeeRefScriptCoinsPerByteL .~ CoinPerByte refScriptFee
+      let refScriptFee = boundRational' 10
```
lehins: There is a handy operator that use can use in tests:
```suggestion
      let refScriptFee = 10 %! 1
```

eras/conway/impl/src/Cardano/Ledger/Conway/Tx.hs
```diff
@@ -82,9 +82,10 @@ getConwayMinFeeTx ::
   Int ->
   Coin
 getConwayMinFeeTx pp tx refScriptsSize =
-  alonzoMinFeeTx pp tx <+> (refScriptsSize <×> refScriptFee)
+  alonzoMinFeeTx pp tx <+> refScriptsFee
   where
-    CoinPerByte refScriptFee = pp ^. ppMinFeeRefScriptCoinsPerByteL
+    refScriptFee = unboundRational (pp ^. ppMinFeeRefScriptCostPerByteL)
+    refScriptsFee = Coin (floor (fromIntegral @Int @Rational refScriptsSize * refScriptFee))
```
lehins: Let's change the binding name to reflect the meaning a little better
```suggestion
    refScriptCostPerByte = unboundRational (pp ^. ppMinFeeRefScriptCostPerByteL)
    refScriptsFee = 
      Coin (floor (fromIntegral @Int @Rational refScriptsSize * refScriptCostPerByte))
```

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/UtxoSpec.hs
```diff
@@ -81,7 +81,9 @@ spec = describe "UTxO" $ do
       minFeeDiff <- conwayDiffMinFee tx
       -- check that the difference is the sum of the sizes of the passed reference scripts
       minFeeDiff
-        `shouldBe` Foldable.foldMap (\s -> scriptSize s <×> refScriptFee) refScriptInToScripts
+        `shouldBe` Foldable.foldMap
+          (\s -> Coin $ floor $ (fromIntegral @Int @Rational $ scriptSize s) * unboundRational refScriptFee)
+          refScriptInToScripts
```
lehins: That formula is incorrect. You first need to sum up all the sizes and only then multiply the result bu the fee, because if you do multiplication and rounding on per script basis, then you can accumulate loss with each script, which will make this invalid.

It worked before, since there was no rounding.

lehins (changes requested):
There is one small issue in the unit test. Which works right now, but if we ever turned it into a property with some randomness in it, the test would break.

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/UtxoSpec.hs
aniketd: Ah! :bow: Thank you!

eras/conway/impl/src/Cardano/Ledger/Conway/Tx.hs
aniketd: Understood :pray:

lehins (approved):
~Needs a rebase, there is some conflict right now. Other than that~ it is ready to go

PR #4056, opened 2024-02-06
Reduce duplication in Conway UTXOW rule

### LINKED ISSUE #3916: Reduce duplication in the `ConwayUTXOW` rule
We also need to reduce duplication in the ConwayUTXOW rule, almost the whole thing is from BabbageUTXOW (related ticket #3485)

~One of the things in ConwayUTXOW rule that is missing that is in BabbageUTXOW is this predicate check. Why is it missing?~

https://github.com/input-output-hk/cardano-ledger/blob/5922372daccb4878c4c088c582b0d56f3a671ae8/eras/babbage/impl/src/Cardano/Ledger/Babbage/Rules/Utxow.hs#L324-L327

As mentioned in https://github.com/IntersectMBO/cardano-ledger/issues/3916#issuecomment-1875187081 it has been moved to a new place

lehins:
# Description

Fixes #3916 

The only difference between Conway's and Babbage's UTXOW rule implementation is that latter has MIR certificate witness validation.

This PR extract MIR validation into a separate action and uses it only in Babbage, while the rest of the logic is reused in Conway.

This PR also fixes a small issue in Alonzo spec, where `costmdls` was applied to `tx` instead of `pp`, which was noted in the rule.

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/eras/babbage/impl/src/Cardano/Ledger/Babbage/Rules/Utxow.hs
+++ b/eras/babbage/impl/src/Cardano/Ledger/Babbage/Rules/Utxow.hs
@@ -248,14 +248,32 @@ validateScriptsWellFormed pp tx =
 -- The tests are very generic and reusable, but the transition
 -- function is very specific to the Babbage Era.
 
--- | A very specialized transitionRule function for the Babbage Era.
+
+babbageUtxowMirTransition ::
+  forall era.
+  ( AlonzoEraTx era
+  , ShelleyEraTxBody era
+  , STS (BabbageUTXOW era)
+  ) =>
+  TransitionRule (BabbageUTXOW era)
+babbageUtxowMirTransition = do
+  TRC (UtxoEnv _ _ certState, st, tx) <- judgmentContext
+  -- check genesis keys signatures for instantaneous rewards certificates
+  {-  genSig := { hashKey gkey | gkey ∈ dom(genDelegs)} ∩ witsKeyHashes  -}
+  {-  { c ∈ txcerts txb ∩ TxCert_mir} ≠ ∅  ⇒ |genSig| ≥ Quorum  -}
+  let genDelegs = dsGenDelegs (certDState certState)
+      witsKeyHashes = witsFromTxWitnesses tx
+  coreNodeQuorum <- liftSTS $ asks quorum
+  runTest $
+    Shelley.validateMIRInsufficientGenesisSigs genDelegs coreNodeQuorum witsKeyHashes tx
+  pure st
+
+-- | UTXOW transition rule that is used in Babbage and Conway era.
 babbageUtxowTransition ::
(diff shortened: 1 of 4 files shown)


eras/babbage/impl/src/Cardano/Ledger/Babbage/Rules/Utxow.hs
```diff
@@ -364,7 +379,7 @@ instance
   type BaseM (BabbageUTXOW era) = ShelleyBase
   type PredicateFailure (BabbageUTXOW era) = BabbageUtxowPredFailure era
   type Event (BabbageUTXOW era) = AlonzoUtxowEvent era
-  transitionRules = [babbageUtxowTransition]
+  transitionRules = [babbageUtxowMirTransition >> babbageUtxowTransition @era]
```
TimSheard, line 382: Not sure if this matters, but by defining it this way, the order that the Predicate tests run now, is not the same as the order they used to be run.  If it doesn't matter, we should say so, and explain why.

TimSheard (approved):
Looks good, but it breaks from a pattern that we have used before. of defining a function like

validateXXX::
  EraTx era =>
  Tx era -> .. -> Test (SomePredFailure era)

and then calling

runTest $ validateXXX

Is there a reason we did not do this here?

eras/babbage/impl/src/Cardano/Ledger/Babbage/Rules/Utxow.hs
lehins: The important part is that predicate is checked. The order in which they are checked only controls the order in which the user sees the failure, so that is not that important. This is a general rule, which doesn't require a special documentation in this case.

In this case especially it doesn't matter, since MIR certificates can only be submitted by IOG.

lehins:
> Is there a reason we did not do this here?

Yes, because this approach is not going to work in this scenario. We need to abstract a collection of rules and reuse them as a whole. 

In other words, the approach of using validations (i.e. the `runTest`) is using them like building blocks, where we can use a whole bunch of similar blocks to construct different walls.

In this case however we want to reuse the whole wall, without one block. Which means we need to split up a rule into a slightly smaller rule, that we can later reuse verbatim in the next era.

eras/babbage/impl/src/Cardano/Ledger/Babbage/Rules/Utxow.hs
lehins: Forgot to mention, although I think it is obvious, but the order matters for any state modification. So, if `babbageUtxowMirTransition` did modify the state in a way that mattered for the order of computation, then clearly such such re-ordering would not be possible. But, this is not the case since it only contains a predicate check and does not affect the state in any way

PR #4058, opened 2024-02-07
Fix `ToJSON` instance of `BoundedRatio` to avoid precision loss

### LINKED ISSUE #4026: Fix `ToJSON` instance for `BoundedRatio` to not loose precision
Current problem is that rational numbers like `2/3` are encoded into json with loss of precision like so: `0.6666666666666666666`. So, natural fix would be to encode `BoundedRatio` as:
```json
{ "numerator": 2
, "denominator": 3
}
```
However, we can't do this blindly due to backwards compatibility. The best course of action is to support both formats: decimal and rational. Especially considering that `FromJSON` already supports that.

So, the fix should involve detecting loss of precision with the help of [`fromRationalRepetendLimited`](https://hackage.haskell.org/package/scientific-0.3.7.0/docs/Data-Scientific.html#v:fromRationalRepetendLimited)
and falling back to `ToJSON` instance for `Rational`, when encoding in decimal form is not possible without loss of precision.

Property tests for this change should go into: `Test.Cardano.Ledger.BaseTypesSpec`

Lucsanszky:
# Description
Resolves #4026 

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/libs/cardano-ledger-core/src/Cardano/Ledger/BaseTypes.hs
+++ b/libs/cardano-ledger-core/src/Cardano/Ledger/BaseTypes.hs
@@ -306,12 +306,12 @@ fromRationalBoundedRatio r
 -- when in doubt.
 fromRatioBoundedRatio ::
   forall b a.
-  (Bounded (BoundedRatio b a), Integral a) =>
+  (Bounded a, Bounded (BoundedRatio b a), Integral a) =>
   Ratio a ->
   Maybe (BoundedRatio b a)
 fromRatioBoundedRatio ratio
-  | r < toRationalBoundedRatio lowerBound
-      || r > toRationalBoundedRatio upperBound =
+  | r < unboundRational lowerBound
+      || r > unboundRational upperBound =
       Nothing -- ensure valid range
   | otherwise = Just $ BoundedRatio ratio
   where
@@ -328,7 +328,7 @@ instance
   where
   decCBOR = do
     r <- decodeRationalWithTag
-    case fromRationalBoundedRatio r of
+    case boundRational r of
       Nothing ->
         cborError $ DecoderErrorCustom "BoundedRatio" (Text.pack $ show r)
       Just u -> pure u
--- a/libs/cardano-ledger-core/src/Cardano/Ledger/BaseTypes.hs
+++ b/libs/cardano-ledger-core/src/Cardano/Ledger/BaseTypes.hs
@@ -138,7 +138,7 @@ import Data.Aeson (
   FromJSON (..),
   KeyValue,
   ToJSON (..),
-  Value,
+  Value (..),
   object,
   pairs,
   withObject,
@@ -161,8 +161,8 @@ import Data.Scientific (
   Scientific,
   base10Exponent,
   coefficient,
+  fromRationalRepetendLimited,
   normalize,
-  scientific,
  )
 import Data.Text (Text)
 import qualified Data.Text as Text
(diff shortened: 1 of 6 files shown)


libs/cardano-ledger-core/test/Test/Cardano/Ledger/BaseTypesSpec.hs
```diff
@@ -56,29 +55,16 @@ boundedRationalTests badJSONValues = do
...
-      prop "ToJSON/FromJSON roundtrip up to an epsilon" $
+      prop "ToJSON/FromJSON roundtrip" $
         forAll genValid $ \(br :: a) ->
           within 500000 $
-            case eitherDecode (encode br) of
+            case boundedFromJSON (encode br) of
               Left err -> error err
-              Right (br' :: a) ->
-                epsilonMaybeEq 1e-18 (unboundRational br) (unboundRational br')
-      prop "Roundtrip to Scientific and back up to an epsilon" $
```
Lucsanszky, line 66: I removed these tests since I believe my changes make them obsolete.

libs/cardano-ledger-core/test/Test/Cardano/Ledger/BaseTypesSpec.hs
```diff
@@ -17,7 +17,6 @@ import Data.Typeable
 import Test.Cardano.Ledger.Binary.RoundTrip (roundTripCborExpectation)
 import Test.Cardano.Ledger.Common
 import Test.Cardano.Ledger.Core.Arbitrary ()
-import Test.Cardano.Ledger.Core.Utils (epsilonMaybeEq)
```
Lucsanszky, line 20: `epsilonMaybeEq` was only used here, so we might want to remove the function itself as well. Unless we think that it might be useful again at some point.

libs/cardano-ledger-core/src/Cardano/Ledger/BaseTypes.hs
```diff
@@ -328,53 +335,42 @@ instance
...
 fromScientificBoundedRatioWord64 (normalize -> sci)
-  | coeff < 0 = failWith "negative"
+  | coeff < 0 = failWith "negative" sci
   | exp10 <= 0 = do
-      when (exp10 < -19) $ failWith "too precise"
+      when (exp10 < -19) $ failWith "too precise" sci
       fromRationalEither (coeff % (10 ^ negate exp10))
   | otherwise = do
-      when (19 < exp10) $ failWith "too big"
+      when (19 < exp10) $ failWith "too big" sci
```
Lucsanszky: `19` is the number of decimal digits for 64 bit values. I think it might be a good idea to consider removing this magic number by introducing a constant. Something along the lines of `numberOfDecimals64 = 19`. Would that be desirable? If so, I can do it here and now or later as part of some refactoring.

Lucsanszky (commented):
This is now ready for review. Added some comments to consider.

libs/cardano-ledger-core/src/Cardano/Ledger/BaseTypes.hs
```diff
@@ -244,6 +245,12 @@ instance Integral a => Ord (BoundedRatio b a) where
 promoteRatio :: Integral a => Ratio a -> Rational
 promoteRatio r = toInteger (numerator r) % toInteger (denominator r)
 
+toRationalJSON :: Rational -> Value
+toRationalJSON r =
+  case fromRationalRepetendLimited 20 r of
```
lehins: I believe we need to use `19` here as well, otherwise there could be issues with roundtripping
```suggestion
  case fromRationalRepetendLimited 19 r of
```

libs/cardano-ledger-core/src/Cardano/Ledger/BaseTypes.hs
lehins: I totally agree with you. Let's create a top level binding with a bit of explanation what this number means. I think doing it in this PR will be good.

libs/cardano-ledger-core/src/Cardano/Ledger/BaseTypes.hs
```diff
@@ -328,53 +335,42 @@ instance
...
-    failWith :: String -> Either String a
-    failWith msg = Left $ "Value is " ++ msg ++ ": " ++ show sci
-    fromRationalEither =
-      maybe (failWith "outside of bounds") Right . fromRationalBoundedRatio
+
+fromRationalEither :: Bounded (BoundedRatio b Word64) => Rational -> Either String (BoundedRatio b Word64)
+fromRationalEither r = maybe (failWith "outside of bounds" r) Right $ boundRational r
+
+failWith :: Show a => String -> a -> Either String b
+failWith msg (show -> val) = Left $ "Value is " <> msg <> ": " <> val
```
lehins: I'd suggest using `ViewPatterns` only in places where they are really useful. Some people get triggered about their usage. (none on our team, from what I know, but it does happen in Haskell community)
```suggestion
failWith msg val = Left $ "Value is " <> msg <> ": " <> show val
```

libs/cardano-ledger-core/src/Cardano/Ledger/Plutus/ExUnits.hs
```diff
@@ -32,6 +32,7 @@ import Cardano.Ledger.BaseTypes (
   BoundedRational (unboundRational),
   NonNegativeInterval,
   boundRational,
+  toRationalJSON,
```
lehins: We no longer need to import it here, because this comment is no longer true:
https://github.com/IntersectMBO/cardano-ledger/blob/21871e9651d8d3fad3383647b4f157f1612ae685/libs/cardano-ledger-core/src/Cardano/Ledger/Plutus/ExUnits.hs#L172

So, we can fix the ToJSON/FromJSON instances for `Prices`, by relying on the underlying instance of `NonNegativeInterval`
```suggestion
```

libs/cardano-ledger-core/test/Test/Cardano/Ledger/BaseTypesSpec.hs
lehins: I think we can remove it. If we ever need it again, we can always bring it back. I have a variant of it in a couple of projects of mine.
I doubt though that we will need it, since it is usually only useful for testing floating point numbers and other computation related to rounding, which is unlikely to be useful in ledger again.

lehins (changes requested):
Very nice!

I left one suggestion that I didn't even think of until reviewing this PR. We can now simplify JSON instances for `Prices`.

Once that's done we can merge the PR

libs/cardano-ledger-core/src/Cardano/Ledger/BaseTypes.hs
```diff
@@ -176,6 +177,9 @@ import NoThunks.Class (NoThunks (..))
 import Numeric.Natural (Natural)
 import Quiet (Quiet (Quiet))
 
+maxScale :: Int
+maxScale = 19
```
lehins: We should not export this biding from the module and it should have a better name, along the line of:
(Edit: tests can define their own binding like that)
```suggestion
numFullDigitsWord64 :: Int
numFullDigitsWord64 = 19
```

Lucsanszky: Renamed it. The "full digits" part sounds strange to me but including `Word64` in the name is good. `maxScale` could indeed be a bit misleading without context.

lehins: `maxBound` for `Word64` has 20 digits:
```haskell
ghci> length (show (maxBound :: Word))
20
```

But with regards to representing decimal numbers it has only 19 digits available:
```
ghci> (2 * 10 ^ 19) > toInteger (maxBound :: Word)
True
ghci> (1 * 10 ^ 19) > toInteger (maxBound :: Word)
False
```
For example, the smallest decimal number you can represent with `Ratio Word64` is this number that has 19 decimal places (despite that numerator has 20 digits :grin:):
```
ghci> 0.0000000000000000001 :: Ratio Word
1 % 10000000000000000000
ghci> 0.00000000000000000001 :: Ratio Word
1 % 7766279631452241920
```

With all this information at your disposal, if you have a better name for the binding, please go for it.

Lucsanszky: Yes, this is what I added to the commit body here: be55cbc14d2fc61970b711d77eb857b70e8c896b . Although not in such detail. :grin:  This is why I picked `maxScale` since you refer to that parameter as scale too in `safe-decimal`: https://hackage.haskell.org/package/safe-decimal-0.2.1.0/docs/Numeric-Decimal.html

How about `maxScaleWord64`? Or `maxDecimalsWord64` since mathematics sometimes refer to the decimal places after the decimal point as decimals? I think this latter one is more well-known.

lehins: `maxDecimalsWord64` sounds ideal!

lehins (approved):
Perfect!

PR #4059, opened 2024-02-08
Enable Plutus v1 reference scripts in Conway

### LINKED ISSUE #3965: Reference scripts for `PlutusV1`
It has been brought to our attention that PlutusV1 scripts occupy around 40% of the whole blockchain since Alonzo. Natural solution to this problem is to enable reference scripts for PlutusV1 in Conway

Here is the relevant CIP: https://github.com/cardano-foundation/CIPs/pull/679

The proposed solution in the above CIP is too complicated for no good reason. We can just enable the reference scripts. However this will require #3952 since we need mitigate the worries that @WhatisRT mentioned where one could use a fee field to infer inside a PlutusV1 script whether the script was provided in a witness set or in a reference input.

In order to confirm the report and the proposed solution of using reference scripts we did some investigation. We wrote a program that replays the chain and counts up the sizes of all of the Plutus scripts in all of the blocks per epoch starting from beginning of Alonzo until Christmas 2023. Here is the plot for the calculated data:

image

* Blue is the total size of the blockchain per Epoch
* Red is the total size of PlutusV1 scripts per epoch that were used as witnesses in transactions
* Yellow (barely visible) is the total size of PlutusV2 scripts per epoch that were used as witnesses in transactions together with PlutusV2 scripts that were placed into TxOuts for future use as Reference Scripts. In other words total contribution of PlutusV2 scripts to on-chain data.
* Green is the total size of PlutusV2 scripts that were actually used as reference scripts. In other words, this would have been the overhead added to the chain, if we didn't have reference scripts capabilities for the PlutusV2 scripts.

It is clear that reference scripts are a great solution to the problem.

This plot is just the zoomed version of only the PlutusV2 sizes:

image

The pie chart below depicts the percentage of the on-chain data occupied by PlutusV1 (red), PlutusV2 (yellow) the rest of the data (blue):

image

It was noted on Github that 60Gb overhead by scripts is not that big when comparing to other blockchains, which is true, but 41% of the total blockchain data is too much.

Soupstraw:
# Description

This PR removes a check from conway that fails a transaction when it's trying to use reference scripts for Plutus V1.

I also added support for reference scripts in ImpTests.

resolves #3965

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/eras/alonzo/impl/testlib/Test/Cardano/Ledger/Alonzo/ImpTest.hs
+++ b/eras/alonzo/impl/testlib/Test/Cardano/Ledger/Alonzo/ImpTest.hs
@@ -14,30 +14,44 @@ module Test.Cardano.Ledger.Alonzo.ImpTest (
   initAlonzoImpNES,
   PlutusArgs (..),
   impAddPlutusScript,
+  impLookupPlutusScript,
+  fixupPPHash,
+  fixupPlutusScripts,
+  addCollateralInput,
   alonzoFixupTx,
+  impGetPlutusContexts,
+  fixupDatums,
 ) where
 
 import Cardano.Crypto.DSIGN.Class (DSIGNAlgorithm (..))
 import Cardano.Crypto.Hash (Hash)
+import Cardano.Ledger.Address (Addr (..))
 import Cardano.Ledger.Alonzo (AlonzoEra)
-import Cardano.Ledger.Alonzo.Core (AlonzoEraScript (..), AlonzoEraTxOut (..), Era (..), EraGov)
+import Cardano.Ledger.Alonzo.Core (AlonzoEraScript (..), AlonzoEraTxOut (..), AsIxItem, Era (..), EraGov)
 import Cardano.Ledger.Alonzo.PParams (AlonzoEraPParams, getLanguageView, ppCostModelsL, ppMaxTxExUnitsL, ppMaxValSizeL)
-import Cardano.Ledger.Alonzo.Scripts (ExUnits (..), plutusScriptLanguage, toAsIx)
+import Cardano.Ledger.Alonzo.Scripts (
+  AsItem (..),
+  ExUnits (..),
+  plutusScriptLanguage,
+  toAsItem,
+  toAsIx,
+ )
 import Cardano.Ledger.Alonzo.Tx (hashScriptIntegrity)
 import Cardano.Ledger.Alonzo.TxBody (AlonzoEraTxBody (..))
 import Cardano.Ledger.Alonzo.TxWits (AlonzoEraTxWits (..), Redeemers (..), TxDats (..))
-import Cardano.Ledger.Alonzo.UTxO (AlonzoScriptsNeeded (..))
+import Cardano.Ledger.Alonzo.UTxO (AlonzoEraUTxO (..), AlonzoScriptsNeeded (..))
 import Cardano.Ledger.Coin (Coin (..))
 import Cardano.Ledger.Core (
   EraIndependentTxBody,
   EraScript (..),
   EraTx (..),
-  EraTxBody (..),
+  EraTxOut (..),
   EraTxWits (..),
   ScriptHash,
  )
+import Cardano.Ledger.Credential (Credential (..))
 import Cardano.Ledger.Crypto (Crypto (..))
-import Cardano.Ledger.Plutus.Data (Data (..), Datum (..), binaryDataToData, hashData)
+import Cardano.Ledger.Plutus.Data (Data (..), Datum (..), hashData)
 import Cardano.Ledger.Plutus.Language (Language (..), Plutus, PlutusLanguage, hashPlutusScript)
 import Cardano.Ledger.Shelley.LedgerState (
   NewEpochState,
@@ -48,12 +62,12 @@ import Cardano.Ledger.Shelley.LedgerState (
   nesEsL,
   utxosUtxoL,
  )
-import Cardano.Ledger.Shelley.UTxO (EraUTxO (..), txinLookup)
+import Cardano.Ledger.Shelley.UTxO (EraUTxO (..))
 import Cardano.Ledger.TxIn (TxIn)
 import Control.Monad (forM)
 import Data.Default.Class (Default)
-import Data.Foldable (Foldable (..))
 import qualified Data.Map.Strict as Map
+import Data.MapExtras (fromElems)
 import Data.Maybe (catMaybes)
 import Data.Set (Set)
 import qualified Data.Set as Set
@@ -62,6 +76,7 @@ import Lens.Micro.Mtl (use, (%=), (.=))
 import Test.Cardano.Ledger.Allegra.ImpTest (impAllegraSatisfyNativeScript)
 import Test.Cardano.Ledger.Alonzo.TreeDiff ()
 import Test.Cardano.Ledger.Common
+import Test.Cardano.Ledger.Imp.Common (expectJust)
 import Test.Cardano.Ledger.Plutus (PlutusArgs (..), ScriptTestContext (..), testingCostModels)
 import Test.Cardano.Ledger.Shelley.ImpTest as ImpTest
 
@@ -112,6 +127,21 @@ impLookupPlutusScript sh = do
     Just (ScriptTestContext plutus _) -> pure $ mkPlutusScript plutus
     Nothing -> pure Nothing
 
+impGetPlutusContexts ::
+  (ScriptsNeeded era ~ AlonzoScriptsNeeded era, EraUTxO era) =>
+  Tx era ->
+  ImpTestM
+    era
+    [(PlutusPurpose AsIxItem era, ScriptHash (EraCrypto era), ScriptTestContext)]
+impGetPlutusContexts tx = do
+  let txBody = tx ^. bodyTxL
+  utxo <- getsNES $ nesEsL . esLStateL . lsUTxOStateL . utxosUtxoL
+  let AlonzoScriptsNeeded asn = getScriptsNeeded utxo txBody
+  mbyContexts <- forM asn $ \(prp, sh) -> do
+    ctx <- getScriptTestContext sh
+    pure $ (prp,sh,) <$> ctx
+  pure $ catMaybes mbyContexts
+
 fixupPlutusScripts ::
   forall era.
   ( EraUTxO era
@@ -123,15 +153,28 @@ fixupPlutusScripts ::
   Tx era ->
   ImpTestM era (Tx era)
 fixupPlutusScripts tx = do
-  utxo <- getsNES $ nesEsL . esLStateL . lsUTxOStateL . utxosUtxoL
+  contexts <- impGetPlutusContexts tx
+  exUnits <- getsNES $ nesEsL . curPParamsEpochStateL . ppMaxTxExUnitsL
   let
-    txBody = tx ^. bodyTxL
-    AlonzoScriptsNeeded asn = getScriptsNeeded utxo txBody
-  mbyContexts <- forM asn $ \(prp, sh) -> do
-    ctx <- getScriptTestContext sh
-    pure $ (prp,sh,) <$> ctx
+    mkNewRedeemers (prpIdx, _, ScriptTestContext _ (PlutusArgs dat _)) =
+      (hoistPlutusPurpose @era toAsIx prpIdx, (Data dat, exUnits))
+    Redeemers oldRedeemers = tx ^. witsTxL . rdmrsTxWitsL
+    newRedeemers = Map.fromList $ mkNewRedeemers <$> contexts
+  pure $
+    tx
+      & witsTxL . rdmrsTxWitsL <>~ Redeemers (Map.union oldRedeemers newRedeemers)
+
+fixupScriptWits ::
+  forall era.
+  ( ScriptsNeeded era ~ AlonzoScriptsNeeded era
+  , AlonzoEraScript era
+  , EraUTxO era
+  ) =>
+  Tx era ->
+  ImpTestM era (Tx era)
+fixupScriptWits tx = do
+  contexts <- impGetPlutusContexts tx
   let
-    contexts = catMaybes mbyContexts
     plutusToScript ::
       forall l.
       PlutusLanguage l =>
@@ -143,31 +186,43 @@ fixupPlutusScripts tx = do
         Nothing -> error "Plutus version not supported by era"
   scriptWits <- forM contexts $ \(_, sh, ScriptTestContext plutus _) ->
     (sh,) <$> plutusToScript plutus
-  exUnits <- getsNES $ nesEsL . curPParamsEpochStateL . ppMaxTxExUnitsL
-  let
-    mkNewRedeemers (prpIdx, _, ScriptTestContext _ (PlutusArgs dat _)) =
-      (hoistPlutusPurpose @era toAsIx prpIdx, (Data dat, exUnits))
-    newRedeemers = Redeemers . Map.fromList $ mkNewRedeemers <$> contexts
-    mkTxDats (_, _, ScriptTestContext _ (PlutusArgs _ datum)) =
-      (\d -> (hashData d, d)) . Data <$> datum
-    txDats = mkTxDats <$> contexts
-    mkInputDats txIn =
-      case txinLookup txIn utxo of
-        Just txOut ->
-          case txOut ^. datumTxOutF of
-            Datum bin -> do
-              let dat = binaryDataToData bin
-              Just (hashData dat, dat)
-            _ -> Nothing
-        Nothing -> error $ "TxIn not found in the UTxO: " <> show txIn
-    inputDats = mkInputDats <$> toList (txBody ^. inputsTxBodyL)
   pure $
     tx
       & witsTxL . scriptTxWitsL <>~ Map.fromList scriptWits
-      & witsTxL . rdmrsTxWitsL <>~ newRedeemers
+
+fixupDatums ::
+  forall era.
+  ( HasCallStack
+  , ScriptsNeeded era ~ AlonzoScriptsNeeded era
+  , AlonzoEraTxWits era
+  , AlonzoEraUTxO era
+  , AlonzoEraTxOut era
+  , ShelleyEraImp era
+  ) =>
+  Tx era ->
+  ImpTestM era (Tx era)
+fixupDatums tx = do
+  contexts <- impGetPlutusContexts tx
+  scripts <- use impScriptsL
+  let
+    txOutScriptHash txOut
+      | Addr _ (ScriptHashObj sh) _ <- txOut ^. addrTxOutL = sh
+      | otherwise = error "TxOut does not have a payment script"
+    spendDatum (ScriptTestContext _ (PlutusArgs _ (Just d))) = Data d
+    spendDatum _ = error "Context does not have a spending datum"
+    filterInline (purpose, _, _) = do
+      AsItem txIn <- expectJust . toSpendingPurpose $ hoistPlutusPurpose toAsItem purpose
+      txOut <- impLookupUTxO @era txIn
+      pure $ case txOut ^. datumTxOutF of
+        DatumHash _dh -> spendDatum <$> Map.lookup (txOutScriptHash txOut) scripts
+        _ -> Nothing
+  datums <- traverse filterInline contexts
+  let TxDats prevDats = tx ^. witsTxL . datsTxWitsL
+  pure $
+    tx
       & witsTxL . datsTxWitsL
-        <>~ TxDats
-          (Map.fromList . catMaybes $ txDats ++ inputDats)
+        .~ TxDats
+          (Map.union prevDats $ fromElems hashData (catMaybes datums))
 
 fixupPPHash ::
   forall era.
@@ -203,6 +258,7 @@ alonzoFixupTx ::
   , AlonzoEraTxWits era
   , AlonzoEraTxBody era
   , ScriptsNeeded era ~ AlonzoScriptsNeeded era
+  , AlonzoEraUTxO era
   ) =>
   Tx era ->
   ImpTestM era (Tx era)
@@ -211,6 +267,8 @@ alonzoFixupTx =
     >=> addCollateralInput
     >=> addRootTxIn
     >=> fixupPlutusScripts
+    >=> fixupScriptWits
+    >=> fixupDatums
     >=> fixupPPHash
     >=> fixupFees
     >=> updateAddrTxWits
--- a/eras/babbage/impl/testlib/Test/Cardano/Ledger/Babbage/ImpTest.hs
+++ b/eras/babbage/impl/testlib/Test/Cardano/Ledger/Babbage/ImpTest.hs
@@ -1,17 +1,38 @@
+{-# LANGUAGE AllowAmbiguousTypes #-}
 {-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE RankNTypes #-}
+{-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE TypeFamilies #-}
+{-# LANGUAGE TypeOperators #-}
 {-# LANGUAGE UndecidableInstances #-}
 {-# OPTIONS_GHC -Wno-orphans #-}
 
-module Test.Cardano.Ledger.Babbage.ImpTest () where
+module Test.Cardano.Ledger.Babbage.ImpTest (babbageFixupTx) where
 
 import Cardano.Crypto.DSIGN.Class (DSIGNAlgorithm (..))
 import Cardano.Crypto.Hash (Hash)
+import Cardano.Ledger.Alonzo.UTxO (AlonzoEraUTxO, AlonzoScriptsNeeded)
 import Cardano.Ledger.Babbage (BabbageEra)
+import Cardano.Ledger.Babbage.Core (
+  AlonzoEraTxWits (..),
+  BabbageEraTxBody (..),
+ )
 import Cardano.Ledger.Core
 import Cardano.Ledger.Crypto (Crypto (..))
+import Cardano.Ledger.Shelley.UTxO (EraUTxO (..))
 import Test.Cardano.Ledger.Allegra.ImpTest (impAllegraSatisfyNativeScript)
-import Test.Cardano.Ledger.Alonzo.ImpTest (alonzoFixupTx, initAlonzoImpNES)
+import Test.Cardano.Ledger.Alonzo.ImpTest (
+  ImpTestM,
+  addCollateralInput,
+  addNativeScriptTxWits,
+  addRootTxIn,
+  fixupDatums,
+  fixupFees,
+  fixupPPHash,
+  fixupPlutusScripts,
+  initAlonzoImpNES,
+  updateAddrTxWits,
+ )
 import Test.Cardano.Ledger.Babbage.TreeDiff ()
 import Test.Cardano.Ledger.Common
 import Test.Cardano.Ledger.Shelley.ImpTest (ShelleyEraImp (..))
@@ -26,4 +47,23 @@ instance
   where
   initImpNES = initAlonzoImpNES
   impSatisfyNativeScript = impAllegraSatisfyNativeScript
-  fixupTx = alonzoFixupTx
+  fixupTx = babbageFixupTx
+
+babbageFixupTx ::
+  ( ShelleyEraImp era
+  , AlonzoEraTxWits era
+  , ScriptsNeeded era ~ AlonzoScriptsNeeded era
+  , BabbageEraTxBody era
+  , AlonzoEraUTxO era
+  ) =>
+  Tx era ->
+  ImpTestM era (Tx era)
+babbageFixupTx =
+  addNativeScriptTxWits
+    >=> addCollateralInput
+    >=> addRootTxIn
+    >=> fixupPlutusScripts
+    >=> fixupDatums
+    >=> fixupPPHash
+    >=> fixupFees
+    >=> updateAddrTxWits
--- a/eras/conway/impl/src/Cardano/Ledger/Conway/TxInfo.hs
+++ b/eras/conway/impl/src/Cardano/Ledger/Conway/TxInfo.hs
@@ -25,6 +25,8 @@ module Cardano.Ledger.Conway.TxInfo (
   transDRep,
   transScriptPurpose,
   transMap,
+  transTxInInfoV1,
+  transTxOutV1,
 ) where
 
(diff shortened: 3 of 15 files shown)


eras/babbage/impl/testlib/Test/Cardano/Ledger/Babbage/ImpTest.hs
```diff
@@ -26,4 +64,82 @@ instance
...
+      forall l.
+      PlutusLanguage l =>
+      Plutus l ->
+      ImpTestM era (Script era)
+    plutusToScript p =
+      case mkPlutusScript @era p of
+        Just x -> pure $ fromPlutusScript x
+        Nothing -> error "Plutus version not supported by era"
+  scriptWits <- forM contexts $ \(_, sh, ScriptTestContext plutus _) ->
+    (sh,) <$> plutusToScript plutus
```
lehins: Reduce duplication by extracting this functionality into a separate function and reusing it here as well as in `fixupScriptWits`

eras/babbage/impl/testlib/Test/Cardano/Ledger/Babbage/ImpTest.hs
```diff
@@ -26,4 +64,82 @@ instance
...
+    refInputs = tx ^. bodyTxL . referenceInputsTxBodyL
+    lookupRefScripts txIn =
+      case txinLookup txIn utxo of
+        Just txOut ->
+          case txOut ^. referenceScriptTxOutL of
+            SJust script -> (hashScript script :)
+            SNothing -> id
+        Nothing -> error $ "TxIn not found in UTxO: " <> show txIn
+  let
+    refScripts = Set.foldr' lookupRefScripts mempty refInputs
```
lehins:
```suggestion
    lookupAddRefScripts txIn acc =
      case txinLookup txIn utxo of
        Just txOut ->
          case txOut ^. referenceScriptTxOutL of
            SJust script -> Set.insert (hashScript script) acc
            SNothing -> acc
        Nothing -> error $ "TxIn not found in UTxO: " <> show txIn
  let
    refScripts = Set.foldr' lookupAddRefScripts Set.empty refInputs
```

eras/babbage/impl/testlib/Test/Cardano/Ledger/Babbage/ImpTest.hs
```diff
@@ -26,4 +64,82 @@ instance
...
+                  : [ (hashData @era (Data sDat), Data sDat)
+                    | Just sDat <- [spendDat]
+                    ]
+          pure $ Map.union dataHashes m
+        Nothing -> error "Script not registered"
+  refDats <- foldrM insertData mempty refScripts
+  pure $
+    tx
+      & witsTxL . scriptTxWitsL
+        <>~ Map.filter filterRefScripts (Map.fromList scriptWits)
```
lehins:
```suggestion
      & witsTxL . scriptTxWitsL %~ Map.withoutKeys refScripts
```

eras/babbage/impl/testlib/Test/Cardano/Ledger/Babbage/ImpTest.hs
```diff
@@ -26,4 +64,82 @@ instance
...
+    lookupRefScripts txIn =
+      case txinLookup txIn utxo of
+        Just txOut ->
+          case txOut ^. referenceScriptTxOutL of
+            SJust script -> (hashScript script :)
+            SNothing -> id
+        Nothing -> error $ "TxIn not found in UTxO: " <> show txIn
+  let
+    refScripts = Set.foldr' lookupRefScripts mempty refInputs
+    filterRefScripts script = hashScript script `notElem` refScripts
```
lehins: Never do membership checks on lists, unless the element you are dealing with does not have a sensible Ord instance, which is almost never the case.
```suggestion
```

eras/babbage/impl/testlib/Test/Cardano/Ledger/Babbage/ImpTest.hs
```diff
@@ -26,4 +64,82 @@ instance
...
+        Just (ScriptTestContext _ (PlutusArgs dat spendDat)) -> do
+          let
+            dataHashes =
+              Map.fromList $
+                (hashData @era (Data dat), Data dat)
+                  : [ (hashData @era (Data sDat), Data sDat)
+                    | Just sDat <- [spendDat]
+                    ]
+          pure $ Map.union dataHashes m
+        Nothing -> error "Script not registered"
```
lehins: It is good to report a bit more info when bad things happen.
```suggestion
        Nothing -> error $ "Script not registered: " ++ show sh
```

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/UtxosSpec.hs
```diff
@@ -0,0 +1,77 @@
...
+      producingTx <-
+        submitTxAnn "Producing transaction" $
+          mkBasicTx mkBasicTxBody
+            & bodyTxL . outputsTxBodyL
+              .~ SSeq.fromList
+                [ refTxOut
+                , scriptLockedTxOut
+                ]
+      let producingTxId = txIdTx producingTx
+      submitTxAnn_ "Consuming transaction" $
```
lehins: We need some more checks here, because from what I can see in the Conway.TxInfo implementation reference scripts are still not enabled for PlutusV1 

Can you please add a check that verifies that there was no witness added for that script, effectively ensuring that it was indeed the reference script that was used.
```suggestion
      tx <- submitTxAnn "Consuming transaction" $
```

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/UtxosSpec.hs
```diff
@@ -0,0 +1,77 @@
...
+        mkBasicTx mkBasicTxBody
+          & bodyTxL . inputsTxBodyL .~ Set.singleton (mkTxInPartial producingTxId 1)
+          & bodyTxL . referenceInputsTxBodyL .~ Set.singleton (mkTxInPartial producingTxId 0)
```
lehins: Could you add another transaction that uses that same script as reference script in the regular `inputsTxBodyL`, effectively spending the txOut, while running the script.

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/UtxosSpec.hs
```diff
@@ -0,0 +1,77 @@
...
+
+spec ::
+  forall era.
+  ( ShelleyEraImp era
+  , BabbageEraTxBody era
+  ) =>
+  SpecWith (ImpTestState era)
+spec =
+  describe "UTXOS" $ do
+    it "can use reference scripts" $ do
```
lehins, line 129: Could you please add one more test that checks that validation fails when a transaction tries to use a TxOut that besides PlutusV1 reference script also contains an inline datum. Since those should still result in a failure.

eras/conway/impl/src/Cardano/Ledger/Conway/TxInfo.hs
```diff
@@ -205,9 +205,6 @@ instance Crypto c => EraPlutusTxInfo 'PlutusV1 (ConwayEra c) where
   toPlutusScriptPurpose proxy = transPlutusPurposeV1V2 proxy . hoistPlutusPurpose toAsItem
 
   toPlutusTxInfo proxy pp epochInfo systemStart utxo tx = do
-    let refInputs = txBody ^. referenceInputsTxBodyL
-    unless (Set.null refInputs) $ Left $ inject $ ReferenceInputsNotSupported @(ConwayEra c) refInputs
-
     timeRange <- Alonzo.transValidityInterval pp epochInfo systemStart (txBody ^. vldtTxBodyL)
     inputs <- mapM (Babbage.transTxInInfoV1 utxo) (Set.toList (txBody ^. inputsTxBodyL))
```
lehins: `Babbage.transTxInInfoV1` will fail for reference scritps. There is more fixing and testing that needs to be done. New functions like these need to be added to Conway.TxInfo

https://github.com/IntersectMBO/cardano-ledger/blob/60e0dcc2da3824e48c46066d917b0eee08947699/eras/babbage/impl/src/Cardano/Ledger/Babbage/TxInfo.hs#L149-L152

https://github.com/IntersectMBO/cardano-ledger/blob/60e0dcc2da3824e48c46066d917b0eee08947699/eras/babbage/impl/src/Cardano/Ledger/Babbage/TxInfo.hs#L101-L102

lehins (changes requested):
It is close, but not quite done

eras/babbage/impl/testlib/Test/Cardano/Ledger/Babbage/ImpTest.hs
```diff
@@ -26,4 +66,83 @@ instance
...
+      mbyStc <- getScriptTestContext sh
+      case mbyStc of
+        Just (ScriptTestContext _ (PlutusArgs dat spendDat)) -> do
+          let
+            dataHashes =
+              Map.fromList $
+                (hashData @era (Data dat), Data dat)
+                  : [ (hashData @era (Data sDat), Data sDat)
+                    | Just sDat <- [spendDat]
+                    ]
```
lehins: I remembered one more simplification:
```suggestion
              fromElems (hashData @era) $
                Data dat : [ Data sDat | Just sDat <- [spendDat]]
```

Which uses this `fromElems`: https://github.com/IntersectMBO/cardano-ledger/blob/60e0dcc2da3824e48c46066d917b0eee08947699/libs/cardano-data/src/Data/MapExtras.hs#L261-L268

eras/babbage/impl/testlib/Test/Cardano/Ledger/Babbage/ImpTest.hs
Soupstraw: I ended up not needing this function at all thanks to one of your other suggestions :slightly_smiling_face:

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/UtxosSpec.hs
```diff
@@ -0,0 +1,157 @@
...
+        submitFailingTx
+          consumingTx
+          [ ConwayUtxowFailure . UtxoFailure . BabbageNonDisjointRefInputs @era . NE.singleton $
+              mkTxInPartial producingTx 0
+          ]
+      pure ()
+    it "fails when using inline datums for PlutusV1" $ do
+      shSpending <- setupPlutusV1Script
+      refTxOut <- mkRefTxOut
+      let
```
lehins: Look at this lonely let! :smile: 
I did not know that his was actually a legal syntax
```suggestion
```

eras/alonzo/impl/testlib/Test/Cardano/Ledger/Alonzo/ImpTest.hs
```diff
@@ -112,6 +117,23 @@ impLookupPlutusScript sh = do
...
+  (ScriptsNeeded era ~ AlonzoScriptsNeeded era, EraUTxO era) =>
+  Tx era ->
+  ImpTestM
+    era
+    [(PlutusPurpose AsIxItem era, ScriptHash (EraCrypto era), ScriptTestContext)]
+impGetPlutusContexts tx = do
+  let
+    txBody = tx ^. bodyTxL
+  utxo <- getsNES $ nesEsL . esLStateL . lsUTxOStateL . utxosUtxoL
+  let
```
aniketd: I don't have any strong opinion on this, but these `let`s on a single line all by themselves look a bit odd.

aniketd (commented):
I don't have enough understanding about scripts in order to submit an approval on this one. 😄 But LGTM @Soupstraw otherwise.

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/UtxosSpec.hs
Soupstraw: Yeah, me neither :smile:

eras/conway/impl/src/Cardano/Ledger/Conway/TxInfo.hs
```diff
@@ -199,20 +205,62 @@ instance
     PlutusPurposeNotSupported purpose ->
       kindObject "PlutusPurposeNotSupported" ["purpose" .= toJSON purpose]
 
+-- | Given a TxOut, translate it for V2 and return (Right transalation).
+-- If the transaction contains any Byron addresses or Babbage features, return Left.
+conwayTransTxOutV1 ::
```
lehins: Functions for all previous eras are imported qualified, so we don't need to disambiguate it in the name of the function. I know it is not how we do the rules, but this is the way txInfo translation is implemented, so it will be good to continue that for consistency. Otherwise we'll end up in the next era with `Conway.conwayTransTxOutV1`, but only for those two functions
```suggestion
transTxOutV1 ::
```

eras/conway/impl/src/Cardano/Ledger/Conway/TxInfo.hs
```diff
@@ -199,20 +205,62 @@ instance
...
+  Either (ContextError era) PV1.TxOut
+conwayTransTxOutV1 txOutSource txOut = do
+  when (isSJust (txOut ^. dataTxOutL)) $ do
+    Left $ inject $ InlineDatumsNotSupported @era txOutSource
+  case Alonzo.transTxOut txOut of
+    Nothing -> Left $ inject $ ByronTxOutInContext @era txOutSource
+    Just plutusTxOut -> Right plutusTxOut
+
+-- | Given a TxIn, look it up in the UTxO. If it exists, translate it to the V1 context
+conwayTransTxInInfoV1 ::
```
lehins: Ditto
```suggestion
transTxInInfoV1 ::
```

eras/conway/impl/src/Cardano/Ledger/Conway/TxInfo.hs
```diff
@@ -199,20 +205,62 @@ instance
...
+        case txinLookup txIn utxo of
+          Just txOut
+            | isSJust (txOut ^. dataTxOutL) -> True
+          _ -> False
+      refInlineDatum = find hasInlineDatum . toList $ txBody ^. referenceInputsTxBodyL
+    case refInlineDatum of
+      Just txIn ->
+        Left . BabbageContextError . InlineDatumsNotSupported $
+          TxOutFromInput txIn
+      Nothing -> pure ()
```
lehins: You need to use `conwayTransTxInInfoV1` translation here on ref inputs while discarding the results, because there are other checks besides the inline datums

lehins: Something along the lines of:
```suggestion
   mapM_ (conwayTransTxInInfoV1 utxo) (Set.toList (txBody ^. referenceInputsTxBodyL))
```

eras/babbage/impl/src/Cardano/Ledger/Babbage/TxInfo.hs
```diff
@@ -88,7 +88,7 @@ transReferenceScript (SJust s) = Just . transScriptHash . hashScript @era $ s
 
 -- | Given a TxOut, translate it for V2 and return (Right transalation).
 -- If the transaction contains any Byron addresses or Babbage features, return Left.
-transTxOutV1 ::
+babbageTransTxOutV1 ::
```
lehins: Going back to my comment about `conway` prefix. Same applies here
```suggestion
transTxOutV1 ::
```

eras/babbage/impl/src/Cardano/Ledger/Babbage/TxInfo.hs
```diff
@@ -97,7 +97,7 @@ transTxOutV1 ::
   TxOutSource (EraCrypto era) ->
   TxOut era ->
   Either (ContextError era) PV1.TxOut
-transTxOutV1 txOutSource txOut = do
+babbageTransTxOutV1 txOutSource txOut = do
```
lehins: Ditto
```suggestion
transTxOutV1 txOutSource txOut = do
```

eras/babbage/impl/testlib/Test/Cardano/Ledger/Babbage/ImpTest.hs
```diff
@@ -26,4 +54,48 @@ instance
...
+        Just txOut ->
+          case txOut ^. referenceScriptTxOutL of
+            SJust script | isPlutusScript script -> Set.insert $ hashScript script
+            _ -> id
+        Nothing -> error $ "TxIn not found in UTxO: " <> show txIn
+  let
+    plutusRefScripts = Set.foldr' lookupAddPlutusRefScripts mempty refInputs
+  pure $
+    tx
+      & witsTxL . scriptTxWitsL %~ (`Map.withoutKeys` plutusRefScripts)
```
lehins: I think this is a suboptimal approach.

I would like for fixup function to be:

1. Idempotent
2. They should only add things that are missing, not remove parts of transaction that are not valid.

This fixup function violates the second point.
For example if I create a transaction and I supply an wrongful witness by myself, the fixup function should not touch it. This is useful for writing negative tests. 


So, I think the best way would be is to adjust the function that adds script witnesses to not add script witnesses whenever reference scripts are available. Something similar in spirit to `fixupDatums`

eras/alonzo/impl/testlib/Test/Cardano/Ledger/Alonzo/ImpTest.hs
```diff
@@ -143,31 +184,43 @@ fixupPlutusScripts tx = do
...
     tx
       & witsTxL . scriptTxWitsL <>~ Map.fromList scriptWits
-      & witsTxL . rdmrsTxWitsL <>~ newRedeemers
-      & witsTxL . datsTxWitsL
-        <>~ TxDats
-          (Map.fromList . catMaybes $ txDats ++ inputDats)
+
+fixupDatums ::
+  forall era.
+  ( ScriptsNeeded era ~ AlonzoScriptsNeeded era
```
lehins:
```suggestion
  ( HasCallStack
  , ScriptsNeeded era ~ AlonzoScriptsNeeded era
```

eras/alonzo/impl/testlib/Test/Cardano/Ledger/Alonzo/ImpTest.hs
```diff
@@ -143,31 +184,43 @@ fixupPlutusScripts tx = do
...
+      | otherwise = error "TxOut does not have a payment script"
+    spendDatum (ScriptTestContext _ (PlutusArgs _ (Just d))) = Data d
+    spendDatum _ = error "Context does not have a spending datum"
+    filterInline (prps, _, _) = do
+      AsItem txIn <- toSpendingPurpose $ hoistPlutusPurpose toAsItem prps
+      let
+        txOut =
+          case txinLookup txIn utxo of
+            Just x -> x
+            Nothing -> error "Could not find TxOut"
```
lehins:
```suggestion
            Nothing -> error $ "Could not find TxOut for: " <> show txIn
```

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp.hs
```diff
@@ -28,6 +33,10 @@ spec ::
   , GovState era ~ ConwayGovState era
   , PredicateFailure (EraRule "LEDGER" era) ~ ConwayLedgerPredFailure era
   , PredicateFailure (EraRule "GOV" era) ~ ConwayGovPredFailure era
+  , PredicateFailure (EraRule "UTXOW" era) ~ BabbageUtxowPredFailure era
+  , PredicateFailure (EraRule "UTXO" era) ~ BabbageUtxoPredFailure era
+  , PredicateFailure (EraRule "UTXOS" era) ~ AlonzoUtxosPredFailure era
+  , ContextError era ~ ConwayContextError era
```
lehins: This would prevent these tests from running in future eras. We need to describe this in the form of `Inject` constraints
```suggestion
```

eras/alonzo/impl/testlib/Test/Cardano/Ledger/Alonzo/ImpTest.hs
```diff
@@ -143,31 +184,43 @@ fixupPlutusScripts tx = do
...
+  utxo <- getUTxO
+  contexts <- impGetPlutusContexts tx
+  scripts <- use impScriptsL
+  let
+    txOutScriptHash txOut
+      | Addr _ (ScriptHashObj sh) _ <- txOut ^. addrTxOutL = sh
+      | otherwise = error "TxOut does not have a payment script"
+    spendDatum (ScriptTestContext _ (PlutusArgs _ (Just d))) = Data d
+    spendDatum _ = error "Context does not have a spending datum"
+    filterInline (prps, _, _) = do
```
lehins: ths s bd nmng cnvntn
```suggestion
    filterInline (purpose, _, _) = do
```

eras/alonzo/impl/testlib/Test/Cardano/Ledger/Alonzo/ImpTest.hs
```diff
@@ -143,31 +184,43 @@ fixupPlutusScripts tx = do
...
+          case txinLookup txIn utxo of
+            Just x -> x
+            Nothing -> error "Could not find TxOut"
+      case txOut ^. datumTxOutF of
+        DatumHash _dh -> spendDatum <$> Map.lookup (txOutScriptHash txOut) scripts
+        _ -> Nothing
+    datums = filterInline <$> contexts
+  pure $
+    tx
+      & witsTxL . datsTxWitsL .~ TxDats (fromElems hashData $ catMaybes datums)
```
lehins: This will overwrite datums that are alread in the transaction. What if I want to test adding datums manually that violate some predicate checks. I believe we only need to add missing datums.

eras/alonzo/impl/testlib/Test/Cardano/Ledger/Alonzo/ImpTest.hs
```diff
@@ -123,15 +152,27 @@ fixupPlutusScripts ::
...
-    AlonzoScriptsNeeded asn = getScriptsNeeded utxo txBody
-  mbyContexts <- forM asn $ \(prp, sh) -> do
-    ctx <- getScriptTestContext sh
-    pure $ (prp,sh,) <$> ctx
+    mkNewRedeemers (prpIdx, _, ScriptTestContext _ (PlutusArgs dat _)) =
+      (hoistPlutusPurpose @era toAsIx prpIdx, (Data dat, exUnits))
+    newRedeemers = Redeemers . Map.fromList $ mkNewRedeemers <$> contexts
+  pure $
+    tx
+      & witsTxL . rdmrsTxWitsL <>~ newRedeemers
```
lehins: Here too, we should only add redeemers that are not already present. Current approach will override any redeemers that have already been added.

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/UtxosSpec.hs
```diff
@@ -0,0 +1,201 @@
...
+scriptLockedTxOut ::
+  forall era.
+  AlonzoEraTxOut era =>
+  ScriptHash (EraCrypto era) ->
+  TxOut era
+scriptLockedTxOut shSpending =
+  mkBasicTxOut
+    (Addr Testnet (ScriptHashObj shSpending) StakeRefNull)
+    (inject $ Coin 1_000_000)
+    & dataHashTxOutL .~ SJust (hashData @era . Data $ P1.I 3)
```
lehins: This will result in a failure if `setupPlutusV1Script` uses a different datum.

I think datum needs to be looked up in the imp state instead (since `ScriptTestContext` for that script has already been added to the state it will be available there)

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/UtxosSpec.hs
```diff
@@ -0,0 +1,201 @@
...
+          consumingTx
+          [ ConwayUtxowFailure
+              . UtxoFailure
+              . AlonzoInBabbageUtxoPredFailure
+              . UtxosFailure
+              $ CollectErrors
+                [ BadTranslation
+                    . BabbageContextError
+                    . InlineDatumsNotSupported
+                    $ TxOutFromInput lockedTxIn
```
lehins: This will allow usage of this test in future eras
```suggestion
                [ BadTranslation
                    . inject
                    $ TxOutFromInput lockedTxIn
```

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/UtxosSpec.hs
```diff
@@ -0,0 +1,201 @@
...
+            & bodyTxL . inputsTxBodyL .~ Set.singleton lockedTxIn
+            & bodyTxL . referenceInputsTxBodyL .~ Set.singleton (mkTxInPartial producingTx 0)
+      impAnn "Consuming transaction" $
+        submitFailingTx
+          consumingTx
+          [ ConwayUtxowFailure
+              . UtxoFailure
+              . AlonzoInBabbageUtxoPredFailure
+              . UtxosFailure
+              $ CollectErrors
```
lehins: #4071 will allow this to be
```suggestion
          [ injectFailure
              $ CollectErrors
```

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/UtxosSpec.hs
```diff
@@ -0,0 +1,200 @@
...
+            & bodyTxL . inputsTxBodyL
+              .~ Set.fromList
+                [ mkTxInPartial producingTx 0
+                , mkTxInPartial producingTx 1
+                ]
+            & bodyTxL . referenceInputsTxBodyL .~ Set.singleton (mkTxInPartial producingTx 0)
+      _ <-
+        submitFailingTx
+          consumingTx
+          [ ConwayUtxowFailure . UtxoFailure . BabbageNonDisjointRefInputs @era $
```
lehins: With #4071 this becomes:
```suggestion
          [ injectFailure . BabbageNonDisjointRefInputs $
```

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/UtxosSpec.hs
Soupstraw: `setupPlutusV1Script` isn't really meant to be used outside of `UtxosSpec`, I just factored it out since it is used in multiple tests in this module.

lehins: Sure, but it is a top level function that has this magic value that is directly related to the same hard coded value in a different top level function. This is the sort of stuff that is omnipresent in the old tests, which we should avoid as much as possible.

Even if we never change this function, for the sake of the future reader it is better to be explicit about their relation.

Soupstraw: I just moved the datum to the top level, I think looking it up from the ImpState would be overkill

libs/cardano-ledger-core/testlib/Test/Cardano/Ledger/Imp/Common.hs
```diff
@@ -298,6 +300,16 @@ shouldBeLeft e x = expectLeft e >>= (`shouldBe` x)
 shouldBeLeftExpr :: (HasCallStack, ToExpr a, ToExpr b, Eq a, MonadIO m) => Either a b -> a -> m ()
 shouldBeLeftExpr e x = expectLeftExpr e >>= (`shouldBeExpr` x)
 
+expectJust :: MonadIO m => Maybe a -> m a
```
lehins: That's prety important for expectations
```suggestion
expectJust :: (MonadIO m, HasCallStack) => Maybe a -> m a
```

libs/cardano-ledger-core/testlib/Test/Cardano/Ledger/Imp/Common.hs
```diff
@@ -298,6 +300,16 @@ shouldBeLeft e x = expectLeft e >>= (`shouldBe` x)
...
+expectJust (Just x) = pure x
+expectJust Nothing = assertFailure "Expected Just, got Nothing"
+
+expectNothing :: (MonadIO m, ToExpr a) => Maybe a -> m ()
```
lehins: For consistency with other functions like `shouldBeLeftExpr` and callstack
```suggestion
expectNothingExpr :: (MonadIO m, ToExpr a, HasCalStack) => Maybe a -> m ()
```

eras/babbage/impl/testlib/Test/Cardano/Ledger/Babbage/ImpTest.hs
```diff
@@ -26,4 +47,24 @@ instance
...
+  , AlonzoEraUTxO era
+  ) =>
+  Tx era ->
+  ImpTestM era (Tx era)
+babbageFixupTx =
+  addNativeScriptTxWits
+    >=> addCollateralInput
+    >=> addRootTxIn
+    >=> fixupPlutusScripts
+    -- >=> babbageFixupScriptWits
```
lehins: Forgotten commented out line
```suggestion
```

eras/alonzo/impl/testlib/Test/Cardano/Ledger/Alonzo/ImpTest.hs
```diff
@@ -143,31 +186,43 @@ fixupPlutusScripts tx = do
...
+      | Addr _ (ScriptHashObj sh) _ <- txOut ^. addrTxOutL = sh
+      | otherwise = error "TxOut does not have a payment script"
+    spendDatum (ScriptTestContext _ (PlutusArgs _ (Just d))) = Data d
+    spendDatum _ = error "Context does not have a spending datum"
+    filterInline (purpose, _, _) = do
+      AsItem txIn <- expectJust . toSpendingPurpose $ hoistPlutusPurpose toAsItem purpose
+      txOut <- impLookupUTxO @era txIn
+      case txOut ^. datumTxOutF of
+        DatumHash _dh -> pure $ spendDatum <$> Map.lookup (txOutScriptHash txOut) scripts
+        _ -> pure Nothing
```
lehins:
```suggestion
      pure $ case txOut ^. datumTxOutF of
        DatumHash _dh -> spendDatum <$> Map.lookup (txOutScriptHash txOut) scripts
        _ -> Nothing
```

lehins (approved):
This is perfect!!!! Thank you very much for this work.

PR #4065, opened 2024-02-09
Remove `ConwayPool` rule definition

lehins:
# Description

Conway's `POOL` is identical to the one in Shelley. There is no need to redefine it.

This PR removes `ConwayPool` in favor of `ShelleyPOOL`

It also:

* Removes usage of `ShelleyTx` in Shelley and Allegra rules in favor of `Tx` type family
* Makes `EraRule rule era` invective in the `rule`. This will help avoid mistakes like this:

```haskell
type instance EraRule "UTXOW" (ConwayEra c) = ConwayUTXOW (ConwayEra c)
type instance EraRule "UTXOS" (ConwayEra c) = ConwayUTXOW (ConwayEra c)
```

Unfortunately it can't be made injective in `era` as well, because of `TICKN`:
https://github.com/IntersectMBO/cardano-ledger/blob/0e65b5e7ad45df9a7ad7ed9bfe8364d5153a2c15/eras/shelley-ma/test-suite/test/Tests.hs#L34-L36

Maybe in the future will be able to solve that one, which will help with type inference

# Checklist

- [x] Self-reviewed the diff


aniketd (approved):
LGTM 👍 Thank you!

PR #4071, opened 2024-02-10
Introduction of InjectRuleFailure

lehins:
# Description

TLDR. Add a consistent way to promote and inject predicate failures across eras and from one rule to another.

Somewhat lengthy description of the problem that this PR solves is below.

When trying to implement a test like that

```haskell
  submitFailingTx txBad [inject $ InvalidWitnessesUTXOW @era [aliceVKey]]
```
I ran into this failure:

```
test/Main.hs:23:9: error:
    • Couldn't match type: BabbageUtxowPredFailure (BabbageEra StandardCrypto)
                     with: ShelleyUtxowPredFailure (BabbageEra StandardCrypto)
```

For reference, this is the predicate failure type expected by the test:
```haskell
data ShelleyUtxowPredFailure era
  = InvalidWitnessesUTXOW
      ![VKey 'Witness (EraCrypto era)]
  | ...
```

For reference here is our rule callstack: 

`BBODY` -> `LEDGERS` -> `LEDGER` -> `UTXOW` -> `UTXO` -> `UTXOS`

PredicateFailure hierarchy:

```haskell
data ShelleyLedgerPredFailure era
  = UtxowFailure (PredicateFailure (EraRule "UTXOW" era))
  | DelegsFailure (PredicateFailure (EraRule "DELEGS" era))

data BabbageUtxowPredFailure era
  = AlonzoInBabbageUtxowPredFailure !(AlonzoUtxowPredFailure era)
  | UtxoFailure !(PredicateFailure (EraRule "UTXO" era))
    ...
```

This is how we inject into: `BabbageUtxowPredFailure`.
```haskell
instance Inject (ShelleyUtxowPredFailure era) (BabbageUtxowPredFailure era) where
  inject = AlonzoInBabbageUtxowPredFailure . ShelleyInAlonzoUtxowPredFailure
```

So far so good. 

Next step is we need to check how we inject the `ShelleyLedgerPredFailure` into
`PredicateFailure (EraRule "LEDGER" era)`, which we need because of our tests executing the `LEDGER` rule. Let's see what Shelley does about it:

```haskel
instance
  PredicateFailure (EraRule "UTXOW" era) ~ ShelleyUtxowPredFailure era =>
  Inject (ShelleyUtxowPredFailure era) (ShelleyLedgerPredFailure era)
  where
  inject = UtxowFailure
```

So, this is our problem. We also need to do this same injection of `ShelleyUtxowPredFailure era` -> `ShelleyLedgerPredFailure era`, but with a different path. In other words we need an instance like this for Babbage:

```haskell
instance
  PredicateFailure (EraRule "UTXOW" era) ~ ShelleyUtxowPredFailure era =>
  Inject (ShelleyUtxowPredFailure era) (ShelleyLedgerPredFailure era)
  where
  inject = UtxowFailure . AlonzoInBabbageUtxowPredFailure . ShelleyInAlonzoUtxowPredFailure
```

Quick and easy solution for this particular case would be fairly simple, namely making those instances era specific:

```haskel
instance
  Inject (ShelleyUtxowPredFailure (ShelleyEra c)) (ShelleyLedgerPredFailure (ShelleyEra c))
  where
  inject = UtxowFailure

instance
  Inject (ShelleyUtxowPredFailure (AllegraEra c)) (ShelleyLedgerPredFailure (AllegraEra c))
  where
  inject = UtxowFailure
 ...

instance
  Inject (ShelleyUtxowPredFailure (BabbageEra c)) (ShelleyLedgerPredFailure (BabbageEra c))
  where
  inject = UtxowFailure . AlonzoInBabbageUtxowPredFailure . ShelleyInAlonzoUtxowPredFailure
```

The problem with quick and easy solution is that there are plenty more cases like this
which we will run into when we try adopt older tests for newer eras. This was a
tricky issue to debug. So to avoid wasting time of other developers going forward I came
up with consistent and exhaustive implementation for all cases, while improving type
inference in the process.

# Checklist

- [x] Self-reviewed the diff


libs/cardano-ledger-core/src/Cardano/Ledger/Core/Era.hs
```diff
@@ -106,6 +114,37 @@ instance Crypto c => Era (ByronEra c) where
...
+type instance EraRuleFailure "TICKF" era = VoidRuleFailure "TICKF" era
+type instance EraRuleFailure "UPEC" era = VoidRuleFailure "UPEC" era
+type instance EraRuleFailure "RUPD" era = VoidRuleFailure "RUPD" era
+type instance EraRuleFailure "POOLREAP" era = VoidRuleFailure "POOLREAP" era
+
+class
+  EraRuleFailure rule era ~ PredicateFailure (EraRule rule era) =>
+  InjectRuleFailure (rule :: Symbol) t era
+  where
+  injectFailure :: t era -> EraRuleFailure rule era
```
Soupstraw, line 147:
```suggestion
  injectFailure :: t era -> EraRuleFailure rule era
  default injectFailure :: t era ~ EraRuleFailure rule era => t era -> EraRuleFailure rule era
  injectFailure = id
```

This might get rid of some of the boilerplate since we always want this to be reflexive for any rule.

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Bbody.hs
```diff
@@ -72,6 +80,38 @@ data ShelleyBbodyPredFailure era
...
+
+instance InjectRuleFailure "BBODY" ShelleyDelplPredFailure (ShelleyEra c) where
+  injectFailure = LedgersFailure . injectFailure
+
+instance InjectRuleFailure "BBODY" ShelleyPoolPredFailure (ShelleyEra c) where
+  injectFailure = LedgersFailure . injectFailure
+
+instance InjectRuleFailure "BBODY" ShelleyDelegPredFailure (ShelleyEra c) where
+  injectFailure = LedgersFailure . injectFailure
+
```
Soupstraw, line 113: Couldn't GHC derive these automatically via transitivity?
```suggestion
instance InjectRuleFailure "LEDGERS" t (ShelleyEra c) => InjectRuleFailure "BBODY" t (ShelleyEra c) where
  injectFailure = LedgersFailure . injectFailure
```

lehins: As discussed during the meeting, this would result in overlapping instances, which we really want to avoid

libs/cardano-ledger-core/src/Cardano/Ledger/Core/Era.hs
lehins: Yeah, this is true. Thanks

Soupstraw (approved):
LGTM

aniketd (approved):
LGTM! 🙌

PR #4081, opened 2024-02-14
Fix MIR rule using wrong values for treasury and reserves

### LINKED ISSUE #4077: alonzo spec delta changes for treasury and reserves from MIR certs are unclear
image

Alonzo spec describes this change for the MIR certs processing, with the new `deltaReserves` and `deltaTreasury` values produced from MIR certs of the type `[ 0 / 1, coin ]` that are supposed to move funds directly between pots. The question is related to the `MIR` transition still using only the `totT` and `totR` values to change the actual pots, and these values are produced only from aggregating the `irwdT` and `irwdR`, so only the certs with staking keys. It seems like the `deltaReserves` and `deltaTreasury` values are only used to calculate the `enough` value and are not used to change the actual pots.

It also seems from the MIR rule that the validation is not correct since if checks if `totR` is less or equal to `treasury + deltaTreasury`, but then produces `acnt'` with `treasury - totT`, which means, for example, the treasury technically might become negative in case `totT > treasury` but `totT <= (treasury + deltaTreasury)`.

Am I missing something?

WhatisRT:
# Description

Fixes #4077.

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/eras/alonzo/formal-spec/mir.tex
+++ b/eras/alonzo/formal-spec/mir.tex
@@ -240,7 +240,7 @@ Lastly, Figure \ref{fig:rules:mir} is the updated MIR rule.
       \\
       \var{acnt'}\leteq
       {\begin{cases}
-          (\var{treasury}-\var{totT},~\var{reserves}-\var{totR})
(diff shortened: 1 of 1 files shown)


lehins (approved):
LGTM.
I can confirm that this makes the spec in line with implementation.

PR #4082, opened 2024-02-14
Deprecate `PPUPPredFailure`

lehins:
# Description

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Utxo.hs
+++ b/eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Utxo.hs
@@ -1,21 +1,27 @@
 {-# LANGUAGE DataKinds #-}
+{-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE MultiParamTypeClasses #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TupleSections #-}
+{-# LANGUAGE TypeApplications #-}
 {-# LANGUAGE TypeFamilies #-}
+{-# LANGUAGE TypeOperators #-}
 {-# OPTIONS_GHC -Wno-orphans #-}
 
-module Cardano.Ledger.Conway.Rules.Utxo () where
+module Cardano.Ledger.Conway.Rules.Utxo (allegraToConwayUtxoPredFailure) where
 
-import Cardano.Ledger.Allegra.Rules (AllegraUtxoPredFailure, shelleyToAllegraUtxoPredFailure)
+import Cardano.Ledger.Allegra.Rules (shelleyToAllegraUtxoPredFailure)
+import qualified Cardano.Ledger.Allegra.Rules as Allegra (AllegraUtxoPredFailure (..))
 import Cardano.Ledger.Alonzo.Rules (
   AlonzoUtxoPredFailure (..),
   AlonzoUtxosPredFailure,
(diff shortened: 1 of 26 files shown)


libs/cardano-ledger-core/src/Cardano/Ledger/Core/Era.hs
```diff
@@ -20,7 +22,8 @@ module Cardano.Ledger.Core.Era (
   -- ** Rules
   EraRule,
   EraRuleFailure,
-  VoidRuleFailure,
```
teodanciu, line 23: Should we mention the removal of `VoidRuleFailure` in CHANGELOG ?   (or its replacement with `VoidEraRule`)

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Utxo.hs
```diff
@@ -27,11 +33,31 @@ instance InjectRuleFailure "UTXO" AlonzoUtxoPredFailure (ConwayEra c) where
...
+  Allegra.OutsideValidityIntervalUTxO vi slotNo -> OutsideValidityIntervalUTxO vi slotNo
+  Allegra.MaxTxSizeUTxO x y -> MaxTxSizeUTxO x y
+  Allegra.InputSetEmptyUTxO -> InputSetEmptyUTxO
+  Allegra.FeeTooSmallUTxO c1 c2 -> FeeTooSmallUTxO c1 c2
+  Allegra.ValueNotConservedUTxO vc vp -> ValueNotConservedUTxO vc vp
+  Allegra.WrongNetwork x y -> WrongNetwork x y
+  Allegra.WrongNetworkWithdrawal x y -> WrongNetworkWithdrawal x y
+  Allegra.OutputTooSmallUTxO x -> OutputTooSmallUTxO x
+  Allegra.UpdateFailure x -> absurdEraRule @"PPUP" @era x
+  Allegra.OutputBootAddrAttrsTooBig xs -> OutputTooBigUTxO (map (0,0,) xs)
```
teodanciu: I wonder  why it's not `OutputBootAddrAttrsTooBig xs` here, but I can see that this is how it is in `allegraToAlonzoUtxoPredFailure`, so it's probably correct.

teodanciu (approved):
Looks good to me

libs/cardano-ledger-core/src/Cardano/Ledger/Core/Era.hs
lehins: Yes, `VoidEraRule` is a replacement. I'll adjust the changelog. thanks for noticing

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Utxo.hs
lehins: That is an interesting observation. I have not looked into why it was that way. I only copy pasted the `allegraToAlonzoUtxoPredFailure` and adjusted the part that was relevant for Conway.

It looks like a mistake in injection that was introduced here: https://github.com/IntersectMBO/cardano-ledger/commit/47d8a35bf31fdd5c1bb5c20b40abd8303884e13f#diff-a9a408a2190f9a7d639380d4faf3d0d0178c9b96d2195c42d8bf7fdd505faa3aR566

I'll fix it for Conway and we can clean it up for older eras later.

lehins: Done

PR #4084, opened 2024-02-14
Fix intermittent test failures in leader-proportion test

### LINKED ISSUE #4076: Fix flakiness in a property test
This test has been failing for years now (See #1638). Which makes it a useless test in presence of randomness

```
      We are elected as leader proportional to our stake:                    FAIL (6.64s)
        *** Failed! Falsified (after 89 tests and 19 shrinks):
        ASC (ActiveSlotCoeff {unActiveSlotVal = 4993816495258313 % 18014398509481984, unActiveSlotLog = -3246400179917455809567734387442081})
        StakeProportion (143882283013 % 274877906944)
        247 /< 378 /< 378 (p=0.15627595358256552)
        Use --quickcheck-replay=529930 to reproduce.
        Use -p '/We are elected as leader proportional to our stake/' to rerun this test only.
```

We do need to get rid of this failure on CI, since it is getting in a way of development.
The proper way would be to investigate the reason and fix in a way that it could never fail in presence of randomness, however that seems to be pretty hard considering the nature of the test.  Therefore, instead of removing or disabling the test we should hard code a set seed on that test which would essentially convert this property test into a unit test.

neilmayhew:
Closes #4076

# Description

# Checklist

- [x] Self-reviewed the diff


lehins (approved):
Perfect!

PR #4086, opened 2024-02-15
Change applySTS to return NonEmpty (PredicateFailure s)

### LINKED ISSUE #4054: Change `applySTS` to return `NonEmpty (PredicateFailure s)`
`applySTSEither` and `applySTS` currently return a `Left [PredicateFailure s]` whenever there is an STS rule failure. This means that there is always this impossible case `Left []` that users need to account for:

https://github.com/IntersectMBO/cardano-ledger/blob/e73acc68d1ee98e59849288a9825b8fe1fb83324/libs/small-steps/src/Control/State/Transition/Extended.hs#L524-L540

We need to fix that by ensuring that empty case cannot happen with the use of `NonEmpty`. This will be a breaking change, but a very useful one.

neilmayhew:
Closes #4054

# Description

This eliminates the anomalous possibility of a `Left` value containing no failures, which we can't treat as a success because it doesn't contain a return value.

I wondered about adding a type alias for `NonEmpty (PredicateFailure f)` or even `Either (NonEmpty (PredicateFailure f)) r` to simplify the function type signatures, but decided that would make the code less clear for someone not familiar with the alias.

This touched a lot more lines than I was expecting.

# Checklist

- [x] Self-reviewed the diff


neilmayhew:
The ghc 8.10.7 builds are failing because there's no `singleton` in `Data.List.NonEmpty`. I assume you'll want us to fix this.

I could fix it by using `fromList`, turning on `OverloadedLists`, using CPP to conditionally define a singleton function, or construct singletons explicitly with `:| []`.

All the uses of `NE.singleton` are in test code, so the partiality of using `fromList` (directly, or via `OverloadedLists`) won't matter.

eras/shelley/test-suite/src/Test/Cardano/Ledger/Shelley/Examples/Consensus.hs
```diff
@@ -135,8 +135,8 @@ defaultShelleyLedgerExamples mkWitnesses mkAlonzoTx value txBody auxData transla
     , sleHashHeader = exampleHashHeader (Proxy @era)
     , sleTx = mkAlonzoTx tx
     , sleApplyTxError =
-        ApplyTxError
-          [DelegsFailure $ DelegateeNotRegisteredDELEG @era (mkKeyHash 1)]
+        ApplyTxError . NE.singleton . DelegsFailure $
```
lehins: As you pointed out in the comment GHC-8.10 doesn't have `singleton` defined. Solution is quite simple. You can either use `pure`, or in tests it is allowed for us to use the partial usage of `OverloadedList`
```suggestion
        ApplyTxError . pure . DelegsFailure $
```
Naturally, same will apply to all other uses of `singleton` in this PR

libs/small-steps/src/Control/State/Transition/Extended.hs
```diff
@@ -526,24 +526,24 @@ applySTSOptsEither ::
...
             ((st, []), evts) -> Right (st, evts)
-            ((_, pfs), _) -> Left pfs
+            ((_, pf : pfs), _) -> Left $ pf :| pfs
 
 applySTS ::
   forall s m rtype.
   (STS s, RuleTypeRep rtype, m ~ BaseM s) =>
   RuleContext rtype s ->
-  m (Either [PredicateFailure s] (State s))
+  m (Either (NonEmpty (PredicateFailure s)) (State s))
```
lehins, line 546: I know I haven't talked about this yet, but developer that introduces changes to the package is responsible for adjusting the version and all the bounds.

For example, this is a breaking change, which besides being recorded in the changelog also requires a major version bump and a lower bound update on all packages in the repo that were affected by this change

libs/cardano-protocol-tpraos/src/Cardano/Protocol/TPraos/API.hs
```diff
@@ -325,7 +326,7 @@ view
 --  application of the TICK rule at the target slot to the curernt ledger state.
 
 newtype FutureLedgerViewError era
-  = FutureLedgerViewError [PredicateFailure (EraRule "TICKF" era)]
+  = FutureLedgerViewError (NonEmpty (PredicateFailure (EraRule "TICKF" era)))
```
lehins, line 329: Same here and for `ChainTransitionError`, this will require a major version bump and a changelog entry. I don't think any packages in this repo will be affected by this change, so I no lower bounds need to be changed

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/Preds/Tx.hs
```diff
@@ -974,12 +973,12 @@ oneTest sizes proof = do
...
-        ( whenFail
-            (putStrLn ("\napplySTS fails\n" ++ unlines (map show errs)) >> (goRepl proof env2 ""))
-            (counterexample ("\napplySTS fails\n" ++ unlines (map show errs)) (True === False))
-        )
+    Left errs -> do
+      let msg = unlines $ "" : "applySTS fails" : map show (toList errs)
+      pure $
+        whenFail
+          (putStrLn msg >> goRepl proof env2 "")
+          (counterexample msg $ True === False)
```
lehins:
```suggestion
          (counterexample msg False)
```

eras/shelley/test-suite/test/Test/Cardano/Ledger/Shelley/RulesTests.hs
```diff
@@ -132,7 +133,7 @@ testAliceSignsAlone =
 
 testAliceDoesntSign :: Assertion
 testAliceDoesntSign =
-  utxoSt' @?= Left [ScriptWitnessNotValidatingUTXOW (Set.singleton $ hashScript @C aliceOnly)]
+  utxoSt' @?= (Left . NE.singleton . ScriptWitnessNotValidatingUTXOW) wits
```
lehins: Maybe enabling `OverloadedLists` in those test modules could be a good idea, as you did in some other places?

eras/shelley/test-suite/test/Test/Cardano/Ledger/Shelley/Examples/Mir.hs
```diff
@@ -271,22 +269,14 @@ mirFailFunds pot treasury llNeeded llReceived =
...
-                )
-            )
-        ]
+        . NE.singleton
+        . BbodyFailure
+        . LedgersFailure
+        . LedgerFailure
+        . DelegsFailure
+        . DelplFailure
+        . DelegFailure
```
lehins: Let's simplify this a bit
```suggestion
        . injectFailure
```

eras/shelley/test-suite/test/Test/Cardano/Ledger/Shelley/Examples/Mir.hs
```diff
@@ -243,15 +244,12 @@ mirFailWits pot =
...
-                        MIRInsufficientGenesisSigsUTXOW ws
-                    )
-                )
-            )
-        ]
+        . NE.singleton
+        . BbodyFailure @(ShelleyEra c)
+        . LedgersFailure
+        . LedgerFailure
+        . UtxowFailure
```
lehins: Let's simplify this one a bit (thanks to the recent change that was merged that I was talking about during the meeting)
```suggestion
        . injectFailure
```

eras/byron/ledger/impl/test/Test/Cardano/Chain/Block/Model.hs
```diff
@@ -273,7 +273,7 @@ classifyTransactions =
 invalidChainTracesAreRejected ::
   TestLimit ->
   [(Int, SignalGenerator CHAIN)] ->
-  ([PredicateFailure CHAIN] -> ChainValidationError -> PropertyT IO ()) ->
+  (NonEmpty (PredicateFailure CHAIN) -> ChainValidationError -> PropertyT IO ()) ->
```
lehins, line 276: It is worth pointing out that a change like this requires two things:
* changelog entry
* major version bump, since it is  breaking change
* lower bounds update on the `small-test` package version (See also my comment on that package)
* lower bound update on any package affected by this change (which I don't think there are any in this repo)

More info can be found on the process in the [RELEASING document](https://github.com/IntersectMBO/cardano-ledger/blob/master/RELEASING.md)

eras/byron/ledger/impl/test/Test/Cardano/Chain/Delegation/Model.hs
```diff
@@ -58,7 +58,7 @@ prop_commandSDELEG = withTests 25 . property $ do
 
 data StateSDELEG (v :: Type -> Type) = StateSDELEG
   { abstractState :: Abstract.DSState
-  , lastAbstractResult :: Either [STS.PredicateFailure SDELEG] (STS.State SDELEG)
+  , lastAbstractResult :: Either (NonEmpty (STS.PredicateFailure SDELEG)) (STS.State SDELEG)
```
lehins, line 61: Ditto, changelog entry

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/ImpTest.hs
```diff
@@ -350,7 +351,7 @@ submitFailingVote ::
   ) =>
   Voter (EraCrypto era) ->
   GovActionId (EraCrypto era) ->
-  [PredicateFailure (EraRule "LEDGER" era)] ->
+  NonEmpty (PredicateFailure (EraRule "LEDGER" era)) ->
```
lehins, line 398: This package, despite introducing a breaking change will not need a version bump, since it has already been increased. Just FYI.

eras/shelley-ma/test-suite/test/Test/Cardano/Ledger/Mary/Examples.hs
```diff
@@ -16,23 +16,24 @@ import Cardano.Ledger.UTxO (UTxO)
...
 import Lens.Micro
 import Test.Cardano.Ledger.Mary.TreeDiff ()
 import Test.Cardano.Ledger.Shelley.Utils (applySTSTest, runShelleyBase)
 import Test.Tasty.HUnit (Assertion, (@?=))
 
 ignoreAllButUTxO ::
-  Either [PredicateFailure (ShelleyLEDGER Mary)] (LedgerState Mary) ->
-  Either [PredicateFailure (ShelleyLEDGER Mary)] (UTxO Mary)
+  Either (NonEmpty (PredicateFailure (ShelleyLEDGER Mary))) (LedgerState Mary) ->
+  Either (NonEmpty (PredicateFailure (ShelleyLEDGER Mary))) (UTxO Mary)
```
lehins, line 28: This is an interesting one with respect to versioning. This is not a user facing change, since it is in the test suite. But, nevertheless it is a change and if we don't bump up the version this package will not be released, which will lead to CHaPs failing CI (due to it trying to build an older version that is not compatible with newer small-steps)

So, this requires a minor version bump. But, because it has already been bumped recently you don't need to do anything about it in this PR :)

This one also I bring up as a good example of our versioning processes.

lehins (changes requested):
Aside from required changes to versions and changelogs this PR looks great.

eras/shelley/test-suite/src/Test/Cardano/Ledger/Shelley/Examples/Consensus.hs
neilmayhew: Using `pure` is a neat idea. It seems like it would be a little obscure, though, since it's not obvious what it's doing. However, most of the time someone reading the code doesn't need to care. I'll think about it, and use either that or overloaded lists.

neilmayhew:
Rebased on `master` and fixed ghc 8.10.7 build

eras/byron/ledger/impl/test/Test/Cardano/Chain/Block/Model.hs
neilmayhew: The changes in `eras/byron/ledger/impl/test` are in functions that aren't exported from their modules, so I don't think they need a changelog entry.

However, I'll add/update lower bounds for `small-steps` and `small-steps-test` anywhere they're used.

eras/byron/ledger/impl/test/Test/Cardano/Chain/Delegation/Model.hs
neilmayhew: Not exported anywhere

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/ImpTest.hs
neilmayhew: Changelog updated

eras/byron/ledger/impl/test/Test/Cardano/Chain/Block/Model.hs
lehins: Byron has a slightly different structure and this module is actually a publicly visible module: https://github.com/IntersectMBO/cardano-ledger/blob/4223cd483af91913c356a882a1ea1ff4357fe3fc/eras/byron/ledger/impl/test/cardano-ledger-byron-test.cabal#L118

libs/cardano-protocol-tpraos/src/Cardano/Protocol/TPraos/API.hs
neilmayhew: `cardano-protocol-tpraos` is already at version `1.1.0.0` so I don't think it needs a version bump

eras/byron/ledger/impl/test/Test/Cardano/Chain/Block/Model.hs
neilmayhew: The module is visible, but AFAICT `invalidChainTracesAreRejected` and `invalidSizesAreRejected` aren't exported from the module

neilmayhew: However, we don't get an unused function warning, and I don't understand why

neilmayhew:
I rebased, but kept the review changes as separate commits. I'll squash before merging.

eras/byron/ledger/impl/test/Test/Cardano/Chain/Block/Model.hs
lehins: Oh yeah, sorry. You are totally right. I misread your first message

lehins (approved):
Beautiful!!
Thank you.

PR #4087, opened 2024-02-16
Move `EpochInterval` to `cardano-base`

### LINKED ISSUE #3944: Move `EpochInterval` to `cardano-base`
The package where `EpochNo` and `EpochSize` are defined is `cardano-slotting`. It only makes sense for `EpochInterval` and `addEpochInterval` to live there as well.

See a related ticket: https://github.com/IntersectMBO/cardano-base/issues/455

Also adding `IsList` instance for `StrictSeq` would be nice too: https://github.com/IntersectMBO/cardano-base/issues/444

This is a multi-stage process:
* [x] - Add changed to the `cardano-base`
* [x] - Release to CHaP
* [x] - Update cabal index
* [x] - Account for the changes in cardano-ledger

Lucsanszky:
# Description
Resolves #3944 

# Checklist

- [x] Self-reviewed the diff


Lucsanszky:
Quickly fixed the things that broke due to the release of `cardano-slotting-0.2.0.0` but I still need to go over the code and do some cleanup, hence the PR is still a draft.

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/NewEpoch.hs
```diff
@@ -162,7 +162,7 @@ newEpochTransition = do
       , eNo
       ) <-
     judgmentContext
-  if eNo /= eNoL + 1
+  if eNo /= EpochNo (unEpochNo eNoL + 1)
```
lehins: I think this is a little bit more type appropriate, despite being a little bit more verbose:
```suggestion
  if eNo /= addEpochInterval eNoL (EpochIntterval 1)
```

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Ppup.hs
```diff
@@ -214,7 +214,7 @@ ppupTransitionNonEmpty = do
               , futureProposals = ProposedPPUpdates fpupS
               }
         else do
-          currentEpoch + 1 == te ?! PPUpdateWrongEpoch currentEpoch te VoteForNextEpoch
+          unEpochNo currentEpoch + 1 == unEpochNo te ?! PPUpdateWrongEpoch currentEpoch te VoteForNextEpoch
```
lehins: Ditto:
```suggestion
          addEpochInteval currentEpoch 1 == te ?! PPUpdateWrongEpoch currentEpoch te VoteForNextEpoch
```

libs/cardano-ledger-binary/src/Cardano/Ledger/Binary/Encoding/EncCBOR.hs
```diff
@@ -1147,6 +1147,8 @@ instance EncCBOR BlockNo where
   encCBOR = fromPlainEncoding . Serialise.encode
   encodedSizeExpr size = encodedSizeExpr size . fmap unBlockNo
 
+deriving instance EncCBOR EpochInterval
```
lehins, line 1150: As you already found out from changes to CHaPs, this change will require a lower bound update on `cardano-slotting`

libs/cardano-ledger-core/src/Cardano/Ledger/BaseTypes.hs
```diff
@@ -53,8 +53,6 @@ module Cardano.Ledger.BaseTypes (
   activeSlotLog,
   module Data.Maybe.Strict,
   BlocksMade (..),
-  EpochInterval (..),
-  addEpochInterval,
```
lehins, line 57: We could continue re-exporting it from here to reduce downstream breakage

lehins (commented):
Some changes to version bounds are needed, but otherwise looks great

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/NewEpoch.hs
Soupstraw: You probably meant 
```haskell
  if eNo /= addEpochInterval eNoL (EpochIntterval 1)
```
I think it's also possible to do
```haskell
  if eNo /= succ eNoL
```

lehins: Yeah, thanks for the correction. Downsides of not having a compiler in github comments :grin: 

`succ` would work too. I just checked and it does have an `Enum` derived. 

Usually I'd oppose using `succ`/`pred`, since those are partial, but here overflowing would be even worse. However, it is impossible possible, so either of the approaches will be good.

eras/shelley/test-suite/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Version history for `cardano-ledger-shelley-test`
 
+## 1.3.0.3
+
+*
+
```
lehins: `1.3.0.2` hasn't been released yet. The rule of thumb is to not add a new section in the changelog on any PR, except the post release PR that does this for all packages that were released
```suggestion
```

lehins (approved):
Great job, Daniel!

Issue #4001, opened 2024-01-22
Newly elected CC members are removed together with other unrecognized CC members


mkoura:
It should be possible for new potential CC members to authorize their hot keys, and the hot keys should stay authorized once the member gets elected.
It is not the case, the authorized key is removed once the "update committee" proposal that added the member gets ratified.

1. Authorize hot keys for two new potential CC members. The members are shown with status `Unrecognized`, and their hot keys have tag `MemberAuthorized`:

```json
{
    "keyHash-b996f3e6e53884a2aafa98ede0a23cb7200542092846033c26f5b18c": {
      "expiration": null,
      "hotCredsAuthStatus": {
        "contents": {
          "keyHash": "44fb79e80bfe48184165ab213ffeacbe5828b6ff6f2a2097f38c16da"
        },
        "tag": "MemberAuthorized"
      },
      "nextEpochChange": "NoChangeExpected",
      "status": "Unrecognized"
    },
    "keyHash-ea6bf31dff5e45a809b98bff7a157b178f429b230b34f4e988861bc0": {
      "expiration": null,
      "hotCredsAuthStatus": {
        "contents": {
          "keyHash": "7f8d2df7ac1080b6e4c6cbb2f28ad2740e728ef2eefa2f7c76987f59"
        },
        "tag": "MemberAuthorized"
      },
      "nextEpochChange": "NoChangeExpected",
      "status": "Unrecognized"
    }
}
```

2. Propose a "update committee" action that adds one of the new potential CC members.
3. Elect the new CC members.
4. Once the proposal is ratified, check that all `Unrecognized` CC members were removed, both the newly elected and the non-elected ones. Both CC members listed above can no longer be queried.
5. Wait until the proposed action gets enacted.
6. Check that the newly elected CC member was added and can be queried. However, its hot key is missing and tag is `MemberNotAuthorized`. Therefore he needs to authorize his hot key again.

```json
{
    "keyHash-b996f3e6e53884a2aafa98ede0a23cb7200542092846033c26f5b18c": {
      "expiration": 5,
      "hotCredsAuthStatus": {
        "tag": "MemberNotAuthorized"
      },
      "nextEpochChange": "NoChangeExpected",
      "status": "Active"
    },
}
```

Full committee state query outputs and governance state dumps: [ledger_issue_cc_members.tar.gz](https://github.com/IntersectMBO/cardano-ledger/files/14016002/ledger_issue_cc_members.tar.gz)

[...]

PR #4088, opened 2024-02-16
Committee query improvements

teodanciu:
# Description

This was initially meant to be a test to verify the behavior reported in this issue: https://github.com/IntersectMBO/cardano-ledger/issues/4001
However  I realized that we could improve the query by returning  results independent of whether there exists a current committee, since it's possible that we have relevant information even when there is no currently elected committee (for example: information about the next epoch committee, or about pre-authorized members). 
It's possible that returning better results helps with clarifying the behavior reported in the bug, for which I added a test in the last commit. 

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/libs/cardano-ledger-api/src/Cardano/Ledger/Api/State/Query.hs
+++ b/libs/cardano-ledger-api/src/Cardano/Ledger/Api/State/Query.hs
@@ -37,7 +37,6 @@ module Cardano.Ledger.Api.State.Query (
   NextEpochChange (..),
 
   -- * For testing
-  getCommitteeMembers,
   getNextEpochCommitteeMembers,
 ) where
 
@@ -48,7 +47,7 @@ import Cardano.Ledger.Api.State.Query.CommitteeMembersState (
   MemberStatus (..),
   NextEpochChange (..),
  )
-import Cardano.Ledger.BaseTypes (EpochNo (EpochNo), UnitInterval, binOpEpochNo, strictMaybeToMaybe)
+import Cardano.Ledger.BaseTypes (EpochNo (EpochNo), binOpEpochNo, strictMaybeToMaybe)
 import Cardano.Ledger.CertState
 import Cardano.Ledger.Coin (Coin)
 import Cardano.Ledger.Compactible (fromCompact)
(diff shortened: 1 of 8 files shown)


libs/cardano-ledger-api/src/Cardano/Ledger/Api/State/Query.hs
```diff
@@ -184,87 +177,96 @@ queryCommitteeMembersState ::
...
   AccountState
 queryAccountState = view $ nesEsL . esAccountStateL
+
+getNextEpochCommitteeMembers ::
+  ConwayEraGov era =>
+  NewEpochState era ->
+  Map (Credential 'ColdCommitteeRole (EraCrypto era)) EpochNo
+getNextEpochCommitteeMembers nes =
+  let ratifyState = snd . finishDRepPulser $ queryGovState nes ^. drepPulsingStateGovStateL
+      committee = strictMaybeToMaybe $ ratifyState ^. rsEnactStateL . ensCommitteeL
```
lehins: Correct me if I am wrong, but I don't think this conversion is necessary
```suggestion
      committee = ratifyState ^. rsEnactStateL . ensCommitteeL
```

libs/cardano-ledger-api/src/Cardano/Ledger/Api/State/Query.hs
```diff
@@ -184,87 +177,96 @@ queryCommitteeMembersState ::
...
-        csCommitteeCreds $
-          nes ^. nesEpochStateL . esLStateL . lsCertStateL . certVStateL . vsCommitteeStateL
-
-      withFilteredColdCreds s
-        | Set.null coldCredsFilter = s
-        | otherwise = s `Set.intersection` coldCredsFilter
+  CommitteeMembersState (EraCrypto era)
+queryCommitteeMembersState coldCredsFilter hotCredsFilter statusFilter nes =
+  let
+    committee = strictMaybeToMaybe $ queryGovState nes ^. committeeGovStateL
```
lehins: Probably unnecessary 
```suggestion
    committee = queryGovState nes ^. committeeGovStateL
```

lehins (approved):
Looks great. Thank you!

libs/cardano-ledger-api/src/Cardano/Ledger/Api/State/Query.hs
teodanciu: Done, thanks!

PR #4096, opened 2024-02-19
Add a collection of test Plutus scripts for use in ImpTests

### LINKED ISSUE #4027: Make a collection of test scripts available for every Plutus Version
We have a tool named `plutus-preprocessor` that is setup to generate Plutus scripts.

We need to expend its functionality to generate a matching script for every Plutus version.

We should start by creating a data type in a new module `cardano-ledger-core:testlib:Test.Cardano.Ledger.Plutus.Examples`:

```haskell
data PlutusArgs = PlutusArgs
  { paData :: Data
  , paSpendDatum :: Maybe Data
  }

data PlutusExample (l :: Language) = PlutusExample 
  { peScript :: Plutus l
  , peValidArgs :: PlutusArgs
  -- ^ arguments that when applied to the script will make it pass validation
  , peInvalidArgs :: PlutusArgs
  -- ^ arguments that when applied to the script will make it fail validation
  }

class PlutusExamples (l :: Language) = PlutusExamples
  { peOddRedeemer :: PlutusExample l
  , ...
```

Further we would use plutus-preprocessor to generate as many modules as there are known Plutus versions that contain identical scripts compiled for different versions of Plutus

Then `Test.Cardano.Ledger.Plutus.Examples` would import them and construct three bindings:

```haskell
examplesPlutusV1 :: PlutusExamples 'PlutusV1
examplesPlutusV1 = ...

examplesPlutusV2 :: PlutusExamples 'PlutusV2
examplesPlutusV2 = ...

examplesPlutusV3 :: PlutusExamples 'PlutusV3
examplesPlutusV3 = ...
```

We could then use those examples for writing more interesting tests that involve Plutus scripts.

Soupstraw:
# Description

This PR improves the Plutus scripts fixup pipeline so that whoever is writing the tests doesn't have to worry about adding the correct datums and redeemers when using one of the provided example scripts. The test writer can still add invalid datums and redeemers that would make the transaction fail and the fixup function will respect that.

resolves #4027

# Checklist

- [x] Self-reviewed the diff


eras/alonzo/impl/testlib/Test/Cardano/Ledger/Alonzo/Imp/UtxosSpec.hs
```diff
@@ -1,66 +1,77 @@
...
  )
 import Test.Cardano.Ledger.Common
 import Test.Cardano.Ledger.Core.Utils (txInAt)
-import Test.Cardano.Ledger.Plutus (ScriptTestContext (..))
 import Test.Cardano.Ledger.Plutus.Examples (guessTheNumber3)
 
+submitProducingTx :: forall era. ShelleyEraImp era => ImpTestM era (Tx era)
+submitProducingTx =
+  submitTxAnn "Sumbit a transaction with a script output" $
+    mkBasicTx mkBasicTxBody
```
lehins, line 44: The whole reason why this tx is submitted is so the txout is placed into utxo, so it makes sense returning the TxIn instead of Tx
```suggestion
    fmap (txInAt (0 :: Int)) $ mkBasicTx mkBasicTxBody
```

libs/plutus-preprocessor/src/Main.hs
```diff
@@ -58,105 +60,138 @@ manylines h n ts' = write (split ts')
...
+    display outh scriptBytesFun scriptArgs name
+
+-- hPutStrLn outh $
+--  unlines
+--    [ "\n -- List of all " <> show plutusVersion <> " scripts"
+--    , "scripts" <> versionSuffix <> " :: [Plutus '" <> show plutusVersion <> "]"
+--    , "scripts" <> versionSuffix <> " = "
+--    , "  [ " <> (intercalate "\n  , " $ (\(_, _, name) -> name <> versionSuffix) <$> scripts)
+--    , "  ]"
+--    ]
```
lehins: Looks like dead code
```suggestion
```

libs/plutus-preprocessor/src/Main.hs
```diff
@@ -14,38 +15,39 @@ module Main where
...
+
+-- let
+--   makeLanguageBinary lang bytes = do
+--     hPutStr h $
+--       concat
+--         [ "Plutus . PlutusBinary . pack $ concat\n"
+--         , "  ["
+--         ]
+--     manylines h 15 (unpack bytes)
+-- undefined
```
lehins: Dead code
```suggestion
```

libs/cardano-ledger-core/testlib/Test/Cardano/Ledger/Plutus/Examples.hs
```diff
@@ -1,118 +1,170 @@
...
+alwaysSucceeds2 :: SLanguage l -> Plutus l
+alwaysSucceeds2 =
+  Plutus
+    . PlutusBinary
+    . pack
+    . ( \case
+          SPlutusV1 -> [71, 1, 0, 0, 34, 18, 0, 17]
+          SPlutusV2 -> [71, 1, 0, 0, 34, 18, 0, 17]
+          SPlutusV3 -> [70, 1, 1, 0, 34, 128, 1]
+      )
```
lehins, line 26: Very nice. I really like what you did here with script generation and picking the correct script with plutus language version.

eras/shelley/impl/testlib/Test/Cardano/Ledger/Shelley/ImpTest.hs
```diff
@@ -338,6 +334,8 @@ class
 
   fixupTx :: Tx era -> ImpTestM era (Tx era)
 
+  scriptTestContexts :: Map (ScriptHash (EraCrypto era)) ScriptTestContext
```
lehins: I think it would make sense to create `class ShelleyEraImp era => AlonzoEraImp era` in `cardano-ledger-alonzo:testlib` and put that function in that type class.

eras/alonzo/impl/testlib/Test/Cardano/Ledger/Alonzo/ImpTest.hs
```diff
@@ -161,13 +195,13 @@ fixupRedeemers tx = do
     newRedeemers = Map.fromList $ mkNewRedeemers <$> contexts
```
lehins: We need to make sure we do not overwrite the already specified redeemers
```suggestion
    newRedeemers = (Map.fromList $ mkNewRedeemers <$> contexts) Map.\\ oldRedeemers
```

It probably would be better to do this exclusion in `impGetPlutusContexts`, but it'll work here as well.

eras/alonzo/impl/testlib/Test/Cardano/Ledger/Alonzo/Imp/UtxosSpec.hs
```diff
@@ -1,66 +1,77 @@
...
-                  (inject (Coin 100))
-                  & dataHashTxOutL .~ SJust (hashData @era $ Data spendDatum)
-              )
+    tx0 <- submitProducingTx
     submitTxAnn_ "Submit a transaction that consumes the script output" $
       mkBasicTx mkBasicTxBody
         & bodyTxL . inputsTxBodyL
           .~ Set.singleton (txInAt (0 :: Int) tx0)
+  it "Invalid plutus script fails in phase 2" $
+    do
```
lehins: Let's stick with the convention of no lonely `do`s on one line :smile:  
```suggestion
  it "Invalid plutus script fails in phase 2" $ do
```

eras/alonzo/impl/testlib/Test/Cardano/Ledger/Alonzo/Imp/UtxosSpec.hs
```diff
@@ -1,66 +1,77 @@
...
+      tx0 <- impAnn "Submitting producing transaction" submitProducingTx
```
lehins: It already has an annotation, I don't think more is needed
```suggestion
      tx0 <- submitProducingTx
```

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/ImpTest.hs
```diff
@@ -202,6 +203,11 @@ instance
 
   fixupTx = alonzoFixupTx
 
+  scriptTestContexts =
+    plutusTestScripts SPlutusV1
+      <> plutusTestScripts SPlutusV2
```
lehins, line 211: I think there might be a problem with his approach, since  PlutusV1 and PlutusV2 scripts are identical, their hash will be identical, which will result in TestContexts for PlutusV2 to disappear. 

That'll probably work out for now, but might become a problem in the future. So, I don't suggest changing the implementation right now, but adding a comment on the `scriptTestContexts` would be useful for the future that this how it works right now.

eras/alonzo/impl/testlib/Test/Cardano/Ledger/Alonzo/ImpTest.hs
```diff
@@ -12,23 +12,35 @@
...
 import Cardano.Ledger.Address (Addr (..))
 import Cardano.Ledger.Alonzo (AlonzoEra)
-import Cardano.Ledger.Alonzo.Core (AlonzoEraScript (..), AlonzoEraTxOut (..), AsIxItem, Era (..), EraGov)
-import Cardano.Ledger.Alonzo.PParams (AlonzoEraPParams, getLanguageView, ppCostModelsL, ppMaxTxExUnitsL, ppMaxValSizeL)
+import Cardano.Ledger.Alonzo.Core (
+  AlonzoEraScript (..),
+  AlonzoEraTxOut (..),
+  AsIxItem,
+  Era (..),
+ )
```
lehins: I suggest having an open Core import, this will avoid a whole bunch of redundant imports, thus reduce the amount if work we need to do when dealing with imports and resolving conflicts when working on the same module
```suggestion
import Cardano.Ledger.Alonzo.Core
```

lehins (approved):
This looks great. Only some suggestion, but nothing critical.

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/ImpTest.hs
Soupstraw: It's probably better to make an issue for this, otherwise nobody is going to notice this. I think it'd make sense to have the key be `(ScriptHash (EraCrypto era), Language)`

eras/alonzo/impl/testlib/Test/Cardano/Ledger/Alonzo/ImpTest.hs
Soupstraw: I don't think it's necessary, I already accounted for `union` being left-biased

PR #4102, opened 2024-02-20
fix `prop_DELEG` in STS tests

MaximilianAlgehed:
# Description

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/V2/Conway/DELEG.hs
+++ b/libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/V2/Conway/DELEG.hs
@@ -36,14 +36,20 @@ delegCertSpec ::
 delegCertSpec pp ds =
   let rewardMap = unUnify $ rewards ds
       delegMap = unUnify $ delegations ds
+      depositOf k =
+        case fromCompact . rdDeposit <$> Map.lookup k rewardMap of
+          Just d | d > 0 -> SJust d
+          _              -> SNothing
    in constrained $ \dc ->
         (caseOn dc)
           -- ConwayRegCert !(StakeCredential c) !(StrictMaybe Coin)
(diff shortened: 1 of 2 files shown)


MaximilianAlgehed:
@lehins here is another flaky test - what does it mean when it fails with a test being "pending"

lehins:
> here is another flaky test - what does it mean when it fails with a test being "pending"

Pending is not a failure. "pending" tests are the ones that have been disabled, because they are not essential and there is some work needed to get them to pass.

Failure CI is due to some byron test failure that I have not seen before:
```
1 out of 51 tests failed (98.82s)
Test suite byron-spec-ledger-test: FAIL
```

But, we don't need to worry about it, since byron code base is mostly read only.

MaximilianAlgehed:
@lehins by earlier logic, shouldn't we turn the tests off as they would be in the way of auto-merge here? Slowing down merging on PRs.

lehins:
> by earlier logic, shouldn't we turn the tests off as they would be in the way of auto-merge here? Slowing down merging on PRs.

The fact that I have never seen it before, means it is indeed an extremely rare failure and is not worth the time. That being said, please make sure to not introduce any new tests like that.

MaximilianAlgehed:
@lehins same failure as here iirc:
https://github.com/IntersectMBO/cardano-ledger/pull/3730

lehins:
> same failure as here iirc:

Please don't make me dig through the logs. Could you please paste the output of the test failure for the one from #3730 and this PR. This will tell us if it is indeed the same.

MaximilianAlgehed:
Unfortunately the logs aren't saved from old CI runs.

> Looks like it, but it is the first time I see this failure, but this is from byron codebase and this is not something we are going to fix. So, no need to worry about it

Anyway, I'm just making sure I understand the standards of this code base and how you want things done.

lehins (approved):
:+1:

PR #4104, opened 2024-02-20
Governance policy tests

### LINKED ISSUE #4089: ImpTest Governance Policy is executed
There are two GovActions that can trigger governance policy: TreasuryWithdrawal and ParameterChange

This ticket tracks theses tests:

* [ ] - Submitting a proposal with those GovActions in presence of alwaysSucceeding governance policy successfully executes the governance policy script and accepts the transaction as valid
* [ ] - Submitting a proposal with those GovActions in presence of alwaysFailing governance policy successfully fails execution of the governance policy script and accepts transaction as invalid (i.e. collects the collateral)
* [ ] - Submitting a proposal with either of those two GovActions with present Governance Policy that is not a PlutusV3 script results in a predicate failure

teodanciu:
# Description

Closes https://github.com/IntersectMBO/cardano-ledger/issues/4089

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/GovSpec.hs
+++ b/eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/GovSpec.hs
@@ -44,6 +44,7 @@ import qualified Data.Sequence.Strict as SSeq
 import qualified Data.Set as Set
 import Data.Tree
 import Lens.Micro
+import Test.Cardano.Ledger.Conway.Arbitrary ()
 import Test.Cardano.Ledger.Conway.ImpTest
 import Test.Cardano.Ledger.Core.KeyPair (mkAddr)
 import Test.Cardano.Ledger.Core.Rational (IsRatio (..))
@@ -628,58 +629,6 @@ spec =
           passNEpochs 3
           fmap (!! 3) getProposalsForest
             `shouldReturn` Node (SJust p116) []
-    describe "Proposals always have valid previous actions" $ do
-      context "Invalid proposals are rejected" $ do
-        it "Invalid Index in GovPurposeId" $ do
-          gaidConstitutionProp <- submitInitConstitutionGovAction
-          curConstitution <- getsNES $ newEpochStateGovStateL . constitutionGovStateL
-          constitutionHash <- freshSafeHash
-          let constitutionActionNext =
-                NewConstitution
-                  (SJust $ GovPurposeId gaidConstitutionProp)
-                  ( Constitution
-                      ( Anchor
-                          (fromJust $ textToUrl 64 "constitution.1")
-                          constitutionHash
-                      )
-                      SNothing
-                  )
-          curConstitution' <- getsNES $ newEpochStateGovStateL . constitutionGovStateL
-          impAnn "Constitution has not been enacted yet" $
-            curConstitution' `shouldBe` curConstitution
-          submitGovAction_ constitutionActionNext
-        it "Enact Constitution and use valid GovPurposeId" $ do
-          (dRep, committeeMember, _) <- electBasicCommittee
-          constitutionHash <- freshSafeHash
-          let constitution =
-                Constitution
-                  ( Anchor
-                      (fromJust $ textToUrl 64 "constitution.0")
-                      constitutionHash
-                  )
-                  SNothing
-              constitutionAction =
-                NewConstitution SNothing constitution
-          gaidConstitutionProp <- submitGovAction constitutionAction
-          submitYesVote_ (DRepVoter dRep) gaidConstitutionProp
-          submitYesVote_ (CommitteeVoter committeeMember) gaidConstitutionProp
-          passEpoch
-          passEpoch
-          curConstitution <- getsNES $ newEpochStateGovStateL . constitutionGovStateL
-          impAnn "Constitution has been enacted" $
-            curConstitution `shouldBe` constitution
-          constitutionHash1 <- freshSafeHash
-          let constitutionAction1 =
-                NewConstitution
-                  (SJust $ GovPurposeId gaidConstitutionProp)
-                  ( Constitution
-                      ( Anchor
-                          (fromJust $ textToUrl 64 "constitution.1")
-                          constitutionHash1
-                      )
-                      SNothing
-                  )
-          submitGovAction_ constitutionAction1
 
     describe "Voting" $ do
       context "fails for" $ do
@@ -851,13 +800,19 @@ spec =
         (pp ^. ppMinFeeAL) `shouldBe` newMinFeeA
 
     describe "Constitution proposals" $ do
-      context "accepted for" $
+      context "accepted for" $ do
         it "empty PrevGovId before the first constitution is enacted" $ do
           --  Initial proposal does not need a GovPurposeId but after it is enacted, the
           --  following ones are not
           _ <- submitConstitution SNothing
           -- Until the first proposal is enacted all proposals with empty GovPurposeIds are valid
           void $ submitConstitution SNothing
+        it "valid GovPurposeId" $ do
+          (dRep, committeeMember, _) <- electBasicCommittee
+          constitution <- arbitrary
+          gaidConstitutionProp <- enactConstitution SNothing constitution dRep committeeMember
+          constitution1 <- arbitrary
+          void $ enactConstitution (SJust $ GovPurposeId gaidConstitutionProp) constitution1 dRep committeeMember
 
       context "rejected for" $ do
         it "empty PrevGovId after the first constitution was enacted" $ do
@@ -866,7 +821,7 @@ spec =
           submitYesVote_ (DRepVoter dRep) govActionId
           submitYesVote_ (CommitteeVoter committeeMember) govActionId
           passNEpochs 2
-          constitution <- newConstitution
+          constitution <- arbitrary
           let invalidNewConstitutionGovAction =
                 NewConstitution
                   SNothing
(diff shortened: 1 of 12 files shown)


eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/ImpTest.hs
```diff
@@ -1041,3 +1051,52 @@ constitutionShouldBe cUrl = do
     getsNES $
       nesEsL . esLStateL . lsUTxOStateL . utxosGovStateL . constitutionGovStateL
   anchorUrl `shouldBe` fromJust (textToUrl 64 $ T.pack cUrl)
+
+setConstitutionScript :: ConwayEraImp era => ScriptHash (EraCrypto era) -> ImpTestM era ()
+setConstitutionScript sh =
+  modifyNES $
+    nesEsL . esLStateL . lsUTxOStateL . utxosGovStateL . constitutionGovStateL
+      .~ Constitution def (SJust sh)
```
lehins: State should not be modified manually. Constitution should be updated with a proposal.

I know that you just extracted the implementation that was already there, but could you please fix that. We should resort to NewEpochState modification only for things that are either too complicated to change by methods that would normally be used on mainnet or not yet supported by ImpTest

Also, instead of `def` generate the hash with `arbitrary`, eg:
```haskell
  constitutionHash <- arbitrary
  pure $ Constitution constitutionHash (SJust sh)
```

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/GovSpec.hs
```diff
@@ -1703,3 +1641,36 @@ ccVoteOnConstitutionFailsWithMultipleVotes = do
...
+mismatchedGovPolicies ::
+  forall era.
+  ( HasCallStack
+  , ConwayEraImp era
+  , InjectRuleFailure "LEDGER" ConwayGovPredFailure era
+  ) =>
+  GovAction era ->
+  ScriptHash (EraCrypto era) ->
+  ImpTestM era ()
+mismatchedGovPolicies govAction scriptHash = do
```
lehins: Combine these two into one test function, since there is dependency between. (Dependency will come from constitution modification which needs to be modified using a proposal)

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/UtxosSpec.hs
```diff
@@ -187,3 +180,70 @@ spec =
               CollectErrors
                 [BadTranslation . inject . InlineDatumsNotSupported @era $ TxOutFromInput lockedTxIn]
           ]
+
+    describe "Plutus gov policies for ParameterChange and TreasuryWithdrawal" $ do
```
lehins: Could you please also add a test that checks that governance policy can be a native script?

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/UtxosSpec.hs
```diff
@@ -187,3 +180,70 @@ spec =
...
+  ) =>
+  GovAction era ->
+  ImpTestM era ()
+alwaysFailsGovPolicyTxAcceptedInvalid govAction = do
+  let plutus = alwaysFailsPlutus @'PlutusV3 0
+  setConstitutionScript $ hashPlutusScript plutus
+  proposal <- proposalWithPlutusGovPolicy govAction plutus
+  let tx =
+        mkBasicTx mkBasicTxBody
+          & bodyTxL . proposalProceduresTxBodyL .~ GHC.fromList [proposal]
```
lehins: Why not use `OverloadedLists`?
```suggestion
          & bodyTxL . proposalProceduresTxBodyL .~ [proposal]
```

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/UtxosSpec.hs
```diff
@@ -187,3 +180,70 @@ spec =
...
+
+alwaysFailsGovPolicyTxAcceptedInvalid ::
+  forall era.
+  ( HasCallStack
+  , ConwayEraImp era
+  ) =>
+  GovAction era ->
+  ImpTestM era ()
+alwaysFailsGovPolicyTxAcceptedInvalid govAction = do
+  let plutus = alwaysFailsPlutus @'PlutusV3 0
```
lehins: It should have two arguments
```suggestion
  let plutus = alwaysFailsPlutus @'PlutusV3 2
```

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/ImpTest.hs
```diff
@@ -1041,3 +1051,52 @@ constitutionShouldBe cUrl = do
...
+  let govAction' = case govAction of
+        ParameterChange gid ppup _ -> ParameterChange gid ppup (SJust sh)
+        TreasuryWithdrawals ws _ -> TreasuryWithdrawals ws (SJust sh)
+        _ -> error "Gov Policy tests are only applicable to ParameterChange and TreasuryWithdrawals actions"
+  pure $
+    ProposalProcedure
+      { pProcReturnAddr = rewardAccount
+      , pProcGovAction = govAction'
+      , pProcDeposit = pp ^. ppGovActionDepositL
+      , pProcAnchor = def
```
lehins: Conversion of `GovAction` to `ProposalProcedure` is a duplicate functionality. `submitGovAction` will already do that.

Also, overwriting values is an anti pattern. It is better to construct GovAction with the desired script hash right off the bat.

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/UtxosSpec.hs
```diff
@@ -187,3 +180,70 @@ spec =
...
+alwaysSucceedsGovPolicyTxAcceptedValid ::
+  forall era.
+  (HasCallStack, ConwayEraImp era) =>
+  GovAction era ->
+  ImpTestM era ()
+alwaysSucceedsGovPolicyTxAcceptedValid govAction = do
+  let plutus = alwaysSucceedsPlutus @'PlutusV3 3
+  setConstitutionScript $ hashPlutusScript plutus
+  proposal <- proposalWithPlutusGovPolicy govAction plutus
+  submitProposal_ proposal
```
lehins: Only spend scripts have 3 arguments. Also the workflow should be adjusted:
* we don't want to overwrite the script hash, we can just accept a closure, eg `ParameterChange SNothing pparamsUpdate . SJust`. This will avoid using `error` in `error "Gov Policy tests are only...`
* We should be getting the ScriptHash by adding the script to the ImpTestState, this ensures we can use it
* we can use the regular `submitGovAction` function, no need to add extra helpers
```suggestion
alwaysSucceedsGovPolicyTxAcceptedValid ::
  (HasCallStack, ConwayEraImp era) =>
  (SciptHash (EraCrypto ers) -> GovAction era) ->
  ImpTestM era ()
alwaysSucceedsGovPolicyTxAcceptedValid mkGovAction = do
  scriptHash <-
    impAddPlutusScript $
      ScriptTestContext
        (alwaysSucceedsPlutus @'PlutusV3 2)
        PlutusArgs
          { paSpendDatum = Nothing
          , paData = P1.I 3
          }
  setConstitutionScript scriptHash
  submitGovAction_ $ mkGovAction scriptHash
```

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/UtxosSpec.hs
```diff
@@ -187,3 +180,70 @@ spec =
...
+  ImpTestM era ()
+alwaysFailsGovPolicyTxAcceptedInvalid govAction = do
+  let plutus = alwaysFailsPlutus @'PlutusV3 0
+  setConstitutionScript $ hashPlutusScript plutus
+  proposal <- proposalWithPlutusGovPolicy govAction plutus
+  let tx =
+        mkBasicTx mkBasicTxBody
+          & bodyTxL . proposalProceduresTxBodyL .~ GHC.fromList [proposal]
+  res <- trySubmitTx tx
+  void $ expectLeft res
```
lehins: Why aren't we checking the expected predicate failure?

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/ImpTest.hs
```diff
@@ -1041,3 +1051,52 @@ constitutionShouldBe cUrl = do
...
+proposalWithPlutusGovPolicy govAction plutus = do
+  sh <-
+    impAddPlutusScript $
+      ScriptTestContext
+        plutus
+        PlutusArgs
+          { paSpendDatum = Nothing
+          , paData = P1.I 3
+          }
+  proposalWithGovPolicy govAction sh
```
lehins: This is not a valid function, since it is not clear will this be a valid proposal or not, which itself will depend on the current constitution. It should be written in the opposite direction. Submit a proposal either with the scripthash that matches the current constitution or with an explicitly mistmatched scripthash. This will make the intent clear at the test site.

lehins (changes requested):
Implemented tests to check more or less what we need, but the way they are written can be improved. Let me know if my comments aren't clear I'll be happy to meet and talk about it.

lehins (approved):
Very nice!!

PR #4105, opened 2024-02-21
Update `RELEASING` with revisioning instructions

Lucsanszky:
# Description

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -172,7 +172,7 @@ CHaP. (TODO: implement a script that lists all of the package that fit the above
 #### Release to CHaP
 
 1. Follow the [CHaP release
-   instructions](https://github.com/input-output-hk/cardano-haskell-packages#-from-github)
+   instructions](https://github.com/intersectmbo/cardano-haskell-packages#-from-github)
 
    For example, to release commit with SHA `deadbeef`:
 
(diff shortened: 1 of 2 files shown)


Lucsanszky:
I will also add a section with general instructions to sum things up after the more specific step-by-step instructions and perhaps add the actual `git` commands in the general setup section but this is roughly it. Hopefully it is clear and detailed enough.

docs/RevisionHowto.md
```diff
@@ -0,0 +1,161 @@
+# Step-by-step example
```
teodanciu: I think it would be useful to add a few details here: 
* what exactly is  the example demonstrating ?    (is it adding revisions for those libraries from cardano-base? 
* under what circumstanced would we need to do this? 
* what is the meaning of the commits mentioned in the code examples, eg: why  checkout  commit `4fab1fe2`? 

Also I don't understand exactly what the relationship of this paragraph is with the one at the end, `# Summary of steps to take when adding and testing revisions`  , is the latter a more abstract way of describing it , whereas the former is a concrete example? If so, maybe we can change the order, so the example would be maybe easier to follow if we knew what we are trying to achieve?

RELEASING.md
teodanciu, line 1: I would add it at the top-level, just because it seems to belong with the others, RELEASING.md,  CONTRIBUTING.md,  etc., which are sort of  "meta" - docs.    The `docs` folder  appears to be more about business logic, and though I don't think it was pre-planned carefully, as things are structured now,  in my opinion it would be more natural place at the top level.

docs/RevisionHowto.md
teodanciu, line 1: I would add it at the top-level, just because it seems to belong with the others, RELEASING.md,  CONTRIBUTING.md,  etc., which are sort of  "meta" - docs.    The `docs` folder  appears to be more about business logic, and though I don't think it was pre-planned carefully, as things are structured now,  in my opinion it would be more natural place at the top level.

teodanciu: And in that case I would maybe name it REVISIONS.md, for consistency with its siblings

teodanciu:
These are great details, thank you for writing them down so thoroughly! 
I just had a suggestion about switching the sections: starting with the overview of what the process is and when we need to do it, and then following with the example  (maybe also, if possible,  naming the specifics of the example, eg: creating a revision for this package at that commit),  just so it's easier to understand for someone who is not familiar with revisions (like me :grimacing: )

docs/RevisionHowto.md
Lucsanszky: Thanks for the detailed review! I addressed your suggestions, hopefully it is now a bit easier to read and more clear. I will add a few more things to the `Building with Nix` instructions (more info on `.#allSmokeTestPackages` and will probably replace that with the exact packages as that obviously builds much faster).

docs/RevisionHowto.md
Lucsanszky: Good idea! I will name it `REVISIONING.md` though as that is more consistent with the naming convention (and also because `REVISIONS.md` could be misleading as it could imply that it lists revisions for some reason rather than discussing the way to do revisions).

teodanciu (approved):
Looks good to me, awesome!

PR #4113, opened 2024-02-24
Add new functionality to constrained generators

TimSheard:
Added newtypes Size, SizeSpec, and class Sized to system.
Added HasSpec Size instance
Added operations And and Or to BoolFn

This PR tries to recover lost changes to Branch ts-new-constraints when Tim's computer failed.
Functionality is now in place, but the lost tests need to be recreated.

# Description

# Checklist

- [x] Self-reviewed the diff


libs/cardano-ledger-test/src/Test/Cardano/Ledger/EraClass.hs
MaximilianAlgehed, line 1: Does this file belong in this PR? @lehins is in a better position to comment on what's going on here than me.

libs/constrained-generators/src/Constrained/Base.hs
```diff
@@ -611,6 +652,11 @@ class
   --   ∀ a. a `conformsTo` spec == a `conformsTo` constrained (\t -> toPreds t spec)
   toPreds :: Term fn a -> TypeSpec fn a -> Pred fn
 
+  -- | compute a SizeSpec, that bounds the size of what (genFromTypeSpec spec) can compute.
```
MaximilianAlgehed, line 655: This is confusing. Does that mean the number of elements in the set of possible values, or does it mean the size of the values themselves? The latter assumes that every type has a size, which is a major change to the model that I would expect at least shows up in some constraint somewhere. 

If it's the latter I would prefer this to be in it's own type class to avoid baking assumptions about types into `HasSpec`.

libs/constrained-generators/src/Constrained/Base.hs
```diff
@@ -2772,128 +2805,125 @@ toPredsNumSpec v (NumSpecInterval ml mu) =
...
   emptySpec = emptyNumSpec
   combineSpec = combineNumSpec
   genFromTypeSpec = genFromNumSpec
   conformsTo = conformsToNumSpec
   toPreds = toPredsNumSpec
+  sizeOfTypeSpec = mempty
 
 ------------------------------------------------------------------------
 -- Function universes
 ------------------------------------------------------------------------
```
MaximilianAlgehed, line 2926: Why was this moved elsewhere?

libs/constrained-generators/src/Constrained/Base.hs
```diff
@@ -382,45 +407,47 @@ data HOLE a b where
 
 toCtx ::
```
MaximilianAlgehed, line 408: Adding the string argument here is not how the `MonadGenError` is meant to be used. The argument `termstring` appears to always be `show` on the `Term` argument at the callsites. If this is the case, then why not just add a show constraint? If this is not the case, consider adding the extra logging information in a call to `explain` at the _call site_ of `toCtx`.

Either way, I would strongly prefer the use site to be wrapped in `explain` instead of this solution, as it pushes information down in the functions and makes them more complicated.

libs/constrained-generators/src/Constrained/Base.hs
```diff
@@ -562,6 +589,20 @@ typeSpec ts = TypeSpec ts mempty
 
 -- The HasSpec Class ------------------------------------------------------
 
+-- | sizeOfSpec generalizes the method (sizeOfTypeSpec :: TypeSpec fn t -> SizeSpec) to (sizeOfSpec:: Spec fn t -> SizeSpec)
```
MaximilianAlgehed, line 592: Can we move this function elsewhere? It would be nice if the `The HasSpec Class` banner is above the `HasSpec` class implementation.

libs/constrained-generators/src/Constrained/Base.hs
MaximilianAlgehed: C.f. `Forallable` and `OrdLike` for inspiration on the way to do this.

libs/constrained-generators/src/Constrained/Base.hs
```diff
@@ -1504,14 +1548,14 @@ letFloating = fold . go []
...
+aggressiveInlining inlineable prd
   | inlined = aggressiveInlining inlineable (simplifyPred pInlined)
-  | otherwise = p
+  | otherwise = prd
   where
-    (pInlined, Any inlined) = runWriter $ go (freeVars p `restrictedTo` Set.fromList inlineable) [] p
+    (pInlined, Any inlined) = runWriter $ go (freeVars prd `restrictedTo` Set.fromList inlineable) [] prd
 
-    underBinder fvs x p = fvs `without` [Name x] <> freeVars p `restrictedTo` (Set.singleton $ Name x)
+    underBinder fvs x ps = fvs `without` [Name x] <> freeVars ps `restrictedTo` (Set.singleton $ Name x)
```
MaximilianAlgehed, line 1558: Why this name change?

libs/constrained-generators/src/Constrained/Base.hs
MaximilianAlgehed, line 592: This docstring doesn't explain what this function does.

libs/constrained-generators/src/Constrained/Base.hs
```diff
@@ -2383,29 +2420,16 @@ instance (Ord a, HasSpec fn a) => Semigroup (SetSpec fn a) where
...
-sizeSpec (ErrorSpec es) = ErrorSpec es
-sizeSpec (TypeSpec (SetSpec must elemSpec sizeSpec) cant) =
-  cantSpec
-    <> sizeSpec
-    <> geqSpec @fn (Set.size must)
-    <> specSizeBound elemSpec
+-- NOTE: No guarantees that this is a tight spec.
+
+-- | Like specSizeBound, but slightly more accurate, since it can analyze the 'cant' part of TypeSpec for Set
+sizeOfSet :: forall fn a. (Ord a, BaseUniverse fn) => Spec fn (Set a) -> Spec fn Size
```
MaximilianAlgehed, line 2426: Above it looks like `specSizeBound` computes the size of the generator, i.e. how many different possible sets can be generated. However, this function used to compute how large generated sets could be.

libs/constrained-generators/src/Constrained/Base.hs
```diff
@@ -562,6 +589,20 @@ typeSpec ts = TypeSpec ts mempty
 
 -- The HasSpec Class ------------------------------------------------------
 
+-- | sizeOfSpec generalizes the method (sizeOfTypeSpec :: TypeSpec fn t -> SizeSpec) to (sizeOfSpec:: Spec fn t -> SizeSpec)
+--   It is not unusual for instances of sizeOfTypeSpec to call sizeOfSpec, some (TypeSpec fn t)'s contain (Spec fn s)
+sizeOfSpec :: forall fn t. Spec fn t -> SizeSpec
+sizeOfSpec TrueSpec = mempty
+sizeOfSpec (MemberSpec xs) = atMostInt (length xs)
```
MaximilianAlgehed, line 596: Why is this `atMost` and not `exactInt (nub $ length xs)`?

libs/constrained-generators/src/Constrained/Base.hs
```diff
@@ -2419,26 +2443,32 @@ instance (Ord a, HasSpec fn a) => HasSpec fn (Set a) where
...
       ]
 
   genFromTypeSpec (SetSpec must e TrueSpec) = (must <>) . Set.fromList <$> listOfT (genFromSpec e)
   genFromTypeSpec ss@(SetSpec must e _) = do
     -- Compute more constrained size spec based on must and elem spec!
-    n <- explain ["Generate set size"] $ genFromSpec (sizeSpec $ typeSpec ss) `suchThatT` (>= Set.size must)
-    go (n - Set.size must) must
+    (Size n) <- explain ["Generate set size"] $ genFromSpec (sizeOfSet $ typeSpec ss) `suchThatT` (>= setSize must)
+    let (Size m) = setSize must
+    go (Size (n - m)) must
```
MaximilianAlgehed, line 2456: If `go` pattern matches on the size constructor everywhere, can't you just call it with `n-m` instead and reduce the code complexity?

libs/constrained-generators/src/Constrained/Base.hs
```diff
@@ -2419,26 +2443,32 @@ instance (Ord a, HasSpec fn a) => HasSpec fn (Set a) where
...
-      go 0 s = pure s
-      go n s = do
-        e <- explain ["generate set member"] $ withMode Strict $ genFromSpec e `suchThatT` (`Set.notMember` s)
-        go (n - 1) (Set.insert e s)
+      go (Size 0) s = pure s
+      go (Size n) s = do
+        ex <- explain ["generate set member"] $ withMode Strict $ genFromSpec e `suchThatT` (`Set.notMember` s)
+        go (Size (n - 1)) (Set.insert ex s)
+
+  toPreds s (SetSpec m es _size)
```
MaximilianAlgehed, line 2463: What happened to `satisfies (size_ s) size`?

libs/constrained-generators/src/Constrained/Base.hs
```diff
@@ -2460,15 +2490,15 @@ instance IsUniverse fn => Functions (SetFn fn) fn where
                     (mapList (\(Value a) -> Lit a) pre)
                     (x' :> mapList (\(Value a) -> Lit a) suf)
              in Let (App (injectFn fn) args) (x :-> p)
-    Size | NilCtx HOLE <- ctx -> typeSpec $ SetSpec mempty TrueSpec spec
+    --    SetSize | NilCtx HOLE <- ctx -> typeSpec $ SetSpec mempty TrueSpec spec
```
MaximilianAlgehed, line 2493: Where did `size_` go?

libs/constrained-generators/src/Constrained/Base.hs
```diff
@@ -2575,19 +2608,19 @@ instance IsUniverse fn => Functions (SetFn fn) fn where
...
 -- List -------------------------------------------------------------------
 
-data ListSpec fn a = ListSpec [a] (Spec fn Int) (Spec fn a) (FoldSpec fn a)
+data ListSpec fn a = ListSpec [a] (Spec fn Size) (Spec fn a) (FoldSpec fn a)
 
 deriving instance Show (FoldSpec fn a)
 deriving instance HasSpec fn a => Show (ListSpec fn a)
 
 instance HasSpec fn a => HasSpec fn [a] where
+  sizeOfTypeSpec (ListSpec xs sz elemS _) =
```
MaximilianAlgehed, line 2620: What is the specification for `sizeOfTypeSpec`?? Here we are mixing the length of the list and the _size_ of the spec for the length of the list?

libs/constrained-generators/src/Constrained/Base.hs
```diff
@@ -2619,23 +2652,23 @@ instance HasSpec fn a => HasSpec fn [a] where
       <> toPredsFoldSpec x foldS
       <> satisfies (length_ x) size
 
-sizeSpecList :: forall fn a. Spec fn [a] -> Spec fn Int
+sizeSpecList :: forall fn a. BaseUniverse fn => Spec fn [a] -> Spec fn Size
 sizeSpecList TrueSpec = TrueSpec
-sizeSpecList (MemberSpec es) = MemberSpec (length <$> es)
+sizeSpecList (MemberSpec es) = exactSizeSpec (listSize es)
```
MaximilianAlgehed, line 2657: This is not the same semantics.

libs/constrained-generators/src/Constrained/Base.hs
```diff
@@ -3189,7 +3219,7 @@ class (Num a, HasSpec fn a) => NumLike fn a where
     Maybe a
   safeSubtract a b = fromSimpleRep <$> safeSubtract @fn @(SimpleRep a) (toSimpleRep a) (toSimpleRep b)
 
-instance NumLike fn a => Num (Term fn a) where
+instance {-# OVERLAPPABLE #-} NumLike fn a => Num (Term fn a) where
```
MaximilianAlgehed, line 3222: What? Why is this necessary?

libs/constrained-generators/src/Constrained/Base.hs
```diff
@@ -3643,3 +3679,262 @@ instance HasSpec fn a => Pretty (Spec fn a) where
...
+
+-- ==================================================================
+-- Size, Sized, and SizeSpec
+-- ==================================================================
+
+newtype Size = Size {unSize :: Int} deriving (Eq, Ord, Num)
+
+instance Show Size where show (Size n) = "(Size " ++ show n ++ ")"
+
+instance {-# INCOHERENT #-} BaseUniverse fn => OrdLike fn Size where
```
MaximilianAlgehed, line 3691: Why does this need to be `INCOHERENT`?

libs/constrained-generators/src/Constrained/Base.hs
```diff
@@ -3643,3 +3679,262 @@ instance HasSpec fn a => Pretty (Spec fn a) where
...
+  type TypeSpec fn Size = SizeSpec
+  type Prerequisites fn Size = ()
+  emptySpec = SizeSpec mempty
+  combineSpec s1 s2 = TypeSpec (s1 <> s2) []
+  genFromTypeSpec = genFromSizeSpec deltaMax
+  conformsTo = conformsToSize
+  toPreds x szspec@(SizeSpec _) = toPredsSizeSpec x szspec
+  sizeOfTypeSpec x = x
+
+instance {-# INCOHERENT #-} BaseUniverse fn => Num (Term fn Size) where
```
MaximilianAlgehed, line 3711: This instance suggests that perhaps it would be better for the size functions to just return an `Int`? It would be easier to work with, produce less complex code, and look more like native Haskell.

libs/constrained-generators/src/Constrained/Base.hs
```diff
@@ -3643,3 +3679,262 @@ instance HasSpec fn a => Pretty (Spec fn a) where
...
+
+-- | SizeSpec forms a Semigroup. It inherits from the Semigroup instance of (NumSpec Word64)
+instance Semigroup SizeSpec where
+  SizeSpec x <> SizeSpec y = SizeSpec (x <> y)
+
+-- | SizeSpec forms a Monoid
+instance Monoid SizeSpec where
+  mempty = SizeSpec mempty
+
+toPredsSizeSpec ::
```
MaximilianAlgehed, line 3778: This should already exist for `NumSpec`, why do we need this again?

libs/constrained-generators/src/Constrained/Base.hs
```diff
@@ -3643,3 +3679,262 @@ instance HasSpec fn a => Pretty (Spec fn a) where
...
+-- | When we generate from an interval like [ x .. infinity ]. What is the practical bound we want
+--   on infinity? So we answer this using deltaMax.
+--   When we generate [ x .. infinity ] we use [ x .. x + deltaMax ]
+--   This creates a uniform distribution between x and (x + deltaMax). If we just choose an arbitray
+--   value greater than 'x', values are not uniform. Just the way arbitrary works on Int.
+deltaMax :: Int
+deltaMax = 10
+
+-- | Gen a random Size, Because Sizes are aways >= 0, raise a failure if that is the case.
+genFromSizeSpec :: MonadGenError m => Int -> SizeSpec -> GenT m Size
```
MaximilianAlgehed, line 3797: This should already exist.

libs/constrained-generators/src/Constrained/Base.hs
```diff
@@ -3643,3 +3679,262 @@ instance HasSpec fn a => Pretty (Spec fn a) where
...
+  | x >= 0 =
+      Size <$> pureGen (choose (x, x + delta))
+genFromSizeSpec _ (SizeSpec interval) =
+  fail ("genFromSizeSpec fails because the interval " ++ show interval ++ " has only negative choices.")
+
+-- =============================================================================
+-- Size Spec acts like an interval, and we can implement interval arithmetic
+-- We are careful not to apply subtraction if it leads to a negative choice.
+
+addM :: (a -> a -> a) -> Maybe a -> Maybe a -> Maybe a
```
MaximilianAlgehed, line 3819: addH f m n = f <$> m <*> n

libs/constrained-generators/src/Constrained/Base.hs
```diff
@@ -3643,3 +3679,262 @@ instance HasSpec fn a => Pretty (Spec fn a) where
...
+  propagateSpecFun _ _ (ErrorSpec err) = ErrorSpec err
+  propagateSpecFun fn (ListCtx pre HOLE suf) (SuspendedSpec x p) =
+    constrained $ \x' ->
+      let args = appendList (mapList (\(Value a) -> Lit a) pre) (x' :> mapList (\(Value a) -> Lit a) suf)
+       in Let (App (injectFn fn) args) (x :-> p)
+  propagateSpecFun SizeOf (NilCtx HOLE) _spec = liftSizeSpec (SizeSpec mempty)
+  propagateSpecFun Plus (HOLE :? Value (s :: Size) :> Nil) spec = minusSpec spec s
+  propagateSpecFun Plus (Value (s :: Size) :! NilCtx HOLE) spec = minusSpec spec s
+  propagateSpecFun Minus (HOLE :? Value (s :: Size) :> Nil) spec = plusSpec s spec
+  propagateSpecFun Minus (Value (s :: Size) :! NilCtx HOLE) spec = plusSpec s spec
```
MaximilianAlgehed, line 3909: These two cases can't be the same.

libs/constrained-generators/src/Constrained/Base.hs
```diff
@@ -3643,3 +3679,262 @@ instance HasSpec fn a => Pretty (Spec fn a) where
...
+--   if the subtraction returns a neagative number.
+minusSpec :: BaseUniverse fn => Spec fn Size -> Size -> Spec fn Size
+minusSpec (TypeSpec spec@(SizeSpec _) ss) sz@(Size w) =
+  case subtractionM spec (exactInt w) of
+    Left msg -> ErrorSpec [msg]
+    Right ans -> TypeSpec ans (map ((-) sz) ss) -- TODO is this right?
+minusSpec (MemberSpec ss) sz = MemberSpec (map ((-) sz) ss)
+minusSpec spec _ = error ("minusSpec, should be handled in propagateSpec " ++ show spec)
+
+{-
```
MaximilianAlgehed, line 3932: Remove this comment.

libs/constrained-generators/src/Constrained/Instances.hs
```diff
@@ -60,12 +60,32 @@ instance IsUniverse fn => Functions (EqFn fn) fn where
...
+  (True, True) -> MemberSpec [True]
+  (True, False) -> MemberSpec [False]
+  (False, False) -> TrueSpec
+  (False, True) -> ErrorSpec [" (and_ " ++ show constant ++ " HOLE) must equal True. That cannot be the case."]
+
+-- | We have something like ('constant' || HOLE) must evaluate to 'need'. Return a (Spec fn Bool) for HOLE, that makes that True.
+okOr :: Bool -> Bool -> Spec fn Bool
+okOr constant need = case (constant, need) of
+  (True, True) -> TrueSpec
+  (True, False) -> MemberSpec [False]
```
MaximilianAlgehed, line 89: This should be error spec

libs/constrained-generators/src/Constrained/Instances.hs
```diff
@@ -60,12 +60,32 @@ instance IsUniverse fn => Functions (EqFn fn) fn where
...
+  (False, False) -> MemberSpec [False]
+  (False, True) -> ErrorSpec [" (or_ " ++ show constant ++ " HOLE) must equal True. That cannot be the case."]
```
MaximilianAlgehed, line 91: This should be const True

libs/constrained-generators/src/Constrained/Spec/Maps.hs
```diff
@@ -161,7 +171,7 @@ instance (Member (MapFn fn) fn, IsUniverse fn) => Functions (MapFn fn) fn where
           , Evidence <- prerequisites @fn @(Map k v) ->
               case spec of
                 MemberSpec [s] ->
-                  typeSpec $ MapSpec s [] (equalSpec $ Set.size s) TrueSpec NoFold
+                  typeSpec $ MapSpec s [] (exactSizeSpec $ setSize s) TrueSpec NoFold
```
MaximilianAlgehed, line 174: Why not `equalSpec`??

libs/constrained-generators/src/Constrained/Test.hs
```diff
@@ -461,13 +462,15 @@ notSubset :: Spec Fn (Set Int)
...
 unionSized :: Spec Fn (Set Int)
 unionSized = constrained $ \s ->
-  10 ==. size_ (s <> lit (Set.fromList [1 .. 8]))
+  Lit (MkSize 10) ==. sizeOf_ @(Set Int) @Fn (s <> lit (Set.fromList [1 .. 8]))
+-}
 
 listSum :: Numbery a => Spec Fn [a]
 listSum = constrained $ \as ->
-  10 <=. sum_ as
+  Lit 10 <=. sum_ as
```
MaximilianAlgehed, line 473: Needing `Lit` here is evidence that the interface has gotten worse. This is probably caused by the `OVERLAPPABLE` and `INCOHERENT` instances for `Num Term`, I would prefer if we didn't do this.

libs/constrained-generators/src/Constrained/Univ.hs
```diff
@@ -270,16 +281,12 @@ elemFn = injectFn $ Elem @_ @fn
...
-sizeFn = injectFn $ Size @_ @fn
-
 data SetFn (fn :: [Type] -> Type -> Type) args res where
   Subset :: Ord a => SetFn fn '[Set a, Set a] Bool
   Disjoint :: Ord a => SetFn fn '[Set a, Set a] Bool
   Member :: Ord a => SetFn fn '[a, Set a] Bool
   Singleton :: Ord a => SetFn fn '[a] (Set a)
-  Union :: Ord a => SetFn fn '[Set a, Set a] (Set a)
-  Size :: Ord a => SetFn fn '[Set a] Int
+  Union :: Ord a => SetFn fn '[Set a, Set a] (Set a) -- HELP ME, I don't think this is usefull
```
MaximilianAlgehed, line 289: Why would being able to specify things about the union not be useful?

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/V2/Conway.hs
```diff
@@ -114,11 +114,11 @@ import Constrained qualified as C
 type ConwayUnivFns = StringFn : DefaultFns
 type ConwayFn = Fix (OneofL ConwayUnivFns)
 
-type IsConwayUniv fn =
+type ConwayUniverse fn =
```
MaximilianAlgehed, line 117: Can we have the changes to names, renaming, etc. in a separate PR. There are many things that need to be resolved about size and having a monster PR that touches everything in the codebase makes that very difficult.

libs/constrained-generators/src/Constrained/Base.hs
```diff
@@ -3286,11 +3316,25 @@ fromGeneric_ ::
 fromGeneric_ = app fromGenericFn
 
 not_ ::
-  IsUniverse fn =>
+  BaseUniverse fn =>
   Term fn Bool ->
   Term fn Bool
 not_ = app notFn
 
+and_ ::
```
MaximilianAlgehed, line 3324: Why not `(&&.)` and `(||.)` to be consistent with the rest of the interface?

MaximilianAlgehed (changes requested):
A few things:
(1) It would be nice if this were four different PRs, one for the size stuff, one for all the renaming etc., one for the boolean stuff, and one for the `EraClass` stuff that I don't quite understand the reason for.
(2) I don't see what's gained from this `Size` exercise (esp. since I think at the moment it breaks things).

libs/cardano-ledger-test/src/Test/Cardano/Ledger/EraClass.hs
lehins: In my opinion this whole file should be removed, but Tim likes to have it in the test suite for some reason. I don't have time to argue about its uselessness, so I choose not to argue about it.

lehins:
Closing in favor of #4121

PR #4114, opened 2024-02-24
Imp Bootstrap address support

lehins:
# Description

This PR has a few important changes to ImpSpec:

* Fix `arbitrary` going into an infinite loop
* Bootstrap address fixup support
* Avoid overwriting witnesses. Instead only missing key witnesses are added.
* Hashes and keys are generated at random using QuickCheck, instead of relying on an index.
* Collateral is now injected on demand, instead of a fixed number upon initialization
* Migration of a couple of Bootstrap address witness tests from `cardano-ledger-shelley-test`
# Checklist

- [x] Self-reviewed the diff


eras/shelley/impl/testlib/Test/Cardano/Ledger/Shelley/ImpTest.hs
```diff
@@ -761,22 +761,19 @@ fixupFees tx = do
...
     changeBeforeFeeTxOut =
       mkBasicTxOut
         (mkAddr (kpSpending, kpStaking))
         changeBeforeFee
   let
-    txNoWits =
-      txWithRoot & bodyTxL . outputsTxBodyL %~ (:|> changeBeforeFeeTxOut)
-    outsBeforeFee = txWithRoot ^. bodyTxL . outputsTxBodyL
+    txNoWits = tx & bodyTxL . outputsTxBodyL %~ (:|> changeBeforeFeeTxOut)
+    outsBeforeFee = tx ^. bodyTxL . outputsTxBodyL
```
teodanciu, line 813: I can see it's correct to remove this (because i tested it myself out of disbelief), but I can't explain why this is the case! The root is in the UTXO, so how come it's all still balancing out?

lehins: Because there is now a separate step that does just that with `addRootTxIn` 
https://github.com/IntersectMBO/cardano-ledger/blob/bf878d234657c50c4bdc748d324460b92d05a522/eras/shelley/impl/testlib/Test/Cardano/Ledger/Shelley/ImpTest.hs#L803

teodanciu: Oh right, I missed this. Thanks!

teodanciu (approved):
Looks good to me (I admit, the keypair generation logic eludes me in some places though)

libs/cardano-ledger-core/testlib/Test/Cardano/Ledger/Core/KeyPair.hs
```diff
@@ -173,13 +181,18 @@ mkKeyHash :: Crypto c => Int -> KeyHash kd c
...
+mkKeyPair = mkKeyPairWithSeed . Plain.serialize'
+
+mkKeyPairWithSeed :: forall r c. Crypto c => BS.ByteString -> KeyPair r c
+mkKeyPairWithSeed inputSeed = KeyPair vk sk
   where
+    seedSize = fromIntegral (DSIGN.seedSizeDSIGN (Proxy @(DSIGN c)))
+    seed
+      | BS.length inputSeed /= seedSize =
+          hashToBytes $ Hash.hashWith @Hash.Blake2b_256 id inputSeed
+      | otherwise = inputSeed
```
Soupstraw: This should probably be extracted into a separate function as it's also used in `mkBootKeyPairWithSeed`

lehins: Oh yeah. Good point. They were added at different times, so I didn't notice the logic is exactly the same

eras/shelley/impl/testlib/Test/Cardano/Ledger/Shelley/ImpTest.hs
```diff
@@ -988,89 +1022,145 @@ withImpState ::
...
-  impFreshIdxL += 1
-  pure $ hashKey kh
+freshKeyHash = fst <$> freshKeyPair
+
+-- | Generate a random KeyPair and add it to the known keys in the Imp state
+freshKeyPair ::
+  (Era era, MonadState (ImpTestState era) m, MonadGen m) =>
+  m (KeyHash r (EraCrypto era), KeyPair r (EraCrypto era))
+freshKeyPair = do
+  keyPair <- arbitrary
```
Soupstraw, line 1101: Is this generated uniformly?

Soupstraw (approved):
Awesome!

eras/shelley/impl/testlib/Test/Cardano/Ledger/Shelley/ImpTest.hs
lehins: yes

libs/cardano-ledger-core/testlib/Test/Cardano/Ledger/Core/KeyPair.hs
lehins: Done

PR #4115, opened 2024-02-25
Switch to using the correct stake pool distribution for voting

### LINKED ISSUE #4108: Stake distribution updates seem to take 4 epochs to reach ratification only for SPO votes
In #4097 , there is a test that demonstrates that, for SPO votes, stake distribution updates (by adding UTxOs) as well as updates to the rewards, take 4 epochs to reach ratification, as opposed to 2 epochs for DRep votes.

- [x] Investigate why there is a difference in the number of epochs needed to ratify
- [ ] ~Either document that this is expected behavior for the future or fix it.~

lehins:
# Description

Fixes #4108 

We were using stake pool distribution that was an epoch too old

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/eras/conway/impl/CHANGELOG.md
+++ b/eras/conway/impl/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## 1.13.0.0
 
+* Switch `EPOCH` rule environment back to `()`. Start using the latest stake pool
+  distribution: #4115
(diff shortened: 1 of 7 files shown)


aniketd (approved):
Awesome! :raised_hands:

PR #4111, opened 2024-02-22
Bump Plutus deps to 1.22.0.0

Unisay:
This PR updates Plutus dependencies versions to 1.22.0.0

Related: https://github.com/IntersectMBO/cardano-haskell-packages/pull/669

PR #4117, opened 2024-02-26
Use plutus-ledger-api 1.22.1

Unisay:
This PR updates Plutus dependencies versions to 1.22.1.0

Related: https://github.com/IntersectMBO/cardano-haskell-packages/pull/681
Supersedes: https://github.com/IntersectMBO/cardano-ledger/pull/4111


eras/alonzo/test-suite/src/Test/Cardano/Ledger/Alonzo/Translation/TranslationInstance.hs
```diff
@@ -73,6 +73,9 @@ instance Cborg.Serialise PV1.TxInInfo
 instance Cborg.Serialise PV1.TxInfo
 instance Cborg.Serialise PV1.TxOut
 instance Cborg.Serialise PV2.ScriptPurpose
+instance Cborg.Serialise PV2.TxId
+instance Cborg.Serialise PV2.TxOutRef
+instance Cborg.Serialise PV2.TxInInfo
```
lehins, line 78: This is strange, aren't these the new types that were defined for PlutusV3? How come they are exported from `PlutusLedgerApi.V2`?

lehins (commented):
Looking good.

This guy also need to be bumped to 233:
https://github.com/IntersectMBO/cardano-ledger/blob/7009e29ae0a9adb472bc7123b96348aa4ce7dcba/eras/conway/impl/cddl-files/conway.cddl#L507

eras/alonzo/test-suite/src/Test/Cardano/Ledger/Alonzo/Translation/TranslationInstance.hs
Unisay: These types existed before in V2, and V3 was using them, but not anymore. Now V3 has its own copies of these types, with the only difference being that copies use V3.TxId instead of the V2.TxId

lehins (approved):
Looks great. Thank you!

PR #4119, opened 2024-02-26
NewConstraints phase1. Add BoolFn And and Or and tests

TimSheard:
Add And and Or to BoolFn
complete the necessary instances
add the test mapSizeConstrained  as a baseline for future work.
add a test for (&&.) and (||.)

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/libs/constrained-generators/src/Constrained/Base.hs
+++ b/libs/constrained-generators/src/Constrained/Base.hs
@@ -3291,6 +3291,22 @@ not_ ::
   Term fn Bool
 not_ = app notFn
 
+infixr 3 &&.
+(&&.) ::
+  IsUniverse fn =>
+  Term fn Bool ->
+  Term fn Bool ->
+  Term fn Bool
+(&&.) = app andFn
+
+infixr 2 ||.
+(||.) ::
+  IsUniverse fn =>
+  Term fn Bool ->
+  Term fn Bool ->
+  Term fn Bool
+(||.) = app orFn
+
 elem_ ::
   forall a fn.
   HasSpec fn a =>
--- a/libs/constrained-generators/src/Constrained/Instances.hs
+++ b/libs/constrained-generators/src/Constrained/Instances.hs
@@ -67,5 +67,25 @@ instance IsUniverse fn => Functions (BoolFn fn) fn where
       let args = appendList (mapList (\(Value a) -> Lit a) pre) (v' :> mapList (\(Value a) -> Lit a) suf)
        in Let (App (injectFn fn) args) (v :-> ps)
   propagateSpecFun Not (NilCtx HOLE) spec = caseBoolSpec spec (equalSpec . not)
+  propagateSpecFun And (HOLE :? Value (s :: Bool) :> Nil) spec = caseBoolSpec spec (okAnd s)
+  propagateSpecFun And (Value (s :: Bool) :! NilCtx HOLE) spec = caseBoolSpec spec (okAnd s)
+  propagateSpecFun Or (HOLE :? Value (s :: Bool) :> Nil) spec = caseBoolSpec spec (okOr s)
+  propagateSpecFun Or (Value (s :: Bool) :! NilCtx HOLE) spec = caseBoolSpec spec (okOr s)
 
   mapTypeSpec Not _ = typeSpec ()
+
+-- | We have something like ('constant' &&. HOLE) must evaluate to 'need'. Return a (Spec fn Bool) for HOLE, that makes that True.
+okAnd :: Bool -> Bool -> Spec fn Bool
+okAnd constant need = case (constant, need) of
+  (True, True) -> MemberSpec [True]
+  (True, False) -> MemberSpec [False]
+  (False, False) -> TrueSpec
+  (False, True) -> ErrorSpec ["(" ++ show constant ++ " &&. HOLE) must equal True. That cannot be the case."]
+
+-- | We have something like ('constant' ||. HOLE) must evaluate to 'need'. Return a (Spec fn Bool) for HOLE, that makes that True.
+okOr :: Bool -> Bool -> Spec fn Bool
+okOr constant need = case (constant, need) of
+  (True, True) -> TrueSpec
+  (True, False) -> ErrorSpec ["(" ++ show constant ++ "||. HOLE) must equal False. That cannot be the case."]
+  (False, False) -> MemberSpec [False]
+  (False, True) -> MemberSpec [True]
--- a/libs/constrained-generators/src/Constrained/Test.hs
+++ b/libs/constrained-generators/src/Constrained/Test.hs
@@ -153,6 +153,8 @@ tests =
     , testSpec "parallelLetPair" parallelLetPair
     , testSpec "mapSizeConstrained" mapSizeConstrained
     , numberyTests
+    , testSpec "andPair" andPair
+    , testSpec "orPair" orPair
     ]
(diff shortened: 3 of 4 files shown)


libs/constrained-generators/src/Constrained/Base.hs
```diff
@@ -3291,6 +3291,20 @@ not_ ::
   Term fn Bool
 not_ = app notFn
 
+and_ ::
```
MaximilianAlgehed: rename `and_` and `or_` `(&&.)` and `(||.)` with the proper precedence to maintain coherence with the rest of the library.

libs/constrained-generators/src/Constrained/Instances.hs
```diff
@@ -67,5 +67,25 @@ instance IsUniverse fn => Functions (BoolFn fn) fn where
...
   propagateSpecFun Not (NilCtx HOLE) spec = caseBoolSpec spec (equalSpec . not)
+  propagateSpecFun And (HOLE :? Value (s :: Bool) :> Nil) spec = caseBoolSpec spec (okAnd s)
+  propagateSpecFun And (Value (s :: Bool) :! NilCtx HOLE) spec = caseBoolSpec spec (okAnd s)
+  propagateSpecFun Or (HOLE :? Value (s :: Bool) :> Nil) spec = caseBoolSpec spec (okOr s)
+  propagateSpecFun Or (Value (s :: Bool) :! NilCtx HOLE) spec = caseBoolSpec spec (okOr s)
 
   mapTypeSpec Not _ = typeSpec ()
+
+-- | We have something like ('constant' &&. HOLE) must evaluate to 'need'. Return a (Spec fn Bool) for HOLE, that makes that True.
+okAnd :: Bool -> Bool -> Spec fn Bool
```
MaximilianAlgehed, line 78: I would put these in a `were` clause to keep the top level interface of the module clean, but that's just a style thing for me.

libs/constrained-generators/src/Constrained/Instances.hs
```diff
@@ -67,5 +67,25 @@ instance IsUniverse fn => Functions (BoolFn fn) fn where
...
+okAnd constant need = case (constant, need) of
+  (True, True) -> MemberSpec [True]
+  (True, False) -> MemberSpec [False]
+  (False, False) -> TrueSpec
+  (False, True) -> ErrorSpec [" (and_ " ++ show constant ++ " HOLE) must equal True. That cannot be the case."]
+
+-- | We have something like ('constant' ||. HOLE) must evaluate to 'need'. Return a (Spec fn Bool) for HOLE, that makes that True.
+okOr :: Bool -> Bool -> Spec fn Bool
+okOr constant need = case (constant, need) of
```
MaximilianAlgehed, line 87: This is not quite correct, c.f. my comments on #4113.

libs/constrained-generators/src/Constrained/Test.hs
```diff
@@ -547,3 +550,22 @@ parallelLetPair = constrained $ \p ->
...
+    or_ (x <=. 5) (y <=. 5)
+
+data Three = One | Two | Three
+  deriving (Ord, Eq, Show, Generic)
+
+instance HasSimpleRep Three
+instance IsUniverse fn => HasSpec fn Three
+
+mapSizeConstrained :: Spec Fn (Map Three Int)
+mapSizeConstrained = constrained $ \m -> size_ (dom_ m) <=. 3
```
MaximilianAlgehed: Remove these from this PR as they are in another PR that shouldn't interfere with this.

libs/constrained-generators/src/Constrained/Test.hs
```diff
@@ -152,6 +152,9 @@ tests =
     , testSpec "listSumPair" (listSumPair @Int)
     , testSpec "parallelLetPair" parallelLetPair
     , numberyTests
+    , testSpec "(And pair (x,y) test -- both (x <=. 5) and (y <=. 1)" andPair
+    , testSpec "(Or  pair (x,y) test -- at least one of (x <=. 5) or (y <=. 5)" orPair
+    , testSpec "size test" mapSizeConstrained
```
MaximilianAlgehed: remove this

MaximilianAlgehed (changes requested):

libs/constrained-generators/src/Constrained/Base.hs
TimSheard: Good idea. It is done

libs/constrained-generators/src/Constrained/Instances.hs
TimSheard: OK, I made some changes. Please check that I got it right.

libs/constrained-generators/src/Constrained/Test.hs
TimSheard: So I based this OR on   PR-map-size-test, so we would have this baseline test at the beginning of the 3 phase (PRS) that lead up to the whole shebang. Isn't that what you wanted?

libs/constrained-generators/src/Constrained/Test.hs
TimSheard: same as conversation above. Don't we want the baseline tests here?

libs/constrained-generators/src/Constrained/Instances.hs
TimSheard: It is hard to get these function in the two scopes where the are called, and still make it clear that propagateSpecFun exhaustively covers all cases, if they are not at the top level. So I plan to leave them where they are, But we could make a more explicit statement of what is exported at the module level. Let me know if you want to do that.

libs/constrained-generators/src/Constrained/Test.hs
MaximilianAlgehed: That works too.

libs/constrained-generators/src/Constrained/Test.hs
MaximilianAlgehed: I think this is fine, if we change the `"size test"` to be `"mapSizeConstrained"` to ensure minimal risk of collision with the other PR.

libs/constrained-generators/src/Constrained/Test.hs
```diff
@@ -152,6 +152,9 @@ tests =
     , testSpec "listSumPair" (listSumPair @Int)
     , testSpec "parallelLetPair" parallelLetPair
     , numberyTests
+    , testSpec "(And pair (x,y) test -- both (x <=. 5) and (y <=. 1)" andPair
+    , testSpec "(Or  pair (x,y) test -- at least one of (x <=. 5) or (y <=. 5)" orPair
```
MaximilianAlgehed: I would prefer we rename these two tests to follow the naming scheme of the rest of the file.

However, I like the idea of having information about what the test contains be reported on failure. I'll add a `counterexample` for the `spec` to the `prop_sound` property so that way we get the same effect there.

MaximilianAlgehed (changes requested):

libs/constrained-generators/src/Constrained/Instances.hs
MaximilianAlgehed: No this is ok, no big deal at all.

MaximilianAlgehed (approved):

libs/constrained-generators/src/Constrained/Test.hs

libs/constrained-generators/src/Constrained/Test.hs

PR #4120, opened 2024-02-27
Newconstaints phase2 rename IsUniverse (BaseUniverse), Fn (BaseFn)

TimSheard:
This is a follow pr based on top of PR #4119 
It renames a few types having to do with the Functions universe, to make explicit that these types
refer to the objects defined in the Base.

# Checklist

- [ ] Self-reviewed the diff


libs/constrained-generators/src/Constrained/Test.hs
```diff
@@ -152,17 +152,20 @@ tests =
     , testSpec "listSumPair" (listSumPair @Int)
     , testSpec "parallelLetPair" parallelLetPair
     , numberyTests
+    , testSpec "(And pair (x,y) test -- both (x <=. 5) and (y <=. 1)" andPair
```
MaximilianAlgehed: Male sure to keep the change of the names here from the other PR

MaximilianAlgehed (approved):

libs/constrained-generators/src/Constrained/Test.hs

PR #4121, opened 2024-02-27
Newconstraints phase3, Add newtypes: Size, SizeSpec and class Sized.

TimSheard:
This is a follow on PR. Originally rebased on top of PR #4120, it is now rebased on the current master, and 4120 has been closed.

Spec does 3 things.
1) Adds generic size functions on numerous types with HasSpec instances, by adding Sized instances. Current
    instances include Int, Set, List, Map, and Size itself.
2) Retracts the And opertor from BoolFn, which turned out to have some technical difficulties.
3) Adds the methods 'cardinalTypeSpec' and 'cardinalTrueSpec',  these allow any HasSpec instance to
    use the function  cardinality :: HasSpec fn t => Spec fn t -> Spec fn Int.
   It computes bounds on the total number of possible solutions generated by 'genFromTypeSpec'
   This supports tighter (more accurate) bounds in a number of places.

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/libs/cardano-ledger-test/src/Test/Cardano/Ledger/STS.hs
+++ b/libs/cardano-ledger-test/src/Test/Cardano/Ledger/STS.hs
@@ -291,3 +291,40 @@ utxoTests =
     , testProperty "prop_UTXOW" prop_UTXOW
     , testProperty "prop_UTXOS" prop_UTXOS
     ]
+
+epoch :: TestTree
+epoch =
+  testGroup
+    "STS property tests"
+    [ govTests
+    , testProperty "prop_EPOCH" prop_EPOCH
+    ]
+
+{-
+zzz :: IO ()
+zzz = go ms
+
+ms ::Spec BaseFn (Map.Map Word8 [Integer])
+ms = TypeSpec (MapSpec
+        {mapSpecMustKeys = fromList [],
+         mapSpecMustValues = [],
+         mapSpecSize = TrueSpec,
+         mapSpecElem = TrueSpec,
+         mapSpecFold = NoFold}) []
+
+epoch :: TestTree
+epoch =
+  testGroup
+    "STS property tests"
+    [ testProperty "prop_EPOCH" prop_EPOCH
+    ]
+
+predmap :: Spec BaseFn Word8
+predmap = constrained $
+              \k -> toPred
+                  [ assert $ not_ (member_ k (Lit $ Set.fromList [1,2,3,4,5::Word8]))
+                  , satisfies (pair_ k (Lit @Word8 7)) TrueSpec
+                  ]
+
+-- type Conway = ConwayEra StandardCrypto
+-}
--- a/libs/cardano-ledger-test/src/Test/Cardano/Ledger/STS.hs
+++ b/libs/cardano-ledger-test/src/Test/Cardano/Ledger/STS.hs
@@ -299,32 +299,3 @@ epoch =
     [ govTests
     , testProperty "prop_EPOCH" prop_EPOCH
     ]
-
-{-
-zzz :: IO ()
-zzz = go ms
-
-ms ::Spec BaseFn (Map.Map Word8 [Integer])
-ms = TypeSpec (MapSpec
-        {mapSpecMustKeys = fromList [],
-         mapSpecMustValues = [],
-         mapSpecSize = TrueSpec,
-         mapSpecElem = TrueSpec,
-         mapSpecFold = NoFold}) []
-
-epoch :: TestTree
-epoch =
-  testGroup
-    "STS property tests"
-    [ testProperty "prop_EPOCH" prop_EPOCH
-    ]
-
-predmap :: Spec BaseFn Word8
-predmap = constrained $
-              \k -> toPred
-                  [ assert $ not_ (member_ k (Lit $ Set.fromList [1,2,3,4,5::Word8]))
-                  , satisfies (pair_ k (Lit @Word8 7)) TrueSpec
-                  ]
-
--- type Conway = ConwayEra StandardCrypto
--}
--- a/libs/constrained-generators/src/Constrained/Base.hs
+++ b/libs/constrained-generators/src/Constrained/Base.hs
@@ -7,6 +7,7 @@
 {-# LANGUAGE FlexibleInstances #-}
 {-# LANGUAGE FunctionalDependencies #-}
 {-# LANGUAGE GADTs #-}
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
 {-# LANGUAGE ImportQualifiedPost #-}
 {-# LANGUAGE ImpredicativeTypes #-}
 {-# LANGUAGE LambdaCase #-}
@@ -2341,7 +2342,7 @@ genFromFold ::
   , HasSpec fn a
   ) =>
   [a] ->
-  Spec fn Int ->
+  Spec fn Size ->
   Spec fn a ->
   fn '[a] b ->
   Spec fn b ->
@@ -2350,7 +2351,7 @@ genFromFold must size elemS fn foldS = do
   let elemS' = mapSpec fn elemS
       mustVal = adds @fn (map (sem fn) must)
       foldS' = propagateSpecFun (theAddFn @fn) (HOLE :? Value mustVal :> Nil) foldS
-  results <- genList elemS' foldS' `suchThatT` (\xs -> length xs `conformsToSpec` size)
+  results <- genList elemS' foldS' `suchThatT` (\xs -> listSize xs `conformsToSpec` size)
   explain
     [ "genInverse"
     , "  fn = " ++ show fn
@@ -2413,7 +2414,7 @@ deriving instance (HasSpec fn a, HasSpec fn b) => Show (SumSpec fn a b)
 
 -- Sets -------------------------------------------------------------------
 
-data SetSpec fn a = SetSpec (Set a) (Spec fn a) (Spec fn Int)
+data SetSpec fn a = SetSpec (Set a) (Spec fn a) (Spec fn Size)
 
 instance (Ord a, HasSpec fn a) => Semigroup (SetSpec fn a) where
   SetSpec must es size <> SetSpec must' es' size' =
@@ -2423,28 +2424,21 @@ instance (Ord a, HasSpec fn a) => Monoid (SetSpec fn a) where
   mempty = SetSpec mempty mempty TrueSpec
 
 -- NOTE: No guarantees that this is a tight spec!
-sizeSpec :: forall fn a. (Ord a, BaseUniverse fn) => Spec fn (Set a) -> Spec fn Int
+sizeSpec :: forall fn a. (Ord a, BaseUniverse fn) => Spec fn (Set a) -> Spec fn Size
 sizeSpec TrueSpec = TrueSpec
-sizeSpec (MemberSpec es) = MemberSpec (Set.size <$> es)
+sizeSpec (MemberSpec es) = MemberSpec (setSize <$> es)
 sizeSpec SuspendedSpec {} = TrueSpec
 sizeSpec (ErrorSpec es) = ErrorSpec es
-sizeSpec (TypeSpec (SetSpec must elemSpec sizeSpec) cant) =
+sizeSpec (TypeSpec (SetSpec must _elemSpec szSpec) cant) =
   cantSpec
-    <> sizeSpec
-    <> geqSpec @fn (Set.size must)
-    <> specSizeBound elemSpec
+    <> szSpec
+    <> geqSpec @fn (setSize must)
   where
+    --  <> specSizeBound elemSpec  -- I don't see how this can have any effect on the sizeSpec
+
     cantSpec
       | mempty `elem` cant = notEqualSpec 0
       | otherwise = TrueSpec
-
-specSizeBound :: forall fn a. (Eq a, BaseUniverse fn) => Spec fn a -> Spec fn Int
-specSizeBound TrueSpec = TrueSpec
-specSizeBound (MemberSpec es) = leqSpec @fn (length (nub es))
-specSizeBound ErrorSpec {} = equalSpec 0
--- NOTE: these are concessions, there is not really a good way to know
-specSizeBound TypeSpec {} = TrueSpec
-specSizeBound SuspendedSpec {} = TrueSpec
 
 instance (Ord a, HasSpec fn a) => HasSpec fn (Set a) where
   type TypeSpec fn (Set a) = SetSpec fn a
@@ -2458,7 +2452,7 @@ instance (Ord a, HasSpec fn a) => HasSpec fn (Set a) where
 
   conformsTo s (SetSpec must es size) =
     and
-      [ Set.size s `conformsToSpec` size
+      [ setSize s `conformsToSpec` size
       , must `Set.isSubsetOf` s
       , all (`conformsToSpec` es) s
       ]
@@ -2466,8 +2460,8 @@ instance (Ord a, HasSpec fn a) => HasSpec fn (Set a) where
   genFromTypeSpec (SetSpec must e TrueSpec) = (must <>) . Set.fromList <$> listOfT (genFromSpec e)
   genFromTypeSpec ss@(SetSpec must e _) = do
     -- Compute more constrained size spec based on must and elem spec!
-    n <- explain ["Generate set size"] $ genFromSpec (sizeSpec $ typeSpec ss) `suchThatT` (>= Set.size must)
-    go (n - Set.size must) must
+    n <- explain ["Generate set size"] $ genFromSpec (sizeSpec $ typeSpec ss) `suchThatT` (>= setSize must)
+    go (n - setSize must) must
     where
       go 0 s = pure s
       go n s = do
@@ -2499,7 +2493,7 @@ instance BaseUniverse fn => Functions (SetFn fn) fn where
                     (mapList (\(Value a) -> Lit a) pre)
                     (x' :> mapList (\(Value a) -> Lit a) suf)
              in Let (App (injectFn fn) args) (x :-> p)
-    Size | NilCtx HOLE <- ctx -> typeSpec $ SetSpec mempty TrueSpec spec
+    SetSize | NilCtx HOLE <- ctx -> typeSpec $ SetSpec mempty TrueSpec spec
     Singleton
       | NilCtx HOLE <- ctx ->
           let singletons = filter ((1 ==) . Set.size)
@@ -2532,7 +2526,7 @@ instance BaseUniverse fn => Functions (SetFn fn) fn where
                     exists (\eval -> pure $ Set.difference (eval x) s) $ \disjoint ->
                       [ assert $ overlap `subset_` Lit s
                       , assert $ disjoint `disjoint_` Lit s
-                      , satisfies (size_ disjoint + Lit (Set.size s)) size
+                      , satisfies (size_ disjoint + Lit (setSize s)) size
                       , assert $ x ==. overlap <> disjoint
                       ]
             -- TODO: shortcut more cases?
@@ -2544,7 +2538,10 @@ instance BaseUniverse fn => Functions (SetFn fn) fn where
     Subset
       | HOLE :? Value (s :: Set a) :> Nil <- ctx
       , Evidence <- prerequisites @fn @(Set a) -> caseBoolSpec spec $ \case
-          True -> typeSpec $ SetSpec mempty (MemberSpec $ Set.toList s) mempty
+          True ->
+            if Set.null s
+              then MemberSpec [Set.empty]
+              else typeSpec $ SetSpec mempty (MemberSpec $ Set.toList s) mempty
           False -> constrained $ \set ->
             exists (\eval -> headGE $ Set.difference (eval set) s) $ \e ->
               [ set `DependsOn` e
@@ -2614,14 +2611,14 @@ instance BaseUniverse fn => Functions (SetFn fn) fn where
       constrained $ \x ->
         unsafeExists $ \x' ->
           assert (x ==. singleton_ x') <> toPreds x' ts
-    Size ->
+    SetSize ->
       constrained $ \x ->
         unsafeExists $ \x' ->
           assert (x ==. size_ x') <> toPreds x' ts
 
 -- List -------------------------------------------------------------------
 
-data ListSpec fn a = ListSpec [a] (Spec fn Int) (Spec fn a) (FoldSpec fn a)
+data ListSpec fn a = ListSpec [a] (Spec fn Size) (Spec fn a) (FoldSpec fn a)
 
 deriving instance Show (FoldSpec fn a)
 deriving instance HasSpec fn a => Show (ListSpec fn a)
@@ -2648,12 +2645,17 @@ instance HasSpec fn a => HasSpec fn [a] where
         (genFromSpec elemS)
         sz
         ((`conformsToSpec` szSpec) . (+ length must))
+{- REDO consistsent with above
+  genFromTypeSpec ss@(ListSpec must _ elemS NoFold) = do
+    MkSize sz <- genFromSpec (sizeSpecList $ typeSpec ss)
+    lst <- vectorOfT sz $ genFromSpec elemS
+-}
     pureGen $ shuffle (must ++ lst)
   genFromTypeSpec (ListSpec must size elemS (FoldSpec f foldS)) = do
     genFromFold must size elemS f foldS
 
   conformsTo xs (ListSpec must size elemS foldS) =
-    length xs `conformsToSpec` size
+    listSize xs `conformsToSpec` size
       && all (`elem` xs) must
       && all (`conformsToSpec` elemS) xs
       && xs `conformsToFoldSpec` foldS
@@ -2661,22 +2663,22 @@ instance HasSpec fn a => HasSpec fn [a] where
   toPreds x (ListSpec must size elemS foldS) =
     (forAll x $ \x' -> assert (elem_ x' (Lit must)) <> satisfies x' elemS)
       <> toPredsFoldSpec x foldS
-      <> satisfies (length_ x) size
+      <> satisfies (sizeOf_ x) size
 
-sizeSpecList :: forall fn a. Spec fn [a] -> Spec fn Int
+sizeSpecList :: forall fn a. Spec fn [a] -> Spec fn Size
 sizeSpecList TrueSpec = TrueSpec
-sizeSpecList (MemberSpec es) = MemberSpec (length <$> es)
+sizeSpecList (MemberSpec es) = MemberSpec (listSize <$> es)
 sizeSpecList (SuspendedSpec x p) =
   constrained $ \len ->
     Exists
       (\_ -> fatalError ["sizeSpecList: Exists"])
-      (x :-> (Assert [] (len ==. length_ (V x)) <> p))
+      (x :-> (Assert [] (len ==. sizeOf_ (V x)) <> p))
 sizeSpecList (ErrorSpec es) = ErrorSpec es
 sizeSpecList (TypeSpec (ListSpec _ _ ErrorSpec {} _) _) = equalSpec 0
 sizeSpecList (TypeSpec (ListSpec must sizeSpec _ _) cant) =
   cantSpec
     <> sizeSpec
-    <> geqSpec @fn (length must)
+    <> geqSpec @fn (listSize must)
   where
     cantSpec
       | mempty `elem` cant = notEqualSpec 0
@@ -2926,7 +2928,7 @@ instance BaseUniverse fn => HasSpec fn Float where
 -- that is an extension of the Base Universe.
 -----------------------------------------------------------------------------------
 
-type BaseFns = '[ListFn, SumFn, MapFn, SetFn, FunFn, GenericsFn, PairFn, IntFn, BoolFn, EqFn, OrdFn]
+type BaseFns = '[ListFn, SumFn, MapFn, SetFn, FunFn, GenericsFn, PairFn, IntFn, BoolFn, EqFn, OrdFn, SizeFn]
 type BaseFn = Fix (OneofL BaseFns)
 
 -- | A minimal Universe contains functions for a bunch of different things.
@@ -2943,6 +2945,7 @@ type BaseUniverse fn =
   , Member (SumFn fn) fn
   , Member (MapFn fn) fn
   , Member (FunFn fn) fn
+  , Member (SizeFn fn) fn
   )
 
 -- Higher order functions -------------------------------------------------
@@ -3142,9 +3145,9 @@ data ListFn fn args res where
     ) =>
     fn '[a] b ->
     ListFn fn '[[a]] b
-  ListSize :: ListFn fn '[[a]] Int
+  ListSize :: ListFn fn '[[a]] Size
 
-listSizeFn :: forall fn a. Member (ListFn fn) fn => fn '[[a]] Int
+listSizeFn :: forall fn a. Member (ListFn fn) fn => fn '[[a]] Size
 listSizeFn = injectFn $ ListSize @fn
 
 deriving instance Show (ListFn fn args res)
@@ -3158,7 +3161,7 @@ instance Typeable fn => Eq (ListFn fn args res) where
 instance FunctionLike fn => FunctionLike (ListFn fn) where
   sem = \case
     FoldMap f -> adds @fn . map (sem f)
-    ListSize -> length
+    ListSize -> listSize
 
 instance BaseUniverse fn => Functions (ListFn fn) fn where
   propagateSpecFun _ _ (ErrorSpec err) = ErrorSpec err
@@ -3406,14 +3409,24 @@ size_ ::
   forall a fn.
   (HasSpec fn (Set a), Ord a) =>
   Term fn (Set a) ->
-  Term fn Int
+  Term fn Size
 size_ = app sizeFn
+
+union_ ::
+  forall a fn.
+  ( HasSpec fn a
+  , Ord a
+  ) =>
+  Term fn (Set a) ->
+  Term fn (Set a) ->
+  Term fn (Set a)
+union_ = app unionFn
 
 length_ ::
   forall a fn.
   HasSpec fn [a] =>
   Term fn [a] ->
-  Term fn Int
+  Term fn Size
 length_ = app listSizeFn
 
 infix 4 <=., <., ==., /=.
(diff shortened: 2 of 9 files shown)


libs/constrained-generators/src/Constrained/Base.hs
```diff
@@ -2422,30 +2437,30 @@ instance (Ord a, HasSpec fn a) => Semigroup (SetSpec fn a) where
...
-    <> sizeSpec
-    <> geqSpec @fn (Set.size must)
-    <> specSizeBound elemSpec
+    <> szSpec
+    <> geqSpec @fn (setSize must)
+    <> specIntToSize (cardinality elemSpec)
   where
+    -- I don't see how this can have any effect on the sizeSpec
+    -- Please help me understand.
+
```
TimSheard: Max, is this a first version of cardinality? Help me understand line 2455

libs/constrained-generators/src/Constrained/Base.hs
```diff
@@ -2372,6 +2383,8 @@ instance BaseUniverse fn => HasSpec fn Bool where
   _ `conformsTo` _ = True
   genFromTypeSpec _ = pureGen arbitrary
   toPreds _ _ = TruePred
+  cardinalTypeSpec _ = MemberSpec [0]
```
MaximilianAlgehed: This should be `MemberSpec [2]`

libs/constrained-generators/src/Constrained/Base.hs
```diff
@@ -2422,30 +2437,30 @@ instance (Ord a, HasSpec fn a) => Semigroup (SetSpec fn a) where
...
+sizeSpec (TypeSpec (SetSpec must elemSpec szSpec) cant) =
   cantSpec
-    <> sizeSpec
-    <> geqSpec @fn (Set.size must)
-    <> specSizeBound elemSpec
+    <> szSpec
+    <> geqSpec @fn (setSize must)
+    <> specIntToSize (cardinality elemSpec)
   where
+    -- I don't see how this can have any effect on the sizeSpec
```
MaximilianAlgehed: What is this referring to? `cantSpec`? In which case, the important thing here is that when we know that `mempty` is not a valid value we make sure that the set must contain _something_.

libs/constrained-generators/src/Constrained/Base.hs
```diff
@@ -2422,30 +2437,30 @@ instance (Ord a, HasSpec fn a) => Semigroup (SetSpec fn a) where
 instance (Ord a, HasSpec fn a) => Monoid (SetSpec fn a) where
   mempty = SetSpec mempty mempty TrueSpec
 
+-- ==================================================
+-- THIS FUNCTION IS ONLY USED ONCE in  genFromTypeSpec  for SETS
+-- I think it might be a first attempt at cardinality.
+-- MAYBE WE CAN REPLACE IT with Cardinality
+-- ===================================================
```
MaximilianAlgehed: This isn't doing cardinality, this is computing the possible size of the generated set. That means it's using the cardinality of the _elemSpec_ in the `TypeSpec` case, but as you can see from the `MemberSpec` case this isn't doing simple cardinality.

libs/constrained-generators/src/Constrained/Base.hs
```diff
@@ -2458,16 +2473,17 @@ instance (Ord a, HasSpec fn a) => HasSpec fn (Set a) where
...
       , all (`conformsToSpec` es) s
       ]
 
   genFromTypeSpec (SetSpec must e TrueSpec) = (must <>) . Set.fromList <$> listOfT (genFromSpec e)
   genFromTypeSpec ss@(SetSpec must e _) = do
     -- Compute more constrained size spec based on must and elem spec!
-    n <- explain ["Generate set size"] $ genFromSpec (sizeSpec $ typeSpec ss) `suchThatT` (>= Set.size must)
-    go (n - Set.size must) must
+    n <- explain ["Generate set size"] $ genFromSpec (sizeSpec $ typeSpec ss) `suchThatT` (>= setSize must)
+    -- \******  ^ should 'sizeSpec' be cardinality? *******
```
MaximilianAlgehed: No, this is telling us how many elements to put in the set, not the number of sets we can generate.

libs/constrained-generators/src/Constrained/Base.hs
```diff
@@ -2847,6 +2867,11 @@ instance BaseUniverse fn => HasSpec fn Natural where
   genFromTypeSpec = genFromNumSpec
   conformsTo = conformsToNumSpec
   toPreds = toPredsNumSpec
+  cardinalTypeSpec (NumSpecInterval (Just lo) (Just hi)) =
+    if hi >= lo then MemberSpec [fromIntegral @Natural @Int (hi - lo)] else MemberSpec [0]
+  cardinalTypeSpec (NumSpecInterval Nothing (Just hi)) =
+    if hi >= 0 then MemberSpec [fromIntegral @Natural @Int hi] else MemberSpec [0]
```
MaximilianAlgehed: This isn't right, there are a lot of values that fall in the interval `[-inf...0]`

libs/constrained-generators/src/Constrained/Base.hs
```diff
@@ -2847,6 +2867,11 @@ instance BaseUniverse fn => HasSpec fn Natural where
   genFromTypeSpec = genFromNumSpec
   conformsTo = conformsToNumSpec
   toPreds = toPredsNumSpec
+  cardinalTypeSpec (NumSpecInterval (Just lo) (Just hi)) =
+    if hi >= lo then MemberSpec [fromIntegral @Natural @Int (hi - lo)] else MemberSpec [0]
```
MaximilianAlgehed: If `hi == lo` this will say `0`, but in fact it's `1` in that case.

libs/constrained-generators/src/Constrained/Base.hs
```diff
@@ -3142,9 +3175,9 @@ data ListFn fn args res where
     ) =>
     fn '[a] b ->
     ListFn fn '[[a]] b
-  ListSize :: ListFn fn '[[a]] Int
+  ListSize :: ListFn fn '[[a]] Size
```
MaximilianAlgehed: Why is this still here but for `Set` we drop it? Shouldn't they both be using what's in `SizeFn`?

libs/constrained-generators/src/Constrained/Base.hs
```diff
@@ -3715,3 +3750,476 @@ instance HasSpec fn a => Pretty (Spec fn a) where
 
 instance HasSpec fn a => Show (Spec fn a) where
   showsPrec d = shows . pretty . WithPrec d
+
+-- ==================================================================
+-- Size, Sized, and SizeSpec
+-- ==================================================================
+
+newtype Size = MkSize {unSize :: Int} deriving (Eq, Ord, Num)
```
MaximilianAlgehed: I'm very uncomfortable having this as a separate type, as it appears to be giving rise to an `{-# INCOHERENT #-}` instance. I would prefer if we replace it with `Integer`.

libs/constrained-generators/src/Constrained/Base.hs
```diff
@@ -3715,3 +3750,476 @@ instance HasSpec fn a => Pretty (Spec fn a) where
...
+-- | sizeOfSpec generalizes the method 'sizeOfTypeSpec'
+--   From (sizeOfTypeSpec :: TypeSpec fn t -> Spec fn Size)
+--   To   (sizeOfSpec     :: Spec fn t     -> Spec fn Size)
+--   It is not unusual for instances (HasSpec fn t) to define sizeOfTypeSpec with calls to sizeOfSpec,
+--   Because many (TypeSpec fn t)'s contain (Spec fn s), for types 's' different from 't'
+sizeOfSpec :: forall fn t. (BaseUniverse fn, Sized t) => Spec fn t -> Spec fn Size
+sizeOfSpec TrueSpec = TrueSpec
+sizeOfSpec (MemberSpec xs) = MemberSpec (map sizeOf xs)
+sizeOfSpec (ErrorSpec xs) = ErrorSpec xs
+sizeOfSpec SuspendedSpec {} = mempty
```
MaximilianAlgehed: Depending on how this is used I would argue for using an `unsafeExists` here instead of `mempty`.

libs/constrained-generators/src/Constrained/Base.hs
```diff
@@ -3715,3 +3750,476 @@ instance HasSpec fn a => Pretty (Spec fn a) where
...
+exactInt :: Int -> SizeSpec
+exactInt a = rangeInt a a
+
+rangeInt :: Int -> Int -> SizeSpec
+rangeInt a b | a < 0 || b < 0 = error ("Negative Int in call to rangeInt: " ++ show a ++ " " ++ show b)
+rangeInt a b = SizeSpec (NumSpecInterval (Just a) (Just b))
+
+atLeastInt :: Int -> SizeSpec
+atLeastInt a | a < 0 = error ("Negative Int in call to atLeastInt: " ++ show a)
+atLeastInt a = SizeSpec (NumSpecInterval (Just a) Nothing)
```
MaximilianAlgehed:
```suggestion
atLeastInt :: Int -> SizeSpec
atLeastInt a
  | a < 0 = error ("Negative Int in call to atLeastInt: " ++ show a)
  | otherwise = SizeSpec (NumSpecInterval (Just a) Nothing)
```

libs/constrained-generators/src/Constrained/Base.hs
```diff
@@ -3715,3 +3750,476 @@ instance HasSpec fn a => Pretty (Spec fn a) where
...
+
+atMostInt :: Int -> SizeSpec
+atMostInt a | a < 0 = error ("Negative Int in call to atMostInt: " ++ show a)
+atMostInt a = SizeSpec (NumSpecInterval Nothing (Just a))
```
MaximilianAlgehed:
```suggestion
atMostInt :: Int -> SizeSpec
atMostInt a
  | a < 0 = error ("Negative Int in call to atMostInt: " ++ show a)
  | otherwise = SizeSpec (NumSpecInterval Nothing (Just a))
```

libs/constrained-generators/src/Constrained/Base.hs
```diff
@@ -3715,3 +3750,476 @@ instance HasSpec fn a => Pretty (Spec fn a) where
...
+
+exactSizeSpec :: BaseUniverse fn => Size -> Spec fn Size
+exactSizeSpec (MkSize a) = TypeSpec (exactInt a) []
```
MaximilianAlgehed:
```suggestion
exactSizeSpec (MkSize a) = MemberSpec [a]
```

libs/constrained-generators/src/Constrained/Base.hs
```diff
@@ -3715,3 +3750,476 @@ instance HasSpec fn a => Pretty (Spec fn a) where
...
+
+-- | SizeSpec forms a Semigroup. It inherits from the Semigroup instance of (NumSpec Word64)
+instance Semigroup SizeSpec where
+  SizeSpec x <> SizeSpec y = SizeSpec (x <> y)
+
+-- | SizeSpec forms a Monoid
+instance Monoid SizeSpec where
+  mempty = SizeSpec mempty
+
+toPredsSizeSpec :: BaseUniverse fn => Term fn Size -> SizeSpec -> Pred fn
```
MaximilianAlgehed: This already exists

libs/constrained-generators/src/Constrained/Base.hs
```diff
@@ -3715,3 +3750,476 @@ instance HasSpec fn a => Pretty (Spec fn a) where
...
+--   That is not very practical because a very large size is not useful in tests.
+--   What is the practical bound we want on ∞ ? So we answer this using deltaMax.
+--   When we generate [ x ..  ∞ ] we use [ x .. x + deltaMax ]
+--   This creates a uniform distribution between x and (x + deltaMax). If we just choose an arbitray
+--   value greater than 'x', values are not uniform. Just the way arbitrary works on Int.
+deltaMax :: Int
+deltaMax = 10
+
+-- | Gen a random Size, Because Sizes are aways >= 0, raise a failure if that is the case.
+genFromSizeSpec :: MonadGenError m => Int -> SizeSpec -> GenT m Size
```
MaximilianAlgehed: This already exists

libs/constrained-generators/src/Constrained/Base.hs
```diff
@@ -3715,3 +3750,476 @@ instance HasSpec fn a => Pretty (Spec fn a) where
...
+subtractionM (SizeSpec x) (SizeSpec y) = checkInterval (subNumSpec (liftNumSpec x) (liftNumSpec y))
+
+addition :: SizeSpec -> SizeSpec -> SizeSpec
+addition (SizeSpec x) (SizeSpec y) = dropSizeSpec (subNumSpec (liftNumSpec x) (liftNumSpec y))
+
+-- =====================================
+-- Size operations
+
+data SizeFn (fn :: [Type] -> Type -> Type) as b where
+  Plus :: SizeFn fn '[Size, Size] Size
```
MaximilianAlgehed: I'm strongly opposed to having _two_ `Plus` functions in the `BaseUniverse`

libs/constrained-generators/src/Constrained/GenT.hs
```diff
@@ -159,6 +159,9 @@ instance Monad m => Monad (GenT m) where
 strictGen :: GenT m a -> Gen (m a)
 strictGen gen = runGenT gen Strict
 
+genFromGenT :: GenT GE a -> Gen a
```
MaximilianAlgehed, line 162: 👍

libs/constrained-generators/src/Constrained/Test.hs
```diff
@@ -192,6 +194,23 @@ numberyTests =
     , testNumberyListSpec "listSumElemRange" listSumElemRange
     ]
 
+sizeTests :: TestTree
+sizeTests =
+  testGroup "SizeTests" $
+    [ testSpec "sizeAddOrSub1" sizeAddOrSub1
+    , testSpec "sizeAddOrSub2" sizeAddOrSub2
+    , testSpec "sizeAddOrSub3" sizeAddOrSub3
+    , -- This should fail, how do I state this , testSpec "sizeAddOrSub4" sizeAddOrSub4
```
MaximilianAlgehed: something like `expectFailure $ prop_soundness <thing>`. You'd have to not use `testSpec`.

libs/constrained-generators/src/Constrained/Test.hs
```diff
@@ -677,3 +691,105 @@ listMustSizeIssue = constrained $ \xs ->
...
+hasSizeList = hasSize (rangeInt 0 4)
+
+hasSizeSet :: Spec BaseFn (Set Int)
+hasSizeSet = hasSize (rangeInt 1 3)
+
+hasSizeMap :: Spec BaseFn (Map Int Int)
+hasSizeMap = hasSize (rangeInt 1 3)
+
+-- ========================================================
+-- Test properties of the instance Num(NumSpec Integer)
```
MaximilianAlgehed, line 737: 👍

libs/constrained-generators/src/Constrained/Test.hs
```diff
@@ -677,3 +691,105 @@ listMustSizeIssue = constrained $ \xs ->
...
+    , testProperty "scaleOne(y = 1 * y)" scaleOne
+    , testProperty "negNagate(x = x == negate (negate x))" negNegate
+    , testProperty "commutesNumSpec(x+y = y+x)" commutesNumSpec
+    , testProperty "assocNumSpec(x+(y+z) == (x+y)+z)" assocNumSpec
+    , testProperty "assocNumSpecTimes(x*(y*z) == (x*y)*z)" assocNumSpecTimes
+    ]
+
+-- ==========================================================
+
+go :: HasSpec BaseFn t => Spec BaseFn t -> IO ()
```
MaximilianAlgehed: This name is perhaps too generic. I would call it `runTestSpec`.

libs/constrained-generators/src/Constrained/Test.hs
```diff
@@ -677,3 +691,105 @@ listMustSizeIssue = constrained $ \xs ->
...
+go spec = defaultMain (testSpec "testing with go" spec)
+
+run :: forall fn a. HasSpec fn a => Spec fn a -> IO ()
+run spec = do
+  x <- generate (genFromGenT (genFromSpec @fn spec))
```
MaximilianAlgehed:
```suggestion
  x <- generate $ genFromSpec_ @fn spec
```

libs/constrained-generators/src/Constrained/Test.hs
```diff
@@ -677,3 +691,105 @@ listMustSizeIssue = constrained $ \xs ->
...
+    , testProperty "assocNumSpec(x+(y+z) == (x+y)+z)" assocNumSpec
+    , testProperty "assocNumSpecTimes(x*(y*z) == (x*y)*z)" assocNumSpecTimes
+    ]
+
+-- ==========================================================
+
+go :: HasSpec BaseFn t => Spec BaseFn t -> IO ()
+go spec = defaultMain (testSpec "testing with go" spec)
+
+run :: forall fn a. HasSpec fn a => Spec fn a -> IO ()
```
MaximilianAlgehed: Same as above, I would prefer a more specific name like maybe `generateSpec`, then you could also do `generateSpec :: HasSpec fn a => Spec fn a -> IO a` instead of `IO ()` as `ghci` will anyway `print` the result.

libs/constrained-generators/src/Constrained/Univ.hs
```diff
@@ -133,23 +134,24 @@ instance FunctionLike (EqFn fn) where
...
-
 orFn :: forall fn. Member (BoolFn fn) fn => fn '[Bool, Bool] Bool
 orFn = injectFn $ Or @fn
 
 data BoolFn (fn :: [Type] -> Type -> Type) as b where
   Not :: BoolFn fn '[Bool] Bool
-  And :: BoolFn fn '[Bool, Bool] Bool
   Or :: BoolFn fn '[Bool, Bool] Bool
 
+-- One might expect   And :: BoolFn fn '[Bool, Bool] Bool
```
MaximilianAlgehed: :+1:

libs/constrained-generators/src/Constrained/Univ.hs
```diff
@@ -133,23 +134,24 @@ instance FunctionLike (EqFn fn) where
...
+-- But this is problematic because a Term like ( p x &&. q y ) has to solve for
+-- x or y first, choose x, so once x is fixed, it might be impossible to solve for y
+-- Luckily we can get conjuction by using Preds, in fact using Preds, the x and y
+-- can even be the same variable.
```
MaximilianAlgehed: We might want a similar explanation about DeMorgan's law and `(||.)` saying that `(||.)` should be used with care.

libs/constrained-generators/src/Constrained/Univ.hs
```diff
@@ -302,7 +300,7 @@ instance FunctionLike (SetFn fn) where
     Member -> Set.member
     Singleton -> Set.singleton
     Union -> (<>)
-    Size -> Set.size
+    --   SetSize -> MkSize . Set.size FIXME
```
MaximilianAlgehed:
```suggestion
```

MaximilianAlgehed (changes requested):
I have left some comments, but I haven't gone over all the changes to `Base` in detail yet. There is a lot of stuff around the `Size` stuff that already exists for `Integer` and the other numeric types. I think it would be wise to drop `Size` as a type, in favour of `Integer` (to avoid wrapping issues when computing `cardinality` for e.g. `Word64`).

MaximilianAlgehed:
After we've revamped `Size` I can take another look at the changes to `Constrained.Base` - they should be simpler then.

libs/constrained-generators/src/Constrained/Base.hs
TimSheard: Of course, silly of me.

libs/constrained-generators/src/Constrained/Base.hs
TimSheard: Sorry.
It is referring to  
  <> specIntToSize (cardinality elemSpec)

elemSpec is about the properties of each element, How can that affect the number of elements?
I suppose if the cardinalty is, say 100, than there can't be more than 100 things in the set, and if the cardinality is 0, then the set must have size 0.

But specIntToSize (cardinality elemSpec)   says it must be exactly 100 (or 0),  so it probably should be, that the set can be no larger than 100 (or 0), to I guess it should be exactly what it was.

 specIntToSize (cardinality elemSpec)   -- recall cardinality is (Spec fn Int) and we need (Spec fn Size)
Thanks for letting me work this out.

libs/constrained-generators/src/Constrained/Base.hs
TimSheard: So I added a comment, and renamed this possibleSetSizes

libs/constrained-generators/src/Constrained/Base.hs
TimSheard: I change the explanation to
explain ["Choose a possible size Bounds for the Sets to be generated"]

libs/constrained-generators/src/Constrained/Base.hs
TimSheard: Ah, but remember this is a Natural, so - inf  is 0, 

So I changed these clause to
```
cardinalTypeSpec (NumSpecInterval (Just lo) (Just hi)) =
    if hi >= lo then MemberSpec [fromIntegral @Natural @Int (hi - lo + 1)] else MemberSpec [0]
  cardinalTypeSpec (NumSpecInterval Nothing (Just hi)) =
    MemberSpec [fromIntegral @Natural @Int hi + 1] 
  cardinalTypeSpec _ = TrueSpec
```

Not I changed (hi -lo) to (hi - lo + 1)

libs/constrained-generators/src/Constrained/Base.hs
TimSheard: Addressed that in the previous comment. Good catch.

libs/constrained-generators/src/Constrained/Base.hs
TimSheard: This was an oversight on my part. I should have removed it. I am surprised though, that leaves only FoldMap as the only operator on Lists.  I would have thought perhaps Append, and definitely NullList.  
I added a TODO about adding those two at some time.

libs/constrained-generators/src/Constrained/Base.hs
TimSheard: OK, will do this in the next pass. I Promise.

libs/constrained-generators/src/Constrained/Base.hs
TimSheard: done

libs/constrained-generators/src/Constrained/Base.hs
TimSheard: done

libs/constrained-generators/src/Constrained/Base.hs
TimSheard: Yes, but for Int (Not Size). This will go away in the next step.

libs/constrained-generators/src/Constrained/Base.hs
TimSheard: Yes and no. We can talk about this later.  Recall if generating a Size, we shouldn't generate negative numbers. I think that was my main reason for wanting a newtype.

libs/constrained-generators/src/Constrained/Base.hs
TimSheard: Yes and No.   Here we are making a Num instance with plus and minus. I think the other uses addition and negate.
The purpose of this is that Size does not really support Negative sizes, so the old instance wouldn't work.
When we replace Size with (NumSpec Integer) we will have to decide which approach to take, and how to deal with negative numbers.

libs/constrained-generators/src/Constrained/GenT.hs
TimSheard: Yeah,  I wrote this 3 times to explore things, and I kept forgetting how to write it. So I decided to add it to GenT.hs

libs/constrained-generators/src/Constrained/Test.hs
TimSheard: Not the most comprehensive tests, but covers the basics.

libs/constrained-generators/src/Constrained/Base.hs
TimSheard: done

libs/constrained-generators/src/Constrained/Test.hs

libs/constrained-generators/src/Constrained/Test.hs

libs/constrained-generators/src/Constrained/Base.hs
MaximilianAlgehed: NullList can be done with `length_ xs ==. 0` and append is really tricky because that would require talking about the order of elements and distributing foldMap over the two lists.

It's possible it can be done, but I don't know if it would be at the cost of the interaction with foldMap. On the other hand, one of the lists is known when doing propagate spec so it might be possible that you never need to distribute and only need to keep a prefix and a suffix in the spec.

The first thing I would try is cons however, that will tell us if it can be done easily.

libs/constrained-generators/src/Constrained/Base.hs
MaximilianAlgehed: It's a good catch that sizes can't be negative. We do already support `Natural` - so I would like to amend my suggestion by replacing `Integer` with `Natural`!

libs/constrained-generators/src/Constrained/Base.hs
MaximilianAlgehed: Yes, there is something wrong here. It should be `atMostSpec (cardinality elemSpec)` for some `atMostSpec :: Spec fn Int -> Spec fn Int`.

MaximilianAlgehed:
@TimSheard how do you feel about `s/Size/Natural` instead of `Integer`? Then the issues with subtraction etc. should be resolved? Possibly we might want to have another look at the `Num` instance for `Term fn Natural` if there is an issue with `negate` vs `(-)` there?

lehins:
> Possibly we might want to have another look at the Num instance for Term fn Natural if there is an issue with negate vs (-) there?

There is ugly issue with `Num` for `Natural`, which is why I usually suggest to avoid it. For example if `f` is a polymorphic reusable function that is potentially defined in another library:
```
ghci> let f = subtract (-1)
ghci> :t f
f :: Num a => a -> a
ghci> map f [10::Natural]
[*** Exception: arithmetic underflow
```

TimSheard:
I am thinking that we make Size = Integer.
Now we have an instance of genFromSpec
genFromSpec ::  Spec fn Integer -> GenT m Integer
 But we write our own function
genFromSize :: Spec fn Integer -> GenT m Integer,  that never generates an Integer less than 0, 
and we use this function at appropriate places in our code.

MaximilianAlgehed:
With that solution `genFromSize = genFromSpec . geqSpec 0`?

TimSheard:
Yes, that would be right.  Any take a look, I have swithced to Size = Integer

libs/constrained-generators/src/Constrained/Test.hs

libs/constrained-generators/src/Constrained/Univ.hs
TimSheard: done

libs/constrained-generators/src/Constrained/Univ.hs
TimSheard: done

libs/constrained-generators/src/Constrained/Base.hs
TimSheard: Here is a try at atMostSpec

```

atMostSpec :: BaseUniverse fn => Spec fn Size -> Spec fn Size
atMostSpec TrueSpec = TrueSpec
atMostSpec (SuspendedSpec _ _) = TrueSpec
atMostSpec (MemberSpec[]) = ErrorSpec ["empty MemberSec in atMostSpec."]
atMostSpec s@(MemberSpec xs) = s <> typeSpec (atMostInt (maximum xs))
atMostSpec s@(TypeSpec (NumSpecInterval _ hi) bad) = TypeSpec (NumSpecInterval Nothing hi) bad
```
What do you think?

libs/constrained-generators/src/Constrained/Base.hs
TimSheard: So now Size is just a type synonym for Integer

libs/constrained-generators/src/Constrained/Base.hs
TimSheard: So it is now gone, since Size is Integer

libs/constrained-generators/src/Constrained/Base.hs
TimSheard: So now we just use the genFromSpec for Integer.
We may wan't to use
```
genFromSize = genFromSpec . geqSpec 0?
```
where we must get a positive number.

libs/constrained-generators/src/Constrained/Base.hs
TimSheard: So now there is just one Plus function. The one for all Num types.
Note that this is used for   Num n => Num(Term fn n)    instances

Size is now Integer, where   (NumSpec Integer) has its own Num instance.
And this is lifted to  Num(Spec fn Integer).  Very usefull

libs/cardano-ledger-test/src/Test/Cardano/Ledger/STS.hs
```diff
@@ -291,3 +291,40 @@ utxoTests =
...
+
+predmap :: Spec BaseFn Word8
+predmap = constrained $
+              \k -> toPred
+                  [ assert $ not_ (member_ k (Lit $ Set.fromList [1,2,3,4,5::Word8]))
+                  , satisfies (pair_ k (Lit @Word8 7)) TrueSpec
+                  ]
+
+-- type Conway = ConwayEra StandardCrypto
+-}
```
MaximilianAlgehed:
```suggestion
```

libs/constrained-generators/src/Constrained/Base.hs
```diff
@@ -26,7 +27,7 @@
 
 -- The pattern completeness checker is much weaker before ghc-9.0. Rather than introducing redundant
 -- cases and turning off the overlap check in newer ghc versions we disable the check for old
--- versions.
+-- versions.io
```
MaximilianAlgehed:
```suggestion
-- versions.
```

libs/constrained-generators/src/Constrained/Base.hs
```diff
@@ -621,6 +622,22 @@ class
   --   ∀ a. a `conformsTo` spec == a `conformsTo` constrained (\t -> toPreds t spec)
   toPreds :: Term fn a -> TypeSpec fn a -> Pred fn
 
+  -- | Compute an upper and lower bound on the number of solutions genFromTypeSpec might return
+  cardinalTypeSpec :: TypeSpec fn a -> Spec fn Integer
+  cardinalTypeSpec _ = TrueSpec
+
+  -- | A bound on the number of solutions (genFromTypeSpec TrueSpec)
```
MaximilianAlgehed:
```suggestion
  -- | A bound on the number of solutions `genFromTypeSpec TrueSpec` can produce.
```

libs/constrained-generators/src/Constrained/Base.hs
```diff
@@ -621,6 +622,22 @@ class
...
+  --   For a type with finite elements, we can get a much more accurate
+  --   answer than TrueSpec
+  cardinalTrueSpec :: Spec fn Integer
+  cardinalTrueSpec = TrueSpec
+
+  -- | For some types (especially finite ones) there may be much better ways to construct
+  --   a Spec than the default method of just adding a large 'bad' list to TypSpec. This
+  --   function allows each HasSpec instance to decide.
+  typeSpecOpt :: TypeSpec fn a -> [a] -> Spec fn a
+  typeSpecOpt tySpec bad = TypeSpec tySpec bad
```
MaximilianAlgehed, line 639: 👍

libs/constrained-generators/src/Constrained/Base.hs
```diff
@@ -2372,6 +2389,16 @@ instance BaseUniverse fn => HasSpec fn Bool where
   _ `conformsTo` _ = True
   genFromTypeSpec _ = pureGen arbitrary
   toPreds _ _ = TruePred
+  cardinalTypeSpec _ = MemberSpec [2]
+  cardinalTrueSpec = exactSizeSpec 2 -- there are exactly two, True and False
+  typeSpecOpt () [True] = MemberSpec [False]
+  typeSpecOpt () [False] = MemberSpec [True]
+  typeSpecOpt () [True, False] = ErrorSpec ["Both True and False are Not in the Spec, so the Spec is inconsistent."]
+  typeSpecOpt () [False, True] = ErrorSpec ["Both False and True are Not in the Spec, so the Spec is inconsistent."]
+  typeSpecOpt () bad = TypeSpec () bad
```
MaximilianAlgehed:
```suggestion
  typeSpecOpt _ bad = MemberSpec (filter (`notElem` bad) [True, False])
```

libs/constrained-generators/src/Constrained/Base.hs
```diff
@@ -2422,30 +2451,28 @@ instance (Ord a, HasSpec fn a) => Semigroup (SetSpec fn a) where
...
-sizeSpec (MemberSpec es) = MemberSpec (Set.size <$> es)
-sizeSpec SuspendedSpec {} = TrueSpec
-sizeSpec (ErrorSpec es) = ErrorSpec es
-sizeSpec (TypeSpec (SetSpec must elemSpec sizeSpec) cant) =
+-- ==================================================
+-- Suppose genFromSpec (Spec fn (Set a)) can generate 3 distinct sets {s1::Set a, s2::Set a, s3::Set a}
+-- where (Set.size s1 = 4) (Set.size s2 = 12) and (Set.size s3 = 0)
+-- then we want possibleSizes to generate the following things {4,12,0}
+-- There are No guarantees that this is a tight spec!
+-- ===================================================
```
MaximilianAlgehed: Perhaps haddock format here?

libs/constrained-generators/src/Constrained/Base.hs
```diff
@@ -2341,7 +2358,7 @@ genFromFold ::
   , HasSpec fn a
   ) =>
   [a] ->
-  Spec fn Int ->
+  Spec fn Size ->
```
MaximilianAlgehed: I would prefer not even having a type synonym for `Size` to be honest.

libs/constrained-generators/src/Constrained/Base.hs
```diff
@@ -2544,7 +2572,10 @@ instance BaseUniverse fn => Functions (SetFn fn) fn where
     Subset
       | HOLE :? Value (s :: Set a) :> Nil <- ctx
       , Evidence <- prerequisites @fn @(Set a) -> caseBoolSpec spec $ \case
-          True -> typeSpec $ SetSpec mempty (MemberSpec $ Set.toList s) mempty
+          True ->
+            if Set.null s
```
MaximilianAlgehed, line 2549: 👍

libs/constrained-generators/src/Constrained/Base.hs
```diff
@@ -2641,42 +2668,47 @@ instance HasSpec fn a => HasSpec fn [a] where
...
-    sz0 <- genFromSpec (sizeSpecList $ typeSpec ss)
-    let sz = sz0 - length must
+    sz0 <- genFromSizeSpec (sizeSpecList $ typeSpec ss)
+    let sz = fromIntegral (sz0 - listSize must)
     lst <-
       listOfUntilLenT
         (genFromSpec elemS)
         sz
-        ((`conformsToSpec` szSpec) . (+ length must))
+        ((`conformsToSpec` szSpec) . (+ listSize must) . fromIntegral)
```
MaximilianAlgehed: I don't see the point of introducing functions `listSize` and `setSize` when there are already `length` and `Set.size`.

libs/constrained-generators/src/Constrained/Base.hs
```diff
@@ -2641,42 +2668,47 @@ instance HasSpec fn a => HasSpec fn [a] where
...
-
-sizeSpecList :: forall fn a. Spec fn [a] -> Spec fn Int
+      <> satisfies (sizeOf_ x) size
+
+-- | This function is a special case of (sizeOfSpec:: Spec fn t -> Spec fn Size), it is identical, except,
+--   since [a] has a monoid instance, it can be more accurate in two ways:
+--   1) by inspecting the cant set of (TypeSpec ListSpec{} cant)
+--   2) by inspecting the SuspendedSpec
+--   Neither of which 'sizeOfSpec' can do, because it does not know what the type 't' is.
+sizeSpecList :: forall fn a. Spec fn [a] -> Spec fn Size
```
MaximilianAlgehed: We could generalize this to any type that admits the use of `sizeOf_`.

libs/constrained-generators/src/Constrained/Base.hs
```diff
@@ -3142,23 +3180,21 @@ data ListFn fn args res where
     ) =>
     fn '[a] b ->
     ListFn fn '[[a]] b
-  ListSize :: ListFn fn '[[a]] Int
 
-listSizeFn :: forall fn a. Member (ListFn fn) fn => fn '[[a]] Int
-listSizeFn = injectFn $ ListSize @fn
+{- TODO add these?
+  AppendFn :: ListFn fn '[[a],[a]] [a]
+  NullFn :: ListFn fn '[[a]] Bool
```
MaximilianAlgehed:
```suggestion
```

libs/constrained-generators/src/Constrained/Base.hs
```diff
@@ -3715,3 +3751,397 @@ instance HasSpec fn a => Pretty (Spec fn a) where
...
+
+sizeOfFn :: forall fn a. (HasSpec fn a, Member (SizeFn fn) fn, Sized a) => fn '[a] Size
+sizeOfFn = injectFn $ SizeOf @fn @a
+
+sizeOf_ ::
+  forall a fn.
+  (HasSpec fn a, Sized a) =>
+  Term fn a ->
+  Term fn Size
+sizeOf_ = app sizeOfFn
```
MaximilianAlgehed: I would prefer to put this together with the other syntax functions exported from this module. Also, here is a place where we could add
```
null_ :: (HasSpec fn a, Sized a) => Term fn a -> Term fn Bool
null_ xs = sizeOf_ xs =. 0
```

libs/constrained-generators/src/Constrained/Base.hs
```diff
@@ -3715,3 +3751,397 @@ instance HasSpec fn a => Pretty (Spec fn a) where
...
+  sizeOf x = x
+  liftSizeSpec x = typeSpec x
+  liftMemberSpec xs = MemberSpec xs
+  sizeOfTypeSpec x = (TypeSpec x [])
+
+instance Sized Int where
+  sizeOf x = fromIntegral x
+  liftSizeSpec (NumSpecInterval a b) = typeSpec (NumSpecInterval (fromIntegral <$> a) (fromIntegral <$> b))
+  liftMemberSpec xs = MemberSpec (map fromIntegral xs)
+  sizeOfTypeSpec (NumSpecInterval a b) = TypeSpec (NumSpecInterval (toInteger <$> a) (toInteger <$> b)) []
```
MaximilianAlgehed: I'm slightly uncomfortable with these instances. Why would one want them?

libs/constrained-generators/src/Constrained/Base.hs
```diff
@@ -3715,3 +3751,397 @@ instance HasSpec fn a => Pretty (Spec fn a) where
...
+-- | sizeOfSpec generalizes the method 'sizeOfTypeSpec'
+--   From (sizeOfTypeSpec :: TypeSpec fn t -> Spec fn Size)
+--   To   (sizeOfSpec     :: Spec fn t     -> Spec fn Size)
+--   It is not unusual for instances (HasSpec fn t) to define sizeOfTypeSpec with calls to sizeOfSpec,
+--   Because many (TypeSpec fn t)'s contain (Spec fn s), for types 's' different from 't'
+sizeOfSpec :: forall fn t. (BaseUniverse fn, Sized t) => Spec fn t -> Spec fn Size
+sizeOfSpec TrueSpec = TrueSpec
+sizeOfSpec (MemberSpec xs) = MemberSpec (map sizeOf xs)
+sizeOfSpec (ErrorSpec xs) = ErrorSpec xs
+sizeOfSpec SuspendedSpec {} = mempty
```
MaximilianAlgehed: Here we can do the same thing we do in the special case for lists you had above.

libs/constrained-generators/src/Constrained/Base.hs
```diff
@@ -3715,3 +3751,397 @@ instance HasSpec fn a => Pretty (Spec fn a) where
...
+sizeOfSpec (TypeSpec x _) = sizeOfTypeSpec @t @fn x
+
+-- | Turn a Size spec into an ErrorSpec if it has negative numbers.
+check :: Spec fn Size -> Spec fn Size
```
MaximilianAlgehed: Can we improve the name of this function?

libs/constrained-generators/src/Constrained/Test.hs
```diff
@@ -356,7 +375,7 @@ mapPairSpec = constrained' $ \m s ->
 
 mapEmptyDomainSpec :: Spec BaseFn (Map Int Int)
 mapEmptyDomainSpec = constrained $ \m ->
-  subset_ (dom_ m) mempty
+  subset_ (dom_ m) (Lit Set.empty)
```
MaximilianAlgehed:
```suggestion
  subset_ (dom_ m) mempty
```

libs/constrained-generators/src/Constrained/Base.hs
```diff
@@ -3406,15 +3432,25 @@ size_ ::
...
+union_ = app unionFn
 
 length_ ::
   forall a fn.
   HasSpec fn [a] =>
   Term fn [a] ->
-  Term fn Int
-length_ = app listSizeFn
+  Term fn Size
+length_ = app sizeOfFn
```
MaximilianAlgehed: We shouldn't have this _and_ `sizeOf_`.

libs/constrained-generators/src/Constrained/Base.hs
```diff
@@ -3715,3 +3751,397 @@ instance HasSpec fn a => Pretty (Spec fn a) where
...
+--   Here we lift the HasSpec methods 'cardinalTrueSpec' and 'cardinalTypeSpec'
+--   from (TypeSpec fn Int) to (Spec fn Int)
+cardinality :: forall fn a. (Eq a, BaseUniverse fn, HasSpec fn a) => Spec fn a -> Spec fn Integer
+cardinality TrueSpec = cardinalTrueSpec @fn @a
+cardinality (MemberSpec es) = exactSizeSpec (listSize (nub es))
+cardinality ErrorSpec {} = equalSpec 0
+-- This is a concession, there is not really a good way to know
+cardinality (TypeSpec s cant) =
+  subSpecInt (cardinalTypeSpec @fn @a s) (exactSizeSpec (listSize (filter (\c -> conformsTo @fn @a c s) cant)))
+cardinality SuspendedSpec {} = TrueSpec
```
MaximilianAlgehed: This can be improved to `cardinalTrueSpec` - as that's the _maximum possible_

libs/constrained-generators/src/Constrained/Base.hs
```diff
@@ -3715,3 +3751,397 @@ instance HasSpec fn a => Pretty (Spec fn a) where
...
+
+-- | Put some (admittedly loose bounds) on the number of solutions that
+--   'genFromTypeSpec' might return. For lots of types, there is no way to be very accurate.
+--   Here we lift the HasSpec methods 'cardinalTrueSpec' and 'cardinalTypeSpec'
+--   from (TypeSpec fn Int) to (Spec fn Int)
+cardinality :: forall fn a. (Eq a, BaseUniverse fn, HasSpec fn a) => Spec fn a -> Spec fn Integer
+cardinality TrueSpec = cardinalTrueSpec @fn @a
+cardinality (MemberSpec es) = exactSizeSpec (listSize (nub es))
+cardinality ErrorSpec {} = equalSpec 0
+-- This is a concession, there is not really a good way to know
```
MaximilianAlgehed:
```suggestion
```

libs/constrained-generators/src/Constrained/Base.hs
```diff
@@ -3715,3 +3751,397 @@ instance HasSpec fn a => Pretty (Spec fn a) where
...
+-- the lower and upper bounds are concrete. We can't play the game we used in cardinalSizeBoundedSpec
+-- Because Integer does not have minBound and maxBound instances.
+-- So if either NumSpec bound is Nothing, then we return TrueSpec.
+-- Just a little bit of infinity, makes the whole thing infinity.
+cardinalSizeIntegerSpec :: forall n fn. Integral n => NumSpec n -> Spec fn Integer
+cardinalSizeIntegerSpec (NumSpecInterval (Just lo) (Just hi)) =
+  if hi >= lo
+    then MemberSpec [fromIntegral @n @Integer (hi - lo)]
+    else MemberSpec [0]
+cardinalSizeIntegerSpec (NumSpecInterval _ _) = TrueSpec
```
MaximilianAlgehed: There is a bunch of other code for the `HasSpec` instances for `Word` and `Natural` etc. that treat bounded and unbounded types uniformly. It would be good if we could do that here too to reduce the amount of code.

libs/constrained-generators/src/Constrained/Base.hs
```diff
@@ -3715,3 +3751,397 @@ instance HasSpec fn a => Pretty (Spec fn a) where
...
+
+exactSizeSpec :: BaseUniverse fn => Size -> Spec fn Size
+exactSizeSpec a = MemberSpec [a]
+
+atMostSpec :: BaseUniverse fn => Spec fn Size -> Spec fn Size
+atMostSpec TrueSpec = TrueSpec
+atMostSpec (SuspendedSpec _ _) = TrueSpec
+atMostSpec (ErrorSpec xs) = ErrorSpec xs
+atMostSpec (MemberSpec []) = ErrorSpec ["empty MemberSec in atMostSpec."]
+atMostSpec s@(MemberSpec xs) = s <> typeSpec (atMostSize (maximum xs))
```
UlfNorell: The `s <>` is wrong, no?

libs/constrained-generators/src/Constrained/Base.hs
```diff
@@ -2824,14 +2856,6 @@ instance BaseUniverse fn => HasSpec fn Int where
   conformsTo = conformsToNumSpec
   toPreds = toPredsNumSpec
 
-instance BaseUniverse fn => HasSpec fn Integer where
-  type TypeSpec fn Integer = NumSpec Integer
-  emptySpec = emptyNumSpec
-  combineSpec = combineNumSpec
-  genFromTypeSpec = genFromNumSpec
-  conformsTo = conformsToNumSpec
-  toPreds = toPredsNumSpec
```
MaximilianAlgehed: I'm not a fan of moving this instance away from all the other instances for numbers

libs/constrained-generators/src/Constrained/Base.hs
```diff
@@ -3715,3 +3751,397 @@ instance HasSpec fn a => Pretty (Spec fn a) where
...
+-- ==================================================================================
+-- (NumSpec Integer) can support interval arithmetic, so we can make a (Num (NumSpec fn Integer)) instance
+-- Given operator #, then (a,b) # (c,d) = (minimum s, maximum s) where s = [a # c, a # d, b # c, b # d]
+-- There are simpler rules for (+) and (-), but for (*) we need to use the general rule.
+
+operM :: (a -> a -> a) -> Maybe a -> Maybe a -> Maybe a
+operM _ Nothing Nothing = Nothing
+operM _ Nothing (Just _) = Nothing
+operM _ (Just _) Nothing = Nothing
+operM f (Just x) (Just y) = Just (f x y)
```
MaximilianAlgehed: `operM` is called `liftM2` in `Control.Monad`, alternatively `f <$> m <*> m'`

libs/constrained-generators/src/Constrained/Base.hs
```diff
@@ -3715,3 +3751,397 @@ instance HasSpec fn a => Pretty (Spec fn a) where
...
+
+-- | Be REALLY CAREFULL, only the Integer instance is immune from Overflow issues
+instance (Show n, Num n, Ord n) => Num (NumSpec n) where
+  (+) = addNumSpec
+  (-) = subNumSpec
+  (*) = multNumSpec
+  negate = negNumSpec
+  fromInteger n = NumSpecInterval (Just (fromInteger n)) (Just (fromInteger n))
+  abs = error "No abs in the Num (NumSpec Integer) instance"
+  signum = error "No signum in the Num (NumSpec Integer) instance"
```
MaximilianAlgehed: Do we use this at any other type than `Integer`? If so, perhaps we shouldn't define it for the unsafe types?

MaximilianAlgehed (changes requested):

libs/constrained-generators/src/Constrained/Base.hs
```diff
@@ -3715,3 +3751,397 @@ instance HasSpec fn a => Pretty (Spec fn a) where
...
+
+atMostSize :: Size -> SizeSpec
+atMostSize a
+  | a < 0 = error ("Negative Int in call to atMostSize: " ++ show a)
+  | otherwise = NumSpecInterval Nothing (Just a)
+
+exactSizeSpec :: BaseUniverse fn => Size -> Spec fn Size
+exactSizeSpec a = MemberSpec [a]
+
+atMostSpec :: BaseUniverse fn => Spec fn Size -> Spec fn Size
```
MaximilianAlgehed: Can we write down the correctness criteria for this as a QuickCheck property?

libs/constrained-generators/src/Constrained/Base.hs
```diff
@@ -3715,3 +3751,397 @@ instance HasSpec fn a => Pretty (Spec fn a) where
...
+atMostSpec TrueSpec = TrueSpec
+atMostSpec (SuspendedSpec _ _) = TrueSpec
```
MaximilianAlgehed: This could be done more tightly with an `unsafeExists`

libs/constrained-generators/src/Constrained/Base.hs
TimSheard: The type synonym was very helpful when making the change from the old newtype Size to Integer, as it marked places where things can go wrong. We can change it now that we have finished that transformation.

libs/constrained-generators/src/Constrained/Base.hs
TimSheard: Well, they return Integer, and the others return Int, so unless you want a bunch of 'fromIntegral' everywhere, this make things cleaner. The other thing is to use sizeOf, the function from the Sized class, but this may introduce Sized constraints.. May I'll try that and see it it happens. Might not since we usually know its a set or list.

libs/constrained-generators/src/Constrained/Base.hs
TimSheard: Have to think about that.
Actually, it has nothing to do with what types have Sized instances. It us used to generate a  (Spec fn Size) consiistent with the TypeSpec passed to `genFromTypeSpec`

libs/constrained-generators/src/Constrained/Base.hs
TimSheard: Not really important. But all the other stuff about Integers, that the other numeric type don't support, is nearby. Keeps it all together. I moved it back.

libs/constrained-generators/src/Constrained/Base.hs
TimSheard: keeps some of the old code, backward compatible. So we don't need changes in the Conway era stuff. Easliy changed.

libs/constrained-generators/src/Constrained/Base.hs
TimSheard: OK good idea.

libs/constrained-generators/src/Constrained/Base.hs
TimSheard: The thought behind this is to use the mechanism that <> provides. To intersect the two specs. If you look what happens with (MemberSpec _ <> TypeSpec _ -) it makes a Member spec. So the result is that same as  MemberSpec [maximum xs], which is probably a better defintion anyway. Glad you pointed this out.

libs/constrained-generators/src/Constrained/Base.hs
TimSheard: In the old system, some times you want to make two things the same size, but wan't to put bounds on what that size is. This lets you do that.

libs/constrained-generators/src/Constrained/Base.hs
TimSheard: Didn't know that.  Hmm, wonder what liftM1 does?  does it return 'x' if one of the operands is (Just x) and the other is Nothing?
I like this one best.  f <$> m <*> m'

libs/constrained-generators/src/Constrained/Base.hs
TimSheard: I don't know the answer to that. It was easy to make the general types. So I did.

libs/constrained-generators/src/Constrained/Base.hs
TimSheard: So there is a design decision here. All the cardinal stuff depends on the Num instances, which we debated above as being a bit dangerous because of overflow.

Here we are more accurate on Bounded types, but they are more dangerous elsewhere.

I guess we have to make a decision if we only want this type at Integer. If so all the other decisions are made for us, and these questions goes away.

libs/constrained-generators/src/Constrained/Base.hs
TimSheard: how about
checkForNegativeSize

libs/constrained-generators/src/Constrained/Test.hs
TimSheard: Using mempty in contexts like this makes me  uncomfortable, with the change I made it is so easy to understand what is going on.  The fact that (Term fn (Set a)) has a Monoid instance is pretty obscure (and surprising) fact.

libs/constrained-generators/src/Constrained/Base.hs
TimSheard: At second thought, I'll go with typeSpec (atMostSize (maximum xs))

libs/constrained-generators/src/Constrained/Base.hs
TimSheard: But the best answer is to literally inline that at the two call sites. I like that best.

libs/constrained-generators/src/Constrained/Base.hs
TimSheard: There is a design decision to be made here.

libs/constrained-generators/src/Constrained/Base.hs
TimSheard: So it works to replace everything with sizeOf.  So I did that.

libs/constrained-generators/src/Constrained/Base.hs
TimSheard: done

libs/constrained-generators/src/Constrained/Base.hs
TimSheard: So the more I think about this,  In order to do that one needs a (HasSpec fn SizeSpec) instance,
we should talk about this.

libs/constrained-generators/src/Constrained/Base.hs
TimSheard: OK, I think I did this right. If not let me know
```
sizeOfSpec :: forall fn t. (BaseUniverse fn, Sized t) => Spec fn t -> Spec fn Size
sizeOfSpec TrueSpec = TrueSpec
sizeOfSpec (MemberSpec xs) = MemberSpec (map sizeOf xs)
sizeOfSpec (ErrorSpec xs) = ErrorSpec xs
sizeOfSpec (SuspendedSpec x p) =
  constrained $ \len ->
    Exists
      (\_ -> fatalError ["sizeOfSpec: Exists"])
      (x :-> (Assert [] (len ==. sizeOf_ (V x)) <> p))
sizeOfSpec (TypeSpec x _) = sizeOfTypeSpec @t @fn x
```

libs/constrained-generators/src/Constrained/Base.hs
TimSheard: Note that length is specialized to List, while sizeOf_   needs a Sized instance.

libs/constrained-generators/src/Constrained/Base.hs
TimSheard: done

libs/constrained-generators/src/Constrained/Base.hs
MaximilianAlgehed: Taking intuitions about size from the old system here is a bit dangerous. I would prefer if we used `genHint` as much as possible for controlling the distribution if things as that won't have a knock on effect on checking.

For the cases where you do want to talk explicitly about the size relationship between things I think you're good to go with what we have. We could argue back and forth between having a size conversion function for ints or just adding `toInteger`. I would be more in favour of the latter to be honest.

libs/constrained-generators/src/Constrained/Base.hs
```diff
@@ -3715,3 +3769,378 @@ instance HasSpec fn a => Pretty (Spec fn a) where
...
+
+instance Sized Size where
+  sizeOf x = x
+  liftSizeSpec x = typeSpec x
+  liftMemberSpec xs = MemberSpec xs
+  sizeOfTypeSpec x = (TypeSpec x [])
+
+instance Sized Int where
+  sizeOf x = fromIntegral x
+  liftSizeSpec (NumSpecInterval a b) = typeSpec (NumSpecInterval (fromIntegral <$> a) (fromIntegral <$> b))
```
MaximilianAlgehed: The uses of fromIntegral here are not safe w.r.t max and min int.

libs/constrained-generators/src/Constrained/Test.hs
MaximilianAlgehed: I strongly disagree, the goal here is to keep things close to Haskell syntax to avoid overhead when reading the code. Introducing extra `Lit` when it's not necessary I think makes the code more difficult to read.

libs/constrained-generators/src/Constrained/Base.hs
MaximilianAlgehed: You're right, keeping `length_` around makes sense to make the cognitive distance from standard Haskell to this as small as possible.

libs/constrained-generators/src/Constrained/Base.hs
TimSheard: Check into MaybeBounded

libs/constrained-generators/src/Constrained/Base.hs
TimSheard: do x <- genFromTypeSpec spec
     y <- genFromTypeSpec (atMostSpec spec)
     conformsTo spec x && conformsTo spec y && x <= y

TimSheard: widest interval whose largest element is admitted by the original spec

libs/constrained-generators/src/Constrained/Base.hs
TimSheard: atMostSpec s@SuspendedSpec{} = constrained $ \ x -> unsafeExists $ \ y -> y `satisfies` s <> x <=. y

libs/constrained-generators/src/Constrained/Base.hs
TimSheard: Now there is one one function.
```
cardinalSizeSpec :: forall n fn. (Integral n, Num n,MaybeBounded n) => NumSpec n -> Spec fn Integer
```

libs/constrained-generators/src/Constrained/Base.hs
TimSheard: We now suupport only Integer instances

libs/constrained-generators/src/Constrained/Base.hs
TimSheard: Ok it is now
```
maxSpec s@(SuspendedSpec z p) =
   constrained $ \ x -> unsafeExists $ \ y -> 
                           [y `satisfies` s
                           , Assert ["maxSpec on SuspendedSpec"] (x <=. y)
                           ]
```

libs/constrained-generators/src/Constrained/Base.hs
TimSheard: we got rid of these 2 instances.

libs/constrained-generators/src/Constrained/Base.hs
TimSheard: They have been removed.

libs/constrained-generators/src/Constrained/Base.hs
TimSheard: There is only one call to this function. In the instance for `genFromTypeSpec ` for lists. It used to be this
```
 genFromTypeSpec ss@(ListSpec must szSpec elemS NoFold) = do
    sz0 <- genFromSizeSpec (sizeSpecList $ typeSpec ss)
   ...
```
But all the information we need is in the TypeSpec `(ListSpec must szSpec elemS NoFold) `

So it now been changed to
```
 genFromTypeSpec (ListSpec must szSpec elemS NoFold) = do
    sz0 <- genFromSizeSpec (szSpec <> geqSpec @fn (sizeOf must) <> maxSpec (cardinality elemS))
```
And the function has been removed.

TimSheard: I'd like confirmation that `maxSpec (cardinality elemS)`  is the right thing to do.  I will not resolve this, But if you agree, you can resolve it when you approve.

libs/constrained-generators/src/Constrained/Base.hs
TimSheard: done

libs/constrained-generators/src/Constrained/Base.hs
TimSheard: A similar transformation happens for the Set instance of genFromTypeSpec

```
genFromTypeSpec ss@(SetSpec must elemS szSpec) = do
    n <- explain ["Choose a possible size Bounds for the Sets to be generated"] $
                 genFromSizeSpec (szSpec <> geqSpec @fn (sizeOf must) <> maxSpec (cardinality elemS))
    go (n - sizeOf must) must
```

libs/constrained-generators/src/Constrained/Base.hs
TimSheard: No more Size type synonym

libs/constrained-generators/src/Constrained/Test.hs
TimSheard: It now reads
```
subset_ (dom_ m) mempty  -- mempty in the Monoid instance (Term fn (Set a))
```

libs/constrained-generators/src/Constrained/Base.hs
TimSheard:
```
sizeOfSpec (SuspendedSpec x p) =
  constrained $ \len ->
    Exists
      (\_ -> fatalError ["sizeOfSpec: Exists"])
      (x :-> (Assert [] (len ==. sizeOf_ (V x)) <> p))
```

libs/constrained-generators/src/Constrained/Test.hs
TimSheard: I expect a negative size for sizeAddOrSub4, so the actual test, tests that. So Now I don't need expectFailure.

libs/constrained-generators/src/Constrained/Base.hs
TimSheard: There should be some similar technique for generating the size of a map, But maps do not use genFromTypeSpec, so I am unsure what to do. I left a TODO comment in the code to address later.

MaximilianAlgehed: > I'd like confirmation that `maxSpec (cardinality elemS)` is the right thing to do. I will not resolve this, But if you agree, you can resolve it when you approve.

Not in the list case, as lists can contain duplicates.

libs/constrained-generators/src/Constrained/Base.hs
```diff
@@ -2422,30 +2447,25 @@ instance (Ord a, HasSpec fn a) => Semigroup (SetSpec fn a) where
...
-    <> sizeSpec
-    <> geqSpec @fn (Set.size must)
-    <> specSizeBound elemSpec
+    <> szSpec
+    <> geqSpec @fn (sizeOf must)
+    <> maxSpec (cardinality elemSpec) -- Can't have more elements, than what (genFromSpec elemSpec) can produce
   where
     cantSpec
       | mempty `elem` cant = notEqualSpec 0
       | otherwise = TrueSpec
```
MaximilianAlgehed: We could remove all this code now that it's not used anywhere.

MaximilianAlgehed (approved):

PR #4123, opened 2024-02-27
Add the ability to do constraints over trees in `constrained-generators`

MaximilianAlgehed:
# Description

Also fixes some minor issues around list size

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/libs/constrained-generators/src/Constrained/Base.hs
+++ b/libs/constrained-generators/src/Constrained/Base.hs
@@ -333,6 +333,11 @@ data Pred (fn :: [Type] -> Type -> Type) where
     -- encoded with `ProdOver` (c.f. `Constrained.Univ`).
     List (Binder fn) as ->
     Pred fn
+  GenHint ::
+    HasGenHint fn a =>
+    Hint a ->
+    Term fn a ->
+    Pred fn
   TruePred :: Pred fn
   FalsePred :: [String] -> Pred fn
 
@@ -439,6 +444,7 @@ checkPred env = \case
   Subst x t p -> checkPred env $ substitutePred x t p
   Assert [] t -> runTerm env t
   Assert es t -> explain ("assert:" : map ("  " ++) es) $ runTerm env t
+  GenHint {} -> pure True
   Reify t f (x :-> p) -> do
     val <- runTerm env t
     checkPred (extendEnv x (f val) env) p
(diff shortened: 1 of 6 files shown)


libs/constrained-generators/src/Constrained/Base.hs
TimSheard, line 1: Why in simplifyPred, simplying a GenHint body to a constant. Is  TruePred.
Maybe a few comments here.

libs/constrained-generators/src/Constrained/GenT.hs
```diff
@@ -170,6 +170,16 @@ listOfT gen = do
   lst <- pureGen . listOf $ runGenT gen Loose
   catGEs lst
 
+-- TODO: possibly one could return "more, fewer, ok" in the `validLen` instead
+-- of `Bool`
+listOfUntilLenT :: MonadGenError m => GenT GE a -> Int -> (Int -> Bool) -> GenT m [a]
```
TimSheard, line 177: please add a comment, that explains what this does. Hard to understand, what we could possibly add, if we don't know what it currently does.

TimSheard (approved):
I love the clarity of the module Tree.hs
The examples are also clear as a bell.
Good to go, Thanks Max for such a sweet extension.

libs/constrained-generators/src/Constrained/Base.hs
MaximilianAlgehed: I'll add a comment, bit basically the idea is that if the body simplifies to a constant that means there is nothing left to generate in that term. Consequently, this genHint doesn't inform a generator any more so it can be safely removed.

UlfNorell:
@lehins I'm having trouble figuring out why CI failed. It says
```
Tests failed for test:cardano-ledger-shelley-ma-test
```
but I can't for the life of me find the failing test in the log.

lehins:
@UlfNorell Here is [the reason](https://github.com/IntersectMBO/cardano-ledger/actions/runs/8138988841/job/22241100364?pr=4123#step:15:4278)

This is the flaky test that has been haunting us for the longest time. It is caused by the older trace generation framework that was written back in times of Shelley.
Once we are done with Conway we'll invest some time into fixing it, but for now we are forced to suffer through it. So, feel free to re-run the build.

UlfNorell:
I don't think I have permissions to restart tests runs. At least I can't find the button where I remember it being.

lehins:
> I don't think I have permissions to restart tests runs.

That is strange. You do have write permission to the repo, so you should be able to restart GithubActions CI (FYI, hydra CI is not required, so there is no need to restart if there are failures) 

>  At least I can't find the button where I remember it being.

It should be right here:

image

MaximilianAlgehed:
I have restart rights. I decided to force push here because it was faster. The time is dominated by the `8.10.7` build - so waiting for that to finish before doing a restart (you can't restart while things are running) would be slower. However, I guess the best option would have been to cancel the 8.10 job and then re-start both the failed and 8.10 job. Good to remember next time.

PR #4129, opened 2024-02-27
Modify PParams to use similar naming to ProtocolParams in `cardano-api` 

palas:
# Description

This PR modifies `FromJSON` and `ToJSON` instances for `PParams` of all eras so that they match those of `ProtocolParams`. That way `cardano-api` can move to use `PParams` directly.
See [the corresponding cardano-api PR](https://github.com/IntersectMBO/cardano-api/pull/457).

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/eras/babbage/impl/src/Cardano/Ledger/Babbage/PParams.hs
+++ b/eras/babbage/impl/src/Cardano/Ledger/Babbage/PParams.hs
@@ -102,7 +102,7 @@ import Data.Aeson as Aeson (
   Key,
   KeyValue ((.=)),
   ToJSON (..),
-  Value (Null),
+  Value,
   object,
   pairs,
   withObject,
@@ -350,12 +350,7 @@ babbagePParamsPairs ::
   PParamsHKD Identity era ->
   [a]
 babbagePParamsPairs pp =
-  uncurry (.=)
-    <$> babbagePParamsHKDPairs (Proxy @Identity) pp
-      ++ [ "decentralization" .= Aeson.Null
-         , "extraPraosEntropy" .= Aeson.Null
-         , "minUTxOValue" .= Aeson.Null
-         ]
+  uncurry (.=) <$> babbagePParamsHKDPairs (Proxy @Identity) pp
 
 instance FromJSON (BabbagePParams Identity era) where
   parseJSON =
--- a/eras/babbage/impl/src/Cardano/Ledger/Babbage/PParams.hs
+++ b/eras/babbage/impl/src/Cardano/Ledger/Babbage/PParams.hs
@@ -66,6 +66,7 @@ import Cardano.Ledger.BaseTypes (
   StrictMaybe (..),
   UnitInterval,
   isSNothing,
+  parseAsRational,
  )
 import Cardano.Ledger.Binary (
   DecCBOR (..),
@@ -102,13 +103,13 @@ import Data.Aeson as Aeson (
   Key,
   KeyValue ((.=)),
   ToJSON (..),
+  Value (Null),
   object,
   pairs,
   withObject,
   (.!=),
   (.:),
  )
-import qualified Data.Aeson as Aeson (Value)
 import Data.Functor.Identity (Identity (..))
 import Data.Proxy (Proxy (Proxy))
 import Data.Word (Word16, Word32)
@@ -350,32 +351,37 @@ babbagePParamsPairs ::
   PParamsHKD Identity era ->
   [a]
 babbagePParamsPairs pp =
-  uncurry (.=) <$> babbagePParamsHKDPairs (Proxy @Identity) pp
+  uncurry (.=)
+    <$> babbagePParamsHKDPairs (Proxy @Identity) pp
+      ++ [ "decentralization" .= Aeson.Null
+         , "extraPraosEntropy" .= Aeson.Null
+         , "minUTxOValue" .= Aeson.Null
+         ]
 
 instance FromJSON (BabbagePParams Identity era) where
   parseJSON =
     withObject "PParams" $ \obj ->
       BabbagePParams
-        <$> obj .: "minFeeA"
-        <*> obj .: "minFeeB"
+        <$> obj .: "txFeePerByte"
+        <*> obj .: "txFeeFixed"
         <*> obj .: "maxBlockBodySize"
         <*> obj .: "maxTxSize"
         <*> obj .: "maxBlockHeaderSize"
-        <*> obj .: "keyDeposit"
-        <*> obj .: "poolDeposit"
-        <*> obj .: "eMax"
-        <*> obj .: "nOpt"
-        <*> obj .: "a0"
-        <*> obj .: "rho"
-        <*> obj .: "tau"
+        <*> obj .: "stakeAddressDeposit"
+        <*> obj .: "stakePoolDeposit"
+        <*> obj .: "poolRetireMaxEpoch"
+        <*> obj .: "stakePoolTargetNum"
+        <*> parseAsRational (obj .: "poolPledgeInfluence")
+        <*> parseAsRational (obj .: "monetaryExpansion")
+        <*> parseAsRational (obj .: "treasuryCut")
         <*> obj .: "protocolVersion"
         <*> obj .: "minPoolCost" .!= mempty
-        <*> obj .: "coinsPerUTxOByte"
-        <*> obj .: "costmdls"
-        <*> obj .: "prices"
-        <*> obj .: "maxTxExUnits"
-        <*> obj .: "maxBlockExUnits"
-        <*> obj .: "maxValSize"
+        <*> obj .: "utxoCostPerByte"
+        <*> obj .: "costModels"
+        <*> obj .: "executionUnitPrices"
+        <*> obj .: "maxTxExecutionUnits"
+        <*> obj .: "maxBlockExecutionUnits"
+        <*> obj .: "maxValueSize"
         <*> obj .: "collateralPercentage"
         <*> obj .: "maxCollateralInputs"
 
@@ -550,7 +556,7 @@ babbageCommonPParamsHKDPairs ::
   [(Key, HKD f Aeson.Value)]
 babbageCommonPParamsHKDPairs px pp =
   alonzoCommonPParamsHKDPairs px pp
-    <> [("coinsPerUTxOByte", hkdMap px (toJSON @CoinPerByte) (pp ^. hkdCoinsPerUTxOByteL @_ @f))]
+    <> [("utxoCostPerByte", hkdMap px (toJSON @CoinPerByte) (pp ^. hkdCoinsPerUTxOByteL @_ @f))]
 
 upgradeBabbagePParams ::
   forall f c.
--- a/eras/babbage/impl/src/Cardano/Ledger/Babbage/PParams.hs
+++ b/eras/babbage/impl/src/Cardano/Ledger/Babbage/PParams.hs
@@ -66,7 +66,6 @@ import Cardano.Ledger.BaseTypes (
   StrictMaybe (..),
   UnitInterval,
   isSNothing,
-  parseAsRational,
  )
 import Cardano.Ledger.Binary (
   DecCBOR (..),
@@ -371,9 +370,9 @@ instance FromJSON (BabbagePParams Identity era) where
         <*> obj .: "stakePoolDeposit"
         <*> obj .: "poolRetireMaxEpoch"
         <*> obj .: "stakePoolTargetNum"
-        <*> parseAsRational (obj .: "poolPledgeInfluence")
-        <*> parseAsRational (obj .: "monetaryExpansion")
-        <*> parseAsRational (obj .: "treasuryCut")
+        <*> obj .: "poolPledgeInfluence"
+        <*> obj .: "monetaryExpansion"
+        <*> obj .: "treasuryCut"
         <*> obj .: "protocolVersion"
         <*> obj .: "minPoolCost" .!= mempty
         <*> obj .: "utxoCostPerByte"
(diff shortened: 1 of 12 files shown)


libs/cardano-ledger-core/src/Cardano/Ledger/BaseTypes.hs
```diff
@@ -844,3 +846,21 @@ class Inject t s where
 -- | Helper function for a common pattern of creating objects
 kindObject :: Text -> [Pair] -> Value
 kindObject name obj = object $ ("kind" .= name) : obj
+
+-- | Loseless way of representing BoundedRational as JSON that uses decimals when feasible
+toRationalJSON :: BoundedRational a => a -> Value
+toRationalJSON br =
+  case fromRationalRepetendLimited 20 r of
```
lehins: The value should be 19, since with 20 digits it is possible to create a decimal number that would not fit in a rational that is backed by Word64. Also less magic numbers, that's why we have defined `maxDecimalsWord64`
```suggestion
  case fromRationalRepetendLimited maxDecimalsWord64 r of
```

eras/babbage/impl/src/Cardano/Ledger/Babbage/PParams.hs
```diff
@@ -350,32 +351,37 @@ babbagePParamsPairs ::
...
-        <*> obj .: "a0"
-        <*> obj .: "rho"
-        <*> obj .: "tau"
+        <*> obj .: "stakeAddressDeposit"
+        <*> obj .: "stakePoolDeposit"
+        <*> obj .: "poolRetireMaxEpoch"
+        <*> obj .: "stakePoolTargetNum"
+        <*> parseAsRational (obj .: "poolPledgeInfluence")
+        <*> parseAsRational (obj .: "monetaryExpansion")
+        <*> parseAsRational (obj .: "treasuryCut")
```
lehins: This is unnecessary, since FromJSON instance will do the same thing
```suggestion
        <*> obj .: "poolPledgeInfluence"
        <*> obj .: "monetaryExpansion"
        <*> obj .: "treasuryCut"
```

eras/babbage/impl/src/Cardano/Ledger/Babbage/PParams.hs
```diff
@@ -350,32 +351,37 @@ babbagePParamsPairs ::
   PParamsHKD Identity era ->
   [a]
 babbagePParamsPairs pp =
-  uncurry (.=) <$> babbagePParamsHKDPairs (Proxy @Identity) pp
+  uncurry (.=)
+    <$> babbagePParamsHKDPairs (Proxy @Identity) pp
+      ++ [ "decentralization" .= Aeson.Null
+         , "extraPraosEntropy" .= Aeson.Null
+         , "minUTxOValue" .= Aeson.Null
```
lehins: Why is this necessary? These have been deprecated and are no longer applicable in Babbage

libs/cardano-ledger-core/src/Cardano/Ledger/BaseTypes.hs
```diff
@@ -844,3 +846,21 @@ class Inject t s where
...
+    r = unboundRational br
+
+-- | Aeson parser that works with the result of toRationalJSON
+-- | (It adapts a BoundedRational so that it uses the Rational parser instead)
+parseAsRational :: BoundedRational r => Parser Rational -> Parser r
+parseAsRational parser = do
+  rational <- parser
+  case boundRational rational of
+    Nothing -> fail "Cannot convert to BoundRational"
+    Just x -> return x
```
lehins: This function isn't needed. `FromJSON` instance for all bounded types will already do that.

libs/cardano-ledger-core/src/Cardano/Ledger/Plutus/ExUnits.hs
```diff
@@ -108,14 +110,14 @@ instance ToJSON ExUnits where
...
+          [ "memory" .= toJSON exUnitsMem
+          , "steps" .= toJSON exUnitsSteps
           ]
 
 instance FromJSON ExUnits where
   parseJSON = withObject "exUnits" $ \o -> do
-    exUnitsMem <- checkWord64Bounds =<< o .: "exUnitsMem"
-    exUnitsSteps <- checkWord64Bounds =<< o .: "exUnitsSteps"
+    exUnitsMem <- checkWord64Bounds =<< o .: "memory"
+    exUnitsSteps <- checkWord64Bounds =<< o .: "steps"
```
lehins: We'll need to accept previous keys as well, since some types like `ExUnits` are also present in the genesis file, which we can't change.
Something like this could do:
```suggestion
    exUnitsMem <- checkWord64Bounds =<< (o .: "memory" <|> o .: "exUnitsMem")
    exUnitsSteps <- checkWord64Bounds =<< (o .: "steps" <|> o .: "exUnitsSteps")
```

Alternatively, we need to use a different parser for genesis FromJSON instances. Which could be safer, I think, but I don't have a strong opinion on it, since I don't know how JSON instances for PParams are used

libs/cardano-ledger-core/src/Cardano/Ledger/Plutus/ExUnits.hs
```diff
@@ -163,9 +165,34 @@ instance NoThunks Prices
...
+    object
+      [ "priceSteps" .= toRationalJSON (unboundRational prSteps)
+      , "priceMemory" .= toRationalJSON (unboundRational prMem)
+      ]
+    where
+      toRationalJSON :: Rational -> Value
+      toRationalJSON r =
+        case fromRationalRepetendLimited 20 r of
+          Right (s, Nothing) -> toJSON s
+          _ -> toJSON r
```
lehins: Yes we can roundtrip. This has been fixed recently in #4058 
```suggestion
    object
      [ "priceSteps" .= prSteps
      , "priceMemory" .= prMem
      ]
```

libs/cardano-ledger-core/src/Cardano/Ledger/Plutus/ExUnits.hs
```diff
@@ -163,9 +165,34 @@ instance NoThunks Prices
...
+      mem <- o .: "priceMemory"
+      prSteps <- checkBoundedRational steps
+      prMem <- checkBoundedRational mem
+      return Prices {prSteps, prMem}
+    where
+      -- We cannot round-trip via NonNegativeInterval, so we go via Rational
+      checkBoundedRational r =
+        case boundRational r of
+          Nothing -> fail ("too much precision for bounded rational: " ++ show r)
+          Just s -> return s
```
lehins: Same comments apply here
```suggestion
      prSteps <- o .: "priceSteps" <|> o .: "prSteps" 
      prMem <- o .: "priceMemory" <|> o .: "prMem" 
      return Prices {prSteps, prMem}
```

eras/shelley/impl/src/Cardano/Ledger/Shelley/PParams.hs
```diff
@@ -479,18 +500,18 @@ shelleyCommonPParamsHKDPairs ::
...
-  , ("a0", hkdMap px (toJSON @NonNegativeInterval) (pp ^. hkdA0L @era @f))
-  , ("rho", hkdMap px (toJSON @UnitInterval) (pp ^. hkdRhoL @era @f))
-  , ("tau", hkdMap px (toJSON @UnitInterval) (pp ^. hkdTauL @era @f))
+  , ("stakeAddressDeposit", hkdMap px (toJSON @Coin) (pp ^. hkdKeyDepositL @era @f))
+  , ("stakePoolDeposit", hkdMap px (toJSON @Coin) (pp ^. hkdPoolDepositL @era @f))
+  , ("poolRetireMaxEpoch", hkdMap px (toJSON @EpochInterval) (pp ^. hkdEMaxL @era @f))
+  , ("stakePoolTargetNum", hkdMap px (toJSON @Natural) (pp ^. hkdNOptL @era @f))
+  , ("poolPledgeInfluence", hkdMap px (toRationalJSON @NonNegativeInterval) (pp ^. hkdA0L @era @f))
+  , ("monetaryExpansion", hkdMap px (toRationalJSON @UnitInterval) (pp ^. hkdRhoL @era @f))
+  , ("treasuryCut", hkdMap px (toRationalJSON @UnitInterval) (pp ^. hkdTauL @era @f))
```
lehins:
```suggestion
  , ("poolPledgeInfluence", hkdMap px (toJSON @NonNegativeInterval) (pp ^. hkdA0L @era @f))
  , ("monetaryExpansion", hkdMap px (toJSON @UnitInterval) (pp ^. hkdRhoL @era @f))
  , ("treasuryCut", hkdMap px (toJSON @UnitInterval) (pp ^. hkdTauL @era @f))
```

eras/shelley/impl/src/Cardano/Ledger/Shelley/PParams.hs
```diff
@@ -457,9 +474,13 @@ shelleyCommonPParamsHKDPairsV6 ::
...
 shelleyCommonPParamsHKDPairsV6 px pp =
-  [ ("decentralisationParam", hkdMap px (toJSON @UnitInterval) (pp ^. hkdDL @era @f))
-  , ("extraEntropy", hkdMap px (toJSON @Nonce) (pp ^. hkdExtraEntropyL @era @f))
+  [ ("decentralization", hkdMap px (toRationalJSON @UnitInterval) (pp ^. hkdDL @era @f))
+  , ("extraPraosEntropy", hkdMap px nonceToJSON (pp ^. hkdExtraEntropyL @era @f))
   ]
+  where
+    nonceToJSON :: Nonce -> Aeson.Value
+    nonceToJSON NeutralNonce = Aeson.Null
+    nonceToJSON (Nonce n) = toJSON n
```
lehins: Why not fix the ToJSON/FromJSON instance for `Nonce` instead of having a custom implementation here?

lehins (changes requested):
Very much in favor of this change. Few important notes on the PR:
* We need to be careful to preserve JSON serialization of Genesis files
* BoundedRational JSON serialization has been fixed in #4058 so we can use it instead doing the conversion to Rational and back everywhere

eras/babbage/impl/src/Cardano/Ledger/Babbage/PParams.hs
palas: If we want to continue generating the same exact format as `cardano-cli` we would need to do this, because it currently uses the same data type for all eras. But maybe we don't have to as long as we can parse it, for example. @Jimbo4350, what do you think?

libs/cardano-ledger-core/src/Cardano/Ledger/Plutus/ExUnits.hs
palas: Thanks, I didn't notice because I was working on top of and old commit

eras/shelley/impl/src/Cardano/Ledger/Shelley/PParams.hs
palas: Good idea :+1:

eras/conway/impl/src/Cardano/Ledger/Conway/PParams.hs
```diff
@@ -840,34 +840,34 @@ conwayPParamsPairs pp =
...
+        <*> obj .: "maxBlockExecutionUnits"
+        <*> obj .: "maxValueSize"
         <*> obj .: "collateralPercentage"
         <*> obj .: "maxCollateralInputs"
         <*> obj .: "poolVotingThresholds"
         <*> obj .: "dRepVotingThresholds"
-        <*> obj .: "committeeMinSize"
-        <*> obj .: "committeeMaxTermLength"
+        <*> obj .: "minCommitteeSize"
+        <*> obj .: "committeeTermLength"
```
lehins: These are new parameters, so we don't have problems with backwards compatibility yet. I would suggest naming them the same way they are in the spec, because the names from cardano-api are quite inconsistent (I believe they reflect the original names that we had that have been improved since then).
```suggestion
        <*> obj .: "committeeMinSize"
        <*> obj .: "committeeMaxTermLength"
```

eras/shelley/impl/src/Cardano/Ledger/Shelley/PParams.hs
```diff
@@ -256,26 +257,36 @@ shelleyPParamsPairs ::
...
+    <$> shelleyPParamsHKDPairs (Proxy @Identity) pp
+      ++ [ "collateralPercentage" .= Aeson.Null
+         , "costModels" .= Aeson.emptyObject
+         , "executionUnitPrices" .= Aeson.Null
+         , "maxBlockExecutionUnits" .= Aeson.Null
+         , "maxCollateralInputs" .= Aeson.Null
+         , "maxTxExecutionUnits" .= Aeson.Null
+         , "maxValueSize" .= Aeson.Null
+         , "utxoCostPerByte" .= Aeson.Null
+         ]
```
lehins: Hey @Jimbo4350 I really don't think these are necessary and they certainly don't follow the ledger mantra of previous eras being unaware of features introduced in the future eras. With this approach we will be forced to add fields in all eras for every new parameter we will add in the future. Which is definitely wrong.

```suggestion
```

lehins (changes requested):
This looks great. There are only two problems left to resolve. We need some input from @Jimbo4350 for one of them

eras/babbage/impl/src/Cardano/Ledger/Babbage/PParams.hs
Jimbo4350: No we shouldn't render deprecated protocol parameters. If `cardano-api` is rendering these in the Babbage era then it's wrong.

eras/shelley/impl/src/Cardano/Ledger/Shelley/PParams.hs
Jimbo4350: I agree. The (justified) confusion stems from the definition of  `ProtocolParameters` in `cardano-api` is not parameterized on `era`. @palas we should favour the ledger's JSON serialization when it comes to which parameters are rendered in a given era.

Jimbo4350 (commented):
Looking good so far. I agree with @lehins wrt not including deprecated protocol parameters. `cardano-api` incorrectly rendered deprecated protocol parameters.

lehins (approved):
Looks great. Thank you!

Issue #4137, opened 2024-02-28
Constraint based test failure on CI


lehins:
There was a [test failure on CI](https://ci.iog.io/build/3058930) and here is a reproducer for current master a807afb1ae466775eceab099158c03ba861f5404:
```
    Constraint soundness:                                             FAIL (0.21s)
      *** Failed! Timeout of 100000 microseconds exceeded. (after 19 tests):
      OrderInfo {sumBeforeParts = False, sizeBeforeArg = False, setBeforeSubset = False}
      (,GenEnv {gOrder = OrderInfo {sumBeforeParts = False, sizeBeforeArg = False, setBeforeSubset = False}, gEnv = A -> Set{▵₳ -887 | size = 29}
      AA -> ▵₳ -68354
      AB -> Map{542 -> ₳ 11 | size = 9, sum = ₳ 86}
      AD -> Map{67 -> 366 | size = 8, sum = 39822}
      AE -> Map{▵₳ -797 -> ▵₳ -740 | size = 13}
      AF -> Map{542 -> ₳ 11 | size = 9, sum = ₳ 86}
      AG -> Map{98 -> ▵₳ -1 | size = 14}
      AI -> Set{₳ 6 | size = 6, sum = ₳ 94}
      AJ -> (Range 6 6)
      AN -> Map{139 -> 9603 | size = 15, sum = 82185}
      AO -> Map{3805 -> ▵₳ -22 | size = 5}
      B -> Map{102 -> ▵₳ 304 | size = 4}
      C -> Map{2527 -> ▵₳ 441 | size = 5}
      D -> Map{₳ 12 -> ₳ 7 | size = 4, sum = ₳ 31}
      E -> ₳ 158
      F -> Set{▵₳ -972 | size = 5}
      G -> Map{13 -> ▵₳ 232 | size = 13}
      H -> ▵₳ 887320
      I -> Map{▵₳ -887 -> ₳ 16 | size = 18, sum = ₳ 149}
      J -> Set{₳ 6 | size = 6, sum = ₳ 94}
      K -> [₳ 11 | size = 11, sum = Coin 121]
      L -> Set[98,1010,3935,4073,5641,7249,9938]
      M -> Map{₳ 1 -> ▵₳ 8 | size = 9}
      N -> Map{946 -> ▵₳ 775 | size = 4}
      O -> (Range 15 15)
      P -> Map{₳ 1 -> ▵₳ -446 | size = 13}
[...]

PR #4139, opened 2024-02-28
Add extra time from 1 sec to 2 sec for prop_soundness test.

TimSheard:
Addressed issue #4137 
Adds more time to prop_soundness.  This is a big test. It looks like it times out infrequently, over running its 1 second
time bound.  I upped the time bound from 1 to 2 seconds.

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/Tests.hs
+++ b/libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/Tests.hs
@@ -4,6 +4,7 @@
 {-# LANGUAGE FunctionalDependencies #-}
 {-# LANGUAGE GADTs #-}
 {-# LANGUAGE LambdaCase #-}
+{-# LANGUAGE NumericUnderscores #-}
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE RankNTypes #-}
 {-# LANGUAGE RecordWildCards #-}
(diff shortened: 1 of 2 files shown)


libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/Tests.hs
```diff
@@ -699,7 +700,7 @@ runPreds ps = do
 -- | Generate a set of satisfiable constraints and check that we can generate a solution and that it
 --   actually satisfies the constraints.
 prop_soundness :: OrderInfo -> Property
-prop_soundness = prop_soundness' False []
+prop_soundness x = within 2_000_000 $ prop_soundness' False [] x
```
lehins: I think it needs to be adjusted in the `prop_soundness'`, because it is an argument to:

https://github.com/IntersectMBO/cardano-ledger/blob/d3e92cf43ab9408180944604fed5210534afdd2b/libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/Tests.hs#L653

TimSheard: So a little bit of research shows there are two ways we might handle this

1) 
within :: [Testable](https://hackage.haskell.org/package/QuickCheck-2.14.3/docs/Test-QuickCheck.html#t:Testable) prop => [Int](https://hackage.haskell.org/package/base-4.16.3.0/docs/Data-Int.html#t:Int) -> prop -> [Property](https://hackage.haskell.org/package/QuickCheck-2.14.3/docs/Test-QuickCheck.html#t:Property)

Considers a property failed if it does not complete [within](https://hackage.haskell.org/package/QuickCheck-2.14.3/docs/Test-QuickCheck.html#v:within) the given number of microseconds.

Note: if the property times out, variables quantified inside the [within](https://hackage.haskell.org/package/QuickCheck-2.14.3/docs/Test-QuickCheck.html#v:within) will not be printed. Therefore, you should use within only in the body of your property.

Good: prop_foo a b c = within 1000000 ...

Bad: prop_foo = within 1000000 $ \a b c -> ...

Bad: prop_foo a b c = ...; main = quickCheck (within 1000000 prop_foo)

So it seems that the inner most within controls things. So the suggested soulution would be the inner one, So that should work.

2)
discardAfter :: [Testable](https://hackage.haskell.org/package/QuickCheck-2.14.3/docs/Test-QuickCheck.html#t:Testable) prop => [Int](https://hackage.haskell.org/package/base-4.16.3.0/docs/Data-Int.html#t:Int) -> prop -> [Property](https://hackage.haskell.org/package/QuickCheck-2.14.3/docs/Test-QuickCheck.html#t:Property)

Discards the test case if it does not complete within the given number of microseconds. This can be useful when testing algorithms that have pathological cases where they run extremely slowly.

So if we add discard after, then the test is discarded. That might be the best solution, since this test succeeds quite quickly in most cases. So if we generate  a pathalogical case, we just discard.

Your call.

lehins: We should fail, instead of discard. 
First of all `discardAfter` could mask a bug that occasionally triggers an infinite loop. Second of all, I always considered discards a hack that people use when they can't figure out how to fix their test properly.

Moreover, even if there is no infinite loop, the pathological case should not take too long to run. 

Note that current value is set to `0.1` second, not `1` second. So increasing it to `2` should give it plenty of time to terminate.

TimSheard: So following Max's advice, we will just double the amount passed to constraintProperty , that way it still protects against infinite loops, and gives extra lee way.

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/Tests.hs
```diff
@@ -708,7 +709,7 @@ defaultWhitelist = ["Size specifications", "The SetSpec's are inconsistent", "Th
 --   constraints.
 prop_soundness' :: Bool -> [String] -> OrderInfo -> Property
 prop_soundness' strict whitelist info =
-  constraintProperty (Just 100000) strict whitelist info $ \preds _graph env ->
+  constraintProperty (Just 200000) strict whitelist info $ \preds _graph env ->
```
lehins: The advise was to remove this argument and a call to `within` from `constraintProperty`. Also .2 seconds is too low. Let's use at least one second.
```suggestion
  within 1_000_000 $
  constraintProperty strict whitelist info $ \preds _graph env ->
```

lehins (changes requested):

lehins (approved):
:+1:

PR #4144, opened 2024-02-29
Add imptests for delaying actions enactment

### LINKED ISSUE #4127: Imp test for checking delaying works as expected
Certain proposal type prevent others from being ratified due to the feature of delaying. We need to write a few unit tests that ensure this indeed happens as expected. Note that there is a corner case of an expiring proposal being delayed (see #4109), that scenario should not be considered in this ticket.

Some scenarios that come to mind:
1. Proposal that triggers a delay prevents another dependent proposal from being enacted until the subsequent epoch, despite both of them having enough votes
2. Proposal delays enactment of all other types of proposals until the next epoch, despite them having enough votes

aniketd:
# Description

Resolves #4127 

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/RatifySpec.hs
+++ b/eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/RatifySpec.hs
@@ -1,0 +1,164 @@
+{-# LANGUAGE NumericUnderscores #-}
+{-# LANGUAGE RankNTypes #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+
+module Test.Cardano.Ledger.Conway.Imp.RatifySpec (spec) where
+
+import Cardano.Ledger.BaseTypes
+import Cardano.Ledger.Coin
+import Cardano.Ledger.Conway.Core
+import Cardano.Ledger.Conway.Governance
+import Data.Default.Class (def)
+import qualified Data.Map.Strict as Map
+import Lens.Micro
+import Test.Cardano.Ledger.Conway.ImpTest
+import Test.Cardano.Ledger.Imp.Common
+
+spec ::
+  forall era.
+  ConwayEraImp era =>
+  SpecWith (ImpTestState era)
+spec =
+  describe "RATIFY" $ do
+    delayingActionsSpec
+
+delayingActionsSpec ::
+  forall era.
+  ConwayEraImp era =>
+  SpecWith (ImpTestState era)
+delayingActionsSpec =
+  describe "Delaying actions" $ do
+    it "A delaying action delays its child even when both ere proposed and ratified in the same epoch" $ do
+      (dRep, committeeMember, _gpi) <- electBasicCommittee
+      modifyPParams $ ppGovActionLifetimeL .~ EpochInterval 5
+      gai0 <- submitInitConstitutionGovAction
+      gai1 <- submitChildConstitutionGovAction gai0
+      gai2 <- submitChildConstitutionGovAction gai1
+      gai3 <- submitChildConstitutionGovAction gai2
+      submitYesVote_ (DRepVoter dRep) gai0
+      submitYesVote_ (CommitteeVoter committeeMember) gai0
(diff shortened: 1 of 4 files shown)


eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/RatifySpec.hs
```diff
@@ -0,0 +1,166 @@
...
+      getLastEnactedParameterChange `shouldReturn` SNothing
+      passEpoch
+      -- all three actions, pGai0, pGai1 and pGai2, are enacted one
+      -- after the other in the same epoch
+      getLastEnactedParameterChange `shouldReturn` SJust (GovPurposeId pGai2)
+      getParameterChangeProposals `shouldReturn` Map.empty
+    describe "An action expires when delayed enough even after being ratified" $ do
+      it "Same lineage" $ do
+        (dRep, committeeMember, _gpi) <- electBasicCommittee
+        conwayModifyPParams $ ppGovActionLifetimeL .~ EpochInterval 2
```
lehins: YOU should be using the type class method instead, also I think this one is redundant, since `electBasicCommittee` already sets the same value
```suggestion
        modifyPParams $ ppGovActionLifetimeL .~ EpochInterval 2
```

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/RatifySpec.hs
```diff
@@ -0,0 +1,166 @@
...
+        passEpoch
+        getLastEnactedConstitution `shouldReturn` SJust (GovPurposeId gai1)
+        passEpoch
+        getLastEnactedConstitution `shouldReturn` SJust (GovPurposeId gai2)
+        getConstitutionProposals `shouldReturn` Map.empty
+        passEpoch
+        getLastEnactedConstitution `shouldReturn` SJust (GovPurposeId gai2)
+      it "Other lineage" $ do
+        (dRep, committeeMember, _gpi) <- electBasicCommittee
+        conwayModifyPParams $ ppGovActionLifetimeL .~ EpochInterval 2
```
lehins: Ditto

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/RatifySpec.hs
```diff
@@ -0,0 +1,166 @@
...
+      passEpoch
+      getLastEnactedConstitution `shouldReturn` SJust (GovPurposeId gai1)
+      passEpoch
+      getLastEnactedConstitution `shouldReturn` SJust (GovPurposeId gai2)
+      passEpoch
+      getLastEnactedConstitution `shouldReturn` SJust (GovPurposeId gai3)
+      getConstitutionProposals `shouldReturn` Map.empty
+    it "A delaying action delays all other actions even when all of them may be ratified in the same epoch" $ do
+      (dRep, committeeMember, _gpi) <- electBasicCommittee
+      void $ conwayModifyPParams $ ppGovActionLifetimeL .~ EpochInterval 5
```
lehins: What's up with `void`? :grin: It returns `()`
```suggestion
      modifyPParams $ ppGovActionLifetimeL .~ EpochInterval 5
```

lehins (approved):
This looks great!

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/RatifySpec.hs
aniketd: Oh that's a left over :sweat: :pray:

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/RatifySpec.hs
aniketd: I thought I would keep it explicit, but I realise after thinking more that it makes little sense to do that, :+1:

PR #4149, opened 2024-02-29
Delay semantics as per spec for expired but already ratified proposals

### LINKED ISSUE #4109: Semantics of how delaying interacts with expiry of GovActions
Current implementation of RATIFY rule removes expired governance actions regardless if they satisfy all of the ratification checks whenever there is another action that triggered a delay.

However, according to the spec, delaying does not trigger the expiration of a proposal whenever a proposal satisfies all of the checks needed for ratification. Which means it is allowed for an expired proposal to be ratified and enacted at any point after expiration, as long as that proposal is valid and it has collected sufficient number of votes.

We need to make sure that the implementation and the spec are in sync. Which means we either need to change the spec or the implementation, which is still to be decided

aniketd:
# Description

Resolves #4109 

- [x] Merge #4144 

# Checklist

- [x] Self-reviewed the diff


lehins:
Original issue #4109 was closed as won't fix

PR #4151, opened 2024-03-04
Fix an issue with txinfo and unusual notation in Babbage UTXO rule

### LINKED ISSUE #4095: Babbage Formal Spec PV2 TxInfo Construction Incorrect for Signatories field
The Babbage formal spec states that the Signatories field of the TxInfo construction uses the key hashes from the transaction witnesses, but I think this should actually say `reqSignerHashes` from the transaction body.

image

WhatisRT:
# Description

Fixes #4095 and replace an unusual notation with something standard.

# Checklist

- [x] Self-reviewed the diff


lehins (approved):
LGTM
Thank you

PR #4152, opened 2024-03-04
Add extra type-safety to ensure that all predicate failures have roundtrip tests

### LINKED ISSUE #3866: Add serialization tests for all Conway predicate failures
We received a bug report (#3865) where we produce `ValueNotConservedUTxO` with zero ADA, which later fails to deserialize. This kind of bug should have been caught by serialization tests

Soupstraw:
# Description

This PR adds the `RuleListEra` typeclass, which has an associated type level list with all the (fallible) rules in that era. This gives us a closed world view of the rules and that makes it easier to write tests that iterate over each rule. See `roundTripAllPredicateFailures` for an example.

resolves #3866 

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/libs/cardano-ledger-core/testlib/Test/Cardano/Ledger/Core/Binary/RoundTrip.hs
+++ b/libs/cardano-ledger-core/testlib/Test/Cardano/Ledger/Core/Binary/RoundTrip.hs
@@ -1,10 +1,20 @@
 {-# LANGUAGE AllowAmbiguousTypes #-}
+{-# LANGUAGE DataKinds #-}
 {-# LANGUAGE FlexibleContexts #-}
-{-# LANGUAGE MonoLocalBinds #-}
+{-# LANGUAGE FlexibleInstances #-}
+{-# LANGUAGE GADTs #-}
+{-# LANGUAGE MultiParamTypeClasses #-}
+{-# LANGUAGE RankNTypes #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE TypeApplications #-}
+{-# LANGUAGE TypeFamilies #-}
+{-# LANGUAGE TypeOperators #-}
+{-# LANGUAGE UndecidableInstances #-}
 
 module Test.Cardano.Ledger.Core.Binary.RoundTrip (
+  RuleListEra (..),
+  EraRuleProof (..),
+
   -- * Spec
   roundTripEraSpec,
   roundTripAnnEraSpec,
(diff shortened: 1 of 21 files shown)


libs/cardano-ledger-core/testlib/Test/Cardano/Ledger/Core/Binary/RoundTrip.hs
```diff
@@ -195,3 +208,55 @@ roundTripCoreEraTypesSpec = do
...
     roundTripShareEraTypeSpec @era @UTxO
+
+data EraRuleProof era (rs :: [Symbol]) where
+  EraRuleProofEmpty :: EraRuleProof era '[]
+  EraRuleProofHead ::
+    ( Show (PredicateFailure (EraRule r era))
+    , Eq (PredicateFailure (EraRule r era))
+    , EncCBOR (PredicateFailure (EraRule r era))
+    , DecCBOR (PredicateFailure (EraRule r era))
+    , Arbitrary (PredicateFailure (EraRule r era))
```
lehins: Let's switch it to `EraRuleFailure`. This will improve inference a bit and will test that we didn't mess up any `EraRuleFailure` instances
```suggestion
    ( Show (EraRuleFailure r era)
    , Eq (EraRuleFailure r era)
    , EncCBOR (EraRuleFailure r era)
    , DecCBOR (EraRuleFailure r era)
    , Arbitrary (EraRuleFailure r era)
    , EraRuleFailure r era ~ PredicateFailure (EraRule r era)
```

lehins (commented):
This is looking great!!! :smiley:

eras/conway/impl/src/Cardano/Ledger/Conway/Era.hs
```diff
@@ -144,5 +145,6 @@ type instance EraRule "SNAP" (ConwayEra c) = ShelleySNAP (ConwayEra c)
 type instance EraRule "TICK" (ConwayEra c) = ShelleyTICK (ConwayEra c)
 
 type instance EraRule "POOL" (ConwayEra c) = ShelleyPOOL (ConwayEra c)
+type instance EraRuleFailure "POOL" (ConwayEra c) = ShelleyPoolPredFailure (ConwayEra c)
```
lehins: We need to define this instance in a separate module for consistency with all previous eras, eg. see: https://github.com/IntersectMBO/cardano-ledger/blob/master/eras/allegra/impl/src/Cardano/Ledger/Allegra/Rules/Pool.hs 

Also, because it will be defined as an orphan the module needs to be imported to `Cardano.Ledger.Conway.Rules` as so: `import Cardano.Ledger.Conway.Rules.Pool ()`
```suggestion
```

lehins (changes requested):
This looks great. Exactly what I envisioned. However, we need the same testing for all previous eras as well not just Conway.

lehins (approved):
Beautiful!

Issue #4006, opened 2024-01-24
HardFork into a new era


lehins:
Current implementation  of Conway will successfully update the protocol version in the Params using HardForkInitiation governance action, which would probably be enough for an intra-era hardfork. The same can't be said about the protocol version update that initiates transition into a new era. The reason for this deficiency is that consensus HardFork combinator relies on internals of the ledger rules and logic of predicting the protocol version update, which has completely changed in Conway.

This needs to be fixed by removing this logic from consensus codebase and abstracting away in ledger in an era agnostic manner the question of: "what will be the protocol version on the next epoch boundary?"

Relevant Consensus ticket: https://github.com/IntersectMBO/ouroboros-consensus/issues/920

PR #4159, opened 2024-03-04
Shelley rules cleanup

lehins:
# Description

Collection of improvements to Shelley rules. This work is done as part of preparation to implementing #4006 

Probably easier to review commit by commit

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Epoch.hs
+++ b/eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Epoch.hs
@@ -27,6 +27,7 @@ import Cardano.Ledger.Shelley.Core
 import Cardano.Ledger.Shelley.Era (ShelleyEPOCH)
 import Cardano.Ledger.Shelley.LedgerState (
   EpochState,
+  LedgerState,
   PState (..),
   UTxOState (utxosDeposited, utxosGovState),
   curPParamsEpochStateL,
@@ -43,7 +44,7 @@ import Cardano.Ledger.Shelley.LedgerState (
   pattern CertState,
   pattern EpochState,
  )
-import Cardano.Ledger.Shelley.LedgerState.Types (esAccountStateL, prevPParamsEpochStateL)
+import Cardano.Ledger.Shelley.LedgerState.Types (prevPParamsEpochStateL)
 import Cardano.Ledger.Shelley.Rewards ()
 import Cardano.Ledger.Shelley.Rules.PoolReap (
   ShelleyPOOLREAP,
@@ -129,7 +130,7 @@ instance
   , State (EraRule "POOLREAP" era) ~ ShelleyPoolreapState era
   , Signal (EraRule "POOLREAP" era) ~ EpochNo
   , Embed (EraRule "UPEC" era) (ShelleyEPOCH era)
-  , Environment (EraRule "UPEC" era) ~ EpochState era
+  , Environment (EraRule "UPEC" era) ~ LedgerState era
   , State (EraRule "UPEC" era) ~ UpecState era
   , Signal (EraRule "UPEC" era) ~ ()
   , Default (PParams era)
@@ -164,7 +165,7 @@ epochTransition ::
   , State (EraRule "POOLREAP" era) ~ ShelleyPoolreapState era
   , Signal (EraRule "POOLREAP" era) ~ EpochNo
   , Embed (EraRule "UPEC" era) (ShelleyEPOCH era)
-  , Environment (EraRule "UPEC" era) ~ EpochState era
+  , Environment (EraRule "UPEC" era) ~ LedgerState era
   , State (EraRule "UPEC" era) ~ UpecState era
   , Signal (EraRule "UPEC" era) ~ ()
   , GovState era ~ ShelleyGovState era
@@ -184,9 +185,8 @@ epochTransition = do
       ) <-
     judgmentContext
   let pp = es ^. curPParamsEpochStateL
-  let pr = es ^. prevPParamsEpochStateL
-  let utxoSt = lsUTxOState ls
-  let CertState vstate pstate dstate = lsCertState ls
+      utxoSt = lsUTxOState ls
+      CertState vstate pstate dstate = lsCertState ls
   ss' <-
     trans @(EraRule "SNAP" era) $ TRC (SnapEnv ls pp, ss, ())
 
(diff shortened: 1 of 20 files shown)


eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Newpp.hs
```diff
@@ -92,26 +89,32 @@ newPpTransition ::
...
-            totalObligation
-              (CertState def pstate dstate)
-              (utxoSt ^. utxosGovStateL)
-      Coin oblgCurr
-        == utxosDeposited utxoSt
-          ?! UnexpectedDepositPot (Coin oblgCurr) (utxosDeposited utxoSt)
+  -- TODO: remove this predicate check. See #4158
+  obligationCurr
+    == utxosDeposited utxoState
+      ?! UnexpectedDepositPot obligationCurr (utxosDeposited utxoState)
```
teodanciu: Is it not an issue that we are changing the conditions under which this PredicateFailure is returned?  (before: only if ppNew was a `Just` and now: anytime). 
Or because we don't expect this to happen anyway, it doesn't matter?

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Newpp.hs
```diff
@@ -92,26 +89,32 @@ newPpTransition ::
...
-        then pure $ NewppState ppNew' (updatePpup ppupSt ppNew')
-        else pure $ NewppState pp (updatePpup ppupSt pp)
-    Nothing -> pure $ NewppState pp (updatePpup ppupSt pp)
+  case mppNew of
+    Just ppNew
+      | toInteger (ppNew ^. ppMaxTxSizeL)
+          + toInteger (ppNew ^. ppMaxBHSizeL)
+          < toInteger (ppNew ^. ppMaxBBSizeL) ->
+          pure $ NewppState ppNew $ updatePpup ppupState ppNew
+    _ -> pure $ NewppState pp ppupStateReset
```
teodanciu: So not only we're not running the `hasLegalProtVerUpdate` check for empty new pparams, but also we are changing the value that gets set? 
Now we are setting directly `emptyPPPUpdates`, but previously, if the `hasLegalProtVerUpdate`  passed, we still set the proposals. 
In what circumstances are we expecting the signal to be empty?  I can see it being set to None in `Upec`, but I don't know what the meaning of it is. 
Just curious what triggered this change, was it causing problems or it's an improvement (removing unnecessary things). But if it's too complicated to explain we can talk about it when you're back, no worries.

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Upec.hs
```diff
@@ -125,24 +125,21 @@ votedValue ::
...
+   in case Map.keys consensus of
         -- NOTE that `quorumN` is a global constant, and that we require
         -- it to be strictly greater than half the number of genesis nodes.
         -- The keys in the `pup` correspond to the genesis nodes,
         -- and therefore either:
         --   1) `consensus` is empty, or
         --   2) `consensus` has exactly one element.
-        1 ->
-          (Just . applyPPUpdates pps . fst . head . Map.toList)
-            consensus
```
teodanciu, line 148: What a strange way to express this!! Happy to see it go 💥

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Epoch.hs
```diff
@@ -202,18 +202,11 @@ epochTransition = do
...
 
   let adjustedCertState = CertState vstate pstate'' dstate'
-      epochState' =
-        EpochState
-          acnt'
-          (ls {lsUTxOState = utxoSt', lsCertState = adjustedCertState})
-          ss'
-          nm
-          & prevPParamsEpochStateL .~ pr
-          & curPParamsEpochStateL .~ pp
```
teodanciu, line 212: I always hated these last two lines! They seem overly defensive but also i felt uneasy to remove them (because what if something gets changed and we forget to set them).
But this now can't happen, good riddance!

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Epoch.hs
```diff
@@ -225,7 +218,7 @@ epochTransition = do
     oblgNew = totalObligation adjustedCertState (utxoSt'' ^. utxosGovStateL)
     utxoSt''' = utxoSt'' {utxosDeposited = oblgNew}
   pure $
-    epochState'
+    EpochState acnt' ls' ss' nm
       & esAccountStateL .~ acnt'
```
teodanciu: Is this line still necessary, since we're building `EpochState` with `acnt'` a line above?

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Upec.hs
```diff
@@ -105,7 +100,7 @@ instance
         coreNodeQuorum <- liftSTS $ asks quorum
 
         let utxoState = lsUTxOState ls
-            pup = proposals $ utxosGovState utxoState
+            pup = proposals ppupState
```
teodanciu: Is this govState `ppupState` from the  `UpecState` the same with the one coming from the environment (as previously used) ?

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Newpp.hs
lehins: It should not matter, since that is the property that should always hold. However, just to be safe I preserve previous semantics. Nice catch.

teodanciu (approved):
Looks good to me, just a few questions.

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Epoch.hs
lehins: Nice catch. It is most certainly redundant

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Newpp.hs
lehins: Thank you for catching this. It was a mistake on my part. With this change we would have lost valid proposals that were from the previous epoch. This is where rotation happens and this call to `updatePpup` is actually necessary

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Upec.hs
lehins: Yes, if you look at the epoch rule, you'll see they are both coming from `utxoSt'`
```haskell
      ls' = ls {lsUTxOState = utxoSt', lsCertState = adjustedCertState}

  UpecState pp' ppupSt' <-
    trans @(EraRule "UPEC" era) $
      TRC (ls', UpecState pp (utxosGovState utxoSt'), ())
```

Same with the TICKF:
```haskell
          ls = esLState es
          updates = utxosGovState $ lsUTxOState ls
      UpecState pp' _ <-
        trans @(EraRule "UPEC" era) $
          TRC (ls, UpecState pp updates, ())
```

teodanciu: Makes sense, thanks!

PR #4160, opened 2024-03-04
Make shelley `TxWits` consistent with other eras

### LINKED ISSUE #3959: Make `TxWits` from `Cardano.Ledger.Shelley.TxWits` consistent with other eras
Definition of witness in Shelley is very outdated and inconsistent with newer eras. We need to fix that. Some of the things that need to be done:

* Remove usage of `HKD`
* Switch to using `MemoBytes`

In other words make it look and feel similar to  `Cardano.Ledger.Alonzo.TxWits`

Lucsanszky:
# Description
Resolves #3959

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/eras/shelley/impl/src/Cardano/Ledger/Shelley/TxWits.hs
+++ b/eras/shelley/impl/src/Cardano/Ledger/Shelley/TxWits.hs
@@ -6,7 +6,6 @@
 {-# LANGUAGE FlexibleInstances #-}
 {-# LANGUAGE GADTs #-}
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
-{-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE MultiParamTypeClasses #-}
 {-# LANGUAGE PatternSynonyms #-}
 {-# LANGUAGE RankNTypes #-}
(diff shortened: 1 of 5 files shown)


Lucsanszky:
I also moved `mapTraverseableDecoderA` from `Alonzo.TxWits` to `Shelley.TxWits`. Although, it might be better to move it to `Decoding` in `cardano-ledger-binary`. Any suggestions?

I'll also update the `CHANGELOG`s and then this will be ready for review.

lehins (commented):
Looking good.

> I also moved mapTraverseableDecoderA from Alonzo.TxWits to Shelley.TxWits. 

That's totally fine


> Although, it might be better to move it to Decoding in cardano-ledger-binary. Any suggestions?

In not so distant future we will be getting rid of `Annotator`, so we don't really need to worry about the ideal placement for this function.

eras/shelley/impl/src/Cardano/Ledger/Shelley/TxWits.hs
```diff
@@ -249,23 +220,26 @@ shelleyEqTxWitsRaw txWits1 txWits2 =
...
 --     Uses the non-exported WitnessSet' constructor.
 prettyWitnessSetParts ::
+  EraScript era =>
   ShelleyTxWits era ->
   ( Set (WitVKey 'Witness (EraCrypto era))
   , Map (ScriptHash (EraCrypto era)) (Script era)
   , Set (BootstrapWitness (EraCrypto era))
   )
-prettyWitnessSetParts (ShelleyTxWitsConstr (WitnessSet' a b c _)) = (a, b, c)
+prettyWitnessSetParts (ShelleyTxWits a b c) = (a, b, c)
```
lehins: I think we can remove this function and just use the `ShelleyTxWits` or lenses in the place where it is being used in the tests

lehins (approved):
Perfect! :+1:

PR #4169, opened 2024-03-06
Add `GovInfoEvent` and add event testing capabilities to `ImpTest`

### LINKED ISSUE #4165: Add a NewEpoch event for proposals and their refunds
In the EPOCH transition for Conway, the enacted and expired proposals refund the active stake addresses. It would be nice to have a ledger event with the ids of these proposals as well as the returned and unclaimed refunds. This could be a nice place for the event https://github.com/IntersectMBO/cardano-ledger/blob/543e734207ebde5cb97edcdafd12974206a72e72/eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Epoch.hs#L316

Soupstraw:
# Description

This PR adds an event to `EPOCH` which contains sets with expired and enacted governance actions and a set with all the governance action IDs that are unclaimed.

I also added a way to write tests about events and cleaned up Conway events. The `ConwayUtxosEvent`is necessary because we have to get rid of `PPUP` events in Conway as we no longer have a `PPUP` rule.

resolves #4165

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/eras/allegra/impl/CHANGELOG.md
+++ b/eras/allegra/impl/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## 1.4.0.0
 
+* Add type `EraRuleEvent` instances for `LEDGER` and `TICK` events
+* Add `Eq` and `NFData` instances for `AllegraUtxoEvent`
 * Add instances for `InjectRuleFailure` and switch to using `injectFailure`
 * Add `shelleyToAllegraUtxoPredFailure`
 * Add `NFData` instance for `AllegraUtxoPredFailure`
@@ -9,6 +11,7 @@
 
 ### `testlib`
 
+* Add `ToExpr` instance for `AllegraUtxoEvent`
 * Add `RuleListEra` instance for Allegra
 
 ## 1.3.0.0
--- a/eras/conway/impl/src/Cardano/Ledger/Conway/Era.hs
+++ b/eras/conway/impl/src/Cardano/Ledger/Conway/Era.hs
@@ -52,20 +52,24 @@ type instance Value (ConwayEra c) = MaryValue c
 -------------------------------------------------------------------------------
 
 type instance EraRule "UPEC" (ConwayEra c) = VoidEraRule "UPEC" (ConwayEra c)
+type instance EraRuleFailure "UPEC" (ConwayEra c) = VoidEraRule "UPEC" (ConwayEra c)
+type instance EraRuleEvent "UPEC" (ConwayEra c) = VoidEraRule "UPEC" (ConwayEra c)
 
 type instance EraRule "NEWPP" (ConwayEra c) = VoidEraRule "NEWPP" (ConwayEra c)
+type instance EraRuleFailure "NEWPP" (ConwayEra c) = VoidEraRule "NEWPP" (ConwayEra c)
+type instance EraRuleEvent "NEWPP" (ConwayEra c) = VoidEraRule "NEWPP" (ConwayEra c)
 
 type instance EraRule "PPUP" (ConwayEra c) = VoidEraRule "PPUP" (ConwayEra c)
 type instance EraRuleFailure "PPUP" (ConwayEra c) = VoidEraRule "PPUP" (ConwayEra c)
+type instance EraRuleEvent "PPUP" (ConwayEra c) = VoidEraRule "PPUP" (ConwayEra c)
 
 type instance EraRule "MIR" (ConwayEra c) = VoidEraRule "MIR" (ConwayEra c)
 type instance EraRuleFailure "MIR" (ConwayEra c) = VoidEraRule "MIR" (ConwayEra c)
+type instance EraRuleEvent "MIR" (ConwayEra c) = VoidEraRule "MIR" (ConwayEra c)
 
 type instance EraRule "DELEGS" (ConwayEra c) = VoidEraRule "DELEGS" (ConwayEra c)
 type instance EraRuleFailure "DELEGS" (ConwayEra c) = VoidEraRule "DELEGS" (ConwayEra c)
-
-type instance EraRule "DELEGS" (ConwayEra c) = VoidEraRule "DELEGS" (ConwayEra c)
-type instance EraRuleFailure "DELEGS" (ConwayEra c) = VoidEraRule "DELEGS" (ConwayEra c)
+type instance EraRuleEvent "DELEGS" (ConwayEra c) = VoidEraRule "DELEGS" (ConwayEra c)
 
 -------------------------------------------------------------------------------
 -- Era Mapping
--- a/eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Deleg.hs
+++ b/eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Deleg.hs
@@ -15,7 +15,6 @@
 
 module Cardano.Ledger.Conway.Rules.Deleg (
   ConwayDELEG,
-  ConwayDelegEvent (..),
   ConwayDelegPredFailure (..),
 ) where
 
@@ -58,6 +57,7 @@ import Control.State.Transition (
  )
 import qualified Data.Map.Strict as Map
 import qualified Data.Set as Set
+import Data.Void (Void)
 import GHC.Generics (Generic)
 import Lens.Micro ((^.))
 import NoThunks.Class (NoThunks)
@@ -101,8 +101,6 @@ instance Era era => DecCBOR (ConwayDelegPredFailure era) where
     5 -> SumD DRepAlreadyRegisteredForStakeKeyDELEG <! From
     n -> Invalid n
 
-newtype ConwayDelegEvent era = DelegEvent (Event (EraRule "DELEG" era))
-
 instance
   ( EraPParams era
   , State (EraRule "DELEG" era) ~ DState era
@@ -117,7 +115,7 @@ instance
   type Environment (ConwayDELEG era) = PParams era
   type BaseM (ConwayDELEG era) = ShelleyBase
   type PredicateFailure (ConwayDELEG era) = ConwayDelegPredFailure era
-  type Event (ConwayDELEG era) = ConwayDelegEvent era
+  type Event (ConwayDELEG era) = Void
 
   transitionRules = [conwayDelegTransition @era]
 
--- a/eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Deleg.hs
+++ b/eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Deleg.hs
@@ -72,6 +72,8 @@ data ConwayDelegPredFailure era
 
 type instance EraRuleFailure "DELEG" (ConwayEra c) = ConwayDelegPredFailure (ConwayEra c)
 
+type instance EraRuleEvent "DELEG" (ConwayEra c) = VoidEraRule "DELEG" (ConwayEra c)
+
 instance InjectRuleFailure "DELEG" ConwayDelegPredFailure (ConwayEra c)
 
 instance NoThunks (ConwayDelegPredFailure era)
--- a/eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Epoch.hs
+++ b/eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Epoch.hs
@@ -1,4 +1,5 @@
 {-# LANGUAGE DataKinds #-}
+{-# LANGUAGE DeriveGeneric #-}
 {-# LANGUAGE DerivingStrategies #-}
 {-# LANGUAGE EmptyCase #-}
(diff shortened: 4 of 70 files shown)


libs/cardano-ledger-core/testlib/Test/Cardano/Ledger/Imp/Common.hs
```diff
@@ -212,6 +214,16 @@ shouldEndWith x y = liftIO $ H.shouldEndWith x y
 shouldContain :: (HasCallStack, Show a, Eq a, MonadIO m) => [a] -> [a] -> m ()
 shouldContain x y = liftIO $ H.shouldContain x y
 
+shouldContainExpr :: (HasCallStack, ToExpr a, Eq a, MonadIO m) => [a] -> [a] -> m ()
+shouldContainExpr x y
+  | y `isInfixOf` x = pure ()
+  | otherwise =
+      assertFailure $
+        "Left does not contain right:\nLeft:\n"
```
teodanciu: I think using `Left` and `Right` here is a little misleading. In the rest of these functions they are used with `Either`, but here it's nothing to do with that, no?

teodanciu:
Will finish reviewing tomorrow, there's a lot here! (some of it above my head)

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Deleg.hs
```diff
@@ -101,8 +101,6 @@ instance Era era => DecCBOR (ConwayDelegPredFailure era) where
     5 -> SumD DRepAlreadyRegisteredForStakeKeyDELEG <! From
     n -> Invalid n
 
-newtype ConwayDelegEvent era = DelegEvent (Event (EraRule "DELEG" era))
```
teodanciu, line 104: We should mention the removal of this in CHANGELOG I think

libs/cardano-ledger-core/src/Cardano/Ledger/Core/Era.hs
```diff
@@ -130,6 +132,8 @@ type family EraRule (rule :: Symbol) era = (r :: Type) | r -> rule
 -- does not provide for us unfortunately.
 type family EraRuleFailure (rule :: Symbol) era = (r :: Type) | r -> rule era
 
+type family EraRuleEvent (rule :: Symbol) era = (r :: Type) | r -> rule era
```
teodanciu, line 135: If it's not too difficult, I think it would be nice to have this change and the changes that it entails in a separate commit, because there seems to be a lot happening here.

teodanciu, line 135: Sorry if it's obvious, but could you please explain what exactly having this type family enables use to do?  Is it related to improving the test framework to support events?

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Epoch.hs
```diff
@@ -172,7 +198,7 @@ returnProposalDeposits removedProposals oldUMap =
                 (RewDepUView um)
             , unclaimed
             )
-        | otherwise = (um, unclaimed <> gasDeposit gas)
+        | otherwise = (um, Map.insertWith (<>) (gasId gas) (gasDeposit gas) unclaimed)
```
teodanciu, line 203: So instead of adding the  values inside the function, we're returning them as a map so we have the values "indexed" by the id, in order to use this info in the Event - is my understanding correct?

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/GovCert.hs
```diff
@@ -140,8 +140,6 @@ instance
         pure (3, ConwayDRepIncorrectRefund refund expectedRefund)
       k -> invalidKey k
 
-newtype ConwayGovCertEvent era = GovCertEvent (Event (EraRule "GOVCERT" era))
```
teodanciu, line 143: We should mention this removal in CHANGELOG

eras/conway/impl/src/Cardano/Ledger/Conway/Era.hs
```diff
@@ -52,20 +52,24 @@ type instance Value (ConwayEra c) = MaryValue c
...
+type instance EraRuleFailure "NEWPP" (ConwayEra c) = VoidEraRule "NEWPP" (ConwayEra c)
+type instance EraRuleEvent "NEWPP" (ConwayEra c) = VoidEraRule "NEWPP" (ConwayEra c)
 
 type instance EraRule "PPUP" (ConwayEra c) = VoidEraRule "PPUP" (ConwayEra c)
 type instance EraRuleFailure "PPUP" (ConwayEra c) = VoidEraRule "PPUP" (ConwayEra c)
+type instance EraRuleEvent "PPUP" (ConwayEra c) = VoidEraRule "PPUP" (ConwayEra c)
 
 type instance EraRule "MIR" (ConwayEra c) = VoidEraRule "MIR" (ConwayEra c)
 type instance EraRuleFailure "MIR" (ConwayEra c) = VoidEraRule "MIR" (ConwayEra c)
+type instance EraRuleEvent "MIR" (ConwayEra c) = VoidEraRule "MIR" (ConwayEra c)
```
teodanciu, line 68: How come we don't have these EraRuleEvent type instances for other eras, for which we have an event, so like MirEvent in Shelley?  Or conversely, why do we need them here?

eras/conway/impl/src/Cardano/Ledger/Conway/Era.hs
```diff
@@ -52,20 +52,24 @@ type instance Value (ConwayEra c) = MaryValue c
...
 
 type instance EraRule "DELEGS" (ConwayEra c) = VoidEraRule "DELEGS" (ConwayEra c)
 type instance EraRuleFailure "DELEGS" (ConwayEra c) = VoidEraRule "DELEGS" (ConwayEra c)
-
-type instance EraRule "DELEGS" (ConwayEra c) = VoidEraRule "DELEGS" (ConwayEra c)
-type instance EraRuleFailure "DELEGS" (ConwayEra c) = VoidEraRule "DELEGS" (ConwayEra c)
```
teodanciu, line 68: Why did we remove these `EraRule` and `EraRuleFailure` instances?

eras/allegra/impl/CHANGELOG.md
```diff
@@ -13,6 +15,7 @@
 
 ## 1.3.0.0
 
+* Add `ToExpr` instance for `AllegraUtxoEvent`
```
teodanciu: Probably should go [below](https://github.com/IntersectMBO/cardano-ledger/pull/4169/files#diff-15549b26d4a93e42ea4073cfa3b2436e03b75f0e9c9e7b97f5e5526964d612d1R26) under "testlib"

eras/shelley/impl/testlib/Test/Cardano/Ledger/Shelley/ImpTest.hs
```diff
@@ -232,6 +236,22 @@ import UnliftIO.Exception (
   throwIO,
  )
 
+data SomeSTSEvent era
```
teodanciu, line 240: The name makes it sound like it's just use for a test, but it's actually something that is part of the testing framework?  I can see it's emitted when submitting an event. Could we find a better name?  What's the purpose of this event?

Soupstraw: I took inspiration from [`SomeException`](https://hackage.haskell.org/package/base-4.19.1.0/docs/Control-Exception.html#t:SomeException), which comes from the `base` package. The idea here is that we hide the actual type inside the existential so it's "some" STS event which has all the constraints that we expect, but we're not able to find out its actual type.

Soupstraw: Actually this approach has some problems right now, because you have to be careful when you make an assertion about events, because the actual event might be injected into some other event. We talked with @lehins about making each event a separate datatype so we can get rid of injections, but we'll do that at some other time.

eras/conway/impl/src/Cardano/Ledger/Conway/Era.hs
Soupstraw: They were already defined on the two lines above

libs/cardano-ledger-core/src/Cardano/Ledger/Core/Era.hs
Soupstraw: The problem is that `STS` isn't aware of the `era` type parameter, so you can't make `Event` injective on `rule era`. Hopefully we can do some refactoring in the future to make an era-aware `STS` class so we can deprecate both `EraRuleFailure` and `EraRuleEvent`.

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Epoch.hs
Soupstraw: Yes, exactly. We need the IDs so we can say which proposals are unclaimed in the `GovInfoEvent`, the behavior of the rule should be unchanged.

libs/cardano-ledger-core/src/Cardano/Ledger/Core/Era.hs
teodanciu: And what do you need the injectivity for? Is it about reusing events in other rules/eras?

eras/conway/impl/src/Cardano/Ledger/Conway/Era.hs
Soupstraw: Ideally they should be defined for all rules in all eras, but right now I only defined `EraRuleEvent` for events that needed it.

Soupstraw: I'll add the others as well so nobody gets confused by this.

libs/cardano-ledger-core/src/Cardano/Ledger/Core/Era.hs
Soupstraw: It helps with type inference so you don't have to add type annotations every time you want to inject an event into another event. GHC has more information to work with so it can derive these types automatically.

teodanciu (approved):
I still don't understand what we wouldn't be able to do without the new `EraRuleEvent`  ( I don't even see any type annotations being removed, if anything, we seem to have added more , like going from:
`mapM_ (tellEvent . SuccessfulPlutusScriptsEvent) (nonEmpty ps)` to `mapM_ (tellEvent . injectEvent @"UTXOS" . SuccessfulPlutusScriptsEvent @era) (nonEmpty ps)`, for example ), 
but I can tell it's the technique used for  `EraRuleFailure`, so I'll approve and hope to understand better one day.

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/EpochSpec.hs
```diff
@@ -304,3 +316,78 @@ depositMovesToTreasuryWhenStakingAddressUnregisters = do
...
+      let actionLifetime = 10
+      modifyPParams $ \pp ->
+        pp
+          & ppGovActionLifetimeL .~ EpochInterval actionLifetime
+          & ppDRepVotingThresholdsL .~ def {dvtPPEconomicGroup = 1 %! 2}
+      propDeposit <- getsNES $ nesEsL . curPParamsEpochStateL . ppGovActionDepositL
+      let
+        proposeCostModel = do
+          newTau <- arbitrary
+          submitGovAction @era $
```
teodanciu: You could probably use `submitParameterChange` if you wanted, though not sure if it would bring much more conciseness

Soupstraw:
> I still don't understand what we wouldn't be able to do without the new `EraRuleEvent` ( I don't even see any type annotations being removed, if anything, we seem to have added more , like going from: `mapM_ (tellEvent . SuccessfulPlutusScriptsEvent) (nonEmpty ps)` to `mapM_ (tellEvent . injectEvent @"UTXOS" . SuccessfulPlutusScriptsEvent @era) (nonEmpty ps)`, for example ), but I can tell it's the technique used for `EraRuleFailure`, so I'll approve and hope to understand better one day.

The main point for adding `EraRuleEvent` is to enable using `injectEvent`. Without it we'd run into the same issue that #4071 was solving, but with events. This `injectEvent` will be useful for writing tests and also makes it easier to reuse existing rules in future eras. It's a bit of an ugly solution, but we should be able to make it better when we refactor after Conway.

libs/cardano-ledger-core/src/Cardano/Ledger/Core/Era.hs
Soupstraw: I think it'd be too much work to split this now, but I'll try to keep that in mind next time

PR #4174, opened 2024-03-06
Convert `small-steps-test` into a sublibrary

### LINKED ISSUE #4068: Convert `small-steps-test` into a sublibrary
Using `cardano-ledger-core` setup as an example, make a `testlib` sublibrary in the `small-steps` package, and move `small-steps-test` functionality into it. Tests should go into a separate `tests` target.

Lucsanszky:
# Description
Resolves #4068 

# Checklist

- [x] Self-reviewed the diff


eras/alonzo/test-suite/cardano-ledger-alonzo-test.cabal
```diff
@@ -1,6 +1,6 @@
 cabal-version: 3.0
 name:          cardano-ledger-alonzo-test
-version:       1.2.0.1
+version:       1.3.0.0
```
teodanciu: Is this a breaking change for this package, or how come we're updating the major  version?

Lucsanszky: Now that you mention it and on a second thought, this might not require a major version bump. Although point 8 in the PVP versioning policy states that when adding a new module from a new package, we should bump the major version to avoid potential name clashes in dependent code but since the newly added module is not re-exported, a minor version bump should be fine here (and really, everywhere else, since the added modules are not reexported anywhere). Thanks!

lehins (approved):
Very nice!!
One less package to worry about!

PR #4175, opened 2024-03-06
Set the `column-limit` in `fourmolu` config

### LINKED ISSUE #4069: Set the `column-limit` in fourmolu
For a while now `fourmolu` had the option of `column-limit`. I believe we should add it to our settings and set the limit to a value of a `100`. 

Also we need to set `respectful` to `false`: https://github.com/IntersectMBO/cardano-ledger/blob/0e65b5e7ad45df9a7ad7ed9bfe8364d5153a2c15/fourmolu.yaml#L13

There was multiple occasions when accidental new line between imports prevented them being formatted properly. It will still be possible to separate import by placing a comment between sections, for those who really want this feature.

Lucsanszky:
# Description
In addition to applying the formatting changes enforced by the added `column-limit` rule, this PR will also:
* Remove the idempotence check when running `fourmolize.sh` as that doesn't play well with `column-limit`.
  See: https://fourmolu.github.io/config/column-limit/
* Add `.git-blame-ignore-revs` file so we can keep track of commits with large formatting changes and ignore them when blaming.
* Make it possible to run `fourmolize.sh` on changed files only by supplying the `--changes` flag.
* Run the `fourmolu` GitHub action on changed files only (compared to `origin/master`).
* Setup `pre-commit` in a non-intrusive way.

For now, I disabled running the `pre-commit-check` as a Hydra job until we decide whether to keep the current setup (running `fourmolize.sh` as a GitHub action) or switch to `pre-commit`. Both options have their advantages, however if at one point we decide to adopt `iogx`, I would advocate for `pre-commit` via Hydra. Regardless, `pre-commit` will be available from now on as a `git hook` and one can opt-in to use it conveniently, albeit in a bit hacky way (see the `SKIP` flag trick detailed in `CONTRIBUTING.md`).

Resolves #4069 

# Checklist

- [x] Self-reviewed the diff


libs/cardano-ledger-core/testlib/Test/Cardano/Ledger/Plutus/Examples.hs
```diff
@@ -258,9 +774,92 @@ evenRedeemer2 =
...
+            , 110
+            , 180
+            , 0
+            , 146
+            , 0
+            , 65
+            , 128
+            , 0
+            , 177
+            ]
```
Lucsanszky, line 862: I'm not sure I'm liking this... :confounded:

Lucsanszky (commented):
Well... There certainly are some interesting formatting changes but I think they are more or less fine. Except for one instance that I pointed out in the code, that one really is an eyesore.

scripts/fourmolize.sh
```diff
@@ -2,6 +2,22 @@
 
 set -euo pipefail
 
-fourmolu -m inplace $(git ls-files -- '*.hs' | grep -v Setup.hs)
+if [[ $# -gt 0 ]]; then
+  case "$1" in
+    --changes)
+      files=$(git diff --diff-filter=MA --name-status origin/master HEAD -- '*.hs' | cut -c 3-)
```
teodanciu: Why not `--name-only` instead of `--name-status` and `cut` after ?

CONTRIBUTING.md
```diff
@@ -290,8 +290,31 @@ and error may be required to make substantive changes!
...
+you can instruct the script to run it only on changed files (compared to `origin/master`)
+by providing the `--changes` flag. If you omit it, then `fourmolu` will format everything.
+You can also run things manually via `pre-commit`. If you previosly unset `SKIP`
+you can just run `pre-commit run --all-files`, if you didn't unset it and only wish
+to unset it for one run, assuming you are using `bash` do the following:
+`SKIP= pre-commit run --all-files`.
+
+You can also just check the formatting without modifying anything by running `nix flake check`
+or if that doesn't work `nix build .#checks.${system}.pre-commit-check`
+(e.g: `nix build .#checks.x86_64-linux.pre-commit-check`).
```
teodanciu, line 317: TIL!  Thanks for the detailed explanations and instructions

libs/cardano-ledger-core/testlib/Test/Cardano/Ledger/Plutus/Examples.hs
teodanciu: Haha, I agree it looks a little awkward!  We could presumably split them in two lists and concatenate them but it's a strange thing to come across. I'd say leave them like this, if they bother us we can come up with something.

.github/workflows/haskell.yml
```diff
@@ -158,7 +158,9 @@ jobs:
         echo "$HOME/.local/bin" >> $GITHUB_PATH
 
     - name: Run fourmolu
-      run: ./scripts/fourmolize.sh
+      run: |
+        git fetch --no-tags --prune --depth=1 origin +refs/heads/master:refs/remotes/origin/master
```
teodanciu, line 162: Why this fetch? I imagine it's gonna have some impact on the `git diff` ran below as part of `formolize.sh`, but how will it make a difference?

teodanciu (approved):
Looks good to me, very generous explanations! 
I'm personally not going to use the pre-commit hook but I will certainly use the `--changes` flag with fourmolu, it will be great to not have to wait so long for the script.

scripts/fourmolize.sh
Lucsanszky: Good catch, thank you! :raised_hands:

.github/workflows/haskell.yml
Lucsanszky: It is needed for the GitHub action. By default, the `checkout@v4` action only fetches a single commit from the feature branch, see: https://github.com/actions/checkout. However, since we want to run `fourmolu` on changed files only compared to `origin/master`, we need to have access to `origin/master` and this is how we achieve that. Otherwise, the action would fail and complain with `bad revision: origin/master`.

libs/cardano-ledger-core/testlib/Test/Cardano/Ledger/Plutus/Examples.hs
lehins: This code is generated by a preprocessor, so I don't think it needs to be pretty or anything.
That being said if we want to make it more readable we could switch to decoding a hex string, rather than packing a list of 8 bit words. This would be more concise.

Issue #3559, opened 2023-07-20
Babbage header CDDL possibly incorrect


rooooooooob:
In the CDDL we have:
```cddl
header_body =
  [ block_number     : uint
  , slot             : uint
  , prev_hash        : $hash32 / null
  , issuer_vkey      : $vkey
  , vrf_vkey         : $vrf_vkey
  , vrf_result       : $vrf_cert ; New, replaces nonce_vrf and leader_vrf
  , block_body_size  : uint
  , block_body_hash  : $hash32 ; merkle triple root
  , operational_cert
  , protocol_version
  ]

operational_cert =
  ( hot_vkey        : $kes_vkey
  , sequence_number : uint
  , kes_period      : uint
  , sigma           : $signature
  )

protocol_version = (major_protocol_version, uint)
```

which would correspond to a size 14 array for `header_body` as `operational_cert` and `protocol_version` are basic groups (note them having _only_ `(` and `)` instead of `[` and `]` or `[(` and `)]`).
However I am encountering a size 10 array for all babbage blocks where it is treated as if they were written instead as:
```cddl
operational_cert =
  [ hot_vkey        : $kes_vkey
  , sequence_number : uint
  , kes_period      : uint
  , sigma           : $signature
  ]

protocol_version = [major_protocol_version, uint]
```
i.e. they are encoded as array groups instead of raw basic groups.

All previous eras seem to have them encoded as basic groups (i.e. their fields are inlined into the header body). It's possible that there's something going on somewhere in the stack I'm receiving it from but it's supposed to be the raw bytes without changes. If that's the case I'll close the issue, but if not we need to update the `babbage.cddl` file.

lehins:
Block serialization logic has moved over to consensus repository starting with Babbage, so that's why CDDL spec can't be checked in ledger code for the block header. 

I just looked at the [implementation](https://github.com/input-output-hk/ouroboros-consensus/blob/fbbaca14339f25872ffbcf41eeb38bcc4957c73e/ouroboros-consensus-protocol/src/ouroboros-consensus-protocol/Ouroboros/Consensus/Protocol/Praos/Header.hs#L195) and at a glance it does seem the cddl spec is off. We'll do a proper investigation into this and adjust the spec accordingly.

Thanks for reporting this issue.

rooooooooob:
This issue also seems to affect the sancho testnet so conway.cddl will need updating as well unless that gets changed before making it to mainnet.

[...]

PR #4178, opened 2024-03-07
Remove code repetition in Conway era CDDL

iccicci:
# Description

`protocol_version` is defined as a group and always referred nested (alone) inside an array.

I introduced `protocol_version_` with the purpose to not change `protocol_version` definition: I don't know if there are some reasons to keep the definitions unchanged between CDDL versions; if that's not the case and I was too scrupulous, we could simplify this PR removing `protocol_version_` and changing `protocol_version` definition.

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/eras/conway/impl/cddl-files/conway.cddl
+++ b/eras/conway/impl/cddl-files/conway.cddl
@@ -34,7 +34,7 @@ header_body =
   , block_body_size  : uint
   , block_body_hash  : $hash32 ; merkle triple root
   , operational_cert
-  , [ protocol_version ]
+  , protocol_version
   ]
 
 operational_cert =
@@ -48,7 +48,7 @@ next_major_protocol_version = 10
 
 major_protocol_version = 1..next_major_protocol_version
 
-protocol_version = (major_protocol_version, uint)
+protocol_version = [(major_protocol_version, uint)]
 
 transaction_body =
   { 0 : set<transaction_input>             ; inputs
@@ -105,7 +105,7 @@ policy_hash = scripthash
 
 parameter_change_action = (0, gov_action_id / null, protocol_param_update, policy_hash / null)
 
-hard_fork_initiation_action = (1, gov_action_id / null, [protocol_version])
+hard_fork_initiation_action = (1, gov_action_id / null, protocol_version)
 
 treasury_withdrawals_action = (2, { reward_account => coin }, policy_hash / null)
 
--- a/eras/babbage/impl/cddl-files/babbage.cddl
+++ b/eras/babbage/impl/cddl-files/babbage.cddl
@@ -34,7 +34,7 @@ header_body =
   , block_body_size  : uint
   , block_body_hash  : $hash32 ; merkle triple root
   , operational_cert
-  , [ protocol_version ]
+  , protocol_version
   ]
 
 operational_cert =
@@ -48,7 +48,7 @@ next_major_protocol_version = 9
 
 major_protocol_version = 1..next_major_protocol_version
 
-protocol_version = (major_protocol_version, uint)
+protocol_version = [(major_protocol_version, uint)]
 
 transaction_body =
   { 0 : set<transaction_input>    ; inputs
@@ -280,7 +280,7 @@ protocol_param_update =
   , ? 9: nonnegative_interval ; pool pledge influence
   , ? 10: unit_interval       ; expansion rate
   , ? 11: unit_interval       ; treasury growth rate
-  , ? 14: [protocol_version]  ; protocol version
+  , ? 14: protocol_version    ; protocol version
   , ? 16: coin                ; min pool cost
   , ? 17: coin                ; ada per utxo byte
   , ? 18: costmdls            ; cost models for script languages


eras/conway/impl/cddl-files/conway.cddl
```diff
@@ -34,7 +34,7 @@ header_body =
   , block_body_size  : uint
   , block_body_hash  : $hash32 ; merkle triple root
   , operational_cert
-  , [ protocol_version ]
+  , protocol_version_
```
teodanciu: Out of curiosity, is the underscore suffix a standard or commonly used way to name singleton lists in cddl? I haven't encountered it before

teodanciu:
I'm not convinced of the benefits of this change. 
`protocol_version_` name is a little inexpressive, in my opinion, for a singleton list of protocol_version (unless there is some convention I'm not aware of). So whenever it's encountered, it has to be looked up to understand what it is. 
Maybe we could find a better name for it?

On the other hand, defining protocol_version itself as a singleton list doesn't seem right either - whereas it would reduce duplication, it can be misleading when reading the types that are referencing it. 

But maybe I'm missing something - is this duplication creating problems in some ways that are not obvious?   Just curious

iccicci:
> On the other hand, defining `protocol_version` itself as a singleton list

From this sentence it seems there are no reasons to keep unchanged definitions between eras; if I am correct, I would prefer this option both for shortness and for clarity.

> it can be misleading when reading the types that are referencing it.

I can't get the point here. Wouldn't it be just as any other defined type? Could you please elaborate? Thank you

> `protocol_version_` name is a little inexpressive...
> Maybe we could find a better name for it?

For sure! Maybe `protocol_version_obj` or `protocol_sem_ver` or any other suggestion will be appreciated.
Anyway, until is possible to change the current definition of `protocol_version`, I would focus on that option.

> is this duplication creating problems in some ways that are not obvious?

In my specific case, I spent more or less one hour trying to understand the reason for this repetition.. and I ended up opening this PR.
In [cardano-js-sdk](https://github.com/input-output-hk/cardano-js-sdk) we started a process to make the SDK as compliant as possible to the Ledger: the presence of this repetition reflects in _forcing_ us to replicate the repetition in SDK or to tolerate a discrepancy (if we want to avid the code repetition).

Thank you

eras/conway/impl/cddl-files/conway.cddl
```diff
@@ -48,7 +48,12 @@ next_major_protocol_version = 10
 
 major_protocol_version = 1..next_major_protocol_version
 
-protocol_version = (major_protocol_version, uint)
+protocol_version =
+  ( major_protocol_version
+  , minor_protocol_version : uint
+  )
+
+protocol_version_ = [ protocol_version ]
```
lehins: There is no place where `(major_protocol_version, uint)` is used without wrapping into an array. Therefore I do not see a reason for this extra indirection.
```suggestion
protocol_version = [(major_protocol_version, uint)]
```

If you'd like to fix this for all previous eras that would be great too. This would make it consistent across eras

lehins (commented):
> I don't know if there are some reasons to keep the definitions unchanged between CDDL versions; if that's not the case and I was too scrupulous, we could simplify this PR removing protocol_version_ and changing protocol_version definition.

I am all in favor of this suggestion.

iccicci:
> is this duplication creating problems in some ways that are not obvious?

Please let me elaborate my thought about this question.

This document is not an end in itself, it is **the refernce** for thousands of other projects. For this reason, in my opinion, it should be kept as clear and minimal as possible.

If I may allow one more suggestion, I would move the big comment it has in its middle to a `README.md` file changing the big comment with a short comment with just a link referencing the section of the chosen README.

teodanciu:
Thank you for the explanation @iccicci,  I concede to the non-indirection approach that you and Alexey suggested, as my concern seems indeed unfounded.

eras/conway/impl/cddl-files/conway.cddl
iccicci: Wait! This suggestion opened my eyes.

Actually I never dealt with an `header_body`, so actually I have no idea of any implications in that context...

The idea of opening this PR cames from the fact that `cardano-node v8.8-pre` accepts transactions encoded with this CBOR:

```
83                          # array(3)
   01                       # unsigned(1)
   F6                       # primitive(22)
   82                       # array(2)
      09                    # unsigned(9)
      00                    # unsigned(0)
```
(if I understood correctly, a CBOR reflecting following syntax)
```
protocol_version = (major_protocol_version, uint)

hard_fork_initiation_action = (1, gov_action_id / null, protocol_version)
```

while if we try to submit a transaction encoded with following CBOR, we get a deserialization error.

```
83                          # array(3)
   01                       # unsigned(1)
   F6                       # primitive(22)
   81                       # array(1)
      82                    # array(2)
         09                 # unsigned(9)
         00                 # unsigned(0)
```
(if I understood correctly, a CBOR reflecting following syntax - current syntax)
```
protocol_version = (major_protocol_version, uint)

hard_fork_initiation_action = (1, gov_action_id / null, [ protocol_version ])
```

Could it be the problem is just a misalignment between this document and `cardano-node`?

lehins: I am not sure I understand the question, but this is the correct serialization for the hard fork initiation:
```
protocol_version = (major_protocol_version, uint)
hard_fork_initiation_action = (1, gov_action_id / null, [protocol_version])
```
or this
```
hard_fork_initiation_action = (1, gov_action_id / null, [major_protocol_version, uint])
```
Both are equivalent and match this cbor:
```
83                          # array(3)
   01                       # unsigned(1)
   F6                       # primitive(22)
   82                       # array(2)
      09                    # unsigned(9)
      00                    # unsigned(0)
```

iccicci: Thank you for your explaination and please ignore my previous question, I'm still learning about CBOR and CDDL...

43b682412 should address your suggestion.

Last, about your proposal:

> If you'd like to fix this for all previous eras that would be great too. This would make it consistent across eras

Actually I tried, but after reading #3559 and #3762 I was no longer sure about the change.

If you can confirm that it is correct to change `protocol_version` and how it is referenced in `header_body` like in this PR for all the eras, I'll be happy to do that.

Thank you

lehins: Oh, yeah. If you'd like you could apply the same change you did here to the Babbage era, but earlier eras had a different encoding in the block body header. They used [grouping](https://github.com/IntersectMBO/cardano-ledger/blob/bc2fdce909fe27e8af93c5c308fda9b16a861bc3/libs/cardano-protocol-tpraos/src/Cardano/Protocol/TPraos/BHeader.hs#L250), which essentially unwraps and inlines the elements of the array. So let's leave pre-Babbage eras as is.

iccicci: f584a37de should address this.

Thank yuo for your patience.

teodanciu:
Rebased against master

teodanciu (approved):
Looks good to me

Issue #4128, opened 2024-02-27
Imp test for competing proposals of the same and different priority


lehins:
We need multiple tests that check interaction of proposals of the same and different priority competing for enactment. 

1. Proposal that was submitted first will be enacted first (i.e. will win) when multiple **competing** proposals of the same priority get enough votes during the same epoch.
2. Proposal that was higher priority but submitted last will be enacted first (i.e. will win) when multiple **competing** proposals of different priority get enough votes during the same epoch (eg. NoConfidence and `UpdateCommittee`).

lehins:
Resolved in #4201

PR #4179, opened 2024-03-07
Imptests: negative treasury, action priorities

### LINKED ISSUE #4133: Imp test that treasury withdrawal cannot put treasury into negative
Submitting a few TreasuryWithdrawal proposals that are by itself within the limit could put the treasury into negative balance if we didn't have appopriate checks. 
We need to write a couple of unit tests that confirms this check:

* Two treasury withdrawals enacted in the same epoch that by themselves are smaller than treasury amount would together put treasury into negative result. In this case only one treasury withdrawal should be enacted, until treasury amount is sufficiently large (eg topped off with a donation)
* Multiple treasury withdrawals that individually are small enough, but together so large that would overflow `maxBound :: Word64`, still result in enactment only as many treasury withdrawals as the actual treasury balance allows

aniketd:
# Description

Resolves:  #4133

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/EnactSpec.hs
+++ b/eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/EnactSpec.hs
@@ -17,9 +17,12 @@ import Cardano.Ledger.Conway.PParams
 import Cardano.Ledger.Conway.Rules
 import Cardano.Ledger.Credential
 import Cardano.Ledger.Shelley.LedgerState
+import Cardano.Ledger.Val
 import Data.Default.Class (def)
 import qualified Data.Map.Strict as Map
 import qualified Data.Sequence as Seq
+import qualified Data.Set as Set
+import Data.Word (Word64)
 import Lens.Micro
 import Test.Cardano.Ledger.Conway.ImpTest
 import Test.Cardano.Ledger.Core.Rational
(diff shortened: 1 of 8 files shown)


eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/ImpTest.hs
```diff
@@ -1262,3 +1269,9 @@ majorFollow pv@(ProtVer x _) = case succVersion x of
 -- | An illegal ProtVer that skips 3 minor versions
 cantFollow :: ProtVer -> ProtVer
 cantFollow (ProtVer x y) = ProtVer x (y + 3)
+
+checkProposalsEmpty :: ConwayEraGov era => ImpTestM era ()
+checkProposalsEmpty =
+  null . proposalsIds
+    <$> getProposals
+      `shouldReturn` True
```
lehins, line 1288: Looks like a stair case :slightly_smiling_face: 
```suggestion
  null . proposalsIds <$> getProposals `shouldReturn` True
```

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/EnactSpec.hs
```diff
@@ -33,6 +36,171 @@ spec ::
...
+            , (rewardAccount5, Coin eachWithdrawal)
+            ]
+        submitYesVote_ (DRepVoter drepC) gpi
+        submitYesVote_ (CommitteeVoter committeeC) gpi
+        passNEpochs 2
+        getsNES (nesEsL . esAccountStateL . asTreasuryL)
+          `shouldReturn` Coin curTreasury
+        modifyNES $
+          nesEsL . esAccountStateL . asTreasuryL
+            %~ (<+> Coin eachWithdrawal)
```
lehins, line 128: We should not be modifying the state, unless there is no other way around it. It is allowed to modify the state as the very first thing in the test, since that is where we are setting up the initial state, but in the middle of the test we could be interfering with the workflow of the ledger rules

If you want to increase the amount in the treasury, why not submit a treasury donation. 
```suggestion
```

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/EnactSpec.hs
```diff
@@ -33,6 +36,171 @@ spec ::
...
+        rewardAccount4 <- registerRewardAccount
+        rewardAccount5 <- registerRewardAccount
+        let eachWithdrawal = curTreasury `div` 4
+        gpi <-
+          submitTreasuryWithdrawals
+            [ (rewardAccount1, Coin eachWithdrawal)
+            , (rewardAccount2, Coin eachWithdrawal)
+            , (rewardAccount3, Coin eachWithdrawal)
+            , (rewardAccount4, Coin eachWithdrawal)
+            , (rewardAccount5, Coin eachWithdrawal)
```
lehins, line 119: Using exact same values can hide subtle bugs. It would be better to have those amounts be slightly different, potentially even generated at random. 

Think in terms of property tests, rather than unit tests.

However if you find it results in a test that is too complicated then don't worry about it.

lehins (commented):
I see tests for #4133 but none for #4128
Those are two totally unrelated tickets, so they shouldn't be solved in one PR either

With respect to #4133 it is looking good. However, CI is failing

teodanciu:
The treasury-withdrawals part of this PR is superseded by this PR:https://github.com/IntersectMBO/cardano-ledger/pull/4189, which includes the commit and adds some improvements on it.

teodanciu:
Closing, as superseded by:
 https://github.com/IntersectMBO/cardano-ledger/pull/4189
https://github.com/IntersectMBO/cardano-ledger/pull/4201

PR #4189, opened 2024-03-08
Imptests -  treasury withdrawals

teodanciu:
# Description

This is the first part of @aniketd 's PR: https://github.com/IntersectMBO/cardano-ledger/pull/4179/commits, closing: https://github.com/IntersectMBO/cardano-ledger/issues/4133

I have re-arranged the test a little bit and added his commit that is actually adding the tests, and improved the tests with the comments from the review.

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/EnactSpec.hs
+++ b/eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/EnactSpec.hs
@@ -3,6 +3,7 @@
 {-# LANGUAGE NumericUnderscores #-}
 {-# LANGUAGE OverloadedLists #-}
 {-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TupleSections #-}
 {-# LANGUAGE TypeApplications #-}
 {-# LANGUAGE TypeFamilies #-}
 {-# LANGUAGE TypeOperators #-}
@@ -19,6 +20,7 @@ import Cardano.Ledger.Conway.Rules
 import Cardano.Ledger.Credential
 import Cardano.Ledger.Shelley.LedgerState
 import Cardano.Ledger.Val (zero, (<->))
+import Control.Monad (forM)
 import Control.State.Transition.Extended (STS (..))
 import Data.Default.Class (def)
 import Data.Foldable (foldl', traverse_)
@@ -106,17 +108,12 @@ treasuryWithdrawalsSpec =
       modifyPParams $ ppGovActionLifetimeL .~ EpochInterval 5
       initialTreasury <- getTreasury
       numWithdrawals <- choose (1, 10)
-      withdrawals <- genWithdrawalsExceeding initialTreasury numWithdrawals 100
+      withdrawals <- genWithdrawalsExceeding initialTreasury numWithdrawals
 
-      gaId <- submitTreasuryWithdrawals withdrawals
-      submitYesVote_ (DRepVoter drepC) gaId
-      submitYesVote_ (CommitteeVoter committeeC) gaId
-      passNEpochs 2
-      getTreasury `shouldReturn` initialTreasury
-      -- reward accounts are empty
-      sumRewardAccounts withdrawals `shouldReturn` zero
+      void $ enactTreasuryWithdrawals withdrawals drepC committeeC
+      checkNoWithdrawal initialTreasury withdrawals
 
-      let sumRequested = mconcat $ snd <$> withdrawals
+      let sumRequested = foldMap snd withdrawals
 
       impAnn "Submit a treasury donation that can cover the withdrawals" $ do
         let tx =
@@ -132,29 +129,16 @@ treasuryWithdrawalsSpec =
       modifyPParams $ ppGovActionLifetimeL .~ EpochInterval 5
       initialTreasury <- getTreasury
       numWithdrawals <- choose (1, 10)
-      withdrawals <- genWithdrawalsExceeding (Coin (fromIntegral (maxBound :: Word64))) numWithdrawals 100
-      gaId <- submitTreasuryWithdrawals withdrawals
-
-      submitYesVote_ (DRepVoter drepC) gaId
-      submitYesVote_ (CommitteeVoter committeeC) gaId
-      passNEpochs 2
-      getTreasury `shouldReturn` initialTreasury
-      sumRewardAccounts withdrawals `shouldReturn` zero
+      withdrawals <- genWithdrawalsExceeding (Coin (fromIntegral (maxBound :: Word64))) numWithdrawals
+      void $ enactTreasuryWithdrawals withdrawals drepC committeeC
+      checkNoWithdrawal initialTreasury withdrawals
 
     it "Withdrawals exceeding treasury submitted in several proposals within the same epoch" $ do
       (drepC, committeeC, _) <- electBasicCommittee
       modifyPParams $ ppGovActionLifetimeL .~ EpochInterval 5
       initialTreasury <- getTreasury
-      -- generate withdrawals which individually do not exceed the treasury, but their sum does
       numWithdrawals <- choose (1, 10)
-      withdrawals <- genWithdrawalsExceeding initialTreasury numWithdrawals 50
-
-      impAnn "submit in the same proposal, with no effect on the treasury" $ do
-        gaId <- submitTreasuryWithdrawals withdrawals
-        submitYesVote_ (DRepVoter drepC) gaId
-        submitYesVote_ (CommitteeVoter committeeC) gaId
-        passNEpochs 2
-        getTreasury `shouldReturn` initialTreasury
+      withdrawals <- genWithdrawalsExceeding initialTreasury numWithdrawals
 
       impAnn "submit in individual proposals in the same epoch" $ do
         traverse_
@@ -169,7 +153,7 @@ treasuryWithdrawalsSpec =
         let expectedTreasury =
               foldl'
                 ( \acc (_, x) ->
-                    if acc <-> x >= zero
+                    if acc >= x
                       then acc <-> x
                       else acc
                 )
@@ -182,13 +166,15 @@ treasuryWithdrawalsSpec =
   where
     getTreasury = getsNES (nesEsL . esAccountStateL . asTreasuryL)
     sumRewardAccounts withdrawals = mconcat <$> traverse (getRewardAccountAmount . fst) withdrawals
-    genWithdrawalsExceeding (Coin val) n maxExcess = do
-      accounts <- replicateM n $ registerRewardAccount @era
-      pcts <- replicateM n (choose (1, 100) :: ImpTestM era Integer)
+    genWithdrawalsExceeding (Coin val) n = do
+      pcts <- replicateM (n - 1) $ choose (1, 100)
       let tot = sum pcts
-      excess <- choose (1, maxExcess) :: ImpTestM era Integer
-      let amounts = fmap (\x -> Coin $ ceiling ((x + excess) % tot * fromIntegral val)) pcts
-      pure $ zip accounts amounts
+      Positive excess <- arbitrary
+      let amounts = Coin excess : map (\x -> Coin $ ceiling ((x * val) % tot)) pcts
+      forM amounts $ \amount -> (,amount) <$> registerRewardAccount
+    checkNoWithdrawal initialTreasury withdrawals = do
+      getTreasury `shouldReturn` initialTreasury
+      sumRewardAccounts withdrawals `shouldReturn` zero
 
 hardForkInitiationSpec :: ConwayEraImp era => SpecWith (ImpTestState era)
 hardForkInitiationSpec =
--- a/eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/EnactSpec.hs
+++ b/eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/EnactSpec.hs
@@ -31,7 +31,6 @@ import Type.Reflection (Typeable)
 spec ::
   forall era.
   ( ConwayEraImp era
-  , GovState era ~ ConwayGovState era
   , NFData (Event (EraRule "ENACT" era))
   , ToExpr (Event (EraRule "ENACT" era))
   , Eq (Event (EraRule "ENACT" era))
@@ -189,23 +188,20 @@ noConfidenceSpec =
     replicateM_ 4 passEpoch
     assertNoCommittee
 
-constitutionSpec ::
-  forall era.
-  ( ConwayEraImp era
-  , GovState era ~ ConwayGovState era
-  ) =>
-  SpecWith (ImpTestState era)
+constitutionSpec :: ConwayEraImp era => SpecWith (ImpTestState era)
 constitutionSpec =
   it "Constitution" $ do
     (dRep, committeeMember, _) <- electBasicCommittee
     (govActionId, constitution) <- submitConstitution SNothing
 
-    ConwayGovState proposalsBeforeVotes enactStateBeforeVotes pulserBeforeVotes _ _ _ <-
-      getsNES $ newEpochStateGovStateL @era
+    proposalsBeforeVotes <- getsNES $ newEpochStateGovStateL . proposalsGovStateL
+    pulserBeforeVotes <- getsNES newEpochStateDRepPulsingStateL
+
     submitYesVote_ (DRepVoter dRep) govActionId
     submitYesVote_ (CommitteeVoter committeeMember) govActionId
-    ConwayGovState proposalsAfterVotes enactStateAfterVotes pulserAfterVotes _ _ _ <-
-      getsNES newEpochStateGovStateL
+
+    proposalsAfterVotes <- getsNES $ newEpochStateGovStateL . proposalsGovStateL
+    pulserAfterVotes <- getsNES newEpochStateDRepPulsingStateL
 
     impAnn "Votes are recorded in the proposals" $ do
       let proposalsWithVotes =
@@ -221,9 +217,6 @@ constitutionSpec =
               )
       proposalsAfterVotes `shouldBe` proposalsWithVotes
 
-    impAnn "EnactState has not changed" $
-      enactStateAfterVotes `shouldBe` enactStateBeforeVotes
-
     impAnn "Pulser has not changed" $
       pulserAfterVotes `shouldBe` pulserBeforeVotes
 
@@ -234,7 +227,7 @@ constitutionSpec =
       constitutionAfterOneEpoch `shouldBe` def
 
     impAnn "Pulser should reflect the constitution to be enacted" $ do
-      ConwayGovState _ _ _ _ _ pulser <- getsNES newEpochStateGovStateL
+      pulser <- getsNES newEpochStateDRepPulsingStateL
       let ratifyState = extractDRepPulsingState pulser
       gasId <$> rsEnacted ratifyState `shouldBe` govActionId Seq.:<| Seq.Empty
       rsEnactState ratifyState ^. ensConstitutionL `shouldBe` constitution
@@ -246,7 +239,7 @@ constitutionSpec =
       curConstitution `shouldBe` constitution
 
     impAnn "Pulser is reset" $ do
-      ConwayGovState _ _ _ _ _ pulser <- getsNES newEpochStateGovStateL
+      pulser <- getsNES newEpochStateDRepPulsingStateL
       let pulserRatifyState = extractDRepPulsingState pulser
       rsEnacted pulserRatifyState `shouldBe` Seq.empty
       enactState <- getEnactState
--- a/eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/EnactSpec.hs
+++ b/eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/EnactSpec.hs
@@ -40,7 +40,23 @@ spec ::
   SpecWith (ImpTestState era)
 spec =
   describe "ENACT" $ do
-    it "TreasuryWithdrawal" $ do
+    treasuryWithdrawalsSpec
+    hardForkInitiationSpec
+    noConfidenceSpec
+    constitutionSpec
+
+treasuryWithdrawalsSpec ::
+  forall era.
+  ( ConwayEraImp era
+  , NFData (Event (EraRule "ENACT" era))
+  , ToExpr (Event (EraRule "ENACT" era))
+  , Eq (Event (EraRule "ENACT" era))
+  , Typeable (Event (EraRule "ENACT" era))
+  ) =>
+  SpecWith (ImpTestState era)
+treasuryWithdrawalsSpec =
+  describe "Treasury withdrawals" $ do
+    it "modify EnactState as expected" $ do
       rewardAcount1 <- registerRewardAccount
       govActionId <- submitTreasuryWithdrawals [(rewardAcount1, Coin 666)]
       gas <- getGovActionState govActionId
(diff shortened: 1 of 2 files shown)


teodanciu:
Sorry, the last two commits were already reviewed as part of https://github.com/IntersectMBO/cardano-ledger/pull/4201, but the base branch was this one instead of master.

Lucsanszky (dismissed):

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/EnactSpec.hs
```diff
@@ -97,95 +99,94 @@ treasuryWithdrawalsSpec =
...
+      submitYesVote_ (CommitteeVoter committeeC) gaId
       passNEpochs 2
-      getsNES (nesEsL . esAccountStateL . asTreasuryL)
-        `shouldReturn` Coin curTreasury
-    it "Proposals in multiple epochs" $ do
-      (drepC, committeeC, _gpi) <- electBasicCommittee
+      getTreasury `shouldReturn` initialTreasury
+      sumRewardAccounts withdrawals `shouldReturn` zero
+
+    it "Wthdrawals exceeding treasury submitted in several proposals within the same epoch" $ do
```
Lucsanszky: Minor typo: 
```suggestion
    it "Withdrawals exceeding treasury submitted in several proposals within the same epoch" $ do
```

Lucsanszky:
I'm currently reviewing this and I intend to finish the review later today. I have some questions and minor suggestions.

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/EnactSpec.hs
```diff
@@ -79,145 +100,343 @@ spec =
...
+        `shouldReturn` val1
+      checkProposalsEmpty
 
-      impAnn "Pulser is reset" $ do
-        ConwayGovState _ _ _ _ _ pulser <- getsNES newEpochStateGovStateL
-        let pulserRatifyState = extractDRepPulsingState pulser
-        rsEnacted pulserRatifyState `shouldBe` Seq.empty
-        enactState <- getEnactState
-        rsEnactState pulserRatifyState `shouldBe` enactState
+checkProposalsEmpty :: ConwayEraGov era => ImpTestM era ()
```
Soupstraw: Let's move this to `Test.Cardano.Ledger.Conway.ImpTest`

Soupstraw: Also I think we should case match on `Proposals` and then give a more informative error message, right now it says `Expected True, got False` or something like that

Soupstraw: You can copy from [my PR](https://github.com/IntersectMBO/cardano-ledger/pull/4205/files#diff-832fb7c0b3f3c6b053b227f99393da3cd4ae7bea71729df757f8b8823139021b)

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/EnactSpec.hs
```diff
@@ -79,145 +100,343 @@ spec =
...
-        proposalsAfterVotes `shouldBe` proposalsWithVotes
+                initialTreasury
+                withdrawals
 
-      impAnn "EnactState has not changed" $
-        enactStateAfterVotes `shouldBe` enactStateBeforeVotes
+        getTreasury `shouldReturn` expectedTreasury
+        -- check that the sum of the rewards matches what was spent from the treasury
+        sumRewardAccounts withdrawals `shouldReturn` (initialTreasury <-> expectedTreasury)
+  where
```
Lucsanszky, line 166: I think introducing this would improve the readability of the tests:
```suggestion
  where
    expectNoChange initialT ws = do
      -- 1. Assert that there are proposals.
      expectCurrentProposals
      -- 2. Assert that there are no proposals from the previous epoch.
      expectNoPulserProposals

      passNEpochs 2

      -- Unmet pre-condition for treasury withdrawal enactment
      -- (insufficient funds) - nothing should change:
      -- 1. Assert that treasury balance remains the same.
      getTreasury `shouldReturn` initialT
      -- 2. Assert that reward accounts are empty.
      sumRewardAccounts ws `shouldReturn` zero
      -- 3. Assert that there are still proposals waiting to be enacted.
      expectCurrentProposals
      -- 4. Assert that there are now proposals in the pulser, awaiting enactment
      expectPulserProposals
```

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/EnactSpec.hs
```diff
@@ -79,145 +100,343 @@ spec =
...
+      (drepC, committeeC, _) <- electBasicCommittee
+      modifyPParams $ ppGovActionLifetimeL .~ EpochInterval 5
+      initialTreasury <- getTreasury
+      numWithdrawals <- choose (1, 10)
+      withdrawals <- genWithdrawalsExceeding initialTreasury numWithdrawals 100
+
+      gaId <- submitTreasuryWithdrawals withdrawals
+      submitYesVote_ (DRepVoter drepC) gaId
+      submitYesVote_ (CommitteeVoter committeeC) gaId
       passNEpochs 2
```
Lucsanszky: Step 1
```suggestion
```

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/EnactSpec.hs
```diff
@@ -79,145 +100,343 @@ spec =
...
-      getsNES (nesEsL . curPParamsEpochStateL . ppProtocolVersionL) `shouldReturn` curProtVer
-      submitYesVote_ (DRepVoter (KeyHashObj dRep2)) govActionId
+      getTreasury `shouldReturn` initialTreasury
+      -- reward accounts are empty
+      sumRewardAccounts withdrawals `shouldReturn` zero
```
Lucsanszky: Step 2
```suggestion

      -- Treasury withdrawals pre-condition should fail
      -- so we are not expecting any changes in the state.
      expectNoChange initialTreasury withdrawals

```

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/EnactSpec.hs
```diff
@@ -79,145 +100,343 @@ spec =
...
+      (drepC, committeeC, _) <- electBasicCommittee
+      modifyPParams $ ppGovActionLifetimeL .~ EpochInterval 5
+      initialTreasury <- getTreasury
+      numWithdrawals <- choose (1, 10)
+      withdrawals <- genWithdrawalsExceeding (Coin (fromIntegral (maxBound :: Word64))) numWithdrawals 100
+      gaId <- submitTreasuryWithdrawals withdrawals
+
+      submitYesVote_ (DRepVoter drepC) gaId
+      submitYesVote_ (CommitteeVoter committeeC) gaId
       passNEpochs 2
```
Lucsanszky: Step 1
```suggestion
```

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/EnactSpec.hs
```diff
@@ -79,145 +100,343 @@ spec =
...
-                (CommitteeVoter committeeMember)
-                VoteYes
-                govActionId
-                ( proposalsAddVote
-                    (DRepVoter dRep)
-                    VoteYes
-                    govActionId
-                    proposalsBeforeVotes
+      getTreasury `shouldReturn` initialTreasury
+      sumRewardAccounts withdrawals `shouldReturn` zero
```
Lucsanszky: Step 2
```suggestion
      -- Treasury withdrawals pre-condition should fail
      -- so we are not expecting any changes in the state.
      expectNoChange initialTreasury withdrawals
```

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/EnactSpec.hs
```diff
@@ -79,145 +100,343 @@ spec =
...
       passNEpochs 2
-      getsNES (nesEsL . curPParamsEpochStateL . ppProtocolVersionL) `shouldReturn` curProtVer
-      submitYesVote_ (DRepVoter (KeyHashObj dRep2)) govActionId
+      getTreasury `shouldReturn` initialTreasury
+      -- reward accounts are empty
+      sumRewardAccounts withdrawals `shouldReturn` zero
+
+      let sumRequested = mconcat $ snd <$> withdrawals
+
+      impAnn "Submit a treasury donation that can conver the withdrawals" $ do
```
Lucsanszky: Minor typo:
```suggestion
      impAnn "Submit a treasury donation that can cover the withdrawals" $ do
```

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/EnactSpec.hs
```diff
@@ -79,145 +100,343 @@ spec =
...
+        let tx =
+              mkBasicTx mkBasicTxBody
+                & bodyTxL . treasuryDonationTxBodyL .~ (sumRequested <-> initialTreasury)
+        submitTx_ tx
       passNEpochs 2
-      getsNES (nesEsL . curPParamsEpochStateL . ppProtocolVersionL) `shouldReturn` curProtVer
-      submitYesVote_ (StakePoolVoter stakePoolId2) govActionId
+      getTreasury `shouldReturn` zero
+      sumRewardAccounts withdrawals `shouldReturn` sumRequested
```
Lucsanszky, line 125:
```suggestion
      -- Pre-condition for treasury withdrawal enactment is now met:
      -- 1. Assert that the total amount of treasury withdrawals was
      -- deducted from the treasury balance.
      getTreasury `shouldReturn` zero
      -- 2. Assert that the total amount of treasury withdrawals (sumRequested)
      -- equals the total balance of the reward accounts.
      sumRewardAccounts withdrawals `shouldReturn` sumRequested
      -- The proposal was enacted so:
      -- 3. Assert that there are no current proposals.
      expectNoCurrentProposals
      -- 4. Assert that the pulser was reset.
      expectNoPulserProposals

```

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/EnactSpec.hs
Lucsanszky: There is also `expectNoCurrentProposals` in `Test.Cardano.Ledger.Conway.ImpTest`.

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/EnactSpec.hs
```diff
@@ -79,145 +100,343 @@ spec =
...
+      checkProposalsEmpty
 
-      impAnn "Pulser is reset" $ do
-        ConwayGovState _ _ _ _ _ pulser <- getsNES newEpochStateGovStateL
-        let pulserRatifyState = extractDRepPulsingState pulser
-        rsEnacted pulserRatifyState `shouldBe` Seq.empty
-        enactState <- getEnactState
-        rsEnactState pulserRatifyState `shouldBe` enactState
+checkProposalsEmpty :: ConwayEraGov era => ImpTestM era ()
+checkProposalsEmpty = null . proposalsIds <$> getProposals `shouldReturn` True
```
Lucsanszky:
```suggestion
```

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/EnactSpec.hs
```diff
@@ -79,145 +100,343 @@ spec =
...
+      traverse_
+        ( \gaid -> do
+            submitYesVote_ (DRepVoter drepC) gaid
+            submitYesVote_ (CommitteeVoter committeeC) gaid
+        )
+        gaids
+      passNEpochs 2
+      getsNES (nesEsL . curPParamsEpochStateL . ppDRepDepositL)
+        `shouldReturn` val1
+      checkProposalsEmpty
```
Lucsanszky:
```suggestion
      expectNoCurrentProposals
```

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/EnactSpec.hs
```diff
@@ -79,145 +100,343 @@ spec =
...
+      traverse_ @[]
+        ( \gaid -> do
+            submitYesVote_ (DRepVoter drepC) gaid
+            submitYesVote_ (CommitteeVoter committeeC) gaid
+        )
+        [pGai0, pGai1, pGai2]
+      passNEpochs 2
+      getLastEnactedParameterChange
+        `shouldReturn` SJust (GovPurposeId pGai2)
+      checkProposalsEmpty
```
Lucsanszky:
```suggestion
      expectNoCurrentProposals
```

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/EnactSpec.hs
```diff
@@ -79,145 +100,343 @@ spec =
...
+        traverse_ @[]
+          ( \gaid -> do
+              submitYesVote_ (DRepVoter drepC) gaid
+              submitYesVote_ (StakePoolVoter poolKH) gaid
+          )
+          [gai1, gai2, gai3]
+        passNEpochs 2
+        getLastEnactedCommittee
+          `shouldReturn` SJust (GovPurposeId gai2)
+        checkProposalsEmpty
```
Lucsanszky:
```suggestion
        expectNoCurrentProposals
```

Lucsanszky (commented):
Sorry, it took me long enough! :grimacing: Overall, I quite like it but since I'm new to the codebase, I felt that some more comments and assertions could improve the readability of the tests. However, my additions might be a bit too verbose so feel free to change or ignore them. Especially, if I didn't get the jargon right. :) 

I mainly looked at the `Treasury withdrawals` spec, here is the gist of my suggestions:
I felt that it might be useful to also assert for the presence/absence of both the current proposals and previous ones (so the ones that are in the `pulser`), so I added those assertions as suggestions and also added an `expectNoChange` assertion that checks multiple things when a `Treasury Withdrawal` governance action's pre-condition is not met and hence the proposal will not be enacted as a result. Hopefully, I did not misunderstand the state changes and my suggestions are indeed helpful rather than misleading. 

It seems like GitHub doesn't support suggestions over deleted lines so sometimes I had to separate suggestions into `Step 1` and `Step 2`.

I was thinking about reorganising and rewording the `it`s and `describe`s as I would have preferred something along the lines of this:
```
        describe "Treasury withdrawals"
          it "modify EnactState as expected"
          describe "in a single proposal"
            it "should only be enacted when treasury has sufficient funds"
            it "should not exceed maxBound Word64"
          describe "in several proposals within the same epoch"
            it "should only be enacted until treasury has sufficient funds"
```
However, it seems like we don't really use this style throughout the codebase, so I opted to leave things intact and I also try to avoid too many nested `describe`s.

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/EnactSpec.hs
teodanciu: Sorry, i didn't understand from the other comment, what is Step 1?  What should be added here?

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/EnactSpec.hs
teodanciu: Thanks!

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/EnactSpec.hs
teodanciu: Nice, thanks both, I replaced with `expectNoCurrentProposals`

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/EnactSpec.hs
teodanciu: I'm not sure about the relevance of asserting proposals, whether current or in Pulser, at this point :thinking:

teodanciu:
@Lucsanszky  Sorry I'm a bit lost  about your comments with Step 1 and Step 2. 
Do you mean, we should extract the check that the withdrawal proposal had no effect? So basically: 
```
      getTreasury `shouldReturn` initialTreasury
      sumRewardAccounts withdrawals `shouldReturn` zero
```  ?

File: eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/EnactSpec.hs (hunk 4 above)
Comment (Lucsanszky) [reply eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/EnactSpec.hs]: No worries! It's a bit unfortunate that GitHub doesn't allow including deleted lines. So Step 1 is to remove `passNEpochs 2` as that is already included in the `expectNoChanges` utility function that I suggested.

File: eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/EnactSpec.hs (hunk 3 above)
Comment (Lucsanszky) [reply eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/EnactSpec.hs:166]: I would not be surprised if I'm wrong. I thought that `current proposals, no proposals in pulsar -> pass 2 epochs -> current proposals, proposals in pulsar` would imply that the proposal was not enacted. 

If that's not the case and you feel that it's irrelevant, then just ignore my `expectNoChanges` suggestion (and related stuff). :)

**Comment by Lucsanszky** [comment]
> @Lucsanszky Sorry I'm a bit lost about your comments with Step 1 and Step 2. Do you mean, we should extract the check that the withdrawal proposal had no effect? So basically:
> 
> ```
>       getTreasury `shouldReturn` initialTreasury
>       sumRewardAccounts withdrawals `shouldReturn` zero
> ```  ? 
> ```

Yes, exactly. I added it because I also added the proposal assertions and at that point I thought that it might worth extracting the whole thing.

File: eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/EnactSpec.hs [hunk 14]
```diff
@@ -79,145 +100,343 @@ spec =
...
 
-      passEpoch
+actionPrioritySpec ::
+  forall era.
+  ConwayEraImp era =>
+  SpecWith (ImpTestState era)
+actionPrioritySpec =
+  describe "Competing proposals ratified in the same epoch" $ do
+    it
+      "The first submitted proposal of higher priority wins, even when submitted after a lower priority one"
```
aniketd:
```suggestion
      "higher action priority wins"
```

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/EnactSpec.hs
```diff
@@ -79,145 +100,343 @@ spec =
...
+        committee `shouldBe` SNothing
+
+    let val1 = Coin 1_000_001
+    let val2 = Coin 1_000_002
+    let val3 = Coin 1_000_003
 
-      impAnn "Constitution is enacted after two epochs" $ do
-        curConstitution <- getsNES $ newEpochStateGovStateL . constitutionGovStateL
-        curConstitution `shouldBe` constitution
+    it "The first submitted proposal of the same priority gets enacted first" $ do
```
aniketd:
```suggestion
    it "proposals of same priority are enacted in order of submission" $ do
```

aniketd (commented):
Just a few suggestions for better test-naming than I originally thought.

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/EnactSpec.hs
```diff
@@ -79,145 +100,343 @@ spec =
...
+      -- generate withdrawals which individually do not exceed the treasury, but their sum does
+      numWithdrawals <- choose (1, 10)
+      withdrawals <- genWithdrawalsExceeding initialTreasury numWithdrawals 50
+
+      impAnn "submit in the same proposal, with no effect on the treasury" $ do
+        gaId <- submitTreasuryWithdrawals withdrawals
+        submitYesVote_ (DRepVoter drepC) gaId
+        submitYesVote_ (CommitteeVoter committeeC) gaId
+        passNEpochs 2
+        getTreasury `shouldReturn` initialTreasury
```
lehins: We already have this check in the test case above: `Withdrawals exceeding maxBound Word64 submitted in a single proposal`, so I don't think we need to do it here again. Unless of course I am missing something.


```suggestion
```

Alternatively we can create a reusable check, similar to @Lucsanszky's suggestions

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/EnactSpec.hs
```diff
@@ -79,145 +100,343 @@ spec =
...
+  where
+    getTreasury = getsNES (nesEsL . esAccountStateL . asTreasuryL)
+    sumRewardAccounts withdrawals = mconcat <$> traverse (getRewardAccountAmount . fst) withdrawals
+    genWithdrawalsExceeding (Coin val) n maxExcess = do
+      accounts <- replicateM n $ registerRewardAccount @era
+      pcts <- replicateM n (choose (1, 100) :: ImpTestM era Integer)
+      let tot = sum pcts
+      excess <- choose (1, maxExcess) :: ImpTestM era Integer
+      let amounts = fmap (\x -> Coin $ ceiling ((x + excess) % tot * fromIntegral val)) pcts
+      pure $ zip accounts amounts
```
lehins: `maxExcess` is overused in here and its meaning is totally unclear. Why would every random `pct` (which I assume stands for percent, but isn't really a percent) include that excess? So, IMHO we don't need to exceed every account by a random 50 to 100 points, or whatever they may mean. It is enough that we have at least one account that exceeds by a tiny bit, it will be enough. Excess can be decided at random too:
```suggestion
    genWithdrawalsExceeding (Coin val) n = do
      pcts <- replicateM (n - 1) $ choose (1, 100)
      let tot = sum pcts
      Positive excess <- arbitrary
      let amounts = excess : map (\x -> Coin $ ceiling ((x * val) % tot)) pcts
      forM amounts $ \ amount -> (,amount) <$> registerRewardAccount
```

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/EnactSpec.hs
```diff
@@ -79,145 +100,343 @@ spec =
...
+              submitYesVote_ (DRepVoter drepC) gaId
+              submitYesVote_ (CommitteeVoter committeeC) gaId
+          )
+          withdrawals
+        passNEpochs 2
+
+        let expectedTreasury =
+              foldl'
+                ( \acc (_, x) ->
+                    if acc <-> x >= zero
```
lehins:
```suggestion
                    if acc >= x
```

lehins (commented):
Looks good overall. Just needs to be finished up and comments need to be addressed

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/EnactSpec.hs
teodanciu: I made the "excess" configurable for the test case above
``` generate withdrawals which individually do not exceed the treasury, but their sum does ```
but I agree i didn't make it clear what it means or how to use it.  I thought it's handy to control the generation in this way, but maybe it's a little excessive.

File: eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/EnactSpec.hs (hunk 16 above)
Comment (teodanciu) [reply eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/EnactSpec.hs]: The difference is how the data is generated  - so in this one, the individual proposals wouldn't exceed the treasury, but their sum does.

File: eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/EnactSpec.hs (hunk 17 above)
Comment (lehins) [reply eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/EnactSpec.hs]: I understand the goal, I am saying that it was unnecessary to achieve the goal.
Instead of incrementing every amount by some arbitrary percentage, it is better to add one amount that puts the whole sum over the top.

Comment (teodanciu) [reply eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/EnactSpec.hs]: Ok :+1:

File: eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/EnactSpec.hs [hunk 19]
```diff
@@ -79,145 +100,343 @@ spec =
...
-      passEpoch
+actionPrioritySpec ::
+  forall era.
+  ConwayEraImp era =>
+  SpecWith (ImpTestState era)
+actionPrioritySpec =
+  describe "Competing proposals ratified in the same epoch" $ do
+    it
+      "The first submitted proposal of higher priority wins, even when submitted after a lower priority one"
+      $ do
```
lehins: Test names should be succinct and unique, so they can be easily found in the codebase. They don't need to describe the full behavior of the test, that can be explained with a comment above the test case.
```suggestion
    -- The first submitted proposal of higher priority wins, even when 
    -- submitted after a lower priority one
    it "Proposal of higher priority wins" $ do
```

lehins: Just read the resolved suggestion, seems like @aniketd suggested the same thing :slightly_smiling_face: 

Notice that allows putting both `it` and `do` on the same line, making the code more readable too.

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/EnactSpec.hs
lehins: Oh yeah. Good point. I knew I was missing something. 
Should this check be extracted into a let binding, so that both test cases can use it?
Feel free to leave it as it is.

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/EnactSpec.hs
teodanciu: I abstracted the generation and checks in the functions: 
`enactTreasuryWithdrawals`  and `checkNoWithdrawal`

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/EnactSpec.hs
teodanciu: I abstracted it in the function `checkNoWithdrawal`

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/EnactSpec.hs
```diff
@@ -79,145 +102,324 @@ spec =
...
+      (drepC, committeeC, _) <- electBasicCommittee
+      modifyPParams $ ppGovActionLifetimeL .~ EpochInterval 5
+      initialTreasury <- getTreasury
+      numWithdrawals <- choose (1, 10)
+      withdrawals <- genWithdrawalsExceeding initialTreasury numWithdrawals
+
+      void $ enactTreasuryWithdrawals withdrawals drepC committeeC
+      checkNoWithdrawal initialTreasury withdrawals
+
+      let sumRequested = mconcat $ snd <$> withdrawals
```
lehins:
```suggestion
      let sumRequested = foldMap snd withdrawals
```

lehins (approved):
Looks great. Thank you!
I made a minor suggestion, but feel free to ignore it

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/EnactSpec.hs
teodanciu: Nice, thank you! Fixed

PR #4194, opened 2024-03-11
Fix epoch boundary check in ImpTest

teodanciu:
# Description

In `ImpTest.passEpoch`, we are checking that the sum of  the AdaPots are preserved before and after the epoch change.
Starting with Conway, we also have donations which get added to the treasury at the epoch boundary, so we have to take them into account if we want to check to hold after donations are being made.

# Checklist

- [ ] Self-reviewed the diff


Lucsanszky (approved):
I haven't checked the related PRs (#4179, #4189) but this LGTM. :+1:

PR #4195, opened 2024-03-12
Fix Haddocks CI

### LINKED ISSUE #4156: Haddock building is broken
It seems that there is an issue with [github pages on CI](https://github.com/IntersectMBO/cardano-ledger/actions/runs/8143972981/job/22257035670), in particular, there is an issue building haddock

### LINKED ISSUE #4193: Benchmark results in GitHub Pages are overwritten by Haddock build
See discussion in [#4156](https://github.com/IntersectMBO/cardano-ledger/issues/4156#issuecomment-1984560085)

neilmayhew:
# Description

Fix the broken Haddocks CI

Resolves #4156
Resolves #4193

* Avoid adding all of the repo files to the gh-pages branch
* Ensure `scripts/haddocks.sh` doesn't fail when optional files don't exist
* Preserve benchmark results in gh-pages when adding Haddocks
* Don't build Haddocks using `nix develop`, to avoid running out of disk space

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/.github/workflows/gh-pages.yml
+++ b/.github/workflows/gh-pages.yml
@@ -50,8 +50,14 @@ jobs:
             git branch -D "$REFNAME"
           done
         git checkout -b gh-pages
+
         git rm -rfq .
-        git commit -qm "Remove all files"
+        git commit -qm "Remove all existing files"
+
+        echo "cardano-ledger.cardano.intersectmbo.org" >CNAME
+        touch .nojekyll
+        git add CNAME .nojekyll
+        git commit -qm "Add CNAME and .nojekyll"
 
         # Preserve benchmark results, if any
         git ls-remote origin --heads gh-pages |
--- a/.github/workflows/gh-pages.yml
+++ b/.github/workflows/gh-pages.yml
@@ -1,4 +1,4 @@
-name: Github Pages
+name: Haddocks to GitHub Pages
 
 on:
   push:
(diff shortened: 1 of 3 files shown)


Lucsanszky (approved):
Thanks for the small, atomic commits! They helped with reviewing.

LGTM :+1:

neilmayhew:
Thanks for reviewing, @Lucsanszky!

Unfortunately, the [gh-pages pipeline](https://github.com/IntersectMBO/cardano-ledger/actions/runs/8258118983) isn't producing quite the right thing, so it's not ready to be merged yet. However, hopefully, it needs only a small amount of tweaking, and having you look over it already is a big help.

.github/workflows/gh-pages.yml
```diff
@@ -23,17 +23,17 @@ jobs:
...
-              git config --local user.email "${{ github.actor }}@users.noreply.github.com"
-              git add -A --force ./haddocks
-              git mv ./haddocks/* .
-              git commit -m "Updated"
-              rm -rf haddocks
+          git config --local user.name ${{ github.actor }}
+          git config --local user.email "${{ github.actor }}@users.noreply.github.com"
+          git add -A --force ./haddocks
+          git mv ./haddocks/* .
+          git commit -m "Updated from ${GITHUB_SHA} via ${GITHUB_EVENT_NAME}"
```
teodanciu: Much better than "Updated' :sweat_smile:

neilmayhew: Thanks! :grin:

.github/workflows/gh-pages.yml
```diff
@@ -25,6 +25,15 @@ jobs:
         run: |
           git config --local user.name ${{ github.actor }}
           git config --local user.email "${{ github.actor }}@users.noreply.github.com"
+          # Preserve benchmark results, if any
```
teodanciu: How is this related to benchmarks? Is it because its' the only ones that could be generated by cabal but not checked in?

neilmayhew: The [benchmarks CI](https://github.com/IntersectMBO/cardano-ledger/actions/workflows/bench.yml) is triggered at the same time as the Haddock one, but finishes first. It saves its [results](https://cardano-ledger.cardano.intersectmbo.org/dev/bench/) into a directory called `dev/bench` and commits that to the `gh-pages` branch without overwriting it. However, when the Haddocks CI finishes it force-pushes the `gh-pages` branch and loses the benchmark results. However, we don't want to keep the commits from the previous Haddocks run because then we'd be filling up our git repo with useless history. This tricky little bit of code gets the `dev` directory from the previous state of the `gh-pages` branch without bringing in its other history.

teodanciu:
Great improvement! 
I started an action on my own fork as well, just to play with it in case we need to change something in Settings, so we can see tomorrow if it's still not working, hopefully we'll figure it out!

neilmayhew:
Sorry for the noise. I'm trying to get this to work properly so I'm experimenting. I hate that it's almost impossible to debug CI workflows locally!

However, I've just realized that I can do this in a new branch that isn't part of a PR, and I can update this PR branch once I've got it working.

I have to trigger the gh-pages workflow manually anyway, so the regular CI isn't useful to me (which is why I've been cancelling it each time).

neilmayhew:
The weird thing I'm struggling with is that the site won't deploy properly (becomes a 404) unless I include the non-Haddock files from the repo. I'm now trying to narrow down which file(s) it actually needs.

neilmayhew:
> Great improvement!

Thanks!

> I started an action on my own fork as well, just to play with it in case we need to change something in Settings, so we can see tomorrow if it's still not working, hopefully we'll figure it out!

That could be very useful.

teodanciu:
> the site won't deploy properly (becomes a 404) unless I include the non-Haddock files from the repo. 

Interesting.  Sorry it's such a headache! 
The other thing that I meant that would bypass the whole gh-pages branch is this: https://github.com/actions/deploy-pages
But no doubt it would bring other problems, so I'm not necessarily suggesting we change  to using it - whichever works!

neilmayhew:
Found it! The file `CNAME` is needed to make the site work with the custom domain name.

It's a bit obscure that this file is checked into the repo along with the code and isn't part of the gh-pages workflow. I'll remove it from the repo and create it on the fly during the workflow.

teodanciu:
> It's a bit obscure that this file is checked into the repo

I admit I missed the part when we added this! Good job finding where the dog is buried. TIL

neilmayhew:
This is working now, but takes a long time to build the Haddocks. It's now 95m when it was 55m before. I think this is because it's no longer getting the benefit of nix caching.

However, I'd like to merge this as-is and add the caching in a separate PR.

teodanciu (approved):
Looks good to me

PR #4197, opened 2024-03-13
add unsafeMkProposals to be used for testing

MaximilianAlgehed:
# Description

When profiling `prop_GOV` that I'm working on over on the branch `PR-fix-prop_GOV` I found that after fixing a bunch of performance bottle necks in `constrained-generators` (now running many X faster!) `checkInvariantAfterAddition` was taking 24% of the time. This seemed excessive as the whole point of the constraint-based generator is to generate correct stuff by default.

This PR adds `unsafeMkProposals` to make it possible to side-step `checkInvariantAfterAddition` when generating test data.

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Proposals.hs
+++ b/eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Proposals.hs
@@ -102,6 +102,7 @@ module Cardano.Ledger.Conway.Governance.Proposals (
   pRootsL,
   pGraphL,
   mkProposals,
+  unsafeMkProposals,
   PRoot (..),
   prRootL,
   prChildrenL,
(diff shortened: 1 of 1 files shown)


eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Proposals.hs
```diff
@@ -238,33 +239,51 @@ proposalsAddAction ::
   GovActionState era ->
   Proposals era ->
   Maybe (Proposals era)
-proposalsAddAction gas ps = withGovActionParent gas (Just psWithGas) update
+proposalsAddAction gas ps = checkInvariantAfterAddition gas ps <$> runProposalsAddAction gas ps
```
teodanciu, line 242: I like this refactoring, it's easier to follow with the concerns neatly separated like this.

eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Proposals.hs
```diff
@@ -238,33 +239,51 @@ proposalsAddAction ::
...
+-- | Add a single @`GovActionState`@ to the @`Proposals`@ forest.
+-- The tree to which it is added is picked according to its
+-- @`GovActionPurpose`@.
+unsafeProposalsAddAction ::
+  forall era.
+  HasCallStack =>
+  GovActionState era ->
+  Proposals era ->
+  Proposals era
+unsafeProposalsAddAction gas ps =
```
teodanciu: It would be nice to have such functions in tests instead, but I guess to do that we'd have to expose `runProposalsAddAction`, which would be confusing since it has the same signature as  `proposalsAddAction` . 
Should we add at least some comment saying that it's for testing only , or the `unsafe` prefix is expressive enough ?

teodanciu: Oh i'm sorry, I see now that there is a comment for `unsafeMkProposals`,  the function invoking this one ( I should really start using "Start a review" instead of "Add single comment".. oops!) . 
Maybe we could define this function nested in `unsafeMkProposals` , to minimize the scope of this function, to pollute less and discourage its usages.

MaximilianAlgehed: Yes, we can move this function to `unsafeMkProposals` instead. I would hesitate to expose enough to re-implement this functionality elsewhere, that makes changing things in the future more annoying. But I'm not firm either way.

teodanciu (approved):
Looks good to me

PR #4200, opened 2024-03-13
Fix `prop_GOV` so that it runs again

MaximilianAlgehed:
# Description

This is kindof a big PR that uses the new `Tree` extension to `constrained-generators` to make sure `prop_GOV` ticks.

To make this work I had to make a bunch of performance improvements to `constrained-generators` that will hopefully be reflected in CI times going forward.

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/V2/Conway.hs
+++ b/libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/V2/Conway.hs
@@ -84,24 +84,29 @@ import Crypto.Hash (Blake2b_224)
 import Data.ByteString qualified as BS
 import Data.ByteString.Short (ShortByteString)
 import Data.ByteString.Short qualified as SBS
-import Data.Default.Class
+import Data.Coerce
 import Data.Foldable
 import Data.Kind
 import Data.Map (Map)
+import Data.Map.Strict qualified as Map
 import Data.Maybe
+import Data.OMap.Strict qualified as OMap
 import Data.OSet.Strict qualified as SOS
 import Data.Sequence (Seq)
 import Data.Sequence qualified as Seq
 import Data.Sequence.Strict (StrictSeq)
 import Data.Sequence.Strict qualified as StrictSeq
 import Data.Set (Set)
+import Data.Set qualified as Set
 import Data.Text (Text)
+import Data.Tree
 import Data.Typeable
 import Data.VMap (VMap)
 import Data.VMap qualified as VMap
 import Data.Word
 import Numeric.Natural (Natural)
 import PlutusLedgerApi.V1 qualified as PV1
+import System.Random
 import Test.Cardano.Ledger.Allegra.Arbitrary ()
 import Test.Cardano.Ledger.Alonzo.Arbitrary ()
 import Test.Cardano.Ledger.Core.Utils
@@ -111,14 +116,15 @@ import Test.QuickCheck hiding (Args, Fun, forAll)
 import Constrained hiding (Value)
 import Constrained qualified as C
 
-type ConwayUnivFns = StringFn : BaseFns
+type ConwayUnivFns = StringFn : RoseTreeFn : BaseFns
 type ConwayFn = Fix (OneofL ConwayUnivFns)
 
 type IsConwayUniv fn =
-  ( Member (StringFn fn) fn
+  ( BaseUniverse fn
+  , Member (StringFn fn) fn
   , Member (MapFn fn) fn
   , Member (FunFn fn) fn
-  , BaseUniverse fn
+  , Member (RoseTreeFn fn) fn
   )
 
 -- TxBody HasSpec instance ------------------------------------------------
@@ -258,6 +264,12 @@ instance (IsConwayUniv fn, HasSpec fn a) => HasSpec fn (Seq a)
 
 instance HasSimpleRep (StrictMaybe a)
 instance HasSpec fn a => HasSpec fn (StrictMaybe a)
+
+cSNothing_ :: HasSpec fn a => Term fn (StrictMaybe a)
+cSNothing_ = con @"SNothing" (lit ())
+
+cSJust_ :: HasSpec fn a => Term fn a -> Term fn (StrictMaybe a)
+cSJust_ = con @"SJust"
 
 instance HasSimpleRep (Sized a)
 instance (IsConwayUniv fn, HasSpec fn a) => HasSpec fn (Sized a)
@@ -683,18 +695,43 @@ instance (IsConwayUniv fn, Crypto c) => HasSpec fn (GovActionId c)
 instance HasSimpleRep GovActionIx
 instance IsConwayUniv fn => HasSpec fn GovActionIx
 
--- TODO: this needs a `canFollow` predicate (that can't be negated).
 instance HasSimpleRep ProtVer
 instance IsConwayUniv fn => HasSpec fn ProtVer
 
--- TODO: this is a bit iffy, Version is really a bounded type
--- that we should be able to represent. Same thing with `UnitInterval`
+-- We do this like this to get the right bounds for `VersionRep`
+-- while ensuring that we don't have to add instances for e.g. `Num`
+-- to version.
+newtype VersionRep = VersionRep Word64
+  deriving (Show, Eq, Ord, Num, Random) via Word64
+instance BaseUniverse fn => HasSpec fn VersionRep where
+  type TypeSpec fn VersionRep = NumSpec VersionRep
+  emptySpec = emptyNumSpec
+  combineSpec = combineNumSpec
+  genFromTypeSpec = genFromNumSpec
+  conformsTo = conformsToNumSpec
+  toPreds = toPredsNumSpec
+instance Bounded VersionRep where
+  minBound = VersionRep $ getVersion64 minBound
+  maxBound = VersionRep $ getVersion64 maxBound
+instance MaybeBounded VersionRep
+
 instance HasSimpleRep Version where
-  type SimpleRep Version = Word64
-  fromSimpleRep = fromJust . mkVersion64
-  toSimpleRep = getVersion64
-instance IsConwayUniv fn => HasSpec fn Version where
-  emptySpec = NumSpecInterval (Just $ getVersion64 minBound) (Just $ getVersion64 maxBound)
+  type SimpleRep Version = VersionRep
+  fromSimpleRep (VersionRep rep) = case mkVersion64 rep of
+    Left err ->
+      error $
+        unlines
+          [ "fromSimpleRep @Version:"
+          , show rep
+          , err
+          ]
+    Right a -> a
+  toSimpleRep = VersionRep . getVersion64
+instance BaseUniverse fn => HasSpec fn Version
+instance BaseUniverse fn => OrdLike fn Version
+
+succV_ :: BaseUniverse fn => Term fn Version -> Term fn Version
+succV_ = fromGeneric_ . (+ 1) . toGeneric_
 
 instance HasSimpleRep (GovPurposeId p era)
 instance (Typeable p, IsConwayUniv fn, Era era) => HasSpec fn (GovPurposeId p era)
@@ -843,6 +880,12 @@ instance
   ) =>
   HasSpec fn (ProposalProcedure era)
 
+pProcGovAction_ ::
+  IsConwayUniv fn =>
+  Term fn (ProposalProcedure (ConwayEra StandardCrypto)) ->
+  Term fn (GovAction (ConwayEra StandardCrypto))
+pProcGovAction_ = sel @2
+
 instance HasSimpleRep ValidityInterval
 instance IsConwayUniv fn => HasSpec fn ValidityInterval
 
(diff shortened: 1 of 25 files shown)


libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/V2/Conway.hs
```diff
@@ -259,6 +265,12 @@ instance (IsConwayUniv fn, HasSpec fn a) => HasSpec fn (Seq a)
 instance HasSimpleRep (StrictMaybe a)
 instance HasSpec fn a => HasSpec fn (StrictMaybe a)
 
+cSNothing_ :: HasSpec fn a => Term fn (StrictMaybe a)
+cSNothing_ = con @"SNothing" (lit ())
+
+cSJust_ :: HasSpec fn a => Term fn a -> Term fn (StrictMaybe a)
+cSJust_ = con @"SJust"
+
```
TimSheard, line 273: Is "con" a new constructor of Term. I haven't run across it before.

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/V2/Conway.hs
```diff
@@ -683,18 +695,43 @@ instance (IsConwayUniv fn, Crypto c) => HasSpec fn (GovActionId c)
...
 instance HasSimpleRep ProtVer
 instance IsConwayUniv fn => HasSpec fn ProtVer
 
--- TODO: this is a bit iffy, Version is really a bounded type
--- that we should be able to represent. Same thing with `UnitInterval`
+-- We do this like this to get the right bounds for `VersionRep`
+-- while ensuring that we don't have to add instances for e.g. `Num`
+-- to version.
+newtype VersionRep = VersionRep Word64
+  deriving (Show, Eq, Ord, Num, Random) via Word64
```
TimSheard: Why word64?

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/V2/Conway/GOV.hs
TimSheard, line 1: So this is the first time I have looked at what it take to make a spec for complicated domain specific data structure.
I am sure there are some rules that are used to do this. I wonder if it might help to write a tutorial for doing the same for something a bit simpler. In the tutorial identify (and name) the steps. An then in a big file like this one could refer to the steps.  I found this pretty fascinating, but I am sure I did not catch all the subtleties .

libs/constrained-generators/src/Constrained/Base.hs
```diff
@@ -396,7 +399,7 @@ toCtx ::
   Var v ->
   Term fn a ->
   m (Ctx fn v a)
-toCtx v = go . simplifyTerm
+toCtx v = go -- . simplifyTerm
```
TimSheard, line 402: Why comment out the simplifyTerm.. Is this a performance issue?
I remember looking at this function when I was working on my PR, and wondering why one called simplifyTerm, and then I thought that it might be necessary, since the go function depends upon some format of the term (Maybe that every application has some special format). Maybe there is no invariant that simplifyTerm creates, But I found this a bit odd.

libs/constrained-generators/src/Constrained/Base.hs
```diff
@@ -548,8 +551,8 @@ instance HasSpec fn a => Semigroup (Spec fn a) where
         (filter (`notElem` cant) as)
   TypeSpec s cant <> MemberSpec as = MemberSpec as <> TypeSpec s cant
   SuspendedSpec v p <> SuspendedSpec v' p' = SuspendedSpec v (p <> rename v' v p')
-  SuspendedSpec v ps <> s = SuspendedSpec v (optimisePred $ ps <> satisfies (V v) s)
-  s <> SuspendedSpec v ps = SuspendedSpec v (optimisePred $ ps <> satisfies (V v) s)
+  SuspendedSpec v ps <> s = SuspendedSpec v (ps <> satisfies (V v) s)
+  s <> SuspendedSpec v ps = SuspendedSpec v (ps <> satisfies (V v) s)
```
TimSheard, line 555: Another performance issue?

libs/constrained-generators/src/Constrained/Base.hs
```diff
@@ -803,6 +806,10 @@ genFromSpec_ spec = do
   res <- strictGen $ genFromSpec spec
   errorGE $ fmap pure res
 
+-- | A version of `genFromSpec` that takes a seed and a size and gives you a result
+genFromSpecWithSeed :: forall fn a. (HasCallStack, HasSpec fn a) => Int -> Int -> Spec fn a -> a
```
TimSheard, line 810: Motivation for this?

libs/constrained-generators/src/Constrained/Base.hs
```diff
@@ -851,41 +858,40 @@ flattenPred pIn = go (freeVarNames pIn) [pIn]
...
-        <> foldMapList (computeBinderDependencies hints) bs
-    TruePred -> mempty
-    FalsePred {} -> mempty
-    Block ps -> foldMap (computeDependencies hints) ps
-    Exists _ b -> computeBinderDependencies hints b
-    Let t b -> noDependencies t <> computeBinderDependencies hints b
-    GenHint _ t -> noDependencies t
+computeDependencies :: Pred fn -> DependGraph fn
+computeDependencies = \case
+  Subst x t p -> computeDependencies (substitutePred x t p)
```
TimSheard, line 863: I see that the `respecting hints `has moved to the function prepareLinearization
Why?

libs/constrained-generators/src/Constrained/Base.hs
```diff
@@ -1104,22 +1124,18 @@ solvableFrom x s g =
 ------------------------------------------------------------------------
 
 freeVarNames :: forall fn t. HasVariables fn t => t -> Set Int
-freeVarNames = Set.map (\(Name v) -> nameOf v) . freeVarSet
+freeVarNames = Set.mapMonotonic (\(Name v) -> nameOf v) . freeVarSet
```
TimSheard, line 1127: Another performance issue?

libs/constrained-generators/src/Constrained/Base.hs
```diff
@@ -1104,22 +1124,18 @@ solvableFrom x s g =
...
-  Name v == Name v'
-    | Just v'' <- cast v = v' == v''
-    | otherwise = False
+  Name (Var i) == Name (Var j) = i == j
 
 instance Ord (Name fn) where
-  compare (Name v) (Name v')
-    | Just v'' <- cast v = compare v'' v'
-    | otherwise = compare (typeRep v) (typeRep v')
+  compare (Name (Var i)) (Name (Var j)) = compare i j
```
TimSheard, line 1138: Seems strange we no longer need to use the typeRep. Is it because compare has type
Name a -> Name a -> Comparison
And since it is never called with with Name of two different types, it never happens?

libs/constrained-generators/src/Constrained/Base.hs
```diff
@@ -1140,32 +1156,64 @@ instance Monoid (FreeVars fn) where
...
   freeVars :: a -> FreeVars fn
-
-freeVarSet :: HasVariables fn a => a -> Set (Name fn)
-freeVarSet = Map.keysSet . unFreeVars . freeVars
+  freeVarSet :: a -> Set (Name fn)
+  freeVarSet = Map.keysSet . unFreeVars . freeVars
+  countOf :: Name fn -> a -> Int
+  countOf n = count n . freeVars
+  appearsIn :: Name fn -> a -> Bool
+  appearsIn n = (> 0) . count n . freeVars
```
TimSheard, line 1175: Motivation for these new methods of HasVariables?
I do notice that they both have default definitions, so there need not be any overhead at instance points.
But I see below that is not the case.
It occurred to me this might be a performance improvement?

libs/constrained-generators/src/Constrained/Base.hs
```diff
@@ -1217,11 +1327,24 @@ backwardsSubstitution sub t =
...
+      | fastInequality t t' = findMatch sub t
       | Just (x', t'') <- cast (x, t')
       , t == t'' =
           Just x'
       | otherwise = findMatch sub t
 
+-- | Sound but not complete inequality on terms
+fastInequality :: Term fn a -> Term fn b -> Bool
+fastInequality (V (Var i)) (V (Var j)) = i /= j
+fastInequality Lit {} Lit {} = False
```
TimSheard, line 1339: So Lit already requires a Show instance. Why not add an Eq instance as well.

libs/constrained-generators/src/Constrained/Base.hs
```diff
@@ -2396,24 +2456,32 @@ genFromFold must size elemS fn foldS = do
...
 
-instance BaseUniverse fn => HasSpec fn Bool where
-  type TypeSpec fn Bool = ()
+instance BaseUniverse fn => HasSpec fn () where
+  type TypeSpec fn () = ()
   emptySpec = ()
   combineSpec _ _ = typeSpec ()
   _ `conformsTo` _ = True
-  genFromTypeSpec _ = pureGen arbitrary
+  genFromTypeSpec _ = pure ()
```
TimSheard, line 2466: Duh, So obvious, why did I not see this

libs/constrained-generators/src/Constrained/Base.hs
```diff
@@ -2396,24 +2456,32 @@ genFromFold must size elemS fn foldS = do
...
   toPreds _ _ = TruePred
-  cardinalTypeSpec _ = MemberSpec [2]
-  cardinalTrueSpec = exactSizeSpec 2 -- there are exactly two, True and False
-  typeSpecOpt _ bad = MemberSpec (filter (`notElem` bad) [True, False])
+  cardinalTypeSpec _ = MemberSpec [1]
+  cardinalTrueSpec = exactSizeSpec 1 -- there are exactly two, True and False
+  typeSpecOpt _ [] = TrueSpec
+  typeSpecOpt _ (_ : _) = MemberSpec []
```
TimSheard, line 2471: I don't understand this. `typeSpecOpt x y` is supposed to be semantically the same as `TypeSpec x y `
Something is wrong here

libs/constrained-generators/src/Constrained/Base.hs
```diff
@@ -2642,50 +2701,61 @@ instance BaseUniverse fn => Functions (SetFn fn) fn where
 
 -- List -------------------------------------------------------------------
 
-data ListSpec fn a = ListSpec [a] (Spec fn Integer) (Spec fn a) (FoldSpec fn a)
+data ListSpec fn a = ListSpec (Maybe Integer) [a] (Spec fn Integer) (Spec fn a) (FoldSpec fn a)
```
TimSheard, line 2704: (Maybe Integer)  An upper bound on the size (if (Just n)), why an upper bound?
What is the realtionship between the first and third argument to ListSpec?

libs/constrained-generators/src/Constrained/Env.hs
```diff
@@ -25,28 +24,21 @@ data EnvValue where
 deriving instance Show EnvValue
 
 data EnvKey where
-  EnvKey :: Typeable a => !(Var a) -> EnvKey
+  EnvKey :: !(Var a) -> EnvKey
```
TimSheard, line 27: Removing of the Typeable constraint is surprising. At first I thought that the Var constructor had the Typeable constraint, so that made it not required, But I see that is not the case. We must enforce this some where.

Is singletonEnv the only place we make Env?

libs/constrained-generators/src/Constrained/Instances.hs
```diff
@@ -74,7 +79,7 @@ instance BaseUniverse fn => Functions (BoolFn fn) fn where
   propagateSpecFun Or (HOLE :? Value (s :: Bool) :> Nil) spec = caseBoolSpec spec (okOr s)
   propagateSpecFun Or (Value (s :: Bool) :! NilCtx HOLE) spec = caseBoolSpec spec (okOr s)
 
-  mapTypeSpec Not _ = typeSpec ()
+  mapTypeSpec Not (SumSpec a b) = typeSpec $ SumSpec b a
```
TimSheard, line 82: Have we changed the TypeSpec of Bool from () to SumSpec () ()
I didn't notice.
Or is there something more subtle here?

libs/constrained-generators/src/Constrained/Spec/Generics.hs
```diff
@@ -414,11 +422,126 @@ instance
...
+  Select fn n (a : a' : as)
+  where
+  select_ = select_ @fn @(n - 1) @(a' : as) . app sndFn
+
+class IsConstrOf (c :: Symbol) b sop where
+  mkCases ::
+    (HasSpec fn b, All (HasSpec fn) (Cases (SOP sop))) =>
+    (forall a. Term fn a -> Pred fn) ->
+    (Term fn b -> Pred fn) ->
+    List (Binder fn) (Cases (SOP sop))
```
TimSheard, line 454: OK, so we are now doing something new for Constructors?
Maybe some explanation.
Was the addition of this  performance issue?

libs/constrained-generators/src/Constrained/Univ.hs
```diff
@@ -30,10 +32,25 @@ class FunctionLike fn where
   -- | The semantics of a function is given by `sem`
   sem :: fn as b -> FunTy as b
 
-class Member (fn :: [Type] -> Type -> Type) fnU where
+-- | Evidence that `fn` exists somewhere in `fnU`. To make it possible
+-- for `fnU` to consist of a balanced tree of functions we need the `path`
+-- to help GHC disambiguate instances of this class during solving.
+class IsMember (fn :: [Type] -> Type -> Type) fnU (path :: [PathElem]) where
   -- | Inject a specific function symbol into a function
   -- universe.
```
TimSheard, line 40: I take this is a performance improvement. Using a Tree rather than a linear list?

libs/constrained-generators/src/Constrained/Univ.hs
```diff
@@ -85,18 +101,58 @@ data
   = OneofLeft !(fn fnU as b)
   | OneofRight !(fn' fnU as b)
 
+-- | Build a balanced tree of `Oneof` from a list of function universes.
+-- NOTE: it is important that this is a balanced tree as that reduces the amount of
+-- overhead in `injectFn` and `extractFn` from `O(n)` to `O(log(n))`. Surprisingly,
+-- we've observed this making more than 10% difference in runtime on some generation
+-- tasks even though the list `as` is typically small (< 20 elements).
```
TimSheard, line 108: So, just reordering a linear list, so that the most common things come earlier in the list could give a 10% improvement, with much simpler code.

TimSheard (approved):
Every thing looks good. It would be nice to record in the comments of the PR what changes were made for performance, and what changes were made to do the Proposals. Also It would be nice to see what was the overall improvement in time , and how each improvement contributed.

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/V2/Conway.hs
MaximilianAlgehed: I think its just because there happened to be some library function for converting versions to Word64. We could also change it to Word8 or something.

libs/constrained-generators/src/Constrained/Base.hs
MaximilianAlgehed: It probably shouldn't be commented out, I can just remove it.

But basically the idea is that we try to reduce the number of calls to simplifyTerm as much as possible (because that in turn reduces the number of traversal of terms etc.) so we are now a but more careful about ensuring that terms are simplified before we end up computing contexts, for example.

libs/constrained-generators/src/Constrained/Base.hs
MaximilianAlgehed: Yeah, better to only run the optimiser at the last moment when you actually want to use a spec.

libs/constrained-generators/src/Constrained/Base.hs
MaximilianAlgehed: Reproducibility when profiling

libs/constrained-generators/src/Constrained/Base.hs
MaximilianAlgehed: Because it wasn't necessary to use it on every recursive call, which created quadratic behaviour.

libs/constrained-generators/src/Constrained/Base.hs
MaximilianAlgehed: Yup

libs/constrained-generators/src/Constrained/Base.hs
MaximilianAlgehed: Basically, the system maintains the invariant that variable indices are not re-used across types (`Var 5` can't appear as both a `Var Int` and a `Var Bool` in the same program) - so there is no need to be super scared about types for comparison.

libs/constrained-generators/src/Constrained/Base.hs
MaximilianAlgehed: Performance, there were places before where we would use the free vars computation that now uses these specialized functions.

libs/constrained-generators/src/Constrained/Base.hs
MaximilianAlgehed: I can't think of a reason off the top of my head so it's probably a good idea! I'll try it.

libs/constrained-generators/src/Constrained/Env.hs
MaximilianAlgehed: Basically the same reason as before, we have the Var index invariant so we can get away with not doing super expensive type casting

libs/constrained-generators/src/Constrained/Instances.hs
MaximilianAlgehed: Yes, it's now a Generic instance. This change in turn allows us to implement ifElse

libs/constrained-generators/src/Constrained/Univ.hs
MaximilianAlgehed: Yes, it makes a big difference. Somewhere in the 10-20% range at the time when I was implementing this particular optimisation

libs/constrained-generators/src/Constrained/Univ.hs
MaximilianAlgehed: Not quite, remember that everything in the system is always traversing these Oneof trees to get at the function symbol, with the consequence that its also traversing dictionaries for classes and a whole bunch of other stuff. But crucially it's doing this _for every function_ no matter where it is in the tree.

Yes, if you were to have a big enough benchmark suite you could probably figure out an "optimal" order, but that would in turn be super fragile for the actual user. If they were doing things that happen to appear late in the list, they would get a performance penalty. When you do it like this performance is much more predictable because everything is at the same depth (and that depth is close to what you would have at the start of the linear list).

libs/constrained-generators/src/Constrained/Spec/Generics.hs
MaximilianAlgehed: It's an expressiveness thing. All in "client code". Basically you want convenient functions for saying "if its this constructor, these constraints apply" and "it has to be this constructor" without having to write the big caseOf statements

libs/constrained-generators/src/Constrained/Base.hs
MaximilianAlgehed: It is semantically the same in this case

MaximilianAlgehed: But note that we are no longer in the instance for Booleans, this is the instance for `()`

libs/constrained-generators/src/Constrained/Base.hs
MaximilianAlgehed: The first is the hint for controlling test case distributio, it's not checked. Whereas the third is what is derived from checked constraints.

There is no particular reason why it had to be a maximum, one could also try to have it be an average. However, implementing average would be _super difficult_ because you don't know the constraints on size and sums etc.

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/V2/Conway.hs
MaximilianAlgehed: It's a shorthand for the thing that builds the constructor term using the right combination of `fromGeneric_` and `InjLeft/InjRight`. It makes it much more convenient to build a lot of things in client code.

libs/constrained-generators/src/Constrained/Base.hs
MaximilianAlgehed: Derp, the literals here are of different type so equality doesn't matter here.

MaximilianAlgehed: You can't know if they're equal because you don't know if they have the same type or not

PR #4207, opened 2024-03-19
Added tests from a bug report

Soupstraw:
# Description

Added some deserialization tests from #4198.

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/libs/cardano-ledger-binary/cardano-ledger-binary.cabal
+++ b/libs/cardano-ledger-binary/cardano-ledger-binary.cabal
@@ -1,6 +1,6 @@
 cabal-version: 3.0
 name:          cardano-ledger-binary
-version:       1.3.1.0
+version:       1.3.2.0
 license:       Apache-2.0
 maintainer:    operations@iohk.io
 author:        IOHK
--- a/libs/cardano-ledger-binary/src/Cardano/Ledger/Binary/Decoding.hs
+++ b/libs/cardano-ledger-binary/src/Cardano/Ledger/Binary/Decoding.hs
@@ -12,6 +12,7 @@ module Cardano.Ledger.Binary.Decoding (
   decodeFullDecoder',
   decodeFullAnnotator,
   decodeFullAnnotatedBytes,
+  decodeFullAnnotatorFromHexText,
   module Cardano.Ledger.Binary.Version,
   module Cardano.Ledger.Binary.Decoding.DecCBOR,
   module Cardano.Ledger.Binary.Decoding.Sharing,
@@ -48,6 +49,7 @@ import Cardano.Ledger.Binary.Decoding.Decoder
 import Cardano.Ledger.Binary.Decoding.Drop
 import Cardano.Ledger.Binary.Decoding.Sharing
 import Cardano.Ledger.Binary.Decoding.Sized
+import Cardano.Ledger.Binary.Plain (withHexText)
 import Cardano.Ledger.Binary.Version
 import Codec.CBOR.Read as Read (DeserialiseFailure, IDecode (..), deserialiseIncremental)
 import Codec.CBOR.Write (toStrictByteString)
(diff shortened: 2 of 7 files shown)


eras/conway/impl/test/Test/Cardano/Ledger/Conway/BinarySpec.hs
```diff
@@ -1,4 +1,5 @@
 {-# LANGUAGE DataKinds #-}
+{-# LANGUAGE RankNTypes #-}
```
lehins: Seems like a redundant addition
```suggestion
```

libs/cardano-ledger-binary/src/Cardano/Ledger/Binary/Decoding.hs
```diff
@@ -188,6 +190,16 @@ decodeFullAnnotatedBytes v lbl decoder bytes =
   annotationBytes bytes <$> decodeFullDecoder v lbl decoder bytes
 {-# INLINE decodeFullAnnotatedBytes #-}
 
+decodeFullAnnotatorFromHexText ::
```
lehins, line 193: Adding this function to a changelog would be a good idea.

Also, we just had a release to CHaPs, so a minor version needs a bump

libs/cardano-ledger-binary/src/Cardano/Ledger/Binary/Plain.hs
```diff
@@ -72,12 +74,15 @@ showDecoderError = formatToString build
 serializeAsHexText :: ToCBOR a => a -> Text.Text
 serializeAsHexText = Text.decodeLatin1 . B16.encode . serialize'
 
--- | Try decoding base16 encode bytes and then try to decoding them as CBOR
-decodeFullFromHexText :: FromCBOR a => Text.Text -> Either DecoderError a
-decodeFullFromHexText txt =
+withHexText :: (ByteString -> Either DecoderError b) -> Text.Text -> Either DecoderError b
+withHexText f txt =
```
lehins, line 78: This one needs a mention in the changelog too

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Binary/Bugs.hs
```diff
@@ -0,0 +1,57 @@
+{-# LANGUAGE AllowAmbiguousTypes #-}
+{-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE RankNTypes #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeApplications #-}
+
+module Test.Cardano.Ledger.Conway.Binary.Bugs where
```
lehins: Bad name for a module, see my other comment
```suggestion
module Test.Cardano.Ledger.Conway.Binary.Regression where
```

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Binary/Bugs.hs
```diff
@@ -0,0 +1,57 @@
...
+import Cardano.Ledger.Conway.Core (EraTx (..), eraProtVerLow)
+import Test.Cardano.Ledger.Common (NFData, Spec, describe, expectRightDeep_, it)
+
+spec ::
+  forall era.
+  ( EraTx era
+  , NFData (Tx era)
+  ) =>
+  Spec
+spec = describe "Bugs" $ do
```
lehins: I would rather call it a regression test (or even a golden), since there aren't any bugs here.
```suggestion
spec = describe "Regression" $ do
```

lehins (commented):
Looks great. 

Side note, I have no idea how to reproduce the bug in #4198 nor do I have any thoughts on how such behavior could be explained

libs/cardano-ledger-binary/cardano-ledger-binary.cabal
```diff
@@ -1,6 +1,6 @@
 cabal-version: 3.0
 name:          cardano-ledger-binary
-version:       1.3.1.0
+version:       1.3.2.0
```
lehins, line 3: One last thing that I didn't notice last time reviewing this PR is that `cardano-ledger-conway` need a lower bound on the `cardano-ledgder-biary >= 1.3.2`, since it is using the new functions

lehins (approved):
This doesn't exactly test the bug described in #4198, but it doesn't mean we can't keep this test around.

PR #4208, opened 2024-03-20
Remove missingScriptsSymmetricDifference

### LINKED ISSUE #3796: Remove `missingScriptsSymmetricDifference`
We no longer need to support an inferior predicate failure:

```haskell
      failureUnless (sNeeded == sProvided) $
        MissingScriptWitnessesUTXOW (sNeeded `Set.difference` sProvided)
```

This also means that `validateMissingScripts` in shelley no longer need to accept PParams as an argument.

neilmayhew:
Closes #3796

# Description

Extraneous scripts have had a specific predicate failure since Babbage and we no longer wish to support the backwards compatibility that this function enables.

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/eras/shelley/impl/CHANGELOG.md
+++ b/eras/shelley/impl/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## 1.10.0.0
 
+* Remove the `PParams` param from `validateMissingScripts`
+* Remove the `missingScriptsSymmetricDifference` function
 * Add `NFData` instance for `AdaPots`, `ShelleyDelegEvent`
 * Add `Generic`, `Eq` and `NFData` instances for:
   * `ShelleyDelegsEvent`
--- a/eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Utxow.hs
+++ b/eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Utxow.hs
@@ -72,7 +72,6 @@ import Cardano.Ledger.Rules.ValidationMode (
 import Cardano.Ledger.SafeHash (extractHash, hashAnnotated)
 import Cardano.Ledger.Shelley.Core
 import Cardano.Ledger.Shelley.Era (ShelleyEra, ShelleyUTXOW)
-import qualified Cardano.Ledger.Shelley.HardForks as HardForks
 import Cardano.Ledger.Shelley.LedgerState.Types (UTxOState (..))
 import Cardano.Ledger.Shelley.PParams (ProposedPPUpdates (ProposedPPUpdates), Update (..))
 import Cardano.Ledger.Shelley.Rules.Ppup (ShelleyPpupPredFailure)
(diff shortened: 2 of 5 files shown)


eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Utxow.hs
```diff
@@ -406,23 +405,16 @@ validateFailedNativeScripts (ScriptsProvided scriptsProvided) tx = do
 {-  sNeeded := scriptsNeeded utxo tx                             -}
 {-  sProvided := Map.keysSet (tx ^. witsTxL . scriptTxWitsL)     -}
 validateMissingScripts ::
-  EraPParams era =>
-  PParams era ->
   ShelleyScriptsNeeded era ->
   ScriptsProvided era ->
   Test (ShelleyUtxowPredFailure era)
-validateMissingScripts pp (ShelleyScriptsNeeded sNeeded) scriptsprovided =
-  if HardForks.missingScriptsSymmetricDifference (pp ^. ppProtocolVersionL)
```
teodanciu, line 415: I wonder if we are sure that this case hasn't happened on mainnet -  in theory we can remove these checks only if that is indeed the case. 
But since there's a ticket about it which doesn't mention anything about testing, I suppose it's safe to  assume that it's safe to do. 
Maybe we can ask @disassembler  at the meeting today.

eras/shelley/impl/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## 1.10.0.0
 
+* Remove `missingScriptsSymmetricDifference` from `Cardano.Ledger.Shelley.HardForks` because
```
teodanciu: It would be great if you could move the explanations to the commit message, and leave in the changelog just the description of the changes
```
* Remove `missingScriptsSymmetricDifference` from `Cardano.Ledger.Shelley.HardForks`
* Remove the `PParams` param from `validateMissingScripts`
```
to keep the changelog slim, because we want it to just describe the "surface changes" for anyone using us.

neilmayhew: Thanks, that's good guidance

eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Utxow.hs
neilmayhew: Summarizing our discussion from the meeting today:

1. The protocol version in `PParams` will change as we process blocks from genesis through to the present.
2. The effect of this PR is to change the specific predicate failures that result from an invalid transaction, but not to change the validity of a given transaction.
3. We _think_ that there will be no invalid transactions in the history before Babbage, but @disassembler thought that it might be possible with some smart contracts. This could be tested, if necessary.
4. We _think_ that the specific predicate failures don't matter, and only the validity matters.
5. We'll wait for @lehins to give an opinion on this next week.

@teodanciu Please add a comment if this isn't accurate.

teodanciu (approved):
Looks good to me

Issue #4171, opened 2024-03-06
Adjust stake pool distribution calculation


lehins:
Current stake distribution calculation does not consider the amount in the reward account if there are no addresses in the UTxO associated with the stake credential

In order to fix that we need to iterate over the delegations in the UMap instead of the incremental stake map: https://github.com/IntersectMBO/cardano-ledger/blob/3311e074cd3ea25262439b284d9966ef60fce4d6/eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState/IncrementalStake.hs#L220

~It is important to fix this behavior for Conway only with the branch on a hardfork.~ Not an issue

This work will also require an imp test that checks that having ADA in the reward account with no associated stake in the UTxO does affect the distribution in Conway, but not in prior eras.

TimSheard:
I am having trouble understanding this. 
The stake distribution is define by this equation
```
(dom poolDelegs) ◁ (aggregate+ (credStake ∪ ptrStake ∪ rewards))
```
We compute it using this equivalent formulation
```
aggregate+ ((dom poolDelegs ◁ credStake) ∪ (dom poolDelegs ◁ ptrStake) ∪ (dom poolDelegs ◁ rewards))
```
The discussion above seems to indicate that there is something wrong with `(dom poolDelegs ◁ credStake) ` but I don't think that is the case.  If we are missing things from the rewards, then those must come from this computation `(dom poolDelegs ◁ rewards))`.

Either the computation is incorrect, or the way we combine the two is incorrect. 

Now the `rewards ` part of the UMap keeps track of both earned rewards and deposits paid to register the credential.  Which of those two do we include in the the distribution attributed to rewards? 

Finally, how do we combine the two? I think this is where we go wrong.

So I created a test that checks the assertion that the _Current stake distribution calculationdoes not consider the amount in the reward account if there are no addresses in the UTxO associated with the stake credential_ to be False. It does indeed add the stake in rewards accounts even if that credential does not appear in any UTxO entry. So its seems that  ` (dom poolDelegs ◁ rewards)` is doing its job correctly.

TimSheard:
I think it is worthwhile to think critically about what goes into a Distribution. A distribution sums up all the Coin controlled by 
each entity, across some Set of Ada Pots. It would have a type like: (Map entityId Coin).

All Coin in the system is stored in one of the 6 Ada pots.
```
data AdaPots = AdaPots
  { treasuryAdaPot :: Coin
[...]

PR #4211, opened 2024-03-21
Adds new tools to create unit tests over Ledger types

TimSheard:
# Description
This was partly addresses issue #4171 .

That issue identified some potential problems with the code in  [eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState/IncrementalStake.hs]
And suggested some tests to resolve the issue. Such a test need specific values with types such as Credential, StakeReference, UtxO,   IncrementalStake, PParams, Rewards, Deposits, all with complex interrelation-ships. 

The module Test.Cardano.Ledger.UnitTestTools provided the tools to make such values clearly and succinctly. Here is an example

```

-- There are 2 Pools
pool1 = poolHash 30
pool2 = poolHash 31

-- There are 5 wallets, each with its own Credential
tom = stakeCred 1
ann = stakeCred 2
ron = stakeCred 3
john = stakeCred 4
mary = stakeCred 5

-- They each own an address
tomAddr = addr 1 0 -- 0 means tomAddr  does not have a StakeReference
annAddr = addr 2 2
ronAddr = addr 3 3
johnAddr = addr 4 4
maryAddr = addr 5 0 -- 0 means maryAddr  does not have a StakeReference

-- Each wallet has registered its credential,
rewards :: Map (Credential 'Staking StandardCrypto) RDPair
rewards = Map.fromList  -- (rdpair reward deposit) 
    [ (tom,rdpair 5 6)  -- only rewards should be distributed
    , (ann,rdpair 7 6)  -- so we should see 53 distributed
    , (ron,rdpair 11 6)
    , (john,rdpair 13 6)
    , (mary,rdpair 17 6) ]
    
rdpair :: Word64 -> Word64 -> RDPair
rdpair x y = RDPair (CompactCoin x) (CompactCoin y)

-- Each wallet delegates to one of the pools
delegations :: Map (Credential 'Staking StandardCrypto) (KeyHash 'StakePool StandardCrypto)
delegations = Map.fromList
    [ (tom,pool1)   -- since every one is delegated 
    , (ann,pool2)   -- no ones stake should be left out
    , (ron,pool1)
    , (john,pool2)
    , (mary,pool2) ]


-- Each wallet has one or more UTxO entries
-- Since tom and mary use a StakeRefNull those entries will not be distributed
utxo = UTxO(Map.fromList
   [(txIn 40 1,txOut tomAddr (Coin 23))  -- Not distrubuted 23
   ,(txIn 40 2,txOut tomAddr (Coin 29))  -- Not distributed 29
   ,(txIn 42 4,txOut annAddr (Coin 31))
   ,(txIn 43 1,txOut ronAddr (Coin 33))
   ,(txIn 44 2,txOut johnAddr (Coin 41))
   ,(txIn 45 1,txOut maryAddr (Coin 43)) -- Not distributed 43
   ])                 --total 200 - 95  105 should be distributed     

umap :: UMap StandardCrypto
umap = unify rewards Map.empty delegations Map.empty
```

# Checklist

- [x] Self-reviewed the diff


Soupstraw:
It looks to me like the tools duplicate a lot of what we already have in `ImpTest`. For example `tomAddr, annAddr, ronAddr, johnAddr, maryAddr` etc. is already doable in ImpTest without having to provide indices manually:
```haskell
pool1 <- freshKeyHash
...
tomAddr <- freshKeyAddr
annAddr <- freshKeyAddr
...
```

Also we're trying to get rid of tests in `test-suite` directories and move them to `testlib`, so it's better to write the tests there instead. New tests should be written using `Spec` from `hspec` instead of `TestTree` from `tasty`, since we agreed to move the tests to `hspec`. It'd be really nice if you could rewrite the tests using `ImpTestM`, it makes the tests much easier to follow and we're planning to move older tests to ImpTest anyways.

You can take a look at some of the existing ImpTests to get an idea:
https://github.com/IntersectMBO/cardano-ledger/blob/master/eras/shelley/impl/testlib/Test/Cardano/Ledger/Shelley/Imp/LedgerSpec.hs
https://github.com/IntersectMBO/cardano-ledger/blob/master/eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/EpochSpec.hs

teodanciu (commented):
Moving `RawSeed` from shelley to core makes sense to me :+1: 

I understand there is the need for fine grained control over the state to allow us to test various scenarios,  however  it seems to me that the style of the new test,  `incrStakeUnitTest`  (and the utilities  involved in writing it) are a little bit in the direction opposite from the one we're trying to move in.  
We even have some tickets to translate  to `Imp`-style ,  tests that construct the state in such a way.

The test even has "characters" , which is great, because it describes a concrete scenario, - so why not have these characters perform the actions that we need using the Imp "primitives" like  `registerPool` or `registerReward` ? I think that would make the test easier to understand and also less prone to errors (like mixing up indices for example, which would create incorrect pre-conditions). 
Maybe we can try to write an equivalent Imp-style test  (we could have an "Imp-party" with @Soupstraw  and everyone interested if you like ) and see how that compares to this test, whether it allows us enough to achieve what is being tested here, what do you think?

libs/cardano-ledger-core/testlib/Test/Cardano/Ledger/UnitTestTools.hs
lehins, line 1: There is no way this ugliness is getting into the cardano-ledger-core:testlib

1. Function naming is terrible
2. `RawSeed` is a complete nonsense that needs to be removed
3. We have quickcheck that can give us random values, why do we need to write tests that hard code integers? All is necessary is changing the unit test into a property test that creates one unit case.


We need to get away from this pattern of using integers for creating hashes. All of those function are really useless and can be in all cases be replaced by using an arbitrary instance for that type

libs/cardano-ledger-core/testlib/Test/Cardano/Ledger/UnitTestTools.hs
```diff
@@ -0,0 +1,172 @@
...
+import Data.Ratio ((%))
+import Data.Word (Word64)
+import Test.Cardano.Ledger.Core.Arbitrary (RawSeed (..))
+import Test.Cardano.Ledger.Core.KeyPair (KeyPair (..), mkKeyPairWithRawSeed, mkWitnessVKey)
+
+-- =======================================================
+
+-- By changing the parameter 'n', we get a different keyPair
+theKeyPair :: Crypto c => Int -> KeyPair kr c
+theKeyPair n = KeyPair a b
```
lehins, line 92: There is already a function in core testlib that does just this: https://github.com/IntersectMBO/cardano-ledger/blob/2924845c389b0890cd3e72a44ffb7e78cf664427/libs/cardano-ledger-core/testlib/Test/Cardano/Ledger/Core/KeyPair.hs#L184-L185

libs/cardano-ledger-core/testlib/Test/Cardano/Ledger/UnitTestTools.hs
```diff
@@ -0,0 +1,172 @@
...
+  PoolParams
+    poolhash
+    (hash 77)
+    (Coin 1)
+    (Coin 3)
+    (fromJust (boundRational (1 % 2)))
+    (rewardAccount rewAcctNum)
+    mempty
+    mempty
+    SNothing
```
lehins, line 172: Stuff like this with magic values does not belong in a reusable library. It is expected for test cases to have hard coded values like that.

eras/shelley/test-suite/src/Test/Cardano/Ledger/Shelley/Rules/IncrementalStake.hs
lehins, line 1: We should not be expending `cardano-ledger-shelley-test`. All new Shelley related tests should go into `cardano-ledger-shelley:tests` target.

eras/shelley/test-suite/src/Test/Cardano/Ledger/Shelley/Rules/IncrementalStake.hs
```diff
@@ -228,3 +233,117 @@ aggregateUtxoCoinByCredential ptrs (UTxO u) initial =
...
+
+-- Each wallet has one or more UTxO entries
+-- Since tom and mary use a StakeRefNull those entries will not be distributed
+utxo1 :: UTxO (ShelleyEra StandardCrypto)
+utxo1 =
+  UTxO
+    ( Map.fromList
+        [ (txIn 40 1, txOut tomAddr (Coin 23)) -- Not distrubuted 23
+        , (txIn 40 2, txOut tomAddr (Coin 29)) -- Not distributed 29
+        , (txIn 42 4, txOut annAddr (Coin 31))
```
lehins, line 294: We should not be writing unit tests with bindings at the top level that depend on othe top level bindings. 

Old test suite is full of this sort of patterns and this makes the test suite very fragile to changes and extremely confusing. We must avoid using this as many magic values to local bindings in specific test cases) as possible.

eras/shelley/test-suite/src/Test/Cardano/Ledger/Shelley/Rules/IncrementalStake.hs
```diff
@@ -228,3 +233,117 @@ aggregateUtxoCoinByCredential ptrs (UTxO u) initial =
...
+
+umap :: UM.UMap StandardCrypto
+umap = UM.unify rewards Map.empty delegations Map.empty
+
+dState :: DState (ShelleyEra StandardCrypto)
+dState = dZero {dsUnified = umap}
+  where
+    dZero = def @(DState (ShelleyEra StandardCrypto))
+
+pState :: PState (ShelleyEra StandardCrypto)
```
lehins, line 315: Restricting this test case to Shelley era makes it not very useful for all future eras.
We should really try writing test cases that work for eras that they must hold for. In this case it is all known eras.

eras/shelley/test-suite/src/Test/Cardano/Ledger/Shelley/Rules/IncrementalStake.hs
```diff
@@ -228,3 +233,117 @@ aggregateUtxoCoinByCredential ptrs (UTxO u) initial =
...
+utxo1 :: UTxO (ShelleyEra StandardCrypto)
+utxo1 =
+  UTxO
+    ( Map.fromList
+        [ (txIn 40 1, txOut tomAddr (Coin 23)) -- Not distrubuted 23
+        , (txIn 40 2, txOut tomAddr (Coin 29)) -- Not distributed 29
+        , (txIn 42 4, txOut annAddr (Coin 31))
+        , (txIn 43 1, txOut ronAddr (Coin 33))
+        , (txIn 44 2, txOut johnAddr (Coin 41))
+        , (txIn 45 1, txOut maryAddr (Coin 43)) -- Not distributed 43
```
lehins, line 297: There is no need for `tom` or `mary` to have anything in the utxo, but we can keep one of them in.
```suggestion
```

lehins (changes requested):
It would have been nicer to have an Imp test for this, but I can see how specialized tests that do not rely on the rules have their own  value as well. If we not gonna have an imp test, then we should at least convert this into a property test instead. The goal is:

* Replace all magic values with ones generated by quickcheck
* Replace creation of hashes with ones generated by quickcheck
* Make the property tests work for all eras
* Start using hspec instead of tasty by not using `cardano-ledger-[era]-test` packages

libs/cardano-ledger-core/testlib/Test/Cardano/Ledger/UnitTestTools.hs
```diff
@@ -0,0 +1,172 @@
...
+theKeyPair :: Crypto c => Int -> KeyPair kr c
+theKeyPair n = KeyPair a b
+  where
+    (b, a) = mkKeyPairWithRawSeed (RawSeed 0 0 0 0 (fromIntegral n))
+
+-- =============================================================
+
+-- | Creates a StakeReference for use in an Addr.
+--   Note that 0 has special meaning, returning StakeRefNull,
+--   Meaning this Addr does not stake it Coin to anything.
```
aniketd, line 100:
```suggestion
--   Meaning this Addr does not stake it's Coin to anything.
```

Soupstraw:
```suggestion
--   Meaning this Addr does not stake its Coin to anything.
```
:nerd_face:

eras/shelley/test-suite/src/Test/Cardano/Ledger/Shelley/Rules/IncrementalStake.hs
TimSheard: done

eras/shelley/test-suite/src/Test/Cardano/Ledger/Shelley/Rules/IncrementalStake.hs
TimSheard: fixed

TimSheard:
Replaced by PR #4242

PR #4212, opened 2024-03-22
Update and reorganize conformance tests

Soupstraw:
# Description

This PR does some reorganizing of the V2 constrained generators and conformance tests.
 * Conway era V2 specs are moved from `cardano-ledger-test` to `cardano-ledger-conway:testlib`
 * Made a `ExecutableSpecRule` typeclass interface to have a common interface for running conformance tests for each rule

This only touches the infrastructure for running conformance tests with V2 constrained generators. The tests themselves will be added in a follow-up PR.

# Checklist

- [ ] Self-reviewed the diff


eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Constrained/Spec/Deleg.hs
```diff
@@ -16,7 +14,10 @@ import Lens.Micro
 
 import Constrained
 
-import Test.Cardano.Ledger.Constrained.V2.Conway
+import Cardano.Ledger.Conway (ConwayEra)
+import Cardano.Ledger.Core (PParams, ppKeyDepositL)
+import Cardano.Ledger.Crypto (StandardCrypto)
```
MaximilianAlgehed: How come all these extra imports are needed?

libs/cardano-ledger-conformance/src/Test/Cardano/Ledger/Conformance/ExecutableSpecRule.hs
```diff
@@ -0,0 +1,138 @@
...
+  logEntry,
+  logToExpr,
+  tryRunImpRule,
+ )
+
+class
+  ( CV2.HasSpec fn (Environment (EraRule rule era))
+  , CV2.HasSpec fn (State (EraRule rule era))
+  , CV2.HasSpec fn (Signal (EraRule rule era))
+  , CV2.HasSpec fn (ExecContext fn rule era)
```
MaximilianAlgehed: Did you ever figure out what was throwing the type checker into a loop here?

libs/cardano-ledger-conformance/src/Test/Cardano/Ledger/Conformance/ExecutableSpecRule.hs
```diff
@@ -0,0 +1,138 @@
...
+  , Inject (ExecContext fn rule era) (SpecTransContext (Environment (EraRule rule era)))
+  , Inject (ExecContext fn rule era) (SpecTransContext (State (EraRule rule era)))
+  , Inject (ExecContext fn rule era) (SpecTransContext (Signal (EraRule rule era)))
+  , Inject (ExecContext fn rule era) (SpecTransContext (PredicateFailure (EraRule rule era)))
```
MaximilianAlgehed: I would consider abstracting over the whole `C` has to hold for `Environment (EraRule rule era)`, `State`, `Signal`, and `PredicateFailure`. That way you'd have one third the number of constraints here.

libs/cardano-ledger-conformance/src/Test/Cardano/Ledger/Conformance/ExecutableSpecRule.hs
Soupstraw: It went into a loop when I used `ConwayFn` directly instead of the `fn` parameter. I didn't really look into why that's so.

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Constrained/Spec/Deleg.hs
Soupstraw: I think they were re-exported by `Cardano.Ledger.Api`, but we lost access to that since this file is no longer in `cardano-ledger-test`

CONTRIBUTING.md
```diff
@@ -272,6 +272,21 @@ from [the Shelley ledger spec](./eras/shelley/formal-spec)).
...
 
+### To update the conformance test
+
+To update the conformance test, do the following:
+
+1. Clone the [Agda specification repo](https://github.com/IntersectMBO/formal-ledger-specifications)
+2. Run `nix-build -A ledger.hsSrc` in the cloned repo, mark down the output path
+   in the nix store
+3. Clone the [executable spec repo](https://github.com/input-output-hk/cardano-ledger-executable-spec)
+4. Delete all the files in the repo and copy
```
teodanciu: I find this a little confusing:
>  Delete all files in the repo and copy .. into the executable spec repo. 

Are we talking about the same repo in both cases (the same as the one cloned at the step above)? 
How about something like
4. Replace the content of the repo cloned above with the files at `/nix/store/<output of the nix-build>/haskell/Ledger/*`
```
rm -rf cardano-ledger-executable-spec/*
cp -r  /nix/store/<output of the nix-build>/haskell/Ledger/*  cardano-ledger-executable-spec
```
?

Soupstraw: Yeah you understood correctly. I'll change that line to what you proposed.

CONTRIBUTING.md
```diff
@@ -272,6 +272,23 @@ from [the Shelley ledger spec](./eras/shelley/formal-spec)).
...
+
+1. Clone the [Agda specification repo](https://github.com/IntersectMBO/formal-ledger-specifications)
+2. Run `nix-build -A ledger.hsSrc` in the cloned repo, mark down the output path
+   in the nix store
+3. Clone the [executable spec repo](https://github.com/input-output-hk/cardano-ledger-executable-spec)
+4. Replace the content of the repo cloned above with the files at `/nix/store/<output of the nix-build>/haskell/Ledger/*`
+   ```bash
+   rm -rf cardano-ledger-executable-spec/*
+   cp -r  /nix/store/<output of the nix-build>/haskell/Ledger/*  cardano-ledger-executable-spec
+   ```
```
teodanciu, line 287: Oops sorry, you also deleted the "commit" instruction.  You can leave it out if you think it's obvious, just pointing it out in case it was by accident.

libs/cardano-ledger-conformance/test/Main.hs
```diff
@@ -1,10 +1,4 @@
...
 
-import qualified Test.Cardano.Ledger.Conformance as Conformance
-import Test.Cardano.Ledger.Imp.Common (describe, ledgerTestMain)
-
 main :: IO ()
-main =
-  ledgerTestMain $
-    describe "Conformance" $ do
-      Conformance.spec
+main = pure () -- TODO remove this module?
```
teodanciu: I would agree to remove this, since, if I understand correctly, we want the entry point of the conformance tests to be in conway tests. 
And if we do that, we can probably remove the whole test suite from the caradano-ledger-conformance.cabal file.

libs/cardano-ledger-core/src/Cardano/Ledger/Tools.hs
```diff
@@ -12,6 +11,7 @@ module Cardano.Ledger.Tools (
   estimateMinFeeTx,
   addDummyWitsTx,
   integralToByteStringN,
+  byteStringToNum,
```
teodanciu, line 14: I think you should bump the version of core to `1.11.1.0`, add the change to changelog, and update the dependency in conway testlib.

eras/conway/impl/cardano-ledger-conway.cabal
```diff
@@ -140,14 +142,20 @@ library testlib
...
+        cardano-crypto-wrapper,
+        cardano-ledger-executable-spec,
+        cardano-ledger-conformance,
         cardano-ledger-allegra:{cardano-ledger-allegra, testlib},
         cardano-ledger-alonzo:{cardano-ledger-alonzo, testlib},
         cardano-ledger-binary,
         cardano-ledger-babbage:{cardano-ledger-babbage, testlib},
+        cardano-ledger-byron,
         cardano-ledger-conway,
         cardano-ledger-core:{cardano-ledger-core, testlib},
```
teodanciu: We should depend on cardano-ledger-core: 1.11.1.0 because of the usage of `byteStringToNum`

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Conformance.hs
```diff
@@ -0,0 +1,6 @@
+module Test.Cardano.Ledger.Conway.Conformance (spec) where
+import Test.Cardano.Ledger.Imp.Common
+
+spec :: Spec
+spec = describe "Conway conformance tests" $ do
+  undefined
```
teodanciu: Wouldn't it be better instead of this, to return a test that is passing, and also plug it in to the `Main` that runs it, to make sure the whole "plumming" is complete ?

eras/conway/impl/cardano-ledger-conway.cabal
```diff
@@ -140,14 +142,20 @@ library testlib
...
         containers,
+        constrained-generators,
+        cryptonite,
         plutus-ledger-api,
         deepseq,
         microlens,
         cardano-crypto-class,
+        cardano-crypto-wrapper,
+        cardano-ledger-executable-spec,
+        cardano-ledger-conformance,
```
teodanciu: How will this work with releasing ledger libraries? Won't we have to add  these three: `constrained-generators`,  `cardano-ledger-executable-spec` and `cardano-ledger-conformance` on Chap, if they are dependencies of testlib?

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Conformance.hs
teodanciu: Aha nevermind, I saw you did this in a next commit.

eras/conway/impl/cardano-ledger-conway.cabal
```diff
@@ -140,23 +155,35 @@ library testlib
         bytestring,
         cardano-data:{cardano-data, testlib},
```
teodanciu: I think we should depend on cardano-data:1.2.2.0 here because of  `toListKV` usage

eras/conway/impl/src/Cardano/Ledger/Conway/Governance.hs
```diff
@@ -35,6 +35,7 @@ module Cardano.Ledger.Conway.Governance (
   VotingProcedure (..),
   VotingProcedures (..),
   foldlVotingProcedures,
+  foldrVotingProcedures,
```
teodanciu, line 38: We should probably bump the version and  add this to changelog. 
cardano-ledger-conway version 1.13.1.0 was just released and the changelog not updated, so I think we need a new version and changelog entry:   1.13.2.0

teodanciu (commented):
Some parts of this PR are over my head, but they are the ones that were moved around, so presumably have been reviewed before, so it shouldn't matter! 
If I understand correctly:  atm in master, cardano-ledger-conformance is providing both: 1) the "glue" to make things work with cardano-ledger-executable-spec   and 2) the actual tests.
So in this PR, we left the more generic glue in cardano-ledger-conformance, and we specialized with instances in cardano-ledger-conway:testlib, as well as moving here the tests and running thereof. 
Is my understanding correct? 

I left some comments about version and dependency bounds updates, but my main concern is about the release processs, since cardano-ledger-conway testlib gets released on Chap, but now depends on 3 things that are not on Chap: 
`constrained-generators`, `cardano-ledger-executable-spec` and `cardano-ledger-conformance`.   Will this work? Maybe I'm missing something, apologies if so!

libs/cardano-ledger-conformance/test/Main.hs
Soupstraw: Yeah, I already removed it in a later commit :slightly_smiling_face:

Soupstraw:
@teodanciu 

 `cardano-ledger-conformance` provides the interface for writing conformance tests. It also adds some orphan instances for the generated Agda types. Now that I look at it, I think it would make sense to move the orphan instances to `cardano-ledger-conway:testlib` and the typeclasses to `cardano-ledger-core:testlib` and then get rid of the package altogether.

MaximilianAlgehed:
Much of the old code here is also over my head. But the parts pertaining to moving the constrained stuff around is mostly uncontroversial from my point of view.

@Soupstraw we discussed the need or lack thereof for this class for running the STS tests. I don't mind one way or another - we can always add a type parameter later if there is a need for modifying the specific specs in any way in tests later. Are there any other parts here that you want my specific input on?

libs/cardano-ledger-test/cardano-ledger-test.cabal
```diff
@@ -54,15 +54,6 @@ library
...
         Test.Cardano.Ledger.Constrained.Solver
-        Test.Cardano.Ledger.Constrained.V2.Conway
-        Test.Cardano.Ledger.Constrained.V2.Conway.CERT
-        Test.Cardano.Ledger.Constrained.V2.Conway.DELEG
-        Test.Cardano.Ledger.Constrained.V2.Conway.GOV
-        Test.Cardano.Ledger.Constrained.V2.Conway.GOVCERT
-        Test.Cardano.Ledger.Constrained.V2.Conway.LEDGER
-        Test.Cardano.Ledger.Constrained.V2.Conway.POOL
-        Test.Cardano.Ledger.Constrained.V2.Conway.PParams
-        Test.Cardano.Ledger.Constrained.V2.Conway.UTXO
```
MaximilianAlgehed, line 65: Actually, here is something I do kind of take issue with. Moving these modules out of here makes it much more difficult to make changes to these constraints as you're working with them. This is because we run the tests in `Test.Cardano.Ledger.STS` and while working on these constraints one typically needs to make many small changes and repeatedly reload ghci. This becomes much more cumbersome if the modules live in a different package as every time you want to re-run a test you can no longer just `:r` but now you need to re-start ghci instead.

lehins (changes requested):
Neither ` cardano-ledger-executable-spec` nor `cardano-ledger-conformance` should be released on CHaPs. For this reason `cardano-ledger-conway` cannot depend on them. This is not only because that conformance functionality is not useful for anyone but us, but also because we don't want to impose those massive dependencies on others for no reason.

So, this has to be reorganized back how it was before. All conformance related stuff needs to live in `cardano-ledger-conformance`.

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Constrained/Spec/Utxo.hs
```diff
@@ -3,13 +3,11 @@
 {-# LANGUAGE GADTs #-}
 {-# LANGUAGE ImportQualifiedPost #-}
 {-# LANGUAGE RecordWildCards #-}
-{-# LANGUAGE TypeApplications #-}
 
 -- | Specs necessary to generate, environment, state, and signal
 -- for the UTXO rule
-module Test.Cardano.Ledger.Constrained.V2.Conway.UTXO where
+module Test.Cardano.Ledger.Conway.Constrained.Spec.Utxo where
```
lehins: FYI. Module structure is fine, eg: `Test.Cardano.Ledger.Conway.Constrained.Spec.Utxo` it just should live in the conformance package

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Conformance/ExecutableSpecRule.hs
```diff
@@ -0,0 +1,141 @@
...
+  inject (ConwayGovExecContext x _) = x
+
+instance Inject (ConwayGovExecContext era) (EnactState era) where
+  inject (ConwayGovExecContext _ x) = x
+
+instance
+  ( NFData (SpecRep (ConwayGovPredFailure Conway))
+  , IsConwayUniv fn
+  ) =>
+  ExecutableSpecRule fn "GOV" Conway
```
lehins: Why not `ExecSpecRule` for consistency with `ExecContext`?
```suggestion
  ExecSpecRule fn "GOV" Conway
```

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Conformance/SpecTranslate.hs
```diff
@@ -0,0 +1,877 @@
...
+
+  toSpecRep x =
+    Agda.MkPParams
+      <$> toSpecRep (cppMinFeeA x)
+      <*> toSpecRep (cppMinFeeB x)
+      <*> pure (toInteger . unTHKD $ cppMaxBBSize x)
+      <*> pure (toInteger . unTHKD $ cppMaxTxSize x)
+      <*> pure (toInteger . unTHKD $ cppMaxBHSize x)
+      <*> pure (toInteger . unTHKD $ cppMaxValSize x)
+      <*> pure 0
```
lehins: Would be nice to explain the magic zero
```suggestion
      <*> pure 0 -- minUTxOValue has been deprecated and is not supported in Conway
```

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Constrained/Spec/Utxo.hs
MaximilianAlgehed: c.f. my comment above - could we perhaps also move the `Test.Cardano.Ledger.STS` tests to the conformance package? That way working on the skeleton part of this (getting the STS rules to execute) wouldn't require jumping too much between the packages?

CONTRIBUTING.md
```diff
@@ -272,6 +272,24 @@ from [the Shelley ledger spec](./eras/shelley/formal-spec)).
 4. Add a link to the package near the bottom of [flake.nix](./flake.nix),
    following the existing examples.
 
+### To update the conformance test
+
+To update the conformance test, do the following:
+
+1. Clone the [Agda specification repo](https://github.com/IntersectMBO/formal-ledger-specifications)
+2. Run `nix-build -A ledger.hsSrc` in the cloned repo, mark down the output path
+   in the nix store
```
aniketd:
```suggestion
2. Run `nix-build -A ledger.hsSrc` in the cloned repo, note the output path
   in the nix store
```
I was a bit confused because of the more familiar usage of `markdown` as a document format.

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Conformance/SpecTranslate.hs
```diff
@@ -0,0 +1,877 @@
...
+import Test.Cardano.Ledger.Conformance (
+  SpecTransM,
+  SpecTranslate (..),
+  SpecTranslationError,
+  askTransCtx,
+ )
+import Test.Cardano.Ledger.Conway.Constrained (IsConwayUniv)
+import Test.Cardano.Ledger.Conway.TreeDiff (ToExpr (..))
+
+-- | OpaqueErrorString is behaves like unit in comparisons, but contains an
```
aniketd:
```suggestion
-- | OpaqueErrorString behaves like unit in comparisons, but contains an
```

libs/cardano-data/src/Data/OMap/Strict.hs
```diff
@@ -375,6 +376,9 @@ instance HasOKey k v => IsList (OMap k v) where
   fromList = fromFoldable
   toList = F.toList
 
+toListKV :: OMap k v -> [(k, v)]
+toListKV = Map.toList . toMap
```
aniketd: I think that canonically this representation is called an [`assocList`](https://hackage.haskell.org/package/Agda-2.6.4.3/docs/Agda-Utils-AssocList.html#t:AssocList), like [in the containers package too](https://hackage.haskell.org/package/containers-0.7/docs/Data-Map-Strict.html#v:toAscList)

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Constrained/Spec/Utxo.hs
Soupstraw: I'll try to move stuff out of `cardano-ledger-test` and put it into conway testlib and `cardano-ledger-conformance`.

aniketd (commented):
Noice! :raised_hands:

libs/cardano-ledger-core/src/Cardano/Ledger/Tools.hs
Soupstraw: Do we want to put bounds on testlib dependencies?

eras/conway/impl/cardano-ledger-conway.cabal
Soupstraw: I got rid of those dependencies.

libs/cardano-data/src/Data/OMap/Strict.hs
Soupstraw: Thanks, fixed!

libs/cardano-ledger-test/cardano-ledger-test.cabal
Soupstraw: I moved everything back to `cardano-ledger-test`, just changed the names of the modules a bit so they're consistent with how we name the rules in the implementation.

CONTRIBUTING.md
```diff
@@ -272,6 +272,24 @@ from [the Shelley ledger spec](./eras/shelley/formal-spec)).
...
+
+1. Clone the [Agda specification repo](https://github.com/IntersectMBO/formal-ledger-specifications)
+2. Run `nix-build -A ledger.hsSrc` in the cloned repo, take note of the output path
+   in the nix store
+3. Clone the [executable spec repo](https://github.com/input-output-hk/cardano-ledger-executable-spec)
+4. Replace the content of the repo cloned above with the files at `/nix/store/<output of the nix-build>/haskell/Ledger/*`
+   ```bash
+   rm -rf cardano-ledger-executable-spec/*
+   cp -r  /nix/store/<output of the nix-build>/haskell/Ledger/*  cardano-ledger-executable-spec
+   ```
```
lehins, line 287: Could you add a bash script to https://github.com/IntersectMBO/formal-ledger-specifications that would do the whole process of building and copying? 
We can leave SRP update (step 5) as a manual step.

libs/cardano-ledger-conformance/src/Test/Cardano/Ledger/Conformance/Spec/Conway.hs
```diff
@@ -0,0 +1,17 @@
...
+import Test.Cardano.Ledger.Conformance.ExecutableSpecRule.Conway ()
+import Test.Cardano.Ledger.Constrained.Conway
+import Test.Cardano.Ledger.Conway.ImpTest (withImpState)
+import Test.Cardano.Ledger.Imp.Common
+
+spec :: Spec
+spec = describe "Conway conformance tests" $ do
+  withImpState @Conway $ do
+    xit "UTXO" . replicateM_ 100 $ conformsToImpl @"UTXO" @ConwayFn
+    xit "GOV" . replicateM_ 100 $ conformsToImpl @"GOV" @ConwayFn
```
lehins, line 17: So, `conformsToImpl` is still failing some tests, right? Do you know what is still missing to convert these to `it`?

libs/cardano-ledger-conformance/src/Test/Cardano/Ledger/Conformance/SpecTranslate/Conway.hs
```diff
@@ -0,0 +1,875 @@
...
+  SpecTranslate (..),
+  SpecTranslationError,
+  askTransCtx,
+ )
+import Test.Cardano.Ledger.Constrained.Conway.Instances (IsConwayUniv)
+import Test.Cardano.Ledger.Conway.TreeDiff (ToExpr (..))
+
+-- | OpaqueErrorString behaves like unit in comparisons, but contains an
+-- error string that can be displayed.
+newtype OpaqueErrorString = OpaqueErrorString String
```
lehins, line 138: So, what is the purpose of this hacky type?

Looks like it is used for translating predicate failures, but why do we need to translate them in the first place?

I though it is only possible to test the successful case of ledger rules against the spec, since the spec cannot fail with any useful information.

libs/cardano-ledger-conformance/src/Test/Cardano/Ledger/Conformance/SpecTranslate/Conway.hs
```diff
@@ -0,0 +1,875 @@
...
+
+  toSpecRep x =
+    Agda.MkPParams
+      <$> toSpecRep (cppMinFeeA x)
+      <*> toSpecRep (cppMinFeeB x)
+      <*> pure (toInteger . unTHKD $ cppMaxBBSize x)
+      <*> pure (toInteger . unTHKD $ cppMaxTxSize x)
+      <*> pure (toInteger . unTHKD $ cppMaxBHSize x)
+      <*> pure (toInteger . unTHKD $ cppMaxValSize x)
+      <*> pure 0
```
lehins: Could you add a comment about it being zero because it is for a parameter that has been removed prior to Conway. 

Related question. Do you know why would Conway spec have this deprecated parameter in the first place?

libs/cardano-ledger-conformance/src/Test/Cardano/Ledger/Conformance/ExecutableSpecRule/Core.hs
```diff
@@ -0,0 +1,131 @@
...
+  logToExpr sig
+  logEntry . show $ signalSpec @fn @rule @era env st
+  (execContext :: ctx) <- liftGen . CV2.genFromSpec_ $ execContextSpec @fn @rule @era
+  agdaRes <-
+    impAnn "Translating spec values to SpecRep" . expectRightExpr . runSpecTransM execContext $
+      runAgdaRule @fn @rule @era <$> toSpecRep env <*> toSpecRep st <*> toSpecRep sig
+  implRes <- fmap fst <$> tryRunImpRule @rule @era env st sig
+  implResTest <-
+    impAnn "Translating implementation values to SpecRep" . expectRightExpr . runSpecTransM execContext $
+      bimapM (traverse toTestRep) toTestRep implRes
```
lehins: I don't think you need to translate the Left case with `toTestRep`. I suspect that removing this translation would let you avoid creating a bogus `OpaqueErrorString` type used for predicate failures.

libs/cardano-ledger-conformance/src/Test/Cardano/Ledger/Conformance/ExecutableSpecRule/Core.hs
```diff
@@ -0,0 +1,131 @@
...
+  ExecutableSpecRule fn rule era =>
+  ImpTestM era ()
+conformsToImpl = impAnn "conformsToImpl" . resize 5 $ do
+  env <- liftGen . CV2.genFromSpec_ $ environmentSpec @fn @rule @era
+  logToExpr env
+  st <- liftGen . CV2.genFromSpec_ $ stateSpec @fn @rule @era env
+  logToExpr st
+  sig <- liftGen . CV2.genFromSpec_ $ signalSpec @fn @rule @era env st
+  logToExpr sig
+  logEntry . show $ signalSpec @fn @rule @era env st
```
lehins: I think signal always has a `ToExpr` instance so we can switch to logging using `logToExpr`, right?:
```suggestion
  logToExpr $ signalSpec @fn @rule @era env st
```

libs/cardano-ledger-conformance/src/Test/Cardano/Ledger/Conformance/ExecutableSpecRule/Core.hs
```diff
@@ -0,0 +1,131 @@
...
+  (execContext :: ctx) <- liftGen . CV2.genFromSpec_ $ execContextSpec @fn @rule @era
+  agdaRes <-
+    impAnn "Translating spec values to SpecRep" . expectRightExpr . runSpecTransM execContext $
+      runAgdaRule @fn @rule @era <$> toSpecRep env <*> toSpecRep st <*> toSpecRep sig
+  implRes <- fmap fst <$> tryRunImpRule @rule @era env st sig
```
lehins: This, I think would simplify the implementation below (see my other comment below)
```suggestion
  implResEither <- fmap fst <$> tryRunImpRule @rule @era env st sig
  implRes <- expectRightExpr implResEither
```

libs/cardano-ledger-conformance/src/Test/Cardano/Ledger/Conformance/SpecTranslate/Conway.hs
```diff
@@ -0,0 +1,875 @@
...
+  , Ord (TestRep (PlutusPurpose AsIx era))
+  ) =>
+  SpecTranslate ctx (AlonzoTxWits era)
+  where
+  type SpecRep (AlonzoTxWits era) = Agda.TxWitnesses
+
+  toSpecRep x =
+    Agda.MkTxWitnesses
+      <$> toSpecRep (txwitsVKey x)
+      <*> pure []
```
lehins, line 482: I though Alonzo has been implemented in Agda. Is there still no support for scripts in the spec, or we are just postponing its testing?

libs/cardano-ledger-conformance/src/Test/Cardano/Ledger/Conformance/SpecTranslate/Conway.hs
```diff
@@ -0,0 +1,875 @@
...
+      <*> toSpecRep d
+      <*> pure () -- TODO Are the anchors supposed to be optional?
+  toSpecRep (ConwayTxCertGov (ConwayUnRegDRep c _)) =
+    Agda.DeRegDRep
+      <$> toSpecRep c
+  toSpecRep (ConwayTxCertGov (ConwayUpdateDRep c _)) =
+    Agda.RegDRep
+      <$> toSpecRep c
+      <*> pure 0
+      <*> pure () -- TODO Are the anchors supposed to be optional?
```
lehins, line 576: Yes, they are suppose to be optional, but I don't think anchors are very useful for Agda anyhow, so that's why they are probably just set to unit.

libs/cardano-ledger-conformance/src/Test/Cardano/Ledger/Conformance/SpecTranslate/Conway.hs
```diff
@@ -0,0 +1,875 @@
...
+  toSpecRep (TreasuryWithdrawals withdrawals _) =
+    Agda.TreasuryWdrl
+      <$> toSpecRep withdrawals
+  toSpecRep (NoConfidence _) = pure Agda.NoConfidence
+  toSpecRep (UpdateCommittee _ remove add threshold) =
+    Agda.NewCommittee
+      <$> toSpecRep add
+      <*> toSpecRep remove
+      <*> toSpecRep threshold
+  toSpecRep (NewConstitution _ (Constitution anchor policy)) =
```
lehins, line 777: How do we handle the hash protection on the Agda side if the previous governance action id is not translated to SpecRep?

lehins (approved):
Looks great. I just a couple of minor suggestions and bunch of questions.

libs/cardano-data/src/Data/OMap/Strict.hs
lehins: @aniketd `toAscList` means to ascending list, not assoc list :wink:

libs/cardano-ledger-conformance/src/Test/Cardano/Ledger/Conformance/SpecTranslate/Conway.hs
Soupstraw: The spec can actually give error messages when any of the predicates are not satisfied. Right now the predicate failures are represented as strings in the Agda spec, we might change them to something better in the future. 

For now all failures are treated as a unit type in comparisons because it's difficult to map these strings to PredicateFailures. With this type I can still print out the error given by the Agda spec if either the implementation or the spec fails and the other one succeds. 

Once we have better types for predicate failures in the Agda spec we can get rid of this type.

libs/cardano-ledger-conformance/src/Test/Cardano/Ledger/Conformance/ExecutableSpecRule/Core.hs
Soupstraw: If I don't do the translation then we can't do negative testing with conformance tests.

libs/cardano-ledger-conformance/src/Test/Cardano/Ledger/Conformance/SpecTranslate/Conway.hs
Soupstraw: I haven't updated all of the types yet, I figured I'll do it in another PR since this one is getting pretty big already.

libs/cardano-ledger-conformance/src/Test/Cardano/Ledger/Conformance/SpecTranslate/Conway.hs
Soupstraw: The translations are incomplete at the moment, I'll deal with them in a future PR.

libs/cardano-ledger-conformance/src/Test/Cardano/Ledger/Conformance/ExecutableSpecRule/Core.hs
Soupstraw: This line prints the signal _spec_, which is a complicated type with GADTs and all sorts of fun stuff. It'd probably be a lot of effort to write `ToExpr` instances for all of the types that it contains.

libs/cardano-ledger-conformance/src/Test/Cardano/Ledger/Conformance/SpecTranslate/Conway.hs
Soupstraw: It's there because the types in the spec are not parametrised over eras like in the implementation. It's meant to be used in earlier eras and is just ignored in Conway.

libs/cardano-ledger-conformance/src/Test/Cardano/Ledger/Conformance/Spec/Conway.hs
Soupstraw: The constraints used in the UTXO rule are still incomplete. GOV rule should have all the necessary constraints, but when I ran it, I got some mismatches between the spec and the implementation. I think some of it might be because I haven't properly updated all of the types after updating. The next PR will be about getting the GOV rule to pass.

libs/cardano-ledger-conformance/src/Test/Cardano/Ledger/Conformance/SpecTranslate/Conway.hs
Soupstraw: Yeah, that's fair, I was asking this because we use a `StrictMaybe` here in the implementation but not in the spec. I guess it doesn't really matter because the rules don't look at the anchors anyways.

CONTRIBUTING.md
Soupstraw: Done

libs/cardano-ledger-conformance/src/Test/Cardano/Ledger/Conformance/Spec/Conway.hs
lehins: Sounds like a good plan

libs/cardano-ledger-conformance/src/Test/Cardano/Ledger/Conformance/SpecTranslate/Conway.hs
lehins: > The spec can actually give error messages when any of the predicates are not satisfied

Aha, that is something that I did not realize. My misconception was that Agda just doesn't build if any pf the predicate checks is not satisfied. It would make sense that the Haskell representation of the spec would produce a runtime failure instead for this case.

> Once we have better types for predicate failures in the Agda spec we can get rid of this type.

Sounds like a good plan

libs/cardano-ledger-conformance/src/Test/Cardano/Ledger/Conformance/ExecutableSpecRule/Core.hs
lehins: That's what I missed. I mistakenly thought it was printing the signal, but it is in fact the spec of the signal :slightly_smiling_face:

libs/cardano-ledger-conformance/src/Test/Cardano/Ledger/Conformance/SpecTranslate/Conway.hs
lehins: That's what I suspected. Thanks

libs/cardano-ledger-conformance/src/Test/Cardano/Ledger/Conformance/ExecutableSpecRule/Core.hs
lehins: Yep, your other comments explained this too. Thanks

libs/cardano-ledger-conformance/src/Test/Cardano/Ledger/Conformance/SpecTranslate/Conway.hs
lehins: Sounds good.

lehins:
There is a missing `TypeOperators` build failure for newer GHCs. Other than that the PR is ready to be merged

libs/cardano-ledger-conformance/src/Test/Cardano/Ledger/Conformance/SpecTranslate/Conway.hs
MaximilianAlgehed: On background (and to check my understanding and recollection, please correct me if Im wrong @Soupstraw): the way the Agda side works (I paired a but with Ulf on this) is basically that there is a proof that the rule is functional (that is in the functional relation) sense of the word. If you have an environment, a state, and a signal you can uniquely determine either a failure to transition or the goal state. The bit that is certified here is that the function that computes this on the agda side is proven to be sound and complete w.r.t. the rule on the agda side.

This function and the relevant types are then compiled from Agda to Haskell and can be invoked in these tests as a Haskell function that implements the STS transition.

Soupstraw: @MaximilianAlgehed Yeah, that's how it works. The rules on the Agda side use a relational approach, but every rule has a `Computational` typeclass. That typeclass has a method that determines the output from all the inputs, and a proof that this method returns a value iff the arguments and the output are related via the relation.

libs/cardano-ledger-conformance/src/Test/Cardano/Ledger/Conformance/ExecutableSpecRule/Core.hs
Soupstraw: We'll keep it the way it is so we can do negative testing.

PR #4216, opened 2024-03-25
improve the `GOV` generator to generate more interesting signals

MaximilianAlgehed:
# Description

Improve the GOV signal generator to make sure we generate signals that can depend on any part of the proposals tree.

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/V2/Conway/GOV.hs
+++ b/libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/V2/Conway/GOV.hs
@@ -8,6 +8,8 @@
 -- for the GOV rule
 module Test.Cardano.Ledger.Constrained.V2.Conway.GOV where
 
+import Data.Foldable
+
 import Data.Coerce
 
 import Cardano.Ledger.Api
@@ -30,12 +32,6 @@ govEnvSpec ::
 govEnvSpec = constrained $ \ge ->
   match ge $ \_ _ pp _ _ _ ->
     satisfies pp pparamsSpec
-
--- TODO:
---  - check `toGovRelationTree` invariants here
---  - check `checkInvariantOnAddition` invariants here
---  - check `checkInvariantOnDeletion` invariants here
---  - check `toPTree` invariants here
 
 -- NOTE: it is probably OK not to check uniqueness of ids here, because a clash
 -- is never going to be generated, and the real representation of `Proposals` doesn't
@@ -79,9 +75,6 @@ govProposalsSpec GovEnv {..} =
                         (minV' ==. 0)
                     , minV' `dependsOn` majV'
                     ]
-          , -- TODO: get rid of this when we've fixed the performance problem around deep forall
-            -- fusion
-            isCon @"HardForkInitiation" (pProcGovAction_ . gasProposalProcedure_ $ gas)
           ]
       , forAll (snd_ hardForkTree) (genHint treeGenHint)
       , genHint listSizeHint (snd_ hardForkTree)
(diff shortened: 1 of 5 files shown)


libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/V2/Conway/GOV.hs
```diff
@@ -283,51 +276,47 @@ govProceduresSpec ge@GovEnv {..} ps =
...
     govAction
     -- ParameterChange
     ( branch $ \mPrevActionId ppUpdate policy ->
-        -- TODO: this should be any ppup proposal in the tree
-        [ assert $ mPrevActionId ==. lit (gePrevGovActionIds ^. grPParamUpdateL)
+        [ assert $ mPrevActionId `elem_` lit ppupIds
         , wfPParamsUpdate ppUpdate
         , assert $ policy ==. lit gePPolicy
         ]
     )
```
TimSheard: I am not a big fan of using field puns and then using Foo{..}, there is no way to tell where the fields are coming from (or even if they are a field. Can't we say something like Foo{ ppupIds }, so then we know it is coming from Foo?.
Every symbol should have a visible binding site. If there is no way to do this, then we could resort to
t@Foo{..}      then use (ppupIds t)

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/V2/Conway/GOV.hs
```diff
@@ -283,51 +276,47 @@ govProceduresSpec ge@GovEnv {..} ps =
...
+              reify mPrevActionId hfIdMinorVer $ \currentMinorVersion ->
+                caseOn
+                  mSuccMajorVersion
+                  ( branch $ \_ ->
+                      [ majorVersion ==. currentMajorVersion
+                      , minorVersion ==. currentMinorVersion + 1
+                      ]
+                  )
+                  ( branch $ \majorVersion' ->
+                      [ assert $ majorVersion `member_` (singleton_ currentMajorVersion <> singleton_ majorVersion')
```
TimSheard, line 314: I guess we really would like some kind of 'or' here 
majorVersion ==. currentMajorVersion ||. majorVersion ==. majorVersion'

libs/constrained-generators/src/Constrained/Spec/Generics.hs
```diff
@@ -383,6 +384,24 @@ constrained' ::
...
+  , IsProd (SimpleRep b)
+  , HasSpec fn a
+  , HasSpec fn b
+  , IsPred p fn
+  ) =>
+  Term fn a ->
+  (a -> b) ->
+  FunTy (MapList (Term fn) (Args (SimpleRep b))) p ->
+  Pred fn
+reify' a r f = reify a r $ \x -> match @fn @p x f
```
TimSheard, line 406: Maybe a comment about what the difference is between  reify and reify'

TimSheard (approved):
Every things looks reasonable.  Please, if at all possible, make binding sites visible.

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/V2/Conway/GOV.hs
MaximilianAlgehed: yeah, that or `list_ :: [Term fn a] -> Term fn [a]`

MaximilianAlgehed: Which can be built if we add `cons_ :: Term fn a -> Term fn [a] -> Term fn [a]`

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/V2/Conway/GOV.hs
MaximilianAlgehed: `ppupIds` is bound in a `where` clause at the bottom of the spec.

PR #4218, opened 2024-03-26
Remove `maxMajorPV` from `Globals`

### LINKED ISSUE #3682: The maxMajorPV field of Globals is unused
Using an up-to-date commit

```
nfrisby@frisbycomp:~/cardano-haskell/cardano-ledger$ git show origin/master --no-patch
commit 63d98c3f8e9eb2878cec3ab71c54fc40c798ac01 (origin/master, origin/HEAD)
Author: Tim Sheard <sheard@pdx.edu>
Date:   Thu Aug 24 15:45:07 2023 -0400

    Fix inactive PoolStake not counting as Drep Stake (#3676)
    
    * Changed freshDRepPulser to use map form IncrementalStake rather than SnapShot.
```

I see that the non-testing code propagates `maxMajorPV` but never actually uses it.

```
nfrisby@frisbycomp:~/cardano-haskell/cardano-ledger$ git grep -C1 -w -e maxMajorPV origin/master -- '*hs' | grep -v -e test-suite -e testlib
origin/master:eras/shelley/impl/src/Cardano/Ledger/Shelley/Genesis.hs-  Globals
origin/master:eras/shelley/impl/src/Cardano/Ledger/Shelley/Genesis.hs:mkShelleyGlobals genesis epochInfoAc maxMajorPV =
origin/master:eras/shelley/impl/src/Cardano/Ledger/Shelley/Genesis.hs-  Globals
--
origin/master:eras/shelley/impl/src/Cardano/Ledger/Shelley/Genesis.hs-    , maxLovelaceSupply = sgMaxLovelaceSupply genesis
origin/master:eras/shelley/impl/src/Cardano/Ledger/Shelley/Genesis.hs:    , maxMajorPV = maxMajorPV
origin/master:eras/shelley/impl/src/Cardano/Ledger/Shelley/Genesis.hs-    , networkId = sgNetworkId genesis
--
--
--
--
--
--
--
--
--
--
--
--
--
origin/master:libs/cardano-ledger-core/src/Cardano/Ledger/BaseTypes.hs-  -- ^ Quorum for update system votes and MIR certificates
origin/master:libs/cardano-ledger-core/src/Cardano/Ledger/BaseTypes.hs:  , maxMajorPV :: !Version
origin/master:libs/cardano-ledger-core/src/Cardano/Ledger/BaseTypes.hs-  -- ^ All blocks invalid after this protocol version
--
origin/master:libs/cardano-ledger-pretty/src/Cardano/Ledger/Pretty.hs-      , ("quorum", pretty quor)
origin/master:libs/cardano-ledger-pretty/src/Cardano/Ledger/Pretty.hs:      , ("maxMajorPV", prettyA maxmaj)
origin/master:libs/cardano-ledger-pretty/src/Cardano/Ledger/Pretty.hs-      , ("maxLovelaceSupply", pretty maxlove)
```

I therefore propose we remove it from the `Globals` data type.

Lucsanszky:
# Description
Resolves #3682 

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/libs/cardano-ledger-core/CHANGELOG.md
+++ b/libs/cardano-ledger-core/CHANGELOG.md
@@ -1,8 +1,8 @@
 # Version history for `cardano-ledger-core`
 
-## 1.14.0.1
+## 1.15.0.0
 
-*
+* Remove `maxMajorPV` from `Globals`
 
 ## 1.14.0.0
 
--- a/eras/allegra/impl/cardano-ledger-allegra.cabal
+++ b/eras/allegra/impl/cardano-ledger-allegra.cabal
@@ -67,7 +67,7 @@ library
         bytestring,
         cardano-crypto-class,
         cardano-ledger-binary ^>=1.3,
-        cardano-ledger-core ^>=1.14,
+        cardano-ledger-core ^>=1.15,
         cardano-ledger-shelley >=1.13 && <1.15,
         cardano-strict-containers,
         cardano-slotting,
--- a/eras/alonzo/impl/cardano-ledger-alonzo.cabal
+++ b/eras/alonzo/impl/cardano-ledger-alonzo.cabal
@@ -79,7 +79,7 @@ library
         cardano-ledger-allegra ^>=1.6,
         cardano-crypto-class,
         cardano-ledger-binary ^>=1.3,
-        cardano-ledger-core ^>=1.14,
+        cardano-ledger-core ^>=1.15,
         cardano-ledger-mary ^>=1.7,
         cardano-ledger-shelley >=1.13 && <1.15,
         cardano-slotting,
--- a/eras/babbage/impl/cardano-ledger-babbage.cabal
+++ b/eras/babbage/impl/cardano-ledger-babbage.cabal
@@ -75,7 +75,7 @@ library
         cardano-ledger-allegra ^>=1.6,
         cardano-ledger-alonzo >=1.9 && <1.11,
         cardano-ledger-binary ^>=1.3,
-        cardano-ledger-core ^>=1.14,
+        cardano-ledger-core ^>=1.15,
         cardano-ledger-mary ^>=1.7,
         cardano-ledger-shelley >=1.13 && <1.15,
         cardano-strict-containers,
--- a/eras/conway/impl/cardano-ledger-conway.cabal
+++ b/eras/conway/impl/cardano-ledger-conway.cabal
@@ -88,7 +88,7 @@ library
         cardano-ledger-allegra ^>=1.6,
         cardano-ledger-alonzo ^>=1.10,
         cardano-ledger-babbage ^>=1.9,
-        cardano-ledger-core ^>=1.14,
+        cardano-ledger-core ^>=1.15,
         cardano-ledger-mary ^>=1.7,
         cardano-ledger-shelley >=1.13 && <1.15,
         cardano-slotting,
--- a/eras/mary/impl/cardano-ledger-mary.cabal
+++ b/eras/mary/impl/cardano-ledger-mary.cabal
@@ -76,7 +76,7 @@ library
         cardano-data ^>=1.2,
         cardano-ledger-allegra ^>=1.6,
         cardano-ledger-binary ^>=1.3,
-        cardano-ledger-core ^>=1.14,
+        cardano-ledger-core ^>=1.15,
         cardano-ledger-shelley >=1.13 && <1.15,
         containers,
         deepseq,
--- a/eras/shelley/impl/CHANGELOG.md
+++ b/eras/shelley/impl/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## 1.14.0.0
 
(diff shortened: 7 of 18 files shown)


Lucsanszky:
@nfrisby if I understand correctly, this should not be merged until some of the concerns you mention [here](https://github.com/IntersectMBO/ouroboros-consensus/issues/324#issuecomment-1695928343) are addressed, right? Especially the obsolete node checks part as this PR would remove the protocol version check from `chainChecks` (which would be okay I guess if we decide to ban intra-era hard forks at some point).

eras/shelley/test-suite/src/Test/Cardano/Ledger/Shelley/Rules/Chain.hs
```diff
@@ -340,8 +336,7 @@ chainTransition =
             chainChecksData = pparamsToChainChecksPParams pp
             bhView = makeHeaderView bh
 
-        maxpv <- liftSTS $ asks maxMajorPV
-        case chainChecks maxpv chainChecksData bhView of
+        case chainChecks chainChecksData bhView of
```
lehins:
```suggestion
        case chainChecks (succ (eraProtVerHigh @era)) chainChecksData bhView of
```

eras/shelley/impl/src/Cardano/Ledger/Chain.hs
```diff
@@ -62,17 +62,13 @@ instance NoThunks ChainPredicateFailure
 
 chainChecks ::
   MonadError ChainPredicateFailure m =>
-  Version ->
   ChainChecksPParams ->
   BHeaderView c ->
   m ()
-chainChecks maxpv ccd bhv = do
```
lehins: Could you please revert changes to this function and all other changes that depended on it

lehins (changes requested):
@Lucsanszky Feedback from consensus folks:

> The idea would be for the ledger PR to remove `maxMajorPV` from `Globals`, but not from `chainChecks`

The only reason why you had to remove `Version` from `chainChecks` is because we used `maxMajorPV` from `Globals` for `chainChecks` in shelley test.

eras/shelley/test-suite/src/Test/Cardano/Ledger/Shelley/Rules/Chain.hs
Lucsanszky: @lehins, it seems like `succ (eraProtVerHigh @era)` evaluates to `3` because the `@era` is `Shelley` and this causes test failures. `pvMajor $ pp ^. ppProtocolVersionL` fixes it but I don't know if this discrepancy between `succ (eraProtVerHigh @era)` and `pvMajor $ pp^. ppProtocolVersionL` is okay / expected, so I thought I would point it out.

lehins: In the matter of fact looking closer at the meaning of the `chainChecks` function, makes me realize it should be `eraProtVerHigh @era`. 

The reason why tests started failing is because generators for the tests are incorrect. With this change you'd get errors like this:

```
ObsoleteNodeCHAIN (Version 9) (Version 2)
```

Which means generators are trying to use protocol version `9` in Shelley, which makes no sense whatsoever.

I am gonna try and fix generators and see if we can have it done properly.

The reason why `pvMajor $ pp ^. ppProtocolVersionL` works, is because then `chainChecks` always compares protocol version to itself, which is sort of useless

Lucsanszky: @lehins ah, I see. I was afraid that this will be the case, that we are asserting wrong behaviour. In the meantime, I'll go back to my other issues but if your hands are full, I'm happy to look at those generators and have a go at fixing this properly.

lehins: > I'm happy to look at those generators and have a go at fixing this properly.

No worries, I had it half way done last week, just didn't have the time to finish it. Here is the PR that you can use to get this one unblocked: https://github.com/IntersectMBO/cardano-ledger/pull/4576

libs/cardano-ledger-core/CHANGELOG.md
```diff
@@ -1,8 +1,8 @@
 # Version history for `cardano-ledger-core`
 
-## 1.14.0.1
+## 1.15.0.0
```
lehins, line 3: Version also need to bumped in the cabal file :wink: which will result in a whole lot of cabal files needing an update too :slightly_smiling_face:

lehins (approved):
Other than `cardano-ledger-core` version this PR is good to go.

libs/cardano-ledger-core/CHANGELOG.md
Lucsanszky: Thanks! However, there is still a conformance test failure. ~It's a flaky test but it seems that it happens quite often so I'll investigate a bit more.~ Nevermind, it seems like this will fix it: https://github.com/IntersectMBO/cardano-ledger/pull/4585

Issue #4192, opened 2024-03-11
Avoid converting to and from compact representation for multi-assets.


dnadales:
Storing the ledger state on disk requires converting from the compact representation of multi-assets, which consumes a significant amount of memory, and raises garbage collection activity significantly. Consequently, [slot leadership checks are missed](https://github.com/IntersectMBO/ouroboros-consensus/issues/868).

TimSheard:
Digging around trying to understand why this might be expensive. So the most expensive thing to Serialize is the UTxO.
The UTxO has TxOut, and TxOut  have Value, and Value might have MultiAsset.  Here is how the Alonzo TxOut is serialized.
```

instance
  (Era era, Val (Value era)) =>
  EncCBOR (AlonzoTxOut era)
  where
  encCBOR (TxOutCompact addr cv) =
    encodeListLen 2
      <> encCBOR addr
      <> encCBOR cv
  encCBOR (TxOutCompactDH addr cv dh) =
    encodeListLen 3
      <> encCBOR addr
      <> encCBOR cv
      <> encCBOR dh
```
We will concentrate on the pattern `TxOutCompact ` (the pattern `TxOutCompactDH ` is similar)
```
pattern TxOutCompact ::
  (Era era, Val (Value era), HasCallStack) =>
  CompactAddr (EraCrypto era) ->
  CompactForm (Value era) ->
  AlonzoTxOut era
pattern TxOutCompact addr vl <-
  (viewCompactTxOut -> (addr, vl, SNothing))
  where
    TxOutCompact cAddr cVal = mkTxOutCompact (decompactAddr cAddr) cAddr cVal SNothing
```

Note how the `cVal` has type `(CompactForm Value)`. How is that serialized? 

```
instance Crypto c => Compactible (MaryValue c) where
  newtype CompactForm (MaryValue c) = CompactValue (CompactValue c)
    deriving (Eq, Typeable, Show, NoThunks, **EncCBOR**, DecCBOR, NFData)
  toCompact x = CompactValue <$> to x
  fromCompact (CompactValue x) = from x
```

So the EncCBOR is derived, so that means it uses the instance (EncCBOR CompactValue) which is this

```
instance Crypto c => EncCBOR (CompactValue c) where
  encCBOR = encCBOR . from
```
So the `from`  translates into a Map of Map and that is serialized.  We could probably just use the bytes inside CompactValue instead, But that would break backward compatibility, and we couldn't synch the chain.

[...]

Issue #3486, opened 2023-06-16
Ensure Conway additions to ledger state use sharing


lehins:
There were a few additions to the NewEpochState due to Conway. This ticket tracks proper `DecShareCBOR` instances for all newly added fields and types in the NewEpochState. A full list should be added to this ticket.

Issue #3220, opened 2022-12-22
New versioned serialization API for data types that don't live on chain.


lehins:
Recently in #3063 and #3138 we introduced versioned serialization mechanism that is necessary for data types which live on chain. Here are the reasons why that versioning scheme is perfect for types that get placed on chain (eg. `TxBody` , `TxAuxData`, etc.):
* It is only natural to switch serialization changes together with all other changes at a hard fork boundary, therefore major protocol version, which is bumped at that boundary, can be used perfectly for versioning serialization
* Deserialization of every type must be backwards compatible at least with the previous era, so there is no need for translation logic, from older version to the newer.
* Every type that lives on chain is parameterized on `era`, therefore it is always possible to get the protocol version for deserialization if necessary. This, however, comes with a caviat that we can't introduce any serialization changes for intra-era hard forks, because as of now we don't know the exact protocol version a the type level, we only know the range.

The other end of the spectrum are the types that do not live on chain, for example most of the contents of the ledger state (i.e. `NewEpochState`) is only stored locally on each node. Such type require a more classical serialization versioning scheme. In such scheme the version that the type was serialized with is encoded in the binary blob together with the data itself, usually in some sort of header. We want to use such scheme because we want to be able to upgrade those local types at a node software version bump, rather than at a hardfork boundary. Due to that fact protocol version is not a suitable candidate. So we need a versioning scheme that:

* Encode a version together with the data. It should probably not be done at the type level (i.e. every individual type is prefixed with a version), because that would introduce quite a bit overhead and is just an overkill.)
* Every type would provide an upgrade logic for going from the previous version of a type to the newer one, whenever there was a breaking change to deserialization.
* Deserialization interface should have native support for improving memory sharing. There is already an implementation of such interface in `FromSharedCBOR`.

lehins:
CC @nc6 @qnikst

PR #4222, opened 2024-03-28
Ts sharing coders

TimSheard:
Extends the Coders library to handle Sharing on the decode side.
Adds some DecShareCBOR instances as described in issue  #3486 using the new tools

Began to address the problems described in issue #4192, about serializing CompactForm MultiAsset, without affecting backward compatibility.

A large part of the changes have to do with the following changes
1.  replacing Ann with Pure
2. replacing (<!*) with (<!>)
3.
The original operators Ann and (<!*) had to do the the Annotator type, but the ones have been generalized to work with any Applicative functor. Including the monad  `(StateT (Share b) (Decoder s) b) ` which is the basis for Sharing.

# Checklist

- [ ] Self-reviewed the diff


lehins:
I am closing this PR as we will not be using Coders with Sharing. The plan is to create a different interface for ledger state serialization. See #3220

PR #4229, opened 2024-04-02
Shrinking for `constrained-generators`

MaximilianAlgehed:
# Description

This implements shrinking for `constrained-generators` by embedding it in the `HasSpec` class. There are still some TODOs that need to be covered but this should be relatively ready to look at.

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/libs/constrained-generators/src/Constrained/Base.hs
+++ b/libs/constrained-generators/src/Constrained/Base.hs
@@ -620,6 +620,9 @@ class
   -- | Check conformance to the spec.
   conformsTo :: HasCallStack => a -> TypeSpec fn a -> Bool
 
+  -- | Shrink an `a` with the aide of a `TypeSpec`
+  shrinkWithTypeSpec :: TypeSpec fn a -> a -> [a]
+
   -- | Convert a spec to predicates:
   -- The key property here is:
   --   ∀ a. a `conformsTo` spec == a `conformsTo` constrained (\t -> toPreds t spec)
@@ -714,6 +717,16 @@ class
     Pred fn
   toPreds v s = toPreds (toGeneric_ v) s
 
+  default shrinkWithTypeSpec ::
+    ( HasSpec fn (SimpleRep a)
+    , TypeSpec fn a ~ TypeSpec fn (SimpleRep a)
+    , HasSimpleRep a
+    ) =>
+    TypeSpec fn a ->
+    a ->
+    [a]
+  shrinkWithTypeSpec spec a = map fromSimpleRep $ shrinkWithTypeSpec @fn spec (toSimpleRep a)
+
 data WithHasSpec fn f a where
   WithHasSpec :: HasSpec fn a => f a -> WithHasSpec fn f a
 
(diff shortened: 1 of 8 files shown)


MaximilianAlgehed:
It might be interesting for @Soupstraw to also have a look at this. Not required but spreading knowledge around isn't a bad thing.

libs/constrained-generators/src/Constrained/Base.hs
```diff
@@ -2570,6 +2603,8 @@ instance (Ord a, HasSpec fn a) => HasSpec fn (Set a) where
               genFromSpec elemS `suchThatT` (`Set.notMember` s)
         go (n - 1) (Set.insert e s)
 
+  shrinkWithTypeSpec (SetSpec _ es _) as = map Set.fromList $ shrinkList (shrinkWithSpec es) (Set.toList as)
```
Soupstraw, line 2606: Wouldn't it make sense to also shrink the size spec here?

Soupstraw: maybe by pushing the upper bound downwards a bit

MaximilianAlgehed: The idea here is not to shrink the spec, only the values.

libs/constrained-generators/src/Constrained/Base.hs
```diff
@@ -2849,6 +2887,10 @@ genFromNumSpec (NumSpecInterval ml mu) = do
   n <- sizeT
   pureGen . choose =<< constrainInterval (ml <|> lowerBound) (mu <|> upperBound) (fromIntegral n)
 
+-- TODO: fixme (?)
+shrinkWithNumSpec :: Arbitrary n => NumSpec n -> n -> [n]
+shrinkWithNumSpec _ = shrink
```
Soupstraw, line 2892: I think this might generate values that do not conform to the spec

MaximilianAlgehed: It will, but thats also not a requirement as we filter on the conformance at the use site. But that's something we should change

Soupstraw (approved):
LGTM

libs/constrained-generators/src/Constrained/Base.hs
```diff
@@ -800,6 +813,21 @@ genFromSpec ts@(TypeSpec s cant) =
     genFromTypeSpec @fn s `suchThatT` (`notElem` cant)
 genFromSpec (ErrorSpec e) = genError e
 
+shrinkWithSpec :: forall fn a. HasSpec fn a => Spec fn a -> a -> [a]
+-- TODO: possibly allow for ignoring the `conformsToSpec` check in the `TypeSpec`
+-- case when you know what you're doing
+shrinkWithSpec spec a = filter (`conformsToSpec` spec) $ case spec of
+  -- TODO: filter on can't if we have a known to be sound shrinker
```
TimSheard, line 820: How does that work? Have another method 
howToConform :: a -> bool
with default implementation (`conformsToSpec` spec)
and allow some types to use (const True)?

libs/constrained-generators/src/Constrained/Test.hs
```diff
@@ -102,6 +102,18 @@ prop_complete s =
...
+  QC.forAll (strictGen $ genFromSpec s) $ \ma -> fromGEDiscard $ do
+    a <- ma
+    let shrinks = shrinkWithSpec s a
+    pure $
+      cover 40 (not $ null shrinks) "non-null shrinks" $
+        if null shrinks
+          then property True
+          else QC.forAll (QC.elements shrinks) $ \a' ->
+            conformsToSpecProp a' s
+
```
TimSheard, line 116: Is there where we add the fact that we must filter by conformsTo.  That works for this Property, but where do we do this for any property that uses shrinking?

libs/constrained-generators/src/Constrained/Test.hs
```diff
@@ -125,45 +137,53 @@ tests =
     , testSpec "maybeSpec" maybeSpec
     , testSpec "eitherSetSpec" eitherSetSpec
     , testSpec "fooSpec" fooSpec
-    , testSpec "intSpec" intSpec
+    , -- TODO: this spec needs double shrinking to shrink properly
+      -- so we need to figure something out about double-shrinking
+      testSpecNoShrink "intSpec" intSpec
```
TimSheard, line 142: What is double shrinking?

TimSheard (approved):
This PR adds a new method to HasSpec, supplies the method for all the instances, and then adds some tests that the shrinking does not affect the results.

Just a few questions to answer. May be add some more comments.

libs/constrained-generators/src/Constrained/Base.hs
MaximilianAlgehed: I think I would go with something like `unsafeShrinkWithTypeSpec` that has a default definition that does the filtering. But the method you propose does have the benefit that it doesn't introduce two different callbacks where one obsolete the other (the unsafe version would have a default implementation).

libs/constrained-generators/src/Constrained/Test.hs
MaximilianAlgehed: I'm not sure I understand what you mean?

libs/constrained-generators/src/Constrained/Test.hs
MaximilianAlgehed: Think about it like this, you have a bug that only materialises when the function under test is given (x, y) such that x == y. If you try to shrink that you won't get far because shrinking of pairs works one component at a time - the property will fail for some (x, x) but shrink to (x', x) (to avoid an exploaion of shrinks paira tyoically shrink one component at a time). With double shrinking you'll try to shrink twice before concluding you're in a local minimum, so you'll try things like (x'', x) and crucially (x', x').

PR #4234, opened 2024-04-03
Add separate `devShell` for `pre-commit`

### LINKED ISSUE #4231: Remove precommit hook
Approach of running fourmolu that was introduced in #4175 using a pre commit hook has some serious downsides. Namely, initial invocation of nix shell will force a user to rely on nix, even if the repo (`.git` folder) migrates to a non-nix computer, because the `pre-commit` will try to run binaries from the nix store.

In other words nix shell would affect users outside of nix, which is something we need to avoid.

Suggested solution is to provide a script that sets up such pre-commit functionality can be executed by any user whenever they want fourmolu to run on every commit.

Lucsanszky:
# Description
Resolves #4231 

# Checklist

- [x] Self-reviewed the diff


flake.nix
```diff
@@ -103,7 +96,6 @@
               fi
             }
             PROMPT_COMMAND=prompt
-            ${pre-commit-check.shellHook}
```
teodanciu, line 106: What about [this](https://github.com/IntersectMBO/cardano-ledger/pull/4234/files#diff-206b9ce276ab5971a2489d75eb1b12999d4bf3843b7988cbe8d687cfde61dea0L33): 
```
        pre-commit-check = inputs.pre-commit-hooks.lib.${system}.run {
          src = ./.;
          hooks = {
            fourmolu.enable = true;
          };
        };
```

Can we get rid of it as well?  Or is it necessary to have there for the alternative dev shell?

teodanciu (approved):
Looks good to me, just a question about a snippet that I'm not sure whether is useful to keep or is part of the previous solution.
I think this solution is elegant and simpler, and this is witnessed by the simplifications in the instructions on enabling/disabling it.

flake.nix
Lucsanszky: That one is still needed. Strictly speaking however, it is not required at that scope. The reason why it's there is because in #4175 I also added a check (see line ~244) that can be run and added the possibility to run `fourmolu` as a hydra job if we wish to. I can remove these if you want to and move the above `pre-commit-check` binding to `mkDevShells`.

Lucsanszky: I added the changes in f9d1a3f. Let me know if that is what you would prefer.

teodanciu: Yes! I think it's better to limit the scope like you did, thank you!

lehins (approved):
Much better :slightly_smiling_face: 
Thank you

flake.nix
```diff
@@ -249,20 +239,30 @@
...
+              pre-commit = let
+                pre-commit-check = inputs.pre-commit-hooks.lib.${system}.run {
+                  src = ./.;
+                  hooks = {
+                    fourmolu.enable = true;
+                  };
+                };
+              in
+                p.shell // (nixpkgs.mkShell {
+                shellHook = p.shell.shellHook + pre-commit-check.shellHook;
```
neilmayhew: I don't think this does what you want. It seems to leave out the Haskell stuff and have just `pre-commit` in it.

I think you need:

```suggestion
                p.shell.overrideAttrs (old: {
                  shellHook = old.shellHook + pre-commit-check.shellHook;
```

neilmayhew (changes requested):

flake.nix
```diff
@@ -249,20 +239,30 @@
...
-            profillingShell = p: {
+            mkDevShells = p: {
               # `nix develop .#profiling` (or `.#ghc928.profiling): a shell with profiling enabled
               profiling = (p.appendModule {modules = [{enableLibraryProfiling = true;}];}).shell;
+              # `nix develop .#pre-commit` (or `.#ghc928.pre-commit): a shell with pre-commit enabled
+              pre-commit = let
+                pre-commit-check = inputs.pre-commit-hooks.lib.${system}.run {
+                  src = ./.;
+                  hooks = {
+                    fourmolu.enable = true;
```
neilmayhew, line 251: I'm wondering about including other tools in this list, such as `cabal-fmt` and `nixpkgs-fmt`. However, I think that should be a follow-up PR.

Lucsanszky: I was thinking about this as well and I agree that this should happen in a separate PR if we decide to add those.

neilmayhew (approved):

PR #4237, opened 2024-04-03
Remove duplicate certificates from conway.cddl

SebastienGllmt:
# Description

Previously, if you wanted to do multiple things at once, you had to submit multiple certificates as part of the transaction

However, in Conway, there is now a combinatorial explosion of certificates that do multiple things at the same time (they're just combinations of `stake_delegation`, `reg_cert`, `unreg_cert`, `vote_deleg_cert`

I feel this is wrong because:
1. It increases protocol complexity despite this being easily handled on the wallet layer
2. This doesn't scale if we add more certificates because every new certificate requires more combinations
3. Trezor, Ledger, Eternl and dcSpark have already said they will not support this feature. This is on-track to be another "pointer address" which is a feature designed to save a few bytes that nobody wants to support and just ends up making everything more complex for no reason

Feel free to let me know if you think differently, but otherwise I propose we just remove these

### DIFF
--- a/eras/conway/impl/cddl-files/conway.cddl
+++ b/eras/conway/impl/cddl-files/conway.cddl
@@ -307,10 +307,6 @@ pool_retirement = (4, pool_keyhash, epoch)
 reg_cert = (7, stake_credential, coin)
 unreg_cert = (8, stake_credential, coin)
 vote_deleg_cert = (9, stake_credential, drep)
-stake_vote_deleg_cert = (10, stake_credential, pool_keyhash, drep)
-stake_reg_deleg_cert = (11, stake_credential, pool_keyhash, coin)
-vote_reg_deleg_cert = (12, stake_credential, drep, coin)
-stake_vote_reg_deleg_cert = (13, stake_credential, pool_keyhash, drep, coin)
 
 ; GOVCERT
 auth_committee_hot_cert = (14, committee_cold_credential, committee_hot_credential)


lehins:
:smile: Removing CDDL alone is only going to break things.

Disclaimer: I personally don't care about this certificate, cause as you said, its functionality can be fully reproduced as a combination of couple of other certificates.

Here is my response of why this change of removal is not going to happen:

1. This is too close to Conway completion for us to change CDDL, implementation of Ledger, Plutus, CLI and Agda specification to make a drastic change like that at this point.
2. Trezor, Ledger, Eternl and dcSpark have not said anything to us about not supporting this certificate, despite that they knew about these certificate type long time ago. This certificate in the CDDL was published almost a year ago: 7f42d20b2ba07bac5e222dfddc856b16c307d62a.
3. If they don't want to support it, that's fine. It does not mean that the protocol cannot not support it. Hardware wallets do not support many other features that protocol provides, like voting for example. It is acceptable for them to only support a subset of the features that their users care about.

So, to sum it up. At this point it doesn't matter if it is a good idea or not to have this feature. If there will be strong support in the community on removing this certificate which can do registration and delegation all at the same time, then we will be able to remove it in the next era. That would now require a CIP and a proper discussion, of course.

If I remember correctly this type of certificate was introduced as an optimization that would reduce the size of the transaction by a tiny bit, thus could save a tiny bit on fees. Was it a right decision, I honestly don't know, time will tell.

KtorZ:
To be fair I brought up the point 8 months ago already with the exact same points that @SebastienGllmt is now making. For the record, this was the question back then: "Why the compound type? Wouldn't it be better to keep the API orthogonal and let users submit two certificates for each situation?" which led to a few questions and answers:

> Why would it be better?

Because it's orthogonal. So there's only one way to do a certain operation and ways to compose things together.

>  It is worse for efficiency: transaction become bigger the more certificates you add to it.

I don't think the case is that common with respect to other on-chain operation to warrant such an optimization? is it? Especially when a vast majority of the existing network has already registered their keys / credentials. Plus, how much bytes are we talking here? This saves 2 bytes per double registrations -- so that's 58 bytes serialized instead of 60 bytes so that's a net gain of -3% :thinking_face:. I don't think the gain in computation / processing are that much significant either?All I am saying is that, I don't quite see the benefits and it seems to be adding some extra complexity that doesn't seem justified. Certificates are already complicated enough and they get an order of magnitude more complicated in Conway so I thought keeping things simple when possible would be best.

> the approach you are suggesting would also allow the user to completely undelegate, but not unregister (thus not return the deposit), which is the state we want to discourage.

I am only suggesting to rely on the current approach which doesn't support un-delegation, but simply de-registration (I think specifically for that reason). As far as I can tell, the new design doesn't really change that as there's still a registration cert, de-registration cert and deleg cert (as per cardano-ledger-conway=1.6.3.0); The only oddity is this compound certificate which doesn't sound like a useful optimization to me as it doesn't solve any problem?

lehins:
@KtorZ Please, go back to the conversation you pulled those quotes from. All of those responses where to your original question:

> What's the intent of the DelegStakeVote !(KeyHash 'StakePool c) !(Credential 'Voting c) Delegatee on Conway? Is it simply a way to batch together a stake delegation and vote delegation?
> Why the compound type? Wouldn't it be better to keep the API orthogonal and let users submit two certificates for each situation?

Which was initially an unclear question about a totally different certificate. Once we actually clarified that you were asking about `ConwayRegDelegCert`, this was my response:

> Oh, so your question is about ConwayRegDelegCert?
> I didn't come up with it. It was designed in this way in the spec and it didn't seem like a bad idea to me.
> There was a desire expressed in ability to register and delegate in one certificate, so that's what we have

I'd appreciate you did not put incorrect feedback on public channels and if you do, please put the full story.

With all that aside, you did state that you found that certificate redundant. And I can agree with both of you @KtorZ and  @SebastienGllmt on that, but unless you want to postpone a release of Conway for another few months, we can't remove it.
FYI, it would be easy for us to remove it in the ledger codebase, so that is absolutely not the reason why I am pushing back on this. We were explicitly asked to not make any changes to the CDDL because HWW depend on its stability and no drastic changes to the spec, since it is going through an audit. Not to mention all the integration with api/cli and plutus.

Again, if you feel strongly about that certificate being removed, then please either talk to product people that are higher than me in the chain of command or submit a CIP for the next era.

PR #4238, opened 2024-04-04
Imptests: ParameterChange affects ratification for in-flight proposals

### LINKED ISSUE #4132: Imp test that ParameterChange affects proposals that follow
Ensure that changing parameters correctly affects other proposals that haven't been ratified in the same epoch:
* Increasing a DRep/StakePool threshold prevents a proposal that had enough votes before parameter change enactment can no longer be ratified
* Decreasing a DRep/StakePool threshold results in enactment of a proposal that did not have enough votes before parameter change enactment

aniketd:
# Description

Resolves #4132 

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/RatifySpec.hs
+++ b/eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/RatifySpec.hs
@@ -1,4 +1,5 @@
 {-# LANGUAGE DataKinds #-}
+{-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE NumericUnderscores #-}
 {-# LANGUAGE OverloadedLists #-}
 {-# LANGUAGE RankNTypes #-}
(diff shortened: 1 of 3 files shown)


eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/RatifySpec.hs
```diff
@@ -35,6 +35,117 @@ spec =
...
+          submitTwoExampleProposalsAndVoteOnTheChild gpiCC drepKH = do
+            committeeC <- KeyHashObj <$> freshKeyHash
+            let updateCC parent = UpdateCommittee parent mempty (Map.singleton committeeC $ EpochNo 5) $ 1 %! 2
+            gai' <- submitGovAction $ updateCC $ SJust gpiCC
+            -- We submit a descendent proposal so that even though it is sufficiently
+            -- voted on, it cannot be ratified before the ParameterChange proposal
+            -- is enacted.
+            gai <- submitGovAction $ updateCC $ SJust $ GovPurposeId gai'
+            submitYesVote_ (DRepVoter $ KeyHashObj drepKH) gai
+            passEpoch -- Make the votes count do
```
Soupstraw:
```suggestion
            passEpoch -- Make the votes count
```

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/RatifySpec.hs
```diff
@@ -35,6 +35,117 @@ spec =
...
+            passNEpochs 2
+          submitTwoExampleProposalsAndVoteOnTheChild gpiCC drepKH = do
+            committeeC <- KeyHashObj <$> freshKeyHash
+            let updateCC parent = UpdateCommittee parent mempty (Map.singleton committeeC $ EpochNo 5) $ 1 %! 2
+            gai' <- submitGovAction $ updateCC $ SJust gpiCC
+            -- We submit a descendent proposal so that even though it is sufficiently
+            -- voted on, it cannot be ratified before the ParameterChange proposal
+            -- is enacted.
+            gai <- submitGovAction $ updateCC $ SJust $ GovPurposeId gai'
+            submitYesVote_ (DRepVoter $ KeyHashObj drepKH) gai
```
Soupstraw:
```suggestion
            gaiParent <- submitGovAction $ updateCC $ SJust gpiCC
            -- We submit a descendent proposal so that even though it is sufficiently
            -- voted on, it cannot be ratified before the ParameterChange proposal
            -- is enacted.
            gaiChild <- submitGovAction $ updateCC $ SJust $ GovPurposeId gaiParent
            submitYesVote_ (DRepVoter $ KeyHashObj drepKH) gaiChild
```

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/RatifySpec.hs
```diff
@@ -35,6 +35,117 @@ spec =
...
+            passEpoch -- Make the votes count do
+            pure gai
```
Soupstraw:
```suggestion
            pure gaiChild
```

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/RatifySpec.hs
```diff
@@ -35,6 +35,117 @@ spec =
...
+      it "Increasing the threshold prevents a hitherto-ratifiable proposal from being ratified" $ do
+        (drepC, hotCommitteeC, gpiCC) <- electBasicCommittee
+        setThreshold smallerThreshold
+        (drepKH, _stakingKH, _paymentKP) <- setupSingleDRep 1_000_000
+        gai <- submitTwoExampleProposalsAndVoteOnTheChild gpiCC drepKH
+        isDRepAccepted gai `shouldReturn` True
+        enactThreshold largerThreshold drepC hotCommitteeC
+        isDRepAccepted gai `shouldReturn` False
```
Soupstraw: Does the parent govaction get enacted here? I think it'd be a good idea to assert that here also if that's the case.

aniketd: The parent-child pair of actions do not get enacted, but remain in-flight, because the parent is never voted on. This had to be done because we want to enact `ParameterChange` in between the checks. But the parent-child pair of actions can be made to be enacted finally, after the intermediate `ParameterChange` has been enacted if that will be a good test, but then this check cannot be performed before the `ParameterChange` since,

1. either the proposal has enough votes, and then a parameter-change caused it to not be enacted - in which case there has to be another condition due to which it didn't already get enacted before the parameter-change itself in that epoch
2. or, the proposal didn't have enough votes, and then a parameter-change just caused it to be enacted in the very next epoch.

Soupstraw: I think it'd be a good idea to also write a test to check what happens when the parent is a `ParameterChange` and gets enacted, but that makes the child proposal not accepted because the parent sets the voting thresholds to a higher value.

aniketd: :100: :1st_place_medal:

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/RatifySpec.hs
```diff
@@ -35,6 +36,140 @@ spec =
...
+      it "Decreasing the threshold ratifies a hitherto-unratifiable proposal" $ do
+        (drepC, hotCommitteeC, gpiCC) <- electBasicCommittee
+        setThreshold largerThreshold
+        (drepKH, _stakingKH, _paymentKP) <- setupSingleDRep 1_000_000
+        gai <- submitTwoExampleProposalsAndVoteOnTheChild gpiCC drepKH
+        isDRepAccepted gai `shouldReturn` False
+        enactThreshold smallerThreshold drepC hotCommitteeC
+        isDRepAccepted gai `shouldReturn` True
```
lehins: Unlike previous test, here we can check that `gai` does get enacted

lehins (approved):
LGTM

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/RatifySpec.hs
lehins: I guess it would require votes from CC members and SPOs, so maybe it would be an overkill for the test, but it might be worth while. I'll leave the decision up to you

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/RatifySpec.hs
aniketd: I have added this test :raised_hands: Awesome suggestion @Soupstraw :rocket:

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/RatifySpec.hs
aniketd: It was easier than I thought :+1: I have made both the tests that decrease the threshold to enact the proposals :raised_hands: :smiling_imp:

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/RatifySpec.hs
```diff
@@ -35,6 +36,191 @@ spec =
...
+      submitYesVote_ (DRepVoter drepC) childGai
+      submitYesVote_ (CommitteeVoter hotCommitteeC) childGai
+      passEpoch
+      logRatificationChecks parentGai
+      logRatificationChecks childGai
+      isDRepAccepted parentGai `shouldReturn` True
+      isDRepAccepted childGai `shouldReturn` True
+      passEpoch
+      getLastEnactedParameterChange `shouldReturn` SJust (GovPurposeId parentGai)
+      Map.member (GovPurposeId childGai) <$> getParameterChangeProposals `shouldReturn` True
```
Soupstraw, line 223:
```suggestion
      Map.member (GovPurposeId childGai) <$> getParameterChangeProposals `shouldReturn` True
      isDRepAccepted childGai `shouldReturn` False
```

Soupstraw (approved):
LGTM

PR #4241, opened 2024-04-05
Add fixup combinators to ImpTest framework

### LINKED ISSUE #4209: Add a `postFixupHook` function to ImpTest
We should add a function that lets you modify the transaction after fixup. For this we should add a function `postFixupHook :: (Tx era -> Tx era) -> ImpTestM era a -> ImpTestM era a` that uses the `local` function to pass a function into the environment of `ImpTestM` that will be applied after the fixup function. This will make it easier to write negative tests, where you actually want a transaction to fail for some specific reason.

teodanciu:
# Description

Resolves https://github.com/IntersectMBO/cardano-ledger/issues/4209

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/eras/shelley/impl/testlib/Test/Cardano/Ledger/Shelley/ImpTest.hs
+++ b/eras/shelley/impl/testlib/Test/Cardano/Ledger/Shelley/ImpTest.hs
@@ -91,7 +91,11 @@ module Test.Cardano.Ledger.Shelley.ImpTest (
   logFeeMismatch,
 
   -- * Combinators
+  withCustomFixup,
+  withFixup,
   withNoFixup,
+  withPostFixup,
+  withPreFixup,
   -- We only export getters, because internal state should not be accessed during testing
   impNESG,
   impLastTickG,
@@ -559,13 +563,11 @@ impWitsVKeyNeeded txBody = do
 
 data ImpTestEnv era = ImpTestEnv
   { iteState :: !(IORef (ImpTestState era))
-  , iteDoTxFixup :: !Bool
-  -- ^ This flag is used to toggle the fixing up of transactions. If it
-  -- is set to False then any transaction should be submitted as-is.
+  , iteFixup :: Tx era -> ImpTestM era (Tx era)
   }
 
-iteDoTxFixupL :: Lens' (ImpTestEnv era) Bool
-iteDoTxFixupL = lens iteDoTxFixup (\x y -> x {iteDoTxFixup = y})
+iteFixupL :: Lens' (ImpTestEnv era) (Tx era -> ImpTestM era (Tx era))
+iteFixupL = lens iteFixup (\x y -> x {iteFixup = y})
 
 newtype ImpTestM era a = ImpTestM {unImpTestM :: ReaderT (ImpTestEnv era) (GenT IO) a}
   deriving
@@ -599,13 +601,13 @@ instance MonadState (ImpTestState era) (ImpTestM era) where
   put x = ImpTestM $ do
     liftIO . flip writeIORef x . iteState =<< ask
 
-instance Example (ImpTestM era ()) where
+instance ShelleyEraImp era => Example (ImpTestM era ()) where
   type Arg (ImpTestM era ()) = ImpTestState era
 
   evaluateExample impTest params =
     evaluateExample (\s -> evalImpTestM (getParamsQCGen params) s impTest) params
 
-instance (Arbitrary a, Show a) => Example (a -> ImpTestM era ()) where
+instance (ShelleyEraImp era, Arbitrary a, Show a) => Example (a -> ImpTestM era ()) where
   type Arg (a -> ImpTestM era ()) = ImpTestState era
 
   evaluateExample impTest params =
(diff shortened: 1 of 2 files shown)


eras/shelley/impl/testlib/Test/Cardano/Ledger/Shelley/ImpTest.hs
```diff
@@ -1311,6 +1322,9 @@ registerPool = do
 withNoFixup :: ImpTestM era a -> ImpTestM era a
 withNoFixup = local $ iteDoTxFixupL .~ False
 
+withPostFixup :: (Tx era -> ImpTestM era (Tx era)) -> ImpTestM era a -> ImpTestM era a
```
teodanciu: Not too happy with the name, because it's in essence also fixup, but it's "manual" rather than "automatic" as opposed to `fixupTx`

Soupstraw: Would `withPostFixupHook` be better?

eras/shelley/impl/testlib/Test/Cardano/Ledger/Shelley/ImpTest.hs
```diff
@@ -913,6 +915,15 @@ trySubmitTx tx = do
...
     pure txFixed
+  where
+    fixedUp = do
+      doFixup <- asks iteDoTxFixup
+      let fx =
+            if doFixup
+              then fixupTx tx
+              else pure tx
+      postFixup <- asks itePostFixup
+      fx >>= postFixup
```
Soupstraw: I think it'd be better to just replace `fixedUp` above with the code here, since it's used only once.

Soupstraw (approved):
Awesome, even better than what I had in mind!

eras/shelley/impl/testlib/Test/Cardano/Ledger/Shelley/ImpTest.hs
teodanciu: Ok, I was thinking maybe it can make the main thing easier to read, by reducing all the fixing up to one invocation, but on the other hand of course, it makes you have to jump to see what's going on. 
Happy to remove the binding, I don't have any strong opinion on it :+1:

eras/shelley/impl/testlib/Test/Cardano/Ledger/Shelley/ImpTest.hs
teodanciu: Not sure if "hook" makes it any better.. I think between "withPostFixup" and "withPostFixupHook" I prefer the former, because we haven't used the "hook" terminology so far, even though I guess "withNoFixup" is also a hook?  So it feels somehow inconsistent.

eras/shelley/impl/testlib/Test/Cardano/Ledger/Shelley/ImpTest.hs
```diff
@@ -562,11 +563,15 @@ data ImpTestEnv era = ImpTestEnv
   , iteDoTxFixup :: !Bool
   -- ^ This flag is used to toggle the fixing up of transactions. If it
   -- is set to False then any transaction should be submitted as-is.
+  , itePostFixup :: Tx era -> (ImpTestM era (Tx era))
```
```suggestion
  , itePostFixup :: Tx era -> ImpTestM era (Tx era)
```

eras/shelley/impl/testlib/Test/Cardano/Ledger/Shelley/ImpTest.hs
lehins: I concur with "hook"less terminology. It is uncommon in Haskell.

lehins (changes requested):
This looks great! I have even a follow up idea how we can make it even better:

Let's get rid of `iteDoTxFixup` and `itePostFixup` in favor of one: `iteFixup :: Tx era -> ImpTestM era (Tx era)`

For times when no fixup is necessary we can just use `pure`, while by default in the `runImpTestM` we can use `fixupTx`.

Then `withPostFixup` will become a composition of the current fixup function that is currently set in `iteFixup` and the one that is supplied as an argument.

@Soupstraw what are your thoughts on this subject? Asking you since you really advocated for a boolean flag.

Soupstraw:
> Let's get rid of iteDoTxFixup and itePostFixup in favor of one: iteFixup :: Tx era -> ImpTestM era (Tx era)

@lehins I like that idea, it's a more flexible version of what we have right now. We can still keep the `withNoFixup` combinator around for convenience, just make it set the fixup function to `pure`.

teodanciu:
I pushed some changes that I believe implement  what @lehins 's  suggestion was . 

I  have doubts about how this will be used! 
So: 
* for no fixup at all:  `withNoFixup`   - as before
* for "auto fixup" and extra things: `withPostFixup` 
* for *only*  extra things, we'll have to do:  `withPostFixup (\f -> withNoFixup ( ..) `  
 Isn't it kinda clunky ?  I have looked at this too much to have an objective view, so happy to go with it if you think so.  @lehins @Soupstraw

eras/shelley/impl/testlib/Test/Cardano/Ledger/Shelley/ImpTest.hs
```diff
@@ -884,11 +897,7 @@ trySubmitTx ::
   Tx era ->
   ImpTestM era (Either (NonEmpty (PredicateFailure (EraRule "LEDGER" era))) (Tx era))
 trySubmitTx tx = do
-  doFixup <- asks iteDoTxFixup
-  txFixed <-
-    if doFixup
-      then fixupTx tx
-      else pure tx
+  txFixed <- asks iteFixup >>= (\f -> f tx)
```
lehins: Here is a cute suggestion :wink: 
```suggestion
  txFixed <- asks iteFixup >>= ($ tx)
```

eras/shelley/impl/testlib/Test/Cardano/Ledger/Shelley/ImpTest.hs
```diff
@@ -1309,7 +1318,10 @@ registerPool = do
 
 -- | Performs the action without running the fix-up function on any transactions
 withNoFixup :: ImpTestM era a -> ImpTestM era a
-withNoFixup = local $ iteDoTxFixupL .~ False
+withNoFixup = local $ iteFixupL .~ pure
+
+withPostFixup :: (Tx era -> ImpTestM era (Tx era)) -> ImpTestM era a -> ImpTestM era a
+withPostFixup f = local $ iteFixupL %~ (f >=>)
```
lehins: Your implementation is actually a `withPreFixup` :smile: It should be `(>=> f)`

My suggestion is to add one `withCustomFixup` that can be exported and would be useful for modification of the fixup function in any way desired. 
```suggestion
withPostFixup f = withCustomFixup (>=> f)

withCustomFixup :: 
  ((Tx era -> ImpTestM era (Tx era)) -> Tx era -> ImpTestM era (Tx era))
  -> ImpTestM era a
  -> ImpTestM era a
withCustomFixup f = local $ iteFixupL %~ f
```

Here is an equivalent implementation for `withNoFixup`, but I don't think it is any simpler than the current implementation:
```haskell
withNoFixup = withCustomFixup (const pure)
```

lehins (commented):
Looks great! One tiny mistake and a suggestion

eras/shelley/impl/testlib/Test/Cardano/Ledger/Shelley/ImpTest.hs
teodanciu: Nice, thanks!

eras/shelley/impl/testlib/Test/Cardano/Ledger/Shelley/ImpTest.hs
teodanciu: Yes! I like the `withCustomFixup`, that solves my concern about this, thank you!

eras/shelley/impl/testlib/Test/Cardano/Ledger/Shelley/ImpTest.hs
```diff
@@ -1309,7 +1318,10 @@ registerPool = do
 
 -- | Performs the action without running the fix-up function on any transactions
 withNoFixup :: ImpTestM era a -> ImpTestM era a
-withNoFixup = local $ iteDoTxFixupL .~ False
+withNoFixup = local $ iteFixupL .~ pure
+
+withPostFixup :: (Tx era -> ImpTestM era (Tx era)) -> ImpTestM era a -> ImpTestM era a
```
Soupstraw:
```suggestion
withFixup :: (Tx era -> ImpTestM era (Tx era)) -> ImpTestM era a -> ImpTestM era a
```

Soupstraw: Oh, sorry nvm, I thought it overwrites the fixup function, but it only composes an arrow with it. I think we should also just have `withFixup` that overwrites the function completely. That would probably solve one of the problems you mentioned.

teodanciu: Yes! Alexey suggested the same, called `withCustomFixup`.  I added this, thanks.

teodanciu: Ah actually `withCustomFixup` is more generic , allowing to compose the configured fixup (or nothing). 
I agree with you, I would like to also have a function that overrides the fixup entirely

Soupstraw: I'd rename `withCustomFixup` to `withModifyFixup`

lehins (approved):
Awesome!

PR #4242, opened 2024-04-05
Added UnitTestTools and IncrementalStakeTest

TimSheard:
This is a second try at addressing #4171.
It replaces PR #4211, which has been abandoned.

Issue #4171 identified some potential problems with the code in [eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState/IncrementalStake.hs]. It suggested some tests to resolve the issue. Such a test need specific values with types such as Credential, StakeReference, UtxO, IncrementalStake, PParams, Rewards, Deposits, all with complex interrelation-ships.

Test.Cardano.Ledger.UnitTestTools makes it easy to express such relationships
Test.Cardano.Ledger.Shelley.UnitTests.IncrementalStakeTest exhibits a test, that resolves the issue.

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/eras/shelley/impl/cardano-ledger-shelley.cabal
+++ b/eras/shelley/impl/cardano-ledger-shelley.cabal
@@ -105,10 +105,10 @@ library
         bytestring,
         cardano-crypto-class,
         cardano-crypto-wrapper,
-        cardano-data ^>=1.2,
+        cardano-data ^>=1.2.2,
         cardano-ledger-binary ^>=1.3,
         cardano-ledger-byron,
-        cardano-ledger-core ^>=1.11,
+        cardano-ledger-core ^>=1.11.1,
         cardano-slotting,
         vector-map ^>=1.1,
         containers,
@@ -144,6 +144,7 @@ library testlib
         Test.Cardano.Ledger.Shelley.Imp.LedgerSpec
         Test.Cardano.Ledger.Shelley.Imp.UtxowSpec
         Test.Cardano.Ledger.Shelley.TreeDiff
+        Test.Cardano.Ledger.Shelley.UnitTests.IncrementalStakeTest
 
     visibility:       public
     hs-source-dirs:   testlib
--- a/eras/shelley/impl/testlib/Test/Cardano/Ledger/Shelley/UnitTests/IncrementalStakeTest.hs
+++ b/eras/shelley/impl/testlib/Test/Cardano/Ledger/Shelley/UnitTests/IncrementalStakeTest.hs
@@ -1,0 +1,138 @@
+{-# LANGUAGE AllowAmbiguousTypes #-}
+{-# LANGUAGE DataKinds #-}
+{-# LANGUAGE RankNTypes #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeApplications #-}
+
+module Test.Cardano.Ledger.Shelley.UnitTests.IncrementalStakeTest (spec) where
+
+import Cardano.Ledger.Address (Addr (..))
+import Cardano.Ledger.BaseTypes (Network (..))
+import Cardano.Ledger.Coin (Coin (..))
+import Cardano.Ledger.Compactible (fromCompact)
+import Cardano.Ledger.Core (Era (..), EraTxOut, emptyPParams, mkCoinTxOut)
+import Cardano.Ledger.Credential (Credential, StakeReference (..))
+import Cardano.Ledger.EpochBoundary (SnapShot (..), Stake (..))
+import Cardano.Ledger.Keys (KeyHash, KeyRole (..))
+import Cardano.Ledger.PoolParams (PoolParams (..))
+import Cardano.Ledger.Shelley.LedgerState (
+  IncrementalStake (..),
+  dsUnifiedL,
+  incrementalStakeDistr,
+  psStakePoolParamsL,
+  updateStakeDistribution,
+ )
+import Cardano.Ledger.TxIn (TxIn)
+import Cardano.Ledger.UMap (RDPair (..), rdRewardCoin)
+import qualified Cardano.Ledger.UMap as UM
+import Cardano.Ledger.UTxO (UTxO (..))
+import Data.Map.Strict (Map)
+import qualified Data.Map.Strict as Map
+import Data.MonoTuple (TupleN)
+import qualified Data.VMap as VMap
+import Lens.Micro
+import Test.Cardano.Ledger.Common
+import Test.Cardano.Ledger.Core.Arbitrary ()
+
+ppIdL :: Lens' (PoolParams c) (KeyHash 'StakePool c)
+ppIdL = lens ppId (\x y -> x {ppId = y})
+
+address :: Credential 'Payment c -> Maybe (Credential 'Staking c) -> Addr c
+address pc Nothing = Addr Testnet pc StakeRefNull
+address pc (Just sc) = Addr Testnet pc (StakeRefBase sc)
+
+arbitraryLens :: Arbitrary a => Lens' a b -> b -> Gen a
+arbitraryLens l b = do a <- arbitrary; pure (a & l .~ b)
+
+-- ===========================================
+
+stakeDistrIncludesRewards :: forall era. EraTxOut era => Gen Property
+stakeDistrIncludesRewards = do
+  (tom, john, ann, ron, mary) <- arbitrary @(TupleN 5 (Credential 'Staking (EraCrypto era)))
+  (tomPay, johnPay, annPay, ronPay) <- arbitrary @(TupleN 4 (Credential 'Payment (EraCrypto era)))
+
+  (pool1, pool2) <- arbitrary @(TupleN 2 (KeyHash 'StakePool (EraCrypto era)))
+  pool1Params <- arbitraryLens ppIdL pool1
+  pool2Params <- arbitraryLens ppIdL pool2
+
+  (tomRD, johnRD, annRD, ronRD, maryRD) <- arbitrary @(TupleN 5 RDPair)
+
+  let tomAddr = address tomPay Nothing -- Nothing means tomAddr does not have a StakeReference
+      johnAddr = address johnPay (Just john)
+      annAddr = address annPay (Just ann)
+      ronAddr = address ronPay (Just ron)
+      -- maryAddr is omitted on purpose. Mary will not have a UTxO entry
+
+      rewards :: Map (Credential 'Staking (EraCrypto era)) RDPair
+      rewards =
+        Map.fromList
+          [ (tom, tomRD)
+          , (john, johnRD)
+          , (ann, annRD)
+          , (ron, ronRD)
+          , (mary, maryRD)
+          ]
+
+      delegations :: Map (Credential 'Staking (EraCrypto era)) (KeyHash 'StakePool (EraCrypto era))
+      delegations =
+        Map.fromList
+          [ (tom, pool1) -- since every one is delegated
+          , (ann, pool2) -- no one's stake should be left out
+          , (ron, pool1)
+          , (john, pool2)
+          , (mary, pool2)
+          ]
+
+  -- Again mary is not included, because she will not have an UTxO entry, but tom will have 2
+  (tomCoin1, tomCoin2, johnCoin, annCoin, ronCoin) <- arbitrary @(TupleN 5 Coin)
+
+  (tomTxIn1, tomTxIn2, johnTxIn, annTxIn, ronTxIn) <- arbitrary @(TupleN 5 (TxIn (EraCrypto era)))
+
+  let
+    -- Each wallet (except mary) has one or more UTxO entries
+    -- Since tom uses a StakeRefNull those entries will not be distributed
+    utxo1 =
+      UTxO @era
+        ( Map.fromList
+            [ (tomTxIn1, mkCoinTxOut tomAddr tomCoin1) -- Not distrubuted, see tomAddr
+            , (tomTxIn2, mkCoinTxOut tomAddr tomCoin2) -- Not distributed, see tomAddr
+            , (annTxIn, mkCoinTxOut annAddr annCoin)
+            , (ronTxIn, mkCoinTxOut ronAddr ronCoin)
+            , (johnTxIn, mkCoinTxOut johnAddr johnCoin)
+            -- Note Mary does not have a UTxO entry, but her rewards are still counted
+            ]
+        )
+
+    pparams = emptyPParams @era
+
+    incrementalStake = updateStakeDistribution pparams (IStake mempty mempty) mempty utxo1
+    umap = UM.unify rewards Map.empty delegations Map.empty
+    poolparamMap = Map.fromList [(pool1, pool1Params), (pool2, pool2Params)]
+  -- We can either use an emptyDstate with just the umap, like this
+  -- dState = (emptyDState {dsUnified = umap}) :: DState era
+  -- Or an arbitrary one, where we overwrite the umap, with the one we need.
+  dState <- arbitraryLens dsUnifiedL umap
+  pState <- arbitraryLens psStakePoolParamsL poolparamMap
+  let computedStakeDistr = Map.map fromCompact (VMap.toMap (unStake (ssStake snap)))
+        where
+          snap =
+            incrementalStakeDistr -- This computes the actual Incremental Stake
+              pparams
+              incrementalStake
+              dState
+              pState
+
+      expectedStakeDistr :: Map (Credential 'Staking (EraCrypto era)) Coin
+      expectedStakeDistr =
+        Map.fromList -- Coin Part is (rdRewardCoin <> utxoCoin)
+          [ (tom, rdRewardCoin tomRD <> Coin 0) -- tom uxtxoCoin is zero because his address has StakeRefNull
+          , (ann, rdRewardCoin annRD <> annCoin)
+          , (ron, rdRewardCoin ronRD <> ronCoin)
+          , (john, rdRewardCoin johnRD <> johnCoin)
+          , (mary, rdRewardCoin maryRD <> Coin 0) -- mary uxtxoCoin is zero because she has no UtxO entry
+          ]
+
+  pure $ (computedStakeDistr === expectedStakeDistr)
+
+spec :: forall era. EraTxOut era => Spec
+spec = prop "StakeDistrIncludesRewards" (stakeDistrIncludesRewards @era)
--- a/libs/cardano-data/CHANGELOG.md
+++ b/libs/cardano-data/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## 1.2.2.0
 
 * Add `assocList`
+* Add module Data.MonoTuple
 
 ## 1.2.1.0
 
--- a/libs/cardano-data/cardano-data.cabal
+++ b/libs/cardano-data/cardano-data.cabal
@@ -24,6 +24,7 @@ library
         Data.Universe
         Data.OSet.Strict
         Data.OMap.Strict
+        Data.MonoTuple
 
     hs-source-dirs:   src
     default-language: Haskell2010
--- a/libs/cardano-data/src/Data/MonoTuple.hs
+++ b/libs/cardano-data/src/Data/MonoTuple.hs
@@ -1,0 +1,29 @@
+{-# LANGUAGE AllowAmbiguousTypes #-}
+{-# LANGUAGE DataKinds #-}
+{-# LANGUAGE FlexibleInstances #-}
+{-# LANGUAGE MultiParamTypeClasses #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeApplications #-}
+{-# LANGUAGE TypeFamilies #-}
+
+-- | Supports creating tuples (where all components have the same type) as Fixed length lists.
+module Data.MonoTuple (
+  TupleN,
+) where
+
+import GHC.TypeNats (Nat)
+
+-- =========================================
+
+type family TupleN (n :: Nat) a where
+  TupleN 0 a = ()
+  TupleN 1 a = a
+  TupleN 2 a = (a, a)
+  TupleN 3 a = (a, a, a)
+  TupleN 4 a = (a, a, a, a)
+  TupleN 5 a = (a, a, a, a, a)
+  TupleN 6 a = (a, a, a, a, a, a)
+  TupleN 7 a = (a, a, a, a, a, a, a)
+  TupleN 8 a = (a, a, a, a, a, a, a, a)
+  TupleN 9 a = (a, a, a, a, a, a, a, a, a)
+  TupleN 10 a = (a, a, a, a, a, a, a, a, a, a)
--- a/libs/cardano-ledger-core/CHANGELOG.md
+++ b/libs/cardano-ledger-core/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 * Add trivial `Inject` instances for `()` and `Void`
 * Add `byteStringToNum`
+* Add functions `rdRewardCoin`, `rdDepositCoin` in UMap.hs
+* Add function `mkCoinTxOut` in Core.hs
 
 ## 1.11.0.0
 
--- a/libs/cardano-ledger-core/src/Cardano/Ledger/Core.hs
+++ b/libs/cardano-ledger-core/src/Cardano/Ledger/Core.hs
@@ -39,6 +39,7 @@ module Cardano.Ledger.Core (
   hashScriptTxWitsL,
   Value,
   EraPParams (..),
+  mkCoinTxOut,
 
   -- * Era
   module Cardano.Ledger.Core.Era,
@@ -104,7 +105,7 @@ import Cardano.Ledger.MemoBytes
 import Cardano.Ledger.Rewards (Reward (..), RewardType (..))
 import Cardano.Ledger.SafeHash (HashAnnotated (..), SafeToHash (..))
 import Cardano.Ledger.TxIn (TxId (..), TxIn (..))
-import Cardano.Ledger.Val (Val (..))
+import Cardano.Ledger.Val (Val (..), inject)
 import Control.DeepSeq (NFData)
 import Data.Aeson (ToJSON)
 import qualified Data.ByteString as BS
@@ -424,6 +425,10 @@ isAdaOnlyTxOutF = to $ \txOut ->
 toCompactPartial :: (HasCallStack, Val a) => a -> CompactForm a
 toCompactPartial v =
   fromMaybe (error $ "Illegal value in TxOut: " <> show v) $ toCompact v
+
+-- A version of mkBasicTxOut, which has only a Coin (no multiAssets) for every EraTxOut era.
+mkCoinTxOut :: EraTxOut era => Addr (EraCrypto era) -> Coin -> TxOut era
+mkCoinTxOut addr = mkBasicTxOut addr . inject
 
 -- | A value is something which quantifies a transaction output.
 type family Value era :: Type
--- a/libs/cardano-ledger-core/src/Cardano/Ledger/UMap.hs
+++ b/libs/cardano-ledger-core/src/Cardano/Ledger/UMap.hs
@@ -25,6 +25,8 @@
 module Cardano.Ledger.UMap (
   -- * Constructing a `UMap`
   RDPair (..),
+  rdRewardCoin,
+  rdDepositCoin,
   UMElem (UMElem),
   umElemRDPair,
   umElemRDActive,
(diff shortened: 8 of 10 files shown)


libs/cardano-ledger-core/testlib/Test/Cardano/Ledger/UnitTestTools.hs
```diff
@@ -0,0 +1,184 @@
...
+import Test.Cardano.Ledger.Core.Arbitrary ()
+import Test.Cardano.Ledger.Core.KeyPair (KeyPair (..), mkWitnessVKey)
+import Test.QuickCheck
+
+-- ==================================================
+
+-- | Supports patterns like
+--   (tom, john, ann, ron, mary) <- fresh @5 @(Credential 'Staking (EraCrypto era))
+class KnownNat n => Fresh (n :: Nat) (a :: Type) where
+  type Tuple n a :: Type
```
Soupstraw:
```suggestion
  type Tuple n a = (r :: Type) | n <- r
```
I wonder if something like this would help get rid of the tuple size type annotations

TimSheard: The problem with type families is that they don't reduce until both arguments are known, so we can't use functional dependency information until both are known. I tried it. It didn't work, but it was a good idea.

eras/shelley/impl/testlib/Test/Cardano/Ledger/Shelley/UnitTests/IncrementalStakeTest.hs
```diff
@@ -0,0 +1,132 @@
...
+
+{-
+InQuickCheck we might say
+property (assertBool (3==4)) :: Property
+in Hspec we might say
+prop "adc" (assertBool "3==4" (3==4)) :: Spec
+or we could say
+prop "abc" (True === False) :: Spec
+
+-}
```
lehins: This is not an exhaustive list of what can be done. In any case, we don't want comments in the code on how to use hspec.
```suggestion
```

libs/cardano-ledger-core/testlib/Test/Cardano/Ledger/UnitTestTools.hs
```diff
@@ -0,0 +1,184 @@
...
+someKeyHashWitVKeyPair = do
+  safehash <- arbitrary @(SafeHash StandardCrypto EraIndependentTxBody)
+  keyHashWitVKeyPairFor safehash
+
+addr :: Credential 'Staking c -> Maybe (Credential 'Staking c) -> Addr c
+addr cr Nothing = Addr Testnet (coerceKeyRole cr) StakeRefNull
+addr c (Just d) = Addr Testnet (coerceKeyRole c) (StakeRefBase d)
+
+txOut :: forall era. EraTxOut era => Addr (EraCrypto era) -> Coin -> TxOut era
+txOut add coinx = mkBasicTxOut add (modifyCoin (const coinx) zero)
```
lehins: `addr` and `txOut` are common names for local bindings, that's why we need to avoid names like that for reusable functions. `mkTestAddr` and `mkCoinTxOut` could be good alternative names. Also why do you need `coerceKeyRole` and `modifyCoin`:

```suggestion
mkTestAddr :: Credential 'Staking c -> Maybe (Credential 'Staking c) -> Addr c
mkTestAddr cred = Addr Testnet cred . maybe StakeRefNull StakeRefBase

mkCoinTxOut :: EraTxOut era => Addr (EraCrypto era) -> Coin -> TxOut era
mkCoinTxOut addr = mkBasicTxOut addr . inject
```

* `mkCoinTxOut` can go into `Cardano.Ledger.Core`, since it can be used outside of testing

* `mkTestAddr` can go into `Test.Cardano.Ledger.Core.KeyPairs` next to `mkAddr` (which itself hopefully we can rename in the future)

libs/cardano-ledger-core/testlib/Test/Cardano/Ledger/UnitTestTools.hs
```diff
@@ -0,0 +1,184 @@
...
+emptyPState = def
+
+emptyDState :: DState era
+emptyDState = def
+
+emptyVState :: VState era
+emptyVState = def
+
+emptyCertState :: CertState era
+emptyCertState = def
```
lehins: We keep removing this kind of functions and you keep adding them. See this example: https://github.com/IntersectMBO/cardano-ledger/blob/1d4ce0853d1136452cd644bd7b0747dd40b976cf/eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState.hs#L184-L186 

Please just use `def @(DState era)` if you need something like that
```suggestion
```

libs/cardano-ledger-core/testlib/Test/Cardano/Ledger/UnitTestTools.hs
```diff
@@ -0,0 +1,184 @@
...
+
+-- | rdReward returns a (CompactForm Coin), This returns a Coin
+rewardCoin :: RDPair -> Coin
+rewardCoin rdp = fromCompact (rdReward rdp)
+
+depositCoin :: RDPair -> Coin
+depositCoin rdp = fromCompact (rdDeposit rdp)
```
lehins: This functions could go into the `UMap` module instead. Better names for consistency would be `rdRewardCoin` and `rdDepositCoin`

libs/cardano-ledger-core/testlib/Test/Cardano/Ledger/UnitTestTools.hs
```diff
@@ -0,0 +1,184 @@
...
+      keyhash = hashKey vk
+  pure (keyhash, witness)
+
+-- | Use this when any TxBody hash will do
+--   You just need a KeyHash and its Witness
+someKeyHashWitVKeyPair ::
+  forall kr. Gen (KeyHash kr StandardCrypto, WitVKey 'Witness StandardCrypto)
+someKeyHashWitVKeyPair = do
+  safehash <- arbitrary @(SafeHash StandardCrypto EraIndependentTxBody)
+  keyHashWitVKeyPairFor safehash
```
lehins: These can go into `Test.Cardano.Ledger.Core.KeyPairs`

libs/cardano-ledger-core/testlib/Test/Cardano/Ledger/UnitTestTools.hs
```diff
@@ -0,0 +1,184 @@
...
+  type Tuple 5 a = Gen (a, a, a, a, a)
+  fresh = (arbitrary @(a, a, a, a, a))
+
+instance Fresh 6 a where
+  type Tuple 6 a = Gen (a, a, a, a, a, a)
+  fresh = (arbitrary @(a, a, a, a, a, a))
+
+instance Fresh 7 a where
+  type Tuple 7 a = Gen (a, a, a, a, a, a, a)
+  fresh = (arbitrary @(a, a, a, a, a, a, a))
```
lehins: This solution is way too complicated than it has to be. This is all it takes to do the same thing:
```haskell
type family TupleN (n :: Nat) a where
  TupleN 0 a = ()
  TupleN 1 a = a
  TupleN 2 a = (a, a)
  TupleN 3 a = (a, a, a)
  TupleN 4 a = (a, a, a, a)
  TupleN 5 a = (a, a, a, a, a)
  TupleN 6 a = (a, a, a, a, a, a)
  TupleN 7 a = (a, a, a, a, a, a, a)
```

Then you can use it in a very similar manner:
```haskell
myGen = do
  (x1, x2, x3) <- arbitrary @(TupleN 3 Int)
  pure $ x1 + x2 + x3
```

Note that this solution is not even specific to `QuickCheck`.

If you ok with switching to what I proposed then we can put `TupleN` into `cardano-data` package: `Data.TupleN`

eras/shelley/impl/testlib/Test/Cardano/Ledger/Shelley/UnitTests/IncrementalStakeTest.hs
```diff
@@ -0,0 +1,132 @@
...
+-- ===========================================
+-- A running example
+
+test1 :: forall era. EraTxOut era => Gen Property
+test1 = do
+  (tom, john, ann, ron, mary) <- fresh @5 @(Credential 'Staking (EraCrypto era))
+  (pool1, pool2) <- fresh @2 @(KeyHash 'StakePool (EraCrypto era))
+  pool1Params <- arbitraryLens ppIdL pool1
+  pool2Params <- arbitraryLens ppIdL pool2
```
lehins, line 56: There are no need to create new abstractions for things that can be easily done without:
```suggestion
  pool1Params <- set ppIdL pool1 <$> arbitrary
  pool2Params <- set ppIdL pool2 <$> arbitrary
```

libs/cardano-ledger-core/testlib/Test/Cardano/Ledger/UnitTestTools.hs
lehins, line 1: With all of my comments this module can be removed

eras/shelley/impl/testlib/Test/Cardano/Ledger/Shelley/UnitTests/IncrementalStakeTest.hs
```diff
@@ -0,0 +1,132 @@
...
+
+  (tomRD, johnRD, annRD, ronRD, maryRD) <- fresh @5 @RDPair
```
lehins: This is how it would look like wit `TupleN` type family
```suggestion
  (tomRD, johnRD, annRD, ronRD, maryRD) <- arbitrary @(TupleN 5 RDPair)
```

eras/shelley/impl/testlib/Test/Cardano/Ledger/Shelley/UnitTests/IncrementalStakeTest.hs
```diff
@@ -0,0 +1,132 @@
...
+          , (ann, rewardCoin annRD <> annCoin)
+          , (ron, rewardCoin ronRD <> ronCoin)
+          , (john, rewardCoin johnRD <> johnCoin)
+          , (mary, rewardCoin maryRD <> Coin 0) -- mary uxtxoCoin is zero because address has StakeRefNull
+          ]
+
+  pure $ property (computedStakeDistr === expectedStakeDistr)
+
+go :: IO ()
+go = hspec (prop "bigtest" (test1 @Shelley))
```
lehins: Please don't forget to remove it
```suggestion
```

eras/shelley/impl/testlib/Test/Cardano/Ledger/Shelley/UnitTests/IncrementalStakeTest.hs
```diff
@@ -0,0 +1,132 @@
...
+  pool1Params <- arbitraryLens ppIdL pool1
+  pool2Params <- arbitraryLens ppIdL pool2
+
+  (tomRD, johnRD, annRD, ronRD, maryRD) <- fresh @5 @RDPair
+
+  let tomAddr = addr tom Nothing -- Nothing means tomAddr does not have a StakeReference
+      johnAddr = addr john (Just john)
+      annAddr = addr ann (Just ann)
+      ronAddr = addr ron (Just ron)
+      _maryAddr = addr mary Nothing
```
lehins: I am sure you are planning to remove those bindings with underscore, but just in case I'll post here a reminder.
```suggestion
```

eras/shelley/impl/testlib/Test/Cardano/Ledger/Shelley/UnitTests/IncrementalStakeTest.hs
```diff
@@ -0,0 +1,132 @@
...
+      expectedStakeDistr =
+        Map.fromList -- Coin Part is (rewardCoin <> utxoCoin)
+          [ (tom, rewardCoin tomRD <> Coin 0) -- tom uxtxoCoin is zero because address has StakeRefNull
+          , (ann, rewardCoin annRD <> annCoin)
+          , (ron, rewardCoin ronRD <> ronCoin)
+          , (john, rewardCoin johnRD <> johnCoin)
+          , (mary, rewardCoin maryRD <> Coin 0) -- mary uxtxoCoin is zero because address has StakeRefNull
+          ]
+
+  pure $ property (computedStakeDistr === expectedStakeDistr)
```
lehins: Redundant
```suggestion
  pure $ computedStakeDistr === expectedStakeDistr
```

libs/cardano-ledger-core/testlib/Test/Cardano/Ledger/UnitTestTools.hs
```diff
@@ -0,0 +1,184 @@
...
+
+genLens :: Gen a -> Lens' a b -> b -> Gen a
+genLens gen l b = do a <- gen; pure (a & l .~ b)
+
+arbitraryLens :: Arbitrary a => Lens' a b -> b -> Gen a
+arbitraryLens = genLens arbitrary
```
lehins: These helpers aren't very useful, see my other comment how `set` can be used to get the desired behavior in much more clear fashion
```suggestion
```

lehins (commented):
The test looks awesome!

As usual I have a bunch of suggestions, but none of them are about the test case. Hopefully you'll like my suggestion to improve the tuple solution :grin:

eras/shelley/impl/testlib/Test/Cardano/Ledger/Shelley/UnitTests/IncrementalStakeTest.hs
TimSheard: done

eras/shelley/impl/testlib/Test/Cardano/Ledger/Shelley/UnitTests/IncrementalStakeTest.hs
TimSheard: well of course it was marked as a Draft PR.

libs/cardano-ledger-core/testlib/Test/Cardano/Ledger/UnitTestTools.hs
TimSheard: So I called the new package Data.MonoTuple

libs/cardano-ledger-core/testlib/Test/Cardano/Ledger/UnitTestTools.hs
TimSheard: done

libs/cardano-ledger-core/testlib/Test/Cardano/Ledger/UnitTestTools.hs
TimSheard: So there is a reason for this.  Suppose you have
```
data PState era
  = PState {psStakePoolParams :: !(Map
                                     (KeyHash 'StakePool (EraCrypto era))
                                     (PoolParams (EraCrypto era))),
            psFutureStakePoolParams :: !(Map
                                           (KeyHash 'StakePool (EraCrypto era))
                                           (PoolParams (EraCrypto era))),
            psRetiring :: !(Map
                              (KeyHash 'StakePool (EraCrypto era))
                              Cardano.Slotting.Slot.EpochNo),
            psDeposits :: !(Map (KeyHash 'StakePool (EraCrypto era)) Coin)}
```
and State has a Default instance.

Now if you write
`def{ppId = x}`

Haskell  cannot type check this, because era is ambiguous.  This is because the data for PState never mentions era, only . (EraCrypto era)

So one has to put a type on def. But there is no room to put a type on def. because the haskell parser does not let you write
`def @(PState era) {def{ppId = x}`
So emptyPState avoids this.  

So to avoid this problem I will  use  `arbritraryLens` instead.

libs/cardano-ledger-core/testlib/Test/Cardano/Ledger/UnitTestTools.hs
TimSheard: so I moved mCoinTxOut to Cardano.Ledger.Core, But the other I just moved to the actual test file. Maybe its not really as reusable as it appears.

libs/cardano-ledger-core/testlib/Test/Cardano/Ledger/UnitTestTools.hs
TimSheard: So I made a better version of these two, and put them in Test.Cardano.Ledger.Core.KeyPairs

eras/shelley/impl/testlib/Test/Cardano/Ledger/Shelley/UnitTests/IncrementalStakeTest.hs
TimSheard: Not sure I agree with this. So I put arbitraryLens in the testfile. It is a 1 liner and it is used several times.

eras/shelley/impl/testlib/Test/Cardano/Ledger/Shelley/UnitTests/IncrementalStakeTest.hs
TimSheard: looks good. That now.

eras/shelley/impl/testlib/Test/Cardano/Ledger/Shelley/UnitTests/IncrementalStakeTest.hs
TimSheard: done

libs/cardano-ledger-core/testlib/Test/Cardano/Ledger/UnitTestTools.hs
TimSheard: see above.

libs/cardano-ledger-core/testlib/Test/Cardano/Ledger/UnitTestTools.hs
TimSheard: done

eras/shelley/impl/testlib/Test/Cardano/Ledger/Shelley/UnitTests/IncrementalStakeTest.hs
TimSheard: This was for me. I never intended to leave it there.

eras/shelley/impl/testlib/Test/Cardano/Ledger/Shelley/UnitTests/IncrementalStakeTest.hs
```diff
@@ -0,0 +1,138 @@
+{-# LANGUAGE AllowAmbiguousTypes #-}
+{-# LANGUAGE DataKinds #-}
+{-# LANGUAGE RankNTypes #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeApplications #-}
+
+module Test.Cardano.Ledger.Shelley.UnitTests.IncrementalStakeTest where
```
lehins: By exporting only the things that we need we reduce the mental overhead we have to deal with when dealing with versioning. Eg. removal of `arbitraryLens` in the future will not require a breaking change on the `cardano-ledger-shelley` package, since it will be local to this module.
```suggestion
module Test.Cardano.Ledger.Shelley.UnitTests.IncrementalStakeTest (spec) where
```

eras/shelley/impl/testlib/Test/Cardano/Ledger/Shelley/UnitTests/IncrementalStakeTest.hs
```diff
@@ -0,0 +1,138 @@
...
+address pc Nothing = Addr Testnet pc StakeRefNull
+address pc (Just sc) = Addr Testnet pc (StakeRefBase sc)
+
+arbitraryLens :: Arbitrary a => Lens' a b -> b -> Gen a
+arbitraryLens l b = do a <- arbitrary; pure (a & l .~ b)
+
+-- ===========================================
+
+test1 :: forall era. EraTxOut era => Gen Property
+test1 = do
```
lehins: Could we use a better name for this, instead of `test1`?

libs/cardano-ledger-core/src/Cardano/Ledger/Core.hs
```diff
@@ -39,6 +39,7 @@ module Cardano.Ledger.Core (
   hashScriptTxWitsL,
   Value,
   EraPParams (..),
+  mkCoinTxOut,
```
lehins, line 42: This needs to be added here and version bumped (in cabal too): https://github.com/IntersectMBO/cardano-ledger/blob/88ff0c593a0e0b4167636ba947d68521b3e19e2a/libs/cardano-ledger-core/CHANGELOG.md?plain=1#L3-L5

```diff
-## 1.11.0.1
+## 1.11.1.0

-*
+* Add `mkCoinTxOut`
```

Also, because `cardano-ledger-shelley` depends on this new function, it needs a new lower bound: `cardano-ledger-core >= 1.11.1`

libs/cardano-ledger-core/src/Cardano/Ledger/UMap.hs
```diff
@@ -141,6 +143,14 @@ data RDPair = RDPair
   }
   deriving (Show, Eq, Ord, Generic, NoThunks, NFData)
 
+-- rdReward and rdDeposit return a (CompactForm Coin), These return a Coin.
+
+rdRewardCoin :: RDPair -> Coin
+rdRewardCoin rdp = fromCompact (rdReward rdp)
+
+rdDepositCoin :: RDPair -> Coin
+rdDepositCoin rdp = fromCompact (rdDeposit rdp)
```
lehins, line 152: These two also need to be mentioned in the changelog

lehins (approved):
Beautiful. 
Thank you for your patience, Tim!

libs/cardano-ledger-core/src/Cardano/Ledger/Core.hs
```diff
@@ -425,6 +426,10 @@ toCompactPartial :: (HasCallStack, Val a) => a -> CompactForm a
 toCompactPartial v =
   fromMaybe (error $ "Illegal value in TxOut: " <> show v) $ toCompact v
 
+-- A version of mkBasicTxOut, which has only a Coin (no multiAssets) for every EraTxOut era.
+mkCoinTxOut :: forall era. EraTxOut era => Addr (EraCrypto era) -> Coin -> TxOut era
+mkCoinTxOut add coinx = mkBasicTxOut add (modifyCoin (const coinx) zero)
```
lehins: Just noticed that one of my previous suggestions wasn't applied, so I'll add it again.
```suggestion
mkCoinTxOut addr = mkBasicTxOut addr . inject
```

libs/cardano-data/cardano-data.cabal
```diff
@@ -24,6 +24,7 @@ library
         Data.Universe
         Data.OSet.Strict
         Data.OMap.Strict
+        Data.MonoTuple
```
lehins, line 27: I think you forgot to `git add` the actual module :slightly_smiling_face: 
Happens to me all the time

lehins: Similarly to `cardano-ledger-core`, this will also mean a minor version bump on `cardano-data` and a lower bound updated for it in `cardano-ledger-shelley`.

Changelog entry will be needed too.

libs/cardano-ledger-core/testlib/Test/Cardano/Ledger/UnitTestTools.hs
lehins: > So one has to put a type on def. But there is no room to put a type on def. because the haskell parser does not let you write
def @(PState era) {def{ppId = x}
So emptyPState avoids this.

There is an easy solution to this, just wrap the expression in parenthesis:

```haskell
(def @(PState era)) {ppId = x}
```

eras/shelley/impl/testlib/Test/Cardano/Ledger/Shelley/UnitTests/IncrementalStakeTest.hs
TimSheard: changed it.

libs/cardano-ledger-core/src/Cardano/Ledger/Core.hs
TimSheard: I think I did the right thing.   Maybe you could check.

libs/cardano-ledger-core/src/Cardano/Ledger/UMap.hs
TimSheard: done

eras/shelley/impl/cardano-ledger-shelley.cabal
```diff
@@ -105,10 +105,10 @@ library
         bytestring,
         cardano-crypto-class,
         cardano-crypto-wrapper,
-        cardano-data ^>=1.2,
+        cardano-data ^>=1.2.1,
         cardano-ledger-binary ^>=1.3,
         cardano-ledger-byron,
-        cardano-ledger-core ^>=1.11,
+        cardano-ledger-core ^>=1.11.0,
```
lehins:
```suggestion
        cardano-ledger-core ^>=1.11.1,
```

libs/cardano-data/cardano-data.cabal
```diff
@@ -1,6 +1,6 @@
 cabal-version:      3.0
 name:               cardano-data
-version:            1.2.1.0
+version:            1.2.1.1
```
lehins: Patch version is for non semantic changes
```suggestion
version:            1.2.2.0
```

libs/cardano-data/CHANGELOG.md
```diff
@@ -1,9 +1,13 @@
...
 
-## 1.2.1.1
+## 1.2.1.2
 
 *
 
+## 1.2.1.1
+
+* Add module Data.MonoTuple
+
```
lehins: New section is NEVER added to the changelog on a regular PR. There is even a reminder on the PR in one of the checkboxes:
> New section is never added with the code changes. (See [RELEASING.md](https://github.com/intersectmbo/cardano-ledger/blob/master/RELEASING.md#changelogmd))

Also, patch version update is only for non-semantic changes

```suggestion
## 1.2.2.0

* Add module Data.MonoTuple

```

libs/cardano-ledger-core/cardano-ledger-core.cabal
```diff
@@ -1,6 +1,6 @@
 cabal-version:      3.0
 name:               cardano-ledger-core
-version:            1.11.0.0
+version:            1.11.0.1
```
lehins:
```suggestion
version:            1.11.1.0
```

libs/cardano-ledger-core/src/Cardano/Ledger/Core.hs
lehins: Nope you, you did not do the right thing. See my suggestions

libs/cardano-ledger-core/src/Cardano/Ledger/Core.hs
```diff
@@ -425,6 +426,10 @@ toCompactPartial :: (HasCallStack, Val a) => a -> CompactForm a
 toCompactPartial v =
   fromMaybe (error $ "Illegal value in TxOut: " <> show v) $ toCompact v
 
+-- A version of mkBasicTxOut, which has only a Coin (no multiAssets) for every EraTxOut era.
+mkCoinTxOut :: forall era. EraTxOut era => Addr (EraCrypto era) -> Coin -> TxOut era
```
lehins: `forall` is redundant
```suggestion
mkCoinTxOut :: EraTxOut era => Addr (EraCrypto era) -> Coin -> TxOut era
```

libs/cardano-ledger-core/CHANGELOG.md
```diff
@@ -1,8 +1,13 @@
 # Version history for `cardano-ledger-core`
 
+## 1.11.0.2
+
+* 
+
```
lehins: Same here, the version should be `1.11.1.0`
```suggestion
```

libs/cardano-data/src/Data/MonoTuple.hs
```diff
@@ -0,0 +1,48 @@
...
+
+instance MonoTuple 2 a where list (x, y) = [x, y]
+instance MonoTuple 3 a where list (x, y, z) = [x, y, z]
+instance MonoTuple 4 a where list (w, x, y, z) = [w, x, y, z]
+instance MonoTuple 5 a where list (v, w, x, y, z) = [v, w, x, y, z]
+instance MonoTuple 6 a where list (u, v, w, x, y, z) = [u, v, w, x, y, z]
+instance MonoTuple 7 a where list (t, u, v, w, x, y, z) = [t, u, v, w, x, y, z]
+instance MonoTuple 8 a where list (s, t, u, v, w, x, y, z) = [s, t, u, v, w, x, y, z]
+instance MonoTuple 9 a where list (r, s, t, u, v, w, x, y, z) = [r, s, t, u, v, w, x, y, z]
+instance MonoTuple 10 a where list (q, r, s, t, u, v, w, x, y, z) = [q, r, s, t, u, v, w, x, y, z]
```
lehins: Dead code. This is not used anywhere. Please don't tell me that it can be useful at some point in the future. If we ever need it, I am sure we can find something similar on hackage. We should not be adding code to the repo on a hypothesis that it might be useful.
```suggestion
```

eras/shelley/impl/cardano-ledger-shelley.cabal
```diff
@@ -105,10 +105,10 @@ library
         bytestring,
         cardano-crypto-class,
         cardano-crypto-wrapper,
-        cardano-data ^>=1.2,
+        cardano-data ^>=1.2.1,
```
lehins:
```suggestion
        cardano-data ^>=1.2.2,
```

lehins (changes requested):
Versioning is off.

lehins (approved):
Perfect

PR #4244, opened 2024-04-06
Imptests: CommitteeMinSize affects in-flight props

### LINKED ISSUE #4145: Imp test that changing `ppCommitteeMinSize` affects in flight proposals
Changing the `ppCommitteeMinSize` can effectively make the committee too small, thus preventing any proposal that a Committee Member can vote on from being enacted.

We would like to test that whenever there is `ParameterChange` proposal with enough votes that changes `ppCommitteeMinSize` and at the same time there is a `TreasuryWithdrawal`:

* that also has enough votes, then enactment of `ParameterChange` that increases the `ppCommitteeMinSize` prevents `TreasuryWithdrawal` from being enacted
* that does not have enough CC votes, then decreasing `ppCommitteeMinSize` allows for `TreasuryWithdrawal` to be enacted on the same epoch boundary as `ParameterChange`

This ticket is similar to #4132

aniketd:
# Description

Resolves #4145 

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/RatifySpec.hs
+++ b/eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/RatifySpec.hs
@@ -17,6 +17,7 @@ import Cardano.Ledger.Credential
 import Cardano.Ledger.Keys
 import Cardano.Ledger.Shelley.LedgerState
 import qualified Cardano.Ledger.UMap as UM
+import Cardano.Ledger.Val ((<->))
 import Data.Default.Class (def)
 import Data.Foldable
 import qualified Data.Map.Strict as Map
@@ -36,6 +37,7 @@ spec =
     votingSpec
     delayingActionsSpec
     spoVotesForHardForkInitiation
+    committeeMinSizeAffectsInFlightProposalsSpec
     paramChangeAffectsProposalsSpec
 
 paramChangeAffectsProposalsSpec ::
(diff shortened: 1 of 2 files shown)


eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/ImpTest.hs
```diff
@@ -93,6 +93,7 @@ module Test.Cardano.Ledger.Conway.ImpTest (
   minorFollow,
   majorFollow,
   cantFollow,
+  isCommitteeAccepted,
```
lehins: Can you move the export next to related `isDRepAccepted` and `isSpoAccepted`?

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/RatifySpec.hs
```diff
@@ -35,6 +36,41 @@ spec =
...
+  SpecWith (ImpTestState era)
+committeeMinSizeAffectsInFlightProposalsSpec =
+  describe "CommitteeMinSize affects in-flight proposals" $ do
+    let setCommitteeMinSize n = modifyPParams $ ppCommitteeMinSizeL .~ n
+        submitATreasuryWithdrawal = do
+          rewardAccount <- registerRewardAccount
+          submitTreasuryWithdrawals [(rewardAccount, Coin 1_000)]
+    it
+      "TreasuryWithdrawal with enough CC votes fails to ratify following an increase in CommitteeMinSize"
+      $ do
```
lehins: The long name and tits formatting as a consequence is triggering OCD in me :smile: 
Maybe something shorter instead?
```suggestion
    it "TreasuryWithdrawal fails to ratify due to an increase in CommitteeMinSize" $ do
```
("enough votes" part I think can be implied)

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/RatifySpec.hs
```diff
@@ -35,6 +36,41 @@ spec =
...
+        (_drepC, hotCommitteeC, _gpiCC) <- electBasicCommittee
+        gaiTW <- submitATreasuryWithdrawal
+        submitYesVote_ (CommitteeVoter hotCommitteeC) gaiTW
+        setCommitteeMinSize 1
+        isCommitteeAccepted gaiTW `shouldReturn` True
+        -- We do not enact the ParameterChange here because that will  cause the
+        -- TreasuryWithdrawal to get enacted
+        setCommitteeMinSize 2
+        isCommitteeAccepted gaiTW `shouldReturn` False
+    it "TreasuryWithdrawal with not enough CC votes ratifies following a decrease in CommitteeMinSize" $ do
```
lehins: Same here:
```suggestion
    it "TreasuryWithdrawal ratifies due to a decrease in CommitteeMinSize" $ do
```

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/RatifySpec.hs
```diff
@@ -35,6 +36,41 @@ spec =
...
+          submitTreasuryWithdrawals [(rewardAccount, Coin 1_000)]
+    it
+      "TreasuryWithdrawal with enough CC votes fails to ratify following an increase in CommitteeMinSize"
+      $ do
+        (_drepC, hotCommitteeC, _gpiCC) <- electBasicCommittee
+        gaiTW <- submitATreasuryWithdrawal
+        submitYesVote_ (CommitteeVoter hotCommitteeC) gaiTW
+        setCommitteeMinSize 1
+        isCommitteeAccepted gaiTW `shouldReturn` True
+        -- We do not enact the ParameterChange here because that will  cause the
```
lehins: I think this comment is actually incorrect.

If we modify the `CommitteeMinSize` parameter using the `ParameterUpdate` governance action it will prevent the treasury withdrawal from getting enacted, because `ParameterUpdate` has higher priority.

So, I think it would be worthwhile avoiding `setCommitteeMinSize` in this test and verify this behavior without manual state modification.

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/RatifySpec.hs
```diff
@@ -35,6 +36,41 @@ spec =
...
+        setCommitteeMinSize 2
+        isCommitteeAccepted gaiTW `shouldReturn` False
+    it "TreasuryWithdrawal with not enough CC votes ratifies following a decrease in CommitteeMinSize" $ do
+      (_drepC, hotCommitteeC, _gpiCC) <- electBasicCommittee
+      gaiTW <- submitATreasuryWithdrawal
+      submitYesVote_ (CommitteeVoter hotCommitteeC) gaiTW
+      setCommitteeMinSize 2
+      getsNES (nesEsL . curPParamsEpochStateL . ppCommitteeMinSizeL) `shouldReturn` 2
+      isCommitteeAccepted gaiTW `shouldReturn` False
+      -- We do not enact the ParameterChange here because that does  not pass
```
lehins:
```suggestion
      -- We do not enact the ParameterChange here because that does not pass
```

lehins (commented):
I think one of the tests can be improved, other than that it looks good.

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/RatifySpec.hs
aniketd: Absolutely :100: :raised_hands: Thanks! :bow:

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/RatifySpec.hs
```diff
@@ -223,6 +225,54 @@ paramChangeAffectsProposalsSpec =
...
+      gaiPC <-
+        submitParameterChange SNothing $
+          emptyPParamsUpdate
+            & ppuCommitteeMinSizeL .~ SJust 2
+      submitYesVote_ (CommitteeVoter hotCommitteeC) gaiPC
+      submitYesVote_ (DRepVoter drepC) gaiPC
+      treasury <- getsNES $ nesEsL . esAccountStateL . asTreasuryL
+      gaiTW <- submitATreasuryWithdrawal
+      submitYesVote_ (CommitteeVoter hotCommitteeC) gaiTW
+      submitYesVote_ (DRepVoter drepC) gaiTW
```
lehins: To make it a little more interesting, let's submit the `TreasuryWithdrawal` before the `ParameterChange` proposal

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/RatifySpec.hs
```diff
@@ -223,6 +225,54 @@ paramChangeAffectsProposalsSpec =
...
+      setCommitteeMinSize 2
+      isCommitteeAccepted gaiTW `shouldReturn` False
+      passNEpochs 2
+      getsNES (nesEsL . esAccountStateL . asTreasuryL) `shouldReturn` treasury
+      -- We do not enact the ParameterChange here because that does not pass
+      -- ratification as the CC size is smaller than MinSize
+      setCommitteeMinSize 1
```
lehins: Instead of changing the parameter (since that is an impossible case) let's instead add more committee members to satisfy the `CommitteeMinSize` and ensure that it also triggers the `TreasuryWithdrawal` enactment

lehins (commented):
Looks good. Just a couple of minor suggestions

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/RatifySpec.hs
```diff
@@ -223,6 +225,66 @@ paramChangeAffectsProposalsSpec =
...
+          UpdateCommittee
+            (SJust gpiCC)
+            Set.empty
+            (Map.singleton coldCommitteeC' $ EpochNo 10)
+            (1 %! 2)
+      submitYesVote_ (DRepVoter drepC) gaiCC
+      passNEpochs 2
+      _hotCommitteeC' <- registerCommitteeHotKey coldCommitteeC'
+      isCommitteeAccepted gaiTW `shouldReturn` True
+      passNEpochs 2
```
lehins, line 285: We only need to pass one epoch here. It is not enacted together with CC update, because it was delayed, so it doesn't need 2 more full epochs. (Could be useful to add it as a comment as well)
```suggestion
      passEpoch
```

lehins (approved):
Looks good. One minor adjustment.

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/RatifySpec.hs
aniketd: Unfortunately, this makes the test fail. I have reverted this change. :+1:

PR #4252, opened 2024-04-10
Add PlutusV3 CostModel to UpgradeConwayPParams

### LINKED ISSUE #4224: Add PlutusV3 CostModel to the Conway genesis
In order to prevent getting stuck, as described in "Unintentional Deadlock" in https://github.com/IntersectMBO/cardano-ledger/issues/4153, we need ability to inject cost model for PlutusV3 when Conway begins

1. Add new required field to Conway Genesis that contains the CostModel only for PlutusV3, prior plutus versions should not be afftected. 
2. Ensure that CostModel for PlutusV3 is set when `ConwayGenesis` is applied during `PParams` translation.

aniketd:
# Description

Resolves #4224

# Checklist

- [x] Self-reviewed the diff


eras/conway/impl/src/Cardano/Ledger/Conway/PParams.hs
```diff
@@ -611,6 +627,17 @@ instance Default (UpgradeConwayPParams Identity) where
       , ucppDRepDeposit = Coin 0
       , ucppDRepActivity = EpochInterval 0
       , ucppMinFeeRefScriptCostPerByte = minBound
+      , ucppPlutusV3CostModel =
+          either
+            ( error
+                . const
+                  ( "Impossible! Implementation of Default for ConwayGenesis. "
+                      <> "mkCostModel PlutusV3 mempty cannot fail."
```
aniketd:
```suggestion
                      <> "mkCostModel PlutusV3 (relicate 233 0) cannot fail."
```

teodanciu (commented):
I reckon this requires the conway version to become 1.14.0.0.
And the Changelog for both the updated conway version and core 1.11.1.0 to be updated.

eras/conway/impl/src/Cardano/Ledger/Conway/PParams.hs
lehins: As I mentioned in the meeting, if we can't remove the `Default` instance for some reason, then I would suggest `ucppPlutusV3CostModel` to contain a `StrictMaybe CostModel` and enforce the presence only in the `FromJSON` instance.

eras/conway/impl/src/Cardano/Ledger/Conway/PParams.hs
```diff
@@ -1179,7 +1215,17 @@ upgradeConwayPParams UpgradeConwayPParams {..} BabbagePParams {..} =
     , cppProtocolVersion = toNoUpdate @f @ProtVer bppProtocolVersion
     , cppMinPoolCost = THKD bppMinPoolCost
     , cppCoinsPerUTxOByte = THKD bppCoinsPerUTxOByte
-    , cppCostModels = THKD bppCostModels
+    , cppCostModels =
```
lehins, line 1193: Could you please add a property test for Conway pparams upgrade that ensures that all fields have been upgraded along the lines of:

```haskell
propConwayPParamsUpgrade :: UpgradeConwayPParams Identity -> PParams Babbage -> Property
propConwayPParamsUpgrade ppu pp = property $ do
  let pp' = upgradePParams ppu pp :: PParams Conway
  pp' ^. ppPoolVotingThresholdsL `shouldBe` ucppPoolVotingThresholds ppu
  ...
  Map.lookup PlutusV3 (costModelsValid (pp' ^. ppCostModels)) `shouldBe` ucppPlutusV3CostModel ppu
```

libs/cardano-ledger-core/src/Cardano/Ledger/HKD.hs
```diff
@@ -62,3 +64,19 @@ instance HKDFunctor StrictMaybe where
   hkdMap _ = fmap
   toNoUpdate _ = NoUpdate
   fromNoUpdate _ = SNothing
+
+class HKDFunctor f => HKDApplicative f where
```
lehins, line 70: This needs an entry in the changelog

eras/conway/impl/src/Cardano/Ledger/Conway/PParams.hs
```diff
@@ -576,6 +591,7 @@ data UpgradeConwayPParams f = UpgradeConwayPParams
   , ucppDRepDeposit :: !(HKD f Coin)
   , ucppDRepActivity :: !(HKD f EpochInterval)
   , ucppMinFeeRefScriptCostPerByte :: !(HKD f NonNegativeInterval)
+  , ucppPlutusV3CostModel :: !(HKD f CostModel)
```
lehins, line 598: This needs an entry in the changelog and a major breaking change version bump

lehins (changes requested):
Couple of things need adjustment, but otherwise looks good.

eras/conway/impl/src/Cardano/Ledger/Conway/PParams.hs
aniketd: I think I was able to get rid of it for `Identity` :+1:

eras/conway/impl/src/Cardano/Ledger/Conway/Genesis.hs
```diff
@@ -98,6 +97,3 @@ toConwayGenesisPairs cg@(ConwayGenesis _ _ _ _ _) =
         ++ ["delegs" .= cgDelegs | not (null cgDelegs)]
         ++ ["initialDReps" .= cgInitialDReps | not (null cgInitialDReps)]
         ++ toUpgradeConwayPParamsUpdatePairs cgUpgradePParams
-
-instance Crypto c => Default (ConwayGenesis c) where
```
lehins, line 102: We should add removal of this instance to the changelog

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Genesis.hs
```diff
@@ -0,0 +1,177 @@
...
+                id
+                $ mkCostModel PlutusV3
+                $ replicate 233 0
+          }
+    }
+
+spec :: Spec
+spec =
+  describe "ConwayGenesis" $ do
+    it "Upgrades" $ property $ propConwayPParamsUpgrade
```
lehins: You don't have to change this, but just a more concise pointer for the future
```suggestion
    prop "Upgrades" $ propConwayPParamsUpgrade
```

libs/cardano-ledger-core/src/Cardano/Ledger/HKD.hs
```diff
@@ -14,9 +15,13 @@ module Cardano.Ledger.HKD (
   HKDNoUpdate,
   HKDFunctor (..),
   NoUpdate (..),
+  HKDApplicative (..),
 )
 where
-
+#if __GLASGOW_HASKELL__ == 964
+#else
```
lehins: The `__GLASGOW_HASKELL__` CPP pragma is not versioned the way you are thinking (or any other normal person would think)

[GHC versioning is `major.minor.patch`](https://ghc.gitlab.haskell.org/ghc/doc/users_guide/intro.html#ghc-version-numbering-policy)

* There is no patch version in the `__GLASGOW_HASKELL__`, so you can't match on the `4` in `9.6.4`, therefore you can only match on `9.6`
* The number in the middle is to account for minor version sometimes taking to digits.

So here are some examples:

* `ghc-9.2.x` - `__GLASGOW_HASKELL__ == 902`
* `ghc-8.10.x` - `__GLASGOW_HASKELL__ == 810`

```suggestion
#if __GLASGOW_HASKELL__ < 906
```

Also with current approach of `==` this pragma would break for ghc 9.8, so we need to use `<` instead

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Genesis.hs
```diff
@@ -0,0 +1,177 @@
...
+                          <> "mkCostModel PlutusV3 mempty cannot fail."
+                      )
+                )
+                id
+                $ mkCostModel PlutusV3
+                $ replicate 233 0
+          }
+    }
+
+spec :: Spec
```
lehins: This `spec` and `propConwayPParamsUpgrade` aren't really reusable, so they should just go into the test suite directly, instead of being exposed from the `testlib` library

eras/conway/test-suite/src/Test/Cardano/Ledger/Conway/Examples/Consensus.hs
```diff
@@ -202,5 +201,5 @@ exampleConwayNewEpochState =
     emptyPParams
     (emptyPParams & ppCoinsPerUTxOByteL .~ CoinPerByte (Coin 1))
 
-exampleConwayGenesis :: Crypto c => ConwayGenesis c
-exampleConwayGenesis = def
+exampleConwayGenesis :: ConwayGenesis StandardCrypto
+exampleConwayGenesis = expectedConwayGenesis
```
lehins, line 205: Note in the changelog that the value has changed, because that will affect consensus golden tests. Also requires a major version bump, since this is semantic change.

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Genesis.hs
```diff
@@ -0,0 +1,177 @@
...
+              either
+                ( error
+                    . const
+                      ( "Impossible! Implementation of Default for ConwayGenesis. "
+                          <> "mkCostModel PlutusV3 mempty cannot fail."
+                      )
+                )
+                id
+                $ mkCostModel PlutusV3
+                $ replicate 233 0
```
lehins: Use [`zeroTestingCostModelV3`](https://github.com/IntersectMBO/cardano-ledger/blob/81f11d9e2c562ee4ad0169e679011c765a1c21e7/libs/cardano-ledger-core/testlib/Test/Cardano/Ledger/Plutus.hs#L93) instead:

```suggestion
          , ucppPlutusV3CostModel = zeroTestingCostModelV3
```

lehins (changes requested):
You where able to remove the `Default` instance. That's good.
Few changes are still necessary

libs/cardano-ledger-core/src/Cardano/Ledger/HKD.hs
aniketd: :bow: Thank you! :heart_on_fire:

eras/conway/impl/CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Version history for `cardano-ledger-conway`
 
+## 1.14.0 0
+
+* Add `ucppPlutusV3CostModel` to `UpgradeConwayPParams` #4252
+  * Remove the `Default` instance for `ConwayGenesis`
+    * Because of this removal `Test.Cardano.Ledger.Conway.Examples.Consensus.exampleConwayGenesis` has been changed to use the value of `expectedConwayGenesis` that is equivalent to  `conway-genesis.json`
```
teodanciu: I think this explanation would be better in the commit message rather than the changelog. 
Or at least shortened , like  `updated  exampleConwayGenesis value to conway-genesis.json` , and moved to the `testlib` section?

eras/conway/impl/CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
...
+
 ## 1.13.1.0
 
 * Add `foldrVotingProcedures`
```
teodanciu: Would you mind moving this line  under `1.14.0 0` ?  The change is actually not in 1.13.1.0, it's just that after that release, the changelog file wasn't updated as it should have, which lead to it being then updated incorrectly.

eras/conway/impl/src/Cardano/Ledger/Conway/PParams.hs
```diff
@@ -1159,10 +1165,15 @@ instance FromJSON (UpgradeConwayPParams Identity) where
         <*> o .: "dRepDeposit"
         <*> o .: "dRepActivity"
         <*> o .: "minFeeRefScriptCostPerByte"
+        <*> do
+          cm <- o .: "plutusV3CostModel"
+          case mkCostModel PlutusV3 cm of
+            Left e -> fail $ show e
+            Right cm' -> pure cm'
```
teodanciu: You could rewrite this as 
`either (fail . show) pure (mkCostModel PlutusV3 cm)` if you fancied.

lehins (approved):
Once you address @teodanciu's comments this PR is ready to go.

eras/conway/impl/CHANGELOG.md
aniketd: Thanks! :raised_hands:

eras/conway/impl/CHANGELOG.md
aniketd: Got it! Moved! :raised_hands:

eras/conway/impl/src/Cardano/Ledger/Conway/PParams.hs
aniketd: Awesome thanks! :raised_hands:

PR #4253, opened 2024-04-10
Hardfork Initiation into a new era

### LINKED ISSUE #4006: HardFork into a new era

lehins:
## Fixes in this PR: 

Resolves #4006

* We have a unified approach to update PParams throughout all eras starting with Shelley,
  thus making HFC combinator much more robust and correct.
* We can remove duplicate logic from consensus that used to count up the genesis key holder votes: https://github.com/IntersectMBO/ouroboros-consensus/pull/1078
* We solve a problem where HFC is triggered in previous era, while TICK happens in the new
  era after translation. Which previously caused protocol version not being updated
  correctly in Conway, since protocol parameter update mechanism was vastly different from
  the one in Babbage.
* We finish implementation of HardForkInitiation into a new era.
* DRep pulser chunks are much more accurate now.

# Description

Outline of the new approach.

### Shelley through Babbage:

1. Genesis key holders submit votes with proposals for new `PParams`. Depending on the
   timeing within the epoch those votes will either go into the current or the future
   proposals bucket.
2. During the first `4k/f` slots in the the PPUP rule PParamUpdate proposals added to
   current proposals. Besides keeping the proposals we also now keep potential protocol
   parameters in the future pparms, but only quorum of genesis votes is reached. Potential
   values for new `PParams` can change if genesis key holders change their votes during
   this period, therefore they cannot yet be considered stable.
3. The very first `TICK` that happens during the last two stability windows before the end
   of the epoch we solidify the proposed `PParams`, thus ensuring they will be applied at
   the next epoch boundary. (See `solidifyNextEpochPParams`)
4. At the epoch boundary during the `NEWPP` rule, instead of counting the votes on the
   proposals, we just looked up in the next `PParams` that were decided earlier and apply
   them. All of the future votes that where potentially submitted before during the last
   two stability windows of past epoch are converted to current votes, which will be
   treated in the same way as if they were submitted in the 2nd step. This also resets the
   future PParams for the next epoch.

### Conway era forward

1. Either `ParameterChange` or `HardForkInitiation` proposal is submitted and votes are
   collected.
2. At the epoch boundary all of the votes and proposals are snapshotted.
3. Pulser starts the work on every `TICK` figuring out the stake distribution for `DReps`,
   calculating the votes and ratifying the proposals. Whenever `ParameterChange` or
   `HardForkInitiation` gets ratified in the `RATIFY` rule, the new values are immediately
   applied to the future enact state by the `ENACT` rule. Therefore we have future
   `PParams` at the latest two stability windows before the end of the epoch.
4. During the first `4k/f` slots on every tick we also lazily update future `PParams`.
5. Just as in Shelley era the very first `TICK` that happens during the last two stability
   windows (`6k/f`) before the end of the epoch we solidify the proposed `PParams`. (See
   `solidifyNextEpochPParams`). Unlike previous eras, in Conway this step is safe to do at
   any point during the initial part of the epoch, because they are considered stable as
   soon as we enter new epoch, however they are expensive to compute during that period,
   that is why we solidify them only when we are pretty confident that the DRep pulser is
   done and `RATIFY` with `ENACT` rules got a chance to be executed.
6. At the epoch boundary we apply the new PParams that where solidified in the previous
   step and reset the future pparams, thus making it ready for the next epoch. The
   important part here is that we do not use the values from the Enact state directly, but
   we take the futurePParams as the source of truth. This allows us to correctly update
   the `PParams` not only using the voting process of Conway era, but allows us to apply
   the `PParams` update from Babbage era.

### Forecast

It is very important that the TICKF rule does the same steps as the TICK and EPOCH (for
Conway) or NEWPP (for pre Conway) rules. In particular same solidification and rotation of
pparams process as in the stepsabove should happen for forcasting to work correctly.

# Checklist

- [ ] Self-reviewed the diff


eras/shelley/impl/src/Cardano/Ledger/Shelley/Governance.hs
```diff
@@ -87,6 +89,18 @@ class
   -- | Lens for accessing the previous protocol parameters
   prevPParamsGovStateL :: Lens' (GovState era) (PParams era)
 
+  -- | Getter for accessing the future protocol parameters.
+  --
+  -- This getter is only reliable and efficient 2 stability before the end of the
```
amesgen: "reliable" is maybe too ambiguous here, maybe "stable"?

eras/shelley/impl/src/Cardano/Ledger/Shelley/Governance.hs
```diff
@@ -87,6 +89,18 @@ class
...
+  -- epoch. Depending on the era, if called earlier in the epoch, it will either produce
+  -- unreliable results or getting those results will be somewhat costly.
+  --
+  -- Whenever called at the earliest two stability before the end of the epoch, then the
+  -- results will be 100% reliable and they will contain either a `Just` value with the
+  -- new `PParams`, when there was an update proposed and `Nothing` whenever PParams will
```
amesgen: "proposed" does not necessarily carry the connotation that it was accepted/reached quorum?

lehins: I am not quite sure if your question is about implementation or about the haddock?

With respect to implementation, it does carry this information that it has reached the quorum, otherwise it will be `Nothing`. The only issue here in pre Conway eras is that it can only be reliable 2 stability windows before the end of the epoch, because any genesis vote can change the proposal before then.

With respect to the haddock. This getter lens produces PParams that will be in the future epoch and it only does that reliably at the earliest two stability windows before the end of the epoch. The "proposed" update does not refer here to a specific mechanism of how it is done (which is very different in Shelley and Conway), but to the fact that there was an accepted proposal to update PParams irrespective of how it was done, but it will take in affect at the epoch boundary.

So, if this explanation makes it a bit clearer, I can include it into the haddock. If you have better ideas on how to word it, I am definitely open to suggestions.

eras/shelley/impl/src/Cardano/Ledger/Shelley/Governance.hs
lehins: I think it is a perfect wording here precisely because it is ambiguous. It means if you call this function earlier then 2 stability before the end of the epoch, you cannot rely on the answer. It is unreliable in the initial part of the epoch by design, so people don't try to invoke it.

amesgen: FTR: Judging from https://github.com/IntersectMBO/ouroboros-consensus/pull/1078#discussion_r1581565528, the goal is now to make it "reliable" everywhere?

eras/shelley/impl/src/Cardano/Ledger/Shelley/Governance.hs
amesgen: It is only about the Haddocks, thanks for clarifying that "proposed" here means "accepted"/"reached the quorum". I think it indeed makes sense to include that more detailed explanation in the Haddocks, thanks :+1:

eras/shelley/impl/src/Cardano/Ledger/Shelley/Governance.hs
```diff
@@ -87,6 +98,16 @@ class
   -- | Lens for accessing the previous protocol parameters
   prevPParamsGovStateL :: Lens' (GovState era) (PParams era)
 
+  -- | Lens for accessing the future protocol parameters.
+  --
+  -- This lens will produce `DefinitePParamsUpdate` whenever we are absolutely sure that
+  -- the new PParams will be updated. Which means there will be no chance of a
+  -- `DefinitePParamsUpdate` value until we are past the point of no return, which is 2
+  -- stability windows before the end of the epoch. This lens is mostly intended for
+  -- ledger usage and `nextEpochUpdatedPParams` should be used instead whenver definite
```
aniketd:
```suggestion
  -- ledger usage and `nextEpochUpdatedPParams` should be used instead whenever definite
```

eras/shelley/impl/src/Cardano/Ledger/Shelley/Governance.hs
```diff
@@ -98,16 +119,95 @@ instance Crypto c => EraGov (ShelleyEra c) where
...
+-- by ledger at any point that is two stability windows before the end of the epoch. If
+-- you need to know if there were actual changes to those PParams then use
+-- `nextEpochUpdatedPParams` instead.
+nextEpochPParams :: EraGov era => GovState era -> PParams era
+nextEpochPParams govState =
+  fromSMaybe (govState ^. curPParamsGovStateL) $ nextEpochUpdatedPParams govState
+
+-- | This function is guaranteed to return updated PParams when it is called during the
+-- last two stability windows of the epoch and there were proposals to update PParams that
+-- all relevnat parties reached consensus on. In other words whenver there is a definite
```
aniketd:
```suggestion
-- all relevant parties reached consensus on. In other words whenever there is a definite
```

Lucsanszky (commented):
Had a first read through but I will definitely have to go through it again to fully understand it. It looks good to me so far but I will wait for the others to approve.

eras/shelley/impl/src/Cardano/Ledger/Shelley/Governance.hs
lehins: Yep, made it reliable everywhere

eras/shelley/impl/src/Cardano/Ledger/Shelley/Governance.hs
lehins: Replaced the implementation and documentation. Should be much more clear now.

eras/shelley/impl/src/Cardano/Ledger/Shelley/Governance.hs
```diff
@@ -98,16 +119,95 @@ instance Crypto c => EraGov (ShelleyEra c) where
...
+solidifyFuturePParams :: FuturePParams era -> FuturePParams era
+solidifyFuturePParams = \case
+  -- Here we convert a potential to a definite update:
+  PotentialPParamsUpdate Nothing -> NoPParamsUpdate
+  PotentialPParamsUpdate (Just pp) -> DefinitePParamsUpdate pp
+  fpp -> fpp
+
+deriving stock instance Eq (PParams era) => Eq (FuturePParams era)
+deriving stock instance Show (PParams era) => Show (FuturePParams era)
+instance NoThunks (PParams era) => NoThunks (FuturePParams era)
```
teodanciu: Should Thunks not be allowed for `Maybe (PParams era)` in  `PotentialPParamsUpdate` ?

docs/adr/2024-04-30_008-pparams-update.md
```diff
@@ -0,0 +1,83 @@
...
+We need an unified approach for predicting and updating protocol parameters.
+
+Outline of the new approach.
+
+### Shelley through Babbage:
+
+1. Genesis key holders submit votes with proposals for new `PParams`. Depending on the
+   timeing within the epoch those votes will either go into the current or the future
+   proposals bucket.
+2. During the first `4k/f` slots in the the PPUP rule PParamUpdate proposals added to
```
teodanciu, line 24: I'm confused between the `*k/f` and `stabilityWindows` terminology. 

`During the first 4k/f slots`  - is this what I see in the code to be: 
`if slot < tooLate`, where `tooLate` is computed  with `getTheSlotOfNoReturn`    (for which I understand the relationship with stability window, thanks to its comprehensive haddock :pray: )

docs/adr/2024-04-30_008-pparams-update.md
```diff
@@ -0,0 +1,83 @@
...
+
+1. Either `ParameterChange` or `HardForkInitiation` proposal is submitted and votes are
+   collected.
+2. At the epoch boundary all of the votes and proposals are snapshotted.
+3. Pulser starts the work on every `TICK` figuring out the stake distribution for `DReps`,
+   calculating the votes and ratifying the proposals. Whenever `ParameterChange` or
+   `HardForkInitiation` gets ratified in the `RATIFY` rule, the new values are immediately
+   applied to the future enact state by the `ENACT` rule. Therefore we have future
+   `PParams` at the latest two stability windows before the end of the epoch.
+4. During the first `4k/f` slots on every tick we also lazily update future `PParams`.
```
teodanciu, line 49: Just for my understanding -  where is this in the code? It is not the `futurePParamsGovStateL %~ solidifyFuturePParams`  from `solidifyNextEpochPParams`  I suppose? (since that's described at the next step)

teodanciu (approved):
Just a few questions from me. 
This PR is so dense :exploding_head: 
Thank you for the comprehensive explanations in the haddocks and in the  ADR!

eras/shelley/impl/src/Cardano/Ledger/Shelley/Governance.hs
lehins: Nice catch. I switched it to deriving through `AllowThunk`

docs/adr/2024-04-30_008-pparams-update.md
lehins: This is a bit confusing, I agree. It took me a while to figure it out and hopefully I did understanding it correctly.

Here is the calculation for the stability window:
https://github.com/IntersectMBO/cardano-ledger/blob/63ddaf6f3102e06d26ac2c69b2d8530d83d26a19/eras/shelley/impl/src/Cardano/Ledger/Shelley/StabilityWindow.hs#L10-L22

So, one stability window is `ceiling (3k/f)` number of slots. Two stability windows before the end of the epoch is: `ceiling (3k/f) + ceiling (3k/f)`. Because there is a total of `10k/f` slots in a the epoch the first `n` slots, where `n = 10k/f - 2 * ceiling (3k/f)` slots, which makes `n =< 4k/f` the period that is before the `tooLate` or what I call the point of no return.

docs/adr/2024-04-30_008-pparams-update.md
lehins: We do the step 4 (lazy update of the future PParams):
* in Shelley:
  * in the PPUP rule: https://github.com/IntersectMBO/cardano-ledger/blob/54dc232375f007322312eb9b766d5abab6b00c4f/eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Ppup.hs#L205-L206
  * and in NEWPP rule: https://github.com/IntersectMBO/cardano-ledger/blob/54dc232375f007322312eb9b766d5abab6b00c4f/eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Newpp.hs#L118-L119
* in Conway we do it in the NEWEPOCH rule: https://github.com/IntersectMBO/cardano-ledger/blob/54dc232375f007322312eb9b766d5abab6b00c4f/eras/conway/impl/src/Cardano/Ledger/Conway/Rules/NewEpoch.hs#L178

Does this answer your question?

docs/adr/2024-04-30_008-pparams-update.md
teodanciu: Thank you, this makes sense!  I  didn't know about the total of `10k/f` slots.

docs/adr/2024-04-30_008-pparams-update.md
teodanciu: Absolutely, thank you!

PR #4259, opened 2024-04-11
Undeprecate redeemerPointer and expose it in cardano-ledger-api

ch1bo:
# Description

User code may still rely on this as there is no other "off-the-shelf" conversion from an AsItem -> AsIx. As the alternative would be to implement this index lookup (e.g. for AlonzoSpending purpose) in user applications, we decide to rather keep it and expose it on the cardano-ledger-api.

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/libs/cardano-ledger-api/cardano-ledger-api.cabal
+++ b/libs/cardano-ledger-api/cardano-ledger-api.cabal
@@ -1,6 +1,6 @@
 cabal-version:      3.0
 name:               cardano-ledger-api
-version:            1.9.0.0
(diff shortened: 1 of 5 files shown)


libs/cardano-ledger-api/cardano-ledger-api.cabal
ch1bo, line 1: Seems my `cabal-fmt` kicked in. Shall I undo these or are you also using `cabal-fmt`?

lehins (approved):
:+1: Thank you

PR #4261, opened 2024-04-12
`constrained-generators` cleanup for hackage

MaximilianAlgehed:
# Description

This is also in preparation for moving the specs for various STS rules out of `cardano-ledger-test`. @lehins, @Soupstraw, and @TimSheard what are your wishes for this? I have the following TODO that I can recall:

- [ ] Type class for specs used in STS rules (@Soupstraw did you end up doing that in the last PR or not?)
- [ ] Move the STS specs out of `cardano-ledger-test`
- [x] Rename the `Spec` type so that it doesn't clash with `HSpec.Spec`
- [x] Migrate away from `tasty` to `hspec`

# Checklist

- [x] Self-reviewed the diff


lehins:
> I have the following TODO that I can recall:

I don't think I can add anything else to the TODO list at the moment. The plan sounds good to me.

MaximilianAlgehed:
In order to focus on fixing a bug for @TimSheard I've decided to push the top two TODOs for the next PR. Could we get this reviewed and merged ASAP?

libs/constrained-generators/src/Constrained/Base.hs
```diff
@@ -2326,17 +2332,17 @@ genNumList ::
...
   normFoldS <- normalize foldSIn
   explain
     [ "Can't generate list of ints with fold constraint"
-    , show $ "  elemSpec =" <+> pretty elemSIn
-    , show $ "  normElemSpec =" <+> pretty normElemS
-    , show $ "  foldSpec =" <+> pretty foldSIn
+    , "  elemSpec = " ++ show elemSIn
+    , "  normElemSpec = " ++ show normElemS
+    , "  foldSpec = " ++ show foldSIn
     ]
```
TimSheard, line 2397: Not a good idea in my opinion to change Pretty to Show.

TimSheard (changes requested):
There seem to be 4 kinds of changes in this PR
1) Changing the type Spec to Specification
2) Adding a bit of new functionality to Conway.Instances
3) Moving from Pretty to Show
4) Splitting the tests into multiple modules. One for List, Set, Map ,Property etc.

Vast majority of changes (> 90%) involve the Spec to Specification

In my opinion moving from Pretty to Show is ill-advised, and should be reverted.

libs/constrained-generators/src/Constrained/Base.hs
MaximilianAlgehed: This is just a refactoring:
```haskell
instance HasSpec fn a => Show (Specification fn a) where
  showsPrec d = shows . pretty . WithPrec d
```

lehins (approved):
Looks awesome!

Here is a side comment to what @TimSheard said:

> In my opinion moving from Pretty to Show is ill-advised, and should be reverted.

I am totally in favor of **not** using the custom Pretty instances. Pretty has huge downsides for the purposes of testing and debugging:
* it is very confusing, because the output is free form and without looking at the actual instance it is hard to infer the meaning of the output or even the type that is being printed.
* it has huge maintenance overhead. Every time any of the ledger types changes we need to go and manually adjust the instance.
* a bug in Pretty instance could lead to either a buggy test or at the very least hard time during debugging of an actual issue.
* Pretty instances themselves are untested and even right now bunch of the instances are incomplete.

My goal is one day to completely get rid of Pretty interface in the ledger codebase.

TimSheard:
We are not paying attention here.  The only Pretty Instances I want to keep are the ones for 
Term, Spec, Pred, etc.  These are the types native to the this package alone. These things can get large
and when a Spec fails, being able to understand what is wrong is helped by the pretty printing. I agree with Alexey, we don't need any Pretty instance for Ledger types.

MaximilianAlgehed:
I agree on the ledger types. For the constraint types you do need pretty instances that make things readable (you don't want to read abstract syntax trees!)

Anyway, @TimSheard could you approve the PR? The issue you were concerned about is moot because the show instance for Spec is just the pretty instance.

TimSheard (approved):
OK I approve.

Issue #4085, opened 2024-02-15
Collection of things to be cleaned up after release of Conway 


lehins:
Things to cleanup:

* [x] - #5260 - Removal of `TriesToForgeADA` for Allegra https://github.com/IntersectMBO/cardano-ledger/blob/b71cb8b155268fce6ac7cf166ae0e485579d264e/eras/allegra/impl/src/Cardano/Ledger/Allegra/Rules/Utxo.hs#L112

* [x] - #5260 - Removal of `TriesToForgeADA` for Alonzo https://github.com/IntersectMBO/cardano-ledger/blob/b71cb8b155268fce6ac7cf166ae0e485579d264e/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Rules/Utxo.hs#L163-L164
* [x] - Remove complex check in plutus evaluation https://github.com/IntersectMBO/cardano-ledger/blob/b71cb8b155268fce6ac7cf166ae0e485579d264e/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Plutus/Evaluate.hs#L153-L156
* [ ] - Fix injection to use `OutputBootAddrAttrsTooBig` https://github.com/IntersectMBO/cardano-ledger/blob/b71cb8b155268fce6ac7cf166ae0e485579d264e/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Rules/Utxo.hs#L740
* [ ] - Remove all predicate failures for rules that run on the epoch boundary and replace their `PredicateFailure` with `VoidEraRule`: https://github.com/IntersectMBO/cardano-ledger/issues/4158
* [x] - #5260 - Change `OutputTooBigUTxO` predicate failure to contain `Int`, instead of `Integer`: https://github.com/IntersectMBO/cardano-ledger/blob/2c9ba10d085331fa99840a63b40b8e8b716fba5d/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Rules/Utxo.hs#L163-L164
* [x] - #5260 - Ensure no Imp test is executed with `withCborRoundTripFailures` and delete this helper function. Serialization for older predicate failures can be fixed now

lehins:
Also #3972 falls into the same category

PR #4265, opened 2024-04-15
Inline UTxO and UTxOW `PredFailure` for `Conway`

### LINKED ISSUE #4170: Stop embedding predicate failures from previous eras
We have a problem in Babbage and consequently in Conway where predicate failure for UTXO and consequently for UTXOS can be injected in two different ways. This is quite dangerous and is most definitely a bug. We can't fix it right now for Babbage, until we hard fork into Conway. However, we must fix it for Conway, otherwise we'll be in the same boat in Conway as we are in Babbage.

Here is a branch that defines the two predicate failures: https://github.com/IntersectMBO/cardano-ledger/tree/inline-conway-predicate-failures

Here are two PRs that do a very similar thing, but for UTXOS rule. Please use it as an example: #4150 and #4168

Injection of failures from previous eras should still work as before.

Lucsanszky:
# Description
Resolves #4170 

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/eras/conway/impl/src/Cardano/Ledger/Conway/Era.hs
+++ b/eras/conway/impl/src/Cardano/Ledger/Conway/Era.hs
@@ -11,6 +11,7 @@ module Cardano.Ledger.Conway.Era (
   ConwayNEWEPOCH,
   ConwayEPOCH,
   ConwayENACT,
+  ConwayUTXO,
   ConwayUTXOS,
   ConwayUTXOW,
   ConwayTICKF,
@@ -20,7 +21,6 @@ module Cardano.Ledger.Conway.Era (
 
 import Cardano.Ledger.Alonzo.Rules (AlonzoBBODY)
 import Cardano.Ledger.Babbage (BabbageEra)
-import Cardano.Ledger.Babbage.Rules (BabbageUTXO)
 import Cardano.Ledger.Core
 import Cardano.Ledger.Crypto (Crypto)
 import Cardano.Ledger.Mary.Value (MaryValue)
@@ -127,9 +127,9 @@ data ConwayUTXOW era
 
 type instance EraRule "UTXOW" (ConwayEra c) = ConwayUTXOW (ConwayEra c)
 
--- Rules inherited from Babbage
+data ConwayUTXO era
 
-type instance EraRule "UTXO" (ConwayEra c) = BabbageUTXO (ConwayEra c)
+type instance EraRule "UTXO" (ConwayEra c) = ConwayUTXO (ConwayEra c)
 
 -- Rules inherited from Alonzo
 
--- a/eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Utxo.hs
+++ b/eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Utxo.hs
@@ -1,58 +1,402 @@
 {-# LANGUAGE DataKinds #-}
+{-# LANGUAGE DeriveGeneric #-}
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE MultiParamTypeClasses #-}
+{-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE StandaloneDeriving #-}
 {-# LANGUAGE TupleSections #-}
 {-# LANGUAGE TypeApplications #-}
 {-# LANGUAGE TypeFamilies #-}
 {-# LANGUAGE TypeOperators #-}
+{-# LANGUAGE UndecidableInstances #-}
 {-# OPTIONS_GHC -Wno-orphans #-}
 
-module Cardano.Ledger.Conway.Rules.Utxo (allegraToConwayUtxoPredFailure) where
+module Cardano.Ledger.Conway.Rules.Utxo (
+  allegraToConwayUtxoPredFailure,
+  babbageToConwayUtxoPredFailure,
+  alonzoToConwayUtxoPredFailure,
+  ConwayUtxoPredFailure (..),
+) where
 
-import Cardano.Ledger.Allegra.Rules (shelleyToAllegraUtxoPredFailure)
+import Cardano.Ledger.Address (Addr, RewardAccount)
+import Cardano.Ledger.Allegra.Rules (AllegraUtxoPredFailure, shelleyToAllegraUtxoPredFailure)
 import qualified Cardano.Ledger.Allegra.Rules as Allegra (AllegraUtxoPredFailure (..))
 import Cardano.Ledger.Alonzo.Rules (
   AlonzoUtxoEvent,
-  AlonzoUtxoPredFailure (..),
+  AlonzoUtxoPredFailure,
   AlonzoUtxosPredFailure,
  )
-import Cardano.Ledger.Babbage.Rules (BabbageUtxoPredFailure (..))
-import Cardano.Ledger.Conway.Era (ConwayEra)
-import Cardano.Ledger.Conway.Rules.Utxos (ConwayUtxosPredFailure)
-import Cardano.Ledger.Core
+import qualified Cardano.Ledger.Alonzo.Rules as Alonzo (
+  AlonzoUtxoEvent (UtxosEvent),
+  AlonzoUtxoPredFailure (..),
+ )
+import Cardano.Ledger.Alonzo.Tx (AlonzoTx (..))
+import Cardano.Ledger.Babbage.Rules (BabbageUtxoPredFailure)
+import qualified Cardano.Ledger.Babbage.Rules as Babbage (
+  BabbageUtxoPredFailure (..),
+  utxoTransition,
+ )
+import Cardano.Ledger.BaseTypes (Network, ShelleyBase, SlotNo)
+import Cardano.Ledger.Binary (DecCBOR (..), EncCBOR (..))
+import Cardano.Ledger.Binary.Coders (
+  Decode (..),
+  Encode (..),
+  decode,
+  encode,
+  (!>),
+  (<!),
+ )
+import Cardano.Ledger.Coin (Coin, DeltaCoin)
+import Cardano.Ledger.Conway.Core
+import Cardano.Ledger.Conway.Era (ConwayEra, ConwayUTXO, ConwayUTXOS)
+import Cardano.Ledger.Conway.Rules.Utxos (
+  ConwayUtxosPredFailure (..),
+ )
+import Cardano.Ledger.Plutus (ExUnits)
+import qualified Cardano.Ledger.Shelley.LedgerState as Shelley (UTxOState)
 import Cardano.Ledger.Shelley.Rules (ShelleyUtxoPredFailure)
+import qualified Cardano.Ledger.Shelley.Rules as Shelley (UtxoEnv)
+import Cardano.Ledger.TxIn (TxIn)
+import Cardano.Ledger.UTxO (EraUTxO, UTxO (..))
+import Control.DeepSeq (NFData)
+import Control.State.Transition.Extended (
+  Embed (..),
+  STS (..),
+ )
+import Data.List.NonEmpty (NonEmpty)
+import Data.Set (Set)
+import GHC.Generics (Generic)
+import GHC.Natural (Natural)
 
-type instance EraRuleFailure "UTXO" (ConwayEra c) = BabbageUtxoPredFailure (ConwayEra c)
+-- ======================================================
+
+-- | Predicate failure for the Conway Era
+data ConwayUtxoPredFailure era
+  = -- | Subtransition Failures
+    UtxosFailure (PredicateFailure (EraRule "UTXOS" era))
+  | -- | The bad transaction inputs
+    BadInputsUTxO
+      !(Set (TxIn (EraCrypto era)))
+  | OutsideValidityIntervalUTxO
+      -- | transaction's validity interval
+      !ValidityInterval
+      -- | current slot
+      !SlotNo
+  | MaxTxSizeUTxO
+      -- | the actual transaction size
+      !Integer
+      -- | the max transaction size
+      !Integer
+  | InputSetEmptyUTxO
+  | FeeTooSmallUTxO
+      -- | the minimum fee for this transaction
+      !Coin
+      -- | the fee supplied in this transaction
+      !Coin
+  | ValueNotConservedUTxO
+      -- | the Coin consumed by this transaction
+      !(Value era)
+      -- | the Coin produced by this transaction
+      !(Value era)
+  | -- | the set of addresses with incorrect network IDs
+    WrongNetwork
+      -- | the expected network id
+      !Network
+      -- | the set of addresses with incorrect network IDs
+      !(Set (Addr (EraCrypto era)))
+  | WrongNetworkWithdrawal
+      -- | the expected network id
+      !Network
+      -- | the set of reward addresses with incorrect network IDs
+      !(Set (RewardAccount (EraCrypto era)))
+  | -- | list of supplied transaction outputs that are too small
+    OutputTooSmallUTxO
+      ![TxOut era]
+  | -- | list of supplied bad transaction outputs
+    OutputBootAddrAttrsTooBig
+      ![TxOut era]
+  | -- | list of supplied bad transaction output triples (actualSize,PParameterMaxValue,TxOut)
+    OutputTooBigUTxO
+      ![(Int, Int, TxOut era)]
+  | InsufficientCollateral
+      -- | balance computed
+      !DeltaCoin
+      -- | the required collateral for the given fee
+      !Coin
+  | -- | The UTxO entries which have the wrong kind of script
+    ScriptsNotPaidUTxO
+      !(UTxO era)
+  | ExUnitsTooBigUTxO
+      -- | Max EXUnits from the protocol parameters
+      !ExUnits
+      -- | EXUnits supplied
+      !ExUnits
+  | -- | The inputs marked for use as fees contain non-ADA tokens
+    CollateralContainsNonADA !(Value era)
+  | -- | Wrong Network ID in body
+    WrongNetworkInTxBody
+      -- | Actual Network ID
+      !Network
+      -- | Network ID in transaction body
+      !Network
+  | -- | slot number outside consensus forecast range
+    OutsideForecast
+      !SlotNo
+  | -- | There are too many collateral inputs
+    TooManyCollateralInputs
+      -- | Max allowed collateral inputs
+      !Natural
+      -- | Number of collateral inputs
+      !Natural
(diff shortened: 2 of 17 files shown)


eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Utxo.hs
```diff
@@ -65,5 +251,273 @@ allegraToConwayUtxoPredFailure = \case
...
+  Alonzo.MaxTxSizeUTxO x y -> MaxTxSizeUTxO x y
+  Alonzo.InputSetEmptyUTxO -> InputSetEmptyUTxO
+  Alonzo.FeeTooSmallUTxO c1 c2 -> FeeTooSmallUTxO c1 c2
+  Alonzo.ValueNotConservedUTxO vc vp -> ValueNotConservedUTxO vc vp
+  Alonzo.WrongNetwork x y -> WrongNetwork x y
+  Alonzo.WrongNetworkWithdrawal x y -> WrongNetworkWithdrawal x y
+  Alonzo.OutputTooSmallUTxO x -> OutputTooSmallUTxO x
+  Alonzo.UtxosFailure x -> UtxosFailure x
+  Alonzo.OutputBootAddrAttrsTooBig xs -> OutputBootAddrAttrsTooBig xs
+  Alonzo.TriesToForgeADA -> error "How should I deal with this?"
```
Lucsanszky: See above

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Utxo.hs
```diff
@@ -65,5 +251,273 @@ allegraToConwayUtxoPredFailure = \case
   Allegra.OutputTooSmallUTxO x -> OutputTooSmallUTxO x
   Allegra.UpdateFailure x -> absurdEraRule @"PPUP" @era x
   Allegra.OutputBootAddrAttrsTooBig xs -> OutputBootAddrAttrsTooBig xs
-  Allegra.TriesToForgeADA -> TriesToForgeADA
+  Allegra.TriesToForgeADA -> error "How should I deal with this?"
```
Lucsanszky: Is this okay or should I try to come up with something along the lines of `absurdEraRule?`

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Utxow.hs
```diff
@@ -118,21 +212,137 @@ instance
...
+  forall era.
+  AlonzoUtxowPredFailure era ->
+  ConwayUtxowPredFailure era
+alonzoToConwayUtxowPredFailure = \case
+  Alonzo.ShelleyInAlonzoUtxowPredFailure f -> shelleyToConwayUtxowPredFailure f
+  Alonzo.MissingRedeemers rs -> MissingRedeemers rs
+  Alonzo.MissingRequiredDatums mds rds -> MissingRequiredDatums mds rds
+  Alonzo.NotAllowedSupplementalDatums uds ads -> NotAllowedSupplementalDatums uds ads
+  Alonzo.PPViewHashesDontMatch a b -> PPViewHashesDontMatch a b
+  Alonzo.MissingRequiredSigners _xs -> error "How should I deal with this?" -- TODO: remove once in Conway. It is now redundant. See #3972
```
Lucsanszky: See above

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Utxow.hs
```diff
@@ -118,21 +212,137 @@ instance
...
+  Alonzo.ExtraRedeemers xs -> ExtraRedeemers xs
+
+shelleyToConwayUtxowPredFailure :: ShelleyUtxowPredFailure era -> ConwayUtxowPredFailure era
+shelleyToConwayUtxowPredFailure = \case
+  Shelley.InvalidWitnessesUTXOW xs -> InvalidWitnessesUTXOW xs
+  Shelley.MissingVKeyWitnessesUTXOW xs -> MissingVKeyWitnessesUTXOW xs
+  Shelley.MissingScriptWitnessesUTXOW xs -> MissingScriptWitnessesUTXOW xs
+  Shelley.ScriptWitnessNotValidatingUTXOW xs -> ScriptWitnessNotValidatingUTXOW xs
+  Shelley.UtxoFailure x -> UtxoFailure x
+  Shelley.MIRInsufficientGenesisSigsUTXOW _xs -> error "How should I deal with this?"
```
Lucsanszky: See above

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Utxo.hs
```diff
@@ -65,5 +251,273 @@ allegraToConwayUtxoPredFailure = \case
...
+  type Signal (ConwayUTXO era) = AlonzoTx era
+  type Environment (ConwayUTXO era) = UtxoEnv era
+  type BaseM (ConwayUTXO era) = ShelleyBase
+  type PredicateFailure (ConwayUTXO era) = ConwayUtxoPredFailure era
+  type Event (ConwayUTXO era) = AlonzoUtxoEvent era
+
+  initialRules = []
+
+  -- TODO: It would be great to be able to use the Alonzo `utxoTransition` here
+  -- i.e: transitionRules = [Alonzo.utxoTransition @era]
```
Lucsanszky: Actually, the `UTXOW` transitions allow this because unlike in `utxoTransition`s, `utxowTransition`s doesn't concretise `EraRule "UTXO" era`. These only get concretised for the instances in `UTXOW`s case which allows the reuse of the transition functions. Would it be possible or okay to modify the `utxoTransition` in `Alonzo` and remove `EraRule "UTXO" era ~ AlonzoUTXO era` so I can reuse that in `Conway`s `utxoTransition`?

Lucsanszky (commented):
I added some comments to facilitate review. Also, I intend to clean up the imports and reorganise some of my added changes.

teodanciu:
Is there a ticket linked to this PR, for a little context?

Lucsanszky:
> Is there a ticket linked to this PR, for a little context?

Sure! I've just updated the description and linked the issue.

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Utxow.hs
lehins:
```suggestion
  Alonzo.MissingRequiredSigners _xs -> 
    error "Impossible case. It will be removed once we are in Conway. See #3972"
```

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Utxow.hs
lehins: For now the only thing we can do is throw an impossible error. There might be a possibility that we can use some type level trickery using GADTs or something to prove that this case is indeed impossible, but for now let's not worry about it, cause this case is indeed impossible. There is no instance for `ShelleyEraTxBody` for Conway, so [`validateMIRInsufficientGenesisSigs`](https://github.com/IntersectMBO/cardano-ledger/blob/2c9ba10d085331fa99840a63b40b8e8b716fba5d/eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Utxow.hs#L524) cannot be invoked.
```suggestion
  Shelley.MIRInsufficientGenesisSigsUTXOW _xs ->
    error "Impossible: MIR has been removed in Conway"
```

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Utxo.hs
```diff
@@ -65,5 +251,273 @@ allegraToConwayUtxoPredFailure = \case
...
+  {- txsize tx ≤ maxTxSize pp -}
+  runTestOnSignal $ Shelley.validateMaxTxSizeUTxO pp tx
+
+  {-   totExunits tx ≤ maxTxExUnits pp    -}
+  runTest $ Alonzo.validateExUnitsTooBigUTxO pp tx
+
+  {-   ‖collateral tx‖  ≤  maxCollInputs pp   -}
+  runTest $ Alonzo.validateTooManyCollateralInputs pp txBody
+
+  trans @(EraRule "UTXOS" era) =<< coerce <$> judgmentContext
```
lehins: The whole point of this predicate injection mechanism is so we can reuse the logic as much as possible. In this case the `UTXO` rule is exactly the same so we can reuse it as-is from Babbage
```suggestion
```

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Utxo.hs
lehins: Yes, setting it to `error "impossible.."` is fine. Adding a note about #4085 would be a good idea too.

eras/conway/impl/src/Cardano/Ledger/Conway/Era.hs
```diff
@@ -127,9 +127,9 @@ data ConwayUTXOW era
 
 type instance EraRule "UTXOW" (ConwayEra c) = ConwayUTXOW (ConwayEra c)
 
--- Rules inherited from Babbage
+data ConwayUTXO era
```
lehins, line 130: You should not be defining the new rule, it is exactly the same as in Babbage, so we can completely reuse it.

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Utxow.hs
```diff
@@ -63,20 +85,81 @@ conwayWitsVKeyNeeded = getConwayWitsVKeyNeeded
...
+    MalformedReferenceScripts
+      !(Set (ScriptHash (EraCrypto era)))
+  deriving (Generic)
+
+type instance EraRuleFailure "UTXOW" (ConwayEra c) = ConwayUtxowPredFailure (ConwayEra c)
 
 type instance EraRuleEvent "UTXOW" (ConwayEra c) = AlonzoUtxowEvent (ConwayEra c)
 
-instance InjectRuleFailure "UTXOW" BabbageUtxowPredFailure (ConwayEra c)
+instance InjectRuleFailure "UTXOW" BabbageUtxowPredFailure (ConwayEra c) where
```
lehins, line 159: With addition of a new predicate failure data type we need `InjectFailure` instances for all possible injections in their respective modules for that type for every rule that can inject it:

* `instance InjectRuleFailure "UTXOW" ConwayUtxowPredFailure (ConwayEra c)`
* `instance InjectRuleFailure "LEDGER" ConwayUtxowPredFailure (ConwayEra c)`
* `instance InjectRuleFailure "LEDGERS" ConwayUtxowPredFailure (ConwayEra c)`
* `instance InjectRuleFailure "BBODY" ConwayUtxowPredFailure (ConwayEra c)`

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Utxo.hs
```diff
@@ -1,58 +1,244 @@
...
+    OutputTooSmallUTxO
+      ![TxOut era]
+  | -- | Subtransition Failures
+    UtxosFailure (PredicateFailure (EraRule "UTXOS" era))
+  | -- | list of supplied bad transaction outputs
+    OutputBootAddrAttrsTooBig
+      ![TxOut era]
+  | -- | list of supplied bad transaction output triples (actualSize,PParameterMaxValue,TxOut)
+    OutputTooBigUTxO
+      ![(Integer, Integer, TxOut era)]
```
lehins: Let's change those to an `Int` instead. There is really no reason for an arbitrary precision `Integer` in here. Note that I added a point to #4085 about doing the same for previous eras, but we'll do that past Conway. This means that you'll have some conversion from `Integer` -> `Int` during injection. Make sure to make a note about this upcoming change and make the conversion monophonic with `fromIntegral :: Integer -> Int`. This way once we change the `OutputTooBigUTxO` from the previous era we will not forget to remove this note and the conversion.
```suggestion
      ![(Int, Int, TxOut era)]
```

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Utxo.hs
```diff
@@ -65,5 +251,273 @@ allegraToConwayUtxoPredFailure = \case
...
+  5 -> SumD ValueNotConservedUTxO <! From <! From
+  6 -> SumD WrongNetwork <! From <! From
+  7 -> SumD WrongNetworkWithdrawal <! From <! From
+  8 -> SumD OutputTooSmallUTxO <! From
+  9 -> SumD UtxosFailure <! From
+  10 -> SumD OutputBootAddrAttrsTooBig <! From
+  11 ->
+    let fromRestricted :: (Int, Int, TxOut era) -> (Integer, Integer, TxOut era)
+        fromRestricted (sz, mv, txOut) = (toInteger sz, toInteger mv, txOut)
+     in SumD OutputTooBigUTxO <! D (map fromRestricted <$> decCBOR)
```
lehins: With my suggestion about a change to the `OutputTooBigUTxO` type
```suggestion
  11 -> SumD OutputTooBigUTxO <! From
```

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Utxo.hs
lehins: You should not be defining `utxoTransition` for `Conway` nor the `ConwayUTXO` rule type. It can all be reused from Babbage. That is the problem here.

lehins (changes requested):
Definition and translation of Predicate failures look good, except some `InjectRuleFailure` instances are missing. Definition of new rule, however, is not necessary.

Let me know if you can't figure it out or if something is preventing you from reusing UTXO rule from Babbage.

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Utxo.hs
Lucsanszky: I was trying to reuse the rule initially but that lead to (for all the `InjectRuleFailure` instances): 
```    
• Couldn't match type: ConwayUtxoPredFailure (ConwayEra c)
                 with: BabbageUtxoPredFailure (ConwayEra c)`
```
Which makes sense, since we are reusing the rule from `Babbage` and in the `STS (BabbageUTXO)` instance we specify the `PredicateFailure` type to be `BabbageUtxoPredFailure era` (line 464 in `Babbage` `Utxo.hs`). This is what prevents me from reusing the rule when I also want to use `ConwayUtxoPredFailure`. That's why I thought that I'll have to do something similar to the `STS (ConwayUTXOW)` instance, which is using `babbageUtxowTransition` for `transitionRules`. However, I also can't reuse `utxoTransition` from `Babbage` (without modifying it) because it has the `EraRule "UTXO" era ~ BabbageUTXO era` constraint.

Is there a way to reuse the `UTXO` rules from `Babbage` while also using `ConwayUtxoPredFailure` for predicate failures?

lehins: > Is there a way to reuse the UTXO rules from Babbage while also using ConwayUtxoPredFailure for predicate failures?

Yes there is and that is exactly the goal.

lehins: Actually, my bad, I do see the problem that you are running into. So, you do need `ConwayUTXO` type and a new `STS` instance, like you did in this PR, because you are setting the new predicate failure. But you can reuse the `utxoTransition` from here: https://github.com/IntersectMBO/cardano-ledger/blob/2c9ba10d085331fa99840a63b40b8e8b716fba5d/eras/babbage/impl/src/Cardano/Ledger/Babbage/Rules/Utxo.hs#L343-L361

The problem is in the fact that the type signature makes this restriction: `EraRule "UTXO" era ~ BabbageUTXO era`, which we can get rid of by:
* removing that type equality constraint
* replacing `BabbageUTXO` with `EraRule "UTXO" era` in `Embed (EraRule "UTXOS" era) (BabbageUTXO era)` 
* and adding a few mores constrains (which ones I can't tell without building the code).

Let me know if it doesn't unblock you and we can talk more about it.

lehins: >  Would it be possible or okay to modify the utxoTransition in Alonzo and remove EraRule "UTXO" era ~ AlonzoUTXO era so I can reuse that in Conways utxoTransition?

Important note on the above question. In Conway we are reusing UTXO rule from Babbage, not Alonzo

eras/conway/impl/src/Cardano/Ledger/Conway/Era.hs
lehins: Yes, that comment is wrong. We do need this new data type.

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Utxo.hs
```diff
@@ -1,58 +1,230 @@
...
+  | WrongNetworkWithdrawal
+      -- | the expected network id
+      !Network
+      -- | the set of reward addresses with incorrect network IDs
+      !(Set (RewardAccount (EraCrypto era)))
+  | -- | list of supplied transaction outputs that are too small
+    OutputTooSmallUTxO
+      ![TxOut era]
+  | -- | Subtransition Failures
+    UtxosFailure (PredicateFailure (EraRule "UTXOS" era))
```
lehins: I would like to suggest move this one to the top of the sum type, just so it stands out. This will affect coders/decoders, but I think it is worth it.

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Utxo.hs
```diff
@@ -1,58 +1,403 @@
...
+  , AlonzoEraTxWits era
+  , Tx era ~ AlonzoTx era
+  , EraRule "UTXO" era ~ ConwayUTXO era
+  , InjectRuleFailure "UTXO" ShelleyUtxoPredFailure era
+  , InjectRuleFailure "UTXO" AllegraUtxoPredFailure era
+  , InjectRuleFailure "UTXO" AlonzoUtxoPredFailure era
+  , InjectRuleFailure "UTXO" BabbageUtxoPredFailure era
+  , InjectRuleFailure "UTXO" ConwayUtxoPredFailure era
+  , -- instructions for calling UTXOS from BabbageUTXO
+    Embed (EraRule "UTXOS" era) (ConwayUTXO era)
```
lehins: That was a pretty useless comment for Babbage, but here it makes even less sense.
```suggestion
  , Embed (EraRule "UTXOS" era) (ConwayUTXO era)
```

lehins (approved):
Perfect! :+1: 
Thank you.

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Utxo.hs
Lucsanszky: Agreed, done.

PR #4266, opened 2024-04-15
Add imptest to propose and enact unknown costmodels

### LINKED ISSUE #4186: Imp test proposals with funny CostModels
There should be an imp test in a Conway imp test suite which proposes and enacts a ParameterUpdate action with a CostModel for a plutus language that is unknown and is unlikely to be known in close future (eg. language tag 255 or even better generated at random in a known bad range, which can be inferred from `eraMaxLanguage`)

aniketd:
# Description

Resolves #4186 

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/GovSpec.hs
+++ b/eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/GovSpec.hs
@@ -23,6 +23,7 @@ import Cardano.Ledger.Conway.Core
 import Cardano.Ledger.Conway.Governance
 import Cardano.Ledger.Conway.Rules (ConwayGovPredFailure (..))
 import Cardano.Ledger.Credential (Credential (KeyHashObj))
+import Cardano.Ledger.Plutus.CostModels (updateCostModels)
 import Cardano.Ledger.Shelley.LedgerState
 import Cardano.Ledger.Val (zero, (<->))
 import Data.Default.Class (Default (..))
@@ -34,6 +35,7 @@ import Data.Tree
 import Lens.Micro
 import Test.Cardano.Ledger.Conway.Arbitrary ()
 import Test.Cardano.Ledger.Conway.ImpTest
+import Test.Cardano.Ledger.Core.Arbitrary (FlexibleCostModels (..))
 import Test.Cardano.Ledger.Core.Rational (IsRatio (..))
 import Test.Cardano.Ledger.Imp.Common hiding (Success)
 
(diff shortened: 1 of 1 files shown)


libs/cardano-ledger-core/testlib/Test/Cardano/Ledger/Core/Arbitrary.hs
```diff
@@ -823,6 +824,12 @@ instance Arbitrary Prices where
 instance Arbitrary CostModel where
   arbitrary = elements nonNativeLanguages >>= genValidCostModel
 
+newtype UnknownCostModels = UnknownCostModels
```
teodanciu: Sorry, just a trivial remark: would you mind moving it a bit down, so that it's next to `FlexibleCostModels` ?  I think that would make it easier to find them and understand the difference. 
Also it seems sort of 'changelog-worthy'.

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/GovSpec.hs
```diff
@@ -54,6 +55,25 @@ spec =
...
+      UnknownCostModels ukcm <- arbitrary
+      (drepC, hotCommitteeC, _gpi) <- electBasicCommittee
+      gai <-
+        submitParameterChange SNothing $
+          emptyPParamsUpdate
+            & ppuCostModelsL .~ SJust ukcm
+      submitYesVote_ (DRepVoter drepC) gai
+      submitYesVote_ (CommitteeVoter hotCommitteeC) gai
+      passNEpochs 2
+      getLastEnactedParameterChange `shouldReturn` SJust (GovPurposeId gai)
```
teodanciu, line 78: Any particular reason to not also check that the enacted parameters are the ones generated ?

libs/cardano-ledger-core/testlib/Test/Cardano/Ledger/Core/Arbitrary.hs
lehins: We actually don't need a specific type of `UnknownCostModels`. It is much better to use `FlexibleCostModels` , because it also can generate only unknown CostModels, but it can also generate other valid cases. So please remove this type and use `FlexibleCostModels` instead for the test case in this PR
```suggestion
```

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/GovSpec.hs
lehins: Yes, we do need to verify that the updates wen't through successfully.

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/GovSpec.hs
```diff
@@ -54,6 +55,25 @@ spec =
...
+    unknownCostModelsSpec
+
+unknownCostModelsSpec ::
+  forall era.
+  ConwayEraImp era =>
+  SpecWith (ImpTestState era)
+unknownCostModelsSpec =
+  describe "Unknown CostModels" $ do
+    it "Accepts them" $ do
+      UnknownCostModels ukcm <- arbitrary
```
lehins: As mentioned in my comment below:
```suggestion
      FlexibleCostModels cms <- arbitrary
```

lehins (changes requested):
It looks good, but we can do a bit better

teodanciu (approved):
Looks good to me

lehins (approved):
:+1:

PR #4267, opened 2024-04-16
Expand TxAuxData interface

lehins:
# Description

Because `TxAuxData` isn't really used in the ledger rules it has been neglected whenever we did the refactor of the ledger interface using lenses.

Now that we expanding our testing and have a need for era agnostic type construction we could use the same interface for `TxAuxData` as for other type families in ledger.

This PR introduces:

* `Cardano.Ledger.Metadata` module with `Metadatum` type moved to it from `cardano-ledger-shelley`
* `mkBasicTxAuxData` and `metadataTxAuxDataL` to `EraTxAuxData` type class.
* `AllegraEraTxAuxData` type class with `timelockScriptsTxAuxDataL` lens
* `AlonzoEraTxAuxData`  type class with `plutusScriptsTxAuxDataL` lens

# Checklist

- [x] Self-reviewed the diff


eras/babbage/impl/cardano-ledger-babbage.cabal
```diff
@@ -75,7 +75,7 @@ library
         cardano-ledger-allegra ^>=1.4,
         cardano-ledger-alonzo ^>=1.8,
         cardano-ledger-binary ^>=1.3,
-        cardano-ledger-core ^>=1.11,
```
teodanciu, line 78: ~Probably this should cause the version of babbage to become `1.8.0.1`~

teodanciu: Nevermind!  I thought we released 1.8.0.0

teodanciu (approved):
Great!  (and almost definitely what I should have done in my other PR instead of assuming there is some reason to not have it and hacking my way around it).
Looks good to me, so nice to have this loose end tied up.

PR #4268, opened 2024-04-16
Enable conformance tests for GOV rule

### LINKED ISSUE #4255: Conformance testing of the GOV rule
Currently testing of GOV rule is disabled in the conformance testing, which needs to be fixed. Quote from https://github.com/IntersectMBO/cardano-ledger/pull/4212#discussion_r1557723923

> GOV rule should have all the necessary constraints, but when I ran it, I got some mismatches between the spec and the implementation. I think some of it might be because I haven't properly updated all of the types after updating. The next PR will be about getting the GOV rule to pass.

Soupstraw:
# Description

This PR adds conformance testing for the `GOV` rule. 

resolves #4255

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/eras/shelley/impl/testlib/Test/Cardano/Ledger/Shelley/ImpTest.hs
+++ b/eras/shelley/impl/testlib/Test/Cardano/Ledger/Shelley/ImpTest.hs
@@ -569,6 +569,14 @@ newtype ImpTestM era a = ImpTestM {_unImpTestM :: ReaderT (ImpTestEnv era) IO a}
     , MonadReader (ImpTestEnv era)
     )
 
+instance (Testable a, ShelleyEraImp era) => Testable (ImpTestM era a) where
+  property m = property $ MkGen $ \qcGen qcSize ->
+    ioProperty $ do
+      preppedState <-
+        execImpTestM (Just qcSize) (mixinCurrentGen def qcGen) $
+          addRootTxOut >> initImpTestState
+      evalImpTestM (Just qcSize) preppedState m
+
 instance MonadWriter [SomeSTSEvent era] (ImpTestM era) where
   writer (x, evs) = (impEventsL %= (<> evs)) $> x
   listen act = do
@@ -1107,43 +1115,53 @@ logEntry e = impLogL %= (<> pretty loc <> "\t" <> pretty e <> line)
 logToExpr :: (HasCallStack, ToExpr a) => a -> ImpTestM era ()
 logToExpr e = logEntry (showExpr e)
 
+instance ShelleyEraImp era => Default (ImpTestState era) where
+  def =
+    ImpTestState
+      { impNES = initShelleyImpNES
+      , impRootTxIn = TxIn (mkTxId 0) minBound
+      , impKeyPairs = mempty
+      , impByronKeyPairs = mempty
+      , impNativeScripts = mempty
+      , impLastTick = 0
+      , impGlobals = testGlobals
+      , impLog = mempty
+      , impGen = mkQCGen 2024
+      , impEvents = mempty
+      }
+
 withImpState ::
   ShelleyEraImp era =>
   SpecWith (ImpTestState era) ->
   Spec
 withImpState = withImpStateModified id
 
+addRootTxOut ::
+  forall era m.
+  ( MonadGen m
+  , MonadState (ImpTestState era) m
+  , ShelleyEraImp era
+  ) =>
+  m ()
+addRootTxOut = do
+  (rootKeyHash, _) <- freshKeyPair @era
+  let rootAddr = Addr Testnet (KeyHashObj rootKeyHash) StakeRefNull
+      rootTxOut = mkBasicTxOut rootAddr $ inject rootCoin
+  impNESL . nesEsL . esLStateL . lsUTxOStateL . utxosUtxoL
+    %= (<> UTxO (Map.singleton (impRootTxIn @era def) rootTxOut))
+  where
+    rootCoin = Coin 1_000_000_000
+
 withImpStateModified ::
+  forall era.
   ShelleyEraImp era =>
   (ImpTestState era -> ImpTestState era) ->
   SpecWith (ImpTestState era) ->
   Spec
 withImpStateModified f =
   beforeAll $
-    execImpTestM Nothing (f impTestState0) $
+    execImpTestM Nothing (f def) $
       addRootTxOut >> initImpTestState
-  where
-    impTestState0 =
-      ImpTestState
-        { impNES = initShelleyImpNES
-        , impRootTxIn = rootTxIn
-        , impKeyPairs = mempty
-        , impByronKeyPairs = mempty
-        , impNativeScripts = mempty
-        , impLastTick = 0
-        , impGlobals = testGlobals
-        , impLog = mempty
-        , impGen = mkQCGen 2024
-        , impEvents = mempty
-        }
-    rootCoin = Coin 1_000_000_000
-    rootTxIn = TxIn (mkTxId 0) minBound
-    addRootTxOut = do
-      (rootKeyHash, _) <- freshKeyPair
-      let rootAddr = Addr Testnet (KeyHashObj rootKeyHash) StakeRefNull
-          rootTxOut = mkBasicTxOut rootAddr $ inject rootCoin
-      impNESL . nesEsL . esLStateL . lsUTxOStateL . utxosUtxoL
-        %= (<> UTxO (Map.singleton rootTxIn rootTxOut))
 
 -- | Creates a fresh @SafeHash@
 freshSafeHash :: Era era => ImpTestM era (SafeHash (EraCrypto era) a)
--- a/libs/cardano-ledger-conformance/src/Test/Cardano/Ledger/Conformance/ExecSpecRule/Core.hs
+++ b/libs/cardano-ledger-conformance/src/Test/Cardano/Ledger/Conformance/ExecSpecRule/Core.hs
@@ -14,34 +14,33 @@ module Test.Cardano.Ledger.Conformance.ExecSpecRule.Core (
   ExecSpecRule (..),
   conformsToImpl,
   computationResultToEither,
+  generatesWithin,
+  runConformance,
+  checkConformance,
 ) where
 
-import Cardano.Ledger.BaseTypes (Inject, ShelleyBase)
+import Cardano.Ledger.BaseTypes (ShelleyBase)
 import Cardano.Ledger.Core (EraRule)
 import qualified Constrained as CV2
-import Constrained.Base (Specification (..))
+import Constrained.Base (shrinkWithSpec, simplifySpec)
 import Control.State.Transition.Extended (STS (..))
 import Data.Bifunctor (Bifunctor (..))
 import Data.Bitraversable (bimapM)
-import Data.Typeable (Typeable)
+import Data.Functor (($>))
+import Data.Typeable (Proxy (..), Typeable, typeRep)
 import GHC.Base (Constraint, NonEmpty, Symbol, Type)
 import GHC.TypeLits (KnownSymbol)
 import qualified Lib as Agda
 import Test.Cardano.Ledger.Conformance.SpecTranslate.Core (SpecTranslate (..), runSpecTransM)
-import Test.Cardano.Ledger.Imp.Common (
-  MonadGen (..),
-  NFData,
-  ToExpr,
-  expectRightExpr,
-  shouldBeExpr,
- )
+import Test.Cardano.Ledger.Imp.Common
 import Test.Cardano.Ledger.Shelley.ImpTest (
   ImpTestM,
+  ShelleyEraImp,
   impAnn,
   logEntry,
-  logToExpr,
   tryRunImpRule,
  )
+import UnliftIO (evaluateDeep)
 
 type ForAllRuleTypes (c :: Type -> Constraint) rule era =
   ( c (Environment (EraRule rule era))
@@ -51,14 +50,8 @@ type ForAllRuleTypes (c :: Type -> Constraint) rule era =
 
 class
   ( ForAllRuleTypes (CV2.HasSpec fn) rule era
-  , ForAllRuleTypes (SpecTranslate (ExecContext fn rule era)) rule era
   , ForAllRuleTypes ToExpr rule era
   , ForAllRuleTypes NFData rule era
-  , CV2.HasSpec fn (ExecContext fn rule era)
-  , Inject (ExecContext fn rule era) (SpecTransContext (Environment (EraRule rule era)))
-  , Inject (ExecContext fn rule era) (SpecTransContext (State (EraRule rule era)))
-  , Inject (ExecContext fn rule era) (SpecTransContext (Signal (EraRule rule era)))
-  , Inject (ExecContext fn rule era) (SpecTransContext (PredicateFailure (EraRule rule era)))
   , Eq (Event (EraRule rule era))
   , Typeable (Event (EraRule rule era))
   , KnownSymbol rule
@@ -73,22 +66,25 @@ class
   type ExecContext fn rule era = ()
 
   environmentSpec ::
+    ExecContext fn rule era ->
     CV2.Specification fn (Environment (EraRule rule era))
 
   stateSpec ::
+    ExecContext fn rule era ->
     Environment (EraRule rule era) ->
     CV2.Specification fn (State (EraRule rule era))
 
   signalSpec ::
+    ExecContext fn rule era ->
     Environment (EraRule rule era) ->
     State (EraRule rule era) ->
     CV2.Specification fn (Signal (EraRule rule era))
 
-  execContextSpec :: CV2.Specification fn (ExecContext fn rule era)
-  default execContextSpec ::
-    ExecContext fn rule era ~ () =>
-    CV2.Specification fn (ExecContext fn rule era)
-  execContextSpec = TrueSpec
+  genExecContext :: Gen (ExecContext fn rule era)
+  default genExecContext ::
+    Arbitrary (ExecContext fn rule era) =>
+    Gen (ExecContext fn rule era)
+  genExecContext = arbitrary
 
   runAgdaRule ::
     SpecRep (Environment (EraRule rule era)) ->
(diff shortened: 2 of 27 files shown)


eras/shelley/impl/testlib/Test/Cardano/Ledger/Shelley/ImpTest.hs
```diff
@@ -1105,37 +1115,36 @@ logEntry e = impLogL %= (<> pretty loc <> "\t" <> pretty e <> line)
...
 withImpState ::
   forall era.
   ShelleyEraImp era =>
   SpecWith (ImpTestState era) ->
   Spec
 withImpState =
-  beforeAll $
-    execImpTestM Nothing impTestState0 $
-      addRootTxOut >> initImpTestState
+  beforeAll $ execImpTestM Nothing mkImpState addRootTxOut
```
teodanciu: Looks like  the call to `initImpTestState` was removed by accident

libs/cardano-data/src/Data/OMap/Strict.hs
```diff
@@ -376,8 +376,8 @@ instance HasOKey k v => IsList (OMap k v) where
   fromList = fromFoldable
   toList = F.toList
 
-assocList :: OMap k v -> [(k, v)]
-assocList = Map.toList . toMap
+assocList :: Ord k => OMap k v -> [(k, v)]
+assocList = toList . toStrictSeqOfPairs
```
teodanciu: Would be nice if we could have this in a separate commit, if it's not too difficult

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/Conway/Gov.hs
```diff
@@ -128,9 +129,8 @@ govProposalsSpec GovEnv {geEpoch, gePPolicy, gePrevGovActionIds} =
       , genHint listSizeHint unorderedProposals
       ]
   where
-    GovRelation {..} = gePrevGovActionIds
-    treeGenHint = (Just 2, 10)
-    listSizeHint = 5
+    treeGenHint = (Just 2, 1)
```
MaximilianAlgehed: These can probably be reverted now that the efficiency issues are fixed.

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/Conway/Gov.hs
```diff
@@ -158,15 +158,13 @@ allGASAndChildInTree t k =
 
 wellFormedChildren ::
   IsConwayUniv fn =>
-  Term fn (StrictMaybe (GovActionId StandardCrypto)) ->
   Term fn ProposalTree ->
   Pred fn
-wellFormedChildren root rootAndTrees =
+wellFormedChildren rootAndTrees =
   match rootAndTrees $ \root' trees ->
-    [ assert $ root ==. root' -- The root matches the root given in the environment
```
MaximilianAlgehed, line 166: What happened to this constraint?

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/Conway/Gov.hs
```diff
@@ -158,15 +158,13 @@ allGASAndChildInTree t k =
...
-    , forAll trees $ \t ->
+    [ forAll trees $ \t ->
         [ -- Every node just below the root has the root as its parent
-          withPrevActId (rootLabel_ t) (assert . (==. root))
+          withPrevActId (rootLabel_ t) (assert . (==. root'))
```
MaximilianAlgehed: s/root'/root

libs/constrained-generators/src/Constrained/Base.hs
```diff
@@ -553,8 +553,8 @@ instance HasSpec fn a => Semigroup (Specification fn a) where
         (filter (`notElem` cant) as)
   TypeSpec s cant <> MemberSpec as = MemberSpec as <> TypeSpec s cant
   SuspendedSpec v p <> SuspendedSpec v' p' = SuspendedSpec v (p <> rename v' v p')
-  SuspendedSpec v ps <> s = SuspendedSpec v (ps <> satisfies (V v) s)
-  s <> SuspendedSpec v ps = SuspendedSpec v (ps <> satisfies (V v) s)
+  SuspendedSpec v ps <> s = SuspendedSpec v (optimisePred $ ps <> satisfies (V v) s)
```
MaximilianAlgehed: Check of this change can be reverted.

libs/constrained-generators/src/Constrained/Base.hs
```diff
@@ -932,7 +932,7 @@ prepareLinearization p = do
 -- | Generate a satisfying `Env` for a `p : Pred fn`. The `Env` contains values for
 -- all the free variables in `flattenPred p`.
 genFromPreds :: (MonadGenError m, BaseUniverse fn) => Pred fn -> GenT m Env
-genFromPreds (optimisePred -> preds) = do
+genFromPreds (optimisePred . optimisePred -> preds) = do
```
MaximilianAlgehed, line 995:
```suggestion
-- TODO: remove this once optimisePred does a proper fixpoint computation
genFromPreds (optimisePred . optimisePred -> preds) = do
```

MaximilianAlgehed (changes requested):

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/Conway/Gov.hs
Soupstraw: I removed `GovRelation` from `GovEnv`, since it was the exact same value as the `GovRelation` in the state. Now this constraint became redundant since we don't have to keep these two values in sync anymore.

libs/cardano-data/src/Data/OMap/Strict.hs
Soupstraw: Done: #4310

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/GovSpec.hs
```diff
@@ -497,6 +502,38 @@ proposalsWithVotingSpec =
...
+      deposit <- getsNES $ nesEsL . curPParamsEpochStateL . ppGovActionDepositL
+      ens <- getEnactState
+      withdrawals <- arbitrary
+      let
+        mkProp name action = do
+          ProposalProcedure
+            { pProcReturnAddr = returnAddr
+            , pProcGovAction = action
+            , pProcDeposit = deposit
+            , pProcAnchor = Anchor (fromJust $ textToUrl 16 name) def
```
teodanciu: You could use arbitrary here if you wanted: 
`anchor <- arbitrary`

teodanciu: Nevermind, I only now see you're using the name, sorry.

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/Conway/Gov.hs
aniketd: This implies that we assume that `proposalsApplyEnactment` is correct, but I suppose even though we eliminate the use of `prevGovActionIds` from the `Environment`, the correctness will still be tested via traces over the `EPOCH` rule.

MaximilianAlgehed (approved):

eras/shelley/impl/testlib/Test/Cardano/Ledger/Shelley/ImpTest.hs
```diff
@@ -569,6 +569,16 @@ newtype ImpTestM era a = ImpTestM {_unImpTestM :: ReaderT (ImpTestEnv era) IO a}
...
 
+instance
+  ( Testable a
+  , ShelleyEraImp era
+  ) =>
+  Testable (ImpTestM era a)
+  where
+  property m = property . fmap ioProperty . runGenT $ do
+    res <- liftGen $ runImpTestGenM mkImpState m
+    liftIO $ fst <$> res
```
lehins: There are a couple of problems with current solution. 

## QuickCheck Gen and Size

Neither quickcheck size nor generator (that get's supplied by the `--seed` argument) are respected

So, this is a better solution:
```haskell
instance (Testable a, ShelleyEraImp era) => Testable (ImpTestM era a) where
  property m = property $ MkGen $ \qcGen qcSize ->
    ioProperty $ evalImpTestM (Just qcSize) (mixinCurrentGen mkImpState qcGen) m
```

## Missing ImpTestState initialization

Another problem with this approach is that we don't do the initialization of the state that normally done by `withImpState`, so this is what I'd recommend in the end:

```suggestion
instance (Testable a, ShelleyEraImp era) => Testable (ImpTestM era a) where
  property m = property $ MkGen $ \qcGen qcSize ->
    ioProperty $ do
      preppedState <-
        execImpTestM (Just qcSize) (mixinCurrentGen mkImpState qcGen) $
          addRootTxOut >> initImpTestState
      evalImpTestM (Just qcSize) preppedState m
```

We'll need to extract the common part and reuse it in the instance above:
```haskell
addRootTxOut :: forall era. ShelleyEraImp era => ImpTestM era ()
addRootTxOut = do
  let rootCoin = Coin 1_000_000_000
  (rootKeyHash, _) <- freshKeyPair
  let rootAddr = Addr Testnet (KeyHashObj rootKeyHash) StakeRefNull
      rootTxOut = mkBasicTxOut rootAddr $ inject rootCoin
  impNESL . nesEsL . esLStateL . lsUTxOStateL . utxosUtxoL
    %= (<> UTxO (Map.singleton (impRootTxIn @era mkImpState) rootTxOut))
```

I am still not too happy that we can't provide a modified state to the property using something like `withImpStateModified`, but that's not a big of a compromise for getting a very useful instance of `Testable` for `ImpTestM`

eras/shelley/impl/testlib/Test/Cardano/Ledger/Shelley/ImpTest.hs
```diff
@@ -1107,43 +1117,45 @@ logEntry e = impLogL %= (<> pretty loc <> "\t" <> pretty e <> line)
 logToExpr :: (HasCallStack, ToExpr a) => a -> ImpTestM era ()
 logToExpr e = logEntry (showExpr e)
 
+mkImpState :: ShelleyEraImp era => ImpTestState era
+mkImpState =
```
lehins: Maybe we should prefix it with something like `def` to indicate that this is the bare minimum initial/default state. 
```suggestion
defImpState :: ShelleyEraImp era => ImpTestState era
defImpState =
```
This is just a suggestion.

libs/cardano-ledger-conformance/src/Test/Cardano/Ledger/Conformance/ExecSpecRule/Core.hs
```diff
@@ -98,36 +110,173 @@ class
...
+      unlines
+        [ ""
+        , "===== DIFF ====="
+        , diffExpr implResTest agdaResTest
+        , ""
+        , "Legend:"
+        , "\t\ESC[91mImplemenation"
+        , "\t\ESC[92mSpecification\ESC[39m"
+        ]
+  unless (implResTest == agdaResTest) $ expectationFailure failMsg
```
lehins: Looks like it would be a good idea to create a reusable function that checks this expectation. There is another copy of the same code earlier in the PR.

Nice touch using escape codes :wink:

libs/cardano-ledger-conformance/src/Test/Cardano/Ledger/Conformance/ExecSpecRule/Core.hs
```diff
@@ -98,36 +110,173 @@ class
...
+  ) =>
+  Gen a ->
+  Int ->
+  Spec
+generatesWithin gen timeout =
+  prop (aName <> " generates in reasonable time")
+    . forAllShow gen showExpr
+    $ \x -> within timeout $ ioProperty (evaluateDeep x $> ())
+  where
+    aName = showsTypeRep (typeRep $ Proxy @a) ""
```
lehins: Slight simplification
```suggestion
    aName = show (typeRep $ Proxy @a)
```

libs/cardano-ledger-conformance/src/Test/Cardano/Ledger/Conformance/ExecSpecRule/Core.hs
```diff
@@ -14,34 +14,49 @@ module Test.Cardano.Ledger.Conformance.ExecSpecRule.Core (
...
 import Data.Bitraversable (bimapM)
-import Data.Typeable (Typeable)
+import Data.Functor (($>))
+import Data.Typeable (Proxy (..), Typeable, showsTypeRep, typeRep)
 import GHC.Base (Constraint, NonEmpty, Symbol, Type)
 import GHC.TypeLits (KnownSymbol)
 import qualified Lib as Agda
+import Test.Cardano.Ledger.Common (Arbitrary (..), Gen, Testable (..), forAllShow)
 import Test.Cardano.Ledger.Conformance.SpecTranslate.Core (SpecTranslate (..), runSpecTransM)
 import Test.Cardano.Ledger.Imp.Common (
```
lehins: You really don't need to be explicit about imports of anything from this module. There is a point why it is called `Common`

libs/cardano-ledger-conformance/src/Test/Cardano/Ledger/Conformance/ExecSpecRule/Core.hs
```diff
@@ -14,34 +14,49 @@ module Test.Cardano.Ledger.Conformance.ExecSpecRule.Core (
...
 import Control.State.Transition.Extended (STS (..))
 import Data.Bifunctor (Bifunctor (..))
 import Data.Bitraversable (bimapM)
-import Data.Typeable (Typeable)
+import Data.Functor (($>))
+import Data.Typeable (Proxy (..), Typeable, showsTypeRep, typeRep)
 import GHC.Base (Constraint, NonEmpty, Symbol, Type)
 import GHC.TypeLits (KnownSymbol)
 import qualified Lib as Agda
+import Test.Cardano.Ledger.Common (Arbitrary (..), Gen, Testable (..), forAllShow)
```
lehins: All of this is re-exported from `Test.Cardano.Ledger.Imp.Common` as well
```suggestion
```

libs/cardano-ledger-conformance/src/Test/Cardano/Ledger/Conformance/ExecSpecRule/Core.hs
```diff
@@ -14,34 +14,49 @@ module Test.Cardano.Ledger.Conformance.ExecSpecRule.Core (
...
+import Cardano.Ledger.BaseTypes (ShelleyBase)
 import Cardano.Ledger.Core (EraRule)
 import qualified Constrained as CV2
-import Constrained.Base (Specification (..))
+import Constrained.Base (shrinkWithSpec, simplifySpec)
 import Control.State.Transition.Extended (STS (..))
 import Data.Bifunctor (Bifunctor (..))
 import Data.Bitraversable (bimapM)
-import Data.Typeable (Typeable)
+import Data.Functor (($>))
```
lehins, line 29: That's why I always use `<$`, so I don't have to explicitly import `$>` :slightly_smiling_face: 

Not suggesting you change it, just an observation

libs/cardano-ledger-conformance/src/Test/Cardano/Ledger/Conformance/Orphans.hs
```diff
@@ -100,10 +104,24 @@ instance NFData Tx
 
 instance NFData UTxOEnv
 
+reverseEndianness :: String -> String
+reverseEndianness (x : y : t) = reverseEndianness t <> [x, y]
+reverseEndianness (x : t) = reverseEndianness t <> ['0', x]
+reverseEndianness [] = []
+
+instance ToExpr Credential where
+  toExpr (KeyHashObj h) = App "KeyHashObj" [toExpr . reverseEndianness $ showHex h ""]
```
lehins: I suggest you convert the `Integer` to `ByteString` while specifying the number of bytes you expect for the type with `integralToByteStringN` instead of relying on an ugly hacks like `reverseEndianness` of showing `Integer` as hex, because current implementation of `showHex` together with `reverseEndianness` will not show correctly hashes represented by small numbers. `ByteString` then can be shown as hex properly with `base16-bytestring`


Fun coincidence. Look what I found in `cardano-crypto-class`: https://github.com/IntersectMBO/cardano-base/blob/02f4a077c4730409ed70ad19da2158d68b9e62ad/cardano-crypto-class/src/Cardano/Crypto/Util.hs#L130-L137

lehins (approved):
There are only two major things I'd like to see changed in this PR:
* Testable instance for `ImpTestM`
* showing of Integers as Hex

Other than that it looks great!
Nice job on figuring conformance tests for GOV rule!

eras/shelley/impl/testlib/Test/Cardano/Ledger/Shelley/ImpTest.hs
Soupstraw: Might as well make a `Default` instance

libs/cardano-ledger-conformance/src/Test/Cardano/Ledger/Conformance/ExecSpecRule/Core.hs
Soupstraw: I like it more when it's pointing to the right, makes it feel like the computation is moving forward rather than backwards :D

libs/cardano-ledger-conformance/src/Test/Cardano/Ledger/Conformance/ExecSpecRule/Core.hs
```diff
@@ -98,36 +94,189 @@ class
...
+checkConformance implResTest agdaResTest = do
+  let
+    failMsg =
+      unlines
+        [ ""
+        , "===== DIFF ====="
+        , diffExpr implResTest agdaResTest
+        , ""
+        , "Legend:"
+        , "\t\ESC[91mImplemenation"
```
teodanciu: You have a small typo here, just in case it annoys you later on, otherwise we can fix it next time

Soupstraw: Thanks, fixed it.

PR #4269, opened 2024-04-16
Fix generation bug for sums of positive member spec

MaximilianAlgehed:
# Description

This unblocks @TimSheard 

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/libs/constrained-generators/src/Constrained/Base.hs
+++ b/libs/constrained-generators/src/Constrained/Base.hs
@@ -539,10 +539,11 @@ instance HasSpec fn a => Semigroup (Spec fn a) where
   ErrorSpec e <> ErrorSpec e' = ErrorSpec (e ++ "" : (replicate 20 '-') : "" : e')
   ErrorSpec e <> _ = ErrorSpec e
   _ <> ErrorSpec e = ErrorSpec e
-  MemberSpec [a] <> spec
-    | a `conformsToSpec` spec = MemberSpec [a]
-    | otherwise = ErrorSpec [show a, "does not conform to", show spec]
-  spec <> MemberSpec [a] = MemberSpec [a] <> spec
+  -- TODO: these can interact poorly with unsafeExists!
+  -- MemberSpec [a] <> spec
+  --   | a `conformsToSpec` spec = MemberSpec [a]
+  --   | otherwise = ErrorSpec [show a, "does not conform to", show spec]
+  -- spec <> MemberSpec [a] = MemberSpec [a] <> spec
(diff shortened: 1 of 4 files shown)


libs/constrained-generators/src/Constrained/Base.hs
```diff
@@ -539,6 +539,7 @@ instance HasSpec fn a => Semigroup (Spec fn a) where
   ErrorSpec e <> ErrorSpec e' = ErrorSpec (e ++ "" : (replicate 20 '-') : "" : e')
   ErrorSpec e <> _ = ErrorSpec e
   _ <> ErrorSpec e = ErrorSpec e
+  -- TODO: these can interact poorly with unsafeExists!
   MemberSpec [a] <> spec
     | a `conformsToSpec` spec = MemberSpec [a]
     | otherwise = ErrorSpec [show a, "does not conform to", show spec]
```
TimSheard: What about adding another guard like
```
memberSpec [a] <>spec
  | not (suspenedSpec spec)
  | a `conformsToSpec` spec = MemberSpec [a]
  | otherwise = ErrorSpec [show a, "does not conform to", show spec]
```
Does that get rid of the poor interaction?

libs/constrained-generators/src/Constrained/Base.hs
```diff
@@ -1534,7 +1544,10 @@ simplifyPred = \case
     t'@App {} -> Let t' (simplifyBinder b)
     -- Variable or literal
     t' | x :-> p <- b -> simplifyPred $ substitutePred x t' p
-  Exists k b -> Exists k (simplifyBinder b)
+  Exists k b -> case simplifyBinder b of
+    _ :-> TruePred -> TruePred
+    x :-> p | Just t <- pinnedBy x p -> simplifyPred $ substitutePred x t p
+    b' -> Exists k b'
```
TimSheard: Maybe put in a comment, with an example that goes down the path that executes this  new code?

TimSheard (approved):
Looks good to me. Add a few comments where indicated.

libs/constrained-generators/src/Constrained/Base.hs
MaximilianAlgehed: yeah, that's a band-aide for now. We can probably refine it later.

MaximilianAlgehed: Problem is that if you have e.g. `Cartesian (SuspendedSpec ...) TrueSpec` that doesn't help

MaximilianAlgehed: I'll comment the cases out for now and we'll think it through a bit more later.

PR #4271, opened 2024-04-16
Implement POC of same number of argumetns for PlutusV3

lehins:
# Description

Proof-of-Concept: making the same number (three) of arguments for PlutusV3, regardless of plutus purpose.

CI is still failing because tests have not been adjusted to account for this change.

# Checklist

- [ ] Self-reviewed the diff


lehins:
This approach is not sufficient for CIP-0069, because it is missing the desired feature of making the spending datum optional, which doe s require changing the ledger rules. Therefore we decided to go with a proper solution, even if it means it will take longer to get it done.

PR #4272, opened 2024-04-17
simplify foldMap interface to higher order syntax

MaximilianAlgehed:
# Description

This is a tiny usability change

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/Conway/Instances.hs
+++ b/libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/Conway/Instances.hs
@@ -37,7 +37,12 @@ module Test.Cardano.Ledger.Constrained.Conway.Instances (
   cSNothing_,
   cSJust_,
   succV_,
+  maryValueCoin_,
   strLen_,
+  sizedValue_,
+  sizedSize_,
+  txOutVal_,
+  pProcDeposit_,
   pProcGovAction_,
   IsConwayUniv,
   gasId_,
@@ -117,6 +122,7 @@ import Data.ByteString.Short (ShortByteString)
 import Data.ByteString.Short qualified as SBS
 import Data.Coerce
 import Data.Foldable
+import Data.Int
 import Data.Kind
 import Data.Map (Map)
 import Data.Map.Strict qualified as Map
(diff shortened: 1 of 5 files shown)


libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/Conway/Instances.hs
```diff
@@ -316,7 +328,7 @@ type BabbageTxOutTypes era =
    , StrictMaybe (Script era)
    ]
 instance (Era era, Val (Value era)) => HasSimpleRep (BabbageTxOut era) where
-  type SimpleRep (BabbageTxOut era) = SOP '["BabbageTxOut" ::: BabbageTxOutTypes era]
+  type TheSop (BabbageTxOut era) = '["BabbageTxOut" ::: BabbageTxOutTypes era]
   toSimpleRep (BabbageTxOut addr val dat msc) =
```
TimSheard, line 332: What is the difference between 'SimplRep' and 'TheSop'?

libs/constrained-generators/src/Constrained/Base.hs
```diff
@@ -3670,17 +3670,25 @@ sum_ ::
...
   Term fn b
-foldMap_ fn = app (foldMapFn fn)
+foldMap_ f = app $ foldMapFn $ toFn $ f (V v)
+  where
+    v = Var (-1) :: Var a
+    toFn :: forall b. HasCallStack => Term fn b -> fn '[a] b
+    toFn (App fn (V v' :> Nil)) | Just Refl <- eqVar v v' = fn
+    toFn (App fn (t :> Nil)) = injectFn $ Compose fn (toFn t)
+    toFn (V v') | Just Refl <- eqVar v v' = idFn
+    toFn _ = error "foldMap_ has not been given a function on the form \\ x -> f (g ... (h x))"
```
TimSheard: 'of the form' better than 'on the form'

libs/constrained-generators/src/Constrained/Base.hs
```diff
@@ -3670,17 +3670,25 @@ sum_ ::
...
   Term fn [a] ->
   Term fn b
-foldMap_ fn = app (foldMapFn fn)
+foldMap_ f = app $ foldMapFn $ toFn $ f (V v)
+  where
+    v = Var (-1) :: Var a
+    toFn :: forall b. HasCallStack => Term fn b -> fn '[a] b
+    toFn (App fn (V v' :> Nil)) | Just Refl <- eqVar v v' = fn
+    toFn (App fn (t :> Nil)) = injectFn $ Compose fn (toFn t)
+    toFn (V v') | Just Refl <- eqVar v v' = idFn
```
TimSheard: Maybe a comment about what is going on here?

TimSheard (approved):
Looks good.  Just add a few comments where indicated.

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/Conway/Instances.hs
MaximilianAlgehed: It's an annoying technical detail - `SimpleRep a = SOP (TheSop a)` and `TheSop = SOPOf (Rep a)` in the generic case - this is what lets you write just `instance ... => HasSimpleRep A` when you don't want a specific instance. When you _do_ want a specific instance you need to define `TheSop` rather than `SimpleRep` if you want `sel`, `con`, and `is/onCon` to work properly. It's something I'm planning on fixing.

PR #4275, opened 2024-04-18
Restrict gov actions during bootstrap

### LINKED ISSUE #3451: Bootstrap period
Right after the hardfork into the conway ledger era there will be an interim period with a limited governance while we wait for the community to register with the governance.

See https://github.com/JaredCorduan/CIPs/blob/voltaire-v1/CIP-1694/README.md#bootstrapping-phase

teodanciu:
# Description

Resolves: #3451 

Implement bootstrap phase, which means, allowing only proposals of type: ParameterChange, HardForkInitiation and Info 
and restricting DRep voting to proposals of type Info. 

The biggest part of the PR is adjusting the tests, namely, 
 *  running all existing tests with protocol version 10 and 
 *  running tests that are relevant (or permitted) during bootstrap phase only with protocol version 9   

I have tried to maximize the number of test we can run for the bootstrap phase, so in some places i have replaced the type of proposal tested (eg: new constitution with parameter update). 
Where this was too complicated (Eg: if drep voting was involved) or the tested proposal is not relevant for bootstrap by the nature of  the proposal, I have separated the tests in two, and ran them with their appropriate protocol version. 
   

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Gov.hs
+++ b/eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Gov.hs
@@ -86,6 +86,7 @@ import Cardano.Ledger.Core
 import Cardano.Ledger.Credential (Credential)
 import Cardano.Ledger.Keys (KeyRole (..))
 import Cardano.Ledger.Rules.ValidationMode (Test, runTest)
+import qualified Cardano.Ledger.Shelley.HardForks as HF (bootstrapPhase)
 import Cardano.Ledger.Shelley.PParams (pvCanFollow)
 import Cardano.Ledger.TxIn (TxId (..))
 import Control.DeepSeq (NFData)
@@ -323,6 +324,16 @@ checkPolicy expectedPolicyHash actualPolicyHash =
   failureUnless (actualPolicyHash == expectedPolicyHash) $
     InvalidPolicyHash actualPolicyHash expectedPolicyHash
 
+checkBootstrapProposal ::
+  EraPParams era =>
+  PParams era ->
+  ProposalProcedure era ->
+  Test (ConwayGovPredFailure era)
+checkBootstrapProposal pp proposal@ProposalProcedure {pProcGovAction}
+  | HF.bootstrapPhase (pp ^. ppProtocolVersionL) =
+      failureUnless (isBootstrapAction pProcGovAction) $ DisallowedProposalDuringBootstrap proposal
+  | otherwise = pure ()
+
 govTransition ::
   forall era.
   ( ConwayEraPParams era
@@ -347,6 +358,8 @@ govTransition = do
   expectedNetworkId <- liftSTS $ asks networkId
 
   let processProposal ps (idx, proposal@ProposalProcedure {..}) = do
+        runTest $ checkBootstrapProposal pp proposal
+
         let newGaid = GovActionId txid idx
 
         -- In a HardFork, check that the ProtVer can follow
@@ -446,6 +459,14 @@ govTransition = do
   tellEvent $ GovNewProposals txid updatedProposalStates
 
   pure updatedProposalStates
+
+isBootstrapAction :: GovAction era -> Bool
+isBootstrapAction =
+  \case
+    ParameterChange {} -> True
+    HardForkInitiation {} -> True
+    InfoAction -> True
+    _ -> False
 
 -- | If the GovAction is a HardFork, then return 3 things (if they exist)
 -- 1) The (StrictMaybe GovPurposeId), pointed to by the HardFork proposal
--- a/eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Gov.hs
+++ b/eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Gov.hs
@@ -163,6 +163,9 @@ data ConwayGovPredFailure era
       (StrictMaybe (ScriptHash (EraCrypto era)))
       -- | The policy script hash of the current constitution
       (StrictMaybe (ScriptHash (EraCrypto era)))
+  | DisallowedProposalDuringBootstrap (ProposalProcedure era)
+  | DisallowedVotesDuringBootstrap
+      (NonEmpty (Voter (EraCrypto era), GovActionId (EraCrypto era)))
   deriving (Eq, Show, Generic)
 
 type instance EraRuleFailure "GOV" (ConwayEra c) = ConwayGovPredFailure (ConwayEra c)
@@ -189,6 +192,8 @@ instance EraPParams era => DecCBOR (ConwayGovPredFailure era) where
     9 -> SumD VotingOnExpiredGovAction <! From
     10 -> SumD ProposalCantFollow <! From <! From <! From
     11 -> SumD InvalidPolicyHash <! From <! From
+    12 -> SumD DisallowedProposalDuringBootstrap <! From
+    13 -> SumD DisallowedVotesDuringBootstrap <! From
     k -> Invalid k
 
 instance EraPParams era => EncCBOR (ConwayGovPredFailure era) where
@@ -221,6 +226,10 @@ instance EraPParams era => EncCBOR (ConwayGovPredFailure era) where
         Sum InvalidPolicyHash 11
           !> To got
           !> To expected
+      DisallowedProposalDuringBootstrap proposal ->
+        Sum DisallowedProposalDuringBootstrap 12 !> To proposal
+      DisallowedVotesDuringBootstrap votes ->
+        Sum DisallowedVotesDuringBootstrap 13 !> To votes
 
 instance EraPParams era => ToCBOR (ConwayGovPredFailure era) where
   toCBOR = toEraCBOR @era
--- a/eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Gov.hs
+++ b/eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Gov.hs
@@ -104,7 +104,6 @@ import Control.State.Transition.Extended (
   tellEvent,
   (?!),
  )
-import Data.Bifunctor (Bifunctor (..))
 import Data.List.NonEmpty (NonEmpty (..))
 import qualified Data.Map.Strict as Map
 import qualified Data.OSet.Strict as OSet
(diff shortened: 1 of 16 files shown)


eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Gov.hs
```diff
@@ -439,6 +475,29 @@ govTransition = do
...
+    InfoAction -> True
+    _ -> False
+
+checkDisallowedVotes ::
+  [(Voter (EraCrypto era), GovActionState era)] ->
+  (Voter (EraCrypto era) -> GovActionState era -> Bool) ->
+  (NonEmpty (Voter (EraCrypto era), GovActionId (EraCrypto era)) -> ConwayGovPredFailure era) ->
+  Test (ConwayGovPredFailure era)
+checkDisallowedVotes votes canVoteOn failure =
+  failureOnNonEmpty disallowedVotes $ failure . fmap (second gasId)
```
lehins: Feel free to keep the usage of `second` if you feel it makes the code clearer
```suggestion
  failureOnNonEmpty disallowedVotes $ failure . fmap (fmap gasId)
```

eras/shelley/impl/src/Cardano/Ledger/Shelley/HardForks.hs
```diff
@@ -64,3 +65,9 @@ forgoPointerAddressResolution ::
   ProtVer ->
   Bool
 forgoPointerAddressResolution pv = pvMajor pv > natVersion @8
+
+-- | Bootstrap phase
+bootstrap ::
```
lehins: Let's name the function as we call it in order to avoid confusion with bootstrap addresses
```suggestion
bootstrapPhase ::
```

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Gov.hs
```diff
@@ -258,28 +268,40 @@ checkVotesAreNotForExpiredActions ::
...
+  Test (ConwayGovPredFailure era)
+checkBootstrapVotes pp votes
+  | HF.bootstrap (pp ^. ppProtocolVersionL) =
+      checkDisallowedVotes votes (flip canVoteOn) DisallowedVotesDuringBootstrap
+  | otherwise = pure ()
+  where
+    canVoteOn gas =
+      \case
+        DRepVoter {} | gasAction gas == InfoAction -> True
+        _ -> isBootstrapAction $ gasAction gas
```
lehins: I recommend changing the order of arguments in `checkDisallowedVotes` in such a way that this is possible:
```suggestion
      checkDisallowedVotes votes DisallowedVotesDuringBootstrap $ \ gas ->
        \case
          DRepVoter {} | gasAction gas == InfoAction -> True
          _ -> isBootstrapAction $ gasAction gas
  | otherwise = pure ()
```

Note swap of arguments of predicate failure and `canVoteOn`, which allows it to become a lambda.

Also swapping arguments of `canVoteOn` is useful for lambda case (the reason why you used flip). This flipping of arguments sort of makes the function read `canBeVotedOnBy` :slightly_smiling_face: 

Same transformation can be then done to `checkVotersAreValid` and `checkVotesAreNotForExpiredActions`

lehins (commented):
This is looking great!!!

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Gov.hs
teodanciu: Definitely liking `fmap` over `second`  (which is somewhat a mystery to me), thanks!

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Gov.hs
teodanciu: That's wonderful, thank you!

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp.hs
```diff
@@ -58,11 +61,21 @@ spec ::
...
+      Enact.spec @era
+      Epoch.spec @era
+      Gov.spec @era
+      GovCert.spec @era
+      Utxo.spec @era
+      Utxos.spec @era
+      Ratify.spec @era
+  describe "ConwayImpSpec - bootstrap phase (protocol version 9)" $
+    withImpState @era $ do
+      Enact.relevantDuringBootstrapSpec @era
```
teodanciu, line 75: Not too happy with this clunky name.. I'd love to rename to something better

Soupstraw: why not just `bootstrapSpec`?

teodanciu: It's just that it sounds like it's testing the bootstrap itself. I guess in a way it is.

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Gov.hs
```diff
@@ -256,10 +265,9 @@ checkVotesAreNotForExpiredActions ::
   EpochNo ->
   [(Voter (EraCrypto era), GovActionState era)] ->
   Test (ConwayGovPredFailure era)
-checkVotesAreNotForExpiredActions curEpoch voters =
-  let votesOnExpiredActions = filter (\(_, GovActionState {gasExpiresAfter}) -> curEpoch > gasExpiresAfter) voters
-   in failureOnNonEmpty votesOnExpiredActions $
-        VotingOnExpiredGovAction . fmap (second gasId)
+checkVotesAreNotForExpiredActions curEpoch votes =
+  checkDisallowedVotes votes VotingOnExpiredGovAction $ \GovActionState {gasExpiresAfter} ->
+    \case _ -> curEpoch <= gasExpiresAfter
```
lehins: Slight simplication
```suggestion
  checkDisallowedVotes votes VotingOnExpiredGovAction $ \GovActionState {gasExpiresAfter} _ ->
    curEpoch <= gasExpiresAfter
```
Alternatively:
```haskell
  checkDisallowedVotes votes VotingOnExpiredGovAction $ \GovActionState {gasExpiresAfter} ->
    const (curEpoch <= gasExpiresAfter)
```

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp.hs
```diff
@@ -58,11 +61,21 @@ spec ::
...
+      Epoch.relevantDuringBootstrapSpec @era
+      Gov.relevantDuringBootstrapSpec @era
+      GovCert.relevantDuringBootstrapSpec @era
+      Utxos.relevantDuringBootstrapSpec @era
+      Utxo.spec @era
```
lehins: I am probably being too pedantic, but could you please order these in the same as the non-bootstrap specs:
```suggestion
      Utxo.spec @era
      Utxos.relevantDuringBootstrapSpec @era
```

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/EnactSpec.hs
```diff
@@ -47,11 +50,18 @@ spec ::
...
     constitutionSpec
-    actionPrioritySpec
+    actionPriorityCommitteePurposeSpec
+    hardForkInitiationSpec
+
+relevantDuringBootstrapSpec ::
+  ConwayEraImp era =>
+  SpecWith (ImpTestState era)
+relevantDuringBootstrapSpec = do
+  actionPrioritySpec
```
lehins, line 64: We do need some tests that check `hardForkInitiation` during bootstrap. It is in fact more relevant for bootstrap, rather than post boostrap at this point

lehins (approved):
This looks great! Thank you @teodanciu !!

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp.hs
teodanciu: 100%, it's bugging me too now that you pointed it out. Done

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/EnactSpec.hs
teodanciu: Good point, I added a test with only SPOs voting  to the bootstrap section of this test.

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/EnactSpec.hs
```diff
@@ -369,64 +391,82 @@ actionPrioritySpec =
...
         )
         gaids
       passNEpochs 2
-      getsNES (nesEsL . curPParamsEpochStateL . ppDRepDepositL)
+      getsNES (nesEsL . curPParamsEpochStateL . ppMinFeeAL)
         `shouldReturn` val1
       expectNoCurrentProposals
+
+protVer :: EraGov era => ImpTestM era ProtVer
+protVer = getsNES $ nesEsL . curPParamsEpochStateL . ppProtocolVersionL
```
lehins: This one definitely needs a better name. Looked like a local binding. Something that indicates that it is a monadic action that will be invoked each time.
```suggestion
getCurProtVerM :: EraGov era => ImpTestM era ProtVer
getCurProtVerM = getsNES $ nesEsL . curPParamsEpochStateL . ppProtocolVersionL
```

eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Gov.hs
teodanciu: Nice, thanks

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/EnactSpec.hs
teodanciu: Maybe worth moving to ShellyImpTest, it's used a lot in these tests

PR #4276, opened 2024-04-18
Use a separate job for each test suite in GitHub CI

### LINKED ISSUE #4039: Create separate GithubActions jobs for each test suite
Current runs on CI as split on per GHC basis. However, all test suites are being executed in parallel in the same run, which has some serious drawbacks:

* Output of test suites is interleaved, which can be very confusing
* There are thousands lines of output which makes it hard to find a failing test
* It is hard to infer which test suite the test failure is coming from

The ideal solution to this would be to execute each test suite as a parallel job that would be clearly visible in the UI. The acceptance requirements for this task are:
* All test suites should run in parallel
* All test suites should be executed and only after the execution is done the decision is made whether to pass or fail CI. Naturally, a failure of any test suite would result in the whole build failure
* Addition of a test suite should not go unnoticed. In other words CI should not be allowed to pass if there is a test suite in the repo that has not been executed. I believe `hie.yaml` could be leveraged somehow to enforce this,  but there are might be other ways by using `cabal` or something.

neilmayhew:
Closes #4039

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/.github/workflows/haskell.yml
+++ b/.github/workflows/haskell.yml
@@ -89,7 +89,6 @@ jobs:
       with:
         path: |
           ${{ steps.setup-haskell.outputs.cabal-store }}
-          dist-newstyle
         # cache is invalidated upon a change to cabal.project (and/or cabal.project.local), a bump to
         # CABAL_CACHE_VERSION or after a week of inactivity
         key: cache-${{ runner.os }}-${{ matrix.ghc }}-${{ env.CABAL_CACHE_VERSION }}-${{ hashFiles('cabal.project*') }}
@@ -102,6 +101,17 @@ jobs:
 
     - name: Build
       run: cabal build all
+
+    # Pass working directory to test jobs
+    - name: Archive working directory
+      run: tar --format pax -cf /var/tmp/state.tar . && mv /var/tmp/state.tar .
+    - name: Upload working directory archive
+      uses: actions/upload-artifact@v4
+      with:
+        name: state-${{ matrix.ghc }}-${{ matrix.os }}
+        path: state.tar
+        overwrite: true
+        retention-days: 1
 
   test:
     needs: build
@@ -158,8 +168,6 @@ jobs:
         shell: bash
 
     steps:
-    - uses: actions/checkout@v4
-
     - name: Install Haskell
       uses: input-output-hk/setup-haskell@v1
       id: setup-haskell
@@ -181,13 +189,6 @@ jobs:
       run: |
         gem install cddl -v 0.10.3
         gem install cbor-diag
-
-    - name: Configure to use libsodium
-      run: |
-        cat >> cabal.project <<EOF
-        package cardano-crypto-praos
-          flags: -external-libsodium-vrf
-        EOF
 
     - name: Download Mainnet Mirror
       run: |
@@ -212,11 +213,13 @@ jobs:
       run: |
         echo "NIGHTLY=true" >> $GITHUB_ENV
 
-    - name: Cabal update
-      run: cabal update
-
-    - name: Cabal configure
-      run: cabal configure --enable-tests --enable-benchmarks --write-ghc-environment-files=always
+    # Retrieve working directory from build jobs
+    - name: Download working directory archive
+      uses: actions/download-artifact@v4
+      with:
+        name: state-${{ matrix.ghc }}-${{ matrix.os }}
+    - name: Unarchive working directory
+      run: tar -xf state.tar && rm state.tar
 
     - uses: actions/cache@v4
       if: matrix.os != 'macos-latest'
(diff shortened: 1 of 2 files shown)


neilmayhew:
This still does some unnecessary rebuilding, but it's relatively minor so I think we should merge it as it is now, and follow up with an improvement to stop the rebuilding.

lehins:
I am not 100% sure about this, but from the looks of it occasionally test suites are not being rebuild, however during restoration of artifacts permissions are probably not restored correctly, which leads to prebuild executable not having an executable permission for the use and leading to Permission denied error. This is just my hypothesis.
```
Running 1 test suites...
Test suite tests: RUNNING...
/home/runner/work/cardano-ledger/cardano-ledger/dist-newstyle/build/x86_64-linux/ghc-9.8.2/vector-map-1.1.0.0/t/tests/build/tests/tests: createProcess: posix_spawnp: permission denied (Permission denied)
Error: cabal: Tests failed for test:tests from vector-map-1.1.0.0.
```

Besides that I love the Haskell script in this PR and overall approach looks great.

neilmayhew:
> I am not 100% sure about this, but from the looks of it occasionally test suites are not being rebuild, however during restoration of artifacts permissions are probably not restored correctly, which leads to prebuild executable not having an executable permission for the use and leading to Permission denied error. This is just my hypothesis.

If this was the case, I wouldn't expect those tests to have succeeded for ghc 8.10.7 and 9.2.8. Also, I've had everything green on a previous test run that used the artifacts approach. I'll retry those tests and see if they pass. That doesn't mean there isn't a problem, of course, but it will shed more light on the problem.

> Besides that I love the Haskell script in this PR and overall approach looks great.

Thanks!

lehins:
> If this was the case, I wouldn't expect those tests to have succeeded for ghc 8.10.7 and 9.2.8.

Not necessarily. See the build for those ghc versions and you will see that the test suite is being recompiled, eg for ghc-9.2.8: 
```
Building test suite 'non-integral-test' for non-integral-1.0.0.0..
[1 of 2] Compiling Tests.Cardano.Ledger.NonIntegral ( test/Tests/Cardano/Ledger/NonIntegral.hs, /home/runner/work/cardano-ledger/cardano-ledger/dist-newstyle/build/x86_64-linux/ghc-9.2.8/non-integral-1.0.0.0/t/non-integral-test/build/non-integral-test/non-integral-test-tmp/Tests/Cardano/Ledger/NonIntegral.o )
[2 of 2] Compiling Main             ( test/Tests.hs, /home/runner/work/cardano-ledger/cardano-ledger/dist-newstyle/build/x86_64-linux/ghc-9.2.8/non-integral-1.0.0.0/t/non-integral-test/build/non-integral-test/non-integral-test-tmp/Main.o )
Linking /home/runner/work/cardano-ledger/cardano-ledger/dist-newstyle/build/x86_64-linux/ghc-9.2.8/non-integral-1.0.0.0/t/non-integral-test/build/non-integral-test/non-integral-test ...
Running 1 test suites...
Test suite non-integral-test: RUNNING...
```

.github/workflows/haskell.yml
```diff
@@ -106,20 +89,110 @@ jobs:
...
+      uses: actions/upload-artifact@v4
+      with:
+        name: state-${{ matrix.ghc }}-${{ matrix.os }}
+        path: state.tar
+        overwrite: true
+        retention-days: 1
+
+  test:
+    needs: build
+
```
lehins, line 121: We need one more job like this:
```yaml
  complete:
    needs: test
    strategy:
      matrix:
        ghc: ["8.10.7", "9.2.8", "9.6.4", "9.8.2"]
        os: [ubuntu-latest]
```

This way we can use that last job for `master` branch protections rules. Unfortunately github does not support wildcards for branch protection rules :cry: 

I wish also there was an easy way to share the `matrix`. However, the approaches I looked at that people invented looked too complicated and aren't worth the trouble in my opinion.

neilmayhew: I'm not quite sure how that would be useful. You realize that `complete` will wait for all the ghc versions of `test` to complete, and then run four identical copies of itself?

lehins: That's exactly what we need. The important part is that it should not run if any of the tests have failed

neilmayhew:
This is working better now that I'm using a `tar` archive. However, I discovered that the default format (`gnu`) has one-second resolution on modification times, and I think that's why there are still some rebuilds. I'm changing it to use `pax` format which has finer resolution.

.github/workflows/haskell.yml
neilmayhew: In that case, I don't think it needs a matrix

neilmayhew: Also, it needs to have some steps or GH will reject it

lehins: >  I don't think it needs a matrix

Very good point, it'll make it even simpler

Step could be:
```yaml 
run: |
  echo Tada
```

:grin:

neilmayhew: I've added that (but haven't pushed it yet). However, I'm curious why we couldn't just use the `test` job in the branch protection rules instead of adding another one that doesn't do anything.

lehins (approved):
This was a long awaited improvement to our CI. 
Thank you for getting it done!

neilmayhew:
A discussed in meetings, the tests now wait for all of the builds to finish instead of starting when the one with the same ghc version finishes. If we ever want to improve on this, there's a solution [here](https://github.com/orgs/community/discussions/42335#discussioncomment-7997348). (Max pointed me to this discussion.)

PR #4279, opened 2024-04-19
`constrained-generators`: Fix bug in toPreds for maps + add additional tests

MaximilianAlgehed:
# Description

There was a silly bug in toPreds whereby the required elements of the domain would become the universe of the domain when you do `toPreds`. This resulted in maps being constrained to be empty under unfortunate circumstances.

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/libs/constrained-generators/src/Constrained/Spec/Maps.hs
+++ b/libs/constrained-generators/src/Constrained/Spec/Maps.hs
@@ -138,13 +138,12 @@ instance
 
   toPreds m (MapSpec mustKeys mustVals size kvs foldSpec) =
     toPred
-      [ assert $ app (subsetFn @fn) (app (domFn @fn) m) (Lit mustKeys)
+      [ assert $ lit mustKeys `subset_` dom_ m
       , forAll (Lit mustVals) $ \val ->
-          app (elemFn @fn) val (app (rngFn @fn) m)
-      , -- TODO: make nice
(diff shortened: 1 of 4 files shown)


libs/constrained-generators/src/Constrained/Spec/Maps.hs
```diff
@@ -138,13 +138,12 @@ instance
 
   toPreds m (MapSpec mustKeys mustVals size kvs foldSpec) =
     toPred
-      [ assert $ app (subsetFn @fn) (app (domFn @fn) m) (Lit mustKeys)
+      [ assert $ lit mustKeys `subset_` dom_ m
```
TimSheard, line 141: So the bug was that the arguments to subset_  were in the wrong order?

TimSheard (approved):
A one line change, a new rewrite rule, and some tests.
Looks good to me.

libs/constrained-generators/src/Constrained/Spec/Maps.hs
MaximilianAlgehed: indeed... such an easy mistake to make, such a stupid mistake to make...

MaximilianAlgehed: I want to build quickcheck tests for this stuff now - I'm getting tired of not finding bugs in this in spite of the 150 or so examples

PR #4281, opened 2024-04-19
Discount expired CC from CC-size calculation

### LINKED ISSUE #4232: Do not count expired committee members towards the committee size
We need to stop counting expired CC members towards the size of the active Committee size, similarly how we stop counting the resigned ones in #4205

In other words this TODO need to be taken care of: https://github.com/IntersectMBO/cardano-ledger/blob/97963a4e83a18d51d069bf5535ee4e9de2cb3def/eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Internal.hs#L446

And an imp test similar to the one in #4205 needs to be written

We do not remove expired CC members on the epoch boundary. Any member can only be removed with an explicit governance action.

On the epoch boundary we remove hot credential authorizations that do not belong to any existing cold credential, but that is only the modification of the committee state. The actual committee members can only be adjusted with an `UpdateCommittee` proposal.

We also need an Imp test that checks this case

_Originally posted by @lehins in https://github.com/IntersectMBO/cardano-ledger/pull/4205#discussion_r1548820405_

aniketd:
# Description

Resolves #4232 

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Internal.hs
+++ b/eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Internal.hs
@@ -397,9 +397,19 @@ votingStakePoolThresholdInternal pp isElectedCommittee action =
         TreasuryWithdrawals {} -> NoVotingAllowed
         InfoAction {} -> NoVotingThreshold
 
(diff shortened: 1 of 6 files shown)


lehins (approved):
Beautiful!
Thank you!

PR #4283, opened 2024-04-19
`constrained-generators`: add new test to test suite

MaximilianAlgehed:
# Description

@TimSheard had a failure in this test on an old branch. It was fixed recently but I think it's none the less a nice minimal test for an old bug so we should keep it around to avoid regressions.

# Checklist

- [x] Self-reviewed the diff


TimSheard (approved):
Always good to have more tests, that test something that was missing.

PR #4284, opened 2024-04-19
Flexible costmodel params

### LINKED ISSUE #2902: Problematic cost model serialization

lehins:
# Description

Fixes in this PR:
* Pump up to Plutus 1.27
* Revert changes in 508f0919c86d03bff9b0a40e31edc1cdd7549637 due to accidental change in Plutus: https://github.com/IntersectMBO/plutus/pull/5932
* Change type for CostModel parameters to `Int64` instead of `Integer`. See an accompanied change in plutus: https://github.com/IntersectMBO/plutus/pull/5920
* Fixes #2902 by allowing variable number of cost model parameters once we are in Conway. See relevant change on the plutus side: https://github.com/IntersectMBO/plutus/pull/5912

# Checklist

- [x] Self-reviewed the diff


libs/cardano-ledger-core/src/Cardano/Ledger/Plutus/CostModels.hs
```diff
@@ -145,11 +135,12 @@ instance FromJSON CostModels where
     cms <- mapM (parseCostModel o) nonNativeLanguages
     let cmsMap = Map.fromList [(cmLanguage cm, cm) | Just cm <- cms]
     unknown <- o .:? "Unknown" .!= mempty
-    pure $ mkCostModels cmsMap <> mkCostModelsLenient unknown
+    unknwonCostmodels <- mkCostModelsLenient unknown
+    pure $ mkCostModels cmsMap <> unknwonCostmodels
```
neilmayhew:
```suggestion
    unknownCostmodels <- mkCostModelsLenient unknown
    pure $ mkCostModels cmsMap <> unknownCostmodels
```

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/Conway/Instances.hs
```diff
@@ -879,23 +879,26 @@ instance IsConwayUniv fn => HasSpec fn CoinPerByte
...
+  genFromTypeSpec _ = pureGen arbitrary
+  shrinkWithTypeSpec _ = shrink
+  conformsTo _ _ = True
+  toPreds _ _ = toPred True
+
+-- FIXME: previous implementation no longer builds
+-- emptySpec =
+--   Cartesian
+--     (constrained $ \m -> size_ (dom_ m) <=. 3)
+--     (constrained $ \p -> size_ (dom_ (fst_ $ fromGeneric_ p)) <=. 3)
```
lehins, line 894: @MaximilianAlgehed I am not so sure why my change worked and I have no idea if it is even correct, but all of the tests seem to be happy about the change. I did it by example of other instances.
Could you please check this change whenever you can and potentially fix it if I screwed it up.

MaximilianAlgehed: For now I think it works because we try to avoid generating cost models because they are insanely expensive to generate. So for now this shouldn't be a problem. However, I can have a closer look at fixing this ASAP. For now this shouldn't break anything.

neilmayhew (commented):
So far all I've seen is this typo

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/Conway/Instances.hs
MaximilianAlgehed: Possibly instead of `arbitrary` you want to use `genValidCostModels` - ther was a comment elsewhere about Arbitrary not working for Conway?

lehins: >  ther was a comment elsewhere about Arbitrary not working for Conway?

That comment is only relevant for serialization, so using `arbitrary` should be fine here

>  I can have a closer look at fixing this ASAP.

There is no rush on it. Whenever you get a chance

eras/alonzo/impl/testlib/Test/Cardano/Ledger/Alonzo/Binary/CostModelsSpec.hs
```diff
@@ -62,38 +59,35 @@ validCostModelProp = do
...
             \cmDecoded ppuDecoded -> do
-              -- pre-Conway we are failing when deserializing invalid costmodels
+              -- pre-Conway we are failing when deserializing underspecified costmodels
               if eraProtVerHigh @era < natVersion @9
-                then expectDeserialiseFailure cmDecoded (Just "CostModels")
+                then do
+                  expectDeserialiseFailure cmDecoded (Just "CostModels")
+                  expectDeserialiseFailure ppuDecoded Nothing
                 else do
                   -- post-Conway, we are collecting CostModels deserialization errors
```
zliu41: This comment should be updated. AFAICT you are verifying that the deserialization succeeded rather than errors being collected.

eras/conway/impl/src/Cardano/Ledger/Conway/PParams.hs
```diff
@@ -113,7 +113,6 @@ import Cardano.Ledger.HKD (
 import Cardano.Ledger.Plutus.CostModels (
   CostModel,
   decodeCostModelFailHard,
```
zliu41: I see that `decodeCostModelFailHard` is still used in this module as well as `Cardano.Ledger.Plutus.Evaluate`. Why do we still need this in Conway?

eras/conway/impl/src/Cardano/Ledger/Conway/PParams.hs
```diff
@@ -113,7 +113,6 @@ import Cardano.Ledger.HKD (
...
-  decodeValidAndUnknownCostModels,
```
zliu41, line 116: Is this function still needed at all?

libs/cardano-ledger-core/src/Cardano/Ledger/Plutus/ToPlutusData.hs
```diff
@@ -77,8 +78,10 @@ instance ToPlutusData NonNegativeInterval where
   fromPlutusData _ = Nothing
 
 instance ToPlutusData CostModels where
-  toPlutusData cmdls = toPlutusData (flattenCostModels cmdls)
-  fromPlutusData x = mkCostModelsLenient <$> fromPlutusData x
+  toPlutusData costModels = toPlutusData $ fmap toInteger <$> flattenCostModels costModels
+  fromPlutusData costModelsData = do
+    costModels :: Map.Map Word8 [Integer] <- fromPlutusData costModelsData
+    mkCostModelsLenient (Map.map (map fromInteger) costModels)
```
zliu41, line 84: Is the `fromInteger` safe wrt overflows?

eras/conway/impl/src/Cardano/Ledger/Conway/PParams.hs
lehins: It's been deprecated.

libs/cardano-ledger-core/src/Cardano/Ledger/Plutus/ToPlutusData.hs
lehins: Yes, because `fromPlutusData` is only used for testing. There will never be a case when we need to translate PlutusContext back to the ledger representation, other than for testing the forward translation

eras/conway/impl/src/Cardano/Ledger/Conway/PParams.hs
lehins: I think the name is a bit too harsh sounding, because it will work fine with variable number of parameters. This is a decoder that will fail for other reasons and it only works for known plutus language versions. We do need it in the ConwayGenesis (usage in `Cardano.Ledger.Conway.PParams` module) and during debugging plutus scripts (usage in `Cardano.Ledger.Plutus.Evaluate`) because at those.

I think I should just rename it to something less scary sounding, something like: `decodeCostModel` :smile:

eras/alonzo/impl/testlib/Test/Cardano/Ledger/Alonzo/Binary/CostModelsSpec.hs
lehins: Yes, errors are no longer collected at all, because `mkEvaluationContext` does not fail anymore for unexpected number of parameters
Comment fixed.

libs/cardano-ledger-core/src/Cardano/Ledger/Plutus/CostModels.hs
lehins: Fixed. Thank you

eras/conway/impl/src/Cardano/Ledger/Conway/PParams.hs
lehins: `decodeCostModelFailHard` is now deprecated in favor of `decodeCostModel`

libs/cardano-ledger-core/src/Cardano/Ledger/Plutus/CostModels.hs
```diff
@@ -115,7 +106,13 @@ import qualified PlutusLedgerApi.V3 as PV3 (ParamName, mkEvaluationContext)
...
 data CostModel = CostModel
   { cmLanguage :: !Language
-  , cmValues :: ![Integer]
+  , cmValues :: ![Int64]
+  -- ^ We retain the orignal values for parameters for the purpose of
+  -- reserialization. Starting with Conway era there might not be an exact amount of
+  -- parameters in this list than what Plutus' smart constructor `mkEvaluationContext`
+  -- expects. This functionality is intentional for allowing the addition of new
+  -- primitives on a hardfork boundary. In case when less than expected number is
+  -- supplied, `maxBound` will be used instead by the Plutus smart constructor.
```
neilmayhew:
```suggestion
  -- ^ We retain the original values for parameters for the purpose of
  -- reserialization. Starting with the Conway era there might not be the exact number
  -- of parameters in this list that Plutus' smart constructor `mkEvaluationContext`
  -- expects. This functionality is intentional for allowing the addition of new
  -- primitives on a hardfork boundary. When less than the expected number is
  -- supplied, `maxBound` will be used instead by the Plutus smart constructor.
```

neilmayhew (approved):
This looks good to me, and I think I understand the changes well enough.

I've made a minor suggestion for improving the wording of a comment, but feel free to ignore this if you don't want to delay merging.

PR #4285, opened 2024-04-21
Start Conway Imp tests with an initial committee and constitution

teodanciu:
# Description

This PR is setting an initial committee and constitution for Conway (emulating the ones that  will be defined in genesis). 
For this, it was necessary to slightly modify the initialization of the NewEpochState in the Imp framework, to allow initialization actions before returning the state. 

Tests were adjusted to only elect a committee if this action is actually meaningful to the test (relying on the initial committee otherwise). 
Going through many tests was an opportunity to make some small  improvements: 
 * an attempt to be more disciplined with setting pparams. I have consolidated a few of them 
 * updating some functions to reduce the noise of converting between KeyHash and Credential
 * removing the duplicated inclusion of a test 
 

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/ImpTest.hs
+++ b/eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/ImpTest.hs
@@ -186,7 +186,6 @@ import Data.List.NonEmpty (NonEmpty)
 import qualified Data.List.NonEmpty as NE
 import qualified Data.Map.Strict as Map
 import Data.Maybe (fromJust, isJust)
-import Data.Maybe.Strict (isSJust)
 import Data.Sequence.Strict (StrictSeq (..))
 import qualified Data.Sequence.Strict as SSeq
 import qualified Data.Set as Set
@@ -927,7 +926,7 @@ getRatifyEnvAndState = do
   pure (ratifyEnv, ratifyState)
 
 -- | Checks whether the governance action has enough DRep votes to be accepted in the next
--- epoch. (Note that no other checks execept DRep votes is used)
+-- epoch. (Note that no other checks except DRep votes are used)
 isDRepAccepted ::
   (HasCallStack, ConwayEraGov era, ConwayEraPParams era) =>
   GovActionId (EraCrypto era) ->
@@ -1095,26 +1094,8 @@ electBasicCommittee = do
       , UpdateCommittee SNothing mempty mempty (1 %! 10)
       ]
   submitYesVote_ (DRepVoter $ KeyHashObj drepKH) gaidCommitteeProp
-
-  let
-    assertNoCommittee = do
-      committee <-
-        getsNES $
-          nesEsL . esLStateL . lsUTxOStateL . utxosGovStateL . committeeGovStateL
-      impAnn "There should not be a committee" $ committee `shouldBe` SNothing
-  logRatificationChecks gaidCommitteeProp
-  assertNoCommittee
-
   passEpoch
-  logRatificationChecks gaidCommitteeProp
-  assertNoCommittee
   passEpoch
-  do
-    committee <-
-      getsNES $
-        nesEsL . esLStateL . lsUTxOStateL . utxosGovStateL . committeeGovStateL
-    impAnn "There should be a committee" $ committee `shouldSatisfy` isSJust
-
   hotCommitteeC <- registerCommitteeHotKey coldCommitteeC
   pure (KeyHashObj drepKH, hotCommitteeC, GovPurposeId gaidCommitteeProp)
 
--- a/eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/ImpTest.hs
+++ b/eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/ImpTest.hs
@@ -236,6 +236,17 @@ instance
               (initAlonzoImpNES nes)
                 & nesEsL . curPParamsEpochStateL . ppDRepActivityL .~ EpochInterval 100
                 & nesEsL . curPParamsEpochStateL . ppGovActionLifetimeL .~ EpochInterval 30
+                & nesEsL . curPParamsEpochStateL . ppGovActionDepositL .~ Coin 123
+                & nesEsL . curPParamsEpochStateL . ppCommitteeMaxTermLengthL .~ EpochInterval 20
+                & nesEsL . curPParamsEpochStateL . ppCommitteeMinSizeL .~ 1
+                & nesEsL . curPParamsEpochStateL . ppDRepVotingThresholdsL
+                  %~ ( \dvt ->
+                        dvt
+                          { dvtCommitteeNormal = 1 %! 1
+                          , dvtCommitteeNoConfidence = 1 %! 2
+                          , dvtUpdateToConstitution = 1 %! 2
+                          }
+                     )
                 & nesEsL . epochStateGovStateL . committeeGovStateL .~ SJust committee
             epochState = newNes ^. nesEsL
             ratifyState =
@@ -1063,20 +1074,7 @@ electBasicCommittee ::
     , GovPurposeId 'CommitteePurpose era
     )
 electBasicCommittee = do
-  logEntry "Setting up PParams and DRep"
-  modifyPParams $ \pp ->
-    pp
-      & ppDRepVotingThresholdsL
-        %~ ( \dvt ->
-              dvt
-                { dvtCommitteeNormal = 1 %! 1
-                , dvtCommitteeNoConfidence = 1 %! 2
-                , dvtUpdateToConstitution = 1 %! 2
-                }
-           )
-      & ppCommitteeMaxTermLengthL .~ EpochInterval 20
-      & ppGovActionLifetimeL .~ EpochInterval 2
-      & ppGovActionDepositL .~ Coin 123
+  logEntry "Setting up a DRep"
   (drep, _, _) <- setupSingleDRep 1_000_000
 
   logEntry "Registering committee member"
--- a/eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/ImpTest.hs
+++ b/eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/ImpTest.hs
@@ -192,6 +192,7 @@ import qualified Data.Text as T
 import Data.Tree
 import qualified GHC.Exts as GHC (fromList)
 import Lens.Micro
+import Lens.Micro.Mtl ((%=))
 import Test.Cardano.Ledger.Babbage.ImpTest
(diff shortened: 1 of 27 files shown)


eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/ImpTest.hs
```diff
@@ -212,6 +214,18 @@ conwayModifyPParams f = modifyNES $ \nes ->
         (snapshot, ratifyState) ->
           DRComplete snapshot (ratifyState & rsEnactStateL . ensCurPParamsL %~ f)
 
+initialCommitteeKeyPair :: Crypto c => KeyPair 'ColdCommitteeRole c
+initialCommitteeKeyPair = mkKeyPair 30
+
+initialCommitteeCred :: forall c. Crypto c => Credential 'ColdCommitteeRole c
+initialCommitteeCred = KeyHashObj (hashKey @c (vKey initialCommitteeKeyPair))
```
lehins: Global bindings like that are not a good idea.

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/ImpTest.hs
```diff
@@ -222,16 +236,21 @@ instance
...
+            & nesEsL . curPParamsEpochStateL . ppCommitteeMaxTermLengthL .~ EpochInterval 20
+            & nesEsL . epochStateGovStateL . committeeGovStateL .~ SJust initialCommittee
+        -- TODO: configure sensible voting thresholds, rather than defaulting to zero
+        -- in order to make voting in tests relevant
         epochState = nes ^. nesEsL
         ratifyState =
           def
             & rsEnactStateL .~ mkEnactState (epochState ^. epochStateGovStateL)
-     in nes & nesEsL .~ setCompleteDRepPulsingState def ratifyState epochState
+    void $ addKeyPair initialCommitteeKeyPair
```
lehins: Let's generate it at random. You'll just need to add `MonadGen`  constraint to `initImpNES`
```suggestion
    committeeKeyPair <- freshKeyPair
```

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/ImpTest.hs
```diff
@@ -268,6 +287,19 @@ instance
   ) =>
   ConwayEraImp (ConwayEra c)
 
+registerInitialCommittee ::
+  forall era.
+  ConwayEraImp era =>
+  ImpTestM era (NonEmpty (Credential 'HotCommitteeRole (EraCrypto era)))
+registerInitialCommittee = do
+  hk <- registerCommitteeHotKey initialCommitteeCred
```
lehins: Global bindings like that is a bad idea. Look up the committee cold credential in the state instead

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/ImpTest.hs
```diff
@@ -268,6 +287,19 @@ instance
...
+  pure $ hk :| []
+
+getInitialCommitteMembers ::
+  Era era =>
+  ImpTestM era (NonEmpty (Credential 'ColdCommitteeRole (EraCrypto era)))
+getInitialCommitteMembers = pure $ initialCommitteeCred :| []
```
lehins: Ditto. Look it up in the state

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/ImpTest.hs
```diff
@@ -212,6 +214,18 @@ conwayModifyPParams f = modifyNES $ \nes ->
         (snapshot, ratifyState) ->
           DRComplete snapshot (ratifyState & rsEnactStateL . ensCurPParamsL %~ f)
 
+initialCommitteeKeyPair :: Crypto c => KeyPair 'ColdCommitteeRole c
+initialCommitteeKeyPair = mkKeyPair 30
```
lehins: These can be generated randomly, no need to use this hack of relying on an integer.

See my comment below on how to do that

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/ImpTest.hs
```diff
@@ -212,6 +214,18 @@ conwayModifyPParams f = modifyNES $ \nes ->
...
+
+initialCommitteeCred :: forall c. Crypto c => Credential 'ColdCommitteeRole c
+initialCommitteeCred = KeyHashObj (hashKey @c (vKey initialCommitteeKeyPair))
+
+initialCommittee :: forall era. Era era => Committee era
+initialCommittee =
```
lehins: Ditto. Global binding that needs to be removed

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/ImpTest.hs
```diff
@@ -997,21 +1029,8 @@ electBasicCommittee ::
...
-  logEntry "Setting up PParams and DRep"
-  modifyPParams $ \pp ->
-    pp
-      & ppDRepVotingThresholdsL
-        %~ ( \dvt ->
-              dvt
-                { dvtCommitteeNormal = 1 %! 1
-                , dvtCommitteeNoConfidence = 1 %! 2
-                , dvtUpdateToConstitution = 1 %! 2
-                }
```
lehins, line 1050: Looks like we lost all of those threshold in the setup of every test that used to `electBasicCommittee`? Are they set somewhere else that I am missing?

eras/shelley/impl/testlib/Test/Cardano/Ledger/Shelley/ImpTest.hs
```diff
@@ -375,7 +376,8 @@ class
   ) =>
   ShelleyEraImp era
   where
-  initImpNES :: NewEpochState era
+  initImpNES ::
+    (Era era, MonadState (ImpTestState era) m) => m (NewEpochState era)
```
lehins: I was thinking more about it and this turns out to be a lot more general function. So, let's rename it and change its semantics to reflect that:
```suggestion
  initImpTestState :: (MonadState (ImpTestState era) m, MonadGen m) => m ()
```

Important notes:

* `Era` is redundant
* `MonadGen` is useful, as mentioned in my other comment
* This action should not return anything, but instead should modify the state with all the initialization that needs to be done.

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/ImpTest.hs
```diff
@@ -222,16 +236,21 @@ instance
...
   where
-  initImpNES =
+  initImpNES = do
     let nes =
           initAlonzoImpNES
             & nesEsL . curPParamsEpochStateL . ppDRepActivityL .~ EpochInterval 100
             & nesEsL . curPParamsEpochStateL . ppGovActionLifetimeL .~ EpochInterval 30
+            & nesEsL . curPParamsEpochStateL . ppCommitteeMaxTermLengthL .~ EpochInterval 20
+            & nesEsL . epochStateGovStateL . committeeGovStateL .~ SJust initialCommittee
+        -- TODO: configure sensible voting thresholds, rather than defaulting to zero
```
lehins: From the looks of it, this PR relaxes a bunch of tests by removing thresholds, which used to be set by a call to `electBasicCommitee`.

I think we need to take care of this TODO in this PR or at the very least make sure we have the same thresholds in tests as before this PR (which probably would mean resetting them to zero in some tests that relied on zero thresholds)

lehins (changes requested):
Looks good, but committee initialization is a bit hacky and it seems that thresholds got relaxed for many tests

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/ImpTest.hs
teodanciu: Indeed! I forgot that these were set here!  Will make sure they are set, thanks!

teodanciu:
* Removed global bindings for committee, and generated them with `freshKeyPair` instead
* Set voting thresholds when initializing and adjusted the tests accordingly
* Added initial constitution 

If you want to have another look @lehins when you can, thanks!

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/ImpTest.hs
```diff
@@ -224,16 +227,36 @@ instance
...
-          initAlonzoImpNES
-            & nesEsL . curPParamsEpochStateL . ppDRepActivityL .~ EpochInterval 100
-            & nesEsL . curPParamsEpochStateL . ppGovActionLifetimeL .~ EpochInterval 30
-        epochState = nes ^. nesEsL
-        ratifyState =
-          def
-            & rsEnactStateL .~ mkEnactState (epochState ^. epochStateGovStateL)
-     in nes & nesEsL .~ setCompleteDRepPulsingState def ratifyState epochState
+  initImpTestState =
+    do
```
lehins: We can make this change in a subsequent PR. Just pointing it out as an inconsistency :wink: 
```suggestion
  initImpTestState = do
```

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/ImpTest.hs
```diff
@@ -224,16 +227,36 @@ instance
...
+      let committee = Committee [(KeyHashObj kh, EpochNo 15)] (1 %! 1)
+      anchor <- arbitrary
+      let constitution = Constitution anchor SNothing
+      impNESL %= initConwayNES committee constitution
+    where
+      initConwayNES committee constitution nes =
+        let newNes =
+              (initAlonzoImpNES nes)
+                & nesEsL . curPParamsEpochStateL . ppDRepActivityL .~ EpochInterval 100
+                & nesEsL . curPParamsEpochStateL . ppGovActionLifetimeL .~ EpochInterval 30
```
lehins, line 241: `electBasicCommittee` used to also set the deposit:
```suggestion
                & nesEsL . curPParamsEpochStateL . ppGovActionLifetimeL .~ EpochInterval 30
                & nesEsL . curPParamsEpochStateL . ppGovActionDepositL .~ Coin 123
```

lehins (approved):
Looks great. Thank you!

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/ImpTest.hs
teodanciu: Fixed, thanks!

PR #4286, opened 2024-04-22
`constrained-generators`: refactor `reify` to reduce the number of binding sites + delay simplification more to avoid variable capture in higher order syntax

MaximilianAlgehed:
# Description

This should unblock @TimSheard 

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/libs/constrained-generators/src/Constrained/Base.hs
+++ b/libs/constrained-generators/src/Constrained/Base.hs
@@ -777,7 +777,7 @@ conformsToSpecM a (SuspendedSpec v ps) = guard =<< checkPred (singletonEnv v a) 
 conformsToSpecM _ (ErrorSpec es) = explain es $ guard False
 
 conformsToSpecProp :: forall fn a. HasSpec fn a => a -> Specification fn a -> Property
-conformsToSpecProp a s = fromGEProp $ conformsToSpecM a s
+conformsToSpecProp a s = fromGEProp $ conformsToSpecM a (simplifySpec s)
 
 conformsToSpec :: forall fn a. HasSpec fn a => a -> Specification fn a -> Bool
 conformsToSpec a s = isOk $ conformsToSpecM a s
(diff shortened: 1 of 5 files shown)


libs/constrained-generators/src/Constrained/Base.hs
```diff
@@ -3763,8 +3772,14 @@ reify ::
...
-reify tm f body =
-  mkReify tm f (bind body)
+reify t f body =
+  exists (\eval -> pure $ f (eval t)) $ \x ->
+    [ reifies x t f
+    , toPred $ body x
+    ]
+
+reifies :: (HasSpec fn a, HasSpec fn b) => Term fn b -> Term fn a -> (a -> b) -> Pred fn
+reifies = Reifies
```
TimSheard, line 3782: I guess this is the heart of the matter. 'reify' has the same type that it had before. Except it is implemented in terms of 
exists and the new Pred 'Reifies'

libs/constrained-generators/src/Constrained/Base.hs
```diff
@@ -1693,12 +1700,12 @@ aggressiveInlining p
...
-        | not (isLit tm)
-        , Lit a <- substituteAndSimplifyTerm sub tm -> do
+      Reifies t' t f
+        | not (isLit t)
+        , Lit a <- substituteAndSimplifyTerm sub t -> do
             tell $ Any True
-            pure $ mkReify (Lit a) f b
-        | otherwise -> Reify tm f <$> goBinder fvs sub b
+            pure $ Reifies t' (Lit a) f
+        | otherwise -> pure $ Reifies t' t f
```
TimSheard, line 1708: Do we have an argument, that eventually the second argument 't' will be a literal?

I assume the new version of simplifyTerm, will try and make that happen?

libs/constrained-generators/src/Constrained/Base.hs
```diff
@@ -1536,7 +1544,7 @@ simplifyPred = \case
     t' -> GenHint h t'
   Subst x t p -> simplifyPred $ substitutePred x t p
   Assert es t -> assertExplain es (simplifyTerm t)
-  Reify tm f b -> mkReify (simplifyTerm tm) f (simplifyBinder b)
+  Reifies t' t f -> Reifies (simplifyTerm t') (simplifyTerm t) f
```
TimSheard, line 1547: This is where we hope (simplifyTerm t)  will eventually be come a literal?

libs/constrained-generators/src/Constrained/Base.hs
```diff
@@ -1432,7 +1440,7 @@ substitutePred x tm = \case
   Let t b -> Let (substituteTerm [x := tm] t) (substituteBinder x tm b)
   ForAll t b -> ForAll (substituteTerm [x := tm] t) (substituteBinder x tm b)
   Case t bs -> Case (substituteTerm [x := tm] t) (mapList (substituteBinder x tm) bs)
-  Reify t f b -> Reify (substituteTerm [x := tm] t) f (substituteBinder x tm b)
+  Reifies t' t f -> Reifies (substituteTerm [x := tm] t') (substituteTerm [x := tm] t) f
```
TimSheard, line 1443: Again we hope the substituion makes it a Literal.

libs/constrained-generators/src/Constrained/Base.hs
```diff
@@ -1006,6 +1006,14 @@ linearize preds graph = do
...
 ------------------------------------------------------------------------
 
+simplifySpec :: HasSpec fn a => Specification fn a -> Specification fn a
+simplifySpec = \case
+  SuspendedSpec x p -> computeSpecSimplified x (optimisePred p)
+  MemberSpec xs -> MemberSpec (nub xs)
+  ErrorSpec es -> ErrorSpec es
+  TypeSpec ts cant -> TypeSpec ts (nub cant)
+  TrueSpec -> TrueSpec
+
```
TimSheard, line 1016: From my observation of the changes.  This simplify spec is used as a view pattern in a few places. Is it used elsewhere?

TimSheard (approved):
Not 100% sure I understand this. I left some comments in what I think are key places. Maybe some additional explanation in those places might help.

libs/constrained-generators/src/Constrained/Base.hs
MaximilianAlgehed: Yes, and the point of that is that it reduces the number of different binding-forms that we have to consider. This simplifies dependency resolution and allows examples like `reifyYucky` to pass.

libs/constrained-generators/src/Constrained/Base.hs
MaximilianAlgehed: Yes, that's right.

libs/constrained-generators/src/Constrained/Base.hs
MaximilianAlgehed: Only if `t` is a variable, the case that is supposed to discharge the `reify` is in `simplifySpec`.

MaximilianAlgehed: Sorry, `simplifyPred`.

libs/constrained-generators/src/Constrained/Base.hs
MaximilianAlgehed: Basically nowhere else.

libs/constrained-generators/src/Constrained/Base.hs
MaximilianAlgehed: Yes, it will become a literal before solving because the dependency order specifies basically `dependsOn t t'`.

PR #4288, opened 2024-04-23
Fix burning tokens predicate failure

lehins:
# Description

## Problem description

When transaction does not balance out in presence of burning the user submitting the transaction will get `DeserialiseFailure 56 "expected word"` deserialization failure instead of `ValueNotConservedUTxO` predicate failure

## Fixes in this PR

* `ValueNotConservedUTxO` cannot handle negative values, therefore we need to fix how this failure is being reported. Instead of reporting negative values for burned tokens in consumed, we change it to report it as positive in the produced.

* Ensure every predicate failure produced in the Imp test is also tested for serialization issues.

* Reorganize a bit how Imp tests are being invoked. This PR changes it in such a way that current era is responsible for calling imp tests from the previous era. This ensures that  we are not gonna forget to run tests from some previous era in a new era.
 
# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/eras/shelley/impl/testlib/Test/Cardano/Ledger/Shelley/ImpTest.hs
+++ b/eras/shelley/impl/testlib/Test/Cardano/Ledger/Shelley/ImpTest.hs
@@ -324,6 +324,7 @@ class
   , ShelleyEraTxCert era
   , ToExpr (Tx era)
   , NFData (Tx era)
+  , ToExpr (TxBody era)
   , ToExpr (TxOut era)
   , ToExpr (Value era)
   , ToExpr (PParams era)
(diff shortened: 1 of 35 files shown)


eras/shelley/impl/testlib/Test/Cardano/Ledger/Shelley/ImpTest.hs
```diff
@@ -917,25 +919,30 @@ trySubmitTx tx = do
...
-    let newRoot
-          | Map.member assumedNewRoot utxo = assumedNewRoot
-          | Map.member impRootTxIn utxo = impRootTxIn
-          | otherwise = error "Root not found in UTxO"
-    impRootTxInL .= newRoot
-    pure txFixed
+  case res of
+    Left predFailures -> do
+      -- Verify that produced predicate failures are ready for the node-to-client protocol
+      liftIO $ forM_ predFailures $ roundTripEraExpectation @era
```
aniketd, line 925: :tada:

aniketd (approved):
Excellent! :bow: :raised_hands:

PR #4292, opened 2024-04-24
`constrained-generators`: add genHint for maps

MaximilianAlgehed:
# Description

This adds `genHint` to `Map`s to make it easier to control the distribution of maps to avoid blowup.

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/libs/constrained-generators/src/Constrained/Spec/Map.hs
+++ b/libs/constrained-generators/src/Constrained/Spec/Map.hs
@@ -52,6 +52,7 @@ data MapSpec fn k v = MapSpec
   , mapSpecFold :: FoldSpec fn v
   }
 
+-- | emptySpec without all the constraints
 defaultMapSpec :: Ord k => MapSpec fn k v
 defaultMapSpec = MapSpec Nothing mempty mempty TrueSpec TrueSpec NoFold
 
@@ -80,6 +81,9 @@ instance
     (MapSpec mHint' mustKeys' mustVals' size' kvs' foldSpec') = fromGE ErrorSpec $ do
       typeSpec
         . MapSpec
+          -- This is min because that allows more compositionality - if a spec specifies a
+          -- low upper bound because some part of the spec will be slow it doesn't make sense
+          -- to increase it somewhere else because that part isn't slow.
           (unionWithMaybe min mHint mHint')
           (mustKeys <> mustKeys')
           (nub $ mustVals <> mustVals')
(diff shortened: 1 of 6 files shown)


libs/constrained-generators/src/Constrained/Spec/Map.hs
```diff
@@ -36,21 +36,25 @@ import Test.QuickCheck (shrinkList)
...
+  , mapSpecMustKeys :: Set k
   , mapSpecMustValues :: [v]
   , mapSpecSize :: Specification fn Integer
   , mapSpecElem :: Specification fn (k, v)
   , mapSpecFold :: FoldSpec fn v
   }
 
+defaultMapSpec :: Ord k => MapSpec fn k v
+defaultMapSpec = MapSpec Nothing mempty mempty TrueSpec TrueSpec NoFold
+
```
TimSheard, line 58: Might be worth a comment, that this is 'emptySpec', but without the HasSpec constraint.

libs/constrained-generators/src/Constrained/Spec/Map.hs
```diff
@@ -69,20 +73,21 @@ instance
...
+  emptySpec = defaultMapSpec
 
   combineSpec
-    (MapSpec mustKeys mustVals size kvs foldSpec)
-    (MapSpec mustKeys' mustVals' size' kvs' foldSpec') = fromGE ErrorSpec $ do
+    (MapSpec mHint mustKeys mustVals size kvs foldSpec)
+    (MapSpec mHint' mustKeys' mustVals' size' kvs' foldSpec') = fromGE ErrorSpec $ do
       typeSpec
         . MapSpec
+          (unionWithMaybe min mHint mHint')
```
TimSheard, line 87: Any particular reason to use 'min', lots of things might make sense (max, average)

libs/constrained-generators/src/Constrained/Spec/Map.hs
```diff
@@ -102,7 +107,7 @@ instance
           -- TODO, we should make sure size' is greater than or equal to 0
           satisfies
             (sz + Lit (sizeOf mustMap))
-            (size <> cardinality (mapSpec fstFn $ mapSpec toGenericFn kvs))
+            (maybe TrueSpec leqSpec mHint <> size <> cardinality (mapSpec fstFn $ mapSpec toGenericFn kvs))
```
TimSheard, line 114: This is where the rubber hits the road.  Maybe a comment about how the hint is used?
I guess this relates to the use of 'min' I talked about earlier.

TimSheard (approved):
Perfect..  Can't wait to try it out.

libs/constrained-generators/src/Constrained/Spec/Map.hs
MaximilianAlgehed: I added a comment:
```
          -- This is min because that allows more compositionality - if a spec specifies a
          -- low upper bound because some part of the spec will be slow it doesn't make sense
          -- to increase it somewhere else because that part isn't slow.
```

PR #4297, opened 2024-04-25
`constrained-generators`: Fix `ifElse` dependencies

MaximilianAlgehed:
# Description

This solves some thorny issues that @TimSheard was having with generating things that have a lot of `ifElse` in them. It generally doesn't make sense to resolve the if scrutinee after resolving the cases - unlike in `caseOn` where you're inspecting variables so you end up deferring things more.

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/libs/constrained-generators/src/Constrained/Base.hs
+++ b/libs/constrained-generators/src/Constrained/Base.hs
@@ -339,6 +339,7 @@ data Pred (fn :: [Type] -> Type -> Type) where
     -- encoded with `ProdOver` (c.f. `Constrained.Univ`).
     List (Binder fn) as ->
     Pred fn
+  -- if-then-else where the branches depend on the scrutinee.
   IfElse ::
     HasSpec fn Bool =>
     Term fn Bool ->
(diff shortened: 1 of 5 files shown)


libs/constrained-generators/src/Constrained/Base.hs
```diff
@@ -337,6 +339,12 @@ data Pred (fn :: [Type] -> Type -> Type) where
     -- encoded with `ProdOver` (c.f. `Constrained.Univ`).
     List (Binder fn) as ->
     Pred fn
+  IfElse ::
+    HasSpec fn Bool =>
+    Term fn Bool ->
+    Pred fn ->
+    Pred fn ->
+    Pred fn
```
TimSheard, line 348: Maybe a comment about how this is intended to be used?

TimSheard (approved):
looks good to me. Good to add a comment about intended uses of new features.

libs/constrained-generators/src/Constrained/Base.hs
```diff
@@ -337,6 +339,12 @@ data Pred (fn :: [Type] -> Type -> Type) where
     -- encoded with `ProdOver` (c.f. `Constrained.Univ`).
     List (Binder fn) as ->
     Pred fn
+  IfElse ::
```
MaximilianAlgehed, line 343:
```suggestion
  -- if-then-else where the branches depend on the scrutinee.
  IfElse ::
```

PR #4299, opened 2024-04-25
Fix strange CI failure.

### LINKED ISSUE #4274: Strange CI failure: Conway/Committee Ratification/Expired or resigned members are not counted/
https://github.com/IntersectMBO/cardano-ledger/actions/runs/8633563774/job/23666930118?pr=4252#step:15:5796

aniketd:
# Description

The range of generated expired epochs and unexpired epochs overlaped at point 100, which caused the `genExpiredOrResigned` to assume that all `Yes` voting committee members are to be expired, but according to the spec we consider members expired only after the current epoch is greater than their expiry epoch, whereas in this case both were 100.

The range of the generator has been adjusted according to these semantics.

Fixes #4274

# Checklist

- [x] Self-reviewed the diff


lehins (approved):
:+1: 
Thank you for figuring it out!

PR #4301, opened 2024-04-26
`constrained-generators`: Add monitoring capability to get a handle on test case distribution

### LINKED ISSUE #4217: Introduce distribution logging to `constrained-generators`
One thing that we're missing in `constrained-generators` is the ability to see what the distribution of generated data is (to spot things where we aren't testing some feature because of some mistake in the generator).

The natural way to implement this would be with something like a `monitor :: (Property -> Property) -> Pred fn` predicate, allowing us to do something like:
```
spec :: Spec fn (Either Int Bool)
spec = constrained $ \ e ->
  caseOn e
    (branch $ \ i -> [ monitor $ label "left", assert $ 0 <. i ])
    (branch $ \  b -> [ monitor $ label "right", assert b])
```
or create re-useable things like:
```
monitorPositive :: Term fn Int -> Pred fn 
monitorPositive i = ifElse (0 <. i) (monitor $ label "positive") (monitor $ label  "non-positive")
```
This could be combined with `reify` to classify any boolean function over stuff as well and would generally be useful to make sure distributions don't suck.

The tricky thing is making the design work out in practise, but I think it can be done by adding a `Property -> Property` field to the constructors of and collecting the coverage information during spec checking: `collectCoverageSpec :: Spec fn a -> a -> Property -> Property` (or with a `Writer` component on one of the established runner functions).

In all, a worth-while thing to do that shouldn't be too hard and will allows us some certainty that our tests are actually meaningful.

MaximilianAlgehed:
distribution

# Description

This PR introduces `monitor :: ((forall a. Term fn a -> a) -> Property -> Property) -> Pred fn` and `monitorSpec :: Specification fn a --> a -> Property -> Property` to `constrained-generators` to make it easier to get a grip on test case distribution in the generators.

closes #4217

# Checklist

- [x] Self-reviewed the diff


libs/constrained-generators/src/Constrained/Base.hs
```diff
@@ -279,6 +279,7 @@ data Binder fn a where
 deriving instance Show (Binder fn a)
 
 data Pred (fn :: [Type] -> Type -> Type) where
+  Monitor :: ((forall a. Term fn a -> a) -> Property -> Property) -> Pred fn
```
TimSheard, line 282: Maybe an example of how this is intended to be used?

TimSheard (approved):
So the most useful parts of the this PR are the tests, the demonstrate how the new feature is to be used. Might add one of these to the PR description.

libs/constrained-generators/src/Constrained/Base.hs
UlfNorell: I gave `monitor` and `monitorSpec` some explanatory haddocks.

Soupstraw (approved):

PR #4306, opened 2024-04-27
Expect minimum deposit for proposals to be higher than the minimum

micahkendall:
Expect minimum deposit for proposals to be higher than the minimum instead of equal. This allows a proposal to leave more funds than necessary. This could be used to draw greater attention to important proposals in the event of a DOS attack - frontends could use a rank by deposit size mechanism.

### DIFF
--- a/eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Gov.hs
+++ b/eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Gov.hs
@@ -355,7 +355,7 @@ govTransition = do
         -- Deposit check
         let expectedDep = pp ^. ppGovActionDepositL
          in pProcDeposit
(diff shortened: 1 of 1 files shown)


lehins:
It is an interesting solution. There is unfortunately a major issue with this proposal that we in Ledger can foresee.

If tooling were to rank proposals by the deposit, that would give significant advantage to proposers who have a lot of funds at their disposal. A whale could easily spam the network with many irrelevant proposals and then create one proposal of interest with a very high deposit, which would put that whale's proposal at the very top.

In any case, if I am wrong about this scenario, we are pretty late in the game to make serious changes like that. If you believe that this is a real issue, I would recommend [submitting a CIP](https://github.com/cardano-foundation/CIPs) and we would be able to tackle it for the next era.

PR #4308, opened 2024-04-29
Add a CI status check to prevent merging PRs that contain merges

### LINKED ISSUE #4307: Add a CI status check to prevent merging PRs that already have merges in them
GH has a button to bring a PR branch up to date with `master` and by default it merges `master` into the PR branch. This causes messy, hard-to-follow history in the repo, and it's not obvious from the UI what GH has done in response to clicking the button.

We will create an additional status check that will prevent merging, as a separate job within the `haskell.yml` workflow, similar to the code-formatting checks. It will check that the branch contains no merges (ie no commits with more than one parent) between `master` and the tip of the PR branch. It won't affect `master` itself because there won't be any commits to check.

If someone needs to bring in commits from another PR or feature branch, they will need to rebase their branch onto the other rather than merging the other into their branch.

The check logic will be:

```shell
! git log --merges --format=%H master..HEAD | grep -q .
```

neilmayhew:
Closes #4307

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/.github/workflows/haskell.yml
+++ b/.github/workflows/haskell.yml
@@ -332,6 +332,20 @@ jobs:
     - name: Regenerate hie.yaml and confirm that it is in sync
       run: ./scripts/gen-hie.sh
 
+  branch-history:
+    name: Check branch history
+    runs-on: ubuntu-latest
+    if: ${{ github.base_ref != '' && github.head_ref != '' }}
+    steps:
+    - uses: actions/checkout@v4
+    - name: Ensure the branch doesn't contain any merges
+      run: |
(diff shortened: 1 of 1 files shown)


lehins (approved):
Awesome! Thank you!

lehins (changes requested):
Looks like this git command does not work as expected: https://github.com/IntersectMBO/cardano-ledger/actions/runs/8886398544/job/24399775489?pr=4308

neilmayhew:
This seems to be working now. It failed (for the right reasons!) on the last run when there was still a merge commit on the branch. If it passes this time, it should be good to go.

lehins (approved):
Perfect!

Issue #4061, opened 2024-02-09
Ensure that proposal deposit does not reduce voting power


lehins:
There is a one line sentence in the CIP-1694 that:
> It [the deposit] will also be counted towards the stake of the reward address it will be paid back to, to not reduce the submitter's voting power to vote on their own (and competing) actions.

It is a bit ambiguous, but after some discussion the intent was that the deposits locked for proposal procedures do not affect the voting power of the decelerator to DReps and StakePools. It should not affect the reward calculation computation in any way.

This only has any sort of significance if gov action deposit will be set to something extremely high, eg. 1M ADA

Implementation of this feature is not a trivial one and could potentially delay Conway era delivery. Therefore it has been decided to postpone it's implementation until the Conway intra-era hardfork, if at all. Before we implement this feature we need to know that:
* there is a chance of really high deposits
* there is indeed interest from the community in implementing this feature

PR #4309, opened 2024-04-30
Add proposal deposits to DRep active voting stake.

aniketd:
# Description

Add proposal-deposits to the DRep distribution.

Partially addresses #4061. The same will be implemented for SPOs in a follow-up.

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/eras/conway/impl/src/Cardano/Ledger/Conway/Governance/DRepPulser.hs
+++ b/eras/conway/impl/src/Cardano/Ledger/Conway/Governance/DRepPulser.hs
@@ -1,4 +1,3 @@
-{-# LANGUAGE AllowAmbiguousTypes #-}
 {-# LANGUAGE BangPatterns #-}
 {-# LANGUAGE ConstraintKinds #-}
 {-# LANGUAGE DataKinds #-}
@@ -38,7 +37,6 @@ module Cardano.Ledger.Conway.Governance.DRepPulser (
   RunConwayRatify (..),
 ) where
 
-import Cardano.Ledger.Address
 import Cardano.Ledger.BaseTypes (EpochNo (..), Globals (..))
 import Cardano.Ledger.Binary (
   DecCBOR (..),
@@ -59,7 +57,7 @@ import Cardano.Ledger.CertState (CommitteeState)
 import Cardano.Ledger.Coin (Coin (..))
 import Cardano.Ledger.Conway.Era (ConwayRATIFY)
 import Cardano.Ledger.Conway.Governance.Internal
-import Cardano.Ledger.Conway.Governance.Procedures (GovActionState, gasDepositL, gasReturnAddrL)
+import Cardano.Ledger.Conway.Governance.Procedures (GovActionState)
 import Cardano.Ledger.Core
 import Cardano.Ledger.Credential (Credential (..))
 import Cardano.Ledger.DRep (DRep (..), DRepState (..))
@@ -72,12 +70,13 @@ import Control.Monad.Trans.Reader (Reader, runReader)
 import Control.State.Transition.Extended
 import Data.Aeson (KeyValue, ToJSON (..), object, pairs, (.=))
 import Data.Default.Class (Default (..))
-import Data.Foldable (foldl', toList)
+import Data.Foldable (toList)
 import Data.Functor.Identity (Identity)
 import Data.Kind (Type)
 import Data.Map.Strict (Map)
 import qualified Data.Map.Strict as Map
 import Data.Maybe (fromMaybe)
+import Data.Proxy (Proxy (Proxy))
 import Data.Pulse (Pulsable (..), pulse)
 import Data.Sequence.Strict (StrictSeq (..))
 import qualified Data.Sequence.Strict as SS
@@ -175,29 +174,16 @@ instance EraPParams era => FromCBOR (PulsingSnapshot era) where
 --   (c) is the size of the DRepDistr, this grows as the accumulator
 computeDRepDistr ::
   forall era.
+  Proxy era ->
   Map (Credential 'Staking (EraCrypto era)) (CompactForm Coin) ->
   Map (Credential 'DRepRole (EraCrypto era)) (DRepState (EraCrypto era)) ->
-  StrictSeq (GovActionState era) ->
+  Map (Credential 'Staking (EraCrypto era)) (CompactForm Coin) ->
   Map (DRep (EraCrypto era)) (CompactForm Coin) ->
   Map (Credential 'Staking (EraCrypto era)) (UMElem (EraCrypto era)) ->
   Map (DRep (EraCrypto era)) (CompactForm Coin)
-computeDRepDistr stakeDistr regDReps gass dRepDistr uMapChunk =
+computeDRepDistr _ stakeDistr regDReps proposalDeposits dRepDistr uMapChunk =
   Map.foldlWithKey' go dRepDistr uMapChunk
   where
-    -- QUESTION: Can I memoize this as a field in DRepPulsar? That causes a
-    -- type-error, though that does _not_ get resolved even with AllowAmbiguousTypes
-    -- as `EraCrypto era` is a non-injective type-family
-    proposalDeposits =
-      foldl'
-        ( \gasMap gas ->
-            Map.insertWith
-              addCompact
-              (gas ^. gasReturnAddrL . rewardAccountCredentialL)
-              (fromMaybe (CompactCoin 0) $ toCompact $ gas ^. gasDepositL)
-              gasMap
-        )
-        mempty
-        gass
     go accum stakeCred umElem =
       let stake = fromMaybe (CompactCoin 0) $ Map.lookup stakeCred stakeDistr
           proposalDeposit = fromMaybe (CompactCoin 0) $ Map.lookup stakeCred proposalDeposits
@@ -246,6 +232,8 @@ data DRepPulser era (m :: Type -> Type) ans where
     -- ^ Snapshot of the EnactState, Used to build the Env of the RATIFY rule
     , dpProposals :: !(StrictSeq (GovActionState era))
     -- ^ Snapshot of the proposals. This is the Signal for the RATIFY rule
+    , dpProposalDeposits :: !(Map (Credential 'Staking (EraCrypto era)) (CompactForm Coin))
+    -- ^ Snapshot of the proposal-deposits per reward-account-staking-credential
     , dpGlobals :: !Globals
     } ->
     DRepPulser era m ans
@@ -262,7 +250,7 @@ instance Pulsable (DRepPulser era) where
     | done pulser = pure pulser {dpIndex = 0}
     | otherwise =
         let !chunk = Map.take dpPulseSize $ Map.drop dpIndex $ UMap.umElems dpUMap
-            dRepDistr = computeDRepDistr dpStakeDistr dpDRepState dpProposals dpDRepDistr chunk
+            dRepDistr = computeDRepDistr (Proxy @era) dpStakeDistr dpDRepState dpProposalDeposits dpDRepDistr chunk
          in pure (pulser {dpIndex = dpIndex + dpPulseSize, dpDRepDistr = dRepDistr})
 
   completeM x@(DRepPulser {}) = pure (snd $ finishDRepPulser @era (DRPulsing x))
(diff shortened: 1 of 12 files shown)


eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/RatifySpec.hs
```diff
@@ -624,6 +627,117 @@ votingSpec =
...
+          it "Directly" $ do
+            -- Only modify the applicable thresholds
+            modifyPParams $ \pp ->
+              pp
+                & ppDRepVotingThresholdsL
+                  .~ def
+                    { dvtCommitteeNormal = 51 %! 100
+                    , dvtCommitteeNoConfidence = 51 %! 100
+                    }
+                & ppCommitteeMaxTermLengthL .~ EpochInterval 20
```
teodanciu: There is no need to modify `ppCommitteeMaxTermLengthL`, because it is set inthis way when initializing Conway.
I assume the `51 %! 100` threshold is relevant for the test, right?  But is the `dvtCommitteeNoConfidence` one relevant as well?

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/RatifySpec.hs
```diff
@@ -624,6 +627,117 @@ votingSpec =
...
+            -- Submit the vote from DRep #1
+            submitVote_ VoteYes (DRepVoter $ KeyHashObj drepKH1) addCCGaid
+            passNEpochs 2
+            -- The vote should not result in a ratification
+            isDRepAccepted addCCGaid `shouldReturn` False
+            getLastEnactedCommittee `shouldReturn` SNothing
+            -- Submit another proposal to bump up the active voting stake
+            anotherCC <- KeyHashObj <$> freshKeyHash
+            let anotherAddCCAction = UpdateCommittee SNothing mempty (Map.singleton anotherCC 10) (75 %! 100)
+            _anotherAddCCGaid <-
```
teodanciu: Why not submitProposal_ instead of binding to this underscore name?

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/RatifySpec.hs
```diff
@@ -624,6 +627,117 @@ votingSpec =
...
+                  , pProcReturnAddr = dRepRewardAccount
+                  , pProcGovAction = addCCAction
+                  , pProcAnchor = def
+                  }
+            -- Submit the vote from DRep #1
+            submitVote_ VoteYes (DRepVoter $ KeyHashObj drepKH1) addCCGaid
+            passNEpochs 2
+            -- The vote should not result in a ratification
+            isDRepAccepted addCCGaid `shouldReturn` False
+            getLastEnactedCommittee `shouldReturn` SNothing
```
teodanciu, line 661: This is just a question for my education, that I was meaning to ask, because I came across it when doing some refactorings on the previous RP. 
I don't understand the meaning of these `getLastEnacted`* functions .
I can see they are looking in :
`nesEsL . esLStateL . lsUTxOStateL . utxosGovStateL . proposalsGovStateL . pRootsL`, and then `grCommitteeL . prRootL` for committee or `grConstitutionL . prRootL` for constitution, etc. 
So this would be the thing that *will* be enacted after one epoch?  What is the difference between this and looking in 
` nesEsL . esLStateL . lsUTxOStateL . utxosGovStateL . committeeGovStateL`  ?  Is the former a sort of preview of the latter?

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/RatifySpec.hs
```diff
@@ -624,6 +627,117 @@ votingSpec =
...
+          it "After switching delegations" $ do
+            -- Only modify the applicable thresholds
+            modifyPParams $ \pp ->
+              pp
+                & ppDRepVotingThresholdsL
+                  .~ def
+                    { dvtCommitteeNormal = 51 %! 100
+                    , dvtCommitteeNoConfidence = 51 %! 100
+                    }
+                & ppCommitteeMaxTermLengthL .~ EpochInterval 20
```
teodanciu: Same comment about unnecessary setting of `ppCommitteeMaxTermLengthL` (and probably also of `dvtCommitteeNoConfidence`)

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/RatifySpec.hs
```diff
@@ -624,6 +627,117 @@ votingSpec =
...
+              submitProposal $
+                ProposalProcedure
+                  { pProcDeposit = Coin 1_000_000
+                  , pProcReturnAddr = dRepRewardAccount1
+                  , pProcGovAction = addCCAction
+                  , pProcAnchor = def
+                  }
+            anotherCC <- KeyHashObj <$> freshKeyHash
+            let anotherAddCCAction = UpdateCommittee SNothing mempty (Map.singleton anotherCC 10) (75 %! 100)
+            _anotherAddCCGaid <-
```
teodanciu: Same as above, you could have less noise with submitProposal_

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/RatifySpec.hs
```diff
@@ -624,6 +627,117 @@ votingSpec =
...
+              submitProposal $
+                ProposalProcedure
+                  { pProcDeposit = Coin 600_000
+                  , pProcReturnAddr = dRepRewardAccount
+                  , pProcGovAction = anotherAddCCAction
+                  , pProcAnchor = def
+                  }
+            passNEpochs 2
+            -- The same vote should now successfully ratify the proposal
```
teodanciu, line 673: So is it the case that more money from the root utxo is now associated with this drep (via the deposit for this proposal), hence the vote of the drep has become more significant because of more stake?

eras/conway/impl/src/Cardano/Ledger/Conway/Governance/DRepPulser.hs
```diff
@@ -173,25 +174,42 @@ instance EraPParams era => FromCBOR (PulsingSnapshot era) where
...
+  StrictSeq (GovActionState era) ->
+  Map (DRep (EraCrypto era)) (CompactForm Coin) ->
+  Map (Credential 'Staking (EraCrypto era)) (UMElem (EraCrypto era)) ->
+  Map (DRep (EraCrypto era)) (CompactForm Coin)
+computeDRepDistr stakeDistr regDReps gass dRepDistr uMapChunk =
   Map.foldlWithKey' go dRepDistr uMapChunk
   where
+    -- QUESTION: Can I memoize this as a field in DRepPulsar? That causes a
+    -- type-error, though that does _not_ get resolved even with AllowAmbiguousTypes
+    -- as `EraCrypto era` is a non-injective type-family
```
teodanciu: So you mean, in DRepPulser, instead of having
```dpProposals :: !(StrictSeq (GovActionState era))`, to have a map from credential to deposit?

File: eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/RatifySpec.hs (hunk 3 above)
Comment (aniketd) [reply eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/RatifySpec.hs:661]: Great questions!

> So this would be the thing that will be enacted after one epoch?

No. These are the last enacted governance action ids, as the name suggests. Which means, they are each returning the last governance action id that was enacted in the last epoch boundary.

>  looking in `nesEsL . esLStateL . lsUTxOStateL . utxosGovStateL . committeeGovStateL`?

This will give you the current state of the committee, as a whole, but not tell you anything about what was the last governance action that was enacted at the previous epoch boundary.

File: eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/RatifySpec.hs (hunk 6 above)
Comment (aniketd) [reply eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/RatifySpec.hs:673]: Precisely, yes.

File: eras/conway/impl/src/Cardano/Ledger/Conway/Governance/DRepPulser.hs (hunk 7 above)
Comment (aniketd) [reply eras/conway/impl/src/Cardano/Ledger/Conway/Governance/DRepPulser.hs]: Not "instead", but "in addition to", because we need `dpProposals` in an ordered container, as they are the signal to the `RATIFY` rule.

File: eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/RatifySpec.hs (hunk 3 above)
Comment (teodanciu) [reply eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/RatifySpec.hs:661]: Oh I see, so if nothing else was enacted in the meantime, I guess those two would coincide, but it's more precise to check with `getLastEnacted`, to verify that the action had indeed the desired effect (rather than it being set in the gov state for other reason). 
Thanks for the explanation!

Comment (aniketd) [reply eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/RatifySpec.hs:661]: > I guess those two would coincide

The functions `getLastEnacted...` return a `GovActionId`, whereas the `Committee` is just the committee. That is also the difference. So they are not the same thing.

File: eras/conway/impl/src/Cardano/Ledger/Conway/Governance/DRepPulser.hs (hunk 7 above)
Comment (lehins) [reply eras/conway/impl/src/Cardano/Ledger/Conway/Governance/DRepPulser.hs]: The answer is it is an absolutele must to compute this during the DRep pulser setup in: `setFreshDRepPulsingState`. Feel free to ping me with a type error if you can't figure it out. 

You might need a `forall` in the `setFreshDRepPulsingState` and a type annotation on the field, but `AllowAmbiguousTypes` is definitely not the right solution here

File: eras/conway/impl/src/Cardano/Ledger/Conway/Governance/DRepPulser.hs [hunk 8]
```diff
@@ -298,20 +316,31 @@ class
...
   where
-    !finalDRepDistr = computeDRepDistr dpStakeDistr dpDRepState dpDRepDistr $ Map.drop dpIndex $ umElems dpUMap
+    !finalDRepDistr =
+      computeDRepDistr
+        dpStakeDistr
+        dpDRepState
+        dpProposals
+        dpDRepDistr
+        $ Map.drop dpIndex
+        $ umElems dpUMap
```
lehins: A local binding will make it possible to avoid this column style
```suggestion
    !leftOver = Map.drop dpIndex $ umElems dpUMap
    !finalDRepDistr =
      computeDRepDistr dpStakeDistr dpDRepState dpProposals dpDRepDistr leftOver
```

eras/conway/impl/src/Cardano/Ledger/Conway/Governance/DRepPulser.hs
```diff
@@ -298,20 +316,31 @@ class
   runConwayRatify globals ratifyEnv ratifyState (RatifySignal ratifySig) =
     let ratifyResult =
           runReader
-            ( applySTS @(ConwayRATIFY era) (TRC (ratifyEnv, ratifyState, RatifySignal $ reorderActions ratifySig))
+            ( applySTS @(ConwayRATIFY era) $
+                TRC (ratifyEnv, ratifyState, RatifySignal $ reorderActions ratifySig)
             )
             globals
      in case ratifyResult of
           Left ps ->
```
Soupstraw:
```suggestion
          Left (x :| _) -> absurd x
```
then you can also get rid of the `error`

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/RatifySpec.hs
aniketd: Thanks for pointing this out :+1:

eras/conway/impl/src/Cardano/Ledger/Conway/Governance/DRepPulser.hs
aniketd: This is now implemented, but I had to use `Proxy` :+1:

eras/conway/impl/src/Cardano/Ledger/Conway/Governance/DRepPulser.hs
aniketd: I did not fully understand the intention and motivation @Soupstraw, sorry :disappointed: . Could you explain a bit?

lehins: `absurd` ensures at compile time that it is an impossible case, so you don't need to provide any error messages. This is possible because:
```haskell
type PredicateFailure (ConwayRATIFY era) = Void
```

lehins (approved):
Looks great. Thank you!

PR #4310, opened 2024-04-30
Fix `OMap.assocList`

Soupstraw:
# Description

This PR changes `assocList` so that the order of key-value pairs is preserved. Also added the `elems` function, which returns a list of elements of an `OMap`.

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/libs/cardano-data/src/Data/OMap/Strict.hs
+++ b/libs/cardano-data/src/Data/OMap/Strict.hs
@@ -27,6 +27,7 @@ module Data.OMap.Strict (
   fromFoldableDuplicates,
   toMap,
   assocList,
+  elems,
   toStrictSeq,
   toStrictSeqOKeys,
   toStrictSeqOfPairs,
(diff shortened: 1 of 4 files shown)


libs/cardano-data/src/Data/OMap/Strict.hs
```diff
@@ -371,13 +372,16 @@ omapl ><| omapr = case omapl of
...
   fromList = fromFoldable
   toList = F.toList
 
-assocList :: OMap k v -> [(k, v)]
-assocList = Map.toList . toMap
+assocList :: Ord k => OMap k v -> [(k, v)]
+assocList = F.toList . toStrictSeqOfPairs
+
+elems :: Ord k => OMap k v -> [v]
+elems = fmap snd . assocList
```
aniketd:
```suggestion
elems = F.toList . toStrictSeq
```

Perhaps this would allocate less? But it is a trivial suggestion, feel free to ignore.

aniketd (approved):
Just a small suggestion, but otherwise looks great! :+1:

PR #4312, opened 2024-05-01
Convert AlonzoValidTxUTXOW to ImpTest (Part 1)

### LINKED ISSUE #4180: Translate all tests from `AlonzoValidTxUTXOW` to Imp test
All of the tests in the [`Test.Cardano.Ledger.Examples.AlonzoValidTxUTXOW`](https://github.com/IntersectMBO/cardano-ledger/blob/master/libs/cardano-ledger-test/src/Test/Cardano/Ledger/Examples/AlonzoValidTxUTXOW.hs) need to be translated to Imp Test in `cardano-ledger-alonzo:test:tests` package into the perspective `[Rule]Spec` modules that are relevant for the test. Note that not all tests are related to UTXOW, despite what the name of the module says.

If for some reason some tests require serious changes to Imp test, they can be left at the original place for the time being, until Imp test framework gets that feature implemented as part of another task.

neilmayhew:
# Description

The tests in [Test.Cardano.Ledger.Examples.AlonzoValidTxUTXOW](https://github.com/IntersectMBO/cardano-ledger/blob/master/libs/cardano-ledger-test/src/Test/Cardano/Ledger/Examples/AlonzoValidTxUTXOW.hs) need to be translated to `ImpTest` and moved from `cardano-ledger-test` to `cardano-ledger-alonzo:testlib`.

Takes care of some part of #4180

# Checklist

- [ ] Self-reviewed the diff


eras/alonzo/impl/testlib/Test/Cardano/Ledger/Alonzo/Imp/UtxowSpec/Valid.hs
```diff
@@ -0,0 +1,323 @@
...
+  , AlonzoEraImp era
+  ) =>
+  SpecWith (ImpTestState era)
+spec = do
+  let
+    tx `shouldProduceState` expectedM = do
+      expected <- expectedM
+      submitTxAnn_ "Submit a transaction with an expected ledger state" tx
+      actual <- getsNES $ nesEsL . esLStateL . lsUTxOStateL
+      actual `shouldBe` expected
```
lehins: This is one of the worst parts of the previous tests. They compare the full state, rather than verify that expected parts of the state have been correctly changed.

We can talk more about it, but in a nutshell tests should check only the relevant part, eg: if we want to test that a transaction spends an output, we apply a transaction and check that the said output was removed from the state.

neilmayhew: I had already planned to approach this differently, as we discussed. For now, I've updated this module to contain even more minimal placeholders.

eras/alonzo/impl/testlib/Test/Cardano/Ledger/Alonzo/Imp/UtxowSpec/Invalid.hs
```diff
@@ -0,0 +1,231 @@
...
+      submitFailingTx tx [injectFailure $ MissingVKeyWitnessesUTXOW [sender]]
+
+  -- This already exists in Shelley.Imp.UtxowSpec
+  it "Missing 1-phase script witness" $ do
+    signer <- impLogToExpr freshKeyHash
+    let scriptHash = hashScript @era $ fromNativeScript $ RequireSignature signer
+    txIn <- produceScript scriptHash
+    submitFailingTx
+      (mkBasicTx $ mkBasicTxBody & inputsTxBodyL .~ [txIn])
+      [injectFailure $ MissingScriptWitnessesUTXOW [scriptHash]]
```
neilmayhew: I still need to check whether we need Alonzo-specific tests for these predicate failures

eras/alonzo/impl/testlib/Test/Cardano/Ledger/Alonzo/Imp/UtxowSpec/Valid.hs
```diff
@@ -0,0 +1,36 @@
...
+    it "Not validating MINT script" $ do
+      const $ pendingWith "not implemented yet"
+    it "Validating scripts everywhere" $ do
+      const $ pendingWith "not implemented yet"
+    it "Acceptable supplimentary datum" $ do
+      const $ pendingWith "not implemented yet"
+    it "Multiple identical certificates" $ do
+      const $ pendingWith "not implemented yet"
+    it "Non-script output with datum" $ do
+      const $ pendingWith "not implemented yet"
```
neilmayhew, line 36: I haven't checked yet whether any of these are redundant

eras/alonzo/impl/testlib/Test/Cardano/Ledger/Alonzo/Imp/UtxowSpec/Invalid.hs
neilmayhew: There's no code in Alonzo, Mary or Allegra that produces these failures, so I'll assume the test in Shelley is sufficient.

eras/shelley/impl/testlib/Test/Cardano/Ledger/Shelley/ImpTest.hs
```diff
@@ -923,7 +926,7 @@ shelleyFixupTx =
     >=> addRootTxIn
     >=> fixupFees
     >=> updateAddrTxWits
-    >=> (\tx -> logFeeMismatch tx $> tx)
+    >=> (logFeeMismatch *> pure)
```
lehins: This change is wrong, so let's bring back the original:
```suggestion
    >=> (\tx -> logFeeMismatch tx $> tx)
```

This is incorrect, because it will totally skip execution of `logFeeMismatch`, because composition of `logFeeMismatch` and `pure` with `*>` actually uses `Applicative` instance for `->`, not the underlying monad. It took me a moment to realize why it even compiles.

Here is an example of how it breaks:

```haskell
ghci> x <- (print *> pure) "foo"
ghci> x
"foo"
```

So, the lesson I learned today is that `*>` for functions simply ignores the first function:
```
1. f *> g
2. (id <$ f) <*> g
3. fmap (const id f) <*> g
4. fmap id <*> g
```

In the matter of fact, things like that is usually the reason why I try to stay away from esoteric instances like Functor/Applicative/Monad/.. for functions, Foldable for tuples, etc.

eras/shelley/impl/testlib/Test/Cardano/Ledger/Shelley/ImpTest.hs
```diff
@@ -1269,11 +1284,15 @@ freshKeyPair = do
...
   pure (keyHash, keyPair)
 
+-- | Generate a random KeyPair and add it to the known keys in the Imp state
 freshKeyAddr :: Era era => ImpTestM era (KeyHash r (EraCrypto era), Addr (EraCrypto era))
 freshKeyAddr = do
   keyHash <- freshKeyHash
   pure (coerceKeyRole keyHash, Addr Testnet (KeyHashObj keyHash) StakeRefNull)
 
+freshAddr :: Era era => ImpTestM era (Addr (EraCrypto era))
+freshAddr = snd <$> freshKeyAddr
```
lehins: `freshKeyAddr` is called that way not because it returns both `KeyHash` and the `Addr`, but because it creates and `Addr` that contains a `KeyHash`, instead of a `ScriptHash`.

So this extra function is certainly helpful, it's name is far from ideal. Let's just give it a suffix of some sort instead:

```suggestion
freshKeyAddr_ :: Era era => ImpTestM era (Addr (EraCrypto era))
freshKeyAddr_ = snd <$> freshKeyAddr
```

There is a totally legitimate implementation of `freshScriptAddr` that we should implement at some point that can produce a random address for us that picks out a known spending script at random

eras/alonzo/impl/testlib/Test/Cardano/Ledger/Alonzo/Imp/UtxowSpec/Valid.hs
lehins: That's totally fine. We can remove any redundant ones later

lehins (changes requested):
Looks good! There is only one actual problem that I found with the PR. 

I like the improvement on the `logToExpr`. It didn't even occur to me that the line was reported incorrectly and addition of blue color is also a nice touch :slightly_smiling_face:

eras/shelley/impl/testlib/Test/Cardano/Ledger/Shelley/ImpTest.hs
neilmayhew: Yikes! That's bad. Thanks for catching this, and for explaining. I'll be more cautious about using the `(->)` instances in future.

It makes sense that `*>` would skip the left argument, because in a pure context there's no point in executing something if you're not going to use the result.

eras/shelley/impl/testlib/Test/Cardano/Ledger/Shelley/ImpTest.hs
neilmayhew: There are now no direct calls to `freshKeyAddr`, so I'm wondering about just changing `freshKeyAddr` to return only the `Addr`. If a caller really needs the `keyHash` they can recover it from the `Addr`. However, that requires some theoretically-failing pattern matching, so if it's likely that this could be needed in future we could keep the original function but rename it to `freshKeyAddrWithHash` or `freshKeyAddr'`. I think it's better that the more frequently used function has the shorter and more memorable name.

neilmayhew: I've implemented my suggestion. Let me know if you prefer what you proposed.

lehins: That's fine. We can add one if we need it later

eras/shelley/impl/CHANGELOG.md
```diff
@@ -18,8 +18,13 @@
   * `UTxOState`
   * `UtxoEnv`
 
-### testlib
+### `testlib`
 
+* Add `impLogToExpr`
```
lehins: This now needs to go into a new section `1.12.1.0`, since `1.12.0.0` has already been released. (See #4429)

eras/shelley/impl/testlib/Test/Cardano/Ledger/Shelley/ImpTest.hs
lehins: Actually, spoke too soon. This becomes a breaking change and we need to bump a major version. Also I really don't like the new name: `freshKeyAddrWithHash`. Let's go with my suggestion instead

eras/shelley/impl/CHANGELOG.md
lehins: Make sure to also add a lower bound on the packages that use the new `getKeyAddr_`

lehins (approved):
Perfect! Thank you!

PR #4314, opened 2024-05-01
Ts newstyle specs

TimSheard:
Use the new style constrained generators to build Specifications of well formed type used in the Cardano Ledger.
Well formed means that constraints that should hold in the NewEpochState on the chain will hold on the randomly generated
values. The (WellFormed t) class has a has a member function with type (  wff :: Gen t )

These Specifications will probably have to be refined over time. But they are complete in that the machinery exists for every
type in the Ledger. Currently they mimic the constraints in the oldstyle constrained generators. They are significantly simpler and much more composeable.

The PR also has a few fixes that were necessary to handle the constraints we wrote. This PR exercised the constrained generators library quite a bit, and it needed some additions. These include

- Filled in missing cases in computation of cardinality, methods cardinalTypeSpec and cardinalTrueSpec
- Better use of constraints like  (x ==. mempty) to get accurate size information
- Added typeSpecOpt to a few HasSpec instances
- Added the function rngSet_ :: Term fn (Map a b) -> Term (Set b)

 

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/Conway/Gov.hs
+++ b/libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/Conway/Gov.hs
@@ -1,13 +1,23 @@
+{-# LANGUAGE BangPatterns #-}
 {-# LANGUAGE DataKinds #-}
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE ImportQualifiedPost #-}
 {-# LANGUAGE NamedFieldPuns #-}
 {-# LANGUAGE RecordWildCards #-}
+{-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE TypeApplications #-}
+-- {-# LANGUAGE TemplateHaskell #-}
+-- {-# LANGUAGE ViewPatterns #-}
+-- {-# LANGUAGE QuasiQuotes #-}
+{-# OPTIONS_GHC -Wno-unused-binds #-}
+{-# OPTIONS_GHC -Wno-unused-imports #-}
+{-# OPTIONS_GHC -Wno-unused-matches #-}
 
 -- | Specs necessary to generate, environment, state, and signal
 -- for the GOV rule
 module Test.Cardano.Ledger.Constrained.Conway.Gov where
+
+import Test.QuickCheck (generate)
 
 import Cardano.Ledger.Shelley.HardForks qualified as HardForks
 import Data.Foldable
@@ -24,12 +34,15 @@ import Lens.Micro
 
 import Constrained
 
+import Cardano.Ledger.Coin (Coin (..))
 import Cardano.Ledger.Conway (ConwayEra)
 import Cardano.Ledger.Conway.Core
 import Cardano.Ledger.Crypto (StandardCrypto)
+import Constrained.Base (printPlan)
 import Lens.Micro qualified as L
 import Test.Cardano.Ledger.Constrained.Conway.Instances
 import Test.Cardano.Ledger.Constrained.Conway.PParams
+import Test.Cardano.Ledger.Generic.PrettyCore (PrettyA (..))
 
 govEnvSpec ::
   IsConwayUniv fn =>
@@ -239,6 +252,16 @@ onHardFork ::
   Pred fn
 onHardFork gas k = onCon @"HardForkInitiation" (pProcGovAction_ . gasProposalProcedure_ $ gas) k
 
+{-
+main :: IO ()
+main = do
+  env <- generate $ genFromSpec $ govEnvSpec @ConwayFn
+  propose <- generate $ genFromSpec $ govProposalsSpec @ConwayFn env
+  procedures <- generate $ genFromSpec $ govProceduresSpec @ConwayFn env propose
+  -- printPlan $ govProceduresSpec @ConwayFn env propose
+  putStrLn (show (prettyA procedures))
+-}
+
 govProceduresSpec ::
   IsConwayUniv fn =>
   GovEnv (ConwayEra StandardCrypto) ->
@@ -264,7 +287,8 @@ govProceduresSpec ge@GovEnv {..} ps =
                 match kvp $ \voter mapActVote ->
                   (caseOn voter)
                     ( branch $ \_c ->
-                        subset_ (dom_ mapActVote) (lit $ Set.fromList committeeVotableActionIds)
+                        [ assert $ subset_ (dom_ mapActVote) (lit $ Set.fromList committeeVotableActionIds)
+                        ]
                     )
                     ( branch $ \_c ->
                         subset_ (dom_ mapActVote) (lit $ Set.fromList drepVotableActionIds)
@@ -401,52 +425,64 @@ wfPParamsUpdate ::
   IsConwayUniv fn =>
   Term fn (PParamsUpdate (ConwayEra StandardCrypto)) ->
   Pred fn
-wfPParamsUpdate ppu =
+wfPParamsUpdate pparamsUpdate =
   toPred
-    [ assert $ ppu /=. lit emptyPParamsUpdate
-    , match ppu $
-        \_cppMinFeeA
-         _cppMinFeeB
-         cppMaxBBSize
-         cppMaxTxSize
-         cppMaxBHSize
-         _cppKeyDeposit
-         cppPoolDeposit
-         _cppEMax
-         _cppNOpt
-         _cppA0
-         _cppRho
-         _cppTau
-         _cppProtocolVersion
-         _cppMinPoolCost
-         _cppCoinsPerUTxOByte
-         cppCostModels
-         _cppPrices
-         _cppMaxTxExUnits
-         _cppMaxBlockExUnits
-         cppMaxValSize
-         cppCollateralPercentage
-         _cppMaxCollateralInputs
-         _cppPoolVotingThresholds
-         _cppDRepVotingThresholds
-         _cppCommitteeMinSize
-         cppCommitteeMaxTermLength
-         cppGovActionLifetime
-         cppGovActionDeposit
-         cppDRepDeposit
-         _cppDRepActivity
-         _cppMinFeeRefScriptCoinsPerByte ->
-            [ cppMaxBBSize /=. lit (THKD $ SJust 0)
-            , cppMaxTxSize /=. lit (THKD $ SJust 0)
-            , cppMaxBHSize /=. lit (THKD $ SJust 0)
-            , cppMaxValSize /=. lit (THKD $ SJust 0)
-            , cppCollateralPercentage /=. lit (THKD $ SJust 0)
-            , cppCommitteeMaxTermLength /=. lit (THKD $ SJust $ EpochInterval 0)
-            , cppGovActionLifetime /=. lit (THKD $ SJust $ EpochInterval 0)
-            , cppPoolDeposit /=. lit (THKD $ SJust mempty)
-            , cppGovActionDeposit /=. lit (THKD $ SJust mempty)
-            , cppDRepDeposit /=. lit (THKD $ SJust mempty)
-            , cppCostModels ==. lit (THKD SNothing) -- NOTE: this is because the cost
-            -- model generator is way too slow
-            ]
+    [ assert $ pparamsUpdate /=. lit emptyPParamsUpdate
+    , match
+        pparamsUpdate
+        ( \ppupdate ->
+            match
+              ppupdate
+              ( \_feeA
+                 _feeB
+                 maxBBSize
+                 maxTxSize
+                 maxBHSize
+                 _keyDeposit
+                 poolDeposit
+                 _emax
+                 _nOpt
+                 _nAo
+                 _nrho
+                 _ntau
+                 _ndecentral
+                 _protocol
+                 _minUtxoVal
+                 _minPoolCost
+                 -- Alonzo
+                 _coinsPerUTxo
+                 costModels
+                 _prices
+                 _maxTx
+                 _maxBlock
+                 maxValSize
+                 collateralPercentage
+                 _maxColInputs
+                 -- Babbage
+                 _coinsperbyte
+                 -- Conway
+                 _poolthresh
+                 _drepthresh
+                 _commMinSize
+                 committeeMaxTermLength
+                 govActionLifetime
+                 govActionDeposit
+                 drepDeposit
+                 _drepAct
+                 _refscriptPerByte ->
+                    [ maxBBSize /=. lit (SJust 0)
+                    , maxTxSize /=. lit (SJust 0)
+                    , maxBHSize /=. lit (SJust 0)
+                    , maxValSize /=. lit (SJust 0)
+                    , collateralPercentage /=. lit (SJust 0)
+                    , committeeMaxTermLength /=. lit (SJust $ EpochInterval 0)
+                    , govActionLifetime /=. lit (SJust $ EpochInterval 0)
+                    , poolDeposit /=. lit (SJust (Coin 0))
+                    , govActionDeposit /=. lit (SJust (Coin 0))
+                    , drepDeposit /=. lit (SJust (Coin 0))
+                    , costModels ==. lit SNothing -- NOTE: this is because the cost
+                    -- model generator is way too slow
+                    ]
+              )
+        )
     ]
--- a/libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/Conway/Instances.hs
+++ b/libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/Conway/Instances.hs
@@ -8,6 +8,7 @@
 {-# LANGUAGE FlexibleInstances #-}
 {-# LANGUAGE GADTs #-}
 {-# LANGUAGE ImportQualifiedPost #-}
+{-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE MultiParamTypeClasses #-}
 {-# LANGUAGE RecordWildCards #-}
 {-# LANGUAGE ScopedTypeVariables #-}
@@ -16,6 +17,7 @@
 {-# LANGUAGE TypeFamilies #-}
 {-# LANGUAGE TypeOperators #-}
 {-# LANGUAGE UndecidableInstances #-}
+{-# LANGUAGE UndecidableSuperClasses #-}
 {-# OPTIONS_GHC -Wno-orphans #-}
 
 -- RecordWildCards cause name shadowing warnings in ghc-8.10.
@@ -29,6 +31,12 @@
 module Test.Cardano.Ledger.Constrained.Conway.Instances (
   ConwayFn,
   StringFn,
+  CoinFn (..),
+  toDelta_,
+  toDeltaFn,
+  EraPP (..),
+  PParamSubset (..),
+  PPUpdate (..),
   ProposalTree,
   onJust',
   onSized,
@@ -52,6 +60,8 @@ module Test.Cardano.Ledger.Constrained.Conway.Instances (
   ProposalsSplit (..),
   genProposalsSplit,
   proposalSplitSum,
+  protocolVersion_,
+  maxTxSize_,
 ) where
 
 import Cardano.Chain.Common (
@@ -66,20 +76,23 @@ import Cardano.Chain.Common (
 import Cardano.Crypto.Hash hiding (Blake2b_224)
 import Cardano.Crypto.Hashing (AbstractHash, abstractHashFromBytes)
 import Cardano.Ledger.Address
+import Cardano.Ledger.Allegra (Allegra)
 import Cardano.Ledger.Allegra.Scripts
+import Cardano.Ledger.Alonzo (Alonzo)
 import Cardano.Ledger.Alonzo.PParams
 import Cardano.Ledger.Alonzo.Scripts (AlonzoScript (..))
 import Cardano.Ledger.Alonzo.Tx
 import Cardano.Ledger.Alonzo.TxAuxData (AlonzoTxAuxData (..), AuxiliaryDataHash)
 import Cardano.Ledger.Alonzo.TxOut
 import Cardano.Ledger.Alonzo.TxWits
+import Cardano.Ledger.Babbage (Babbage)
 import Cardano.Ledger.Babbage.TxBody (BabbageTxOut (..))
 import Cardano.Ledger.BaseTypes hiding (inject)
 import Cardano.Ledger.Binary (Sized (..))
 import Cardano.Ledger.CertState
 import Cardano.Ledger.Coin
 import Cardano.Ledger.Compactible
-import Cardano.Ledger.Conway (ConwayEra)
+import Cardano.Ledger.Conway (Conway, ConwayEra)
 import Cardano.Ledger.Conway.Core
 import Cardano.Ledger.Conway.Governance
 import Cardano.Ledger.Conway.PParams
@@ -99,6 +112,7 @@ import Cardano.Ledger.Keys (
   KeyRole (..),
   WitVKey,
  )
+import Cardano.Ledger.Mary (Mary)
 import Cardano.Ledger.Mary.Value (AssetName (..), MaryValue (..), MultiAsset (..), PolicyID (..))
 import Cardano.Ledger.MemoBytes
 import Cardano.Ledger.Plutus.CostModels
@@ -108,18 +122,24 @@ import Cardano.Ledger.Plutus.Language
 import Cardano.Ledger.PoolDistr
 import Cardano.Ledger.PoolParams
 import Cardano.Ledger.SafeHash
+import Cardano.Ledger.Shelley (Shelley)
 import Cardano.Ledger.Shelley.LedgerState hiding (ptrMap)
+import Cardano.Ledger.Shelley.PParams
 import Cardano.Ledger.Shelley.PoolRank
 import Cardano.Ledger.Shelley.Rules
 import Cardano.Ledger.Shelley.TxAuxData (Metadatum)
+import Cardano.Ledger.Shelley.TxOut (ShelleyTxOut (..))
 import Cardano.Ledger.TxIn (TxId (..), TxIn (..))
+
+-- import Cardano.Ledger.UMap(UMap(..),RDPair(..),UMElem(..))
 import Cardano.Ledger.UMap
 import Cardano.Ledger.UTxO
 import Cardano.Ledger.Val (Val)
 import Constrained hiding (Value)
 import Constrained qualified as C
-import Constrained.Base (HasGenHint (..))
+import Constrained.Base (Binder (..), HasGenHint (..), Pred (..), Term (..))
 import Constrained.Spec.Map
+import Constrained.Univ ()
 import Control.Monad.Trans.Fail.String
 import Crypto.Hash (Blake2b_224)
 import Data.ByteString qualified as BS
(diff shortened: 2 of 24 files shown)


libs/constrained-generators/src/Constrained/Spec/Map.hs
```diff
@@ -214,9 +216,28 @@ instance BaseUniverse fn => Functions (MapFn fn) fn where
...
                 _ -> ErrorSpec ["Rng on bad map spec", show spec]
+    RngSet ->
+      case fn of
+        (_ :: MapFn fn '[Map k v] (Set v))
+          | NilCtx HOLE <- ctx
+          , Evidence <- prerequisites @fn @(Map k v) ->
+              case spec of
+                MemberSpec [r] ->
+                  typeSpec $ MapSpec Nothing Set.empty (Set.toList r) (equalSpec $ sizeOf r) TrueSpec NoFold
+                TypeSpec (SetSpec must elemspec _size) [] ->
```
MaximilianAlgehed: Ignoring the `size` here is not sound.

libs/constrained-generators/src/Constrained/Spec/Map.hs
```diff
@@ -214,9 +216,28 @@ instance BaseUniverse fn => Functions (MapFn fn) fn where
                       Set.empty
                       must
                       size
-                      (constrained $ \kv -> satisfies (app (sndFn @fn) (app (toGenericFn @fn) kv)) elemspec)
+                      -- (constrained $ \kv -> satisfies (app (sndFn @fn) (app (toGenericFn @fn) kv)) elemspec)
```
MaximilianAlgehed:
```suggestion
```

libs/constrained-generators/src/Constrained/Spec/Map.hs
```diff
@@ -123,6 +123,8 @@ instance
             mustVals'
             size'
             (simplifySpec $ constrained $ \v -> unsafeExists $ \k -> pair_ k v `satisfies` kvs)
+            -- old before rebase on master
+            -- (constrained $ \v -> unsafeExists $ \p -> p `satisfies` kvs <> assert (snd_ p ==. v))
```
MaximilianAlgehed:
```suggestion
```

libs/constrained-generators/src/Constrained/Spec/Map.hs
MaximilianAlgehed: The problem with making this sound is that you have to deal with the case where there are positive lower bounds on the size. This in turn means that the size constrains the number of _different_ elements in the range.

libs/constrained-generators/src/Constrained/Base.hs
```diff
@@ -1100,9 +1128,9 @@ simplifySpec spec = case spec of
 -- | Precondition: the `Pred fn` defines the `Var a`
 computeSpecSimplified ::
   forall fn a. (HasSpec fn a, HasCallStack) => Var a -> Pred fn -> Specification fn a
-computeSpecSimplified x p = fromGESpec $ case p of
-  Monitor {} -> pure mempty
+computeSpecSimplified x p = fromGE ErrorSpec $ explain ["While simplifying", show p] $ case p of
```
MaximilianAlgehed:
```suggestion
computeSpecSimplified x p = fromGESpec $ case p of
```

This is being addressed in #4317

libs/constrained-generators/src/Constrained/Base.hs
```diff
@@ -860,7 +888,7 @@ genFromSpec ::
 genFromSpec (simplifySpec -> spec) = case spec of
   TrueSpec -> genFromSpec @fn (typeSpec $ emptySpec @fn @a)
   MemberSpec as
-    | null as -> genError ["MemberSpec {}"]
+    | null as -> genError ["MemberSpec {}"] -- HERE IS WHERE THE ERROR COMES FROM
```
MaximilianAlgehed:
```suggestion
    | null as -> genError ["MemberSpec {}"]
```

libs/constrained-generators/src/Constrained/Base.hs
```diff
@@ -596,15 +603,17 @@ instance HasSpec fn a => Semigroup (Specification fn a) where
...
+    [] -> nullMemberMember as as'
+    other -> MemberSpec other
+  rhs@(MemberSpec as) <> lhs@(TypeSpec s cant) =
+    case filter (flip (conformsTo @fn) s) (filter (`notElem` cant) as) of
+      x@(_ : _) -> MemberSpec x
+      [] ->
+        ErrorSpec
+          [ "\nNull intersection of MemberSpecs and TypeSpec"
+          , "  " ++ show rhs ++ " <> " ++ show lhs
+          ]
```
MaximilianAlgehed, line 635: This is being addressed in #4317

libs/constrained-generators/src/Constrained/Base.hs
```diff
@@ -618,6 +627,25 @@ instance HasSpec fn a => Semigroup (Specification fn a) where
...
+  where
+    more = case (xs, ys) of
+      ([x], [_]) ->
+        [ "  One possible cause, a term 'x' with two equality predicates, Something like this:"
+        , "  [" ++ show x ++ " ==. x, x ==. y]"
+        , "  (==.) forces the right argument to be solved for first. Causing the conflict"
+        , "  [" ++ show x ++ " ==. x, y ==. x].  Might be solveable depending on how constrained 'y' is."
+        ]
+      _ -> []
+
```
MaximilianAlgehed, line 766: This is being addressed in #4317

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/Conway/Gov.hs
```diff
@@ -241,6 +242,22 @@ onHardFork ::
...
+      [ assert $ majV <=. majV'
+      , ifElse
+          (majV ==. lit maxBound)
+          (majV' ==. majV)
+          (majV' <=. succV_ majV)
+      , ifElse
+          (majV ==. majV')
+          (minV' ==. minV + 1)
+          (minV' ==. 0)
+      , minV' `dependsOn` majV'
```
MaximilianAlgehed: I can't find a use site for this anywhere?

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/Conway/Instances.hs
```diff
@@ -1339,3 +1444,440 @@ onSized ::
...
+  updateToPPU x = (uLiftConway x . uLiftBabbage x . uLiftAlonzo x . uLiftShelley) x
+  ppuToUpdate x = uDropConway x $ uDropBabbage x $ uDropAlonzo x $ uDropShelley x
+
+-- ===============================================================
+
+-- | SimpleRep instance for PParams
+instance EraPP era => HasSimpleRep (PParams era) where
+  type SimpleRep (PParams era) = PParamSubset
+  toSimpleRep = ppToSubset
+  fromSimpleRep = subsetToPP
```
MaximilianAlgehed, line 1689: It might be worth while to add the selectors like
```haskell
minFeeA_ :: (IsConwayUniv fn, EraPP era) => Term fn (PParams era) -> Term fn Coin
minFeeA_ = sel @0
```

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/Conway/Instances.hs
```diff
@@ -1339,3 +1444,440 @@ onSized ::
...
+uLiftConway pps pp =
+  pp
+    & ppuCommitteeMaxTermLengthL .~ (ucommitteeMaxTermLength pps)
+    & ppuGovActionLifetimeL .~ (ugovActionLifetime pps)
+    & ppuGovActionDepositL .~ (ugovActionDeposit pps)
+    & ppuDRepDepositL .~ (udRepDeposit pps)
+
+-- | Use then generic HasSimpleRep and HasSpec instances for PParamSubset
+instance HasSimpleRep PPUpdate
+
```
MaximilianAlgehed, line 2059:
```suggestion
```

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/Conway/Instances.hs
```diff
@@ -1339,3 +1444,440 @@ onSized ::
...
+protocolVersion_ :: IsConwayUniv fn => Term fn PParamSubset -> Term fn ProtVer
+protocolVersion_ = sel @13
+
+maxTxSize_ :: IsConwayUniv fn => Term fn PParamSubset -> Term fn Word32
+maxTxSize_ = sel @3
+
+-- | Use then generic HasSimpleRep and HasSpec instances for PParamSubset
+instance HasSimpleRep PParamSubset
+
```
MaximilianAlgehed, line 1747:
```suggestion
```

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/Conway/Gov.hs
```diff
@@ -402,52 +419,48 @@ wfPParamsUpdate ::
   IsConwayUniv fn =>
   Term fn (PParamsUpdate (ConwayEra StandardCrypto)) ->
   Pred fn
-wfPParamsUpdate ppu =
+wfPParamsUpdate pparamsupdate =
```
MaximilianAlgehed: If we are going to have multi-word names I would strongly prefer them to be `camelCase` to make them easier to read.

libs/constrained-generators/src/Constrained/Base.hs
MaximilianAlgehed: Not the latter case, sorry. That one might make sense to include here. However, the extra explanation in `MemberSpec` shouldn't be necessary now because we provide much more contexts in the failing cases now.

TimSheard: OK, I will leave this here for now. After PR #4317 is merged into master, I will rebase and set things straight.

libs/constrained-generators/src/Constrained/Base.hs
TimSheard: Same as before. Fix it when we rebase on master, after PR #4317 is merged into master.

libs/constrained-generators/src/Constrained/Spec/Map.hs
TimSheard: Might we try this:
```
TypeSpec (SetSpec must elemspec setsize) [] ->
                  typeSpec $
                    MapSpec
                      Nothing
                      Set.empty
                      (Set.toList must)
                      ((geqSpec (fromIntegral (Set.size must))) <> setsize) -- This might be overconstrained.
                      (constrained $ \kv -> satisfies (app (sndFn @fn) (app (toGenericFn @fn) kv)) elemspec)
                      NoFold
```
I would reason as follows. The set must contain the elements in `must` so the smallest the mat can be is the length of `must`.  It may contain other things, So if it does then`setsize` constrains how many. So if the SetSpec is consistent, the new MapSpec should be too.  Is there something else I am missing here?

MaximilianAlgehed: I think it would be a good idea to use `prop_sound` to test your ideas here. A good basic test is:
```haskell
rngSpec :: Specification BaseFn (Map Int Int)
rngSpec = constrained $ \ m -> 3 <=. sizeOf_ (rngSet_ m)
```

MaximilianAlgehed:
@TimSheard may I suggest - to make the `rngSet_` thing easier to implement - that you rebase this on #4325 and add a `TestableFn` for `rngSetFn`? That way you'd get a lot more certainty about the implementation.

Another thing I've been thinking about is it it wouldn't be better to just bite the bullet and implement `toSet_ :: Term fn [a] -> Term fn (Set a)` instead of special casing it for the range of maps?

libs/constrained-generators/src/Constrained/Spec/Map.hs
TimSheard: Ok I added this test, made every thing work. So this should be resolved

TimSheard:
So my experience with maps suggests a solution for List. We can add that later.

libs/constrained-generators/src/Constrained/GenT.hs
```diff
@@ -200,13 +203,13 @@ suchThatT g p = do
...
   let (n, cont) = case mode of
         Strict -> (100, fatalError)
-        Loose -> (1 :: Int, genError) -- TODO: Maybe 1 is not the right number here!
-  go n cont
-  where
-    go 0 cont = cont ["Ran out of tries on suchThatT"]
-    go n cont = do
-      a <- g
-      if p a then pure a else scaleT (+ 1) $ go (n - 1) cont
+        Loose -> (100 :: Int, genError) -- TODO: Maybe 1 is not the right number here!
```
Soupstraw, line 206: This comment seems outdated

libs/constrained-generators/src/Constrained/Examples/Map.hs
```diff
@@ -66,3 +71,72 @@ rangeSumSize = constrained $ \m ->
...
+
+w5 :: Specification BaseFn (Map Int Int)
+w5 = constrained $ \m -> [assert $ m /=. lit mempty]
+
+w6 :: Specification BaseFn (Map Int Int)
+w6 = constrained $ \m -> [assert $ m ==. lit mempty]
+
+-- | When a test on a Map spec fails, this function traces the steps, to help
+--   discover what went wrong.
+manualMapTest ::
```
Soupstraw, line 140: Could this be turned into a test instead?

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/Conway/Utxo.hs
```diff
@@ -10,70 +10,33 @@ module Test.Cardano.Ledger.Constrained.Conway.Utxo where
...
-                 _cppDRepVotingThresholds
-                 _cppCommitteeMinSize
-                 _cppCommitteeMaxTermLength
-                 _cppGovActionLifetime
-                 _cppGovActionDeposit
-                 _cppDRepDeposit
-                 _cppDRepActivity
-                 _cppMinFeeRefScriptCoinsPerByte ->
-                    -- NOTE: this is for testing only! We should figure out a nicer way
-                    -- of splitting generation and checking constraints here!
```
MaximilianAlgehed, line 74: It would be nice if we could retain this comment.

libs/constrained-generators/src/Constrained/Base.hs
```diff
@@ -648,9 +657,9 @@ instance (HasSpec fn a, Arbitrary (TypeSpec fn a)) => Arbitrary (Specification f
               TypeSpec <$> arbitrary <*> vectorOf len (genFromSpec @fn TrueSpec)
           )
         , (1, ErrorSpec <$> arbitrary)
-        , (1, pure $ MemberSpec [])
-        , -- Recurse to make sure we apply the tricks for generating suspended specs multiple times
-          (1, arbitrary)
+        , --  , (1, pure $ MemberSpec []) -- This is always inconsistent, So I have temporarily remove it subject to discussion T.S.
+          -- Recurse to make sure we apply the tricks for generating suspended specs multiple times
+          (4, arbitrary)
```
MaximilianAlgehed: We probably want to undo these changes.

libs/constrained-generators/src/Constrained/Base.hs
```diff
@@ -3317,14 +3343,18 @@ instance BaseUniverse fn => Functions (SetFn fn) fn where
               ]
     Member
       | HOLE :? Value s :> Nil <- ctx -> caseBoolSpec spec $ \case
-          True -> MemberSpec $ Set.toList s
+          True -> case (Set.toList s) of
+            [] -> ErrorSpec ["member_ HOLE x, where x is the emptyset"]
```
MaximilianAlgehed, line 3420: 👍

libs/constrained-generators/src/Constrained/Base.hs
```diff
@@ -3958,11 +3988,15 @@ instance (BaseUniverse fn, Member (FunFn fn) fn) => Functions (FunFn fn) fn wher
 lessOrEqualFn :: forall fn a. (Ord a, OrdLike fn a) => fn '[a, a] Bool
 lessOrEqualFn = injectFn (LessOrEqual @_ @fn)
 
+greaterOrEqualFn :: forall fn a. (Ord a, OrdLike fn a) => fn '[a, a] Bool
+greaterOrEqualFn = injectFn (GreaterOrEqual @_ @fn)
+
```
MaximilianAlgehed, line 4066: Not necessary any more with the introduction of `flip_`. Let's remove it.

libs/constrained-generators/src/Constrained/Base.hs
```diff
@@ -4807,13 +4847,14 @@ rangeSize a b | a < 0 || b < 0 = error ("Negative Int in call to rangeSize: " ++
...
 -- | The widest interval whose largest element is admitted by the original spec
+--   This is a Size, so if the largest element is 0, then (NumSpecInterval (Just 0) (Just 0))
 maxSpec :: BaseUniverse fn => Specification fn Integer -> Specification fn Integer
 maxSpec TrueSpec = TrueSpec
 maxSpec s@(SuspendedSpec _ _) =
   constrained $ \x -> unsafeExists $ \y -> [y `satisfies` s, Assert ["maxSpec on SuspendedSpec"] (x <=. y)]
 maxSpec (ErrorSpec xs) = ErrorSpec xs
 maxSpec (MemberSpec []) = ErrorSpec ["empty MemberSec in maxSpec."]
-maxSpec (MemberSpec xs) = leqSpec (maximum xs)
+maxSpec (MemberSpec xs) = leqSpec (maximum (0 : xs)) -- TODO IS this right
```
MaximilianAlgehed: This is not right in general, it makes it impossible to use this function when we aren't talking about sizes.

libs/constrained-generators/src/Constrained/GenT.hs
MaximilianAlgehed, line 206: Why is it necessary to change this to `100`? It will slow a lot of things down by about a factor 100.

libs/constrained-generators/src/Constrained/Examples/Map.hs
```diff
@@ -66,3 +71,72 @@ rangeSumSize = constrained $ \m ->
...
+
+w4 :: Specification BaseFn (Map Int Int)
+w4 = constrained $ \m ->
+  [assert $ sizeOf_ (rngSet_ m) ==. 4]
+
+w5 :: Specification BaseFn (Map Int Int)
+w5 = constrained $ \m -> [assert $ m /=. lit mempty]
+
+w6 :: Specification BaseFn (Map Int Int)
+w6 = constrained $ \m -> [assert $ m ==. lit mempty]
```
MaximilianAlgehed: I would like these to have proper names and for all new examples to be added to the test suite if possible.

libs/constrained-generators/src/Constrained/Examples/Set.hs
```diff
@@ -125,10 +125,17 @@ setSpec = constrained $ \ss ->
...
+size100Spec = constrained $ \s -> [assert $ size_ s <=. 100]
+
 compositionalSpec :: Specification BaseFn (Set Int)
 compositionalSpec = constrained $ \x ->
   [ satisfies x setSpec
   , assert $ 0 `member_` x
+  , satisfies x size100Spec -- Has a tendency to occaisionally fail when sizes over 100 are picked
+  -- Because it resorts to picking something not in a large set. We only get
+  -- 100 trys, so if the set is large the probability increases. Adding this
+  -- should not affect the property being tested.
```
MaximilianAlgehed, line 138: This didn't used to be an issue, if it is now that's a regression we need to avoid.

libs/constrained-generators/src/Constrained/Spec/Map.hs
```diff
@@ -52,21 +62,33 @@ data MapSpec fn k v = MapSpec
   , mapSpecSize :: Specification fn Integer
   , mapSpecElem :: Specification fn (k, v)
   , mapSpecFold :: FoldSpec fn v
+  , mapSpecMinValSize :: Maybe (MinValSize fn v)
+  -- Specifaction of the number of unique Values in the range of the Map that meet the Mapspec.
+  -- The Maybe is there because Nothing means (MinValSpec TrueSpec),
+  -- but does require the (Ord v) constraint
```
MaximilianAlgehed, line 76: does not require the Ord v constraint?

libs/constrained-generators/src/Constrained/Spec/Map.hs
```diff
@@ -308,7 +350,537 @@ dom_ ::
...
+          then 0 :: Integer
+          else sum mustval `div` toInteger (length mustval)
+      goodMust = constrained (\n -> n >=. (lit avg))
+  mustkey <- genFromGenT $ (Set.fromList <$> genDistinctList @_ @fn goodMust mustKeyLen [])
+  let kvSpec =
+        constrained
+          ( \p ->
+              match p (\key val -> [assert $ key >=. val])
+          )
+  pure (MapSpec Nothing mustkey mustval mapsize kvSpec NoFold valsize)
```
MaximilianAlgehed, line 840: I would be good with replacing the current arbitrary instance with this if we can also generate the old garbage map instance ~1/3 of the time to make sure we cover all the really nasty cases.

libs/constrained-generators/src/Constrained/Spec/Map.hs
```diff
@@ -308,7 +350,537 @@ dom_ ::
...
+    tell :: String -> Bool -> Maybe String
+    tell _ True = Nothing
+    tell msg False = Just msg
+
+    allOf :: [Maybe String] -> Maybe String
+    allOf xs = foldr accum Nothing xs
+      where
+        accum Nothing ans = ans
+        accum (Just s) Nothing = Just s
+        accum (Just s) (Just more) = Just (s ++ "\n" ++ more)
```
MaximilianAlgehed, line 935: Perhaps better to do this using the `GE` machinery we already use extensively in this code base?

libs/constrained-generators/test/Constrained/Test.hs
```diff
@@ -151,10 +155,11 @@ tests nightly =
       prop "[Int]" $ prop_conformEmpty @BaseFn @[Int]
       prop "[(Int, Int)]" $ prop_conformEmpty @BaseFn @[(Int, Int)]
     prop "prop_univSound @BaseFn" $
-      withMaxSuccess (if nightly then 100_000 else 10_000) $
+      withMaxSuccess 100 $ -- (if nightly then 100_000 else 10_000) $
```
MaximilianAlgehed: Remove this.

libs/constrained-generators/test/Constrained/Test.hs
```diff
@@ -151,10 +155,11 @@ tests nightly =
...
         prop_univSound @BaseFn
     describe "prop_gen_sound @BaseFn" $ do
-      modifyMaxSuccess (const $ if nightly then 10_000 else 1000) $ do
+      modifyMaxSuccess (const 100) $ do
+        -- (const $ if nightly then 10_000 else 1000) $ do
```
MaximilianAlgehed: Remove this.

libs/constrained-generators/src/Constrained/Spec/Map.hs
```diff
@@ -271,12 +278,40 @@ instance BaseUniverse fn => Functions (MapFn fn) fn where
...
                       size
-                      (constrained $ \kv -> satisfies (app (sndFn @fn) (app (toGenericFn @fn) kv)) elemspec)
+                      (constrained $ \kv -> satisfies (snd_ kv) elemspec)
                       foldspec
+                      Nothing
                 -- NOTE: you'd think `MemberSpec [r]` was a safe and easy case. However, that requires not only that the elements
                 -- of the map are fixed to what is in `r`, but they appear in the order that they are in `r`. That's
                 -- very difficult to achieve!
                 _ -> ErrorSpec ["Rng on bad map spec", show spec]
+    RngSet ->
```
MaximilianAlgehed, line 289: What would we need to add elsewhere to be able to use `fromList_ . rng_` instead of `rngSet_` to make things more general?

MaximilianAlgehed (changes requested):
I haven't had time to look at the new style generators for maps in detail yet, but I trust that once we run enough tests we'll be quite sure they work correctly enough. It's important that we don't reduce the number of tests we run here because this is very tricky territory.

libs/constrained-generators/src/Constrained/GenT.hs
TimSheard: Yes it is. Been removed.

libs/constrained-generators/src/Constrained/Examples/Map.hs
TimSheard: This is a tool. It helps when a (Specification fn (Map k v)) fails to generate something. It helps debug what went wrong.

libs/cardano-ledger-test/src/Test/Cardano/Ledger/Constrained/Conway/Utxo.hs
TimSheard: OK I added back, in an appropriate place.

libs/constrained-generators/src/Constrained/Base.hs
TimSheard: I guess since this is an arbitrary instance. We want it to generate insconsistent things once and a while. I added it back

libs/constrained-generators/src/Constrained/Base.hs
TimSheard: This is only used when computing cardianlities. So it is always a Size.  I have renamed it to maxSpecSize, and added a comment about how it should be used.

libs/constrained-generators/src/Constrained/GenT.hs
TimSheard: I have examples, run in the Strict mode that needed multiple tries to succeed. I don't think this is called alot, in strict mode, (but it is called), and I didn't notice any slowdown.

libs/constrained-generators/src/Constrained/Examples/Map.hs
TimSheard: done

libs/constrained-generators/src/Constrained/Spec/Map.hs
TimSheard: The (Ord v) constraint is carried by the MinValSize constructor, So it is satisfied locally, only when it is used.  If that was not the case, every MapSpec would need an (Ord v) constraint.
If this isn't clear we shpould talk about it.

libs/constrained-generators/src/Constrained/Spec/Map.hs
TimSheard: That is not what I am getting at.  When come tries to compute the most accurate Size for a MapSpec,
many things constribute to the answer. If one of those things is incossitent, there is no way to tell. Things just return ErrorSpec. This  labels aech thing and reports which ones fail.

libs/constrained-generators/test/Constrained/Test.hs
TimSheard: done

libs/constrained-generators/test/Constrained/Test.hs
TimSheard: done

libs/constrained-generators/src/Constrained/Spec/Map.hs
TimSheard: I think this just works.

libs/constrained-generators/src/Constrained/Spec/Map.hs
TimSheard: We would have to generalize it from (Map Integer Integer) to (Map k v), lets do this in a follow on PR.

libs/constrained-generators/src/Constrained/Base.hs
TimSheard: follow on PR

lehins:
Closing in favor of #4575

PR #4315, opened 2024-05-02
`constrained-generators`: Improve error messages and make the tree generator reasonably sized

MaximilianAlgehed:
# Description

The tree generator used to produce trees of size at most 8000, now it defaults to 20 unless otherwise specified.

This also makes it so that the `ErrorSpec` we get from a bunch of operations actually carries the `explain` calls with it to
improve error logging.

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/libs/constrained-generators/src/Constrained/Spec/Tree.hs
+++ b/libs/constrained-generators/src/Constrained/Spec/Tree.hs
@@ -45,7 +45,7 @@ data BinTreeSpec fn a = BinTreeSpec (Maybe Integer) (Specification fn (BinTree a
   deriving (Show)
 
 instance Forallable (BinTree a) (BinTree a, a, BinTree a) where
-  forAllSpec = typeSpec . BinTreeSpec Nothing
+  fromForAllSpec = typeSpec . BinTreeSpec Nothing
   forAllToList BinTip = []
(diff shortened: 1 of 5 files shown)


libs/constrained-generators/src/Constrained/Spec/Tree.hs
```diff
@@ -93,29 +94,29 @@ instance HasSpec fn a => HasSpec fn (BinTree a) where
       : (BinNode left a <$> shrinkWithTypeSpec s right)
       ++ ((\l -> BinNode l a right) <$> shrinkWithTypeSpec s left)
 
-  toPreds t (BinTreeSpec sz s) =
+  toPreds t (BinTreeSpec msz s) =
     (forAll t $ \n -> n `satisfies` s)
-      <> genHint sz t
+      <> maybe TruePred (flip genHint t) msz
```
Soupstraw, line 99: What's the idea here? Why do we only give the generator a hint if the size is given in the spec and set it to `TrueSpec` otherwise?

Soupstraw (approved):

libs/constrained-generators/src/Constrained/Spec/Tree.hs
MaximilianAlgehed: The idea is that because `genHint` is optional we only want to have it in the predicates if the user has explicitly given one.

PR #4317, opened 2024-05-03
`constrained-generators`: Fix bug in reifies

MaximilianAlgehed:
# Description

When looking at improving some error messages I encountered a bug in reifies whereby `reifies 10 x id` would simplify away to `TrueSpec`. This was an oopsie introduced by yours truly but here is a fix!

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/libs/constrained-generators/src/Constrained/Base.hs
+++ b/libs/constrained-generators/src/Constrained/Base.hs
@@ -1113,6 +1113,8 @@ simplifySpec spec = case spec of
   TrueSpec -> TrueSpec
 
 -- | Precondition: the `Pred fn` defines the `Var a`
+--
+-- Runs in `GE` in order for us to have detailed context on failure.
 computeSpecSimplified ::
   forall fn a. (HasSpec fn a, HasCallStack) => Var a -> Pred fn -> GE (Specification fn a)
 computeSpecSimplified x p = localGESpec $ case p of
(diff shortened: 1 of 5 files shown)


libs/constrained-generators/src/Constrained/Spec/Generics.hs
```diff
@@ -372,7 +372,7 @@ forAll' xs f = forAll xs $ \x -> match @fn @p x f
 
 -- | Like `constrained` but pattern matches on the bound `Term fn a`
 constrained' ::
-  forall fn a p.
```
MaximilianAlgehed, line 375: This is just to make sure the order of type arguments is the same in `constrained` and `constrained'`

MaximilianAlgehed:
Note, this also significantly improves error messages by turning what used to be a local call to `error` into a much later call to `error` - thereby preserving a bunch of context that would have disappeared before.

MaximilianAlgehed:
I've also added a bunch more error reporting and fixed a bug in the dependency calculator here too.

libs/constrained-generators/src/Constrained/Base.hs
```diff
@@ -1092,60 +1099,76 @@ explainSpec _ s = s
...
   MemberSpec xs -> MemberSpec (nub xs)
   ErrorSpec es -> ErrorSpec es
   TypeSpec ts cant -> TypeSpec ts (nub cant)
   TrueSpec -> TrueSpec
 
 -- | Precondition: the `Pred fn` defines the `Var a`
 computeSpecSimplified ::
-  forall fn a. (HasSpec fn a, HasCallStack) => Var a -> Pred fn -> Specification fn a
-computeSpecSimplified x p = fromGESpec $ case p of
+  forall fn a. (HasSpec fn a, HasCallStack) => Var a -> Pred fn -> GE (Specification fn a)
```
TimSheard: Comment that we run this in the GE  monad, in order to collect and display more accurate context when we report errors.

libs/constrained-generators/src/Constrained/Spec/Generics.hs
TimSheard: seems like a good idea

TimSheard (approved):
Changes some of the code that simplifies Preds to run in the GE monad, to enable the collection (and display) of of more accurate context when an error occurs. Also includes some better pretty printers so that the context displayed is easier to read and understand.

Only time will tell if this is the right level of detail.

PR #4320, opened 2024-05-06
CERT conformance

Soupstraw:
# Description

Adds conformance testing for `CERT` rule. 

There is an issue with DRep registration certificates that I couldn't figure out so I just modified the constraints so that these certificates aren't generated until we figure out what's causing the problem. 

I have also gotten rid of the `TestRep` associated type, since it was annoying to work with. I've replaced `TestRep` with the `FixupSpecRep` typeclass, which also has a generic instance which fixes up these values recursively. 

This PR also moves the check that checks if a stake pool delegatee is registered from `CERTS` rule to `DELEG` rule since it makes a lot more sense there and has already caused much confusion.

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -13,8 +13,8 @@ repository cardano-haskell-packages
 source-repository-package
   type: git
   location: https://github.com/input-output-hk/cardano-ledger-executable-spec.git
-  tag: ac7af152d7d4c7b3ef1f3513e637c2c1e86ca30e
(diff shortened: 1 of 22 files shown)


MaximilianAlgehed (approved):
I don't understand all the back-and-forth to Agda bits of this (perhaps we should sit down some day and go over it together?) but the constraint bits make perfect sense to me.

aniketd (approved):
Awesome :bow:

PR #4323, opened 2024-05-07
`constrained-generators`: clean up interface

MaximilianAlgehed:
# Description

Removes some TODOs in the interface and cleans things up a bit to reduce the scope for introducing breaking changes later on.

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/libs/constrained-generators/src/Constrained.hs
+++ b/libs/constrained-generators/src/Constrained.hs
@@ -5,10 +5,9 @@
 -- It is supposed to contain everything you need to write constraints
 -- and implement `HasSpec` and `HasSimpleRep` instances.
 module Constrained (
-  Specification (..), -- TODO: it would be nice not to expose this whole thing
-  Pred (..), -- TODO: it would be nice not to expose this whole thing
+  Specification (..),
+  Pred,
   Term,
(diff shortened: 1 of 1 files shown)


libs/constrained-generators/src/Constrained.hs
```diff
@@ -112,13 +111,8 @@ module Constrained (
   shrinkWithNumSpec,
   conformsToNumSpec,
   toPredsNumSpec,
-  -- TODO: this is super yucky, it would be good to implement
-  -- the linear lambda thing to get rid of this.
```
teodanciu, line 116: I don't know what this means but I support the principle: the fewer super yucky things the better!

teodanciu (approved):
Looks good to me

teodanciu (commented):
I guess we're not yet releasing this on Chap, so no need to increment the version (as it would the case if we releasing)?

MaximilianAlgehed:
> I guess we're not yet releasing this on Chap, so no need to increment the version (as it would the case if we releasing)?

Indeed. I'm hoping to clean as much of this stuff up as possible before we release on chap.

PR #4325, opened 2024-05-08
`constrained-generators`: soundness tests and bugfixes

MaximilianAlgehed:
# Description

In this PR we introduce a basic soundness testing scheme for `propagateSpecFun`. The idea is that you take a random function symbol `fn`, a random context `ctx = v0 v1 ... x ... vn` such that `fn v0 v1 ... x ... vn` is well typed, you generate a random `spec` for the result type of `fn` and you propagate it through the `fn` and the `ctx` to give you a spec for `x`. You then generate a value `a` for `x` and evaluate `fn v0 v1 ... a ... vn` and check that the result conforms to `spec`.

This has found a bunch of tiny soundness bugs and a nasty little bug in the optimizer. It will greatly help us avoid regressions, esp. when we add new function symbols.

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/libs/constrained-generators/src/Constrained/List.hs
+++ b/libs/constrained-generators/src/Constrained/List.hs
@@ -17,6 +17,7 @@ module Constrained.List where
 
 import Data.Functor.Const
 import Data.Kind
+import Data.Semigroup (Sum (..))
 
 -- | A heterogeneous list / an inductive tuple.
 -- We use this heavily to represent arguments to
@@ -44,6 +45,15 @@ mapMList :: Applicative m => (forall a. f a -> m (g a)) -> List f as -> m (List 
 mapMList _ Nil = pure Nil
 mapMList f (x :> xs) = (:>) <$> f x <*> mapMList f xs
 
+mapMListC ::
+  forall c as f g m.
+  (Applicative m, All c as) =>
+  (forall a. c a => f a -> m (g a)) ->
+  List f as ->
+  m (List g as)
+mapMListC _ Nil = pure Nil
+mapMListC f (x :> xs) = (:>) <$> f x <*> mapMListC @c f xs
+
 foldMapList :: Monoid b => (forall a. f a -> b) -> List f as -> b
 foldMapList _ Nil = mempty
 foldMapList f (a :> as) = f a <> foldMapList f as
@@ -56,6 +66,9 @@ foldMapListC f (a :> as) = f a <> foldMapListC @c f as
 appendList :: List f as -> List f bs -> List f (Append as bs)
 appendList Nil bs = bs
 appendList (a :> as) bs = a :> appendList as bs
+
+lengthList :: List f as -> Int
+lengthList = getSum . foldMapList (const $ Sum 1)
 
 -- | Append two type-level lists
 type family Append as as' where
@@ -112,6 +125,17 @@ toWholeCtx (x :! xs)
 fromWholeCtx :: ListCtxWhole f as c -> ListCtx f as c
 fromWholeCtx (ListCtxWhole Nil hole suf) = hole :? suf
 fromWholeCtx (ListCtxWhole (x :> pre) hole suf) = x :! fromWholeCtx (ListCtxWhole pre hole suf)
+
+fillListCtx :: ListCtx f as c -> (forall a. c a -> f a) -> List f as
+fillListCtx (ListCtx pre c post) f = appendList pre (f c :> post)
+
+mapListCtx :: (forall a. f a -> g a) -> ListCtx f as c -> ListCtx g as c
+mapListCtx nt (ListCtx pre c post) = ListCtx (mapList nt pre) c (mapList nt post)
+
+mapListCtxC ::
+  forall c as f g h. All c as => (forall a. c a => f a -> g a) -> ListCtx f as h -> ListCtx g as h
+mapListCtxC nt (h :? as) = h :? mapListC @c nt as
+mapListCtxC nt (a :! ctx) = nt a :! mapListCtxC @c nt ctx
 
 -- | A function type from `ts` to `res`:
 --    FunTy '[Int, Bool] Double = Int -> Bool -> Double
--- a/libs/constrained-generators/src/Constrained/Properties.hs
+++ b/libs/constrained-generators/src/Constrained/Properties.hs
@@ -1,14 +1,25 @@
 {-# LANGUAGE AllowAmbiguousTypes #-}
 {-# LANGUAGE DataKinds #-}
+{-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE FlexibleInstances #-}
+{-# LANGUAGE GADTs #-}
 {-# LANGUAGE ImportQualifiedPost #-}
+{-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE RankNTypes #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE TypeApplications #-}
+{-# LANGUAGE TypeOperators #-}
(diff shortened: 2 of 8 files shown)


libs/constrained-generators/src/Constrained/List.hs
```diff
@@ -113,6 +126,17 @@ fromWholeCtx :: ListCtxWhole f as c -> ListCtx f as c
 fromWholeCtx (ListCtxWhole Nil hole suf) = hole :? suf
 fromWholeCtx (ListCtxWhole (x :> pre) hole suf) = x :! fromWholeCtx (ListCtxWhole pre hole suf)
 
+fillListCtx :: ListCtx f as c -> (forall a. c a -> f a) -> List f as
+fillListCtx (ListCtx pre c post) f = appendList pre (f c :> post)
+
+mapListCtx :: (forall a. f a -> g a) -> ListCtx f as c -> ListCtx g as c
+mapListCtx nt (ListCtx pre c post) = ListCtx (mapList nt pre) c (mapList nt post)
```
Soupstraw, line 133: I wonder if it's possible to define this in terms of `mapListCtxC`, since it looks like a more constrained version of that function.

MaximilianAlgehed: Annoyingly, not quite (see the duplication in the other functions of similar type). The problem is that this would require `All SomeTrivialConstraint as` as a precondition to `mapListCtx`. Unfortunately (for good reason) you can't write `All (\_ -> ()) as` and have it not pollute everything.

MaximilianAlgehed: Your intuition is right, but GHCs hamstrung dependent types make it impossible in practice.

libs/constrained-generators/src/Constrained/Spec/Map.hs
```diff
@@ -56,13 +56,59 @@ data MapSpec fn k v = MapSpec
...
+    MapSpec
+      <$> arbitrary
+      <*> arbitrary
+      <*> arbitrary
+      <*> arbitrary
+      <*> arbitrary
+      <*> frequency [(5, pure NoFold), (1, arbitrary)]
+
+instance Arbitrary (FoldSpec fn (Map k v)) where
+  arbitrary = pure NoFold
```
Soupstraw, line 82: I guess you'll want to make this more interesting in the future? Maybe add a `TODO` so you don't forget.

MaximilianAlgehed: Right now it works well enough, the problem is that it causes unnecessarily many calls to `genFromSpec_` to fail.

MaximilianAlgehed: But I've added a TODO.

libs/constrained-generators/src/Constrained/Properties.hs
```diff
@@ -69,20 +91,130 @@ prop_conformEmpty ::
...
+            let sspec = simplifySpec (propagateSpecFun fn ctx spec)
+             in QC.counterexample ("\n" ++ show ("propagateSpecFun fn ctx spec =" /> pretty sspec)) $
+                  QC.forAllBlind (strictGen $ genFromSpec @_ @_ @GE sspec) $ \ge ->
+                    fromGEDiscard $ do
+                      a <- ge
+                      let res = uncurryList_ unValue (sem fn) $ fillListCtx ctx $ \HOLE -> Value a
+                      pure $
+                        QC.counterexample ("\ngenerated value: a = " ++ show a) $
+                          QC.counterexample ("\nfn ctx[a] = " ++ show res) $
+                            conformsToSpecProp res spec
```
Soupstraw, line 110: I think it would make sense to create a monad that can get rid of this staircasing, since this pattern appears quite often.

MaximilianAlgehed: It would indeed, but thats a QuickCheck problem I think rather than something to worry about here. I'm more inclined in this case to say that it's a formatter issue to be honest.

libs/constrained-generators/src/Constrained/Spec/Map.hs
Soupstraw: Yeah, I just figured you wanted to add something more here, since just above this you have 
```haskell
frequency [(5, pure NoFold), (1, arbitrary)]
```
which is as good as having `arbitrary` there instead

Soupstraw (approved):

libs/constrained-generators/src/Constrained/Spec/Map.hs
MaximilianAlgehed: Yeah, this was actually an attempt to solve this exact problem of unsat distributions. If you have too many fold constraints you end up with most maps being completely unsolveable. That's why this tweaks the underlying distribution to generate meaningful constraints here more rarely.

libs/constrained-generators/src/Constrained/List.hs
```diff
@@ -57,6 +67,9 @@ appendList :: List f as -> List f bs -> List f (Append as bs)
 appendList Nil bs = bs
 appendList (a :> as) bs = a :> appendList as bs
 
+lengthList :: List f as -> Int
+lengthList = getSum . foldMapList (const $ Sum 1)
```
TimSheard, line 71: Do we want this to be Integer, most of our sizees use Integer, not Int

libs/constrained-generators/src/Constrained/Properties.hs
```diff
@@ -69,20 +91,130 @@ prop_conformEmpty ::
...
+    , Typeable as
+    , QC.Arbitrary (Specification fn b)
+    , Typeable (FunTy as b)
+    ) =>
+    fn as b ->
+    TestableFn fn
+
+instance BaseUniverse fn => Show (TestableFn fn) where
+  show (TestableFn (fn :: fn as b)) =
+    show fn ++ " :: " ++ show (typeOf (undefined :: FunTy as b))
```
TimSheard, line 159: Will the (BaseUniverse fn) restict us in any way?
Are there some Conway types, that have functions?

libs/constrained-generators/test/Constrained/Test.hs
```diff
@@ -135,6 +135,24 @@ tests =
...
+        prop "Map Int Int" $ prop_gen_sound @BaseFn @(Map Int Int)
+        prop "Set Int" $ prop_gen_sound @BaseFn @(Set Int)
+        prop "Set Bool" $ prop_gen_sound @BaseFn @(Set Bool)
+        prop "[Int]" $ prop_gen_sound @BaseFn @[Int]
+        prop "[(Int, Int)]" $ prop_gen_sound @BaseFn @[(Int, Int)]
+        prop "Map Bool Int" $ prop_gen_sound @BaseFn @(Map Bool Int)
+      -- Slow tests that shouldn't run 1000 times
+      prop "Map (Set Int) Int" $ prop_gen_sound @BaseFn @(Map (Set Int) Int)
+      prop "[(Set Int, Set Bool)]" $ prop_gen_sound @BaseFn @[(Set Int, Set Bool)]
+      prop "Set (Set Bool)" $ prop_gen_sound @BaseFn @(Set (Set Bool))
```
TimSheard, line 155: Of course that means we might not be getting very good coverage.

TimSheard (commented):
Looks good to me.

libs/constrained-generators/src/Constrained/Properties.hs
lehins: Thank god for the formatter, otherwise it would be unreadable :stuck_out_tongue_winking_eye: 

In any case, there is no need to complicate the code any more than necessary with extra monads. There is nothing wrong with the indentation and "staircasing"

PR #4328, opened 2024-05-08
Disable drep thresholds in bootstrap

### LINKED ISSUE #4321: Ensure DRep thresholds are ignored in the bootstrap period
It was planned to have DRep thresholds set to 0 for the bootstrap, effectively disabling their requirement to vote. However, that poses a slight problem during the period whenever we will be leaving boorstrap period, because those threshold would have to be enabled only after the hardfork into protocol version 10.

It much simpler an safer to enforce threshold for DRep to zero programmatically in the RATIFY rule when we are in the bootstrap period.

We want make a change to our imp test to reflect this reality, namely setting non-zero DRep thresholds as the default initial state of PParams.

teodanciu:
# Description

Closes https://github.com/IntersectMBO/cardano-ledger/issues/4321

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Internal.hs
+++ b/eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Internal.hs
@@ -305,17 +305,17 @@ toRatifyStatePairs cg@(RatifyState _ _ _ _) =
 pparamsUpdateThreshold ::
   forall era.
   ConwayEraPParams era =>
-  PParams era ->
+  DRepVotingThresholds ->
   PParamsUpdate era ->
   UnitInterval
-pparamsUpdateThreshold pp ppu =
+pparamsUpdateThreshold thresholds ppu =
   let thresholdLens = \case
         NetworkGroup -> dvtPPNetworkGroupL
         GovGroup -> dvtPPGovGroupL
         TechnicalGroup -> dvtPPTechnicalGroupL
         EconomicGroup -> dvtPPEconomicGroupL
       lookupGroupThreshold (PPGroups grp _) =
-        pp ^. ppDRepVotingThresholdsL . thresholdLens grp
+        thresholds ^. thresholdLens grp
    in Set.foldr' max minBound $
         Set.map lookupGroupThreshold $
           modifiedPPGroups @era ppu
@@ -497,7 +497,7 @@ votingDRepThresholdInternal ::
   GovAction era ->
   VotingThreshold
 votingDRepThresholdInternal pp isElectedCommittee action =
-  let DRepVotingThresholds
+  let thresholds@DRepVotingThresholds
         { dvtCommitteeNoConfidence
         , dvtCommitteeNormal
         , dvtUpdateToConstitution
@@ -515,7 +515,7 @@ votingDRepThresholdInternal pp isElectedCommittee action =
               else dvtCommitteeNoConfidence
         NewConstitution {} -> VotingThreshold dvtUpdateToConstitution
         HardForkInitiation {} -> VotingThreshold dvtHardForkInitiation
-        ParameterChange _ ppu _ -> VotingThreshold $ pparamsUpdateThreshold pp ppu
+        ParameterChange _ ppu _ -> VotingThreshold $ pparamsUpdateThreshold thresholds ppu
         TreasuryWithdrawals {} -> VotingThreshold dvtTreasuryWithdrawal
         InfoAction {} -> NoVotingThreshold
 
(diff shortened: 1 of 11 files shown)


eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/EnactSpec.hs
```diff
@@ -61,8 +60,9 @@ relevantDuringBootstrapSpec ::
   ConwayEraImp era =>
   SpecWith (ImpTestState era)
 relevantDuringBootstrapSpec = do
-  actionPrioritySpec
-  hardForkInitiationNoDRepsSpec
+  describe "ENACT (bootstrap phase)" $ do
+    actionPrioritySpec
+    hardForkInitiationNoDRepsSpec
```
teodanciu: This is what I mean in the meeting: these tests are particularly "engineered" to do without DReps, so there is no point to run them if the DRep thresholds are non-zero (because they will fail)

eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Internal.hs
```diff
@@ -512,7 +515,7 @@ votingDRepThresholdInternal pp isElectedCommittee action =
               else dvtCommitteeNoConfidence
         NewConstitution {} -> VotingThreshold dvtUpdateToConstitution
         HardForkInitiation {} -> VotingThreshold dvtHardForkInitiation
-        ParameterChange _ ppu _ -> VotingThreshold $ pparamsUpdateThreshold pp ppu
+        ParameterChange _ ppu _ -> VotingThreshold $ pparamsUpdateThreshold thresholds ppu
```
lehins, line 518: Nice catch! :+1:

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/EnactSpec.hs
```diff
@@ -50,7 +50,6 @@ spec ::
   SpecWith (ImpTestState era)
 spec =
   describe "ENACT" $ do
-    relevantDuringBootstrapSpec
```
lehins, line 53: Instead of removing these, I'd recommend adjusting each affected test and overwrite the threasholds to `0`, whenever `curPV > ProtVer 9`

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/EnactSpec.hs
```diff
@@ -61,8 +60,9 @@ relevantDuringBootstrapSpec ::
   ConwayEraImp era =>
   SpecWith (ImpTestState era)
 relevantDuringBootstrapSpec = do
-  actionPrioritySpec
-  hardForkInitiationNoDRepsSpec
+  describe "ENACT (bootstrap phase)" $ do
```
lehins: It will be better to nest `ENACT` in the `Imp` module
```suggestion
  describe "bootstrap phase" $ do
```

eg.
```haskell
    withImpState @era $ do
      describe "ENACT" $ Enact.relevantDuringBootstrapSpec @era
```

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/EpochSpec.hs
```diff
@@ -63,9 +60,10 @@ relevantDuringBootstrapSpec ::
   ) =>
   SpecWith (ImpTestState era)
 relevantDuringBootstrapSpec = do
-  proposalsSpec
-  dRepSpec
-  eventsSpec
+  describe "EPOCH (bootstrap phase)" $ do
```
lehins: Ditto
```suggestion
  describe "bootstrap phase" $ do
```

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/EnactSpec.hs
lehins: That's why we need to overwrite the thresholds for those tests, instead of completely disabling them

lehins (commented):
Looks good :+1:

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/EnactSpec.hs
teodanciu: done, thank you

eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/Imp/EnactSpec.hs
teodanciu: Yes! This has been bothering me from the beginning - either not being able to run all enact tests with "ENACT" OR getting   "ENACT / ENACT" in the test results - both suck! 
I applied your suggestion and added a "describe RULE" in `Imp` for both versions, so now the results are neat

lehins (approved):
Beautiful! Thank you

PR #4334, opened 2024-05-14
`RATIFY` conformance

Soupstraw:
# Description

Add conformance testing for `RATIFY`

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/libs/cardano-ledger-conformance/src/Test/Cardano/Ledger/Conformance/ExecSpecRule/Conway.hs
+++ b/libs/cardano-ledger-conformance/src/Test/Cardano/Ledger/Conformance/ExecSpecRule/Conway.hs
@@ -1,10 +1,13 @@
 {-# LANGUAGE DataKinds #-}
 {-# LANGUAGE DeriveGeneric #-}
+{-# LANGUAGE EmptyCase #-}
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE FlexibleInstances #-}
 {-# LANGUAGE GADTs #-}
 {-# LANGUAGE InstanceSigs #-}
+{-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE MultiParamTypeClasses #-}
+{-# LANGUAGE NamedFieldPuns #-}
 {-# LANGUAGE RankNTypes #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE StandaloneDeriving #-}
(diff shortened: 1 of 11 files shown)


libs/cardano-ledger-conformance/src/Test/Cardano/Ledger/Conformance/ExecSpecRule/Conway.hs
```diff
@@ -328,3 +356,181 @@ instance
...
+  , Arbitrary (PParamsHKD StrictMaybe era)
+  ) =>
+  Arbitrary (ConwayRatifyExecContext era)
+  where
+  arbitrary =
+    ConwayRatifyExecContext
+      <$> arbitrary
+      <*> (Map.mapWithKey fixupGAIDs <$> arbitrary)
+    where
+      fixupGAIDs gaId gas = gas {gasId = gaId}
```
aniketd, line 453: `assignGAIs` would be a better name, since `fixup` is not specific enough. It would be helpful to refrain from over-using the word `fixup` in the code base, so we don't always need to check the implementation. It would be gainful in the long run to name these functions more specifically, whenever we can. What do you think? :slightly_smiling_face:

libs/cardano-ledger-conformance/src/Test/Cardano/Ledger/Conformance/ExecSpecRule/Conway.hs
```diff
@@ -328,3 +356,181 @@ instance
...
+ratifySignalSpec ::
+  IsConwayUniv fn =>
+  ConwayRatifyExecContext Conway ->
+  Specification fn (RatifySignal Conway)
+ratifySignalSpec ConwayRatifyExecContext {crecGovActionMap} =
+  constrained $ \sig ->
+    match sig $ \gasS ->
+      match gasS $ \gasL ->
+        forAll' gasL $ \gaId _ _ _ _ _ _ ->
+          gaId `member_` dom_ (lit crecGovActionMap)
```
aniketd, line 495: Does this mean we want `gaId` to _be_ a member of `crecGovActionMap`? Just asking on a hunch, because I thought of the likelihood of the intention being a `not member` predicate too. :grinning:

libs/cardano-ledger-conformance/src/Test/Cardano/Ledger/Conformance/Spec/Conway.hs
```diff
@@ -16,6 +16,8 @@ spec :: Spec
 spec = describe "Conway conformance tests" $ do
   xprop "UTXO" $ conformsToImpl @"UTXO" @ConwayFn @Conway
   prop "GOV" $ conformsToImpl @"GOV" @ConwayFn @Conway
+  prop "CERT" $ conformsToImpl @"CERT" @ConwayFn @Conway
+  xprop "RATIFY" $ conformsToImpl @"RATIFY" @ConwayFn @Conway
```
aniketd, line 20: :memo: :raised_hands:

aniketd (approved):
Just some minor comments. Awesome work! :raised_hands:

libs/cardano-ledger-conformance/src/Test/Cardano/Ledger/Conformance/ExecSpecRule/Conway.hs
Soupstraw: This is to work around the fact that in the implementation we only store the `GovActionId`s, but the spec also needs the corresponding `GovActionState` alongside it. The `crecGovActionMap` just provides some sort of mapping from `GovActionId` to `GovActionState` to make it possible to translate to `SpecRep`

libs/cardano-ledger-conformance/src/Test/Cardano/Ledger/Conformance/ExecSpecRule/Conway.hs
Soupstraw: I don't think it really matters here, since this is just a local function, but in general I think it's a good idea to not overuse `fixup` when naming functions

PR #4337, opened 2024-05-17
Fix RATIFY conformance

aniketd:
# Description

This still requires the corresponding `HSSet` de-duplication to work.

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -13,8 +13,7 @@ repository cardano-haskell-packages
 source-repository-package
   type: git
   location: https://github.com/input-output-hk/cardano-ledger-executable-spec.git
-  tag: 81b976c40c2726998a4828220231acf75b3a685e
-  --sha256: sha256-1CiTiLzROoHz8GxVPCouofUFzwzWUclQUq5AthtwPz4=
+  tag: d6effa703ed06ffc6239a963fdbccccd25dfa663
 
 index-state:
   -- Bump this if you need newer packages from Hackage
--- a/cabal.project
+++ b/cabal.project
@@ -10,10 +10,15 @@ repository cardano-haskell-packages
     c00aae8461a256275598500ea0e187588c35a5d5d7454fb57eac18d9edb86a56
     d4a35cd3121aa00d18544bb0ac01c3e1691d618f462c46129271bccf39f7e8ee
 
+-- While using SRPs one can obtain the `--sha256` value for a package
+-- by first setting it to some random value and letting the tooling
+-- tell you what it should be, for example, using `nix develop` will
+-- throw an error with the correct value to use
 source-repository-package
   type: git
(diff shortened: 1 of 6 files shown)


MaximilianAlgehed:
What is the reason the tests are failing now?

aniketd:
They seem to be passing. Perhaps only a review is required? @MaximilianAlgehed 

cc. @Soupstraw

MaximilianAlgehed:
I mean before this change - what's the underlying problem you're fixing here (I'm curious and trying to understand)

cabal.project
```diff
@@ -13,8 +13,7 @@ repository cardano-haskell-packages
 source-repository-package
   type: git
   location: https://github.com/input-output-hk/cardano-ledger-executable-spec.git
-  tag: 81b976c40c2726998a4828220231acf75b3a685e
-  --sha256: sha256-1CiTiLzROoHz8GxVPCouofUFzwzWUclQUq5AthtwPz4=
```
lehins, line 17: You should update the sha256 instead of removing it

lehins (commented):
Looks good.
Let's wait for @Soupstraw to approve it before merging.

libs/cardano-ledger-conformance/src/Test/Cardano/Ledger/Conformance/ExecSpecRule/Conway.hs
```diff
@@ -478,10 +473,13 @@ ratifyStateSpec ::
...
   Specification fn (RatifyState Conway)
-ratifyStateSpec ConwayRatifyExecContext {crecGovActionMap} =
+ratifyStateSpec _ =
   constrained $ \st ->
-    match st $ \_ _ expired _ ->
-      expired `subset_` dom_ (lit crecGovActionMap)
+    match st $ \_ enacted expired _ ->
+      mconcat
+        [ assert $ enacted ==. lit mempty
+        , assert $ expired ==. lit mempty
```
Soupstraw, line 481: These seem to be overconstraining the state. We want to get cases where there are values in `enacted` and `expired`, but right now it's only generating cases where these fields are empty.

Soupstraw (changes requested):
Looks good overall, but I think the generators are overconstrained.

Soupstraw:
@MaximilianAlgehed, the confomance tests for RATIFY didn't pass before, but this PR aims to fix this problem.

libs/cardano-ledger-conformance/src/Test/Cardano/Ledger/Conformance/ExecSpecRule/Conway.hs
aniketd: We discussed this on a call when you were on holiday 😜. Let's discuss in a huddle today?

Soupstraw (approved):

PR #4338, opened 2024-05-17
Add conformance testing for `ENACT`

Soupstraw:
# Description

Adds conformance testing for the `ENACT` rule. Because the environment of `ENACT` is the unit type, I had to add associated types to `ExecSpecRule` that enable overwriting the types that the constrained generator generates as long as there's a way to inject that type into the environment/state/signal that we need in order to run the rule. With this change we should be able to get rid of `ExecContext` at some point when we get to refactoring the conformance tests.

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/libs/cardano-ledger-conformance/src/Test/Cardano/Ledger/Conformance/Orphans.hs
+++ b/libs/cardano-ledger-conformance/src/Test/Cardano/Ledger/Conformance/Orphans.hs
@@ -1,4 +1,6 @@
 {-# LANGUAGE DeriveGeneric #-}
+{-# LANGUAGE FlexibleInstances #-}
+{-# LANGUAGE MultiParamTypeClasses #-}
 {-# LANGUAGE StandaloneDeriving #-}
 {-# OPTIONS_GHC -Wno-orphans #-}
 
@@ -48,6 +50,8 @@ deriving instance Generic RatifyEnv
 deriving instance Generic RatifyState
 
 deriving instance Generic StakeDistrs
+
+deriving instance Generic EnactEnv
 
 deriving instance Ord Tag
 
@@ -171,6 +175,8 @@ instance NFData RatifyEnv
 
 instance NFData RatifyState
 
+instance NFData EnactEnv
+
 instance ToExpr a => ToExpr (HSSet a)
 
 instance ToExpr Credential where
@@ -235,6 +241,8 @@ instance ToExpr StakeDistrs
 instance ToExpr RatifyEnv
 
 instance ToExpr RatifyState
+
+instance ToExpr EnactEnv
 
 instance Default (HSMap k v)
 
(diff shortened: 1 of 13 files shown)


eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Tx.hs
```diff
@@ -164,6 +165,8 @@ data AlonzoTx era = AlonzoTx
   }
   deriving (Generic)
 
+instance Inject (AlonzoTx era) (AlonzoTx era)
```
lehins: Instead of adding a whole bunch of instances like that I think we can just add one `Inject a a`. Could you try my PR: #4350 and see if it'll work for you

libs/cardano-ledger-conformance/src/Test/Cardano/Ledger/Conformance/Orphans.hs
```diff
@@ -305,3 +314,34 @@ instance FixupSpecRep EnactState
...
+instance Inject GovSignal GovSignal
+
+instance Inject Tx Tx
+
+instance Inject GovAction GovAction
+
+-- GovState
+instance Inject [(GovActionID, GovActionState)] [(GovActionID, GovActionState)]
+
+instance Inject [GovSignal] [GovSignal]
```
lehins: All of these and many more can be removed with #4350

libs/cardano-ledger-conformance/src/Test/Cardano/Ledger/Conformance/Spec/Conway.hs
```diff
@@ -18,6 +18,7 @@ spec = describe "Conway conformance tests" $ do
   prop "GOV" $ conformsToImpl @"GOV" @ConwayFn @Conway
   prop "CERT" $ conformsToImpl @"CERT" @ConwayFn @Conway
   prop "RATIFY" $ conformsToImpl @"RATIFY" @ConwayFn @Conway
+  xprop "ENACT" $ conformsToImpl @"ENACT" @ConwayFn @Conway
```
lehins, line 22: Is it not ready to be enabled yet?
```suggestion
  prop "ENACT" $ conformsToImpl @"ENACT" @ConwayFn @Conway
```

lehins (commented):
Could you please try removing all of those `Inject x x` instances in favor of #4350 approach. I think it would be much cleaner that way. (Side note observation: it would have been nice if it was a separate commit that introduced those instances, which would allow to revert all that work much easier)

MaximilianAlgehed (approved):

libs/cardano-ledger-conformance/src/Test/Cardano/Ledger/Conformance/Spec/Conway.hs
Soupstraw: Yeah, it's still failing, I'm just doing the translations in this PR

eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Tx.hs
lehins: #4350 is now merged.

PR #4344, opened 2024-05-21
`constrained-generators`: identify and fix an issue with big bodies to ifElse

MaximilianAlgehed:
# Description

This was identified when trying to add `explain :: [String] -> Pred fn -> Pred fn` but that's a bit more complicated to deal with so we start here.

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/libs/constrained-generators/src/Constrained/Base.hs
+++ b/libs/constrained-generators/src/Constrained/Base.hs
@@ -342,11 +342,11 @@ data Pred (fn :: [Type] -> Type -> Type) where
     -- encoded with `ProdOver` (c.f. `Constrained.Univ`).
     List (Weighted (Binder fn)) as ->
     Pred fn
-  -- if-then-else where the branches depend on the scrutinee.
-  IfElse ::
+  -- monadic-style `when` - if the first argument is False the second
+  -- doesn't apply.
+  When ::
     HasSpec fn Bool =>
     Term fn Bool ->
-    Pred fn ->
     Pred fn ->
     Pred fn
   GenHint ::
@@ -491,9 +491,9 @@ monitorPred env = \case
   Case t bs -> do
     v <- runTerm env t
     runCaseOn v (mapList thing bs) (\x val ps -> monitorPred (extendEnv x val env) ps)
-  IfElse b p q -> do
+  When b p -> do
     v <- runTerm env b
-    if v then monitorPred env p else monitorPred env q
+    if v then monitorPred env p else pure id
   TruePred -> pure id
   FalsePred {} -> pure id
   DependsOn {} -> pure id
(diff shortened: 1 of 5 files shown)


libs/constrained-generators/src/Constrained/Base.hs
```diff
@@ -343,11 +343,10 @@ data Pred (fn :: [Type] -> Type -> Type) where
     List (Weighted (Binder fn)) as ->
     Pred fn
   -- if-then-else where the branches depend on the scrutinee.
-  IfElse ::
+  When ::
     HasSpec fn Bool =>
     Term fn Bool ->
     Pred fn ->
-    Pred fn ->
     Pred fn
```
TimSheard, line 351: This seems to have lost a Pred for the "else" clause,
Perhaps this is more like the Monadic "when" where the missing else clause  silently returns ()
I would assume here, that the missing else clause means No constraints?
Some comments would clarify this.

libs/constrained-generators/src/Constrained/Base.hs
```diff
@@ -4196,12 +4196,14 @@ lit :: Show a => a -> Term fn a
 lit = Lit
 
 ifElse :: (BaseUniverse fn, IsPred p fn, IsPred q fn) => Term fn Bool -> p -> q -> Pred fn
-ifElse (Lit True) (toPred -> p) _ = p
-ifElse (Lit False) _ (toPred -> q) = q
-ifElse b (toPred -> p) (toPred -> q)
-  | FalsePred _ <- p = assert (not_ b) <> q
-  | FalsePred _ <- q = assert b <> p
-  | otherwise = IfElse b p q
+ifElse b p q = whenTrue b p <> whenTrue (not_ b) q
```
TimSheard, line 4200: So the Pred IfElse, is replaced by When, but the function ifElse is defined in terms of When

libs/constrained-generators/src/Constrained/Graph.hs
```diff
@@ -30,7 +31,11 @@ instance Ord node => Monoid (Graph node) where
   mempty = Graph mempty mempty
 
 instance Pretty n => Pretty (Graph n) where
-  pretty gr = vsep [nest 2 $ pretty n <> " <- " <> pretty (Set.toList ns) | (n, ns) <- Map.toList (edges gr)]
+  pretty gr =
+    fold $
+      punctuate
+        hardline
+        [nest 2 $ pretty n <> " <- " <> pretty (Set.toList ns) | (n, ns) <- Map.toList (edges gr)]
```
TimSheard, line 38: Finally, the missing commas return!

TimSheard (approved):
Good job, just add a few comments about When

libs/constrained-generators/src/Constrained/Base.hs
MaximilianAlgehed: This is on purpose. We have to rewrite `When b (Block [p, q])` to `Block [When b p, When b q]` and if there was an else clause things would start blowing up in annoying ways.

libs/constrained-generators/src/Constrained/Base.hs
MaximilianAlgehed: Indeed.

PR #4345, opened 2024-05-21
`constrained-generators`: Add `explanation :: [String] -> Pred fn -> Pred fn`

MaximilianAlgehed:
# Description

By request from @TimSheard

# Checklist

- [x] Self-reviewed the diff


libs/constrained-generators/src/Constrained/Base.hs
```diff
@@ -1790,6 +1817,19 @@ simplifyPred = \case
...
+  -- know what to do.
+  Explain es p ->
+    let p' = simplifyPred p
+     in case simplifyPred p of
+          FalsePred es' -> FalsePred (es ++ es')
+          TruePred {} -> p'
+          GenHint {} -> p'
+          Monitor {} -> p'
+          Explain es' p'' -> Explain (es ++ es') p''
+          p' -> Explain es p'
```
TimSheard: It looks like, you are not getting rid of Explain, just simplifying underneath. What would getting rid of it look like? I am not sure I understand the breadth of this comment.

libs/constrained-generators/src/Constrained/Base.hs
```diff
@@ -4173,6 +4230,13 @@ reify t f body =
     , toPred $ body x
     ]
 
+explanation :: [String] -> Pred fn -> Pred fn
+explanation _ p@DependsOn {} = p
+explanation _ TruePred = TruePred
+explanation es (FalsePred es') = FalsePred (es ++ es')
+explanation es (Assert es' t) = Assert (es ++ es') t
+explanation es p = Explain es p
+
```
TimSheard, line 4379: So I guess this is where it all comes together. propagating explanations so they don't get lost.

TimSheard (approved):
Anything that makes better explanations is Good!

libs/constrained-generators/src/Constrained/Base.hs
MaximilianAlgehed: I've updated the comment to be more clear. The issue is that it would be nice to just drop `Explain es` here and return `simplifyPred p`, but that risks creating an issue where explanations are removed too soon.

libs/constrained-generators/src/Constrained/Base.hs
MaximilianAlgehed: Yes, one of many places

PR #4350, opened 2024-05-22
Add identity instance for `Inject`

lehins:
# Description

The only correct way to inject a type into itself is with `id`, therefore we can safely add a catch all instance for `Inject a a`

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/libs/cardano-ledger-core/src/Cardano/Ledger/BaseTypes.hs
+++ b/libs/cardano-ledger-core/src/Cardano/Ledger/BaseTypes.hs
@@ -846,14 +846,15 @@ instance Default Network where
 
 class Inject t s where
   inject :: t -> s
-  default inject :: t ~ s => t -> s
-  inject = id
 
 instance Inject t () where
   inject = const ()
 
 instance Inject Void s where
   inject = absurd
+
+instance Inject a a where
+  inject = id
 
 -- | Helper function for a common pattern of creating objects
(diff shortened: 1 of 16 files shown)


teodanciu (approved):
Looks good to me

libs/cardano-ledger-core/src/Cardano/Ledger/BaseTypes.hs
```diff
@@ -846,14 +845,9 @@ instance Default Network where
...
 class Inject t s where
   inject :: t -> s
-  default inject :: t ~ s => t -> s
-  inject = id
-
-instance Inject t () where
-  inject = const ()
 
-instance Inject Void s where
-  inject = absurd
```
Soupstraw, line 856: Why are we removing these?

Soupstraw: oh, I guess because of the overlapping instances :/

Soupstraw (approved):

libs/cardano-ledger-core/src/Cardano/Ledger/BaseTypes.hs
Soupstraw: I think it's justified since we are using the identity case more often than initial/terminal type cases, but it's unfortunate that we can't have both

